PR #3367, opened 2023-03-14
CORE-11651 - Rename metrics tags to conform to the micrometer guidelines.

driessamyn:
Some of our metrics tag names do not follow the convention stated on the [MicroMeter documentation](https://micrometer.io/docs/concepts#_naming_meters). 

This PR renames them.

### DIFF
--- a/libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
+++ b/libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
@@ -21,6 +21,8 @@ object CordaMetrics {
             return MeterBuilder(this.metricsName, this.meter)
         }
 
+        // NOTE: please ensure the metric names adhere to the convensions described on https://micrometer.io/docs/concepts#_naming_meters
+
         /**
          * Number of HTTP Requests.
          */
@@ -85,6 +87,12 @@ object CordaMetrics {
         object InboundSessionCount: Metric<SettableGauge>("p2p.session.inbound", CordaMetrics::settableGauge)
     }
 
+    /**
+     * Metric tag names
+     *
+     * NOTE: please ensure the metric names adhere to the convensions described on
+     * https://micrometer.io/docs/concepts#_naming_meters
+     */
     enum class Tag(val value: String) {
         /**
          * Address for which the metric is applicable.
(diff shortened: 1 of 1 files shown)


libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
```diff
@@ -21,6 +21,8 @@ object CordaMetrics {
             return MeterBuilder(this.metricsName, this.meter)
         }
 
+        // NOTE: please ensure the metric names adhere to the convensions described on https://micrometer.io/docs/concepts#_naming_meters
```
davidcurrie:
```suggestion
        // NOTE: please ensure the metric names adhere to the conventions described on https://micrometer.io/docs/concepts#_naming_meters
```

libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
```diff
@@ -93,56 +101,56 @@ object CordaMetrics {
...
          * Flow Id for which the metric is applicable.
          */
-        FlowId("flowId"),
+        FlowId("id"),
 
         /**
          * The status of the operation. Can be used to indicate whether an operation was successful or failed.
          */
-        OperationStatus("operationStatus"),
+        OperationStatus("status"),
```
davidcurrie, line 143: why the short names for these three?

davidcurrie (changes requested):

libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
driessamyn: I just thought the what the id or status was for would be obvious from the context and the short form is maybe more readable. 
But happy to change this back, maybe @LWogan has a view, I think he introduced these.

davidcurrie: It was a genuine question as I can see benefits either way. As you say, given the context, the short names are undoubtedly preferable. Are there scenarios though where you might use these same tags (i.e. the same values) but where the meaning may not be clear? If so, it would be preferable to use the same tag name in both places. I may just be overthinking this though!

driessamyn: > would be preferable to use the same tag name in both places. I may just be overthinking this though!

I think it would even just from a ux point of view as it's less terms to be familiar with.

I didn't change the Enum name as I think the full context there is probably useful from a code perspective.

davidcurrie (approved):

PR #3369, opened 2023-03-14
CORE-11551: Throw HTTP API exception in the tryWithExceptionHandling

yift-r3:
When we are getting an `HttpApiException`  we can throw it as is.

Tested by running the generate CSR API with non-hex key ID.


Tom-Fitzpatrick (approved):

dimosr (approved):
LGTM - can you add testing notes in the PR description please?

PR #3370, opened 2023-03-14
CORE-11065 `utxo_relevant_transaction_state` new columns

lankydan:
Store data in the `custom_representation` and `consumed` columns.

Fill `custom_representation` with temporary data.

The original `consumed` column now stores a timestamp instead of a boolean.

### DIFF
--- a/components/ledger/ledger-persistence/src/integrationTest/kotlin/net/corda/ledger/persistence/utxo/tests/UtxoPersistenceServiceImplTest.kt
+++ b/components/ledger/ledger-persistence/src/integrationTest/kotlin/net/corda/ledger/persistence/utxo/tests/UtxoPersistenceServiceImplTest.kt
@@ -14,10 +14,11 @@ import net.corda.ledger.common.testkit.getPrivacySalt
 import net.corda.ledger.common.testkit.getSignatureWithMetadataExample
 import net.corda.ledger.common.testkit.transactionMetadataExample
 import net.corda.ledger.persistence.consensual.tests.datamodel.field
+import net.corda.ledger.persistence.utxo.StateCustomJson
 import net.corda.ledger.persistence.utxo.UtxoPersistenceService
+import net.corda.ledger.persistence.utxo.UtxoRepository
 import net.corda.ledger.persistence.utxo.UtxoTransactionReader
 import net.corda.ledger.persistence.utxo.impl.UtxoPersistenceServiceImpl
-import net.corda.ledger.persistence.utxo.impl.UtxoRepositoryImpl
 import net.corda.ledger.persistence.utxo.tests.datamodel.UtxoEntityFactory
 import net.corda.ledger.utxo.data.state.StateAndRefImpl
 import net.corda.ledger.utxo.data.transaction.UtxoComponentGroup
@@ -88,6 +89,7 @@ class UtxoPersistenceServiceImplTest {
     private lateinit var digestService: DigestService
     private lateinit var serializationService: SerializationService
     private lateinit var entityManagerFactory: EntityManagerFactory
+    private lateinit var repository: UtxoRepository
     private val emConfig = DbUtils.getEntityManagerConfiguration("ledger_db_for_test")
 
     companion object {
@@ -128,7 +130,7 @@ class UtxoPersistenceServiceImplTest {
             digestService = ctx.getSandboxSingletonService()
             serializationService = ctx.getSerializationService()
             entityManagerFactory = ctx.getEntityManagerFactory()
-            val repository = UtxoRepositoryImpl(digestService, serializationService, wireTransactionFactory)
+            repository = ctx.getSandboxSingletonService()
             persistenceService = UtxoPersistenceServiceImpl(
                 entityManagerFactory,
                 repository,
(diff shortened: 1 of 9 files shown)


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
```diff
@@ -147,6 +148,7 @@ class UtxoPersistenceServiceImpl constructor(
                 UtxoComponentGroup.OUTPUTS.ordinal,
```
vlajos, line 148: It is unrelated to this PR, but should `relevant` be replaced with `visible` in these parts of the code too?
Then I guess we would need a ticket for that.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
```diff
@@ -202,15 +203,17 @@ class UtxoRepositoryImpl @Activate constructor(
 
     override fun markTransactionRelevantStatesConsumed(
         entityManager: EntityManager,
-        stateRefs: List<StateRef>
+        stateRefs: List<StateRef>,
+        timestamp: Instant
```
vlajos, line 207: I would rename this variable to `consumedTimestamp`. Simply `timestamp` is too generic.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
```diff
@@ -331,21 +334,32 @@ class UtxoRepositoryImpl @Activate constructor(
...
+            )
             ON CONFLICT DO NOTHING"""
         )
             .setParameter("transactionId", transactionId)
             .setParameter("groupIndex", groupIndex)
             .setParameter("leafIndex", leafIndex)
             .setParameter("consumed", consumed)
+            .setParameter("custom", stateCustomJson.value)
             .setParameter("createdAt", timestamp)
+            .run { if (consumed) setParameter("consumedTimestamp", timestamp) else this }
```
vlajos: Is it OK that `createdAt` and `consumedTimestamp` can be the same here?

lankydan: I agree, I don't think there is actually a situation in our code where this happens, but I didn't want to do any refactoring as part of my change.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
lankydan: @blsemo @mnesbit @MrMatthewLayton Probably.

blsemo: Yes, it should be.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
blsemo: Interesting question - we have now a couple of operation on the `UtxoRepository` that all take a `timestamp` - and whether that's a created or consumed timestamp should be clear from the function name. So I think that is ok.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
blsemo: I don't think created and consumed timestamp being the same is a problem. Wouldn't that be the case for anything you get as part of a backchain resolution? Or do we send the actual timestamps as part of that?

blsemo (dismissed):
Looks ok to me.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
lankydan: The states aren't relevant/visible to you in that scenario, so you don't go into this method.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
lankydan: Shall I do this now or raise a ticket? It would be a case of renaming the table + a few methods.

blsemo: Can you raise a ticket for it, please? I'd rather not have it mixed up in an unrelated PR.

nkovacsx (dismissed):

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
lankydan: Left as a single timestamp.

vlajos (dismissed):

components/ledger/ledger-persistence/src/integrationTest/kotlin/net/corda/ledger/persistence/utxo/tests/UtxoPersistenceServiceImplTest.kt
```diff
@@ -128,7 +130,7 @@ class UtxoPersistenceServiceImplTest {
             digestService = ctx.getSandboxSingletonService()
             serializationService = ctx.getSerializationService()
             entityManagerFactory = ctx.getEntityManagerFactory()
-            val repository = UtxoRepositoryImpl(digestService, serializationService, wireTransactionFactory)
+            repository = ctx.getSandboxSingletonService()
```
MrMatthewLayton, line 133: What is `ctx`?

MrMatthewLayton (dismissed):

components/ledger/ledger-persistence/src/integrationTest/kotlin/net/corda/ledger/persistence/utxo/tests/UtxoPersistenceServiceImplTest.kt
lankydan: sandbox context

blsemo (approved):
Reapproved

PR #3373, opened 2023-03-14
CORE-10424 Interop synched with latest release/5.0

SaurabhR3:
1. Synched with release/5.0
2. Changed inerop service to work with synched changed


szymonsztuka (commented):
I think we need to do the other way around, merge our changes onto a new branch which was created from the most recent main, and probably the best as single commit.
Here we have history with compilation error which spans across 200 commits.

PR #3376, opened 2023-03-15
CORE-11639: Make `SerializedBytes` an interface on public API

anien:
- Implements `SerializedBytes` interface and updates usages
- Adds `SerializedBytes.unwrap` function to downcast from `SerializedBytes` to `SerializedBytesImpl` in `SerializedBytes` consumer sites since `SerializedBytes` no longer extends `OpaqueBytes` and `ByteSequence`
- Adds `unwrapSerializedBytes` test function for java classes to use `SerializedBytes.unwrap` function

api PR: [#948](https://github.com/corda/corda-api/pull/948)


libs/serialization/serialization-internal/src/main/kotlin/net/corda/serialization/SerializedBytesImpl.kt
```diff
@@ -0,0 +1,14 @@
+package net.corda.serialization
```
kyriathar: I think `SerializedBytesImpl` is not used in this module at all, so maybe should be moved. In non test code, it is only ever getting created by `SerializationOutput` I think so maybe we could consider moving to `:libs:serialization:serialization-amqp` even try make it internal to that module? Then test uses of it outside of this module could be made mocks, since they shouldn't be testing functionality but rather mocking `SerializationService` API.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/SerializationServiceImpl.kt
```diff
@@ -30,7 +31,7 @@ class SerializationServiceImpl(
     override fun <T : Any> deserialize(serializedBytes: SerializedBytes<T>, clazz: Class<T>): T {
         return try {
             AccessController.doPrivileged(PrivilegedExceptionAction {
-                deserializationInput.deserialize(serializedBytes, clazz, context)
+                deserializationInput.deserialize(OpaqueBytes(serializedBytes.bytes), clazz, context)
```
kyriathar: Could we do it differently this here to look a bit nicer? Perhaps we could delegate to other overload?

libs/serialization/serialization-internal/src/main/kotlin/net/corda/serialization/SerializedBytesImpl.kt
anien: Really good spot! I was not sure what would be the best place to put the implementation class to. I will try to move it and use mocks in tests in other modules.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/SerializationServiceImpl.kt
blsemo: I think this looks wrong - the whole point of serialized bytes was that it was a specialization of the generic `ByteSequence` the serializiation service accepts. If you make `SerializedBytes` an interface, then this signature needs to changes (or `ByteSequence` needs to be an interface as well).

kyriathar: Thanks @blsemo, I think we could change it to:

`override fun <T : Any> deserialize(serializedBytes: SerializedBytes<T>, clazz: Class<T>): T =
        deserialize(serializedBytes.bytes, clazz)` ?

Then it means we delegate to the more primitive one which is the one that converts to `ByteSequence`.

Maybe what you mean is that now it might no longer make much sense for `SerializedBytesImpl` to extend `ByteSequence`, since no longer using much `ByteSequence` in terms of `SerializedBytes`, which I think that is right. 

But we still need the `OpaqueBytes` which offers us the `getBytes` to return a copy instead of the underlying bytes which then makes `SerializedBytes` immutable. And `OpaqueBytes` extends `ByteSequence`. But the plan here is to, in the end (after this one gets merged it's only `DigitalSignature` left to extend `OpaqueBytes`) to internalize both `OpaqueBytes` and `ByteSequence` and then we could maybe rework later internally?

kyriathar: `ByteSequence` looks to me like utilities around bytes (okay it also provides some non primitive abstraction for bytes), I feel we should better kill at some point and convert those things to static utilities, and also port its `equals`, `hashCode` to things extending it.

rick-r3: How about on the first line of the method:

`serializedBytes as? SerializedBytesImpl ?: throw IllegalArgumentException("User defined subclasses of SerializedBytes are not permitted")`

and from then on the compiler knows it is `SerializedBytesImpl` really, and implements `OpaqueBytes` and line 33/34 change not needed.  That should also work for other types you want to apply the same change to later. I'm assuming there's no desire for customers to create their own implementations of these interfaces?

You may want to create an internal helper function `unwrap` that codifies the above every time you go from a public API to private API (or implementation) conversion, like this would be.  A lot of classes in Hibernate / JPA have `unwrap` method for this purpose I think, but in Kotlin you can just have an extension function without having to put it on all of your classes that need it.

Then maybe drop access to the bytes on the public API.  

Background / folklore: The `ByteSequence` / `OpaqueBytes` stuff was designed to protect the underlying `ByteArray` from mutation (intended or otherwise) and has methods to wrap as read-only buffers, write to streams etc without having to reveal / copy the underlying bytes.  It's quite confusing code though because we retrofitted `ByteSequence` later so internally we could have mutable forms too and maintaining API compatibility was quite hard I recall, which probably led to some ugliness.

kyriathar: Sure will look into applying, good point to remove `SerializedBytes.getBytes` since it should only be `SerializationService` to consume it. So no use for users to acquire those bytes.

kyriathar: Yes users should not implement `SerializedBytes` (should be a `DoNotImplement`).

libs/serialization/serialization-internal/src/main/kotlin/net/corda/serialization/SerializedBytesImpl.kt
anien: This is now moved. I haven't use mock in other tests modules though as I don't think it's that neccessary.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/SerializationServiceImpl.kt
anien: The `DoNotImplement` annotation is now added to the interface.
The `SerializedBytes.getBytes` is kept after the offline conversation as it's also required elsewhere.
New function `unwrap` is implemented to downcast from `SerializedBytes` to `SerializedBytesImpl`.
In java classes, usual downcasting was used.

kyriathar:
At a later point (because this is all internal) we need to tidy up `DeserializationInput.deserialize` API to be able to take `SerializedBytes` without unwrapping (i.e. make `unwrap` private to `DeserializationInput`). This is recorded with [CORE-11965](https://r3-cev.atlassian.net/browse/CORE-11965)

[CORE-11965]: https://r3-cev.atlassian.net/browse/CORE-11965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

kyriathar (dismissed):

driessamyn (commented):
LGTM.

Will need re-approval after the version number is bumped to beta again.

wzur-r3 (dismissed):
LGTM for Gradle files

kyriathar (approved):

driessamyn (approved):
LGTM

wzur-r3 (approved):
LGTM for Gradle files

PR #3377, opened 2023-03-15
CORE-11523: Introduce additional input parameters for REST worker for TLS

vkolomeyko:
Introduces ability to pass PKCS12 keystore as an optional command line argument to the REST worker(and Combined Worker).

C5 cluster setup with Combined Worker for Smoke Tests been changed to these new optional parameters.
Our K8s cluster E2E tests still running against REST worker with keystore sourced from the classpath as before. There is a separate [follow-up Jira](https://r3-cev.atlassian.net/browse/CORE-11739) to make an improvement.


gradle.properties
```diff
@@ -40,7 +40,7 @@ bouncycastleVersion=1.72
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.0.0.xx-SNAPSHOT to pick up maven local published copy
 #cordaApiVersion=5.0.0.xxx-SNAPSHOT
-cordaApiVersion=5.0.0.716-beta+
+cordaApiVersion=5.0.0.717-alpha-1678906617452
```
vkolomeyko: - [x] revert this change once `corda-api` PR is merged.

libs/rest/ssl-cert-read-impl/src/main/kotlin/net/corda/rest/ssl/impl/SslCertReadServiceImpl.kt
```diff
@@ -46,13 +53,27 @@ class SslCertReadServiceImpl(private val createDirectory: () -> Path) : SslCertR
     }
 
     override fun getOrCreateKeyStoreInfo(config: SmartConfig): KeyStoreInfo {
-        if (keyStoreInfo == null) {
+        var localKeyStoreInfo = keyStoreInfo
+
+        if (localKeyStoreInfo != null) return localKeyStoreInfo
+
+        localKeyStoreInfo = if (config.hasPath(BOOT_REST_TLS_KEYSTORE_FILE_PATH)) {
```
vkolomeyko, line 60: The "meat" of the change.

applications/workers/worker-common/src/main/resources/net/corda/applications/workers/workercommon/boot/corda.boot.json
```diff
@@ -127,7 +127,46 @@
...
+          "properties": {
+            "tls" : {
+              "description": "Transport Layer Security (TLS) params",
+              "properties": {
+                "keystore": {
+                  "description": "Keystore info",
+                  "properties": {
+                    "file": {
+                      "properties": {
+                        "path": {
```
davidcurrie: nit: are you expecting any other properties of a `file`?

libs/rest/ssl-cert-read-impl/src/main/kotlin/net/corda/rest/ssl/impl/SslCertReadServiceImpl.kt
```diff
@@ -46,13 +53,27 @@ class SslCertReadServiceImpl(private val createDirectory: () -> Path) : SslCertR
...
+            val bootKeyStorePath = config.getString(BOOT_REST_TLS_KEYSTORE_FILE_PATH)
+            val keyStorePassword = config.getString(BOOT_REST_TLS_KEYSTORE_PASSWORD)
+            KeyStoreInfo(Path.of(bootKeyStorePath), keyStorePassword)
+        } else {
+            log.warn(
+                "Using default self-signed TLS certificate. To stop seeing this message, please use bootstrap " +
```
davidcurrie: nit: would be good if this message said what the self-signed certificate was being used for

davidcurrie (commented):
I'll let someone else provide the approval. The one observation I'd make is that, as mentioned in the Jira, PKCS12 is not particularly k8s friendly.

vkolomeyko:
> I'll let someone else provide the approval. The one observation I'd make is that, as mentioned in the Jira, PKCS12 is not particularly k8s friendly.

Thank you @davidcurrie. Yes, as discussed on the Jira native K8s way would be to pass `.crt` and `.key` files, but I decided against this approach for now because:
1. This will assume that we only can use self signed certificate vs. certificate path like we do at the moment: `CA->Intermediate Cert(s)->Rest Endpoint Cert`;
2. If we do pass  `.crt` and `.key` files, we would still need to assemble Keystore inside JVM such that Javalin/Jetty can use this data for TLS handshake. To convert  `.crt` and `.key` files into Keystore would involve adding dependency on libraries like BouncyCastle to perform conversion. Such conversion will have to happen every time on REST Worker start-up, which might slow down start-up sequence.

On a balance, at least for now, I decided to add an option to pass PKCS12 keystore to start with. Especially given that it is simpler to do than `.crt` and `.key` files. However, if during work on https://r3-cev.atlassian.net/browse/CORE-11739 I will discover that working with PKCS12 within Helm is too difficult, I may add another option of passing `.crt` and `.key` files in addition to PKCS12 keystore.

applications/workers/worker-common/src/main/resources/net/corda/applications/workers/workercommon/boot/corda.boot.json
vkolomeyko: Unlikely to be any other properties 😃 At the moment it only exist to allow passing as `file.path`. An alternative could `filepath`, but I am not sure which one is better.

libs/rest/ssl-cert-read-impl/src/main/kotlin/net/corda/rest/ssl/impl/SslCertReadServiceImpl.kt
vkolomeyko: Sure - I will change.

applications/workers/worker-common/src/main/resources/net/corda/applications/workers/workercommon/boot/corda.boot.json
davidcurrie: or `file` or `path` or, taking a cue from the Kafka CLI, `location`

vkolomeyko: Let's change it to `path`.

vkolomeyko: Created PR: https://github.com/corda/corda-api/pull/960

davidcurrie:
> Thank you @davidcurrie. Yes, as discussed on the Jira native K8s way would be to pass `.crt` and `.key` files, but I decided against this approach for now because:
> 
> 1. This will assume that we only can use self signed certificate vs. certificate path like we do at the moment: `CA->Intermediate Cert(s)->Rest Endpoint Cert`;

It is usual for the certificate file to contain multiple certificates representing the chain with the root certificate coming last in the file.

> 2. If we do pass  `.crt` and `.key` files, we would still need to assemble Keystore inside JVM such that Javalin/Jetty can use this data for TLS handshake. To convert  `.crt` and `.key` files into Keystore would involve adding dependency on libraries like BouncyCastle to perform conversion. Such conversion will have to happen every time on REST Worker start-up, which might slow down start-up sequence.

I don't think you'll need Bouncy Castle.

> On a balance, at least for now, I decided to add an option to pass PKCS12 keystore to start with. Especially given that it is simpler to do than `.crt` and `.key` files. However, if during work on https://r3-cev.atlassian.net/browse/CORE-11739 I will discover that working with PKCS12 within Helm is too difficult, I may add another option of passing `.crt` and `.key` files in addition to PKCS12 keystore.

I think you'll struggle to make a PKCS12 format certificate directly with Helm. (The Bitnami charts, for example, accept both formats when a secret is provided with the certificate but only PEM when generating it.)

Anyway, I'm not saying don't do PKCS12 - just that I think it'll only get you so far.

wzur-r3 (approved):
LGTM for Jenkinsfile

libs/rest/ssl-cert-read-impl/src/test/kotlin/net/corda/rest/ssl/impl/SslCertReadServiceImplTest.kt
```diff
@@ -1,16 +1,20 @@
...
 import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
 import java.io.File
 import java.nio.file.Files
 import java.nio.file.Path
 
 class SslCertReadServiceImplTest {
 
-    private lateinit var tempDirectoryPath: Path
+    private var tempDirectoryPath: Path? = null
```
josephzunigadaly, line 17: There is a junit `TempDir` annotation that we use in some other tests. If we annotated this field with that, we could delete the create directory and delete directory code.

josephzunigadaly (approved):
Looks good. I left one minor comment but I'm happy either way.

libs/rest/ssl-cert-read-impl/src/test/kotlin/net/corda/rest/ssl/impl/SslCertReadServiceImplTest.kt
vkolomeyko: Yes, sure - I could have used that one as well.

PR #3382, opened 2023-03-15
CORE-11773: move crypto repository code into softhsm impl

dickon:
Move the crypto repository code into SoftHSM where it will be used. This is simpler and possibly faster to bootstrap than referencing a separate OSGi package. I tried using the separate OSGi approach first and got stuck with OSGi errors, and it turned out to be a larger step.

### DIFF
--- a/components/crypto/crypto-persistence/build.gradle
+++ b/components/crypto/crypto-persistence/build.gradle
@@ -28,7 +28,6 @@ dependencies {
     implementation project(":libs:crypto:crypto-core")
     implementation project(":libs:lifecycle:lifecycle")
     implementation project(':libs:virtual-node:virtual-node-info')
-    implementation project(":components:virtual-node:virtual-node-info-read-service")
 
     testImplementation "org.assertj:assertj-core:$assertjVersion"
     testImplementation "org.jetbrains.kotlin:kotlin-test:$kotlinVersion"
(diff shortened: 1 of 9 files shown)


components/crypto/crypto-persistence/build.gradle
```diff
@@ -21,14 +22,14 @@ dependencies {
...
-    implementation project(":components:db:db-connection-manager")
     implementation project(":libs:layered-property-map")
     implementation project(":libs:configuration:configuration-core")
     implementation project(":libs:crypto:crypto-config-impl")
     implementation project(":libs:crypto:crypto-core")
     implementation project(":libs:lifecycle:lifecycle")
     implementation project(':libs:virtual-node:virtual-node-info')
-    implementation project(":components:virtual-node:virtual-node-info-read-service")
+    implementation project(':components:db:db-connection-manager')
+    implementation project(':components:virtual-node:virtual-node-info-read-service')
```
fowlerrr: Total pedantry, but I tend to put these in alphabetical order.  Simply so I know where to look when I'm looking.

dickon: done

fowlerrr (approved):
Simple move of the files before other changes go in and that LGTM.

PR #3386, opened 2023-03-15
CORE-9565 Implement findTransactionWithStatus and use it in Back Chain resolution

vlajos:
CORE-9565 Implement findTransactionWithStatus which returns the transaction's status if that does not match with the expected and use it in Back Chain resolution

### DIFF
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
@@ -12,6 +12,7 @@ import net.corda.utilities.serialization.deserialize
 import net.corda.utilities.time.Clock
 import net.corda.v5.application.crypto.DigestService
 import net.corda.v5.application.serialization.SerializationService
+import net.corda.v5.base.exceptions.CordaRuntimeException
 import net.corda.v5.crypto.DigestAlgorithmName
 import net.corda.v5.ledger.common.transaction.CordaPackageSummary
 import net.corda.v5.ledger.utxo.ContractState
@@ -35,6 +36,7 @@ class UtxoPersistenceServiceImpl constructor(
             val status = repository.findTransactionStatus(em, id)
             if (status == transactionStatus.value) {
                 repository.findTransaction(em, id)
+                    ?: throw CordaRuntimeException("Transaction $id in status $status has disappeared from the database")
             } else {
                 null
             } to status
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
@@ -2,6 +2,7 @@ package net.corda.ledger.persistence.utxo.impl
 
 import net.corda.ledger.common.data.transaction.SignedTransactionContainer
 import net.corda.ledger.common.data.transaction.TransactionStatus
+import net.corda.ledger.persistence.common.InconsistentLedgerStateException
 import net.corda.ledger.persistence.utxo.CustomRepresentation
 import net.corda.ledger.persistence.utxo.UtxoPersistenceService
 import net.corda.ledger.persistence.utxo.UtxoRepository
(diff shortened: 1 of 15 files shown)


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
```diff
@@ -27,14 +27,17 @@ class UtxoPersistenceServiceImpl constructor(
     private val utcClock: Clock
 ) : UtxoPersistenceService {
 
-    override fun findTransaction(id: String, transactionStatus: TransactionStatus): SignedTransactionContainer? {
+    override fun findTransaction(
+        id: String,
+        transactionStatus: TransactionStatus
+    ): Pair<SignedTransactionContainer?, String?> {
```
lankydan, line 34: I think this should return `Pair<SignedTransactionContainer, String>?`. Where its `null` if there is no matching status and if there is no transaction that matches an existing status record, then an error should be thrown. This should be a fatal error that goes back to the flow imo. We shouldn't be trying to handle this case here because its a genuine error scenario.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainVerifierImpl.kt
```diff
@@ -39,24 +39,47 @@ class TransactionBackchainVerifierImpl @Activate constructor(
...
-                            "${e.message}"
+                    "Backchain resolution of $initialTransactionIds - Verification of transaction $transactionId failed. " +
+                            "The transaction is already invalid."
                 )
                 return false
             }
-            utxoLedgerPersistenceService.updateStatus(transactionId, VERIFIED)
-            log.trace { "Backchain resolution of $initialTransactionIds - Updated status of transaction $transactionId to verified" }
+            if (status == VERIFIED) {
+                log.trace { "Backchain resolution of $initialTransactionIds - The transaction is already verified $transactionId." }
```
lankydan: Can we have the message like `"Backchain resolution of $initialTransactionIds - Transaction $transactionId is already verified, skipping verification"`

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainVerifierImpl.kt
```diff
@@ -39,24 +39,47 @@ class TransactionBackchainVerifierImpl @Activate constructor(
...
+            } else {
+                if (status == null || transaction == null) {
```
lankydan: This will change if my previous comment is addressed.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainVerifierImpl.kt
```diff
@@ -39,24 +39,47 @@ class TransactionBackchainVerifierImpl @Activate constructor(
...
-                    "Backchain resolution of $initialTransactionIds - Verification of transaction $transactionId failed, message: " +
-                            "${e.message}"
+                    "Backchain resolution of $initialTransactionIds - Verification of transaction $transactionId failed. " +
+                            "The transaction is already invalid."
                 )
                 return false
             }
-            utxoLedgerPersistenceService.updateStatus(transactionId, VERIFIED)
-            log.trace { "Backchain resolution of $initialTransactionIds - Updated status of transaction $transactionId to verified" }
+            if (status == VERIFIED) {
```
lankydan: We should use `if else` or a `when` block here because each branch cannot happen if any of the others are executed.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
blsemo: I think what Lajos did here is:
- if we find a tx of the requested status, we return the TX and the status
- if we find a tx that has a different status, we just return the status. That includes returning a status of null if no trace of the TX can be found.

lankydan (approved):

PR #3387, opened 2023-03-16
CORE-7994 Replace key with key id in digital signature packet

kyriathar:
WIP

Still resolving key id to key mappings.

### DIFF
--- a/components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/TransactionSignatureServiceImpl.kt
+++ b/components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/TransactionSignatureServiceImpl.kt
@@ -1,5 +1,6 @@
 package net.corda.ledger.common.flow.impl.transaction
 
+import net.corda.crypto.cipher.suite.KeyEncodingService
 import net.corda.crypto.cipher.suite.merkle.MerkleTreeProvider
 import net.corda.crypto.core.SecureHashImpl
 import net.corda.ledger.common.data.transaction.BATCH_MERKLE_TREE_DIGEST_OPTIONS_LEAF_PREFIX_B64_KEY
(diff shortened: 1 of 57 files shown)


components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/TransactionSignatureServiceImpl.kt
```diff
@@ -153,11 +167,23 @@ class TransactionSignatureServiceImpl @Activate constructor(
...
+            keyUsedToSign,
             signatureSpec
         )
     }
 
+    @Suspendable
+    private fun MemberLookup.lookup(keyId: SecureHash): PublicKey {
+        val digestAlgorithmOfKeyId = keyId.algorithm
+        val knownKeysByKeyIds = lookup().flatMap {
+            it.ledgerKeys
```
vlajos, line 179: `verifySignature(...)` might be called against the notary signature as well.
And that one is not created with a ledgerKey, but something else.
(*)

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityBaseV1.kt
```diff
@@ -153,4 +161,21 @@ abstract class UtxoFinalityBaseV1 : SubFlow<UtxoSignedTransaction> {
             verifySignature(initialTransaction, it, sessionToNotify)
         }
     }
+
+    @Suspendable
+    private fun MemberLookup.lookup(keyId: SecureHash): PublicKey {
+        val digestAlgorithmOfKeyId = keyId.algorithm
+        val knownKeysByKeyIds = lookup().flatMap {
+            it.ledgerKeys
```
vlajos, line 169: This lookup is used only in the verifyAndAddNotarySignature(), so this one needs only the notary keys, and the ledgerkeys are not needed at all. (A bit like (*) above.)

components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/TransactionSignatureServiceImpl.kt
```diff
@@ -153,11 +167,23 @@ class TransactionSignatureServiceImpl @Activate constructor(
...
+        }.associateBy {
+            it.fullIdHash(keyEncodingService, digestService, digestAlgorithmOfKeyId)
+        }
+
+        return knownKeysByKeyIds[keyId] ?: error("Member for consensual signature not found")
+    }
```
vlajos, line 185: In the caller places of `TransactionSignature.verifySignature(...)`, we have a list of public keys. (The signatories in signed transactions and finality flows.)
So in theory we could use those public keys to lookup these keyIds, which (I suppose) would be much faster than Memberlookup.

But
1. those lists may have composite keys. (notary key)
2. those lists are not necessarily exhaustive. It is possible that a transaction has a signature from non-participants.

So I expect that on a bit longer term we'll be able to implement similar functions:
```
val publicKey = signedTransaction.getSignatoryByKeyId(keyId)
transactionSignatureService.verifySignature(transaction, signature, publicKey)
```
But that would be a new API.
And I expect that we won't be able to remove the old one due to the requirement to support verifying non-participant's signatures.
So I think it is fine to carry-on with the member lookup based solution for now. And later the optimization can be added.

@mnesbit @blsemo could you please check my above thinking?

kyriathar:
Closing as [#3400](https://github.com/corda/corda-runtime-os/pull/3400) was used to address this work.

PR #3390, opened 2023-03-16
CORE-11751: Enable crypto-utils publishing

mcgovc:
Issue flagged testing CSDE as `crypto-utils` dependency is required in `simulator`.


mcgovc:
Raised #3399 as this should be targeted to the release branch and not Beta2

PR #3394, opened 2023-03-16
CORE-11278: Enable session key rotation

yift-r3:
Enable session key rotation with a preferred session key and alternative session keys. The link manager will use the preferred key to sign messages and will try to use all the keys to validate the message.

The API pull request can be found in https://github.com/corda/corda-api/pull/943


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -787,18 +784,19 @@ internal class SessionManagerImpl(
         }
 
         val tenantId = ourIdentityInfo.holdingIdentity.shortHash.value
+        val ourIdentitySessionKey = ourIdentityInfo.preferredSessionKey
```
dimosr: I think in that case, we don't want to use the preferred key. We want to look amongst the list of preferred + alternative and find the one whose hash matches the one sent by the peer.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -826,48 +824,42 @@ internal class SessionManagerImpl(
...
-        } catch (exception: WrongPublicKeyHashException) {
-            logger.error("The message was discarded. ${exception.message}")
-            null
-        } catch (exception: InvalidHandshakeMessageException) {
-            logger.validationFailedWarning(message::class.java.simpleName, message.header.sessionId, exception.message)
-            null
-        } catch (exception: InvalidPeerCertificate) {
-            logger.validationFailedWarning(message::class.java.simpleName, message.header.sessionId, exception.message)
-            null
+        val exceptions = peer.sessionInitiationKeys.map { sessionInitiationKey ->
```
dimosr: I think we have the information necessary at this point that indicates which key the peer is using, since they sent the key hash in the initiator hello message. So, we could avoid trying all of them, but needs some re-work to keep track of it as it's not directly available at the moment here.

One way I've thought about is to expose it from `AuthenticationProtocolResponder`. And then here, we'd need to perform a member lookup to find the member using that key hash -> get all session initiation keys from their `MemberInfo` -> filter those to find the one that has a matching key hash.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -826,48 +824,42 @@ internal class SessionManagerImpl(
...
-        } catch (exception: InvalidHandshakeResponderKeyHash) {
-            logger.validationFailedWarning(message::class.java.simpleName, message.header.sessionId, exception.message)
-            false
-        } catch (exception: InvalidHandshakeMessageException) {
-            logger.validationFailedWarning(message::class.java.simpleName, message.header.sessionId, exception.message)
-            false
-        } catch (exception: InvalidPeerCertificate) {
-            logger.validationFailedWarning(message::class.java.simpleName, message.header.sessionId, exception.message)
-            false
+        val messages = memberInfo.sessionInitiationKeys.map { sessionInitiationKey ->
```
dimosr: Similarly here, we could avoid the loop. We're specifying to the other peer which key to use previously (as part of the InitiatorHandshakeMessage). So, we could keep a reference to it in `AuthenticationProtocolInitiator` and expose it so that we can read it back here.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/hosting/HostingMapListener.kt
```diff
@@ -9,6 +9,11 @@ interface HostingMapListener {
         val holdingIdentity: HoldingIdentity,
         val tlsCertificates: List<PemCertificate>,
         val tlsTenantId: String,
+        val preferredSessionKey: SessionKey,
+        val alternativeSessionKeys: List<SessionKey>,
+    )
+
+    data class SessionKey(
```
dimosr:
```suggestion
    data class SessionKeyAndCert(
```

dimosr: Also, now that we're at it, should we also rename `sessionCertificates` below to `sessionCertificateChain`? It will be a bit less ambiguous.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/hosting/LinkManagerHostingMapImpl.kt
```diff
@@ -115,26 +117,43 @@ internal class LinkManagerHostingMapImpl(
...
     }
 
     private fun addEntry(entry: HostedIdentityEntry) {
         val info = HostingMapListener.IdentityInfo(
             holdingIdentity = entry.holdingIdentity.toCorda(),
             tlsCertificates = entry.tlsCertificates,
             tlsTenantId = entry.tlsTenantId,
-            sessionPublicKey = publicKeyReader.loadPublicKey(entry.sessionPublicKey),
-            sessionCertificates = entry.sessionCertificates
+            preferredSessionKey = entry.prefferedSessionKey.toData(),
```
dimosr:
```suggestion
            preferredSessionKey = entry.preferredSessionKey.toData(),
```

components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/CertificatesClientImpl.kt
```diff
@@ -118,15 +118,15 @@ class CertificatesClientImpl @Activate constructor(
         holdingIdentityShortHash: ShortHash,
         p2pTlsCertificateChainAlias: String,
         useClusterLevelTlsCertificateAndKey: Boolean,
-        sessionKeyId: ShortHash?,
-        sessionCertificateChainAlias: String?
+        preferredSessionKey: CertificatesClient.SessionKey?,
```
dimosr: Shouldn't we always have a preferred key declared, i.e. have this non-nullable?

components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/HostedIdentityEntryFactory.kt
```diff
@@ -112,7 +119,34 @@ internal class HostedIdentityEntryFactory(
...
+                .setSessionCertificates(certificates)
+                .build()
+        } else {
+            buildHostedIdentitySessionKey(
+                preferredSessionKey,
+                holdingIdentityShortHash,
+                nodeInfo,
+                policy
+            )
+        }
```
dimosr: related to my other comment - I don't think this is needed given there must always be a preferred key declared.

components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/HostedIdentityEntryFactory.kt
```diff
@@ -47,28 +50,33 @@ internal class HostedIdentityEntryFactory(
...
+            tenantId = tenantId,
+            0,
+            1,
+            CryptoKeyOrderBy.NONE,
+            mapOf(CryptoConsts.SigningKeyFilters.CATEGORY_FILTER to CryptoConsts.Categories.SESSION_INIT),
+        ).toPem(tenantId)
+    }
+    private fun Collection<CryptoSigningKey>.toPem(tenantId: String): String {
+        val key = this.firstOrNull()
+            ?: throw CertificatesResourceNotFoundException("Can not find session key for $tenantId")
```
dimosr: minor: I think it would be a bit easier to follow if the error handling was done outside this method right after retrieving the key from the lookup. And then the method could also be adjusted to operate on a single key, instead of a collection.

components/membership/certificates-client/src/main/kotlin/net/corda/membership/certificate/client/CertificatesClient.kt
```diff
@@ -8,6 +8,18 @@ import net.corda.lifecycle.Lifecycle
...
 interface CertificatesClient : Lifecycle, DbCertificateClient {
 
+    /**
+     * A session key and certificate.
+     *
+     * @param sessionKeyId The session key ID.
+     * @param sessionCertificateChainAlias The certificate chain alias of the Session Key.
+     *   Should be null if no PKI is used for sessions.
+     */
+    data class SessionKey(
```
dimosr:
```suggestion
    data class SessionKeyAndCert(
```

components/membership/certificates-client/src/main/kotlin/net/corda/membership/certificate/client/CertificatesClient.kt
```diff
@@ -16,16 +28,16 @@ interface CertificatesClient : Lifecycle, DbCertificateClient {
      * @param p2pTlsCertificateChainAlias The certificates chain alias.
      * @param useClusterLevelTlsCertificateAndKey Should we use the P2P cluster level TLS certificate type and P2P key or
      *   the virtual node certificate and key.
-     * @param sessionKeyId The session key ID (will use the first one if null).
-     * @param sessionCertificateChainAlias The certificate chain alias of the Session Key. Should be null if no PKI is used for sessions.
+     * @param preferredSessionKey The preferred session keys. If null the first session key will be used.
```
dimosr: I think it's better if we require an explicit declaration of preferred key, instead of trying to make automated guesses. Leaves less room for accidental errors caused by us and given the criticality of it good to have the operator thinking well about it.

components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/converter/PartyConverter.kt
```diff
@@ -29,9 +30,10 @@ class PartyConverter @Activate constructor(
         val name = context.value(NAME)?.let { MemberX500Name.parse(it) }
             ?: throw ValueNotFoundException("'$NAME' is null or missing")
 
-        val key = context.value(SESSION_KEY)?.let { keyEncodingService.decodePublicKey(it) }
+        val key = context.value(SESSION_KEYS)
+            ?: context.value(SESSION_KEY)
```
dimosr: we can remove the old attribute (`SESSION_KEY`) now?

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/NetworkRestResourceImpl.kt
```diff
@@ -41,13 +42,22 @@ class NetworkRestResourceImpl @Activate constructor(
...
         val operation = "set up locally hosted identities"
+        val preferredSessionKeys = request.sessionKeys.filter {
+            it.preferred
+        }
+        val alternativeSessionKeys = request.sessionKeys.filter {
+            !it.preferred
+        }
+        if (preferredSessionKeys.size > 1) {
+            throw BadRequestException("Can not have more than one preferred  session key.")
+        }
```
dimosr: Related to my other comment - I think we need to enforce exactly one preferred key here.

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/types/request/HostedIdentitySetupRequest.kt
```diff
@@ -6,12 +6,23 @@ package net.corda.membership.rest.v1.types.request
...
- * @param sessionKeyId The session key identifier (will use the first one if null).
- * @param sessionCertificateChainAlias The certificate chain alias of the Session Key. Should be null if no PKI is used for sessions.
+ * @param sessionKeys The list of session keys and certificates. If empty the first session key will be used.
  */
 data class HostedIdentitySetupRequest(
     val p2pTlsCertificateChainAlias: String,
     val useClusterLevelTlsCertificateAndKey: Boolean?,
-    val sessionKeyId: String?,
-    val sessionCertificateChainAlias: String? = null,
+    val sessionKeys: List<HostedIdentitySessionKey> = emptyList(),
```
dimosr:
```suggestion
    val sessionKeysAndCerts: List<HostedIdentitySessionKeyAndCert>,
```
I can't think of a scenario where a user would want to specify no keys at all, so empty list doesn't look like a great default if that's the case.

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/types/request/HostedIdentitySetupRequest.kt
```diff
@@ -6,12 +6,23 @@ package net.corda.membership.rest.v1.types.request
...
 )
+
+/**
+ * @param sessionKeyId The session key identifier.
+ * @param sessionCertificateChainAlias The certificate chain alias of the Session Key.
+ *   Should be null if no PKI is used for sessions.
+ * @param preferred true if this key is the preferred key. If no key is preferred the
+ *    first one will be assumed as preferred.
+ */
+data class HostedIdentitySessionKey(
```
dimosr:
```suggestion
 * @param preferred true if this key is the preferred key. 
 */
data class HostedIdentitySessionKeyAndCert(
```

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/MemberInfoExtension.kt
```diff
@@ -38,13 +38,15 @@ class MemberInfoExtension {
 
         /** Key name for party property. */
         const val PARTY_NAME = "corda.name"
-        const val PARTY_SESSION_KEY = "corda.session.key"
+        const val SESSION_KEYS = "corda.session.keys"
+        const val PARTY_SESSION_KEYS = "corda.session.keys.%s"
```
dimosr: would it be better to do `$SESSION_KEYS.%s` for this one and the others below?

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/MemberInfoExtension.kt
```diff
@@ -38,13 +38,15 @@ class MemberInfoExtension {
...
+        const val PARTY_SESSION_KEYS_PEM = "corda.session.keys.%s.pem"
```
dimosr: curious - what is there a reason we introduced the `.pem` suffix instead of keeping it as it was before?

tools/plugins/mgm/src/main/kotlin/net/corda/cli/plugins/mgm/BaseOnboard.kt
```diff
@@ -335,15 +350,19 @@ abstract class BaseOnboard : Runnable {
...
-                throw OnboardException("Registration has been declined.")
-            } else {
-                println("Status of $x500Name registration is $registrationStatus")
-                Thread.sleep(400)
+            when (registrationStatus) {
+                "APPROVED" -> {
+                    return
+                }
+                "DECLINED", "INVALID" -> {
+                    throw OnboardException("Registration has been $registrationStatus.")
```
dimosr:
```suggestion
                    throw OnboardException("Status of registration is $registrationStatus.")
```
otherwise we'd end up with "Registration has been INVALID"

tools/plugins/mgm/src/main/kotlin/net/corda/cli/plugins/mgm/BaseOnboard.kt
```diff
@@ -99,6 +100,12 @@ abstract class BaseOnboard : Runnable {
     )
     var mtls: Boolean = false
 
+    @Option(
+        names = ["--session_keys_count"],
```
dimosr: Do we really need to add this? 

Given this is only designed to be used for rotation for now, it might be better to avoid adding it here to keep things simple. And if we decide to provide support for rotation in the plugin, we can do it properly at that point in time.

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/types/request/HostedIdentitySetupRequest.kt
dimosr: We should also validate that btw, now that we have a list as I don't think it can be validated as part of the JSON schema.

dimosr (commented):
I'll need to do another scan, since there have been a few commits since I last looked. But, sharing a few comments in the meanwhile (I tried to proactively delete the ones that have already been resolved as part of the recent commits).

dimosr:
Also, given it's an API-breaking change can you kick off a [build of the e2e tests](https://github.com/corda/corda-e2e-tests) with the version produced by this PR to confirm nothing will break there too, and link the build here?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/hosting/LinkManagerHostingMapImpl.kt
```diff
@@ -115,26 +117,40 @@ internal class LinkManagerHostingMapImpl(
...
-        publicHashToIdentityInfo[entry.toGroupIdWithPublicKeyHash()] = info
+        entry.toGroupIdWithPublicKeyHash().forEach {
+            publicHashToIdentityInfo[it] = info
+        }
         listeners.forEach {
             it.identityAdded(info)
         }
     }
+
+    private fun HostedIdentitySessionKey.toData(): HostingMapListener.SessionKey {
```
dimosr:
```suggestion
    private fun HostedIdentitySessionKey.toData(): HostingMapListener.SessionKey {
```
```suggestion
    private fun HostedIdentitySessionKey.toCorda(): HostingMapListener.SessionKey {
```
minor, but for consistency with other converter functions.

components/membership/certificates-client/src/main/kotlin/net/corda/membership/certificate/client/CertificatesClient.kt
dimosr: If we want to optimise the experience for the happy path case (single key), which will apply most of the time, by automatically inferring the preferred key if there's only one specified I would be fine with it. 

But, when there are multiple keys specified, the main scenario we have at the moment is rotation and we shouldn't be making automated guesses at that point. The operator should be explicitly asked to make that decision.

yift-r3:
e2e tests in https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda5-e2e-tests/detail/release%2F5.0/262/pipeline

components/membership/membership-p2p/src/main/kotlin/net/corda/membership/p2p/helpers/SignerFactory.kt
```diff
@@ -12,7 +12,7 @@ class SignerFactory(
     ): Signer =
         Signer(
             mgm.id,
-            mgm.sessionInitiationKey,
+            mgm.sessionInitiationKeys.first(),
```
dimosr, line 16: Here, we should be using the preferred key instead. Given we don't have locally hosted identity data here though, it might be a more extensive change.

So, we can create a ticket to change that as a follow-up, but I think it needs to be addressed before GA if it's technically possible to register multiple keys.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/mgm/MGMRegistrationContextValidator.kt
```diff
@@ -87,6 +86,10 @@ internal class MGMRegistrationContextValidator(
             require(isNotEmpty()) { "No TLS trust store was provided." }
             require(orderVerifier.isOrdered(this, 4)) { "Provided TLS trust stores are incorrectly numbered." }
         }
+        context.keys.filter { SESSION_KEY_IDS.format("[0-9]+").toRegex().matches(it) }.apply {
+            require(isNotEmpty()) { "No session was provided." }
```
dimosr:
```suggestion
            require(isNotEmpty()) { "No session key was provided." }
```

dimosr (commented):
A few more comments from the follow-up scan on the latest

components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/converter/PartyConverter.kt
yift-r3: Party is dead - https://github.com/corda/corda-runtime-os/pull/3416

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/MemberInfoExtension.kt
yift-r3: I'm not sure why, but in order to make the key converter work with a list the context needs to have a `pem`.

tools/plugins/mgm/src/main/kotlin/net/corda/cli/plugins/mgm/BaseOnboard.kt
yift-r3: I will remove it (I'm keeping it, for now, to allow myself to test my changes)

components/membership/membership-p2p/src/main/kotlin/net/corda/membership/p2p/helpers/SignerFactory.kt
yift-r3: https://r3-cev.atlassian.net/browse/CORE-11935 created

components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/CertificatesClientImpl.kt
yift-r3: It's nullable for an empty list.

components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/HostedIdentityEntryFactory.kt
yift-r3: We have a case when the user has only one session key, in which case he can commit the session key, and we will choose the only one.

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/types/request/HostedIdentitySetupRequest.kt
yift-r3: The validation seems to work.

components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/CertificatesClientImpl.kt
dimosr: But, is there a valid use case where one would have an empty list of keys? Or should we prevent that, and then make this non-nullable?

components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/HostedIdentityEntryFactory.kt
dimosr: In the same spirit as in other comments, I think we should be requiring the operator to explicitly specify at least one key in the network setup.

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/types/request/HostedIdentitySetupRequest.kt
dimosr: which validation works? I referred to verifying the list contains at least one element.

yift-r3: We have this validation in the rest implementation.

components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/HostedIdentityEntryFactory.kt
yift-r3: I've changed it. Now it's either one ore the list is empty (for the case that you have one key).

libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/WrongPublicKeyHashException.kt
```diff
@@ -3,9 +3,9 @@ package net.corda.p2p.crypto.protocol.api
 import net.corda.v5.base.exceptions.CordaRuntimeException
 import java.util.*
 
-class WrongPublicKeyHashException(expectedHash: ByteArray?, actualHash: ByteArray):
+class WrongPublicKeyHashException(expectedHash: ByteArray?, actualHashs: Collection<ByteArray>):
```
dimosr:
```suggestion
class WrongPublicKeyHashException(expectedHash: ByteArray?, actualHashes: Collection<ByteArray>):
```

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/NetworkRestResourceImpl.kt
```diff
@@ -41,13 +42,31 @@ class NetworkRestResourceImpl @Activate constructor(
...
+            if (preferredSessionKeys.isEmpty()) {
+                throw BadRequestException("No preferred session key was selected.")
+            }
+            if (preferredSessionKeys.size > 1) {
+                throw BadRequestException("Can not have more than one preferred session key.")
+            }
+            val preferredSessionKey = preferredSessionKeys.first()
+            val alternativeSessionKeys = request.sessionKeysAndCertificates.filter {
+                it != preferredSessionKey && !it.preferred
+            }
```
dimosr:
```suggestion
            val alternativeSessionKeys = request.sessionKeysAndCertificates.filter { !it.preferred }
```
wouldn't this be sufficient?

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/types/request/HostedIdentitySetupRequest.kt
dimosr: at which place exactly? Looking at `NetworkRestResourceImpl.setupHostedIdentities()` I can see we are accepting empty list from the user without returning an error.

The conversation [here](https://github.com/corda/corda-runtime-os/pull/3394#discussion_r1140329312) is about the same thing, I believe. I am not sure I got my point across there. I think we shouldn't be allowing the operator to specify empty lists (even when there's a single session key in the cluster). The REST endpoint should mandate at least one session initiation key be specified, given the sensitivity of these operations.

components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/HostedIdentityEntryFactory.kt
yift-r3: Currently, the API support no session key, in that case, we will use the first session init key that we have in the crypto database (and if there are none we will throw an exception). Do we want to remove this behaviour?

dimosr: Yes, I'd prefer to always require at least one session key specified by the user. Similar to the current behaviour of that endpoint (by current I mean without these changes).

yift-r3: What I'm saying is that currently, we do allow null session key (https://github.com/corda/corda-runtime-os/blob/release/os/5.0/components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/types/request/HostedIdentitySetupRequest.kt)

dimosr: ah, I see.. I must have gotten confused with the registration endpoint.

In any case, I think it's safer to always require a session key to be specified.

dimosr (commented):
LGTM - just need to revert the plugin changes and happy to approve after that

dimosr (dismissed):

vkolomeyko (dismissed):

dimosr (dismissed):

applications/workers/release/combined-worker/build.gradle
```diff
@@ -31,6 +31,7 @@ quasar {
             'liquibase**',
             'net.bytebuddy**',
             'net.corda.membership**',
+            'net.corda.p2p**',
```
wzur-r3, line 34: I would feel much better if @chrisr3 gave his stamp of approval here...

vkolomeyko (approved):
LGTM

dimosr (approved):

wzur-r3 (approved):
LGTM

PR #3396, opened 2023-03-16
CORE-11744 Version transaction builder flows

lankydan:


vlajos (dismissed):
LGTM.
It does not have smoke tests in the runtime-os repo, but the e2e test repo has a test against tx builder sending.
So probably it would worth checking if that breaks. Or running that in advance.

mnesbit (approved):
OK

PR #3397, opened 2023-03-16
CORE-9981 

pmck91:
Ledger needs access to the currently running CPI’s information to be able to save it into the metadata of transactions and digital signatures. (name, version, signerSummaryHash, fileChecksum)
Ideally this should be populated into the context of the flow when the flow gets started and made accessible in the flow.


components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
```diff
@@ -113,4 +132,22 @@ class FlowRunnerImpl @Activate constructor(
...
+
+    private fun getPropertiesWithCpiMetadata(
+        holdingIdentity: HoldingIdentity,
+        contextProperties: KeyValuePairList
+    ): KeyValuePairList {
+        return virtualNodeInfoReadService.get(holdingIdentity).let {
+            if (it == null) contextProperties
+            else {
+                cpiInfoReadService.get(it.cpiIdentifier)?.let { metadata ->
+                    contextProperties.put("corda.cpiName", metadata.cpiId.name)
```
LWogan: I don't think we can use KeyValuePairList like this. Im new enough to it myself but I think you need to build 
KeyValuePair objects and add them to a list "List<.KeyValuePair> items;" which is a field in KeyValuePairList. The put() method used here is some method inherited by some avro super type

Simon added a corda helper class to make life easier called KeyValueStore which has an api closer to a Map and has a .avro method to convert to a KeyValuePairList. Example::
https://github.com/corda/corda-runtime-os/blob/65e55b30bdb85e32f3b362b6d5fadd20ca0604da/components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/context/FlowServiceTestContext.kt#L267

LWogan: We should probably store the keys in a constants file somewhere accessible by  flow code and flow-engine code..... maybe in the flow-api project, net.corda.flow.state package, not sure.

LWogan (commented):
The location of the changes and the values set make sense to me.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
```diff
@@ -118,4 +138,27 @@ class FlowRunnerImpl @Activate constructor(
...
         return flowFiberFactory.createAndResumeFlowFiber(fiberContext, flowContinuation)
     }
+
+    @Suppress("NestedBlockDepth")
+    private fun getPropertiesWithCpiMetadata(
+        holdingIdentity: HoldingIdentity,
+        contextProperties: KeyValuePairList
+    ): KeyValuePairList {
+        return virtualNodeInfoReadService.get(holdingIdentity).let {
+            if (it == null) contextProperties
```
simon-johnson-r3: This null check on `it` is not so Kotlin like, it would be better as:
```
return virtualNodeInfoReadService.get(holdingIdentity)?.let {
...
} ?: contextProperties
```
Then you don't need the double return in the let, or the one on line 147, or the return type so I think:
```
private fun getPropertiesWithCpiMetadata(
        ...
    ) = virtualNodeInfoReadService.get(holdingIdentity)?.let {
          KeyValueStore().apply {
               ...
          }.avro
        } ?: contextProperties
```
would do it, and then you might be able to remove the @Suppress too

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
```diff
@@ -118,4 +138,27 @@ class FlowRunnerImpl @Activate constructor(
...
+            else {
+                return KeyValueStore().apply {
+                    cpiInfoReadService.get(it.cpiIdentifier)?.let { metadata ->
+                        this["corda.cpiName"] = metadata.cpiId.name
+                        this["corda.cpiVersion"] = metadata.version.toString()
+                        this["corda.cpiSignerSummaryHash"] = metadata.cpiId.signerSummaryHash.toHexString()
+                        this["corda.cpiFileChecksum"] = metadata.fileChecksum.toHexString()
+                        contextProperties.items.forEach { prop ->
```
simon-johnson-r3: should this part about contextProperties be out of the `cpiInfoReadService.get(it.cpiIdentifier)?.let` ? Otherwise you'll lose them unnecessarily if that is null.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
```diff
@@ -440,6 +440,19 @@ class FlowTests {
...
+    fun `CPI metadata is available in a flow`() {
+        val requestBody = RpcSmokeTestInput().apply {
+            command = "get_cpi_metadata"
+            data = emptyMap()
+        }
+
+        val requestId = startRpcFlow(bobHoldingId, requestBody)
+
+        val result = awaitRpcFlowFinished(bobHoldingId, requestId)
+        assertThat(result.flowResult).contains(applicationCpiName)
```
simon-johnson-r3: It feels like the two ends of this test don't quite meet up. You could output the data from the flow as JSON and the parse that in here quite easily. If you didn't want to check exact values then you could check they were populated with something at least.

Or you could just output the application name as that's all you check here.

I guess you output the full set because it helps with manual testing, but I think there should be some match up between the automated test and the smoke test output. Future maintainers might be wondering what their responsibility is if new fields are added.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
```diff
@@ -67,9 +86,15 @@ class FlowRunnerImpl @Activate constructor(
         val flowStartContext = context.checkpoint.flowStartContext
         val sessionId = flowStartContext.statusKey.id
 
+        val remotePlatformProperties =
```
simon-johnson-r3: Are these really the remote flow properties, you appear to pass the holding Id of the flow about to be executed? Is remote implying counterparty here?

Note that `remoteToLocalContextMapper` and the use of its return type is quite basic at the moment, it uses all platform properties as both current and counterparty context because they didn't differ at the point that method was written. The method was split out in order to make this logic modifiable at a later date.

At line 106 `localContext.counterpartySessionProperties` you'll basically pass `remotePlatformProperties` as counterparty session properties and I don't think that is your intent necessarily.

The intent when this was written was the flow context would eventually contain not just the current flow context, but also the context of the counterparty that initiated you. But that is not yet implemented because it is only now in this PR that flow context gets some of that information which differs across the 2 flows. This is probably the time to mull over and implement how both of these sets of context are exposed to the cordapp developer. Might be worth checking with @mnesbit what expectations are.

simon-johnson-r3 (changes requested):

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
mnesbit: My intent was that in a flow corda.platformVersion and corda.cpiVersion (or whatever name we settled on) and similar platform properties are always of local system. In an initiated flow we would send across the platform+user properties from initiator. We would start forming the context properties for initiated flow. The initiator platform properties would be renamed from corda.xxx to corda.initiator.xxx (we don't have to preserve more than one level in this i.e. don't send from initiator any corda.initiator items). The local platform properties would be written into context as normal (possibly squashing any unrescued initiator values). The user properties should be copied over as as, so that they continue to be available in initiated flow as they were to initiator.

pmck91: thanks for clearing that up! What Ill do is copy over the initiator details and rename them and place them in the remotePlatformContextProperties, I can do this in [remoteToLocalContextMapper](https://github.com/corda/corda-runtime-os/blob/85526d0e5131ae707cffd73db7a74ea34b4ea0ec/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/RemoteToLocalContextMapper.kt), then I can fill in the others as normal into platformProperties?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/RemoteToLocalContextMapper.kt
```diff
@@ -18,13 +19,39 @@ fun remoteToLocalContextMapper(
...
+ * This will only rename one level deep.
+ * e.g. we wont see corda.initiator.initiator
+ *
+ * @param keyValuePairList A KVP List that will have the keys renamed.
+ * @return The newly re-keyed KVP List
+ */
+@Suppress("NestedBlockDepth")
+fun renameInitiatorProps(keyValuePairList: KeyValuePairList) = KeyValueStore().apply {
+    keyValuePairList.items.forEach { kvp ->
+        if (!kvp.key.contains("corda.initiator")) {
```
mnesbit: use startsWith rather than contains for these

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/RemoteToLocalContextMapper.kt
```diff
@@ -18,13 +19,39 @@ fun remoteToLocalContextMapper(
     remoteUserContextProperties: KeyValuePairList,
     remotePlatformContextProperties: KeyValuePairList
 ): LocalContext {
+
+    // replace 'corda.' with 'corda.initiator.'
+    val initiatorPlatformContextProperties = renameInitiatorProps(remotePlatformContextProperties)
```
mnesbit, line 24: Where do we set the local platform properties? Shouldn't we do that in this function too?

pmck91: they're set later after the mapping, back in the flowRunnerImpl

simon-johnson-r3 (dismissed):

ronanbrowne (approved):
lgtm (build change)

PR #3399, opened 2023-03-16
CORE-11751: Enable crypto-utils publishing

mcgovc:
Issue flagged testing CSDE as crypto-utils dependency is required in simulator.

### DIFF
--- a/libs/crypto/crypto-utils/build.gradle
+++ b/libs/crypto/crypto-utils/build.gradle
@@ -1,6 +1,11 @@
 plugins {
     id 'corda.common-publishing'
     id 'corda.common-library'
+    id 'corda.javadoc-generation'
+}
+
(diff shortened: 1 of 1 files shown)


libs/crypto/crypto-utils/build.gradle
```diff
@@ -1,6 +1,11 @@
 plugins {
     id 'corda.common-publishing'
     id 'corda.common-library'
+    id 'corda.javadoc-generation'
+}
+
+ext {
+    releasable = true
```
driessamyn, line 8: We should remember to undo these when we remove simulator as as a customer artifact.

We don't treat this module as API, so there will be breaking changes. 

FYI. @mattbradburyr3

mcgovc:
No longer required

PR #3400, opened 2023-03-16
CORE-7994 Replace key with key id in digital signature packet

vlajos:
At the moment, the `SigningService.sign` returned signature packet includes the signing key (public key) and the signature packet will be sent to over the wire to counter parties. This means that currently, the receiver of that signature packet can grab that key and use it to verify the signature. However, they should be checking that this key exists in the locally known/ trusted counter parties keys. By changing the key to key id it means the consumer of the signature has to to resolve the key out of the key id to proceed with signature verification.

- refactors `TransactionSignatureService.verifySignature` and use it in `UtxoSignedTransactionImpl`, `ConsensualSignedTransactionImpl` and `UtxoFinalityBaseV1` so as to go through the transactions for signature verification because they hold the signatories and therefore that is where we can map the signature's key id to keys
- maps key id to key in `DigitalSignature.WithKeyId` consuming sites
- statically converts key to key id in `SigningServiceImpl.sign` return type (`DigitalSignature.WithKeyId`) 
- internalizes `DigitalSignature.WithKey` to continue using it internally

api PR: [#956](https://github.com/corda/corda-api/pull/956)


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
```diff
@@ -107,34 +110,61 @@ data class UtxoSignedTransactionImpl(
...
-            } catch (e: Exception) {
-                false
+        val keyIdToSignatory = signatories.associateBy {
+            transactionSignatureService.getIdOfPublicKey(
+                it,
+                DigestAlgorithmName.SHA2_256.name // TODO: Where should this come from???
+            )
+        }
+
+        val appliedSignatories = signatures.mapNotNull {
```
blsemo: I think that is a bad name - you can't apply signatories. It should be something like "appliedRequiredSignatures" or so,

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
```diff
@@ -107,34 +110,61 @@ data class UtxoSignedTransactionImpl(
...
+            val publicKey = keyIdToSignatory[it.by]
+            if (publicKey == null) {
+                null
+            } else {
+                try {
+                    transactionSignatureService.verifySignature(this, it, publicKey)
+                    publicKey
+                } catch (e: Exception) {
+                    null
+                }
```
blsemo, line 139: Not sure this is correct - if a signatory has "signed" the transaction with an invalid signature, that isn't just a missing signature, but an error, isn't it?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
```diff
@@ -158,6 +210,63 @@ data class UtxoSignedTransactionImpl(
         }
     }
 
+    @Suspendable
+    override fun verifyNotarySignature(signature: DigitalSignatureAndMetadata) {
+        val keyIdToSignatory = getNotaryKeys().associateBy {
+            transactionSignatureService.getIdOfPublicKey(
+                it,
+                DigestAlgorithmName.SHA2_256.name // TODO: Where should this come from???
```
blsemo: In the case where you pass in a `DigitalSignatureAndMetadata`, that surely should have the digest algorithm?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
```diff
@@ -107,34 +110,61 @@ data class UtxoSignedTransactionImpl(
...
-        val appliedSignatories = signatures.filter {
-            try {
-                transactionSignatureService.verifySignature(this, it)
-                true
-            } catch (e: Exception) {
-                false
+        val keyIdToSignatory = signatories.associateBy {
+            transactionSignatureService.getIdOfPublicKey(
+                it,
+                DigestAlgorithmName.SHA2_256.name // TODO: Where should this come from???
```
blsemo: Does this need to be passed in as a constructor paramters to the `SignedTransactionImpl` - this must be configured somewhere in the SigningService? 
Clearly, we can't hard code it.
Also wondering whether we should calculate the `keyIdToSignatory` as a lazy field - it is used a few times, and it can't change once the SignedTransaction is built.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
vlajos: Previously we just ignored them. Since we are interested in the missing ones here.
And we treat these ones missing.
Do you say that raising an exception would be better in this case?
(Probably I would do that separately if so.)

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
vlajos: I did not check thoroughly but I think we should have some signatures all the places where we need this algorithm, and the keyId's are relevant only in the context of those signatures which ones have a by field which has an algorithm field what we could reuse. Probably.

kyriathar: I understand `signatories` contains keys that need to sign the transaction. And then we have received signatures. With changing signatures to now hold key id, means we also have access to the digest algorithm used to hash the key (key id). So I think this static mapping here by SHA_256 won't work if the signature is of some existing key (in signatories) but hashed with different digest algorithm, e.g. SHA_512.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
vlajos: Yes. I have updated these bits.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
vlajos: Yes. That was just a short cut. I have updated these bits to get the algorithm from the hashes.
(It could use some lazy loading though.)

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
kyriathar: Renamed

components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/TransactionSignatureServiceImpl.kt
```diff
@@ -257,10 +280,8 @@ class TransactionSignatureServiceImpl @Activate constructor(
         }
     }
 
-    private fun checkAndGetSignatureSpec(signatureWithMetadata: DigitalSignatureAndMetadata): SignatureSpec {
-        val signatureSpec = signatureWithMetadata.metadata.signatureSpec
-
-        val compatibleSpecs = signatureSpecService.compatibleSignatureSpecs(signatureWithMetadata.by)
+    private fun checkAndGetSignatureSpec(signatureSpec: SignatureSpec, signingKey: PublicKey): SignatureSpec {
```
vlajos: This one looks like checking only the signature spec now.
```suggestion
    private fun checkSignatureSpec(signatureSpec: SignatureSpec, signingKey: PublicKey): SignatureSpec {
```
Maybe?

kyriathar: I think it still also returns the `SignatureSpec` though?

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityBaseV1.kt
```diff
@@ -36,11 +32,16 @@ abstract class ConsensualFinalityBaseV1 : SubFlow<ConsensualSignedTransaction> {
         sessionToNotify: FlowSession? = null
     ) {
         try {
-            transactionSignatureService.verifySignature(transaction, signature)
-            log.debug { "Successfully verified signature($signature) by ${signature.by.encoded} (encoded) for transaction $transaction.id" }
+            transaction.verifySignature(signature)
+            // TODO The below logging needs to be moved in the transaction
+//            log.debug {
+//                "Successfully verified signature($signature) by ${signature.by.encoded} (encoded) for transaction $transaction.id"
+//            }
```
vlajos: Hmm. Why should these be moved into the transaction classes?

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityBaseV1.kt
```diff
@@ -36,11 +32,16 @@ abstract class ConsensualFinalityBaseV1 : SubFlow<ConsensualSignedTransaction> {
...
         } catch (e: Exception) {
-            val message = "Failed to verify transaction's signature($signature) by ${signature.by.encoded} (encoded) for " +
-                    "transaction ${transaction.id}. Message: ${e.message}"
+            // TODO The below logging needs to be moved in the transaction
+//            val message = "Failed to verify transaction's signature($signature) by ${signature.by.encoded} (encoded) for " +
+//                    "transaction ${transaction.id}. Message: ${e.message}"
+            val message = ""
```
vlajos: Same as https://github.com/corda/corda-runtime-os/pull/3400/files#r1149314235 ?

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityBaseV1.kt
vlajos: I have just read the context from slack.
I think the best is to keep the logging here for now.
The key itself is not super important for successful cases.

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityBaseV1.kt
vlajos: Also I would keep this one too. But the SignedTransaction's verifySignatures method's error message could be improved to mention the publickey. Then it will be available here automatically too.

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityFlowV1.kt
```diff
@@ -110,25 +110,32 @@ class ConsensualFinalityFlowV1(
...
-            val counterpartiesToSignatoriesMessages = signaturesReceivedFromSessions.map { (session, signatures) ->
-                "${session.counterparty} provided ${signatures.size} signature(s) to satisfy the signatories (encoded) " +
-                        signatures.map { it.by.encoded }
-            }
-            val counterpartiesToSignatoriesMessage = if (counterpartiesToSignatoriesMessages.isNotEmpty()) {
-                "\n${counterpartiesToSignatoriesMessages.joinToString(separator = "\n")}"
-            } else {
-                "[]"
-            }
+            // TODO The below logging needs to be moved in the transaction or we switch to logging key ids,
```
vlajos: I would keep the logging here and adjust it to log the keyIds.
Maybe creating a tech debt jira ticket for it.
It is possible that the counterparties send back a signature which cannot be resolved against the signatories of the transaction or may be completely unknown for the MGM.
I think keyIds are the safest.

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityBaseV1.kt
kyriathar: We used to log the key here (through the signature). However, the signature now contains the key id. It doesn't seem very right to try and look up key from other source (e.g. `MemberLookup`). Key, however is available in transaction.

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityFlowV1.kt
kyriathar: That's what I thought, however then, we end up with not consistent logging (elsewhere we still logging keys). So maybe needs more proper fixing. Also if we go for key ids it means we might get keys hashed with different algorithms so different key id forms. Seems like converting them back to keys gives us some "canonicalization".

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/transaction/ConsensualSignedTransactionImpl.kt
```diff
@@ -58,44 +58,61 @@ class ConsensualSignedTransactionImpl(
...
-                false
-            }
-        }.map { it.by }.toSet()
+        // TODO See if adding the following as property breaks CordaSerializable
         val requiredSignatories = this.toLedgerTransaction().requiredSignatories
-        return requiredSignatories.filter {
-            !KeyUtils.isKeyFulfilledBy(
-                it,
-                appliedSignatories
-            ) // isFulfilledBy() helps to make this working with CompositeKeys.
```
vlajos, line 75: I think the function's name in this comment changed. (It may be unrelated to this PR though.)
(It is copy pasted to a few places.)

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityFlowV1.kt
kyriathar: In the scenario you described indeed, key id is safest then. Maybe we could opt for logging by key id only when key is missing?

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/transaction/ConsensualSignedTransactionImpl.kt
```diff
@@ -58,44 +58,61 @@ class ConsensualSignedTransactionImpl(
...
     override fun getMissingSignatories(): Set<PublicKey> {
-        val appliedSignatories = signatures.filter {
-            try {
-                transactionSignatureService.verifySignature(this, it)
-                true
-            } catch (e: Exception) {
-                false
-            }
-        }.map { it.by }.toSet()
+        // TODO See if adding the following as property breaks CordaSerializable
```
vlajos, line 72: I suspect it would involve some work around the serializers, but it should be possible.
On the other hand, it is probably not the highest priority.
I think a tech debt Jira would be enough for it.

kyriathar: Raised [CORE-12088](https://r3-cev.atlassian.net/browse/CORE-12088) for it

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
```diff
@@ -112,34 +113,57 @@ data class UtxoSignedTransactionImpl(
...
+            if (publicKey == null) {
+                null// We do not care about non-notary/non-signatory keys
+            } else {
+                try {
+                    transactionSignatureServiceInternal.verifySignature(this, it, publicKey)
+                    publicKey
+                } catch (e: Exception) {
+                    throw TransactionSignatureException(
+                        id,
+                        "Failed to verify signature of ${it.signature} for transaction $id. Message: ${e.message}",
```
vlajos: I think this should cover the negative cases (re logging):
```suggestion
                        "Failed to verify signature of ${it.signature} from $publicKey for transaction $id. Message: ${e.message}",
```

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityBaseV1.kt
kyriathar: Done

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityBaseV1.kt
kyriathar: Done

components/flow/flow-service/src/test/kotlin/net/corda/flow/application/crypto/SigningServiceImplTest.kt
```diff
@@ -24,13 +26,18 @@ class SigningServiceImplTest {
...
+            whenever(it.encoded).thenReturn(byteArrayOf(0x00))
+        }
+        val signature = DigitalSignatureWithKey(signingKey, byteArrayOf(1))
         val publicKey = mock<PublicKey>()
         val encodedPublicKeyBytes = byteArrayOf(2)
         whenever(keyEncodingService.encodeAsByteArray(publicKey)).thenReturn(encodedPublicKeyBytes)
         whenever(externalEventExecutor.execute(eq(CreateSignatureExternalEventFactory::class.java), captor.capture()))
             .thenReturn(signature)
-        assertEquals(signature, signingService.sign(byteArrayOf(1), publicKey, mock()))
+        // the below will be changed to signatureWithKeyId
```
blsemo: What's the point of this comment? Leftover note from implenting this?

components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/TransactionSignatureServiceImpl.kt
```diff
@@ -129,9 +135,19 @@ class TransactionSignatureServiceImpl @Activate constructor(
 
     override fun verifySignature(
         transaction: TransactionWithMetadata,
-        signatureWithMetadata: DigitalSignatureAndMetadata
+        signatureWithMetadata: DigitalSignatureAndMetadata,
+        publicKey: PublicKey
     ) {
-        val signatureSpec = checkAndGetSignatureSpec(signatureWithMetadata)
+        val digestAlgorithmOfKeyId = signatureWithMetadata.signature.by.algorithm
```
blsemo, line 141: Do we need to check that the digestAlgorithm is from an allow list of acceptable algorithms?

components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/TransactionSignatureServiceImpl.kt
blsemo: Yes, it returns the signatureSpec, but you have to feed it in, and it just passes it through when the check works, so it does not get the signature spec for you. I think the renaming proposal is correct.

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityBaseV1.kt
```diff
@@ -36,10 +32,12 @@ abstract class ConsensualFinalityBaseV1 : SubFlow<ConsensualSignedTransaction> {
...
         try {
-            transactionSignatureService.verifySignature(transaction, signature)
-            log.debug { "Successfully verified signature($signature) by ${signature.by.encoded} (encoded) for transaction $transaction.id" }
+            transaction.verifySignature(signature)
+            log.debug {
+                "Successfully verified signature($signature) by ${signature.by} (key id) for transaction $transaction.id"
+            }
         } catch (e: Exception) {
-            val message = "Failed to verify transaction's signature($signature) by ${signature.by.encoded} (encoded) for " +
+            val message = "Failed to verify transaction's signature($signature) by ${signature.by} (key id) for " +
```
blsemo:
```suggestion
            val message = "Failed to verify transaction signature($signature) by ${signature.by} (key id) for " +
```

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityFlowV1.kt
```diff
@@ -100,7 +100,7 @@ class ConsensualFinalityFlowV1(
             signatures.forEach { signature ->
                 transaction = verifyAndAddSignature(transaction, signature)
                 log.debug {
-                    "Added signature by ${signature.by.encoded} (encoded) from ${session.counterparty} of $signature for transaction " +
+                    "Added signature by ${signature.by} (key id) from ${session.counterparty} of $signature for transaction " +
```
blsemo:
```suggestion
                    "Added signature $signature by (key id) ${signature.by} from ${session.counterparty} for transaction " +
```

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/transaction/ConsensualSignedTransactionImpl.kt
```diff
@@ -58,44 +58,61 @@ class ConsensualSignedTransactionImpl(
...
         val requiredSignatories = this.toLedgerTransaction().requiredSignatories
-        return requiredSignatories.filter {
-            !KeyUtils.isKeyFulfilledBy(
-                it,
-                appliedSignatories
-            ) // isFulfilledBy() helps to make this working with CompositeKeys.
+        val appliedSignatures = signatures.mapNotNull {
+            val signatureKey = getSignatoryKeyFromKeyId(it.by, requiredSignatories)
```
blsemo, line 75: If you call `getSignatoryKeyFromKeyId` in a loop, you should definitely cache the result of mapping all the keys to their hashes. As the set of signatories doesn't change once a transaction is built, this should be straight forward.

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/transaction/ConsensualSignedTransactionInternal.kt
```diff
@@ -52,4 +52,6 @@ interface ConsensualSignedTransactionInternal: ConsensualSignedTransaction {
      */
     @Suspendable
     fun verifySignatures()
+
+    fun verifySignature(signature: DigitalSignatureAndMetadata)
```
blsemo, line 56: Where is this used? Why would we need to verify a signature that is not part of the signed transaction?

components/ledger/ledger-consensual-flow/src/test/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityFlowV1Test.kt
```diff
@@ -249,12 +270,14 @@ class ConsensualFinalityFlowV1Test {
...
         assertThatThrownBy { callFinalityFlow(signedTransaction, listOf(sessionAlice, sessionBob)) }
             .isInstanceOf(TransactionMissingSignaturesException::class.java)
-            .hasMessageContainingAll(
-                "Transaction $TX_ID is missing signatures for signatories (encoded) ${setOf(publicKeyBob).map { it.encoded }}",
-                "The following counterparties provided signatures while finalizing the transaction:",
-                "$ALICE provided 2 signature(s) to satisfy the signatories (encoded) ${aliceSignatures.map { it.by.encoded }}",
-                "$BOB provided 0 signature(s) to satisfy the signatories (encoded) []"
-            )
+//        assertThatThrownBy { callFinalityFlow(signedTransaction, listOf(sessionAlice, sessionBob)) }
+//            .isInstanceOf(TransactionMissingSignaturesException::class.java)
```
blsemo: Why did you have to disable this message check?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityBaseV1.kt
```diff
@@ -71,11 +65,14 @@ abstract class UtxoFinalityBaseV1 : SubFlow<UtxoSignedTransaction> {
     ) {
         try {
             log.debug { "Verifying signature($signature) of transaction: $transaction.id" }
-            transactionSignatureService.verifySignature(transaction, signature)
-            log.debug { "Successfully verified signature($signature) by ${signature.by.encoded} (encoded) for transaction $transaction.id" }
+            transaction.verifySignatorySignature(signature)
+            // TODO The below logging needs to be moved in the transaction
```
blsemo: There are a lot of TODOs here, without any ticket next to them, and commented out code - that needs to be cleaned up before merging.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
blsemo: Ok, fine. Can you raise a ticket to re-assess later, please?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
```diff
@@ -112,34 +113,57 @@ data class UtxoSignedTransactionImpl(
         )
     }
 
+    private fun getSignatoryKeyFromKeyId(keyId: SecureHash): PublicKey? {
+        val keyIdsToSignatories = signatories.associateBy {// todo cache this
```
blsemo, line 117: This definitely needs to be cached, which should be easy as it doesn't change..

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
```diff
@@ -163,6 +220,49 @@ data class UtxoSignedTransactionImpl(
         }
     }
 
+    @Suspendable
+    override fun verifyNotarySignature(signature: DigitalSignatureAndMetadata) {
+        val publicKey = getNotaryPublicKeyByKeyId(signature.by)
+            ?: throw CordaRuntimeException( // todo transition to TransactionSignatureException
+                "Notary's signature has not been created by the transaction's notary. " +
+                        "Notary's public key: $notaryKey " +
+                        "Notary signature's key Id: ${signature.by}"
```
blsemo:
```suggestion
                "Notary signature has not been created by the notary for this transaction. " +
                        "Notary public key: $notaryKey " +
                        "Notary signature key Id: ${signature.by}"
```

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
```diff
@@ -163,6 +220,49 @@ data class UtxoSignedTransactionImpl(
...
+            )
+
+        if (!KeyUtils.isKeyInSet(notaryKey, listOf(publicKey))) {
+            throw CordaRuntimeException( // todo transition to TransactionSignatureException
+                "Notary's signature has not been created by the transaction's notary. " +
+                        "Notary's public key: $notaryKey " +
+                        "Notary signature's key: $publicKey"
```
blsemo:
```suggestion
                "Notary signature has not been created by the notary for this transaction. " +
                        "Notary public key: $notaryKey " +
                        "Notary signature key: $publicKey"
```

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1Test.kt
```diff
@@ -609,65 +675,69 @@ class UtxoFinalityFlowV1Test {
...
-        verify(flowMessaging).sendAllMap(
-            mapOf(
-                sessionAlice to listOf(signatureBob),
-                sessionBob to listOf(signatureAlice1, signatureAlice2)
-            )
-        )
-        verify(flowMessaging).sendAll(any<Payload.Failure<List<DigitalSignatureAndMetadata>>>(), eq(sessions))
-        verify(flowMessaging, never()).sendAll(eq(Payload.Success(listOf(signatureNotary))), any())
-    }
+    // TODO Test needs to move to `UtxoSignedTransactionImpl`. Same behavior is already tested here in
```
blsemo: Either do it or raise a ticket for it - a TODO/commented out code is not a great look.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1Test.kt
```diff
@@ -825,18 +917,21 @@ class UtxoFinalityFlowV1Test {
...
 
         assertThatThrownBy { callFinalityFlow(initialTx, listOf(sessionAlice, sessionBob)) }
             .isInstanceOf(TransactionMissingSignaturesException::class.java)
-            .hasMessageContainingAll(
-                "Transaction $TX_ID is missing signatures for signatories (encoded) ${setOf(publicKeyBob).map { it.encoded }}",
-                "The following counterparties provided signatures while finalizing the transaction:",
-                "$ALICE provided 2 signature(s) to satisfy the signatories (encoded) ${aliceSignatures.map { it.by.encoded }}",
-                "$BOB provided 0 signature(s) to satisfy the signatories (encoded) []"
-            )
+
```
blsemo: We should still check that we get meaningful messages in the exceptions.

notary-plugins/notary-plugin-common/src/main/kotlin/com/r3/corda/notary/plugin/common/PluggableNotaryFlowHelpers.kt
```diff
@@ -41,11 +41,15 @@ fun validateRequestSignature(notarizationRequest: NotarizationRequest,
...
-    if (requestingPartyKey != digitalSignature.by) {
+    val digestAlgorithmOfSignatureKeyId = digitalSignature.by.algorithm
+    val requestingPartyKeyId =
+        requestingPartyKey.fullIdHash(digestService, digestAlgorithmOfSignatureKeyId)
+    if (requestingPartyKeyId != digitalSignature.by) {
         throw IllegalStateException(
-            "Expected a signature by ${requestingPartyKey.publicKeyId(digestService)}, " +
-                    "but received by ${digitalSignature.by.publicKeyId(digestService)}}"
+            // TODO The below message used to print keys and now it prints key ids.
+            //  Maybe change to only mentioned that did not receive signature from expected?
```
blsemo: I think this always printed keyIds, i.e. hashes? The left hand side hashes the key before printing it. Definitely better to print a key id than just a generic "didn't receive from expected parties".

simulator/runtime/src/main/kotlin/net/corda/simulator/runtime/ledger/utxo/UtxoTransactionFinalityHandler.kt
```diff
@@ -122,20 +123,29 @@ class UtxoTransactionFinalityHandler(
...
             throw CordaRuntimeException("Transaction ${finalTransaction.id} " +
                     "is missing signatures for signatories ${missingSignatories.map { memberLookup.lookup(it) }}")
         }
     }
 
+    private fun signatorySignatureExists(key: PublicKey, appliedSignatures: List<SecureHash>): Boolean =
+        appliedSignatures.any {
+            // TODO Need to use digest service instead below
+            val digestAlgorithm = it.algorithm
+            val keyId = MessageDigest.getInstance(digestAlgorithm).digest(key.encoded)
```
blsemo, line 145: Do this once outside the inner loop!

blsemo (commented):
There's a bit more work required around transacton verification etc, also caching of key look-ups.
@vlajos and @kyriathar can you co-ordinate to do that either in this PR or in a follow up that is to come in very shortly, please?

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityFlowV1.kt
kyriathar: Done

components/flow/flow-service/src/test/kotlin/net/corda/flow/application/crypto/SigningServiceImplTest.kt
kyriathar: Yes, this seems obsolete, removed

components/ledger/ledger-consensual-flow/src/test/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityFlowV1Test.kt
kyriathar: This is now put back

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityBaseV1.kt
kyriathar: All of these TODOs around logging are now removed and logging is put back (using key ids).

components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/TransactionSignatureServiceImpl.kt
kyriathar: Ah yes indeed, proposal will be applied.

components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/TransactionSignatureServiceImpl.kt
kyriathar: I don't think we currently have such service. This is to come with [CORE-8151](https://r3-cev.atlassian.net/browse/CORE-8151)

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/transaction/ConsensualSignedTransactionInternal.kt
kyriathar: This is used in `ConsensualFinalityBaseV1.verifySignature`. @vlajos will correct me If I am wrong, this is added because the transaction holds the signatories (i.e. keys) and then we receive the transaction which now holds the key ids. So we pass it in the transaction to achieve the mapping from key id to key, resolve the signature's key and proceed with transaction verification.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
kyriathar: It's the signature's key id digest algorithm that might be different each time.

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/transaction/ConsensualSignedTransactionInternal.kt
kyriathar: I just noticed that `UtxoSignedTransaction.getSignatories` is publicly exposed. So maybe we could avoid this `verifySignature` taking a signature @vlajos? In `ConsensualSignedTransactionInternal` we have `getMissingSignatories` which should also be good enough to also avoid the same there? On the other hand it does encapsulate verification better this way. Not sure.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1Test.kt
kyriathar: This is now put back

notary-plugins/notary-plugin-common/src/main/kotlin/com/r3/corda/notary/plugin/common/PluggableNotaryFlowHelpers.kt
kyriathar: Indeed you are right, not sure what I was thinking. Removing wrong comment.

simulator/runtime/src/main/kotlin/net/corda/simulator/runtime/ledger/utxo/UtxoTransactionFinalityHandler.kt
kyriathar: I think we can't, digest algorithm might be different among signatures key ids.

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/transaction/ConsensualSignedTransactionImpl.kt
kyriathar: I think this rather says that the signatures keys could be of different digest algorithms. So yes some caching needs to be done. Probably the caching needs to be handled all together under a different PR. Registered [CORE-12099](https://r3-cev.atlassian.net/browse/CORE-12099) for it.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
kyriathar: (Same as above) registered [CORE-12099](https://r3-cev.atlassian.net/browse/CORE-12099) for it.

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/flows/finality/v1/ConsensualFinalityFlowV1.kt
kyriathar: Registered [CORE-12100](https://r3-cev.atlassian.net/browse/CORE-12100) to keep recorded that we log both in key and key ids and maybe revisit to find a better way to be more consistent.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1Test.kt
kyriathar: Test removed from here and ported in `UtxoSignedTransactionImplTest`.

components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/TransactionSignatureServiceImpl.kt
mnesbit: https://github.com/corda/corda-api/blob/release/os/5.0/application/src/main/java/net/corda/v5/application/crypto/SignatureSpecService.java#L45 is that list of allowed SignatureSpecs. Basically for proposed publickey and expected hash that is the list of allowed SignatureSpecs. Though it we probably do need to have an idea what hash we expected to be used

nkovacsx (dismissed):
Notary side LGTM

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/transaction/ConsensualSignedTransactionImpl.kt
```diff
@@ -58,44 +58,61 @@ class ConsensualSignedTransactionImpl(
...
-                transactionSignatureService.verifySignature(this, it)
-                true
-            } catch (e: Exception) {
-                throw TransactionSignatureException(
-                    id,
-                    "Failed to verify signature of ${it.signature} for transaction $id. Message: ${e.message}",
-                    e
-                )
+        // TODO See if adding the following as property breaks CordaSerializable
+        val requiredSignatories = this.toLedgerTransaction().requiredSignatories
```
mnesbit, line 95: Think this loop is likely wrong way around for efficiency. We should work through the list of required signers. Hash it (and/or it's leaves if composite) to get id. Locate a signature with that id (or else exception). Verify that signature and remove signature from the list of signatures eligible to match another .

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -126,6 +126,7 @@ class UtxoFinalityFlowV1(
         return transaction to signaturesReceivedFromSessions
     }
 
+    @Suppress("unused_parameter")
```
vlajos: I guess this `Suppress` is not needed anymore.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
```diff
@@ -184,4 +277,11 @@ data class UtxoSignedTransactionImpl(
     override fun toString(): String {
         return "UtxoSignedTransactionImpl(id=$id, signatures=$signatures, wireTransaction=$wireTransaction)"
     }
+
+    private fun getNotaryKeys(): List<PublicKey> {
+        return when (val owningKey = notaryKey) {
```
vlajos, line 282: Are there maybe simpler way to do this?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
kyriathar: Removed

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/transaction/ConsensualSignedTransactionImpl.kt
kyriathar: Good catch. Agreed offline to revisit/ rework key ids to keys mapping in follow-up work.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
kyriathar: I believe this looks good enough

mnesbit (dismissed):
OK

mnesbit (approved):
OK, though we should keep reviewing uses of DigitalSignatureWithKey to ensure we reduce usage to minimum

yift-r3 (approved):
LGTM (p2p and membership code)

driessamyn (approved):
LGTM (crypto changes)

PR #3402, opened 2023-03-17
CORE-11773: instantiate and pass around CryptoRepository

dickon:
Remove WrappingKeyStore - the interface and test coverage, since all that functionality is provided by CryptoRepository instead.

For simplicity, CryptoRepository is a simple function, which takes the references needed to construct itself as function parameters rather than OSGi dependency injection. 

Rely on integration and smoke tests for covering lifecycle and dynamic config, dropping the duplicate alternative implementation in SoftCryptoServiceProviderImpl. This means we have integration and smoke test coverage but not unit test coverage of SoftCryptoServiceProviderImpl, which is a piece of glue logic to plug a couple of caches into SoftCryptoService and pass in some OSGi dependencies.


components/crypto/crypto-softhsm-impl/src/test/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoServiceOperationsTests.kt
```diff
@@ -424,16 +421,16 @@ class SoftCryptoServiceOperationsTests {
...
                 knownWrappingKey.algorithm + "!",
                 rootWrappingKey.wrap(knownWrappingKey)
             )
         )
-        assertThrows<IllegalArgumentException> {
-            cryptoService.generateKeyPair(KeyGenerationSpec(rsaScheme, "key1", alias), emptyMap())
-        }
+//        assertThrows<IllegalArgumentException> {
+//            cryptoService.generateKeyPair(KeyGenerationSpec(rsaScheme, "key1", alias), emptyMap())
+//        }
```
fowlerrr: Can we delete this

osgi-framework-bootstrap/src/main/resources/log4j2-console.xml
```diff
@@ -7,21 +7,21 @@
     </Appenders>
     <Loggers>
         <logger name="Console">
-            <AppenderRef ref="Console" level="info"/>
+            <AppenderRef ref="Console" level="trace"/>
```
fowlerrr: You'll want to revert this

fowlerrr: You'll want to revert this

fowlerrr: Probably the whole file

components/crypto/crypto-softhsm-impl/src/test/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoServiceOperationsTests.kt
dickon: I'll restore the code, it works now.

dickon: done

osgi-framework-bootstrap/src/main/resources/log4j2-console.xml
dickon: done

dickon: done

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/CryptoRepositoryFactory.kt
```diff
@@ -1,11 +1,74 @@
...
+// reference as an argument (i.e. being higher order).
+
+fun cryptoRepositoryFactory(
+    tenantId: String,
+    dbConnectionManager: DbConnectionManager,
+    jpaEntitiesRegistry: JpaEntitiesRegistry,
+    virtualNodeInfoReadService: VirtualNodeInfoReadService,
+): CryptoRepository {
+    val onCluster = CryptoTenants.isClusterTenant(tenantId)
+    val entityManagerFactory = if (onCluster) {
```
driessamyn, line 40: Something that came to mind (but don't change it as part of this PR).
`onCluster` is quite a confusing term as typically vnode DBs will also be part of the Corda cluster.
I think maybe it'd be better to refer to `vnodeTenants`, so maybe `isVnodeTenant` and `CryptoTenants.isVnodeTenant(...)` would be less confusing?

driessamyn (approved):
LGTM

fowlerrr (approved):

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/CryptoRepositoryFactory.kt
dickon: Yes, something to revisit.

PR #3404, opened 2023-03-17
CORE-9159 - Remove PENDING_APPROVAL_FLOW stage for registration

dimosr:
Remove PENDING_APPROVAL_FLOW stage from registration, which will not exist anymore.

corda-api PR: https://github.com/corda/corda-api/pull/958


nikinagy (dismissed):
LGTM

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/RegistrationStatusHelper.kt
```diff
@@ -17,12 +17,11 @@ internal object RegistrationStatusHelper {
             RegistrationStatus.SENT_TO_MGM -> 1
             RegistrationStatus.RECEIVED_BY_MGM -> 1
             RegistrationStatus.PENDING_MEMBER_VERIFICATION -> 2
-            RegistrationStatus.PENDING_APPROVAL_FLOW -> 3
-            RegistrationStatus.PENDING_MANUAL_APPROVAL -> 4
-            RegistrationStatus.PENDING_AUTO_APPROVAL -> 4
-            RegistrationStatus.DECLINED -> 5
-            RegistrationStatus.INVALID -> 5
-            RegistrationStatus.APPROVED -> 5
+            RegistrationStatus.PENDING_MANUAL_APPROVAL -> 3
```
yift-r3, line 20: Maybe those should be `Double` instead of `Int` (to allow adding or removing statuses with minimal change)?

yift-r3 (dismissed):

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/RegistrationStatusHelper.kt
dimosr: hmm.. what I like about the current scheme with integers is it maps cleanly to modelling a process with sequential steps and makes it a bit easier to reason about. Given that's just internal logic and not encoded in wire formats or anything like that (which would make changes more painful due to evolution), I think being "forced" to think about it when adding/removing stages might not be necessarily a bad thing.

vkolomeyko (dismissed):
LGTM

vkolomeyko (approved):
LGTM

nikinagy (approved):
LGTM

PR #3406, opened 2023-03-17
CORE-10613: Removal of old cli plugins during installation

lkalneus:


vkolomeyko (approved):
LGTM

Many thanks @lkalneus for your contribution.

vkolomeyko:
Looks like there is some problem with CI build for this PR.
@corda/blt - would you know what the problem might be?

Also we tend not to use private forks of the repo, but tend to create branches straight off the main repo, this could be the reason.

vkolomeyko:
Superseded by: https://github.com/corda/corda-runtime-os/pull/3473

PR #3407, opened 2023-03-17
CORE-11709: Move implementation from SigningKeyStore, HSMStore into CryptoRepository

fowlerrr:
Port the code (and dependencies) to the implementation of the more unified `CryptoRepository` API


components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/V1CryptoRepositoryImpl.kt
```diff
@@ -1,41 +1,104 @@
...
-        entityManagerFactory.createEntityManager().use { em ->
-            em.find(WrappingKeyEntity::class.java, alias)?.let { rec ->
-                WrappingKeyInfo(
-                    encodingVersion = rec.encodingVersion,
-                    algorithmName = rec.algorithmName,
-                    keyMaterial = rec.keyMaterial,
-                )
-            }
-        }
+    private val wrappingKeyStore = V1WrappingKeyStore(entityManagerFactory)
```
dickon, line 44: I'm not sure about whether this is a good idea, but following this structure for SigningKeyStore and HSMStore will make the PRs smaller, so maybe we'll revisit and collapse down these store operations.

On the other hand, there's a few too many class attributes for comfort here and the 3 stores feel somewhat unconnected, so maybe let's try this approach for a while.

So, let's go in this direction for now.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/V1HSMStore.kt
```diff
@@ -0,0 +1,128 @@
...
+import net.corda.v5.base.util.EncodingUtils.toHex
+
+class V1HSMStore(
+    private val entityManagerFactory: EntityManagerFactory,
+) {
+    fun findTenantAssociation(tenantId: String, category: String): HSMAssociationInfo? =
+        entityManagerFactory.createEntityManager().use {
+            val result = it.createQuery(
+                """
+            SELECT ca FROM HSMCategoryAssociationEntity ca
```
dickon, line 24: This is just the old code moved, but it occurs to me that we may be able to get Hibernate to do this join with a lazy attribute.

dickon (approved):
LGTM. I looked at the form but not the full content of all the new unit tests, which are much better than the nothing we had before even if there are still gaps. We can fill in the gaps as we make signing key changes, and i'm not so concerned with HSMStore unit tests since that whole table may get dropped at some point soon.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/V1HSMStore.kt
driessamyn: That is different behaviour. I believe this query eagerly fetches it, which is probably more effiient if you know you will have to.
I suggest not optimising it until you need to. you can sink an awefull lot of time in optimising Hibernate and ending up in a rabbit hole.

PR #3409, opened 2023-03-17
CORE-11769 - support file boot config

driessamyn:
Refactor boot config so that it can be provided by one or more files as well as the command line arguments.
This will be needed for being able to set the hashicorp vault auth token using a side container.

API PR: https://github.com/corda/corda-api/pull/962

Docs:
This feature introduces an alternative way of passing in boot configuration to the Corda Workers.
The `-f` or `--values` flags let you specify a path to a configuration file.
This configuration file must be in JSON or properties file format and can contain any of the other configuration options' keys in the format of `<longFormConfigFlag>` or `<longFormConfigFlag>.<paramName>`.
Note that the keys will use [lower camel case ](https://en.wikipedia.org/wiki/Camel_case)

For example:

- `--topic-prefix foo` becomes `topicPrefix=foo` or `{"topicPrefix": "foo"}`
- `--secrets salt=pepper` becomes `secrets.salt=pepper` or ```{"secrets":  { "salt": "pepper"} }`

When keys are provided in multiple ways, priority means:
 
- Files to the right (i.e. specified last) take priority over those to the left.
- Keys provided on the command line take priority over those provided in files.

### DIFF
--- a/applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
+++ b/applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
@@ -27,9 +27,7 @@ import net.corda.processors.p2p.linkmanager.LinkManagerProcessor
 import net.corda.processors.rest.RestProcessor
 import net.corda.processors.uniqueness.UniquenessProcessor
 import net.corda.processors.verification.VerificationProcessor
-import net.corda.schema.configuration.BootConfig.BOOT_CRYPTO
-import net.corda.schema.configuration.BootConfig.BOOT_DB_PARAMS
-import net.corda.schema.configuration.BootConfig.BOOT_REST_PARAMS
+import net.corda.schema.configuration.BootConfig
 import net.corda.schema.configuration.DatabaseConfig
 import org.osgi.service.component.annotations.Activate
 import org.osgi.service.component.annotations.Component
@@ -110,9 +108,9 @@ class CombinedWorker @Activate constructor(
             // the combined worker may use SOFT HSM by default unlike the crypto worker
             params.hsmId = SOFT_HSM_ID
         }
-        val databaseConfig = PathAndConfig(BOOT_DB_PARAMS, params.databaseParams)
-        val cryptoConfig = PathAndConfig(BOOT_CRYPTO, createCryptoBootstrapParamsMap(params.hsmId))
-        val restConfig = PathAndConfig(BOOT_REST_PARAMS, params.restParams)
+        val databaseConfig = PathAndConfig(BootConfig.BOOT_DB, params.databaseParams)
+        val cryptoConfig = PathAndConfig(BootConfig.BOOT_CRYPTO, createCryptoBootstrapParamsMap(params.hsmId))
+        val restConfig = PathAndConfig(BootConfig.BOOT_REST, params.restParams)
         val config = getBootstrapConfig(
             secretsServiceFactoryResolver,
             params.defaultParams,
@@ -123,10 +121,10 @@ class CombinedWorker @Activate constructor(
         val superUser = System.getenv("CORDA_DEV_POSTGRES_USER") ?: "postgres"
         val superUserPassword = System.getenv("CORDA_DEV_POSTGRES_PASSWORD") ?: "password"
         val dbName = dbUrl.split("/").last().split("?").first()
-        val dbAdmin = if (config.getConfig(BOOT_DB_PARAMS).hasPath(DatabaseConfig.DB_USER))
-            config.getConfig(BOOT_DB_PARAMS).getString(DatabaseConfig.DB_USER) else "user"
-        val dbAdminPassword = if (config.getConfig(BOOT_DB_PARAMS).hasPath(DatabaseConfig.DB_PASS))
-            config.getConfig(BOOT_DB_PARAMS).getString(DatabaseConfig.DB_PASS) else "password"
+        val dbAdmin = if (config.getConfig(BootConfig.BOOT_DB).hasPath(DatabaseConfig.DB_USER))
+            config.getConfig(BootConfig.BOOT_DB).getString(DatabaseConfig.DB_USER) else "user"
+        val dbAdminPassword = if (config.getConfig(BootConfig.BOOT_DB).hasPath(DatabaseConfig.DB_PASS))
+            config.getConfig(BootConfig.BOOT_DB).getString(DatabaseConfig.DB_PASS) else "password"
 
         // Part of DB setup is to generate defaults for the crypto code. That currently includes a
         // default master wrapping key passphrase and salt, which we want to keep secret, and so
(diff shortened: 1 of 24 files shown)


applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/WorkerHelpers.kt
```diff
@@ -127,22 +136,7 @@ class WorkerHelpers {
...
-         */
-        private fun makeSecrets(smartConfigFactory: SmartConfigFactory, config: SmartConfig): SmartConfig {
-            return if (config.hasPath(BOOT_JDBC_PASS)) {
-                smartConfigFactory.makeSecret(config.getString(BOOT_JDBC_PASS), "boot-db-pass").atPath(BOOT_JDBC_PASS)
-                    .withFallback(config)
-            } else {
-                config
-            }
+            return bootConfig
         }
```
driessamyn, line 140: @LWogan , @JamesHR3 I removed this while refactoring for this PR as we shouldn't be doing this. This isn't always going to work, for example when using HashiCorp Vault and when we don't have permissions to write into the vault.
If we want to mask values, we just need to make sure they are configured as secrets from the start.

libs/rest/ssl-cert-read-impl/src/main/kotlin/net/corda/rest/ssl/impl/SslCertReadServiceImpl.kt
```diff
@@ -25,7 +23,7 @@ class SslCertReadServiceImpl(private val createDirectory: () -> Path) : SslCertR
 
         private val String.withoutPrefix: String
             get() {
-                return this.removePrefix("$BOOT_REST_PARAMS.")
+                return this.removePrefix("${BootConfig.BOOT_REST}.")
```
driessamyn, line 26: @vkolomeyko this is changing some of this recently added config, but I don't think it functionally changes what you are doing. It just removes the `params` bit we are now wanting to support configuring by file and the file should match the cmd line structure.

applications/workers/worker-common/src/test/resources/example-config.json
```diff
@@ -0,0 +1,42 @@
+{
```
driessamyn, line 1: @davidcurrie this is an example of a file based configuration.
This is a full file, but it doesn't have to be.
There is also an example of a property file format rather than json in this test.

applications/workers/release/crypto-worker/src/main/kotlin/net/corda/applications/workers/crypto/CryptoWorker.kt
```diff
@@ -85,7 +86,7 @@ class CryptoWorker @Activate constructor(
         params.defaultParams,
         configurationValidatorFactory.createConfigValidator(),
         listOf(
-            PathAndConfig(BootConfig.BOOT_DB_PARAMS, params.databaseParams),
+            PathAndConfig(BootConfig.BOOT_DB, params.databaseParams),
```
driessamyn, line 89: @dickon @fowlerrr some changes here on the crypto config. I don't think this impacts or will conflict with the config changes needed for key rotation as I think those config settings are part of the dynamic configuration rather than boot.

davidcurrie (commented):
LGTM from a UX perspective. Having said that Vault is capable of templating any file format, seeing the example for full config does make me wonder again whether there is value in also having the ability to reference the contents of a file for an individual config property. This would allow, for example, direct access to secrets mounted as files.

driessamyn:
> LGTM from a UX perspective. Having said that Vault is capable of templating any file format, seeing the example for full config does make me wonder again whether there is value in also having the ability to reference the contents of a file for an individual config property. This would allow, for example, direct access to secrets mounted as files.

I wonder if that's something we could add in 5.1 if it isn't too much of an issue at the moment?
Just trying to limit the amount of work needed in the next couple of weeks.

davidcurrie:
> I wonder if that's something we could add in 5.1 if it isn't too much of an issue at the moment?
> Just trying to limit the amount of work needed in the next couple of weeks.

It would be good to decide if this was something we were going to do as it might influence some of the current naming (@vkolomeyko's TLS certs would be an example of this). The security team did mention AWS CSI Driver (mounting secrets from AWS Secret Manager as files) but then things went quiet again.

driessamyn:
> > I wonder if that's something we could add in 5.1 if it isn't too much of an issue at the moment?
> > Just trying to limit the amount of work needed in the next couple of weeks.
> 
> It would be good to decide if this was something we were going to do as it might influence some of the current naming (@vkolomeyko's TLS certs would be an example of this). The security team did mention AWS CSI Driver (mounting secrets from AWS Secret Manager as files) but then things went quiet again.

I think this would be quite simple to add as individual flags, but as a generic thing may need a bit of thinking and probably quite a bit of testing.

For example. Would it be that something like  `-s salt=file://var/foo/bar` would be recognised as something that would load the value of `/var/foo/bar` into the `secrets.salt` config? Or would it be a different flag of e.g. `--file-source name=secrets.salt value=/var/foo/bar`?

I'm really not against supporting either, just reluctant to take this in in the remaining 10 working days before code freeze.

davidcurrie:
> For example. Would it be that something like -s salt=file://var/foo/bar would be recognised as something that would load the value of /var/foo/bar into the secrets.salt config? Or would it be a different flag of e.g. --file-source name=secrets.salt value=/var/foo/bar?

Given what we already have for the secret provider, I would have gone for something like the former.

> I'm really not against supporting either, just reluctant to take this in in the remaining 10 working days before code freeze.

Yes - sorry - when I said "whether we are going to do this" I meant at some point in the future, not in this release.

vkolomeyko (dismissed):
LGTM

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
```diff
@@ -110,9 +108,9 @@ class CombinedWorker @Activate constructor(
             // the combined worker may use SOFT HSM by default unlike the crypto worker
             params.hsmId = SOFT_HSM_ID
         }
-        val databaseConfig = PathAndConfig(BOOT_DB_PARAMS, params.databaseParams)
-        val cryptoConfig = PathAndConfig(BOOT_CRYPTO, createCryptoBootstrapParamsMap(params.hsmId))
-        val restConfig = PathAndConfig(BOOT_REST_PARAMS, params.restParams)
+        val databaseConfig = PathAndConfig(BootConfig.BOOT_DB, params.databaseParams)
```
dickon, line 111: Seems a bit arbitrary, not sure that change was really necessary, but I'm okay with the change.

dickon: also not justified in the PR description. Not a merge blocker, just makes the PR bigger than it coudl be.

applications/workers/release/crypto-worker/src/main/kotlin/net/corda/applications/workers/crypto/CryptoWorker.kt
dickon: Yeah, provided this gets through CI it should be fine.

dickon (commented):
LGTM for crypto changes. I didn't take the  time to look at the rest of this.

dimosr (dismissed):
LGTM (reviewed @corda/corda-platform-network-team owned files only)

vkolomeyko (approved):
LGTM

dimosr (approved):

dickon (commented):
LGTM for crypto

dickon (approved):

PR #3412, opened 2023-03-20
CORE-11786 Remove `SecureHash.getBytes`

kyriathar:
`SecureHash` API should be enough to users to just provide the hex string form of the hash.

- adds internal API `SecureHash.bytes` to replace runtime-os uses of public API `SecureHash.getBytes()`

api PR: [#964](https://github.com/corda/corda-api/pull/964)

### DIFF
--- a/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
+++ b/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
@@ -41,3 +41,10 @@ fun parseSecureHash(algoNameAndHexString: String): SecureHash {
         SecureHashImpl(algorithm, data)
     }
 }
+
+val SecureHash.bytes: ByteArray
+    get() =
+        (this as? SecureHashImpl)?.bytes
+            ?: throw IllegalArgumentException(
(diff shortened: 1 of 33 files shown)


libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
```diff
@@ -40,4 +40,11 @@ fun parseSecureHash(algoNameAndHexString: String): SecureHash {
         val data = ByteArrays.parseAsHex(value)
         SecureHashImpl(algorithm, data)
     }
-}
\ No newline at end of file
+}
+
+val SecureHash.bytes: ByteArray
+    get() =
```
mnesbit, line 46: This has to return a copy to protect from modification

kyriathar: I believe it does a return a copy of the bytes since `SecureHashImpl` extends `OpaqueBytes`.

yift-r3 (dismissed):
LGTM (network changes)

anien (dismissed):
LGTM!

yift-r3 (approved):

anien (approved):

driessamyn (approved):

PR #3413, opened 2023-03-20
CORE-9215 Clearing DLQ in StateAndEventSubscription and DurableSubscription after outputting items

ben-millar:
Jira: https://r3-cev.atlassian.net/browse/CORE-9215

#### Context:
As stated in the linked ticket, the in-memory DLQ was not being cleared in the `StateAndEventSubscriptionImpl` and `DurableSubscriptionImpl`.

Fixing this issue caused some integ tests to fail for both the `StateAndEventProcessor` and the `DurableProcessor` which uncovered some additional underlying issues with the `DBCordaConsumerImpl.poll()` whereby it was returning all records, even if they had been DLQ'd during serialization.

Additionally, the test processor used by the `DurableSubscriptionIntegrationTest` (`TestDurableDummyMessageProcessor`) was setup incorrectly for processing dummy records (which are `Wakeup()` events).

#### Changes:
+ Added line to clear `DeadLetterRecords` queues after they'd been submitted to the publisher.
+ Refactored `DBCordaConsumerImpl` with the help of @mbrkic-r3 to prune DLQ'd messages from results. We do this by filtering records which had a non-null value before serialization which became null following serialization, as this is how serialization errors are handled.
+ Added filter to `DBCordaConsumerImpl.poll()` to filter out null, non-DLQ messages.
+ Fixed <K, V> type in `TestDurableDummyMessageProcessor` by making `Wakeup` the expected value.

### DIFF
--- a/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImpl.kt
+++ b/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImpl.kt
@@ -169,18 +169,17 @@ internal class DBCordaConsumerImpl<K : Any, V : Any> constructor(
             it.transactionId.state == TransactionState.COMMITTED
         }
 
-        val result = dbRecords.map { dbRecord ->
-            CordaConsumerRecord(
-                dbRecord.topic,
-                dbRecord.partition,
-                dbRecord.recordOffset,
-                deserializeKey(dbRecord.key),
-                deserializeValue(dbRecord.value),
-                dbRecord.timestamp.toEpochMilli(),
-                headerSerializer.deserialize(dbRecord.headers ?: "{}")
-            )
-        }.filter {
-            it.value != null || it.topic.endsWith(".dlq")
+        val result = dbRecords.mapNotNull { dbRecord ->
+            deserializeValue(dbRecord.value)?.let {
+                CordaConsumerRecord(
+                    dbRecord.topic,
+                    dbRecord.partition,
+                    dbRecord.recordOffset,
+                    deserializeKey(dbRecord.key),
+                    it,
+                    dbRecord.timestamp.toEpochMilli()
+                )
+            }
         }
 
         if (dbRecords.isNotEmpty()) {
--- a/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImpl.kt
+++ b/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImpl.kt
@@ -180,7 +180,8 @@ internal class DBCordaConsumerImpl<K : Any, V : Any> constructor(
                     dbRecord.recordOffset,
                     deserializeKey(dbRecord.key),
                     deserializedValue,
-                    dbRecord.timestamp.toEpochMilli()
+                    dbRecord.timestamp.toEpochMilli(),
+                    headerSerializer.deserialize(dbRecord.headers ?: "{}")
                 )
             } else {
                 null
--- a/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImpl.kt
+++ b/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImpl.kt
@@ -8,7 +8,6 @@ import net.corda.messagebus.api.consumer.CordaConsumerRecord
 import net.corda.messagebus.api.consumer.CordaOffsetResetStrategy
 import net.corda.messagebus.db.configuration.ResolvedConsumerConfig
 import net.corda.messagebus.db.datamodel.CommittedPositionEntry
-import net.corda.messagebus.db.datamodel.TopicRecordEntry
 import net.corda.messagebus.db.datamodel.TransactionState
 import net.corda.messagebus.db.persistence.DBAccess
 import net.corda.messagebus.db.persistence.DBAccess.Companion.ATOMIC_TRANSACTION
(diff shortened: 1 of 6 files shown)


testing/message-patterns/src/integrationTest/kotlin/net/corda/messaging/integration/subscription/StateAndEventSubscriptionIntegrationTest.kt
```diff
@@ -445,7 +445,7 @@ class StateAndEventSubscriptionIntegrationTest {
         stateEventSub1.close()
 
         val durableLatch = CountDownLatch(10)
-        val dlqLatch = CountDownLatch(10)
+        val dlqLatch = CountDownLatch(5)
```
ben-millar: I found in testing that only 5 events are errored and sent to the DLQ. When this latch value was 10, this check was passing because the same 5 events were being processed multiple times (often more than 10 times).

That being said, I was unable to determine why we're only having 5 out of 10 events error here – I'm concerned that this behaviour might be incorrect! Would be great to get some insight from others here.

emilybowe (dismissed):
the changes LGTM in terms of CORE-9215 and what was described as the issue, but I agree with you that you may have uncovered something else here that also needs attention.

testing/message-patterns/src/integrationTest/kotlin/net/corda/messaging/integration/subscription/StateAndEventSubscriptionIntegrationTest.kt
mbrkic-r3: Problem seems to be that when deserialization of value in [DBCordaConsumerImpl](https://github.com/corda/corda-runtime-os/blob/d3423277751f8070ed65fa4b3b4fb2718d8174e5/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImpl.kt#L176) fails, [StateAndEventSubscriptionImpl](https://github.com/corda/corda-runtime-os/blob/411b9a6dc8bec4b5f7532615c855850cfa0877dd/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt#L132) will log error and add the record to `deadLetterRecords`, but the same record will still be returned by the `DBCordaConsumerImpl`, with value `null`.

Filtering out those records should fix the issue:
```kotlin
override fun poll(timeout: Duration): List<CordaConsumerRecord<K, V>> {
      val topicPartition = getNextTopicPartition() ?: return emptyList()
      val fromOffset = position(topicPartition)

      if (!recordsAvailable(fromOffset, topicPartition, timeout)) return emptyList()

      val dbRecords = dbAccess.readRecords(fromOffset, topicPartition, maxPollRecords).takeWhile {
          it.transactionId.state == TransactionState.COMMITTED
      }

      val result = dbRecords.map { dbRecord ->
          CordaConsumerRecord(
              dbRecord.topic,
              dbRecord.partition,
              dbRecord.recordOffset,
              deserializeKey(dbRecord.key),
              deserializeValue(dbRecord.value),
              dbRecord.timestamp.toEpochMilli()
          )
      }.filter { it.value != null }

      if (dbRecords.isNotEmpty()) {
          seek(topicPartition, dbRecords.last().recordOffset + 1)
      }
      return result
  }
```
However, I'm not sure whether we have code that expects to get back all records, including those with `null` value. This should also be consistent with Kafka implementation. @JamesHR3 Please advise.

Test could be improved to check what is actually received, not only the number of messages.

JamesHR3: I believe the Kafka implementation doesn't return the records with `null` in. It might be worth checking if the test passes when the integration tests are run against Kafka instead (there's instructions on how to do this in `testing:message-patterns`.)

libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImpl.kt
```diff
@@ -176,9 +176,12 @@ internal class DBCordaConsumerImpl<K : Any, V : Any> constructor(
                 deserializeValue(dbRecord.value),
                 dbRecord.timestamp.toEpochMilli()
             )
+        }.filter {
+            it.value != null || it.topic.endsWith(".dlq")
```
ben-millar: Is there a sensible place to add `".dlq"` as a constant?

mbrkic-r3: I don't think that business logic here should depend on the queue name. If such a feature is desirable it should be handled via factory parameters that will e.g. set a flag in constructed `DBCordaConsumerImpl` that would be used in filtering.
But, as mentioned above, all implementations should behave in the same way. So if Kafka is filtering out messages with `null` values then DB implementation should do it as well.
Also, when processing DLQ topic, we would most likely want to see original messages, i.e. with original values (e.g. to analyze what caused deserialization error), not messages with `null` values. With current implementation, DLQ processor would probably have to provide no-op serializers.
In case we want to keep messages with null values (in all implementations), we would need to modify serailizers to provide serialization result, not just silently return `null` value.

ben-millar: Great points; I think the correct fix here then is to change how the DLQ messages from the serialiser are constructed to prevent them from being null and therefore filtered out (and non-useful...)

testing/message-patterns/src/integrationTest/kotlin/net/corda/messaging/integration/processors/TestDurableProcessor.kt
```diff
@@ -58,13 +59,13 @@ class TestDurableStringProcessor(
 
 class TestDurableDummyMessageProcessor(
     private val latch: CountDownLatch, private val outputTopic: String? = null, private val delayProcessor: Long? = null
-) : DurableProcessor<String, DemoStateRecord> {
+) : DurableProcessor<String, Wakeup> {
```
ben-millar, line 62: See [getDummyRecords()](https://github.com/corda/corda-runtime-os/blob/release/os/5.0/testing/message-patterns/src/integrationTest/kotlin/net/corda/messaging/integration/TestUtils.kt#L32)

mbrkic-r3 (approved):
LGTM

PR #3415, opened 2023-03-20
CORE-11493: Add an option to pass `secureSSL` option to `RestClient`

vkolomeyko:
Changes been done to `RestClient` only for now.

CLI tool will be updated with `--insecure` option in the PR to follow.


filipesoliveira (approved):
LGTM

PR #3416, opened 2023-03-20
CORE-11792 remove party

blsemo:
API PR: https://github.com/corda/corda-api/pull/966
Remove the Party object from the API as it gives a false impression of an identity model that doesn't really exist.


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoTransactionBuilderImpl.kt
```diff
@@ -186,6 +190,9 @@ class UtxoTransactionBuilderImpl(
     @Suspendable
     private fun sign(): UtxoSignedTransaction {
         check(!alreadySigned) { "The transaction cannot be signed twice." }
+
+        require(notaryKey == null){ "The notary must not be set by user code"}
```
relyafi: It feels wrong to me that we need to do this check at all, and stems from the fact that we've added `notaryKey` to `UtxoTransactionBuilderInternal`. As we cannot set this at construction time, we could make `notaryKey` private, and add a getter. Preventing bad usage seems better than failing with a runtime exception.

Also, would it make sense to set this as part of `setNotary` rather than waiting until we sign? If someone does want to query the key then from a correctness perspective it should be available once the notary is set, not once the transaction is signed.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoBaselinedTransactionBuilder.kt
```diff
@@ -20,6 +20,8 @@ class UtxoBaselinedTransactionBuilder private constructor(
     private val currentTransactionBuilder: UtxoTransactionBuilderInternal,
 ) : UtxoTransactionBuilderInternal {
 
+    override var notaryKey: PublicKey? = null
```
relyafi: Not really sure why this class needs `notaryKey` given it's never actually used within this class? If we can call `setNotary` in this class then shouldn't it be set at that point (ties into my other comment in `UtxoTransactionBuilderImpl`)

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/filtered/factory/UtxoFilteredTransactionFactoryImpl.kt
```diff
@@ -41,7 +42,8 @@ class UtxoFilteredTransactionFactoryImpl @Activate constructor(
     ): UtxoFilteredTransaction {
         val notaryAndTimeWindow = if (filteredTransactionBuilder.notary || filteredTransactionBuilder.timeWindow) {
             ComponentGroupFilterParameters.AuditProof(NOTARY.ordinal, Any::class.java) {
-                filteredTransactionBuilder.notary && it is Party || filteredTransactionBuilder.timeWindow && it is TimeWindow
+                filteredTransactionBuilder.notary && (it is MemberX500Name || it is PublicKey ) // notary components
```
relyafi, line 45: Is the or on type correct? `PublicKey` isn't necessarily an identifier for the notary public key?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoTransactionBuilderVerifier.kt
```diff
@@ -35,6 +37,11 @@ class UtxoTransactionBuilderVerifier(private val transactionBuilder: UtxoTransac
         checkNotNull(transactionBuilder.notary) {
             "The notary of the current $subjectClass must not be null."
         }
+        checkNotNull(transactionBuilder.notaryKey) {
+            "The notary ${transactionBuilder.notary} of the current $subjectClass cannot be found or is not a valid notary."
```
relyafi: Should this message reflect that it is specifically the key that cannot be found?

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImplTest.kt
```diff
@@ -27,17 +29,23 @@ internal class UtxoSignedTransactionImplTest: UtxoLedgerTest() {
...
         signedTransaction = UtxoTransactionBuilderImpl(
-            utxoSignedTransactionFactory
+            utxoSignedTransactionFactory,
+            mockNotaryLookup
         )
             .setNotary(notary)
             .setTimeWindowBetween(utxoTimeWindowExample.from, utxoTimeWindowExample.until)
             .addOutputState(getUtxoStateExample())
-            .addSignatories(listOf(publicKeyExample))
+            .addSignatories(listOf(anotherPublicKeyExample))
```
relyafi, line 48: Why was this change needed?

components/ledger/notary-worker-selection-impl/src/main/kotlin/net/corda/ledger/notary/worker/selection/impl/NotaryVirtualNodeSelectorServiceImpl.kt
```diff
@@ -28,11 +28,11 @@ class NotaryVirtualNodeSelectorServiceImpl @Activate constructor(
...
      */
-    override fun selectVirtualNode(serviceIdentity: Party): Party {
+    override fun selectVirtualNode(serviceIdentity: MemberX500Name): MemberX500Name {
         val selectedMember = memberLookup.lookup().filter {
-            it.notaryDetails?.serviceName == serviceIdentity.name
+            it.notaryDetails?.serviceName == serviceIdentity
         }.random()
 
-        return Party(selectedMember.name, selectedMember.sessionInitiationKey)
+        return selectedMember.name
```
relyafi: Can now simplify and do away with `selectedMember` and just return.

libs/ledger/ledger-utxo-transaction-verifier/src/main/kotlin/net/corda/ledger/utxo/transaction/verifier/UtxoLedgerTransactionVerifier.kt
```diff
@@ -46,10 +46,11 @@ class UtxoLedgerTransactionVerifier(
         if (allInputs.isEmpty()) {
             return
         }
-        check(allInputs.map { it.notary }.distinct().size == 1) {
+        check(allInputs.map { Pair(it.notaryName, it.notaryKey) }.distinct().size == 1) {
             "Input and reference states' notaries need to be the same."
         }
-        check(allInputs.first().notary == transaction.notary) {
+        check(allInputs.first().notaryName == transaction.notaryName
+                && allInputs.first().notaryKey == transaction.notaryKey ) {
```
relyafi, line 53: Interesting, not sure this is new, but does this actually work? I think only the name can be guaranteed to be the same. The key could be different if:

- An additional notary worker is added, changing the public composite key.
- Key rotation occurs which replaces an existing part of the composite key.

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
```diff
@@ -105,11 +104,11 @@ class NonValidatingNotaryServerFlowImpl() : ResponderFlow {
             val otherMemberInfo = memberLookup.lookup(session.counterparty)
                 ?: throw IllegalStateException("Could not find counterparty on the network: ${session.counterparty}")
 
-            val otherParty = Party(otherMemberInfo.name, otherMemberInfo.sessionInitiationKey)
+            val otherPartySessionKey = otherMemberInfo.sessionInitiationKey
```
relyafi, line 107: I think we discussed that we should remove use of session key like this. Do we want to tackle in this PR, or raise another ticket?

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
```diff
@@ -180,10 +179,15 @@ class NonValidatingNotaryServerFlowImpl() : ResponderFlow {
     private fun extractParts(requestPayload: NonValidatingNotarizationPayload): NonValidatingNotaryTransactionDetails {
         val filteredTx = requestPayload.transaction as UtxoFilteredTransaction
         // The notary component is not needed by us but we validate that it is present just in case
-        requireNotNull(filteredTx.notary) {
+        requireNotNull(filteredTx.notaryName) {
             "Notary component could not be found on the transaction"
```
relyafi:
```suggestion
            "Notary name component could not be found on the transaction"
```

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
```diff
@@ -180,10 +179,15 @@ class NonValidatingNotaryServerFlowImpl() : ResponderFlow {
...
         }
 
+        requireNotNull(filteredTx.notaryKey) {
+            "Notary component could not be found on the transaction"
```
relyafi:
```suggestion
            "Notary key component could not be found on the transaction"
```

libs/ledger/ledger-utxo-transaction-verifier/src/main/kotlin/net/corda/ledger/utxo/transaction/verifier/UtxoLedgerTransactionVerifier.kt
blsemo: Yes - this is just replicating the existing logic - the logic is not quite right, though. We need a ticket to fix this after 5.0 but before key rotation becomes available.

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
blsemo: In a separate PR, please! I have raised https://r3-cev.atlassian.net/browse/CORE-11852

libs/ledger/ledger-utxo-transaction-verifier/src/main/kotlin/net/corda/ledger/utxo/transaction/verifier/UtxoLedgerTransactionVerifier.kt
blsemo: https://r3-cev.atlassian.net/browse/CORE-11853

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/filtered/factory/UtxoFilteredTransactionFactoryImpl.kt
blsemo: If there's an object in the notary component group, and it's of type `PublicKey`, then it's the notary service key? Or am I missing something here?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoTransactionBuilderVerifier.kt
blsemo: Not necessarily - if the key is null, then no notary info for this notary has been found.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImplTest.kt
blsemo: Long story... because we now look up the notary key under the hood, and the mockNotaryLookUp is configured to return the `publicKeyExample` for the default `notaryX500Name` to make all the other tests work, we have to add a different signature for the signatories - other wise the check if the notary has signed finds the signature from the signatory and confuses it with the notary signature...

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/filtered/factory/UtxoFilteredTransactionFactoryImpl.kt
relyafi: Hmm, I think I got confused that we now have two notary "components" (name and key) but are still retaining one component group. So I think this is probably ok.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoTransactionBuilderImpl.kt
```diff
@@ -150,15 +149,21 @@ class UtxoTransactionBuilderImpl(
...
-    override fun getNotary(): MemberX500Name? {
-        return notary
+    override fun getNotaryName(): MemberX500Name? {
+        return notaryName
     }
 
     override fun setNotary(notary: MemberX500Name): UtxoTransactionBuilder {
-        this.notary = notary
+        this.notaryName = notary
+        notaryKey = notaryLookup.lookup(notary)?.publicKey
```
relyafi: I'd use `this` for both `notaryName` and `notaryKey`, or not at all.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoTransactionBuilderContainer.kt
```diff
@@ -1,16 +1,18 @@
...
 import net.corda.v5.ledger.utxo.StateRef
 import net.corda.v5.ledger.utxo.TimeWindow
 import java.security.PublicKey
+import kotlin.reflect.jvm.internal.impl.descriptors.Visibilities.Public
 
 @CordaSerializable
 data class UtxoTransactionBuilderContainer(
-    private val notary: Party? = null,
+    private val notaryName: MemberX500Name? = null,
+    private val notaryKey: PublicKey? = null,
```
mnesbit: One of these should be (lazy) derived from signed group parameters, we shouldn't store both independently

relyafi (commented):
One minor comment, otherwise LGTM

mnesbit (dismissed):
OK

mnesbit (dismissed):
OK

relyafi (dismissed):

relyafi (dismissed):

mag1c-48 (dismissed):
LGTM - build only

relyafi (approved):

mnesbit (approved):
OK

mag1c-48 (approved):
LGTM - build

PR #3418, opened 2023-03-20
CORE-6545: Review HTTP REST error codes

emilybowe:
A review was made of implementations of the interfaces listed below to ensure that correct HTTP 4XX code is returned in the case of an error condition, as QA reported error code 5XX was returned for various operations.

REST resources in scope:
FlowClassRestResourceImpl
FlowRestResourceImpl
ConfigRestResourceImpl 

Ticket: [CORE-6545](https://r3-cev.atlassian.net/jira/software/c/projects/CORE/boards/691?modal=detail&selectedIssue=CORE-6545)

[CORE-6545]: https://r3-cev.atlassian.net/browse/CORE-6545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/flow/flow-rest-resource-service-impl/src/main/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImpl.kt
```diff
@@ -102,15 +105,15 @@ class FlowRestResourceImpl @Activate constructor(
         startFlow: StartFlowParameters
     ): ResponseEntity<FlowStatusResponse> {
         if (publisher == null) {
-            throw FlowRestResourceServiceException("FlowRestResource has not been initialised ")
+            throw ServiceUnavailableException("FlowRestResource has not been initialised ")
```
ben-millar: Nit: Could also move messages like this to the new `FlowRestExceptionConstants` file. Something like `UNINITIALIZED_ERROR`

ben-millar (dismissed):

components/flow/flow-rest-resource-service-impl/src/main/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImpl.kt
```diff
@@ -102,15 +105,15 @@ class FlowRestResourceImpl @Activate constructor(
...
         }
         if (fatalErrorOccurred) {
             // If Kafka has told us this publisher should not attempt a retry, most likely we have already been
             // replaced by another worker and have been "fenced". In that case it would be unsafe to create another
             // producer, because we'd attempt to replace our replacement. Most likely service orchestration has already
             // replaced us - nothing else should lead to us being fenced - and therefore should be responsible for
             // closing us down soon. There are other fatal error types, but none are recoverable by definition.
-            throw FlowRestResourceServiceException("Fatal error occurred, can no longer start flows from this worker")
+            throw InternalServerException(FlowRestExceptionConstants.FATAL_ERROR)
```
sprintengine: I think we could throw the `ServiceUnavailableException` here since, by the sounds of that comment above, this situation could occur when the service orchestrator has replaced this worker but not finished closing it down. Sounds like if the user retries they will reach the replacement worker and the request will succeed.

And the definition of a [503](https://httpwg.org/specs/rfc9110.html#status.503):
> The 503 (Service Unavailable) status code indicates that the server is currently unable to handle the request due to a temporary overload or scheduled maintenance, which will likely be alleviated after some delay. The server MAY send a [Retry-After](https://httpwg.org/specs/rfc9110.html#field.retry-after) header field ([Section 10.2.3](https://httpwg.org/specs/rfc9110.html#field.retry-after)) to suggest an appropriate amount of time for the client to wait before retrying the request.
 Note: The existence of the 503 status code does not imply that a server has to use it when becoming overloaded. Some servers might simply refuse the connection.

This sounds quite spot on with what we want to return to the user.

components/flow/flow-rest-resource-service-impl/src/main/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImpl.kt
```diff
@@ -182,26 +189,23 @@ class FlowRestResourceImpl @Activate constructor(
...
         fatalErrorOccurred = true
-        log.error("Fatal error occurred, FlowRestResource can no longer start flows, worker expected to terminate.", exception)
+        log.error(FlowRestExceptionConstants.FATAL_ERROR, exception)
         onFatalError()
-        return failureException(exception)
+        throw InternalServerException(FlowRestExceptionConstants.FATAL_ERROR)
     }
 
-    private fun failureException(cause: Exception): Exception =
-        FlowRestResourceServiceException("Failed to publish the Start Flow event.", cause)
```
sprintengine, line 203: You should be able to delete FlowRestResourceServiceException now

components/flow/flow-rest-resource-service-impl/src/main/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImpl.kt
```diff
@@ -182,26 +189,23 @@ class FlowRestResourceImpl @Activate constructor(
...
         }
-        waitOnPublisherFutures(recordFutures, PUBLICATION_TIMEOUT_SECONDS, TimeUnit.SECONDS) { ex, failureIsTerminal ->
-            if (failureIsTerminal) {
+        waitOnPublisherFutures(recordFutures, PUBLICATION_TIMEOUT_SECONDS, TimeUnit.SECONDS) { ex, failureIsFatal ->
+            if (failureIsFatal) {
                 throw markFatalAndReturnFailureException(ex)
             } else {
-                throw failureException(ex)
+                throw InternalServerException(FlowRestExceptionConstants.FATAL_ERROR)
             }
```
sprintengine, line 205: I'm not sure this is correct. If the failure is fatal, we mark fatal and return fatal failure, otherwise, we return a 500 with FATAL_ERROR message. 

This means we lose the error message returned in `ex` provided by `waitOnPublisherFutures`.

Looks to me like we should change the message to something like this:

```
            if (failureIsFatal) {
                throw markFatalAndReturnFailureException(ex)
            } else {
                val msg = ex.message ?: ""
                throw InternalServerException(
                    "Non-fatal error occurred when starting flow.",
                    mapOf("cause" to ex::class.java.simpleName, "reason" to msg)
                )
            }
```

conal-smith (approved):
lgtm

ben-millar (approved):
LGTM, love the constants 😁

PR #3419, opened 2023-03-20
CORE-11849: Prevent Kryo from accidentally serializing Avro classes.

chrisr3:
Remove some stray Avro classes from the stack so that Kryo cannot try to serialize them accidentally.

The latest Quasar SNAPSHOT also allows us to provide Quasar's `FiberSerializer` with the `ClassResolver` instance that we want its Kryo to use, which means we no longer need to reset this `final` field via Java reflection :relieved:.


lankydan (approved):
Flow related changes makes sense. Other stuff looks ok from my perspective.

fowlerrr (approved):
LGTM

PR #3421, opened 2023-03-20
CORE-11493: Make CLI commands SSL secure by default

vkolomeyko:
Usages within E2E tests been updated to use `--insecure` option as they are often executed against `localhost`.

Also, with my changes, I have tested locally.

Using CLI tool as before gives:
```
corda-cli network -u admin -p <pwd> -t https://localhost:8888 members-list -h 698A22FA9260
StdErr> Unable to verify server's SSL certificate. Please check the target parameter or use '--insecure' option.
```
Then, as suggested, there are two ways of improving the situation:
1. To use an endpoint with a valid SSL certificate:
```
corda-cli network -u admin -p <pwd> -t https://<UUID>.eu.ngrok.io members-list -h 698A22FA9260
```

or

2. Use `--insecure` option:
``` 
corda-cli network -u admin -p <pwd> -t https://localhost:8888 --insecure members-list -h 698A22FA9260
````


filipesoliveira (approved):
LGTM

dimosr (approved):

PR #3422, opened 2023-03-20
CORE-11681 Erroring session state before throwing exception when flow unconfirmed counterparty timeout reached

ben-millar:
This PR addresses [CORE-11681](https://r3-cev.atlassian.net/browse/CORE-11681).

#### Context:
Previously, changes were made [HERE](https://github.com/corda/corda-runtime-os/pull/3413) as part of CORE-9215 which errors sessions early if the counterparty doesn't materialise within some timeout window, however the sessionState was not updated to `ERROR` before erroring the session. We need to update this state first to provide quicker feedback to customers.

#### Changes:
Added an `errorSession(sessionState: String)` function to the `SessionManager` to allow for erroring of session state through the manager. This is now called in the FlowGlobabPostProcessor when we timeout a session for missing counterparty.

[CORE-11681]: https://r3-cev.atlassian.net/browse/CORE-11681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/flows/session-manager/src/main/kotlin/net/corda/session/manager/SessionManager.kt
+++ b/libs/flows/session-manager/src/main/kotlin/net/corda/session/manager/SessionManager.kt
@@ -3,6 +3,7 @@ package net.corda.session.manager
 import java.time.Instant
 import net.corda.data.flow.event.SessionEvent
 import net.corda.data.flow.state.session.SessionState
+import net.corda.data.flow.state.session.SessionStateType
 import net.corda.data.identity.HoldingIdentity
 import net.corda.libs.configuration.SmartConfig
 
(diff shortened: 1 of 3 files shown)


libs/flows/session-manager/src/main/kotlin/net/corda/session/manager/SessionManager.kt
```diff
@@ -84,4 +85,10 @@ interface SessionManager {
      */
     fun getMessagesToSend(sessionState: SessionState, instant: Instant, config: SmartConfig, identity: HoldingIdentity): Pair<SessionState,
             List<SessionEvent>>
+
+    /**
+     * Sets the [SessionState.status] of the passed [SessionState] to [SessionStateType.ERROR].
+     * @param sessionState The session state.
+     */
+    fun errorSession(sessionState: SessionState)
```
LWogan: I think we should stick to the existing pattern for session manager functions which is to return the SessionState object. 
Then after you return the SessionState object, set the SessionState back into the Checkpoint using the  checkpoint.putSessionState() function. 

I think (but not 100% sure) your change might not propagate past the toAvro() function in FlowStateManager? Either way lets stick to the patterns that exist for now, even if the "update in place" style function works. We migth need to refactor this stuff in the future and keeping it consistent will help that.

LWogan (approved):
lgtm

PR #3428, opened 2023-03-21
CORE-1068 Introduce Packaging Test Application

simon-johnson-r3:
This is a simple ledger test application which will form the basis of packaging testing for multiple versions of contracts. Its intended use is something along the lines of:

- Upload v1 of the app to Cluster B
- Create vnode Bob in Cluster B
- Upload v2 of the app to Cluster A
- Create vnode Alice in Cluster A
- v1 of the app will mint states for Bob on Cluster B, verifying v1 of the contract works for Bob
- Transfer states to Alice on Cluster A, who has v2 contract, which should fail
- Upload v1 of the app to Cluster A
- Transfer states to Alice on Cluster A, who now has access to v1 contract, so now should pass
- State transfers for each vnode can be tested with the "report states" flow

At the moment this is a single Cordapp. A clone must be made and built to create v2 of the application.

To mint states you execute the flow like so, where the list of integers is a list of state values to mint:
```
{
    "startFlow": {
        "clientRequestId" : "q3",
        "flowClassName" : "net.cordapp.testing.packagingverification.MintFlow",
        "requestBody" : "{ \"stateValues\" : [10, 20] }"
    }
}
```

To transfer states execute the following flow, specifying the issuer, recipient and the amount to transfer. Note there is little error checking so don't try to trick the application:
```
{
    "startFlow": {
        "clientRequestId" : "p1",
        "flowClassName" : "net.cordapp.testing.packagingverification.TransferStatesFlow",
        "requestBody" : "{ \"issuerX500Name\" : \"C=GB, L=London, O=Bob\", \"recipientX500Name\" : \"C=GB, L=London, O=Alice\", \"value\" : 10 }"
    }
}
```

To report states in any vnode, execute the following flow:
```
{
    "startFlow": {
        "clientRequestId" : "t5",
        "flowClassName" : "net.cordapp.testing.packagingverification.ReportStatesFlow",
        "requestBody" : "{}"
    }
}
```


testing/cpbs/packaging-verification-app-v1/src/main/kotlin/net/cordapp/testing/packagingverification/MintFlow.kt
```diff
@@ -0,0 +1,86 @@
...
+    @Suspendable
+    override fun call(requestBody: ClientRequestBody): String {
+        log.info("Minting states")
+
+        val myInfo = memberLookup.myInfo()
+
+        val mintRequest = requestBody.getRequestBodyAs(jsonMarshallingService, MintRequest::class.java)
+        if (mintRequest.stateValues.isEmpty()) {
+            throw CordaRuntimeException("No state values to mint")
+        }
```
fowlerrr: typically I think we use `require` here.  Not terribly important but IIRC it's sort of a convention.

testing/cpbs/packaging-verification-app-v1/src/main/kotlin/net/cordapp/testing/packagingverification/ReceiveTransferFlow.kt
```diff
@@ -0,0 +1,35 @@
...
+    @CordaInject
+    lateinit var utxoLedgerService: UtxoLedgerService
+
+    override fun call(session: FlowSession) {
+        log.info("Receiving states")
+
+        utxoLedgerService.receiveFinality(session) { ledgerTransaction ->
+            log.info("Validating finality with ${ledgerTransaction.outputContractStates.size} output states")
+
+            if (ledgerTransaction.outputContractStates.size > 2) {
```
fowlerrr: Again, `require` for these is the convention, I think.

fowlerrr (commented):
Couple of very small things

driessamyn:
This looks good to me.
Probably worth getting a review from @blsemo or somebody in his team from a ledger perspective.

simon-johnson-r3:
> This looks good to me. Probably worth getting a review from @blsemo or somebody in his team from a ledger perspective.

Added @blsemo, if you want to nominate someone else from your team please go ahead.

testing/cpbs/packaging-verification-app-v1/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+}
+
+dependencies {
+    cordaProvided platform("net.corda:corda-api:$cordaApiVersion")
+    cordaProvided 'net.corda:corda-application'
+    cordaProvided 'org.jetbrains.kotlin:kotlin-osgi-bundle'
+
+    cordapp project(':testing:cpbs:packaging-verification-contract-v1')
+    cordapp project(':notary-plugins:notary-plugin-non-validating:notary-plugin-non-validating-client')
+    cordapp project(':notary-plugins:notary-plugin-non-validating:notary-plugin-non-validating-server')
```
blsemo, line 22: Why do you need the notary server in your CorDapp? That looks a  bit dodgy.
@nkovacsx @relyafi Do you have an opinion on this?

testing/cpbs/packaging-verification-app-v1/src/main/kotlin/net/cordapp/testing/packagingverification/MintFlow.kt
```diff
@@ -0,0 +1,82 @@
...
+    override fun call(requestBody: ClientRequestBody): String {
+        log.info("Minting states")
+
+        val myInfo = memberLookup.myInfo()
+
+        val mintRequest = requestBody.getRequestBodyAs(jsonMarshallingService, MintRequest::class.java)
+        require(mintRequest.stateValues.isNotEmpty()) { "No state values to mint" }
+
+        val myPublicKey = myInfo.ledgerKeys.first()
+        val notary = notaryLookup.notaryServices.single()
```
blsemo: Don't use single - this may break surprisingly should a network introduce more than one notary.
Either have an explicit choice of notary (e.g. via the request or config), or use `.first()`.

testing/cpbs/packaging-verification-app-v1/src/main/kotlin/net/cordapp/testing/packagingverification/MintFlow.kt
```diff
@@ -0,0 +1,82 @@
...
+
+        mintRequest.stateValues.forEach { stateValue ->
+            log.info("Minting state with value $stateValue")
+            mintState(stateValue, myPublicKey, notary, myInfo.name)
```
blsemo: Are you calling this in a loop to cause load? In which case, it might be fine.
Otherwise, I would recommend to min many states in one transaction.

testing/cpbs/packaging-verification-app-v1/src/main/kotlin/net/cordapp/testing/packagingverification/TransferStatesFlow.kt
```diff
@@ -0,0 +1,131 @@
...
+            val session = flowMessaging.initiateFlow(recipientX500)
+            log.info("Finalizing transaction")
+            utxoLedgerService.finalize(signedTransaction, listOf(session))
+            log.info("Finalized transaction")
+        } catch (ex: Exception) {
+            log.info("TransferStatesFlow failed", ex)
+        }
+        finally {
+            // Release the claim on the tokens' states
+            tokenClaim?.useAndRelease(tokenClaim.claimedTokens.map { it.stateRef })
```
blsemo: I think that is wrong - if you've thrown a transaction, you shouldn't use the tokens, just release them.

testing/cpbs/packaging-verification-contract-v1/src/main/kotlin/net/cordapp/testing/packagingverification.contract/SimpleCommand.kt
```diff
@@ -0,0 +1,5 @@
+package net.cordapp.testing.packagingverification.contract
+
+import net.corda.v5.ledger.utxo.Command
+
+class SimpleCommand : Command
```
blsemo: I would recommend having a `Mint` and a `Transfer` command so you can verify the transaction meaningfully.

testing/cpbs/packaging-verification-contract-v1/src/main/kotlin/net/cordapp/testing/packagingverification.contract/SimpleContract.kt
```diff
@@ -0,0 +1,15 @@
...
+import net.corda.v5.ledger.utxo.transaction.UtxoLedgerTransaction
+import org.slf4j.LoggerFactory
+
+class SimpleContract : Contract {
+    private companion object {
+        val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
+    }
+
+    override fun verify(transaction: UtxoLedgerTransaction) {
+        log.info("Verify called on SimpleContract")
```
blsemo, line 13: This should verify your ledger transaction:
E.g. for a mint that there's not input and all coins have the same issuer.
For a transfer, check that there's 1 or 2 outputs, and that the sum of inputs is the same as the sum of outputs.

testing/cpbs/packaging-verification-app-v1/src/main/kotlin/net/cordapp/testing/packagingverification/TransferStatesFlow.kt
```diff
@@ -0,0 +1,131 @@
...
+        val counterpartyMember = memberLookup.lookup(recipientX500)
+            ?: throw IllegalArgumentException("Member info cannot be found for $recipientX500")
+
+        val notary = notaryLookup.notaryServices.single()
+
+        val myInfo = memberLookup.myInfo()
+        // Tests use dynamic names for vnodes to avoid clashes so we have to treat the vnode running this flow as the
+        // issuer, we cannot know it in advance. This means this flow will only work executed as the same vnode the mint
+        // flow was executed in. If the test was ever extended the issuer would have to be passed to the flow request body.
+        val issuerName = myInfo.name
```
blsemo: Hmm, the issuer should be configured somehow, this is super fragile and unrealistic.
The mint flow could e.g. return the issuer X500 name and you put that in all subsequent requests? That way, you would be able to pass tokens around from nodes that received them rather than minted them.
For creating outputs/change, you'd use the same issuer as the inputs (and all inputs need the same issuer), and the contract needs to check you don't change the issuer when transferring tokens.

blsemo (commented):
This app has a few issues where it does not behave very UTXO-y. Depending on how important that is for your tests, it's either ok, and we need a rewrite once the advanced state model is available, or we need to have a chat about how to improve this.

testing/cpbs/packaging-verification-app-v1/build.gradle
simon-johnson-r3: It eases testing as you only need to upload one CPI, there's no deeper reason. The automated execution of this application cannot yet be implemented due to other blockers, it's not entirely clear yet whether the test execution framework will take over responsibility here.

testing/cpbs/packaging-verification-app-v1/src/main/kotlin/net/cordapp/testing/packagingverification/TransferStatesFlow.kt
simon-johnson-r3: My understanding is if the transaction is successful the states these tokens represent can never be claimed again because they are of course spent at that point. If the transaction was not successful (because of a throw), these tokens are released here and the states are available to use again. I checked this with @owenstanford prior to use, but I'll double check to make sure.

There is no other mutable method available to call on TokenClaim, only a getter which returns immutable objects.

testing/cpbs/packaging-verification-contract-v1/src/main/kotlin/net/cordapp/testing/packagingverification.contract/SimpleContract.kt
simon-johnson-r3: I've put a couple of examples in, and commented other steps would be done in a real contract. This isn't important to the test, but if any one looks at the application at least they will get a feel for what should be in here.

testing/cpbs/packaging-verification-app-v1/src/main/kotlin/net/cordapp/testing/packagingverification/TransferStatesFlow.kt
simon-johnson-r3: I know the issuer in advance so I don't need to return it, I just didn't want the overhead of passing more parameters in the test. However you have convinced me, so I moved this to the transferRequest and I can delete the comment.

testing/cpbs/packaging-verification-app-v1/src/main/kotlin/net/cordapp/testing/packagingverification/TransferStatesFlow.kt
owenstanford: Sorry if I gave you bad advice, this is not quite correct, the useAndRelease takes a list of spent tokens (releasing any that were unspent) in this case, where we plan to spend everything, we should use `tokenClaim?.useAndRelease(tokenClaim.claimedTokens.map { it.stateRef })` if the TX was successful `tokenClaim?.useAndRelease(listOf())`  if it failed.

simon-johnson-r3: Ah ok, thanks!

testing/cpbs/packaging-verification-app-v1/build.gradle
relyafi: My advice would be that we shouldn't be doing this, and should have two separate CPIs. Doing this is not especially onerous as there are already examples of doing this in our other tests, e.g. smoke tests. This would be a more realistic test case.

When we discussed verbally, @simon-johnson-r3 did say this was for expedience. However I would caution that were we to run into problems due to this setup, our first answer would likely be this is not a supported configuration, make the setup realistic.

driessamyn (approved):
LGTM

mcgovc (approved):
LGT - build changes.

PR #3431, opened 2023-03-21
CORE-11868: Remove CryptoConnectionsFactory from use

fowlerrr:
Remove all use of `CryptoConnectionsFactory` as it's no longer required.


components/crypto/crypto-persistence-impl/src/main/kotlin/net/corda/crypto/persistence/impl/SigningKeyStoreImpl.kt
```diff
@@ -112,22 +121,31 @@ class SigningKeyStoreImpl @Activate constructor(
     override fun lookupByFullIds(tenantId: String, fullKeyIds: List<SecureHash>): Collection<SigningCachedKey> =
         impl.lookupByFullKeyIds(tenantId, fullKeyIds.toSet())
 
+    private fun entityManager(tenantId: String) = getEntityManagerFactory(
```
dickon: I wonder if this should be called createEntityManager or entityManagerFactory? Trouble is we have so many factories. Right now this reads like an attribute, but it isn't since it takes tenantId as name

components/crypto/crypto-persistence-impl/src/integrationTest/kotlin/net/corda/crypto/persistence/impl/tests/PersistenceTests.kt
```diff
@@ -159,8 +161,9 @@ class PersistenceTests {
 
         private fun randomTenantId() = publicKeyIdFromBytes(UUID.randomUUID().toString().toByteArray())
 
-        private fun cryptoDbEmf(): EntityManagerFactory = cryptoConnectionsFactory.getEntityManagerFactory(
-            CryptoTenants.CRYPTO
+        private fun cryptoDbEmf(): EntityManagerFactory = getEntityManagerFactory(
```
dickon, line 164: I thoguth we'd end up constructing a  CryptoRepository here, but it doesn't matter too much in test code.

components/crypto/crypto-persistence-impl/src/main/kotlin/net/corda/crypto/persistence/impl/HSMStoreImpl.kt
```diff
@@ -1,43 +1,59 @@
...
-        setOf(LifecycleCoordinatorName.forComponent<CryptoConnectionsFactory>())
+        setOf(
+            LifecycleCoordinatorName.forComponent<DbConnectionManager>(),
+            LifecycleCoordinatorName.forComponent<VirtualNodeInfoReadService>(),
+        )
     )
 ), HSMStore {
 
-    override fun createActiveImpl(): Impl = Impl(connectionsFactory)
+    override fun createActiveImpl(): Impl = Impl(getEntityManagerFactory(
```
dickon, line 51: Again, I was thinking this would be constructing a CryptoRepository and using that here.

Having said that this approach could be clean if we do it consistently, it's just not clear that we want both this approach and CryptoRepository.

components/crypto/crypto-persistence-impl/src/main/kotlin/net/corda/crypto/persistence/impl/SigningKeyStoreImpl.kt
fowlerrr: Yeah, I struggled with that.  Originally it would have been `entityManagerFactory()` but had the same problem.  I'll make it `createEntityManager()`

components/crypto/crypto-persistence-impl/src/main/kotlin/net/corda/crypto/persistence/impl/HSMStoreImpl.kt
fowlerrr: The thing here is that it's an `HSMStore` that uses a `CryptoRepository` which is, effectively, an _HSMStore_.  So I decided to avoid that circular logic.  My hope is that we'll be able to trash this class soon enough anyway.

dickon: Okay, for now.

dickon (approved):
LGTM, as a step forward.

PR #3432, opened 2023-03-21
CORE-11502: use new schema for wrapping keys, signing keys,  and config schema

dickon:
Behaviour change: use the tenant ID specified by the context parameter when creating a wrapping key, rather than always storing wrapping keys in the CRYPTO tenant in the cluster datbaase. 

Split out wrapping key material into its own table so that we can store multiple representations at once, for later use during key rotation when that is implemented. (Allowing easy joins motivated the behaviour change on where the wrapping keys are stored, but the behaviour change itself is appropriate.)

Add a rotation date column in the wrapping keys, initially set to a long time in the future. Nothing yet checks the rotation date.

Allow a set of root wrapping keys to be passed into the crypto processor, with one selected to be the default. This required supporting secrets in arrays during validation.

Now that we have removed all OSGi and Lifecycle logic from the repository layer of the crypto processor it it possible to have fast integration tests that consturct an in-memory database directly and exercise the repository layer code, and can be debugged without needing to go multi-process. So, rework the RepositoryTests like that. Take the opportunity to group them according to the repository class they are testing, in the standard structure.

Stop allowing querying the signing repository by master key alias (aka the warpping key), since it is possible with the new strucutre that there are multiple master keys for the same signing key.

Use data classes for the *Info DTOs, and take care over their equqals and hashCode methods where ByteArray is concerned. 

Make the repository operations return the DTO, to align better with other JPA code in Corda. This might sometimes make calling code more compact since it can use the Fluent style.

Disable certain PersistenceTests for now, which are mostly duplicates of the fast new non-OSGi integration tests.

Disable unit tests for the SigningRepository, since their mocking requires updates.

Add some tracing code (at trace log level) for wrapping key handling.

Fix typo in TestWrappingRepoitory filename.


components/crypto/crypto-persistence-model/src/main/kotlin/net/corda/crypto/persistence/db/model/SigningKeyEntity.kt
```diff
@@ -22,20 +20,24 @@ import javax.persistence.Table
...
 @Table(name = DbSchema.CRYPTO_SIGNING_KEY_TABLE)
-@IdClass(SigningKeyEntityPrimaryKey::class)
 @Suppress("LongParameterList")
 class SigningKeyEntity(
+    /*
+     * Synthetic UUID of the key pair
+     */
+    @Id
+    @Column(name = "id", nullable = false, updatable = false)
+    var id: UUID,
```
driessamyn: I think the `id` one is the only one that can be `val` rather than `var` (probably doesn't matter too much though).

components/crypto/crypto-persistence-model/src/main/kotlin/net/corda/crypto/persistence/db/model/SigningKeyEntity.kt
```diff
@@ -22,20 +20,24 @@ import javax.persistence.Table
...
 @Entity
 @Table(name = DbSchema.CRYPTO_SIGNING_KEY_TABLE)
-@IdClass(SigningKeyEntityPrimaryKey::class)
 @Suppress("LongParameterList")
 class SigningKeyEntity(
+    /*
+     * Synthetic UUID of the key pair
+     */
+    @Id
+    @Column(name = "id", nullable = false, updatable = false)
```
driessamyn: I think `updatable = false` is redundant here (but probably does no harm)

components/crypto/crypto-persistence-model/src/main/kotlin/net/corda/crypto/persistence/db/model/SigningKeyEntity.kt
```diff
@@ -48,8 +50,14 @@ class SigningKeyEntity(
...
-    @Column(name = "timestamp", nullable = false, updatable = false)
-    var timestamp: Instant,
+    @Column(name = "created", nullable = false)
+    var created: Instant,
+
+    /**
+     * When the key was last accessed.
+     */
+    @Column(name = "accessed", nullable = false)
+    var accessed: Instant,
```
driessamyn: As per API PR. I really don't think this should be on this table.

It means it has to be updated each time it is queried, but it is also confusing. Do we count, for example, the times the key was used from cache?

components/crypto/crypto-persistence-model/src/main/kotlin/net/corda/crypto/persistence/db/model/SigningKeyMaterialEntity.kt
```diff
@@ -0,0 +1,62 @@
...
+     * Time signigng key material was created
+     */
+    @Column(name = "created", nullable = false)
+    var created: Instant,
+
+    /**
+     * Last time this signing key material was unwrapped
+     */
+    @Column(name = "accessed", nullable = false)
+    var accessed: Instant,
```
driessamyn: As above.

components/crypto/crypto-persistence-model/src/main/kotlin/net/corda/crypto/persistence/db/model/WrappingKeyEntity.kt
```diff
@@ -14,21 +15,34 @@ import javax.persistence.Table
  * The records are immutable.
  */
 @Entity
+@Suppress("LongParameterList")
 @Table(name = DbSchema.CRYPTO_WRAPPING_KEY_TABLE)
 class WrappingKeyEntity(
+    @Id
+    @Column(name = "id", nullable = false, updatable = false, length = 64)
```
driessamyn: length and updatable are redundant.

Probably doesn't do harm, but length is not consistent with the other UUID columns.

components/crypto/crypto-persistence-model/src/main/kotlin/net/corda/crypto/persistence/db/model/WrappingKeyEntity.kt
```diff
@@ -42,11 +56,30 @@ class WrappingKeyEntity(
     var algorithmName: String,
 
     /**
-     * Key material for the wrapping key. It's encrypted by by another key which is obtained through the configuration.
+     * Key material for the wrapping key. It's encrypted by another wrapping key.
      */
     @Column(name = "key_material", nullable = false, updatable = false, columnDefinition = "BLOB")
```
driessamyn, line 55: `columnDefinition = "BLOB"` is incorrect here because the table definition is a VARBINARY.
BLOB type is different, can be off any size and in most DBs is stored differently.

components/crypto/crypto-persistence/src/main/kotlin/net/corda/crypto/persistence/SigningWrappedKeySaveContext.kt
```diff
@@ -6,7 +6,7 @@ import net.corda.crypto.cipher.suite.schemes.KeyScheme
 @Suppress("LongParameterList")
 class SigningWrappedKeySaveContext(
     val key: GeneratedWrappedKey,
-    val masterKeyAlias: String?,
+    val masterKeyAlias: String?, // TODO rename wrappingKeyAlias
```
driessamyn, line 9: Did you mean to leave todo in here?

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/infra/TestSigningRepository.kt
```diff
@@ -114,13 +114,11 @@ class TestSigningRepository: SigningRepository {
             SigningKeyOrderBy.CATEGORY -> filtered.sortedBy { it.category }
             SigningKeyOrderBy.SCHEME_CODE_NAME -> filtered.sortedBy { it.schemeCodeName }
             SigningKeyOrderBy.ALIAS -> filtered.sortedBy { it.alias }
-            SigningKeyOrderBy.MASTER_KEY_ALIAS -> filtered.sortedBy { it.masterKeyAlias }
```
driessamyn, line 117: Not really specific to this PR, but this class is a bit weird. You're basically re-implementing DB behaviour which probably doesn't have much value. 
All this should be covered by integration tests against the in-memory DB, and the repository itself should be able to be mocked out.
I don't think we don anything like this elsewhere.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
```diff
@@ -183,37 +245,51 @@ class SigningRepositoryImpl(
...
         tenantId = tenantId,
         category = category,
         alias = alias,
         hsmAlias = hsmAlias,
         publicKey = publicKey,
-        keyMaterial = keyMaterial,
+        keyMaterial = signingKeyMaterialEntity?.keyMaterial,
         schemeCodeName = schemeCodeName,
-        masterKeyAlias = masterKeyAlias,
+        masterKeyAlias = wrappingKey?.alias,
```
driessamyn, line 284: When are these ever null?

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
```diff
@@ -72,72 +79,124 @@ class SigningRepositoryImpl(
...
+                    .setMaxResults(1).resultList.singleOrNull()
+                    ?.let {
+                        it.id
+                    }
+            } ?: throw InvalidParamsException("unable to find master wrapping key ${context.masterKeyAlias}")
+        }
+
+        val materialEntity = SigningKeyMaterialEntity(
+            signingKeyId = keyMainId,
+            wrappingKeyId = wrappingKeyId,
```
driessamyn, line 121: I'm surprised this works because the EM has been closed. If you persist a graph that has a one-to-many or many-to-one relationship, you should do this within a single EM so that they exist in the same session.

I think this works because you are setting the ID, but the JPA way would be to set the the `WrappingKey` object, not the ID.

components/crypto/crypto-softhsm-impl/src/test/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryTest.kt
```diff
@@ -154,37 +168,46 @@ class SigningRepositoryTest {
...
-            mock { on { hash(any<ByteArray>(), any()) } doReturn hash },
-            mock(),
-        )
-
-        repo.lookupByPublicKeyHashes(keys)
-
-        assertThat(tenantCap.allValues.single()).isEqualTo("tenant")
-        assertThat(fullIdsCap.allValues.single()).isEqualTo(listOf(hashA.toString(), hashB.toString()))
-    }
+//    @Test
```
driessamyn: Is this meant to be commented out?

libs/crypto/crypto-config-impl/src/main/kotlin/net/corda/crypto/config/impl/CryptoConfigUtils.kt
```diff
@@ -271,6 +272,10 @@ fun createDefaultCryptoConfig(wrappingKeyPassphrase: Any, wrappingKeySalt: Any):
                 )
             )
         )
+        .withValue("defaultWrappingKey", ConfigValueFactory.fromAnyRef("root1"))
```
driessamyn: Is this right?

components/crypto/crypto-persistence-impl/src/integrationTest/kotlin/net/corda/crypto/persistence/impl/tests/PersistenceTests.kt
driessamyn, line 1: Does this need to be an OSGi test?

Is in or purpose a lot of the tests are commented out?

driessamyn (commented):
Where are the repository integration tests?

dickon:
I have reviewed and developed further the additions from @driessamyn yesterday.

components/crypto/crypto-persistence-model/src/main/kotlin/net/corda/crypto/persistence/db/model/SigningKeyEntity.kt
dickon: Removed.

components/crypto/crypto-persistence-model/src/main/kotlin/net/corda/crypto/persistence/db/model/SigningKeyMaterialEntity.kt
dickon: Deleted

components/crypto/crypto-persistence-model/src/main/kotlin/net/corda/crypto/persistence/db/model/WrappingKeyEntity.kt
dickon: deleted

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
dickon: Good question. In addition to storing the full signing key pairs, the crypto processor, since Alexey's time, has supported storing only public keys for signing keys which means that the key material is not available, and we don't know the wrapping key for it (masterKeyAlias as it is called in the code). That's what `SigningRepository.savePublicKey` is used for. 

I suppose this may only be necessary when storing data about keys which are handled on an external HSM, which we don't really fully support yet, but I've not taken the time yet to prove that to myself and rip it all out.

components/crypto/crypto-softhsm-impl/src/test/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryTest.kt
dickon: Not sure, but since you commented out the whole file it is a bit irrelevant now.

components/crypto/crypto-persistence-impl/src/integrationTest/kotlin/net/corda/crypto/persistence/impl/tests/PersistenceTests.kt
dickon: Many of them need updating. With the direction now of implementing non-OSGi versions of these tests we may end up deleting them since they are slower and duplicate new tests.

libs/crypto/crypto-config-impl/src/main/kotlin/net/corda/crypto/config/impl/CryptoConfigUtils.kt
dickon: Deleted. (I'd moved it to a new path in the config, but it looks like a merge error or something similar bought it back).

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
dickon: Yeah, maybe we do want to have a relationship here, but what we have now works at this point, so maybe better to get this merged then go back and add the relationships.

components/crypto/crypto-persistence-model/src/main/kotlin/net/corda/crypto/persistence/db/model/SigningKeyEntity.kt
dickon: fixed

components/crypto/crypto-persistence-model/src/main/kotlin/net/corda/crypto/persistence/db/model/SigningKeyEntity.kt
dickon: fixed

driessamyn (commented):
LGTM

PR #3436, opened 2023-03-21
CORE-9596 Vault named query parsing

lankydan:
Adds vault named query parsing of a Postgres syntax into an expression and back into a Postgres query to be executed at some point.

`VaultNamedQueryParser` provides the entry point for parsing a query. The implementation of this interface then delegates to a `VaultNamedQueryExpressionParser` to convert a query string into an expression, a `VaultNamedQueryExpressionValidator` to validate the expression and a `VaultNamedQueryConverter` that converts the expression into an output query string.

The `PostgresVaultNamedQueryExpressionParser` converts a Postgres query string into an expression, and `PostgresVaultNamedQueryConverter` converts that expression into a Postgres query string.

These interfaces give us the flexibility to parse from different formats and output queries for different databases.

### DIFF
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/impl/VaultNamedQueryBuilderUtils.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/impl/VaultNamedQueryBuilderUtils.kt
@@ -1,0 +1,18 @@
+package net.corda.ledger.persistence.query.impl
+
+import org.slf4j.LoggerFactory
+
+private val log = LoggerFactory.getLogger("net.corda.ledger.persistence.query.impl.VaultNamedQueryBuilderUtils")
+
+fun logQueryRegistration(name: String, query: VaultNamedQuery.ParsedQuery) {
+    when {
+        log.isTraceEnabled -> log.trace(
+            "Registering vault named query with name: $name, original query: ${query.originalQuery.replace(
+                    "\n",
+                    " "
+                )
+            }, parsed query: ${query.query}"
+        )
+        log.isDebugEnabled -> log.debug("Registering vault named query with name: $name")
+    }
+}
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/impl/parsing/expressions/PostgresVaultNamedQueryExpressionParser.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/impl/parsing/expressions/PostgresVaultNamedQueryExpressionParser.kt
@@ -1,15 +1,33 @@
 package net.corda.ledger.persistence.query.impl.parsing.expressions
 
+import net.corda.ledger.persistence.query.impl.parsing.And
+import net.corda.ledger.persistence.query.impl.parsing.As
+import net.corda.ledger.persistence.query.impl.parsing.Equals
+import net.corda.ledger.persistence.query.impl.parsing.From
+import net.corda.ledger.persistence.query.impl.parsing.GreaterThan
+import net.corda.ledger.persistence.query.impl.parsing.GreaterThanEquals
+import net.corda.ledger.persistence.query.impl.parsing.In
+import net.corda.ledger.persistence.query.impl.parsing.IsNotNull
+import net.corda.ledger.persistence.query.impl.parsing.IsNull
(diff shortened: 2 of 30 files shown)


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/impl/VaultNamedQueryBuilderUtils.kt
```diff
@@ -0,0 +1,18 @@
...
+fun logQueryRegistration(name: String, query: VaultNamedQuery.ParsedQuery) {
+    when {
+        log.isTraceEnabled -> log.trace(
+            "Registering vault named query with name: $name, original query: ${query.originalQuery.replace(
+                    "\n",
+                    " "
+                )
+            }, parsed query: ${query.query}"
+        )
+        log.isDebugEnabled -> log.debug("Registering vault named query with name: $name")
```
blsemo, line 16: I am not sure I like this behaviour where the DEBUG message disappears when moving the log level to TRACE.
Going to trace should add stuff, not change higher level log messages.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/impl/parsing/expressions/PostgresVaultNamedQueryExpressionParser.kt
```diff
@@ -0,0 +1,164 @@
...
+    private val numberPattern = Regex(
+        """(?<num>[0-9]+(\.[0-9]+)?([eE]-?[0-9]+)?)"""
+    )
+
+    @Suppress("MaxLineLength")
+    private val opsPattern = Regex(
+        """(?<op>(->>)|[+-/*=?]|<(=)?|>(=)?|==|!(=)?|(?i)\bas\b|(?i)\bfrom\b|(?i)\bselect\b|(?i)\bwhere\b|(?i)\band\b|(?i)\bor\b|(?i)\bis null\b|(?i)\bis not null\b|(?i)\bin\b|(?i)\blike\b)"""
+    )
+
+    private val jsonCastPattern = Regex("""(?<cast>::.*?)((->>)|[+*=]|&&|\|\||<(=)?|>(=)?|==|!(=)?|\s)""")
```
blsemo:
```suggestion
    private val jsonCastPattern = Regex("""::(?<cast>.*?)((->>)|[+*=]|&&|\|\||<(=)?|>(=)?|==|!(=)?|\s)""")
```
That way you get the colon removed.

blsemo: I believe if you start your regexes with `^` they will only match at the given start position, so you can drop the index == startPosition.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/impl/VaultNamedQueryBuilderUtils.kt
lankydan: Mmmm, I don't think there is a nicer way to do this, because I don't want the extra information of the query from the trace message to be far away from the debug log if both were output. Plus wouldn't that make it seem like the same query was registered twice if I logged both of them?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/impl/parsing/expressions/PostgresVaultNamedQueryExpressionParser.kt
lankydan: > I believe if you start your regexes with ^ they will only match at the given start position, so you can drop the index == startPosition.

I couldn't get this working myself.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/impl/VaultNamedQueryBuilderUtils.kt
blsemo: Ok, fine

blsemo (approved):
LGTM

kyriathar (approved):
LGTM (db and packaging files)

PR #3439, opened 2023-03-21
CORE-11833 - add metric for fiber serialization and deserialization

conal-smith:
Acquired the following data from these metrics:

image

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberImpl.kt
@@ -136,7 +136,7 @@ class FlowFiberImpl(
             log.trace { "Parking..." }
             val fiberState = CordaMetrics.Metric.FlowFiberSerializationTime.builder()
                 .forVirtualNode(getExecutionContext().flowCheckpoint.holdingIdentity.shortHash.toString())
-                .withTag(CordaMetrics.Tag.FlowId, getExecutionContext().flowCheckpoint.flowId)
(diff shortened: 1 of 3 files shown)


components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberImpl.kt
```diff
@@ -224,8 +231,6 @@ class FlowFiberImpl(
         getExecutionContext().currentSandboxGroupContext.remove()
     }
 
-    private fun Throwable.isUnrecoverable(): Boolean = this is VirtualMachineError && this !is StackOverflowError
```
sprintengine, line 227: This is unused, intellij checks wouldn't let me make this commit without removing this!

libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
```diff
@@ -46,6 +46,16 @@ object CordaMetrics {
...
 
+        /**
+         * Metric for flow fiber serialization.
+         */
+        object FlowFiberSerializationTime : Metric<Timer>("flow.fiber.serialization.time", Metrics::timer)
+
+        /**
+         * Metric for flow fiber deserialization.
+         */
+        object FlowFiberDeserializationTime : Metric<Timer>("flow.fiber.deserialization.time", Metrics::timer)
```
sprintengine, line 57: Decided to have two separate metrics for serialization and deserialization, rather than one "flow.fiber.time" metric with "serialization" and "deserialization" direction labels.

davidcurrie (dismissed):
LGTM

JamesHR3 (approved):

PR #3440, opened 2023-03-21
CORE-11886 Reduce plugin log volume

davidcurrie:


tools/plugins/db-config/src/main/kotlin/net/corda/cli/plugins/dbconfig/DatabaseBootstrapAndUpgrade.kt
```diff
@@ -16,11 +16,11 @@ class DatabaseBootstrapAndUpgrade(wrapper: PluginWrapper) : Plugin(wrapper) {
     }
 
     override fun start() {
-        logger.info("Bootstrap plugin started.")
+        logger.debug("Bootstrap plugin started.")
```
fowlerrr, line 19: For these it's probably better to use the format:

```
        logger.debug { "Bootstrap plugin started." }
```

I don't remember all of the rules but you might save processing time (I think for any String building over 15 characters or with any substitution)

davidcurrie: I don't think the extensions to support this lazy logging syntax are available in the plugins.

fowlerrr: No, you're right.  I just checked and it's in `corda-runtime-os`.  I totally would have thought it would be more available than that.

fowlerrr (approved):

vadanx (approved):

PR #3441, opened 2023-03-22
CORE-11709: Remove HsmStore from usage in favour of CryptoRepository

fowlerrr:
PersistenceTests has also been moved to a more appropriate location


components/crypto/crypto-service-impl/src/integrationTest/kotlin/net/corda/crypto/persistence/impl/tests/PersistenceTests.kt
```diff
@@ -1,5 +1,6 @@
 package net.corda.crypto.persistence.impl.tests
```
dickon, line 1: Should the package name have changed to match the new code structure? Not necessarily worth waiting for a new green build to fix on this PR since I would be fine with merging it like this then fixing later.

fowlerrr: It's nowhere near green.  Some of the tests written around this expected way to much "real" stuff and are proving difficult. I'll fix it.

dickon (approved):
LGTM, other than the nitpick over package name we can fix in a later PR.

components/crypto/crypto-service-impl/src/integrationTest/kotlin/net/corda/crypto/persistence/impl/tests/PersistenceTests.kt
fowlerrr: Wait.  The package name matches.

fowlerrr:
This is proving difficult because of how many tests rely on HSMStore and expect it.  So I'm going to invert the logic for now and make the HSMStore itself use the CryptoRepository.  That's a more minimal change and we can plan to migrate services away from HSMStore later.

PR #3443, opened 2023-03-22
CORE-10794 Uniqueness metrics

relyafi:
### Summary

Metrics have been added to provide an insight into uniqueness processing. These have been added at 3 levels:

- The uniqueness checker client (metrics starting with `ledger.uniqueness.client`), which are metrics captured from the perspective of a notary virtual node notarisation flow, making uniqueness check requests
- Uniqueness checker (metrics starting with `uniqueness.checker`), which are metrics relating to the business logic / processing of uniqueness check requests
- Backing store (metrics starting with `uniqueness.backingstore`), which are metrics relating to the low level access and operations against the uniqueness database.

For full details, please see https://github.com/corda/platform-eng-design/blob/master/core/corda-5/corda-5.0/notary/metrics-design-c5.0/index.md

Other changes that have been made as part of this PR:

- A new "Uniqueness" section of the development Grafana dashboard has also been added to provide a visualisation of these metrics.
- All `Timer` metrics, both existing and new, have been updated to publish 50th, 95th and 99th percentile information.

### Testing

The backing store logic was modified to inject random `EntityExistsException` failures, to allow testing of the backing store retry and error metrics, as these cannot easily be invoked under normal running conditions. Metrics were observed to update correctly as part of these changes. Example panel data:

metrics1
metrics2
metrics3

### DIFF
--- a/components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImpl.kt
+++ b/components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImpl.kt
@@ -14,6 +14,7 @@ import net.corda.lifecycle.RegistrationStatusChangeEvent
 import net.corda.lifecycle.StartEvent
 import net.corda.lifecycle.StopEvent
 import net.corda.lifecycle.createCoordinator
+import net.corda.metrics.CordaMetrics
 import net.corda.orm.JpaEntitiesRegistry
 import net.corda.orm.JpaEntitiesSet
 import net.corda.uniqueness.backingstore.BackingStore
@@ -42,13 +43,14 @@ import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
+import java.time.Duration
+import java.time.Instant
 import javax.persistence.EntityExistsException
 import javax.persistence.EntityManager
 import javax.persistence.OptimisticLockException
 import javax.persistence.RollbackException
 
 @Suppress("ForbiddenComment")
-// TODO: Reimplement metrics, config
 /**
  * JPA backing store implementation, which uses a JPA compliant database to persist data.
  */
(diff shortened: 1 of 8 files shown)


relyafi:
@davidcurrie ideally I'd rather have "delta" rather than "running total" graphs for some of these metrics, but I was unable to get the delta / increase functionality in the prometheus query to work properly (it would often report 0 incorrectly). It seemed to be because the series were going out of existence when no data was reported due to using tag filtering. If you have any suggestions on how to make this work, it would be appreciated.

components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImpl.kt
```diff
@@ -42,13 +43,14 @@ import org.osgi.service.component.annotations.Component
...
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
+import java.time.Duration
+import java.time.Instant
 import javax.persistence.EntityExistsException
 import javax.persistence.EntityManager
 import javax.persistence.OptimisticLockException
 import javax.persistence.RollbackException
 
 @Suppress("ForbiddenComment")
```
davidcurrie, line 53: Can this go now?

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/BatchedUniquenessCheckerImpl.kt
```diff
@@ -217,21 +242,53 @@ class BatchedUniquenessCheckerImpl(
                                 }
                             )
                         )
+
+                        // IsDuplicate tag is omitted, as we do not have the information to deduce this in
+                        // an unhandled exception scenario
+                        CordaMetrics.Metric.UniquenessCheckerRequestCount
+                            .builder()
+                            .withTag(CordaMetrics.Tag.SourceVirtualNode, cordaHoldingIdentity.shortHash.toString())
+                            .withTag(CordaMetrics.Tag.ResultType, "UniquenessCheckResultUnhandledException")
```
davidcurrie: I let the method names pass but this feels like it should be a constant somewhere.

metrics/grafana/provisioning/dashboards/corda.json
davidcurrie, line 1: Too big a change to review from source.

components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImpl.kt
relyafi: Unfortunately not, because there's still several other TODOs in this file.

davidcurrie (approved):
LGTM

wzur-r3 (approved):
LGTM for Gradle file changes

PR #3446, opened 2023-03-22
CORE-11936: Add extra logging to SandboxSetupImpl.getService()

josephzunigadaly:
- Log how many times we tried to load the service
- Log which bundles were loaded at the time we stopped trying to load the service


josephzunigadaly:
Closing this PR because the bundle list may be quite long so we wouldn't want to put it into the exception and the timeouts are long enough that we expect multiple runs of the loop.

PR #3447, opened 2023-03-22
CORE-9467: Make vnode test timeouts consistent

josephzunigadaly:
Changes:
- Add immutable properties for timeout and interval
- Use new properties throughout the vnode test code

I've used the longest timeout and shortest interval. My reasoning is:
- The long timeout was added by someone who must have seen we needed it
- All the tests do similar work so they should use the same timeout
- Tests will exit as soon as their success condition passes so we won't run for `timeout` seconds
- We've seen failures in tests lacking the timeout, hopefully this will help improve stability
- The interval should be short to exit the test as soon as success is achieved


josephzunigadaly:
Also applied to the e2e test repo:
https://github.com/corda/corda-e2e-tests/pull/41

vkolomeyko (approved):
LGTM

PR #3448, opened 2023-03-22
CORE-11709: Point HSMStore to use CryptoRepository

fowlerrr:
Pointing those two classes to use the CryptoRepository means less code duplication and a path towards eventually removing them.  There's a lot of tests that need to be managed, however, in order to get that to work so this is a simpler and more expedient way to achieve that result.

Also ensuring that uses of CryptoRepository are closed to save database connections.


components/crypto/crypto-softhsm-impl/src/test/kotlin/net/corda/crypto/softhsm/impl/SigningKeyStoreUnitTest.kt
fowlerrr, line 1: I need more time looking at these tests but I think they're already covered so I might just deleted the file rather than make them fit what we have now

components/crypto/crypto-persistence-impl/build.gradle
```diff
@@ -24,6 +24,7 @@ dependencies {
     implementation project(":components:crypto:crypto-component-core-impl")
     implementation project(":components:crypto:crypto-persistence-model")
     implementation project(":components:crypto:crypto-persistence")
+    implementation project(":components:crypto:crypto-softhsm-impl")
```
dickon, line 27: surprised that's not a cycle

components/crypto/crypto-softhsm-impl/src/test/kotlin/net/corda/crypto/softhsm/impl/V1SigningKeyStoreTest.kt
```diff
@@ -208,4 +207,6 @@ class V1SigningKeyStoreTest {
         assertThat(tenantCap.allValues.single()).isEqualTo("tenant")
         assertThat(fullIdsCap.allValues.single()).isEqualTo(listOf(hashB.toString()))
     }
+
```
dickon, line 210: unnecessary whitespace

(not a merge blocker)

dickon (approved):
LGTM, nice

PR #3451, opened 2023-03-22
CORE-11868: Fix PersistenceTests to close EntityManagerFactory

fowlerrr:
This should ensure another EMF gets closed.


dickon (approved):
I probably would have done it all with Kotlin `use`, but this should work and is a smaller change to code we want to rework anyway, so: LGTM

PR #3452, opened 2023-03-22
CORE-11113 Support for notary protocol versioning

YashNabar:
Adds the following registration context fields to support notary protocol versioning -

- `corda.notary.service.flow.protocol.name`: Renamed from `corda.notary.service.plugin`. This is the name of the flow protocol used by the notary service.
- `corda.notary.service.flow.protocol.version.%s`: List of versions supported for the above flow.

Updates `NotaryInfo`, group parameters construction in static and dynamic registration.

https://github.com/corda/corda-api/pull/955


components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
```diff
@@ -462,6 +464,9 @@ class DynamicMemberRegistrationService @Activate constructor(
                     require(isNotEmpty()) { "No notary key ID was provided." }
                     require(orderVerifier.isOrdered(this, 3)) { "Provided notary key IDs are incorrectly numbered." }
                 }
+                context.keys.filter { notaryProtocolVersionsRegex.matches(it) }.apply {
+                    require(orderVerifier.isOrdered(this, 6)) { "Provided notary protocol versions are incorrectly numbered." }
```
nikinagy, line 474: can this be an empty list? or is it mandatory to provide a list when registering a notary?

components/flow/flow-service/src/test/kotlin/net/corda/flow/application/services/NotaryLookupImplTest.kt
```diff
@@ -63,11 +63,11 @@ class NotaryLookupImplTest {
 
         assertThat(notaries).anySatisfy {
             assertThat(it.name).isEqualTo(alice)
-            assertThat(it.pluginClass).isEqualTo("net.corda.Plugin1")
+            assertThat(it.protocol).isEqualTo("net.corda.Plugin1")
```
charlieR3, line 68: Could be good to check for protocol versions too

components/membership/membership-persistence-client/src/main/kotlin/net/corda/membership/persistence/client/MembershipPersistenceClient.kt
```diff
@@ -107,7 +107,7 @@ interface MembershipPersistenceClient : Lifecycle {
      */
     fun addNotaryToGroupParameters(
         viewOwningIdentity: HoldingIdentity,
-        notary: MemberInfo
+        notary: MemberInfo,
```
charlieR3: Seems fine but I'm not sure why `MembershipPersistenceClient.kt`, `MembershipPersistenceClientImplTest.kt` and `MembershipPersistenceClientImpl.kt` all needed to be changed. Was there going to be changes here that were reverted?

components/membership/membership-persistence-service-impl/src/integrationTest/kotlin/net/corda/membership/impl/persistence/service/MembershipPersistenceTest.kt
```diff
@@ -683,17 +695,19 @@ class MembershipPersistenceTest {
...
+            KeyValuePair(String.format(NOTARY_SERVICE_PROTOCOL_VERSIONS_KEY, 0, 0), "1"),
         )
 
         val persisted = membershipPersistenceClientWrapper.addNotaryToGroupParameters(viewOwningHoldingIdentity, notary)
 
         assertThat(persisted).isInstanceOf(MembershipPersistenceResult.Success::class.java)
         with((persisted as? MembershipPersistenceResult.Success<KeyValuePairList>)!!.payload.items) {
-            assertThat(size).isEqualTo(5)
-            assertThat(containsAll(expectedGroupParameters))
+            assertThat(this).anyMatch { it.key == net.corda.membership.lib.MODIFIED_TIME_KEY }
```
charlieR3: Could `MODIFIED_TIME_KEY` be imported and removed from the companion object?
```suggestion
            assertThat(this).anyMatch { it.key == MODIFIED_TIME_KEY }
```

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
```diff
@@ -68,15 +72,39 @@ internal class AddNotaryToGroupParametersHandler(
...
             }
             val (epoch, groupParameters) = if (notaryServiceNumber != null) {
                 // Add notary to existing notary service, or update notary with rotated keys
-                updateExistingNotaryService(parametersMap, notary, notaryServiceNumber, keyEncodingService, logger).apply {
+                val memberQueryBuilder = criteriaBuilder.createQuery(MemberInfoEntity::class.java)
+                val memberQuery = memberQueryBuilder.select(memberQueryBuilder.from(MemberInfoEntity::class.java))
+                val members = em.createQuery(memberQuery)
+                    .setLockMode(LockModeType.PESSIMISTIC_WRITE)
+                    .resultList.map {
+                        persistenceHandlerServices.memberInfoFactory.create(
```
charlieR3:
```suggestion
                        memberInfoFactory.create(
```

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
```diff
@@ -68,15 +72,39 @@ internal class AddNotaryToGroupParametersHandler(
...
+                            deserializeProperties(it.memberContext).toSortedMap(),
+                            deserializeProperties(it.mgmContext).toSortedMap(),
+                        )
+                    }
+                val currentProtocolVersions = members.filter {
+                        it.notaryDetails?.serviceName == notary.serviceName &&
+                        it.name != notaryInfo.name &&
+                        it.status == MEMBER_STATUS_ACTIVE
+                    }.flatMap { it.notaryDetails!!.serviceProtocolVersions }.toSet()
```
charlieR3: Doesn't this combine all versions rather than intersect them?

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandlerTest.kt
charlieR3, line 1: I'd like to see a test in here verifying that we get the intersection of versions when combined multiple virtual nodes under one service. Even though multiple vnode are not supported at the moment, the class under test is written to support it so I think we should have a test for it.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/staticnetwork/StaticMemberRegistrationService.kt
```diff
@@ -243,12 +243,20 @@ class StaticMemberRegistrationService @Activate constructor(
...
+        staticMemberList: List<StaticMember>,
+        groupReader: MembershipGroupReader,
+    ) {
         val cache = lifecycleHandler.groupParametersCache
         val holdingIdentity = memberInfo.holdingIdentity
         val groupParametersList = cache.getOrCreateGroupParameters(holdingIdentity).run {
-            memberInfo.notaryDetails?.let {
-                cache.addNotary(memberInfo)
+            memberInfo.notaryDetails?.let { notary ->
+                val currentProtocolVersions = groupReader.lookup()
```
charlieR3: It might be better to use `groupReader.notaryVirtualNodeLookup`

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/staticnetwork/StaticMemberRegistrationService.kt
```diff
@@ -243,12 +243,20 @@ class StaticMemberRegistrationService @Activate constructor(
...
+                    .filter { it.notaryDetails?.serviceName == notary.serviceName && it.name != memberInfo.name }
+                    .flatMap { it.notaryDetails!!.serviceProtocolVersions }
+                    .toSet()
```
charlieR3: Wouldn't this be a union of versions rather than intersection?

components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/ApproveRegistrationHandlerTest.kt
```diff
@@ -111,8 +120,29 @@ class ApproveRegistrationHandlerTest {
...
+    private val mgmContext: MGMContext = mock {
+        on { entries } doReturn mapOf("a" to "b").entries
+    }
+    private val otherNotary: MemberInfo = mock {
+        on { memberProvidedContext } doReturn memberContext
+        on { mgmProvidedContext } doReturn mgmContext
+    }
     private val groupReader: MembershipGroupReader = mock {
         on { groupParameters } doReturn mockGroupParameters
+        on { lookup() } doReturn listOf(otherNotary)
```
charlieR3: Was this necessary to make existing test pass or was it intended that additional assertions would be added for this in tests?

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/GroupParametersUtils.kt
```diff
@@ -19,19 +20,39 @@ const val NOTARIES_KEY = "corda.notary.service"
...
     logger.info("Adding notary to group parameters under existing notary service '$notaryServiceName'.")
-    notaryDetails.servicePlugin?.let {
-        require(currentParameters[String.format(NOTARY_SERVICE_PLUGIN_KEY, notaryServiceNumber)].toString() == it) {
+    notaryDetails.serviceProtocol?.let {
+        require(currentParameters[String.format(NOTARY_SERVICE_PROTOCOL_KEY, notaryServiceNumber)].toString() == it) {
             throw MembershipPersistenceException("Cannot add notary to notary service " +
-                    "'$notaryServiceName' - plugin types do not match.")
+                    "'$notaryServiceName' - protocols do not match.")
+        }
+        require(notaryDetails.serviceProtocolVersions.isNotEmpty()) {
```
charlieR3, line 40: Should this be outside `notaryDetails.serviceProtocol?.let`?

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/GroupParametersUtils.kt
```diff
@@ -19,19 +20,39 @@ const val NOTARIES_KEY = "corda.notary.service"
...
+            throw MembershipPersistenceException("Cannot add notary to notary service '$notaryServiceName' - protocol" +
+                    "  versions are missing.")
+        }
+    }
+    val versionsRegex = NOTARY_SERVICE_PROTOCOL_VERSIONS_KEY.format(notaryServiceNumber, "([0-9]+)").toRegex()
+    val updatedVersions = notaryDetails.serviceProtocolVersions.let {
+        val versionIntersection = if (currentProtocolVersions.isEmpty()) {
+            it
+        } else {
+            it.intersect(currentProtocolVersions.toSet())
```
charlieR3, line 50: If I understand the changes in the PR correctly, this will intersect the versions of the new notary with all versions of all existing notaries. It should be an intersection of all notaries. 

The way I'm reading this is that if we have notaryA with versions [1, 2, 3], notaryB with versions [2, 3, 4], and we try to add notaryC with versions [3, 4], the resulting list will be [3, 4] rather than the correct list which would be [3] since the areas of the code creating the current versions aren't intersecting.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterBuilder.kt
```diff
@@ -154,7 +154,7 @@ class ClusterBuilder {
             |    "corda.key.scheme" : "CORDA.ECDSA.SECP256R1", 
             |    "corda.roles.0" : "notary",
             |    "corda.notary.service.name" : "O=MyNotaryService, L=London, C=GB",
-            |    "corda.notary.service.plugin" : "net.corda.notary.NonValidatingNotary"
+            |    "corda.notary.service.flow.protocol.name" : "net.corda.notary.NonValidatingNotary"
```
charlieR3: Won't this need a version also?

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MembershipUtils.kt
```diff
@@ -111,7 +111,7 @@ fun onboardNotaryMember(
     mapOf(
         "corda.roles.0" to "notary",
         "corda.notary.service.name" to MemberX500Name.parse("O=NotaryService, L=London, C=GB").toString(),
-        "corda.notary.service.plugin" to "net.corda.notary.NonValidatingNotary",
+        "corda.notary.service.flow.protocol.name" to "net.corda.notary.NonValidatingNotary",
```
charlieR3, line 114: Won't this need a version also?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
YashNabar: I believe it would be empty if the protocol is not specified

nikinagy: but must you specify the protocol when you create a notary?

nikinagy: also I can see that in the converter we throw an exception if it was empty:
`val protocolVersions = context.map.parseList(PROTOCOL_VERSIONS_PREFIX, Int::class.java).toSet().apply {
             require(isNotEmpty()) { throw ValueNotFoundException("'$PROTOCOL_VERSIONS_PREFIX' is empty.") }
         }`
        wouldn't it be better to throw this exception here? or if it can be empty then default to emptyList() in the notary info converter
        (although I think parseList automatically returns empty list if the key wasn't found)

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/GroupParametersUtils.kt
YashNabar: Addressed in [1cb2291](https://github.com/corda/corda-runtime-os/pull/3452/commits/1cb22911f92226d0c91dd8f746d0f09498862388)

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/GroupParametersUtils.kt
YashNabar: Protocol version is required only when protocol name has been specified

components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/ApproveRegistrationHandlerTest.kt
YashNabar: That shouldn't be here, removed

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandlerTest.kt
YashNabar: Added in [1cb2291](https://github.com/corda/corda-runtime-os/pull/3452/commits/1cb22911f92226d0c91dd8f746d0f09498862388)

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
YashNabar: Addressed in [1cb2291](https://github.com/corda/corda-runtime-os/pull/3452/commits/1cb22911f92226d0c91dd8f746d0f09498862388)

components/membership/membership-persistence-client/src/main/kotlin/net/corda/membership/persistence/client/MembershipPersistenceClient.kt
YashNabar: Yes, removed now

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
YashNabar: My understanding is that it is tied to the protocol name (which is not mandatory when adding a notary to an existing notary service). So if protocol name is not provided, versions can be missing in the registration context. However while adding a new notary service, protocol name is required (hence, protocol version is required as well) - so versions shouldn't be an empty list in `NotaryInfo`

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/staticnetwork/StaticMemberRegistrationService.kt
YashNabar: Addressed in [1cb2291](https://github.com/corda/corda-runtime-os/pull/3452/commits/1cb22911f92226d0c91dd8f746d0f09498862388)

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandlerTest.kt
charlieR3, line 1: I notice there's a lot of mocking done per test. It could be a good idea to pull that out in to common code unless there's a reason each test needs unique mocking set up. It would just make it more readable.

charlieR3 (dismissed):
LGTM

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandlerTest.kt
```diff
@@ -296,53 +314,34 @@ class AddNotaryToGroupParametersHandlerTest {
         }
     }
 
+    @Test
+    fun `invoke sets notary protocol versions to intersection of protocol versions of individual notary vnodes`() {
```
relyafi, line 318: Whilst it doesn't really matter for now, since we only support a single virtual node for a service, I don't think intersection behavior will ever be correct. Notarisation requests are randomly assigned to different virtual nodes, so in the scenario where the virtual nodes are out of step, we will get non-deterministic failures.

This shows one of the weaknesses of having all this information set at the virtual node level, as I think any solution is going to be a bit clunky. Maybe we leave this as-is for now, but raise a ticket to address? We really need to think through what this looks like in the vnode amendment scenario too.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/MemberRole.kt
```diff
@@ -50,10 +51,14 @@ internal sealed class MemberRole {
...
-            val plugin = context[NOTARY_SERVICE_PLUGIN]
+            val protocol = context[NOTARY_SERVICE_PROTOCOL]
+            val protocolVersions = NOTARY_SERVICE_PROTOCOL_VERSIONS.format("([0-9]+)").toRegex().let { regex ->
+                context.filter { it.key.matches(regex) }.mapTo(mutableSetOf()) { it.value.toInt() }
+            }
             return Notary(
                 serviceName = MemberX500Name.parse(serviceName),
-                plugin = plugin,
+                protocol = protocol,
+                protocolVersions = protocolVersions,
```
relyafi: Unnecessary trailing comma

charlieR3 (dismissed):

relyafi (dismissed):
Mostly LGTM, couple of very minor comments

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandlerTest.kt
YashNabar: It's not quite clear to me why the intersection of individual virtual nodes' versions wouldn't work correctly. The supported versions for the notary service are re-calculated while updating the group parameters, such that the versions list only contains those versions which are supported by each of the virtual nodes under that notary service. This means that any virtual node selected at random for processing a particular notarisation request supports the flow protocol version.
Happy to leave this as is for now since this path is disabled, but can we have a discussion at some point to land on a suitable solution?

charlieR3 (dismissed):

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandlerTest.kt
relyafi: Bah, this is what happens when I raise comments before having my morning coffee - I'd read intersection as _union_. Intersection makes sense, so long as we are retaining the context of each virtual nodes supported versions and recalculating that on each change, as opposed to the intersection the previous _service level_ supported version and the virtual node being updated.

YashNabar: Yes we recalculate every time based on the versions defined in each virtual node's `MemberInfo`

relyafi (approved):

charlieR3 (approved):

vkolomeyko (approved):

PR #3453, opened 2023-03-22
CORE-11502: pull signing store cache up, clean up interfaces and implementations

dickon:
I reviewed the signing keys parts of the CryptoRepository interface. It turns out that the signing keys were all in the cluster database (since the tenantId was specified as CRYPTO when picking a database), irrespecitve of the tenant ID passed in when calling the SigningService. That's a bug, since virutal node data must be stored in the virtual node database. So, there's no reason for the tenant ID parameter to the signing methods, since an instance of `CryptoRepository` embeds a tenant ID, and so it is ambigious to specify it again. To be clear, this *is* a behaviour change.  

Currently `V1SigningKeyStore` contains both database code and a caching layer, mixed up together. This makes it hard to maintain the database code, and that's pressing since we need to make significant changes. It also means that it is necessary to configure that cache, which brings Corda config into the repository layer. SImplify this code by pushing the caching up to `SigningServiceImpl`. This matches the pattern where the peer code in `SoftCryptoService` does caching and calls down to the repository layer purely for database access. 

A previous attempt to solve this problem involved pulling out some methods to `SigningKeyRepositoryImpl` in order to make those methods unit testable without needing to involve lifecycle.  I believe the root cause of this was the number of OSGi and lifecycle components involved. So, I've removed `SigningServiceFactoryImpl` and so OSGi is no longer involved at this level, so the code is easy to test without needing lifecycle.


components/crypto/crypto-persistence-impl/src/main/kotlin/net/corda/crypto/persistence/impl/HSMStoreImpl.kt
```diff
@@ -84,23 +74,23 @@ class HSMStoreImpl @Activate constructor(
     ): HSMAssociationInfo = impl.associate(tenantId, category, hsmId, masterKeyPolicy)
 
     class Impl(
-        private val cryptoRepository: CryptoRepository,
+        private val cryptoHSMRRepository: HSMRepository,
```
fowlerrr: typo in `cryptoHSMRepository`

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoOpsBusProcessor.kt
```diff
@@ -38,36 +41,50 @@ import net.corda.data.crypto.wire.ops.rpc.commands.SignRpcCommand
...
-    event: ConfigChangedEvent
+    event: ConfigChangedEvent,
+    cryptoServiceFactory: CryptoServiceFactory,
+    schemeMetadata: CipherSchemeMetadata,
+    digestService: PlatformDigestService,
+    dbConnectionManager: DbConnectionManager,
+    jpaEntitiesRegistry: JpaEntitiesRegistry,
+    keyEncodingService: KeyEncodingService,
+    virtualNodeInfoReadService: VirtualNodeInfoReadService,
+    layeredPropertyMapFactory: LayeredPropertyMapFactory,
```
fowlerrr: I would argue that most (or all) of these are only for creating the `SigningService`, and therefore don't need to be passed in.  Let them remain with the caller.  I know that would interfere with the factory function below, but I think it's better to keep all of that _out_ of this class.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoOpsBusServiceImpl.kt
```diff
@@ -54,7 +75,17 @@ class CryptoOpsBusServiceImpl @Activate constructor(
...
+            event = event,
+            cryptoServiceFactory = cryptoServiceFactory,
+            schemeMetadata = schemeMetadata,
+            digestService = digestService,
+            keyEncodingService = keyEncodingService,
+            dbConnectionManager = dbConnectionManager,
+            jpaEntitiesRegistry = jpaEntitiesRegistry,
+            virtualNodeInfoReadService = virtualNodeInfoReadService,
+            layeredPropertyMapFactory = layeredPropertyMapFactory
+        )
```
fowlerrr, line 98: Yeah, I'd say create the `SigningService` here and just pass it in.  It's mockable for tests so there's no need for the factory function.

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/SigningServiceGeneralTests.kt
```diff
@@ -394,4 +394,97 @@ class SigningServiceGeneralTests {
...
+//
+//        repo.lookupSigningKeysByPublicKeyHashes("tenant", keys)
+//
+//        val cacheKeys = setOf(V1SigningKeyStore.CacheKey("tenant", shortA), V1SigningKeyStore.CacheKey("tenant", shortB))
+//        queryCap.allValues.single().forEach {
+//            assertThat(it in cacheKeys)
+//        }
+//        assertThat(tenantCap.allValues.single()).isEqualTo("tenant")
+//        assertThat(fullIdsCap.allValues.single()).isEqualTo(listOf(hashB.toString()))
+//    }
```
fowlerrr: We've already chatted, but just noting that these will come back in some form (in some place)

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/infra/TestWrappingRepository.kt
```diff
@@ -0,0 +1,18 @@
...
+class TestWrappingRepository(
+    val keys: ConcurrentHashMap<String, WrappingKeyInfo> = ConcurrentHashMap(),
+) : SigningRepository {
+    override fun saveKey(alias: String, key: WrappingKeyInfo) {
+        keys[alias] = key
+    }
+
+    override fun findKey(alias: String): WrappingKeyInfo? = keys[alias]
+
+    override fun close() {
```
fowlerrr, line 16: meh.  I suppose you could clear the map here.  But it probably doesn't matter much.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/WrappingRepository.kt
```diff
@@ -0,0 +1,9 @@
+package net.corda.crypto.softhsm
+
+import net.corda.crypto.persistence.WrappingKeyInfo
+import java.io.Closeable
+
+interface WrappingRepository : Closeable {
```
fowlerrr, line 13: Kdocs!!!!

components/crypto/crypto-softhsm-impl/src/test/kotlin/net/corda/crypto/softhsm/impl/TestSigningRepositoryImpl.kt
```diff
@@ -20,37 +18,32 @@ import org.mockito.kotlin.times
...
         verify(dbConnectionManager).getOrCreateEntityManagerFactory(CordaDb.Crypto, DbPrivilege.DML)
         verifyNoMoreInteractions(dbConnectionManager)
-        cryptoRepositoryFactory(CryptoTenants.P2P, config, dbConnectionManager, mock(), mock(), mock(), mock(), mock())
+        SigningRepositoryImpl(
+            entityManagerFactory = getEntityManagerFactory(CryptoTenants.P2P, dbConnectionManager, mock(), mock()),
+            tenantId = CryptoTenants.P2P,
+            keyEncodingService = mock(),
+            digestService = mock(),
+            layeredPropertyMapFactory = mock()
+        )
```
fowlerrr, line 53: Er, did you want this again?

components/crypto/crypto-softhsm-impl/src/test/kotlin/net/corda/crypto/softhsm/impl/TestV1SigningRepositoryImpl.kt
```diff
@@ -13,7 +13,7 @@ import org.mockito.kotlin.mock
 import kotlin.test.assertNotNull
 
 
-class TestV1CryptoRepositoryImpl {
+class TestV1SigningRepositoryImpl {
```
fowlerrr: Probably don't need the `V1` now

fowlerrr: Also, is this class for testing wrapping keys or signing keys?

gradle.properties
```diff
@@ -6,6 +6,7 @@ kotlin.stdlib.default.dependency=false
 kotlinMetadataVersion = 0.6.0
 
 org.gradle.jvmargs=-Dfile.encoding=UTF-8
+org.gradle.parallel=true
```
fowlerrr: Uncommit!!  Uncommit!!!

fowlerrr (commented):
I think I prematurely reviewed this and a lot of it would have been fixed anyway, but here are some comments.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoOpsBusProcessor.kt
dickon: yeah, much simpler now. I'm not sure how I missed that.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoOpsBusServiceImpl.kt
dickon: done

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/SigningServiceGeneralTests.kt
dickon: What do you think about leaving out the commented out new code in the signing service tests in this PR and then that's another piece either of us can do in parallel with other work? It's not exaclty like we're deleting tests, since that coverage didn't exist at this level

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/infra/TestWrappingRepository.kt
dickon: I don't see the point.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/WrappingRepository.kt
dickon: They didn't exist before.

dickon: done

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/SigningServiceGeneralTests.kt
fowlerrr: Yeah, I don't mind coming back to it.  I just wanted to go on record (for others reading this) that we have "a plan"

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoOpsBusServiceImpl.kt
```diff
@@ -76,15 +78,18 @@ class CryptoOpsBusServiceImpl @Activate constructor(
...
+                cryptoServiceFactory = cryptoServiceFactory,
+                schemeMetadata = schemeMetadata,
+                digestService = digestService,
+                keyEncodingService = keyEncodingService,
+                dbConnectionManager = dbConnectionManager,
+                jpaEntitiesRegistry = jpaEntitiesRegistry,
+                virtualNodeInfoReadService = virtualNodeInfoReadService,
+                layeredPropertyMapFactory = layeredPropertyMapFactory,
+                config = event.config.get(CRYPTO_CONFIG)
+                    ?: throw InvalidParameterException("$CRYPTO_CONFIG missing from config")
```
fowlerrr: There's a conveience function for this.
```suggestion
               config = event.config.toCryptoConfig()
```

Different exception but...

dickon:
I review @fowlerrr's commits to date, and they look good to me.

fowlerrr (approved):
I've been going over this as we worked on it and it's in good shape to merge for the work that will follow

PR #3454, opened 2023-03-23
CORE-10689 - Remove signature table

nikinagy:
Remove signature table and move signature information into registration request and member info database tables.

API PR: https://github.com/corda/corda-api/pull/990

**Testing:**

MGM's ID: 190D53DAC64F
Member's ID: EE0990007E14

After successfully registering MGM and member:
- registration lookup API for member
registration_lookup

- member lookup API for member
member_lookup

- registration request db table (MGM's signature is empty by default by our current behaviour)
registration_request_table

- member info db table (MGM's signature is empty by default by our current behaviour)
member_info_table


libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/registration/RegistrationRequestStatus.kt
```diff
@@ -11,6 +13,8 @@ data class RegistrationRequestStatus(
     val status: RegistrationStatus,
     val registrationId: String,
     val memberContext: KeyValuePairList,
+    val memberSignature: CryptoSignatureWithKey,
+    val memberSignatureSpec: CryptoSignatureSpec,
```
charlieR3: Same question as on the API - I'm curious why we need the signature returned as part of the registration status

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistMemberInfoHandler.kt
```diff
@@ -78,9 +78,12 @@ internal class PersistMemberInfoHandler(
                         memberInfo.status == MEMBER_STATUS_PENDING,
                         memberInfo.status,
                         clock.instant(),
-                        serializeContext(it.memberContext),
-                        serializeContext(it.mgmContext),
-                        memberInfo.serial
+                        serializeContext(it.persistentMemberInfo.memberContext),
```
charlieR3, line 81: I think if we're persisting the signature, we need to persist the byte array that the member signed over instead of creating the byte array ourselves. Signature verification could potentially fail if we don't store the byte array that the member signed over.

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/SignedMemberInfo.kt
```diff
@@ -0,0 +1,11 @@
+package net.corda.membership.lib
+
+import net.corda.data.crypto.wire.CryptoSignatureSpec
+import net.corda.data.crypto.wire.CryptoSignatureWithKey
+import net.corda.v5.membership.MemberInfo
+
+data class SignedMemberInfo (
```
charlieR3, line 7: This will need to store the signed data in it's original form (i.e. serialized)

charlieR3 (commented):
I'm not sure the signed data is being handled correctly. At the moment, the member is serializing it's member context, signing it, sending it to the MGM, the MGM persists the signed data and the signature with the registration request, but when the registration starts the signed data is dropped. It's worth a discussion on this but it feels to me like the serialized data from the member needs to become the source of truth (i.e. what we store in the database, how we construct the MemberInfo objects, and what we distribute) if we want to store the signatures with the member info. This is similar to what I have been doing for group parameters.

components/membership/membership-persistence-client-impl/src/main/kotlin/net/corda/membership/impl/persistence/client/MembershipQueryClientImpl.kt
charlieR3, line 1: It might cause issues if we query member info and signatures separately. Not sure if it's something we should do now in this, but the signature could be stored in the member info so that the signature matches the data. As it currently is there could be a case where the member info held in memory is different to the persisted member info and the signature will not match in that case.

charlieR3: (In which case `queryMembersSignatures` would be removed from this class)

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/registration/RegistrationRequestStatus.kt
nikinagy: I am planning to get the signatures for StartRegistrationHandler from the RegistrationRequest table by using the registration id. The functions on the persistence client all return RegistrationRequestStatus type, so I added the signature informations there. At this point we won't have anything in the MemberInfo table, so I won't be able to use the queryMemberSignatures() function.
I think probably changing the name of RegistrationRequestStatus could help.

nikinagy: Removed RegistrationRequestStatus in [8381c89](https://github.com/corda/corda-runtime-os/pull/3454/commits/8381c896ef94fa1393bedb36093dbfb4a2f5b676) and started using a renamed Avro scheme instead

nikinagy:
Other comments should be addressed by follow-up ticket: https://r3-cev.atlassian.net/browse/CORE-11990

charlieR3 (dismissed):
LGTM as long as we have follow up tickets to change how the signed data is handled (i.e. persist the signed data and don't recalculate the signed data on the MGM side), and to include the signature querying as part of the member/registration request lookup instead of having a separate signature query function.

charlieR3 (dismissed):

charlieR3 (approved):

PR #3455, opened 2023-03-23
CORE-10049 Add HashiCorp Vault addon format secret Config support to cli secret-config

simon-johnson-r3:
Add support to the `corda-cli secret-config` command to generate Config snippets for extracting secrets out of HashiCorp Vault using the Enterprise only Vault addon to Corda 5.

Because no secrets are encrypted into these Config entries, the `decrypt` command cannot be used for VAULT type.
The only mandatory parameter is --vault-path, and <value> which is the key for Vault to use (as opposed to the value to encrypt as it is with the built in Corda secrets service).

Example of use:
```
simon.johnson@LDNM-WDD239QFH2 corda % corda-cli secret-config -v myPath -t vault passwordKey create
```
output:
```
{"configSecret":{"vaultKey":"passwordKey","vaultPath":"myPath"}}
```


tools/plugins/secret-config/src/test/kotlin/net/corda/cli/plugin/secretconfig/TestSecretConfigPlugin.kt
```diff
@@ -51,65 +66,62 @@ class TestSecretConfigPlugin {
...
 
         println("Output:\n$outText")
         println("Expected:\n$expectedError")
-        assertThat(outText).isEqualTo(expectedError).withFailMessage(caseName)
+        assertThat(outText).contains(expectedError).withFailMessage(caseName)
     }
 
     companion object {
 
-        private val helpText =
```
simon-johnson-r3, line 71: removed this, the help is now much larger and it isn't helpful to try and do an exact string match on it for every failure scenario - not least because it's auto generated for us, which is one of the features of Picocli - instead the test passes as long as the output `contains` the string

dickon (dismissed):
LGTM

tools/plugins/secret-config/src/main/kotlin/net/corda/cli/plugin/secretconfig/VaultSecretConfigGenerator.kt
```diff
@@ -0,0 +1,17 @@
+package net.corda.cli.plugin.secretconfig
+
+import com.typesafe.config.Config
+import com.typesafe.config.ConfigFactory
+import net.corda.libs.configuration.SmartConfig
+
+private const val VAULT_PATH = "vaultPath"
+private const val VAULT_KEY = "vaultKey"
```
driessamyn: These should really be constants [in the API repo](https://github.com/corda/corda-api/blob/dfe78200c0973bc3baff87bceba6e2df8477cede/data/config-schema/src/main/java/net/corda/schema/configuration/ConfigKeys.java#L37)

I notice they are hardcoded here as well: https://github.com/corda/corda5-enterprise/blob/ada3c14d94234067170b0a0884e437ce20ef9414/addons/configuration/hashicorp-vault/src/main/kotlin/com/r3/corda/configuration/hashicorpvault/VaultSecretsServiceImpl.kt#L93

tools/plugins/secret-config/src/main/kotlin/net/corda/cli/plugin/secretconfig/SecretConfigGenerator.kt
```diff
@@ -0,0 +1,14 @@
+package net.corda.cli.plugin.secretconfig
+
+import com.typesafe.config.Config
+
+interface SecretConfigGenerator {
```
driessamyn: Isn't this pretty much a duplicate of `SecretsCreateService`?

tools/plugins/secret-config/src/main/kotlin/net/corda/cli/plugin/secretconfig/SecretConfigPlugin.kt
```diff
@@ -2,76 +2,114 @@ package net.corda.cli.plugin.secretconfig
...
         )
         fun decrypt() {
-            if (salt.isNullOrBlank()) {
-                throw ParameterException(spec.commandLine(), "A salt must be provided")
-            }
-            if (passphrase.isNullOrBlank()) {
-                throw ParameterException(spec.commandLine(), "A passphrase must be provided")
+            if (type != SecretsServiceType.CORDA) {
+                println("'decrypt' command can only be run on secrets created by CORDA secrets service.")
+                return
```
driessamyn, line 99: Would it be worth adding where in Vault the secret can be found? I.e. compile the "path" in the way the AddOn does?

simon-johnson-r3: `decrypt` takes a value not a Config for the Corda secrets service. To do what you suggest for VAULT type would require to accept a Config, which makes it a little asymmetrical IMHO. In other words `decrypt` as it stands is not the opposite of `create`. I guess it's really there to manually decrypt secrets you have in your Config. I think having `decrypt` for CORDA type accepting a Config would make it a little harder to use too.

So if you really wanted this, I'd suggest a different subcommand which accepts a VAULT Config and resolves it to a URL. However then worth noting all I can really do is add `/v1/secret/data/` to the path, I can't know the host here. So I'm not sure how much this is worth the effort. As a user it's not offering much, I'm not sure I'd be minded to run this command.

What are your thoughts on this?

driessamyn: Y. You're right. I don't think it's worth it.
Thinking about it again, decrypt doesn't make sense in the context of vault. I'm fine leaving as is.

libs/cache/cache-caffeine/src/main/kotlin/net/corda/cache/caffeine/CacheFactory.kt
```diff
@@ -14,4 +16,13 @@ interface CacheFactory {
      * @param caffeine Caffeine cache builder.
      */
     fun <K, V> build(name: String, caffeine: Caffeine<in K, in V>): Cache<K, V>
-}
\ No newline at end of file
+
```
simon-johnson-r3, line 19: This relates to the comment on the accompanying enterprise PR: https://github.com/corda/corda5-enterprise/pull/25/files#r1145963146

The CacheFactory requires the feature to create a loading cache.

libs/configuration/configuration-core/src/main/kotlin/net/corda/libs/configuration/SmartConfigFactory.kt
```diff
@@ -25,11 +27,14 @@ interface SmartConfigFactory {
             // select type from the config, and fall back on the EncryptionSecretsServiceFactory
             val type = secretsServiceConfig.getStringOrDefault(SECRET_SERVICE_TYPE, EncryptionSecretsServiceFactory.TYPE)
 
-            val secretsServiceFactory = secretsServiceFactories
-                .singleOrNull { it.type == type }
-                ?: throw SecretsConfigurationException("SecretsServiceFactory of type $type is not available.")
-
-            return SmartConfigFactoryImpl(secretsServiceFactory.create(secretsServiceConfig))
```
simon-johnson-r3, line 32: This relates to testing of the enterprise Combined Worker. It is possible to get multiple plugins advertising the same type. Previously we would halt execution of Corda at this point, now we are more resilient: log the problem and choose the first.

libs/configuration/configuration-core/src/main/kotlin/net/corda/libs/configuration/SmartConfig.kt
```diff
@@ -11,10 +11,6 @@ import com.typesafe.config.ConfigValue
  */
 @Suppress("TooManyFunctions")
 interface SmartConfig : Config {
```
simon-johnson-r3, line 13: This relates to https://github.com/corda/corda-runtime-os/pull/3455#discussion_r1147256452 this constant was moved to the Api repo with the others.

driessamyn (approved):
LGTM

mcgovc (approved):
LGTM - build file(s) changes

PR #3456, opened 2023-03-23
CORE-11739: Change Helm chart to use custom TLS certificate for REST

vkolomeyko:
I have done local K8s deployment and certificate presented looks as expected:
image


libs/rest/ssl-cert-read-impl/src/main/kotlin/net/corda/rest/ssl/impl/SslCertReadServiceImpl.kt
```diff
@@ -140,7 +141,12 @@ class SslCertReadServiceImpl(private val createDirectory: () -> Path) : SslCertR
 
         val privateKeyInfo = File(tlsKeyPath).reader().use { keyReader ->
             val pemParser = PEMParser(keyReader)
-            PrivateKeyInfo.getInstance(pemParser.readObject())
+            val pemObject = pemParser.readObject()
+            if (pemObject is PEMKeyPair) {
```
vkolomeyko, line 145: This was necessary as Helm's `genSignedCert` produces `KeyPair` when `$cert.Key` is used.

charts/corda-lib/templates/_worker.tpl
```diff
@@ -195,6 +195,11 @@ spec:
           - "-ddatabase.jdbc.directory=/opt/jdbc-driver"
           - "-ddatabase.pool.max_size={{ .clusterDbConnectionPool.maxSize }}"
           {{- end }}
+          {{- if $optionalArgs.restTlsSecret }}
+          - "-rtls.crt.path=/rest/tls/tls.crt"
```
vkolomeyko, line 199: The "meat" of the change where new arguments are passed to the image at start-up.

charts/corda/values.yaml
```diff
@@ -96,7 +96,7 @@ db:
 # Kafka configuration
 kafka:
   # -- comma-separated list of Kafka bootstrap servers (required); for example: "broker1.example.com:9092,broker2.example.com:9092,broker3.example.com:9092"
-  bootstrapServers: ""
+  bootstrapServers: "change-me-kafka:9092"
```
vkolomeyko: Having an empty string was violating constrains for this field.

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -119,6 +119,34 @@ Initial REST API admin password secret name
...
+*/}}
+{{- define "corda.restTlsKeySecretKey" -}}
+{{ .Values.workers.rest.tls.key.valueFrom.secretKeyRef.key | default "key" }}
+{{- end }}
+
+{{/*
+REST TLS CA cert secret key
+*/}}
+{{- define "corda.restTlsCaSecretKey" -}}
+{{ .Values.workers.rest.tls.ca.valueFrom.secretKeyRef.key | default "ca" }}
```
davidcurrie: Defacto key names for these in k8s would be `tls.crt`, `tls.key`, and `ca.crt`. (The first two, at least, are what you get when you create a Kubernetes secret of type `tls`.)

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -119,6 +119,34 @@ Initial REST API admin password secret name
 {{ .Values.bootstrap.restApiAdmin.password.valueFrom.secretKeyRef.name | default (include "corda.restApiAdminSecretName" .) }}
 {{- end }}
 
+{{/*
+REST TLS keystore secret name
+*/}}
+{{- define "corda.restTlsSecretName" -}}
+{{ .Values.workers.rest.tls.crt.valueFrom.secretKeyRef.name | default (printf "%s-rest-tls" (include "corda.fullname" .)) }}
```
davidcurrie: At least as the configuration currently stands, you could have different user-provided secret names for each of the three files.

charts/corda-lib/templates/_worker.tpl
```diff
@@ -327,6 +337,18 @@ spec:
               - key: {{ $.Values.kafka.tls.truststore.valueFrom.secretKeyRef.key | quote }}
                 path: "ca.crt"
         {{- end -}}
+        {{- if $optionalArgs.restTlsSecret }}
+        - name: resttls
+          secret:
+            secretName: {{ include "corda.restTlsSecretName" $ | quote }}
```
davidcurrie: As above, these files could be in three different secrets.

charts/corda/templates/secrets.yaml
```diff
@@ -52,3 +52,24 @@
     ( dict "username" ( dict "required" true ) "password" ( dict "generate" 12 ) )
   )
 }}
+{{- $restWorkerName := "rest-worker" }}
+{{- $ca := genCA "REST Worker Certification Authority" 1000 }}
+{{- $interCert := genSignedCert "REST Worker Intermediary Cert" nil nil 365 $ca }}
```
davidcurrie: What are we gaining from having an intermediate certificate here when we put the root certificate in the chain anyway?

charts/corda/templates/secrets.yaml
```diff
@@ -52,3 +52,24 @@
...
+{{- $altNames := list ( printf "%s.c5.r3.com" .Release.Namespace ) ( printf "%s.%s.svc" $restWorkerName .Release.Namespace ) "localhost" }}
```
davidcurrie: What's the purpose of the `c5.r3.com` domain here? It's probably worth including the `%s` and `%s.%s` options here and maybe also giving them the option of specifying the cluster domain name as well (usually `cluster.local` by default but is often configurable). I think it tends to divide opinion but I would not include `localhost` in this list.

charts/corda/templates/secrets.yaml
```diff
@@ -52,3 +52,24 @@
     ( dict "username" ( dict "required" true ) "password" ( dict "generate" 12 ) )
   )
 }}
+{{- $restWorkerName := "rest-worker" }}
```
davidcurrie: This should be the service name (currently embedded as `$workerName` in `corda.worker` so would need extracting into its own function if this stays here).

charts/corda/templates/secrets.yaml
```diff
@@ -52,3 +52,24 @@
...
+{{- $ca := genCA "REST Worker Certification Authority" 1000 }}
+{{- $interCert := genSignedCert "REST Worker Intermediary Cert" nil nil 365 $ca }}
+{{- $altNames := list ( printf "%s.c5.r3.com" .Release.Namespace ) ( printf "%s.%s.svc" $restWorkerName .Release.Namespace ) "localhost" }}
+{{- $cert := genSignedCert $restWorkerName (list "127.0.0.1") $altNames 365 $interCert }}
```
davidcurrie: I'd probably just leave the list of IPs as `nil` as this is not a helpful IP to use.

charts/corda/values.yaml
davidcurrie: That's deliberate - it means that the user has to provide a value before they can perform an install.

charts/corda/values.schema.json
```diff
@@ -1287,6 +1287,28 @@
...
+                                },
+                                "tls": {
+                                    "type": "object",
+                                    "title": "REST Worker TLS configuration",
+                                    "additionalProperties": false,
+                                    "properties": {
+                                        "crt": {
+                                            "title": "TLS Certificate configuration",
+                                            "additionalProperties": false,
+                                            "$ref": "#/$defs/config"
```
davidcurrie: the config schema supports both `value` and `valueFrom` whereas your implementation does not

charts/corda/values.yaml
```diff
@@ -740,6 +740,38 @@ workers:
...
+      # Certification Authority(CA) certificates chain configuration
+      ca:
+        # the CA certificates chain secret configuration
+        valueFrom:
+          # the CA certificates chain secret key reference
+          secretKeyRef:
+            # -- the CA certificates chain secret name
+            name: ""
+            # -- the CA certificates chain secret key
+            key: ""
```
davidcurrie: I'm inclined to think that this pattern has outlived its usefulness here. When specifying a username and password it was possible that you might want to provide the username as a value and the password via a secret. I'm not convinced that it's useful to be able to specify the certificates/key in different secrets here. I'm not even sure whether, given there is a standard TLS secret format, it's useful to be able to override the key names. Perhaps we should just have a `tlsSecretName` field and be done with it?! I'll sleep on that one...

charts/corda/templates/workers.yaml
```diff
@@ -11,5 +11,5 @@
 ) }}
 {{- include "corda.worker" ( list $ .Values.workers.p2pLinkManager "p2pLinkManager" ) }}
 {{- include "corda.worker" ( list $ .Values.workers.rest "rest"
-  ( dict "httpPort" 8888 )
+  ( dict "httpPort" 8888 "restTlsSecret" true )
```
davidcurrie: A better abstraction here would be to use `additionalWorkArgs` to pass the three extra parameters. I'd then just have the generic worker function look for the existence of a `tls` field in the values to see whether it should mount a secret (and drop the `rest` from the mount location).

charts/corda/templates/secrets.yaml
davidcurrie: I'd push all of this into `secrets.yaml` and make a generic `corda.tlsSecret` function.

charts/corda/templates/secrets.yaml
```diff
@@ -52,3 +52,24 @@
     ( dict "username" ( dict "required" true ) "password" ( dict "generate" 12 ) )
   )
 }}
+{{- $restWorkerName := "rest-worker" }}
+{{- $ca := genCA "REST Worker Certification Authority" 1000 }}
```
davidcurrie: Is it worth having something in here that makes it clear that this is self-signed?

davidcurrie (changes requested):

charts/corda/values.yaml
vkolomeyko: Sure - I will undo this change.

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: I think I will change this approach to a **single** secret name only which will have to contain all 3 keys.

charts/corda-lib/templates/_worker.tpl
vkolomeyko: As above, this will be changed to a single secret.

charts/corda/templates/secrets.yaml
vkolomeyko: It was just to illustrate that we can have longer certificate chains. I can eliminate intermediate bit, if there is an objection.

charts/corda/templates/secrets.yaml
vkolomeyko: Sure - I will add an option allowing to specify the domain name.
Will remove `localhost` as well.

charts/corda/templates/secrets.yaml
vkolomeyko: Sure - I will remove IPs.

charts/corda/values.schema.json
vkolomeyko: As per above, I will change this schema to have just a single secret name for all 3 keys.

charts/corda/values.yaml
vkolomeyko: Yeap, one secret name is the way to go, I think. It was mentioned on multiple instances above as well.

charts/corda/templates/workers.yaml
vkolomeyko: Sure - I will change.

charts/corda/templates/secrets.yaml
vkolomeyko: Sure - I will change.

charts/corda/templates/secrets.yaml
vkolomeyko: Sure, I will change it to `REST Worker Self-Signed Certification Authority`.

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -447,7 +475,7 @@ Secret creation
 {{-     if and $v.required ( not $field.value ) }}
 {{-       fail ( printf "Must specify %s.%s.valueFrom.secretKeyRef.name or %s.%s.value" $path $k $path $k ) }}
 {{-     end }}
-{{-     if or $field.value $v.generate }}
+{{-     if or $field.value $v.generate $v.explicit }}
```
vkolomeyko: I no longer need change for `explicit` value passing.
However, I thought of it as being a useful facility for the future.
Happy to undo it, if people feel this is unnecessary.

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -119,6 +119,34 @@ Initial REST API admin password secret name
...
+{{- define "corda.restTlsKeySecretKey" -}}
+{{ .Values.workers.rest.tls.key.secretKey | default "key" }}
+{{- end }}
+
+{{/*
+REST TLS CA cert secret key
+*/}}
+{{- define "corda.restTlsCaSecretKey" -}}
+{{ .Values.workers.rest.tls.ca.secretKey | default "ca" }}
+{{- end }}
```
davidcurrie: I'm still not sold on the need to be able to have user-definable key values here. I think they've sufficiently standard that it is fair to just say that the secret needs to have three data fields called `tls.crt`, `tls.key`, and `ca.crt`. See for example: https://cert-manager.io/docs/concepts/certificate/

If you do want them to be user-configurable, then I think `tls.crt`, `tls.key`, and `ca.crt` should still be the default values. The default values should live in `values.yaml` where the user can see them.

charts/corda-lib/templates/_helpers.tpl
davidcurrie: Why not have `corda.tlsSecret` call this function once it has prepared the certificates?

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -479,11 +507,51 @@ data:
...
+{{- $secretName := index . 3 }}
+{{- $crtSecretKey := index . 4 }}
+{{- $keySecretKey := index . 5 }}
+{{- $caSecretKey := index . 6 }}
+{{- $existingSecret := lookup "v1" "Secret" $.Release.Namespace $secretName }}
+{{- if not $existingSecret }}
+{{- $caName := printf "%s Self-Signed Certification Authority" $kind }}
+{{- $ca := genCA $caName 1000 }}
+{{- $interName := printf  "%s Intermediary Cert" $kind }}
+{{- $interCert := genSignedCert $interName nil nil 365 $ca }}
```
davidcurrie: My question still remains: why have an intermediary certificate here?

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -479,11 +507,51 @@ data:
...
 {{- end }}
 {{- end }}
 {{- end }}
 {{- end }}
+{{/*
+TLS Secret creation
+*/}}
+{{- define "corda.tlsSecret" -}}
+{{- $ := index . 0 }}
+{{- $kind := index . 1 }}
```
davidcurrie: `kind` has a very specific meaning in Kubernetes resources so it might be worth calling this something different

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -479,11 +507,51 @@ data:
...
+{{- $altNames := index . 2 }}
+{{- $secretName := index . 3 }}
+{{- $crtSecretKey := index . 4 }}
+{{- $keySecretKey := index . 5 }}
+{{- $caSecretKey := index . 6 }}
+{{- $existingSecret := lookup "v1" "Secret" $.Release.Namespace $secretName }}
```
davidcurrie, line 535: What happens if the existing secret has different key names or alt names to the values that are now specified on the chart?

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -479,11 +507,51 @@ data:
...
+{{- $cert := genSignedCert $kind nil $altNames 365 $interCert }}
+{{- $certChain :=  printf "%s\n%s" $interCert.Cert $ca.Cert }}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ $secretName }}
+  annotations:
+    "helm.sh/hook-weight": "-1"
+    "helm.sh/hook": pre-install
```
davidcurrie: These can go as I don't think this secret needs to be available at pre-install time. Kubernetes will ensure that the REST worker does not start until the secret is available.

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -479,11 +507,51 @@ data:
...
+{{- $crtSecretKey := index . 4 }}
+{{- $keySecretKey := index . 5 }}
+{{- $caSecretKey := index . 6 }}
+{{- $existingSecret := lookup "v1" "Secret" $.Release.Namespace $secretName }}
+{{- if not $existingSecret }}
+{{- $caName := printf "%s Self-Signed Certification Authority" $kind }}
+{{- $ca := genCA $caName 1000 }}
+{{- $interName := printf  "%s Intermediary Cert" $kind }}
+{{- $interCert := genSignedCert $interName nil nil 365 $ca }}
+{{- $cert := genSignedCert $kind nil $altNames 365 $interCert }}
```
davidcurrie: It is fine to allow additional or overridden `altNames` to be configured but the internal DNS names that Kubernetes will allocate should certainly be the defaults (as you had started to do previously). Likewise, I'd expect the CN should be the service name.

charts/corda-lib/templates/_worker.tpl
```diff
@@ -327,6 +332,18 @@ spec:
...
+            items:
+              - key: {{ include "corda.restTlsCrtSecretKey" $ | quote }}
+                path: "tls.crt"
+              - key: {{ include "corda.restTlsKeySecretKey" $ | quote }}
+                path: "tls.key"
+              - key: {{ include "corda.restTlsCaSecretKey" $ | quote }}
+                path: "ca.crt"
```
davidcurrie: If the key names are fixed, this all gets a lot simpler but I'd still like to see this not explicitly depend on the worker type here. So either pass the secret name (and keys if needed) as parameters into the `corda.worker` function or swap the conditional for a `with .tls`.

charts/corda/values.yaml
```diff
@@ -740,6 +740,26 @@ workers:
...
+        # -- the TLS Certificate secret key
+        secretKey: ""
+      # TLS Certificate private key configuration
+      key:
+        # -- the TLS Certificate private key secret key
+        secretKey: ""
+      # Certification Authority(CA) certificates chain configuration
+      ca:
+        # -- the CA certificates chain secret key
+        secretKey: ""
```
davidcurrie: As above, if these are staying then they should have default values in them

charts/corda/values.yaml
```diff
@@ -740,6 +740,26 @@ workers:
               name: ""
               # -- the REST worker SASL password secret key for client connection to Kafka
               key: ""
+    # REST Worker TLS configuration
+    tls:
+      # the name of the secret to derive information below
```
davidcurrie: "leaf" comments should begin `--`

davidcurrie: the comment is used in generated documentation where you can't rely on "below" still being meaningful

davidcurrie: the comment should make it clear that you need to provide the name of an existing secret containing the credentials or that one will get generated using self-signed certificates

charts/corda/values.yaml
```diff
@@ -740,6 +740,26 @@ workers:
...
+      secretName: ""
+      generation:
+        altNames:
+          - "*.c5.r3.com"
+          - "*.another.one.r3.com"
```
davidcurrie: We shouldn't have `r3.com` in here. As per the comments above, my preference would be for this to be an empty list by default and to have the default names based on the Kubernetes internal service names. With the ingress route we discussed, the only thing accessing these certificates will be something like nginx running in the same Kubernetes cluster as the workers.

davidcurrie (changes requested):

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: Sure - I have replied here: https://github.com/corda/corda-runtime-os/pull/3456#discussion_r1150305230

I can remove intermediate bit, if necessary.

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: How about changing this to `purpose` ?

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: I found this function rather complicated as it is at the moment.
Also this function takes `.Values...` as an input parameter and then it assumes that secret keys will be stored in a particular way under `.Values`.
Hence I decided against using it.

charts/corda-lib/templates/_helpers.tpl
davidcurrie: Sorry - I didn't see your earlier response. I don't think we should be using this as an example of how to create a certificate. This should just be the minimum needed when they haven't done so.

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: OK, sure.
Let me rename those secret names and move them to `values.yaml`.

charts/corda-lib/templates/_helpers.tpl
davidcurrie: OK - that's fine. In which case, yes, could you please remove the `explicit` logic again.

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: If a secret exists and keys mismatch then in this particular function nothing is going to happen, i.e. no new secret will be created.
Then at REST worker pod start-up time an error is likely to occur when trying to mount a secret with non-existent key name. Which I think is fine.
@davidcurrie - what would you rather happen? Do you want another clause to this function to validate existence of the secret keys here?

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: Sure - I will remove.

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: Sure - I will change that.

charts/corda-lib/templates/_worker.tpl
vkolomeyko: Sure - I see your point. Let me re-work.

charts/corda/values.yaml
vkolomeyko: Sure - will change.

charts/corda/values.yaml
vkolomeyko: Sure, I will change.

charts/corda/values.yaml
vkolomeyko: Sure - I will change.

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: Removed.

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: Removed.

charts/corda-lib/templates/_helpers.tpl
davidcurrie: As you say, if the `keys` are changed then it will fail later and I think it's fine to expect the user just to correct them as they would if they were using their own certificate. The `altNames` is more problematic as, if this changes, they probably are expecting the certificate to be regenerated. I'm guessing extracting these from the existing certificate is non-trivial in this context so perhaps it is worth writing these into the secret as well, or maybe adding them as annotations?

vkolomeyko: Sure - I will add `altNames` as annotations.

charts/corda/values.yaml
```diff
@@ -743,6 +743,22 @@ workers:
               name: ""
               # -- the REST worker SASL password secret key for client connection to Kafka
               key: ""
+    # REST Worker TLS configuration
+    tls:
```
davidcurrie, line 747: include the optional properties here as well, particularly as the README is generated from the comments in this file, not the schema - so something like:
```
  # ...
  generation:
     # -- ....
     altNames: []
```

charts/corda/values.schema.json
```diff
@@ -1337,6 +1337,62 @@
...
+                                            "title": "The name of the secret to derive information below. If missing, a new secret will be created and certificates/keys generated."
+                                        },
+                                        "generation": {
+                                            "type": "object",
+                                            "title": "Optional parameters to be used during TLS certificate generation.",
+                                            "additionalProperties": false,
+                                            "properties": {
+                                                "altNames": {
+                                                    "type": "array",
+                                                    "title": "Alternative names to be include into a certificate when generated."
```
davidcurrie:
```suggestion
                                                    "title": "Alternative names to be included in a certificate when generated."
```

charts/corda/values.schema.json
```diff
@@ -1337,6 +1337,62 @@
...
+                                },
+                                "tls": {
+                                    "type": "object",
+                                    "title": "REST Worker TLS configuration",
+                                    "additionalProperties": false,
+                                    "properties": {
+                                        "secretName": {
+                                            "type": "string",
+                                            "default": "",
+                                            "title": "The name of the secret to derive information below. If missing, a new secret will be created and certificates/keys generated."
```
davidcurrie:
```suggestion
                                            "title": "The name of the Kubernetes secret containing the certificate, private key and CA certificate. If not specified, a new secret will be created and certificates/keys generated."
```

charts/corda/values.yaml
```diff
@@ -743,6 +743,22 @@ workers:
...
+      # -- the name of the K8s secret to derive information about certificate, private key and CA certificate
```
davidcurrie:
```suggestion
      # -- the name of the Kubernetes secret containing the certificate, private key and CA certificate; if not specified, a new secret will be created and certificates/keys generated
```

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -506,3 +513,40 @@ data:
...
+{{- $ := index . 0 }}
+{{- $purpose := index . 1 }}
+{{- $serviceName := index . 2 }}
+{{- $altNames := index . 3 }}
+{{- $secretName := index . 4 }}
+{{- $crtSecretKey := index . 5 }}
+{{- $keySecretKey := index . 6 }}
+{{- $caSecretKey := index . 7 }}
+{{- $existingSecret := lookup "v1" "Secret" $.Release.Namespace $secretName }}
+{{- if not $existingSecret }}
```
davidcurrie: We generate a certificate if they give us a name but a secret doesn't exist with that name. I think we should only generate if they don't give us a name AND the generated secret doesn't already exist. If they gave us a name and the secret doesn't exist (maybe there was a typo), it should just error out when the worker tries to start.

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -506,3 +513,40 @@ data:
...
+{{- $caName := printf "%s Self-Signed Certification Authority" $purpose }}
+{{- $ca := genCA $caName 1000 }}
+{{- if not $altNames }}
+{{-   $altNames = list }}
+{{- end }}
```
davidcurrie, line 531: You cover `altNames` being empty elsewhere so we should always have a list by now

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -506,3 +513,40 @@ data:
...
+{{- $altNames = ( append $altNames ( printf "%s.%s.svc" $serviceName $.Release.Namespace ) ) }}
```
davidcurrie: The list should also contain `( printf "%s.%s" $serviceName $.Release.Namespace )`

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -506,3 +513,40 @@ data:
...
+{{- $cert := genSignedCert $serviceName nil $altNames 365 $ca }}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ $secretName }}
+  annotations:
+    certificate/altNames: {{ toString ( toStrings ( $altNames ) ) | quote }}
```
davidcurrie: The purpose of adding the annotations was to check that they are the same as the current value of `altNames` when retrieving an existing secret.

davidcurrie: Isn't `altNames` already a list of strings? I would expect to see something more like `join "," $altNames`

davidcurrie (changes requested):

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: Not quite.
```
{{- $altNames := list }}
{{- if .Values.workers.rest.tls.generation }}
{{-   $altNames = .Values.workers.rest.tls.generation.altNames }}
{{- end}}
```
covers the case when `generation` section is missing. However, there can be a situation when `generation` section is present, but `altNames` inside inside it were not provided.
Therefore, I think this is still necessary.

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: Sure - let me alter logic here.

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -506,3 +513,50 @@ data:
...
+{{- $crtSecretKey := index . 6 }}
+{{- $keySecretKey := index . 7 }}
+{{- $caSecretKey := index . 8 }}
+{{- $existingSecret := "undefined" }}
+{{- if $maybeExplicitSecretName }}
+{{-   if ( lookup "v1" "Secret" $.Release.Namespace $maybeExplicitSecretName ) }}
+{{-     $existingSecret = true }}
+{{-   else }}
+{{-     fail ( printf ( "The name of the existing secret '%s' been explicitly provided but it does not exist." $maybeExplicitSecretName ) ) }}
+{{-   end }}
```
davidcurrie: Sorry - I've made this more complicated than it needs be. I'd just skip calling `corda.tlsSecret` if `secretName` has been set. It's better to let Kubernetes handle the existence or otherwise of that secret. By putting a check here you're adding an explicit ordering constraint (the secret must exist before Corda is deployed) which is a bit of an anti-pattern.

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -506,3 +513,50 @@ data:
...
+{{- else }}
+{{-   $existingSecret = ( lookup "v1" "Secret" $.Release.Namespace $newSecretName ) }}
+{{- end }}
+{{- if not $existingSecret }}
```
davidcurrie: this needs to be "or there is an existing secret but with the wrong altNames"

charts/corda/values.yaml
```diff
@@ -743,6 +743,25 @@ workers:
               name: ""
               # -- the REST worker SASL password secret key for client connection to Kafka
               key: ""
+    # REST Worker TLS configuration
+    tls:
+      # -- the name of the Kubernetes secret containing the certificate, private key and CA certificate; if not specified, a new secret will be created and certificates/keys generated
+      secretName: ""
+      generation:
```
davidcurrie, line 751: missing comment

davidcurrie (changes requested):

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: Changed.

charts/corda-lib/templates/_helpers.tpl
vkolomeyko: Sure, I have incorporated annotations diffing as well.

davidcurrie (approved):
LGTM 🚀 

Tested this locally by deploying a cluster with `--set "workers.rest.tls.generation.altNames[0]=localhost"`. I could then execute a "secure" curl against the REST API as follows:
```
kubectl port-forward --namespace dcurrie deployment/corda-rest-worker 8888 &
curl --cacert <(kubectl get secret corda-rest-tls -o go-template='{{ index .data "ca.crt" | base64decode }}') -X POST https://localhost:8888/api/v1/hello\?addressee\=David -u admin:admin
Handling connection for 8888
Hello, David! (from admin)
```
The only issue I've identified is that, if `altNames` is updated, the certificate is regenerated but the REST worker deployment is not rolled over. We can either document that the deployment needs to be rolled manually or use [this trick](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments). Either way, I'm happy to see this merged as-is.

filipesoliveira (approved):

PR #3458, opened 2023-03-23
CORE-10385 Custom query execution (part 1)

nkovacsx:
### Overview

This PR is adding the very basic custom query execution. 

### Test notes

Added a new test flow called `UtxoCustomQueryDemoFlow` which will execute the dummy named query (`DummyUtxoVaultNamedQueryFactory`) that is basically doing a select on the JSON column.

This flow is executed in two test cases, one checks the offset/limit as well.


applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/ledger/UtxoLedgerTests.kt
```diff
@@ -91,6 +91,48 @@ class UtxoLedgerTests {
...
+
+        val customQueryFlowResult = awaitRpcFlowFinished(aliceHoldingId, customQueryFlowId)
+        assertThat(customQueryFlowResult.flowStatus).isEqualTo(RPC_FLOW_STATUS_SUCCESS)
+        assertThat(customQueryFlowResult.flowError).isNull()
+
+        val parsedResponse = objectMapper.readValue(customQueryFlowResult.flowResult!!, CustomQueryFlowResponse::class.java)
+
+        assertThat(parsedResponse.results).isNotEmpty
+        assertThat(parsedResponse.results).hasSize(1)
+    }
```
lankydan: These tests should probably be moved to the e2e tests. We should keep a simple, can execute a query and get results in the smoke tests though.

components/ledger/ledger-persistence/build.gradle
```diff
@@ -30,11 +30,13 @@ dependencies {
 
     implementation project(':libs:crypto:cipher-suite')
     implementation project(':libs:flows:external-event-responses')
+    implementation project(':libs:flows:flow-api')
```
lankydan: This doesn't seem right. What do you need the `flow-api` for?

components/ledger/ledger-persistence/build.gradle
```diff
@@ -30,11 +30,13 @@ dependencies {
...
     implementation project(':libs:ledger:ledger-common-data')
     implementation project(':libs:ledger:ledger-consensual-data')
     implementation project(':libs:ledger:ledger-utxo-data')
     implementation project(':libs:messaging:messaging')
     implementation project(":libs:serialization:json-validator")
+    implementation project(':libs:serialization:serialization-checkpoint-api')
```
lankydan: Same here, checkpoints only occur in the flow worker.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/data/VaultNamedQuery.kt
```diff
@@ -10,7 +10,7 @@ import net.corda.v5.ledger.utxo.query.VaultNamedQueryTransformer
  */
 data class VaultNamedQuery(
     val name: String,
-    val jsonString: String?,
+    val whereJson: String?,
```
lankydan: I changed this in my PR, so that it can support different types of queries in the future.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -0,0 +1,81 @@
...
+    service = [
+        VaultNamedQueryExecutor::class,
+        UsedByPersistence::class
+    ],
+    property = [
+        SandboxConstants.CORDA_MARKER_ONLY_SERVICE
+    ],
+    scope = ServiceScope.PROTOTYPE
+)
+class VaultNamedQueryExecutorImpl @Activate constructor(
```
lankydan, line 32: Is there a reason to separate this from the `UtxoRequestHandlerSelectorImpl` path that the current utxo persistence code takes. I also think the execution of the query code itself should occur in `UtxoPersistenceService`.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -0,0 +1,81 @@
...
+        log.debug("Executing query: ${request.queryName}")
+
+        val vaultNamedQuery = registry.getQuery(request.queryName)
+
+        require(vaultNamedQuery != null) { "Query with name ${request.queryName} could not be found!" }
+
+        val resultList = entitySandboxService.get(holdingIdentity.toCorda())
+            .getEntityManagerFactory()
+            .transaction { em ->
+                // TODO For now we only select the tx id but we might need the whole entity?
```
lankydan: Yeah, we need the whole entity so it can be fed into the filtering and transforming code.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoExecuteCustomQueryHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+    override fun execute(): List<Record<*, *>> {
+
+        return try {
+            val response = vaultNamedQueryExecutor.executeQuery(holdingIdentity, request)
+            listOf(externalEventResponseFactory.success(
+                externalEventContext,
+                response
+            ))
+        } catch (e: Exception) {
+            listOf(externalEventResponseFactory.platformError(
```
lankydan: I think we just rethrow here and default to the current exception handling. This means we'll do retries in case of database errors etc. Or just remove the try/catch.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRequestHandlerSelectorImpl.kt
```diff
@@ -96,6 +101,15 @@ class UtxoRequestHandlerSelectorImpl @Activate constructor(
                     persistenceService
                 )
             }
+            is FindWithNamedQuery -> {
+                UtxoExecuteCustomQueryHandler(
```
lankydan: `UtxoExecuteNamedQueryHandler`.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/LedgerResultSetImpl.kt
```diff
@@ -0,0 +1,10 @@
+package net.corda.ledger.utxo.flow.impl.persistence
+
+import net.corda.v5.application.persistence.PagedQuery
+
+// TODO Should we have a common `PagedQuery.ResultSet` implementation?
+data class LedgerResultSetImpl<R>(
```
lankydan: Yes, move `ResultSetImpl` into `flow-api`, this module should already depend on it. It might look funny but its a pojo at the end of the day. We could add a factory but don't want to waste time on that.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/VaultNamedParameterizedQueryImpl.kt
```diff
@@ -0,0 +1,82 @@
...
+import net.corda.ledger.utxo.flow.impl.persistence.external.events.VaultNamedQueryExternalEventFactory
+import net.corda.v5.application.persistence.PagedQuery
+import net.corda.v5.application.persistence.ParameterizedQuery
+import net.corda.v5.application.serialization.SerializationService
+import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.ledger.utxo.query.VaultNamedParameterizedQuery
+import java.nio.ByteBuffer
+import java.time.Instant
+
+class VaultNamedParameterizedQueryImpl<T>(
```
lankydan, line 15: We might be able to use delegation to simplify this.

```kotlin
class VaultNamedParameterizedQueryImpl<T>(
  private val delegate: ParameterizedQuery,
  ....
) VaultNamedParameterizedQuery, NamedParameterizedQuery by delegate {

 // still need to override execute because its implementation will be different.
```

To do this though, we need to move `PagedQueryFactory` into `flow-api` so that we can call it from our ledger persistence service. But, I'd rather get this done and change that later.

simulator/runtime/src/main/kotlin/net/corda/simulator/runtime/ledger/utxo/SimUtxoLedgerService.kt
```diff
@@ -77,6 +78,10 @@ class SimUtxoLedgerService(
         TODO("Not yet implemented")
     }
 
+    override fun <R : Any?> query(queryName: String, resultClass: Class<R>): VaultNamedParameterizedQuery<R> {
+        TODO("Not yet implemented")
+    }
```
lankydan, line 84: @blsemo What do we want to do here? This is not something that can be easily implemented.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -0,0 +1,81 @@
...
+                )
+
+                request.parameters.forEach {
+                    query.setParameter(it.key, serializationService.deserialize(it.value.array(), Any::class.java))
+                }
+
+                query.firstResult = request.offset
+                query.maxResults = request.limit
+
+                query.resultList
```
lankydan: We need to pass the results through the filtering, transforming and collecting.

lankydan: Just occurred to me, in the builder.

```
override fun filter(filter: VaultNamedQueryFilter<*>): VaultNamedQueryBuilder {
        this.filter = filter
        return this
    }

    override fun map(mapper: VaultNamedQueryTransformer<*, *>): VaultNamedQueryBuilder {
        this.mapper = mapper
        return this
    }
```

Can we make sure we document that filtering is always applied before mapping.

Can we also make it throw if we try to set either of these when they're already been set, we can consider tidying this up later if we have time.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -0,0 +1,81 @@
...
+            .getEntityManagerFactory()
+            .transaction { em ->
+                // TODO For now we only select the tx id but we might need the whole entity?
+                val query = em.createNativeQuery(
+                    "SELECT transaction_id FROM $UTXO_VISIBLE_TX_TABLE " +
+                            vaultNamedQuery.whereJson // TODO Add timestamp limit logic
+                )
+
+                request.parameters.forEach {
+                    query.setParameter(it.key, serializationService.deserialize(it.value.array(), Any::class.java))
```
lankydan: We need to extract the named parameter arguments from the input request and call `setParameter` on each of them individually.

lankydan: See `PersistenceServiceInternal.findWithNamedQuery` for how we did this for normal queries, for now just copy its code and take what you need.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/VaultNamedParameterizedQueryImpl.kt
lankydan: TODO, probably raise a ticket to address this later.

components/ledger/ledger-persistence/build.gradle
nkovacsx: Removed

components/ledger/ledger-persistence/build.gradle
nkovacsx: Removed

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/VaultNamedParameterizedQueryImpl.kt
nkovacsx: Raised CORE-12032

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/ledger/UtxoLedgerTests.kt
nkovacsx: Removed one test, kept one

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -51,31 +57,108 @@ class VaultNamedQueryExecutorImpl @Activate constructor(
...
+            vaultNamedQuery.filter?.filter(it, deserializedParams) ?: true
+        }.map {
+            vaultNamedQuery.mapper?.transform(it, deserializedParams) ?: it
+        }
+
+        // Once filtering and transforming are done collector function can be applied (if present)
+        val collectedResults = vaultNamedQuery.collector?.collect(
+            filteredAndTransformedResults,
+            deserializedParams
+        )?.results?.filterNotNull() ?: filteredAndTransformedResults
```
lankydan, line 79: I think we should allow `null`s to be returned if that is what the user wanted to do. We should test that as well to see what happens at serialization (we might have to skip it when serializing).

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -51,31 +57,108 @@ class VaultNamedQueryExecutorImpl @Activate constructor(
...
+    ): List<String> {
+        return entitySandboxService.get(holdingIdentity.toCorda())
             .getEntityManagerFactory()
             .transaction { em ->
-                // TODO For now we only select the tx id but we might need the whole entity?
-                val query = em.createNativeQuery(
+                var query = em.createNativeQuery(
                     "SELECT transaction_id FROM $UTXO_VISIBLE_TX_TABLE " +
-                            vaultNamedQuery.whereJson // TODO Add timestamp limit logic
+                            whereJson
```
lankydan: You removed the TODO for the timestamp limit but didn't implement it.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -51,31 +57,108 @@ class VaultNamedQueryExecutorImpl @Activate constructor(
...
 
-        return EntityResponse(
-            resultList.filterNotNull().map { ByteBuffer.wrap(serializationService.serialize(it).bytes) }
-        )
+    /**
+     * A function that fetches the contract states that belong to the given transaction IDs. The data stored in the
+     * component table will be deserialized into contract states using component groups.
+     */
+    private fun fetchContractStates(
+        txIds: List<String>,
```
lankydan: can we go for `transactionIds` (I've always haved the `txIds` naming)

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -103,19 +103,27 @@ class VaultNamedQueryExecutorImpl @Activate constructor(
...
             .transaction { em ->
-                var query = em.createNativeQuery(
+                val query = em.createNativeQuery(
                     "SELECT transaction_id FROM $UTXO_VISIBLE_TX_TABLE " +
                             whereJson
                 )
 
-                deserializedParams.forEach {
-                    query = query.setParameter(it.key, it.value)
+                request.parameters.filter { it.value != null}.forEach { rec ->
```
lankydan: Formatting (just reformat the whole file)

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/registration/impl/VaultNamedQueryBuilderImpl.kt
```diff
@@ -30,11 +30,17 @@ class VaultNamedQueryBuilderImpl(
     }
 
     override fun filter(filter: VaultNamedQueryFilter<*>): VaultNamedQueryBuilder {
+        require(this.filter == null) {
+            "Mapper function has already been set!"
```
lankydan: Should be "Filter"

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/registration/impl/VaultNamedQueryBuilderImpl.kt
```diff
@@ -30,11 +30,17 @@ class VaultNamedQueryBuilderImpl(
...
+        }
         this.filter = filter
         return this
     }
 
     override fun map(mapper: VaultNamedQueryTransformer<*, *>): VaultNamedQueryBuilder {
+        require(this.mapper == null) {
```
lankydan: Can we do some renaming internally and the API here so its `transformer: VaultNamedQueryTransformer` instead of `mapper`. There's a chance it needs an API bump because of the build failing if there are any warnings....

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/data/VaultNamedQuery.kt
```diff
@@ -11,9 +12,9 @@ import net.corda.v5.ledger.utxo.query.VaultNamedQueryTransformer
 data class VaultNamedQuery(
     val name: String,
     val query: ParsedQuery,
-    val filter: VaultNamedQueryFilter<*>?,
-    val mapper: VaultNamedQueryTransformer<*, *>?,
-    val collector: VaultNamedQueryCollector<*, *>?
+    val filter: VaultNamedQueryStateAndRefFilter<ContractState>?,
+    val mapper: VaultNamedQueryStateAndRefTransformer<ContractState, Any>?,
```
lankydan: We should still be able to use the `VaultNamedQueryFilter` and `VaultNamedQueryStateAndRefTransformer` interfaces here but with `Any` as their generic types.

simulator/runtime/src/main/kotlin/net/corda/simulator/runtime/ledger/utxo/SimUtxoLedgerService.kt
blsemo: As far as I know, the simulator won't be part of the Corda 5.0 release, so not implementing this should be ok for now.
@mattbradburyr3, can you confirm?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/VaultNamedParameterizedQueryImpl.kt
```diff
@@ -0,0 +1,99 @@
...
+            "Offset needs to be provided and needs to be a positive number to execute the query."
+        }
+        require(limitValue != null && limitValue > 0) {
+            "Limit needs to be provided and needs to be a positive number to execute the query."
+        }
+
+        // If timestamp limit is not set we default to the current time
+        if (timestampLimit == null) {
+            setCreatedTimestampLimit(Instant.now())
+        }
```
lankydan: Remember to validate that its not in the future at this point as well.

lankydan (commented):
Left 2 comments but looks good once addressed.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/VaultNamedParameterizedQueryImpl.kt
nkovacsx: I'm not exactly sure what you mean? We're setting it to `Instant.now()` if it was not provided before. I might be missing something though.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
nkovacsx: Actually I just realised that `null`s will not be returned even if I remove `filterNotNull` because `fetchStateAndRefs` won't allow `null`s to be returned.  That's the same logic in `UtxoPersistenceServiceImpl` so I'm not sure if we want to tweak that to allow nulls? 

I'm not entirely sure how nulls could be fetched in the first place because the table (`utxo_transaction_component`) does not allow null values.

The only scenario I can imagine if the `fetchTransactionIdsForRequest` would somehow return null objects.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/VaultNamedParameterizedQueryImpl.kt
blsemo: If one has been provided, it must not be in the future. Providing a future timestamp is not allowed as that means that the query content can change on reruns.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
lankydan: The transformations users apply to their queries could return nulls as they can return whatever they want. If we're going to filter nulls out after transforming and collecting then we should document on their APIs that nulls will be filtered out and not returned.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/AmqpSerializationTests.kt
```diff
@@ -12,10 +12,12 @@ import net.corda.e2etest.utilities.startRpcFlow
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Assertions.assertAll
 import org.junit.jupiter.api.BeforeAll
+import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 import java.util.UUID
 
 @Suppress("Unused")
+@Disabled
```
lankydan: Remember to remove these `@Disabled` annotations.

lankydan: Nvm you did.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/VaultNamedParameterizedQueryImpl.kt
```diff
@@ -82,14 +86,15 @@ class VaultNamedParameterizedQueryImpl<T>(
     }
 
     override fun setCreatedTimestampLimit(timestampLimit: Instant): VaultNamedParameterizedQuery<T> {
-        require(this.timestampLimit == null) { "Timestamp limit is already set." }
+        require(getTimestampLimitParameter() == null) { "Timestamp limit is already set." }
```
lankydan: I think we should allow repeated setting of the timestamp, there is no issue with setting it multiple times imo

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/VaultNamedParameterizedQueryImpl.kt
```diff
@@ -67,10 +66,15 @@ class VaultNamedParameterizedQueryImpl<T>(
         }
 
         // If timestamp limit is not set we default to the current time
-        if (timestampLimit == null) {
+        if (getTimestampLimitParameter() == null) {
```
lankydan: Can we change this logic a bit, which can mean we don't need the `!!` and also don't have to validate the timestamp if we just set it.

```kotlin
getTimestampLimitParameter()?.let { 
  require(it <= Instant.now()) {
            "Timestamp limit must not be in the future."
        }
} ?: setCreatedTimestampLimit(Instant.now())
```

Something like that should work.

Can we also rename `getTimestampLimitParameter` to `getCreatedTimestampLimit` so it matches the setter.

nkovacsx: Good idea thanks

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/VaultNamedParameterizedQueryImpl.kt
nkovacsx: ok

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
nkovacsx: That's a good point, but with the current definition of filter/mapper/collector in the `VaultNamedQuery` we don't allow `null`s as `Any` is not nullable:
```
val filter: VaultNamedQueryFilter<Any>?,
val mapper: VaultNamedQueryTransformer<Any, Any>?,
val collector: VaultNamedQueryCollector<Any, Any>?
```

Should we change these to `Any?` ?

components/ledger/ledger-persistence/src/main/java/net/corda/ledger/persistence/query/execution/impl/package-info.java
```diff
@@ -0,0 +1,4 @@
+@Export
+package net.corda.ledger.persistence.query.execution.impl;
```
blsemo: Why does an impl package need to be exported? That is a smell.

components/ledger/ledger-persistence/src/main/java/net/corda/ledger/persistence/query/parsing/impl/package-info.java
```diff
@@ -0,0 +1,4 @@
+@Export
+package net.corda.ledger.persistence.query.parsing.impl;
```
blsemo: Another exported impl package.

components/ledger/ledger-persistence/src/main/java/net/corda/ledger/persistence/query/registration/impl/package-info.java
```diff
@@ -0,0 +1,4 @@
+@Export
+package net.corda.ledger.persistence.query.registration.impl;
```
blsemo: And another exported impl package.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -0,0 +1,208 @@
...
+
+        // Fetch the transaction IDs for the given request
+        val transactionIds = fetchTransactionIdsForRequest(
+            request,
+            vaultNamedQuery.query.query,
+            holdingIdentity
+        )
+
+        // Fetch the contract states for the given transaction IDs
+        val contractStateResults = fetchStateAndRefs(transactionIds, holdingIdentity)
```
blsemo: I am not 100% convinced that this is right? So we use the JSON query (that works on states) to fetch transaction ids, and then load all states from these transaction ids? Wouldn't we end up with a lot of states we don't actually want?

lankydan (dismissed):
LGTM, waiting for @blsemo to finish reviewing.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -0,0 +1,208 @@
...
+            throw NullParameterException(msg)
+        }
+
+        return entitySandboxService.get(holdingIdentity.toCorda())
+            .getEntityManagerFactory()
+            .transaction { em ->
+                val query = em.createNativeQuery(
+                    "SELECT transaction_id FROM $UTXO_VISIBLE_TX_TABLE " +
+                            whereJson +
+                            " AND consumed <= :$TIMESTAMP_LIMIT_PARAM_NAME"
```
lankydan: As @blsemo pointed out, we need to get the leaf index of a state along with its transaction id. Without this pair, we're just finding all states within a transaction that met the custom where clause, rather than just the states we're looking for.

blsemo (commented):
The rest looks fine

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -0,0 +1,189 @@
...
+                ComponentLeafDto(
+                    t[0] as String, // transactionId
+                    (t[1] as Number).toInt(), // groupIndex
+                    (t[2] as Number).toInt(), // leafIndex
+                    t[3] as ByteArray // data
+                )
+            }.groupBy { it.groupIndex }
+
+        val outputInfos = componentGroups[UtxoComponentGroup.OUTPUTS_INFO.ordinal]
+            ?.associate { Pair(it.leafIndex, it.data) }
```
blsemo: Doesn't this break if there's data from more than one transaction, with the same leaf index?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -0,0 +1,189 @@
...
+
+                request.parameters.filter { it.value != null }.forEach { rec ->
+                    val bytes = rec.value.array()
+                    query.setParameter(rec.key, serializationService.deserialize(bytes))
+                }
+
+                query.firstResult = request.offset
+                query.maxResults = request.limit
+
+                query.resultList as List<Tuple>
```
blsemo, line 122: Does this execute the query under the hood or are we missing a query.execute()?

blsemo (commented):
It looks like it should work apart from that the result handling needs to take transaction ids into account as well - outputs are uniquely keyed by transactionId/groupIndex/leafIndex.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
lankydan: I think it executes it. Pretty sure Nandor tested it so it should work.

nkovacsx: It does execute it.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -0,0 +1,189 @@
...
+            .getEntityManagerFactory()
+            .transaction { em ->
+                val query = em.createNativeQuery(
+                    "SELECT tc.transaction_id, tc.group_idx, tc.leaf_idx, tc.data FROM " +
+                            "$UTXO_TX_COMPONENT_TABLE AS tc " +
+                            "JOIN $UTXO_VISIBLE_TX_TABLE AS visible_states " +
+                                "ON visible_states.transaction_id = tc.transaction_id " +
+                            "$whereJson " +
+                            "AND tc.group_idx IN (:groupIndices) " +
+                            "AND visible_states.consumed <= :$TIMESTAMP_LIMIT_PARAM_NAME",
```
lankydan: `visible_states.consumed` needs to be `visible_states.created`

lankydan (dismissed):

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -0,0 +1,190 @@
...
+
+        return utxoComponentDtos.map {
+            val info = serializationService.deserialize<UtxoOutputInfoComponent>(it.info)
+            val contractState = serializationService.deserialize<ContractState>(it.data)
+            StateAndRefImpl(
+                state = TransactionStateImpl(contractState, info.notaryName, info.notaryKey, info.getEncumbranceGroup()),
+                ref = StateRef(parseSecureHash(it.transactionId), it.leafIndex)
+            )
+        }
+    }
```
blsemo: Why do we need two mapping steps here? I think we could construct the StateAndRefs in one go.
```suggestion
        return componentGroups[UtxoComponentGroup.OUTPUTS.ordinal]?.map {
            val serializedInfo = outputInfos[it.transactionId to it.leafIndex]

            requireNotNull(serializedInfo) {
                "Missing output info at index [${it.leafIndex}] for UTXO transaction with ID [${it.transactionId}]"
            }
            val info = serizializationService.deserialize<UtxoOutputInfoComponent>(serializedInfo.data)
            val contractState = serializationService.deserialize<ContractState>(it.data)
            StateAndRefImpl(
                state = TransactionStateImpl(contractState, info.notaryName, info.notaryKey, info.getEncumbranceGroup()),
                ref = StateRef(parseSecureHash(it.transactionId), it.leafIndex)
            )
          } ?: emptyList()
    }
```

blsemo (dismissed):
Looks much better - one question about the query execution left, but it's correct as is, so approving tentatively.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
nkovacsx: good point!

blsemo (dismissed):

blsemo (dismissed):
Re-approved

lankydan (dismissed):

lankydan (approved):

ronanbrowne (approved):

PR #3461, opened 2023-03-24
CORE-10575 Add a membership actions processor to the member worker

williamvigorr3:
Moved the logic which distributes the updated member info from the registration processor into the member ship actions processor.  This is action is now triggered by the Registration processor. This code requeues distribute member info request if the group parameters epoch is smaller than the expected epoch. So we can handle delayed publication of the group parameters. Previously, we would wait for an exact version of the group parameters (but this might mean we try and distribute forever if a new group parameters gets published simultaneously). I also added code to requeue the request if the member info version is serial is smaller than expected (for the same reason). Requires: https://github.com/corda/corda-api/pull/992


components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/MemberOpsServiceImpl.kt
```diff
@@ -58,12 +66,21 @@ class MemberOpsServiceImpl @Activate constructor(
...
+    private val cordaAvroSerializationFactory: CordaAvroSerializationFactory,
+    @Reference(service = MerkleTreeProvider::class)
+    private val merkleTreeProvider: MerkleTreeProvider,
 ) : MemberOpsService {
     private companion object {
         private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
         const val RPC_GROUP_NAME = "membership.ops.rpc"
         const val RPC_CLIENT_NAME = "membership.ops.rpc"
         const val ASYNC_GROUP_NAME = "membership.ops.async"
+        const val ACTIONS_GROUP_NAME = "membership.actions"
```
nikinagy: I wonder if we need to call this as "membership.ops.actions" to match the others.

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/MembershipActionsProcessor.kt
```diff
@@ -0,0 +1,64 @@
...
+class MembershipActionsProcessor(
+    membershipQueryClient: MembershipQueryClient,
+    cipherSchemeMetadata: CipherSchemeMetadata,
+    clock: Clock,
+    cryptoOpsClient: CryptoOpsClient,
+    cordaAvroSerializationFactory: CordaAvroSerializationFactory,
+    merkleTreeProvider: MerkleTreeProvider,
+    membershipConfig: SmartConfig,
+    groupReaderProvider: MembershipGroupReaderProvider
+): DurableProcessor<String, MembershipActionsRequest> {
```
nikinagy, line 29: why we ended up with using DurableProcessor? why not use a compacted topic and processor instead?

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoAction.kt
```diff
@@ -0,0 +1,191 @@
...
+                            "available.")
+                }
+            }
+        } ?: return recordToRequeueDistribution(key, request) {
+            logger.info("Retrieved group parameters are null. Republishing the distribute command to be processed later when set of " +
+                    "group parameters with epoch ${request.minimumGroupParametersEpoch} is available.")
+        }
+        val allMembers = groupReader.lookup()
+        val allActiveMembersExcludingMgm = allMembers.filter {
+            it.status == MemberInfoExtension.MEMBER_STATUS_ACTIVE && !it.isMgm
```
nikinagy: You don't need to filter here for active status, the lookup() by default does that for you.

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoAction.kt
```diff
@@ -0,0 +1,191 @@
...
+    fun process(key: String, request: DistributeMemberInfo): List<Record<String, *>> {
+        val approvedBy = request.mgm
+        val updatedMember = request.updatedMember
+
+        val groupReader = groupReaderProvider.getGroupReader(approvedBy.toCorda())
+
+        // Verify that the group parameters from the reader are the ones persisted during registration approval.
+        // If not, republish the distribute command to be processed later when the updated set of group parameters
+        // is available.
+        val groupParameters = groupReader.groupParameters?.apply {
```
nikinagy: We can probably move this check a bit down and only do this validation if the updated member was a notary? Unless, there is a special and strong reason why we need to do this check every time an update comes in.

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoAction.kt
```diff
@@ -0,0 +1,191 @@
...
+import net.corda.v5.membership.MemberInfo
+import net.corda.virtualnode.HoldingIdentity
+import net.corda.virtualnode.toAvro
+import net.corda.virtualnode.toCorda
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import java.util.UUID
+
+@Suppress("LongParameterList")
+class DistributeMemberInfoAction(
```
nikinagy: `DistributeMemberInfoAction` sounds more like a command, than something that processes the requests and does the actual distribution. Can we find a more suitable name?

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoAction.kt
```diff
@@ -0,0 +1,191 @@
...
+            logger.info("Retrieved group parameters are null. Republishing the distribute command to be processed later when set of " +
+                    "group parameters with epoch ${request.minimumGroupParametersEpoch} is available.")
+        }
+        val allMembers = groupReader.lookup()
+        val allActiveMembersExcludingMgm = allMembers.filter {
+            it.status == MemberInfoExtension.MEMBER_STATUS_ACTIVE && !it.isMgm
+        }
+        val updatedMemberInfo = allActiveMembersExcludingMgm.firstOrNull { it.name.toString() == updatedMember.x500Name }
+            ?: return recordToRequeueDistribution(key, request) {
+                logger.info("The MemberInfo retrieved from the message bus for ${updatedMember.x500Name} is null. Republishing the " +
```
nikinagy:
```suggestion
                logger.info("The MemberInfo retrieved from the message bus for ${updatedMember.x500Name} is not present yet. Republishing the " +
```

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoAction.kt
```diff
@@ -0,0 +1,191 @@
...
+                        "distribute command to be processed later when the MemberInfo is available.")
+            }
+        if (request.minimumUpdatedMemberSerial > updatedMemberInfo.serial) {
+            return recordToRequeueDistribution(key, request) {
+                logger.info("The MemberInfo retrieved from the message bus for ${updatedMember.x500Name} has serial" +
+                        " ${updatedMemberInfo.serial}. Republishing the distribute command to be processed later when the MemberInfo with" +
```
nikinagy:
```suggestion
                        " ${updatedMemberInfo.serial}, which is an old version. Republishing the distribute command to be processed later when the MemberInfo with" +
```

components/membership/membership-service-impl/src/test/kotlin/net/corda/membership/service/impl/MembershipActionsProcessorTest.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+    private val processor = MembershipActionsProcessor(mock(), mock(), mock(), mock(), mock(), mock(), mock(), mock())
+
+    @AfterEach
+    fun cleanUp() {
+        distributeMemberInfoAction.close()
+    }
+
+    @Test
+    fun `processor forwards distribute member info request to DistributeMemberInfoAction`() {
```
nikinagy, line 35: Please test the non-happy too with having the invalid request type to increase test coverage.

components/membership/membership-service-impl/src/test/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoActionTest.kt
```diff
@@ -0,0 +1,385 @@
...
+            on { name } doReturn holdingIdentity.x500Name
+            on { groupId } doReturn holdingIdentity.groupId
+            on { serial } doReturn MEMBER_INFO_SERIAL
+        }
+    }
+
+    private val owner = createHoldingIdentity("owner")
+    private val member = createHoldingIdentity("member")
+    private val action = DistributeMemberInfo(owner.toAvro(), member.toAvro(), EPOCH, MEMBER_INFO_SERIAL)
+    private val key = "key"
```
nikinagy: can be const

components/membership/membership-service-impl/src/test/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoActionTest.kt
```diff
@@ -0,0 +1,385 @@
...
+            .hasSize(1)
+            .allSatisfy {
+                assertThat(it.topic).isEqualTo(Schemas.Membership.MEMBERSHIP_ACTIONS_TOPIC)
+                assertThat(it.key).isEqualTo(key)
+                assertThat((it.value as? MembershipActionsRequest)?.request).isEqualTo(action)
+            }
+    }
+
+    @Test
+    fun `process republishes the distribute command if expected member info serial is not available via the group reader`() {
```
nikinagy:
```suggestion
    fun `process republishes the distribute command if expected member info version is not available via the group reader`() {
```

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoAction.kt
```diff
@@ -0,0 +1,191 @@
...
+            is MembershipQueryResult.Failure -> return recordToRequeueDistribution(key, request) {
+                logger.warn("Failed to query for the members signature: ${membersSignaturesQuery.errorMsg}. Distributing the member info " +
+                        "will be reattempted.")
+            }
+        }
+        val membershipPackageFactory = createMembershipPackageFactory(mgm, allActiveMembersExcludingMgm, membersSignatures)
+
+        // Send all approved members from the same group to the newly approved member over P2P
+        val allMembersExcludingMgmPackage = try {
+            membershipPackageFactory.invoke(allActiveMembersExcludingMgm, groupParameters)
```
nikinagy: what is invoke calling here exactly?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/RegistrationManagementServiceImpl.kt
```diff
@@ -167,9 +167,6 @@ class RegistrationManagementServiceImpl @Activate constructor(
                         cordaAvroSerializationFactory,
                         membershipPersistenceClient,
                         membershipQueryClient,
-                        cryptoOpsClient,
```
nikinagy, line 173: Please remove these from the constructor as well.

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/MembershipActionsProcessor.kt
williamvigorr3: If we use a compacted processor, we can miss events (if we add more actions). If someone added two different actions we might miss the first one. At the moment we are keying this by `"${approvedMember.x500Name}-${approvedMember.groupId}",` changing that and using a random ID would fix that but then all members won't processed by the same member worker.

dimosr: In general, durable subscriptions / processors are a better fit for processing events with at-least-once / exactly-once guarantees (where you don't want to miss events), while compacted subscriptions / processors are a better fit for keeping track of a KV dataset (where you are ok with missing events as long as you can view the latest value for each key).

nikinagy: Okay, I kind of missed that we wanted to do make sure we process all of the events. I thought we just wanted to send the latest action (so latest version of member) in case if they were more than 1 update for the member nearly at the same time.

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoAction.kt
williamvigorr3: I think it is better to make minimumGroupParametersEpoch nullable (and do the check if it is not null). Still a bit on the fence about if we should do this or leave this as it is i.e. make the user to specify I minimum version to be distributed). If we decide we need update the group parameters for some other reason (e.g. this https://r3-cev.atlassian.net/browse/CORE-9519 ) then we won't need to change this code later.

nikinagy: Good point, that makes sense.

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoAction.kt
williamvigorr3: I renamed this `DistributeMemberInfoActionHandler`.

components/membership/membership-service-impl/src/test/kotlin/net/corda/membership/service/impl/MembershipActionsProcessorTest.kt
williamvigorr3: Good point. I added `processor returns an empty list if unknown action`.

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoAction.kt
williamvigorr3: I removed this (it was left over from the old implementation). It is calling the factory.

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoAction.kt
williamvigorr3: I will make both the `minimumGroupParametersEpoch` and the `minimumUpdatedMemberSerial` nullable. If null we will just use the latest version (from the message bus).

nikinagy: I left a comment on the api PR. I don't think they will be null ever as both are non-nullable.

components/membership/membership-service-impl/src/test/kotlin/net/corda/membership/service/impl/MembershipActionsProcessorTest.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    @Test
+    fun `processor returns an empty list if unknown action`() {
+        val result = processor.onNext(listOf(Record(
+            Schemas.Membership.MEMBERSHIP_ACTIONS_TOPIC,
+            KEY,
+            MembershipActionsRequest(Unit),
+        )))
+
+        assertThat(result).isEmpty()
```
nikinagy, line 56: Can we check with verify that the handler.process() was never called? My issue with this test is that we are returning emptyList() when the call was successful too.

nikinagy (dismissed):
LGTM

dimosr (approved):

PR #3463, opened 2023-03-24
NOTICK Change Helm version scheme in order not to collide with main/beta builds.

szymonsztuka:
Change Helm version to 5.0.0-Interop so it doesn't clash with builds with main branch /beta.


davidcurrie (approved):
👍

wzur-r3 (approved):
LGTM

PR #3464, opened 2023-03-24
CORE-10424 : Changes to manage session for interop request

SaurabhR3:
1. Changes to manage session for interop request
2. Changed `InteropProcessor` from `DurableProcessor` to `StateAndEventProcessor` to maintain the states
3. In the `InteropListener` you can see `FlowMapperEvent` which is used for time being, that can be replaced when changes will be merged for CORE-10420
4. Removed previous implementation


szymonsztuka (commented):
I thought that we decided to work on solution ready for groups bigger than 2 members, so the mechanism using `StateAndEventProcessor`, which  will invalidate this PR entirely.

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropAliasProcessor.kt
```diff
@@ -24,6 +24,11 @@ class InteropAliasProcessor : CompactedProcessor<String, HostedIdentityEntry> {
         fun getRealHoldingIdentity(recipientId: String?): HoldingIdentity? {
             return identityMappingCache[recipientId]
         }
+
+        fun findFakeHoldingIdentityByGroupId(groupId: String, fakeIdentity: String) : HoldingIdentity  {
+            val filteredMap = identityMappingCache.filter {(key, value) -> key != fakeIdentity && value.groupId == groupId }
```
yarivus: Why are you searching the inequality of fakeIdentity ?
Shouldn't it be equal by fakeIdentity && groupId ?

yarivus (changes requested):

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropAliasProcessor.kt
SaurabhR3: Its removed now

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropService.kt
```diff
@@ -58,6 +60,7 @@ class InteropService @Activate constructor(
         private const val REGISTRATION = "REGISTRATION"
         private const val CONFIG_HANDLE = "CONFIG_HANDLE"
         private const val GROUP_NAME = "interop_alias_translator"
+        private const val CLEANUP_TASK = "TASK"
```
yarivus:
```suggestion
        private const val CLEANUP_TASK = "CLEANUP_TASK"
```

gradle.properties
```diff
@@ -40,7 +40,7 @@ bouncycastleVersion=1.72
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.0.0.xx-SNAPSHOT to pick up maven local published copy
 #cordaApiVersion=5.0.0.xxx-SNAPSHOT
-cordaApiVersion=5.0.0-INTEROP.0-beta+
+cordaApiVersion=5.0.0-INTEROP.0-alpha-1680006496201
```
yarivus: why are you going backwards ?

yarivus (changes requested):

gradle.properties
ac101m: We need to track API versioning separately for the interop feature branch. Sharing the version number with mainline is likely to cause problems for us in the the future.

When we merge from mainline or make a breaking change, we'll increment our version number. When it comes time to merge back into mainline, we'll increment the mainline version number.

Collisions shouldn't occur because of the `-INTEROP` suffix to the corda product version.

components/interop/interop-service/src/integrationTest/kotlin/net/corda/interop/service/integration/InteropServiceIntegrationTest.kt
```diff
@@ -166,15 +168,15 @@ class InteropServiceIntegrationTest {
         flowMapperExpectedOutputMessages.let { expectedMessageCount ->
             val mapperLatch = CountDownLatch(expectedMessageCount)
             val testProcessor = P2POutMessageCounter(aliceX500Name, mapperLatch, expectedMessageCount)
-            val eventTopic = subscriptionFactory.createDurableSubscription(
+            val eventTopic = subscriptionFactory.createStateAndEventSubscription(
```
szymonsztuka: The test reader doesn't need to be of type`StateAndEventProcessor`, leave it as`DurableProcessor`. It checks what it on `P2P_OUT_TOPIC`, and this PR doesn't alter what is send on that topic and it doesn't store our internal state etc.

szymonsztuka (changes requested):

gradle.properties
SaurabhR3: When I created corda-api PR it generates the artefacts with name `alpha` : [https://ci02.dev.r3.com/job/Corda5/job/corda-api/job/PR-1010/1/](url) 
So I have to use `alpha` to make this build happy.

OR probably I need to merge the corda-api PR and then it will generate the `beta` and I can use it. I am not sure about it. Need to check.

yarivus (commented):
LGTM but pending comments

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropService.kt
SaurabhR3: Fixed in 910e547

gradle.properties
SaurabhR3: This will be fixed as part of last commit, when corda-api changes will be merged.

components/interop/interop-service/src/integrationTest/kotlin/net/corda/interop/service/integration/InteropServiceIntegrationTest.kt
SaurabhR3: Fixed in 910e547

gradle.properties
```diff
@@ -40,7 +40,7 @@ bouncycastleVersion=1.72
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.0.0.xx-SNAPSHOT to pick up maven local published copy
 #cordaApiVersion=5.0.0.xxx-SNAPSHOT
-cordaApiVersion=5.0.0-INTEROP.1-beta+
+cordaApiVersion=5.0.0-INTEROP.3-SNAPSHOT
```
szymonsztuka: You need to use corresponding Alpha build (with precise number) from https://github.com/corda/corda-api/pull/1010, after that PR is merged, you need to switch to `5.0.0-INTEROP.3-beta+` in this PR.

SaurabhR3: I have triggered the alpha build now, I will change it once it is available. Till the time build will be red

ac101m: Should also note, there's no need to pick a specific version number until tests are passing and the PR has been approved. When you get to the point of merging your api PR, just set the version number to the next available one (probably going to be 2 in this case).

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropProcessor.kt
```diff
@@ -107,28 +119,43 @@ class InteropProcessor(
...
-                    SUBSYSTEM,
-                    MembershipStatusFilter.ACTIVE
-                ), ByteBuffer.wrap(sessionEventSerializer.serialize(sessionEvent))))
-            )
+                AppMessage(
+                    AuthenticatedMessage(
+                        AuthenticatedMessageHeader(
+                            destinationIdentity,
+                            net.corda.data.identity.HoldingIdentity(
+                                state?.aliasHoldingIdentity,
```
szymonsztuka: In general it's a bit tricky reasoning when we have a saved state already, as first message may arrived to the processor from both sides (INBOUND and OUTBOUND). At least we need to prevent what if a state is null? If it's null then use value as in the previous code.
`state?.aliasHoldingIdentity ?: destinationIdentity.x500Name,`

szymonsztuka (changes requested):

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropProcessor.kt
```diff
@@ -1,5 +1,7 @@
 package net.corda.interop
 
+//TODO import commented out - see TODO adding FLOW_CONFIG below:
```
jmacmahonr3: Please add a ticket number of the ticket where we are going to do this work. If this ticket doesn't exist we can create one.

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropProcessor.kt
```diff
@@ -107,28 +119,43 @@ class InteropProcessor(
...
-                ), ByteBuffer.wrap(sessionEventSerializer.serialize(sessionEvent))))
-            )
+                AppMessage(
+                    AuthenticatedMessage(
+                        AuthenticatedMessageHeader(
+                            destinationIdentity,
+                            net.corda.data.identity.HoldingIdentity(
+                                state?.aliasHoldingIdentity,
+                                destinationIdentity.groupId
+                            ), //TODO replace groups with alias one
```
jmacmahonr3: Add Ticket Please

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropListener.kt
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    override fun onPartitionSynced(states: Map<String, InteropState>) {
+        log.info("Synced states $states")
+        val currentTime = clock.millis()
+        for (stateEntry in states) {
+            val key = stateEntry.key
+            val state = stateEntry.value
+            val expiryTime = state.expiryTime
+            if (expiryTime != null && state.status == InteropStateType.INVALID) {
```
jmacmahonr3, line 33: Maybe a dumb question but do we care if we have an expiry time but the state is still invalid ?

jmacmahonr3: I see we check it on post commit but do we need to check it here too ?

jmacmahonr3 (changes requested):
Some small comments but looking good

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropProcessor.kt
SaurabhR3: This is not my change it is from Szymon's changes, @szymonsztuka do we have ticket number for this?

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropProcessor.kt
SaurabhR3: Removed comment as we are setting alias here.

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropListener.kt
SaurabhR3: I am not 100% sure here, I just keep it consistent with flow mapper processor.

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropProcessor.kt
SaurabhR3: I believe in outbound message state must be available. If we set the `destinationIdentity.x500Name` wouldn't it be wrong? Source and Destination will be destinationIdentity.

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropProcessor.kt
szymonsztuka: CORE-12208

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropProcessor.kt
SaurabhR3: Fixed in eded1b3

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropProcessor.kt
SaurabhR3: FIxed in eded1b3

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropProcessor.kt
szymonsztuka: 1. The status for outbound message is present only if this is a message created from a responder flow, if a message is created from an initiating flow then this is the first message ever hitting the processor, and it's still outbound.
2. Even if the 1st) statement would be incorrect,  the compiler tells you this can be null `state?`, so you have second reason to have a remedy on this.

szymonsztuka (changes requested):

jmacmahonr3 (approved):
Make sure you resolve any remaining comments but LGTM

components/interop/interop-service/src/main/kotlin/net/corda/interop/InteropProcessor.kt
SaurabhR3: I am still unsure how initiating flow request can be outbound message? Also even initiating flow will be outbound still we can't just keep the source and destination same while moving message to p2p-out topic. Better to discuss it on call.

SaurabhR3: @szymonsztuka thanks for a call, fixed in c52dc76

szymonsztuka (approved):
before the merge, mereg corda-api first with increased APi version, then here change `cordaApiVersion=5.0.0-INTEROP.3-alpha-1680260631242`to  `cordaApiVersion=5.0.0-INTEROP.3-beta+`

gradle.properties
SaurabhR3: Fixed in 4054268

yarivus (approved):
LGTM

PR #3465, opened 2023-03-24
CORE-11278: Fix e2e tests

yift-r3:
The `net.corda.e2etest.membership.RegistrationApprovalTest` tests fail (for example - [here](https://ci02.dev.r3.com/job/Corda5/job/corda5-e2e-tests/view/change-requests/job/PR-47/4/)). This is because the network setup API had changed as part of the session key ID rotation.


dimosr (approved):
Code change LGTM

Can you add some description, please? Which end-to-end tests failed (the ones in the same repo or in the separate one) and could you add a link to the failed build?

PR #3466, opened 2023-03-24
CORE-9955: Allow multiple servers for p2p Gateway

yift-r3:
API pull request in https://github.com/corda/corda-api/pull/984.

Open more than one HTTP server in the gateway.


components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/GatewayConfiguration.kt
```diff
@@ -5,17 +5,9 @@ import java.time.Duration
...
-    /**
-     * Port number used when binding the HTTP server
-     */
-    val hostPort: Int,
-    /**
-     * The URL path the gateway HTTP server will listen to for requests.
+     * The gateway configurations
      */
-    val urlPath: String,
+    val servers: Collection<GatewayServerConfiguration>,
```
dimosr:
```suggestion
    val serversConfiguration: Collection<GatewayServerConfiguration>,
```

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/ReconfigurableHttpServer.kt
```diff
@@ -62,55 +53,53 @@ internal class ReconfigurableHttpServer(
...
-                    logger.info(
-                        "New server configuration for ${dominoTile.coordinatorName} on the same port, " +
-                            "HTTP server will have to go down"
-                    )
-                    serverLock.write {
-                        val oldServer = httpServer
-                        httpServer = null
+                val newServersConfiguration = newConfiguration.servers
+                if (newServersConfiguration.isEmpty()) {
+                    throw CordaRuntimeException("No servers defined!")
```
dimosr:
```suggestion
                    throw IllegalArgumentException("No server configurations defined!")
```
I believe `CordaRuntimeException` is a type supposed to be used by flows, so not sure it should be used here.

Also, given the schema you have on corda-api, I understand this is just for the sake of defensive programming. But, empty lists will practically be rejected at the point the user specifies the configuration and it won't reach that point at all. Let me know if not, because ideally that should be the way it works.

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/ReconfigurableHttpServer.kt
```diff
@@ -62,55 +53,53 @@ internal class ReconfigurableHttpServer(
...
+                }
+                newServersConfiguration.groupBy {
+                    it.hostAddress to it.hostPort
+                }.forEach { (address, configurations) ->
+                    if (configurations.size != 1) {
+                        throw CordaRuntimeException("Can not have more than one server listening to: $address")
+                    }
```
dimosr: A bit on the fence on this..

But, assuming we cannot prevent this via JSON schema, I think it's better if we are a bit more graceful with this scenario, i.e. select the first amongst them and log a warning indicating that there were multiple ones, which one was selected and which were ignored.

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/ReconfigurableHttpServer.kt
```diff
@@ -62,55 +53,53 @@ internal class ReconfigurableHttpServer(
...
+                }
+                val mutualTlsTrustManager = createTrustManagerIfNeeded(
+                    newConfiguration.sslConfig,
+                    commonComponents.trustStoresMap,
+                    dynamicCertificateSubjectStore,
+                )
+                val serversToStop = httpServers.filterKeys { configuration ->
```
dimosr, line 88: what is this logic trying to achieve?

I think the previous logic here was trying to optimise the time of unavailability in the scenario where the new configuration differs from the existing server by starting the new one before stopping the old one. 

But, looking at the logic here I am not sure a similar thing is achieved here. What I think this is doing is it's stopping first existing servers that don't have a configuration matching with the new configurations and then proceeding with the rest.


In order for do what we were doing previously, I think we'd need to 1. filter all entries in the new configuration that don't match an existing server and start servers with these configurations 2. then stop previously running servers 3. start the servers for the remaining configurations that were matching existing servers.

Btw, it could be optimised a bit further by going through the individual servers in 2 & 3 and stopping, starting one-by-one. But, probably not worth the extra effort doing that now, since people are expected to do these transitions by having two different configurations for upgrades etc.

tools/plugins/mgm/src/main/kotlin/net/corda/cli/plugins/mgm/Helpers.kt
```diff
@@ -17,13 +17,10 @@ internal object Helpers {
                 "go-template={{ .data.password | base64decode }}"
             ).start()
             if (getSecret.waitFor() != 0) {
-                throw BaseOnboard.OnboardException(
-                    "Can not get admin password. ${
-                    getSecret.errorStream.reader().readText()
-                    }"
-                )
+                "admin"
```
dimosr, line 20: what is the context behind this change?

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/ReconfigurableHttpServer.kt
```diff
@@ -1,37 +1,35 @@
...
+    private val listener: RequestListener,
     private val commonComponents: CommonComponents,
     private val dynamicCertificateSubjectStore: DynamicCertificateSubjectStore,
 ) : LifecycleWithDominoTile {
 
-    @Volatile
-    private var httpServer: HttpServer? = null
-    private val serverLock = ReentrantReadWriteLock()
+    // Map from the server configuration to a server
+    private var httpServers = ConcurrentHashMap<GatewayServerConfiguration, HttpServer>()
```
williamvigorr3: Can be `val`.

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/ReconfigurableHttpServer.kt
```diff
@@ -62,55 +53,53 @@ internal class ReconfigurableHttpServer(
...
-                        httpServer = null
+                val newServersConfiguration = newConfiguration.servers
+                if (newServersConfiguration.isEmpty()) {
+                    throw CordaRuntimeException("No servers defined!")
+                }
+                newServersConfiguration.groupBy {
+                    it.hostAddress to it.hostPort
+                }.forEach { (address, configurations) ->
+                    if (configurations.size != 1) {
+                        throw CordaRuntimeException("Can not have more than one server listening to: $address")
```
williamvigorr3: Suggest  instead of `$address`, `$address.first:address.secound` so it prints something like `0.0.0.0:8888`.

tools/plugins/mgm/src/main/kotlin/net/corda/cli/plugins/mgm/Helpers.kt
yift-r3: I don't know... Now I'm getting an error in this after using the deploy script.

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/ReconfigurableHttpServer.kt
yift-r3: The JSON scheme should prevent the user from having an empty list. But, because it's in another library, I think having another level of protection here is better.

dimosr: yes, I am happy with having an extra layer of protection here. As long as, we know an empty list will be rejected at the early point as well now.

But, I think we should move away from `CordaRuntimeException` here for the reason described above.

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/internal/InboundMessageHandler.kt
```diff
@@ -120,7 +120,7 @@ internal class InboundMessageHandler(
             gatewayMessage to LinkInMessage(gatewayMessage.payload)
         } catch (e: Throwable) {
             logger.warn("Received invalid message, which could not be deserialized", e)
-            server.writeResponse(HttpResponseStatus.BAD_REQUEST, request.source)
+            httpWriter.write(HttpResponseStatus.BAD_REQUEST, request.source )
```
williamvigorr3: Nit: remove extra whitespace after `request.source`.

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/ReconfigurableHttpServer.kt
```diff
@@ -62,55 +52,65 @@ internal class ReconfigurableHttpServer(
...
+                        httpServers.compute(configuration) { _, oldServer ->
+                            oldServer?.close()
+                            logger.info(
+                                "New server configuration, ${dominoTile.coordinatorName} will be connected to " +
+                                    "${configuration.hostAddress}:${configuration.hostPort}${configuration.urlPath}"
+                            )
+                            HttpServer(
+                                listener,
+                                newConfiguration.maxRequestSize,
+                                configuration,
```
williamvigorr3: Rename `configuration` -> `serverConfiguration`?

dimosr (dismissed):
LGTM

tools/plugins/mgm/src/main/kotlin/net/corda/cli/plugins/mgm/Helpers.kt
williamvigorr3, line 20: Did you forget to revert this?

components/gateway/src/test/kotlin/net/corda/p2p/gateway/messaging/http/ReconfigurableHttpServerTest.kt
```diff
@@ -143,17 +146,33 @@ class ReconfigurableHttpServerTest {
...
+
         configHandler.applyNewConfiguration(configuration, null, resourcesHolder)
 
-        verify(resourcesHolder).keep(serverMock.constructed().last())
+        closeable.firstValue.close()
+        verify(serverMock.constructed().last()).close()
+    }
+
+    @Test
+    fun `applyNewConfiguration will remember to forget about the servers`() {
```
williamvigorr3: This test is the same as the one above.

components/gateway/src/test/kotlin/net/corda/p2p/gateway/messaging/http/ReconfigurableHttpServerTest.kt
```diff
@@ -162,4 +181,67 @@ class ReconfigurableHttpServerTest {
...
+            ),
+            null,
+            resourcesHolder
+        )
+
+        assertThat(serverMock.constructed()).hasSize(3)
+    }
+
+    @Test
+    fun `applyNewConfiguration will throw an error if there is more than one server that is using the same host andport`() {
```
williamvigorr3: Doesn't this contradict the test above (I think you forgot to update the test name)?

tools/plugins/mgm/src/main/kotlin/net/corda/cli/plugins/mgm/Helpers.kt
yift-r3: No, we need to keep it (the password is admin when the secret is not installed)

williamvigorr3 (dismissed):
LGTM

williamvigorr3 (approved):
LGTM!

dimosr (approved):

PR #3467, opened 2023-03-24
CORE-11035 - additional-db-pool-configuration 

driessamyn:
Expose additional DB Pool configuration to the Pool Factory API.
These configuration settings broadly line up with Hikari's configuration and will allow us to tune the pool.

A follow-up PR will expose this configuration to workers and the vnode DB configuration.

The test that as added is not meant to be run as part of the PR gate. It's not really a useful test for CI, but it is useful for observing/verifying Hikari's behaviour. The reason for me doing this was to verify the pool can even scale down to zero connections when they are idle, which may be useful for vnode DB connections as the pool in that case is cached along with the Sandbox.

### DIFF
--- a/libs/db/db-core/src/main/kotlin/net/corda/db/core/DataSourceFactory.kt
+++ b/libs/db/db-core/src/main/kotlin/net/corda/db/core/DataSourceFactory.kt
@@ -1,6 +1,37 @@
 package net.corda.db.core
 
+import java.time.Duration
+import kotlin.time.DurationUnit
+import kotlin.time.toDuration
+import kotlin.time.toJavaDuration
+
 interface DataSourceFactory {
+    /**
+     * Create a new [CloseableDataSource] with the given configuration
+     *
+     * @param driverClass type of JDBC driver to use.
+     * @param jdbcUrl
+     * @param username
+     * @param password
+     * @param isAutoCommit default auto-commit behaviour.
+     * @param isReadOnly when set, and supported by the database, the pool will return read-only connections. Default - false.
+     * @param maximumPoolSize the maximum number of connection in the pool. Default - 10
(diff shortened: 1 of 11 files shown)


driessamyn:
@JamesHR3 tagging you here as this may be of interest on the performance work.
Obviously not easily consumed until I expose the config for it, but just so it's on your radar.

Dito @davidcurrie

libs/db/db-core/src/main/kotlin/net/corda/db/core/DataSourceFactory.kt
```diff
@@ -1,13 +1,50 @@
...
         driverClass: String,
         jdbcUrl: String,
         username: String,
         password: String,
         isAutoCommit: Boolean = false,
-        maximumPoolSize: Int = 10
+        isReadOnly: Boolean = false,
+        maximumPoolSize: Int = 10,
+        minimumPoolSize: Int = maximumPoolSize,
+        idleTimeout: Duration = 2.toDuration(DurationUnit.MINUTES).toJavaDuration(),
```
simon-johnson-r3: `Duration.ofMinutes(2)` would be a more concise way to do the same thing I believe
(same with other lines)

simon-johnson-r3 (commented):
One minor comment

libs/db/db-core/src/main/kotlin/net/corda/db/core/DataSourceFactory.kt
driessamyn: Y. Will change. 
I first was using kotlin.Duration, but that was a nightmare to get working in Mockito when I realised there was little benefit in using the Kotlin type.

simon-johnson-r3 (approved):

PR #3468, opened 2023-03-24
CORE-11833 - use custom kryo serializer

conal-smith:
Appears locally to improve both serialization and deserialization time:
Before:
image

After:
image

Will need a run of soak tests on this branch to gauge any improvements.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberImpl.kt
@@ -1,6 +1,5 @@
 package net.corda.flow.fiber
 
-import co.paralleluniverse.fibers.CustomFiberWriter
 import co.paralleluniverse.fibers.Fiber
 import co.paralleluniverse.fibers.FiberScheduler
 import net.corda.data.flow.state.checkpoint.FlowStackItem
@@ -13,7 +12,6 @@ import net.corda.utilities.debug
 import net.corda.utilities.trace
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
-import java.io.Serializable
 import java.nio.ByteBuffer
 import java.util.UUID
 import java.util.concurrent.CompletableFuture
@@ -25,8 +23,6 @@ class FlowFiberImpl(
     override val flowLogic: FlowLogicAndArgs,
     scheduler: FiberScheduler
 ) : Fiber<Unit>(flowId.toString(), scheduler), FlowFiber, Interruptable {
-
-    private fun interface CordaCustomFiberWriter : CustomFiberWriter, Serializable
 
     companion object {
         private val log: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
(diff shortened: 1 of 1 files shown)


chrisr3:
Quasar's `parkAndCustomSerialize` API was created by @blsemo back in 2018:
```
commit 73f5cef96ecf8b57e4184bc3a255d225b03d5600
Author: Christian Sailer <christian.sailer@r3.com>
Date:   Tue Sep 11 17:13:31 2018 +0100

    Replaced Runnable with CustomFiberWriter and added unit test for custom serialization.
```
And yet AFAICT, even Corda 4.10 still isn't using it :confused:. I don't know the history here, but it seems odd that we didn't make this change _years_ ago. Can you shed any light on this please, Christian?

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberImpl.kt
```diff
@@ -27,7 +26,7 @@ class FlowFiberImpl(
     scheduler: FiberScheduler
 ) : Fiber<Unit>(flowId.toString(), scheduler), FlowFiber, Interruptable {
 
-    private fun interface SerializableFiberWriter : FiberWriter, Serializable
+    private fun interface CordaCustomFiberWriter : CustomFiberWriter, Serializable
```
chrisr3: Do you need `CordaCustomFiberWriter` at all? You should be able to use `CustomFiberWriter` directly because it now extends `java.io.Serializable`.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberImpl.kt
```diff
@@ -131,7 +130,7 @@ class FlowFiberImpl(
     @Suspendable
     override fun <SUSPENDRETURN> suspend(request: FlowIORequest<SUSPENDRETURN>): SUSPENDRETURN {
         removeCurrentSandboxGroupContext()
-        parkAndSerialize(SerializableFiberWriter { _, _ ->
+        parkAndCustomSerialize(CordaCustomFiberWriter { _ ->
```
chrisr3: I think we can rewrite this now (i.e. with modern Kotlin) as:
```kotlin
parkAndCustomSerialize {
    // etc...
}
```

chrisr3 (commented):
There's a bit of history here. (And also a brand new Quasar 0.8.10_r3-SNAPSHOT...)

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberImpl.kt
sprintengine: You're right, good spot thanks!

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberImpl.kt
sprintengine: Yes that's much nicer, thanks!

jujoramos (approved):
I'm okay with the changes, just wish there was a quick way of adding a unit test or something to make sure we get some red flags if the serialiser gets accidentally changed in the future.

JamesHR3 (approved):

PR #3469, opened 2023-03-24
CORE-8712 - Switch to use MemberInfo with latest serial when processing initiator hello

dimosr:
Switch to use MemberInfo with latest serial when processing initiator hello.

At the point we receive the session initiator hello, we don't know the identity on our side the session is established with. This is only done at the later phases of the session handshake. Previously, we were picking the Member Info as viewed from a randomly selected locally hosted identity (the first one) up until that point. The new logic here picks the Member Info with the latest serial, so that we make sure we use the latest view. This will be a bit more tolerant against cases where peers have upgraded their Member Info and some locally hosted identities are on stale versions.


nikinagy:
Can we add some description why this change was needed? It's not that clear to me what was the issue with the previous approach.

dimosr:
sure, let me update the description

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerTest.kt
```diff
@@ -560,6 +561,34 @@ class SessionManagerTest {
         val responseMessage = sessionManager.processSessionMessage(LinkInMessage(initiatorHelloMsg))
 
         assertThat(responseMessage!!.payload).isInstanceOf(ResponderHelloMessage::class.java)
+        assertThat(responseMessage.header.address)
+            .isEqualTo(PEER_ENDPOINT)
+    }
+
+    @Test
+    fun `when an initiator hello is received, a responder hello is return`() {
```
williamvigorr3: Did you forget to update the test name?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -688,20 +688,16 @@ internal class SessionManagerImpl(
...
-                        MembershipStatusFilter.ACTIVE_IF_PRESENT_OR_PENDING,
-                    )
-                if (member == null) {
-                    null
-                } else {
-                    hostedIdentityInSameGroup to member
-                }
-            } ?: let {
+        val hostedIdentityInSameGroup = hostedIdentitiesInSameGroup.first()
+        val peerMemberInfo = hostedIdentitiesInSameGroup.mapNotNull { localIdentity ->
```
williamvigorr3: In theory `peerMemberInfo` might not be lookup-able inside the membership group reader for `hostedIdentityInSameGroup`. Would it be simpler conceptually to use something like (I haven't tested this code):
```
val (hostedIdentityInSameGroup, peerMemberInfo) = hostedIdentitiesInSameGroup.mapNotNull { localIdentity ->
             localIdentity to membershipGroupReaderProvider.lookupByKey(
                 localIdentity,
                 message.source.initiatorPublicKeyHash.array(),
                 MembershipStatusFilter.ACTIVE_IF_PRESENT_OR_PENDING
             )
         }.maxByOrNull { it.second.serial }
```
What you have already is functionally Okay, I think (having thought about it for a bit).

williamvigorr3 (commented):
This makes sense to me. e.g. using the latest version will help route messages if a member updates its gateway endpoint.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
dimosr: hm, not sure I understood what you mean by:

> In theory peerMemberInfo might not be lookup-able inside the membership group reader for hostedIdentityInSameGroup

This piece of the code performs a lookup with all the locally hosted identities and filters out all the successful lookups.

`hostedIdentityInSameGroup` is only used later on for the group policy lookup, not any membership lookup here. I guess with your suggestion we could ensure we use the same local identity for the group lookup as well, but given we use it for the pkiMode that's set for all the lifetime of a network, it wouldn't make a big difference. Also, I was hoping we would tackle https://r3-cev.atlassian.net/browse/CORE-8714 before GA, which would eliminate the need for doing this for group policy lookup completely.

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerTest.kt
dimosr: yes, I did - sorry, I'll update

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
williamvigorr3: What I mean is (if you did):
```
val member = membershipGroupReaderProvider.lookupByKey(
                 hostedIdentityInSameGroup,
                 message.source.initiatorPublicKeyHash.array(),
                 MembershipStatusFilter.ACTIVE_IF_PRESENT_OR_PENDING
             )
```
 `member` might be different to `peerMemberInfo` (or null even). So `peerMemberInfo` might not be in `hostedIdentityInSameGroup` view of the group (and this might make debugging more difficult and is a bit more complicated to think about).
 
`hostedIdentityInSameGroup` also sets the source identity in the message, from which the gateway does a lookup of the truststore (but that shouldn't be a problem).

williamvigorr3 (approved):
LGTM

PR #3476, opened 2023-03-27
CORE-11615 Need a way to create composite keys in flows

kyriathar:
- makes `CompositeKeyGeneratorImpl` injectable to flows

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/CompositeKeyGeneratorImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/CompositeKeyGeneratorImpl.kt
@@ -1,6 +1,7 @@
 package net.corda.flow.application.crypto
 
 import net.corda.crypto.core.CompositeKeyProvider
+import net.corda.sandbox.type.UsedByFlow
 import net.corda.v5.application.crypto.CompositeKeyGenerator
 import net.corda.v5.base.annotations.Suspendable
 import net.corda.v5.crypto.CompositeKeyNodeAndWeight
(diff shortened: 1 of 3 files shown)


components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/CompositeKeyGeneratorImpl.kt
```diff
@@ -15,13 +16,13 @@ import java.security.PublicKey
  * corda implementation so that we have flexibility to add extra checks or other
  * work at this point.
  */
-@Component(service = [ CompositeKeyGenerator::class, SingletonSerializeAsToken::class ])
+@Component(service = [CompositeKeyGenerator::class, UsedByFlow::class])
```
chrisr3: Add `scope = PROTOTYPE` here.

Every sandbox will create its own `CompositeKeyGeneratorImpl` instance, but they will all share the platform's `CompositeKeyProvider` service.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/CompositeKeyGeneratorImpl.kt
```diff
@@ -15,13 +16,13 @@ import java.security.PublicKey
...
 class CompositeKeyGeneratorImpl
 @Activate
 constructor(
     @Reference(service = CompositeKeyProvider::class)
     val provider: CompositeKeyProvider
```
chrisr3: This can be `private`:
```kotlin
@Reference(service = CompositeKeyProvider::class)
private val provider: CompositeKeyProvider
```

chrisr3 (commented):
Almost there...

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/CompositeKeyGeneratorImpl.kt
kyriathar: Done

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/CompositeKeyGeneratorImpl.kt
kyriathar: Done

chrisr3 (approved):
LGTM

PR #3479, opened 2023-03-27
CORE-3819: Update Gradle API usage for configurations.

chrisr3:
`with` has been deprecated, and replaced by `using`.

Also prefer `configureEach` over `all`.


mcgovc (approved):
LGTM

PR #3480, opened 2023-03-27
CORE-12069 - change confusing use function on EntityManagerFactory

driessamyn:
Fix up the `use` extension method on `EntityManagerFactory` as its current usage is confusing as it doesn't close the `EntityManagerFactory`, only the `EntityManager` it creates.

This makes the usages slightly more verbose, but more explicit and will hopefully lead to fixing some failures to close connection pools in the integration tests.

### DIFF
--- a/components/membership/membership-persistence-service-impl/src/integrationTest/kotlin/net/corda/membership/impl/persistence/service/MembershipPersistenceTest.kt
+++ b/components/membership/membership-persistence-service-impl/src/integrationTest/kotlin/net/corda/membership/impl/persistence/service/MembershipPersistenceTest.kt
@@ -111,7 +111,7 @@ import org.slf4j.LoggerFactory
 import java.nio.ByteBuffer
 import java.security.KeyPairGenerator
 import java.time.Instant
-import java.util.UUID
+import java.util.*
 import java.util.UUID.randomUUID
 import javax.persistence.EntityManagerFactory
 
@@ -525,7 +525,7 @@ class MembershipPersistenceTest {
 
         assertThat(result).isInstanceOf(MembershipPersistenceResult.Success::class.java)
 
-        val persistedEntity = vnodeEmf.use {
+        val persistedEntity = vnodeEmf.createEntityManager().use {
             it.find(RegistrationRequestEntity::class.java, registrationId)
         }
         assertThat(persistedEntity).isNotNull
@@ -554,13 +554,13 @@ class MembershipPersistenceTest {
         val persisted2 = membershipPersistenceClientWrapper.persistGroupPolicy(viewOwningHoldingIdentity, groupPolicy2, 2)
         assertThat(persisted2).isInstanceOf(MembershipPersistenceResult.Success::class.java)
 
-        val persistedEntity = vnodeEmf.use {
+        val persistedEntity = vnodeEmf.createEntityManager().use {
             it.find(GroupPolicyEntity::class.java, 1L)
         }
         assertThat(cordaAvroDeserializer.deserialize(persistedEntity.properties)!!.toMap()).isEqualTo(
             groupPolicy1.entries.associate { it.key to it.value }
         )
-        val secondPersistedEntity = vnodeEmf.use {
+        val secondPersistedEntity = vnodeEmf.createEntityManager().use {
             it.find(GroupPolicyEntity::class.java, 2L)
         }
         assertThat(cordaAvroDeserializer.deserialize(secondPersistedEntity.properties)!!.toMap()).isEqualTo(
@@ -576,7 +576,7 @@ class MembershipPersistenceTest {
         val persisted = membershipPersistenceClientWrapper.persistGroupParametersInitialSnapshot(viewOwningHoldingIdentity)
         assertThat(persisted).isInstanceOf(MembershipPersistenceResult.Success::class.java)
 
-        val persistedEntity = vnodeEmf.use {
+        val persistedEntity = vnodeEmf.createEntityManager().use {
             it.find(
                 GroupParametersEntity::class.java,
                 1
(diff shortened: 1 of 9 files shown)


components/membership/membership-persistence-service-impl/src/integrationTest/kotlin/net/corda/membership/impl/persistence/service/MembershipPersistenceTest.kt
```diff
@@ -111,7 +111,7 @@ import org.slf4j.LoggerFactory
 import java.nio.ByteBuffer
 import java.security.KeyPairGenerator
 import java.time.Instant
-import java.util.UUID
+import java.util.*
```
vkolomeyko: This wildcard import is probably not necessary. I suggest reverting this change.
I note it has been done in few other files.

vkolomeyko (commented):
LGTM

Minor comment posted.

components/membership/membership-persistence-service-impl/src/integrationTest/kotlin/net/corda/membership/impl/persistence/service/MembershipPersistenceTest.kt
dickon: There's some config in IntelliJ which does this by default, it's annoying and detekt complains.

libs/db/db-orm-impl/src/integrationTest/kotlin/net/corda/orm/impl/EntityManagerUtilsTest.kt
```diff
@@ -153,45 +154,47 @@ class EntityManagerUtilsTest {
     @Test
     fun `can load JPA entities using EntityManager#use`() {
 
-        val emf = createEntityManagerFactory()
+        createEntityManagerFactory().use { emf ->
```
dickon, line 162: So with this your change it is now appears that this test leaks an EMF. That's true in the baseline as well.

libs/permissions/permission-datamodel/src/integrationTest/kotlin/net/corda/permissions/model/test/RbacEntitiesTest.kt
```diff
@@ -97,7 +97,7 @@ class RbacEntitiesTest {
             null
         )
         emf.transaction { em -> em.persist(user) }
-        emf.use { em ->
+        emf.createEntityManager().use { em ->
```
dickon, line 100: we could explicitly check that the EMF got closed?

components/membership/membership-persistence-service-impl/src/integrationTest/kotlin/net/corda/membership/impl/persistence/service/MembershipPersistenceTest.kt
driessamyn: Sorry, yes, changed over laptops and haven't set the Intellij preference yet. Will change it.

libs/db/db-orm-impl/src/integrationTest/kotlin/net/corda/orm/impl/EntityManagerUtilsTest.kt
driessamyn: You're right. They must have been leaking beforehand too. Let me review.

dickon (dismissed):
This removes the confusion so is a step forward.

The new `use` could do with more coverage, and it seems we do have EMF connection leaks in various tests. 

I'd be inclined to pull in `consume` from the crypto processor, but that doesn't have to be done now. We should make tickets to clear up all EMFs in places that leak them.

libs/permissions/permission-datamodel/src/integrationTest/kotlin/net/corda/permissions/model/test/RbacEntitiesTest.kt
driessamyn: That one gets closed in the `@AfterAll` function.

I don't think it's in scope of this test to validate the behaviour of the utils.

driessamyn:
> I'd be inclined to pull in `consume` from the crypto processor, but that doesn't have to be done now. 

I don't agree. It doesn't add much and hides away what is created and closed, which I think leads to some confusion.
It's also rare you want to create an EMF, create a EM and close both in a single statement. We should discourage that pattern especially as we should be looking at caching the EMFs more effectively.
But maybe that's a discussion for another day.

> We should make tickets to clear up all EMFs in places that leak them.

I think we should fix them when we see them. It'd be more work to create the tickets than fix them as the biggest amount of work would be to find them.

vkolomeyko (dismissed):
LGTM

libs/db/db-orm/src/main/kotlin/net/corda/orm/utils/EntityManagerUtils.kt
```diff
@@ -15,8 +15,12 @@ import javax.persistence.EntityManagerFactory
...
  * @see transaction
  */
-inline fun <R> EntityManagerFactory.use(block: (EntityManager) -> R): R {
-    return createEntityManager().use(block)
+inline fun <R> EntityManagerFactory.use(block: (EntityManagerFactory) -> R): R {
+    return try {
+        block(this)
+    } finally {
+        close()
+    }
```
kyriathar, line 23: Should we just remove this function now? Does it add much? If yes then its kDoc needs updating I think.

driessamyn: You're right, the KDOC needed updating. That's done now. Thanks for spotting this!

I think the function is useful because `EntityManagerFactory` does not implement `AutoCloseable`, you cannot use the native `use` extension function.

charlieR3 (approved):
Looks good from a corda-platform-network-team point of view

kyriathar (approved):

vkolomeyko (approved):
LGTM

PR #3482, opened 2023-03-27
CORE-9703: Remove action from registration request

yift-r3:
API pull request in https://github.com/corda/corda-api/pull/1000 (!)

This will remove the unused action from the membership registration request.

e2e tests passed in https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda5-e2e-tests/detail/PR-49/3/pipeline


components/membership/membership-client/src/main/kotlin/net/corda/membership/client/MemberResourceClient.kt
```diff
@@ -14,11 +13,15 @@ interface MemberResourceClient : Lifecycle {
     /**
      * Starts the registration process for a member.
      *
-     * @param memberRegistrationRequest Data necessary to include in order to initiate registration.
+     * @param holdingIdentityShortHash The ID of the holding identity the member is using.
+     * @param registrationContext The member or MGM context required for on-boarding within a group.
```
nikinagy: I am not sure about this docs, because via registration members could only submit the 'member provided context'. I guess you meant here that MGMs need to provide their registration contexts too, in this case it just needs a bit rephrasing.

nikinagy (commented):
LGTM in general. One minor comment. Also, don't forget we should update the wiki page and let QA know about this change.

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/types/request/MemberRegistrationRequest.kt
```diff
@@ -1,12 +0,0 @@
-package net.corda.membership.rest.v1.types.request
-
-/**
- * Request sent during member registration.
- *
- * @param action The action to take during registration.
- * @param context The member or MGM context required for on-boarding within a group.
- */
-data class MemberRegistrationRequest(
-    val action: String,
```
vkolomeyko, line 10: I would suggest keeping `MemberRegistrationRequest` as before, removing `action` from it.

IMO it is better to have a stronger type for the specific operation rather than `Map<String, String>`, also it makes OpenAPI definition look neater.

components/membership/membership-client/src/main/kotlin/net/corda/membership/client/MemberResourceClient.kt
yift-r3: Do you have any better phrasing?

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/types/request/MemberRegistrationRequest.kt
yift-r3: OK

processors/rest-processor/src/integrationTest/resources/swaggerBaseline.json
```diff
@@ -4605,7 +4602,7 @@
             "example" : "No example available for this type"
           }
         },
-        "description" : "The request sent during registration which contains the requested registration action (e.g. 'requestJoin') along with a context map containing data required to initiate the registration process."
+        "description" : "The request sent during registration which contains the requested registration  context map containing data required to initiate the registration process."
```
charlieR3: Very nit picky
```suggestion
        "description" : "The request sent during registration which contains the requested registration context map containing data required to initiate the registration process."
```

charlieR3 (dismissed):
LGTM. One very very nit picky comment that I won't block on

components/membership/membership-client/src/main/kotlin/net/corda/membership/client/MemberResourceClient.kt
nikinagy: Maybe copy from the other docs: 
`The request sent during registration which contains the requested registration context map containing data required to initiate the registration process.`
Or if we want to stick with the original version: 
`The member's or MGM's registration context required for on-boarding within a group, which will be later part of the member provided context of the [MemberInfo]. In case of MGM registration some parts of the context will be used for building the [GroupPolicy] as well.`

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/MemberRegistrationRestResource.kt
```diff
@@ -60,10 +60,10 @@ interface MemberRegistrationRestResource : RestResource {
         @RestPathParameter(description = "The holding identity ID of the requesting virtual node")
         holdingIdentityShortHash: String,
         @ClientRequestBodyParameter(
-            description = "The request sent during registration which contains the requested registration action" +
-                    " (e.g. 'requestJoin') along with a context map containing data required to initiate the registration process."
+            description = "The request sent during registration which contains the requested registration " +
+                " context map containing data required to initiate the registration process."
```
vkolomeyko:
```suggestion
                "context map containing data required to initiate the registration process."
```

processors/rest-processor/src/integrationTest/resources/swaggerBaseline.json
vkolomeyko: This change needs to happen in the annotation of the underlying interface. I have added a comment to that effect.

vkolomeyko (dismissed):

nikinagy (dismissed):
LGTM

vkolomeyko (approved):
LGTM

nikinagy (approved):
LGTM

PR #3483, opened 2023-03-27
CORE-12098: Remove superfluous "immediate=true" from referenced components.

chrisr3:
By default, an OSGi component is declared as `immediate = false`, which means that the OSGi framework will not instantiate it until something - most likely another component - asks for an instance. We would prefer this to be when our OSGi bootstrapper requests the `Application` service instance, as all bundles will have been safely activated by this time.

Conversely, if an OSGi component is declared as `immediate = true` then the OSGi framework will try to instantiate it as soon as all of its (known) requirements have been satisfied. This could be _before_ the rest of the application's bundles have been activated, if the framework believes it can activate a component without them.

The Quasar instrumentor injects dynamic references to the `co.paralleluniverse.fibers.suspend.SuspendExecution` class into the classes that it instruments, along with a `DynamicImport-Package` reference to the `co.paralleluniverse.fibers.suspend` package. However, this requires that the `quasar-core-osgi` bundle is already active to _export_ this package, and there's no way to guarantee that. And not least because `quasar-core-osgi` has a few bundle dependencies of its own which need to be installed first too.

Reset as many components as possible back to using `immediate = false` to prevent the `combined-worker` from failing to start because it cannot find the `SuspendExecution` class.

### DIFF
--- a/applications/tools/p2p-test/app-simulator/src/main/kotlin/net/corda/p2p/app/simulator/AppSimulator.kt
+++ b/applications/tools/p2p-test/app-simulator/src/main/kotlin/net/corda/p2p/app/simulator/AppSimulator.kt
@@ -42,7 +42,7 @@ import java.time.Duration
 import java.time.Instant
 import kotlin.random.Random
 
-@Component(immediate = true)
+@Component
 class AppSimulator @Activate constructor(
     @Reference(service = Shutdown::class)
     private val shutDownService: Shutdown,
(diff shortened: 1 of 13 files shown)


driessamyn (approved):
LGTM

owenstanford (approved):
LGTM

dimosr (approved):
LGTM

PR #3487, opened 2023-03-28
CORE-12083 - Replaced the constructor call VirtualNodeInfoAvro with the builder.

filipesoliveira:
Replaced the constructor call `VirtualNodeInfoAvro` with the builder. This change is required to prevent breaking changes when a new field with a default value is added to the `VirtualNodeInfo` schema.


filipesoliveira:
This PR is required so https://github.com/corda/corda-api/pull/1002 can be merged without breaking the runtime-os.

owenstanford (approved):

PR #3488, opened 2023-03-28
CORE-10691 - avro schema registry from kafka

driessamyn:
We need to make sure that all workers have a schema of any message that can be posted in a cluster.  This is to ensure that, as messages evolve, that they can still be deserialized.  Avro expects to have both the original and expected schema for deserialization.  So when a message evolves we need to be able to have both sides available.

There should be an enhancement such that the schema registry can receive all of the schemas and add any new ones for which is doesn’t already know about from initialization.

This change makes all processors fetch existing schemas from the schemas (compacted) topic and publishes any new schemas to that topic.

Manually E2E tested:

image

### DIFF
--- a/components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/AvroSchemaProcessor.kt
+++ b/components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/AvroSchemaProcessor.kt
@@ -1,0 +1,87 @@
+package net.corda.configuration.read.impl
+
+import net.corda.data.Fingerprint
+import net.corda.lifecycle.LifecycleCoordinator
+import net.corda.messaging.api.processor.CompactedProcessor
+import net.corda.messaging.api.publisher.Publisher
+import net.corda.messaging.api.records.Record
+import net.corda.schema.Schemas.AvroSchemas.Companion.AVRO_SCHEMA_TOPIC
+import net.corda.schema.registry.AvroSchemaRegistry
+import org.apache.avro.Schema
+
+import org.slf4j.LoggerFactory
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.TimeUnit
+
+internal class AvroSchemaProcessor(
+    private val coordinator: LifecycleCoordinator,
+    private val avroSchemaRegistry: AvroSchemaRegistry
+) : CompactedProcessor<Fingerprint, String> {
+
+    private companion object {
+        private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+        const val PUBLISH_TIMEOUT = 5L
+    }
+
+    override val keyClass: Class<Fingerprint>
+        get() = Fingerprint::class.java
+    override val valueClass: Class<String>
+        get() = String::class.java
+
+    private val recordsToWrite = mutableListOf<Record<Fingerprint, String>>()
+    private var readSchemas = false
+
+    override fun onSnapshot(currentData: Map<Fingerprint, String>) {
+        logger.info("Received ${currentData.size} Avro schemas from topic in initial snapshot.")
+
+        val externalFingerprints = mutableSetOf<Fingerprint>()
+        // Read external schemas
+        currentData.forEach { (fingerprint, schemaJson) ->
+            externalFingerprints.add(fingerprint)
+            avroSchemaRegistry.addSchemaOnly(Schema.Parser().parse(schemaJson))
+        }
+
+        // Record any schemas we know about that aren't already in the topic
+        avroSchemaRegistry.schemasByFingerprintSnapshot.forEach { (fingerprint, schema) ->
+            if (!externalFingerprints.contains(fingerprint)) {
+                recordsToWrite.add(Record(AVRO_SCHEMA_TOPIC, fingerprint, schema.toString()))
+            }
+        }
+
+        if (recordsToWrite.isNotEmpty()) {
+            logger.info("Queued ${recordsToWrite.size} Avro schemas to write to topic.")
+        }
+
+        readSchemas = true
+        coordinator.postEvent(SetupConfigSubscription())
+    }
+
+    override fun onNext(newRecord: Record<Fingerprint, String>, oldValue: String?, currentData: Map<Fingerprint, String>) {
+        avroSchemaRegistry.addSchemaOnly(Schema.Parser().parse(newRecord.value))
+    }
+
+    fun publishNewSchemas(publisher: Publisher) {
+        if (!readSchemas) {
+            logger.info("Attempt to publish new schemas made prior to reading.")
+            return
+        }
+
+        if (recordsToWrite.isNotEmpty()) {
+            logger.info("Publishing ${recordsToWrite.size} queued Avro schemas to topic.")
+
+            val futures = publisher.publish(recordsToWrite)
+            try {
+                CompletableFuture.allOf(*futures.toTypedArray()).get(PUBLISH_TIMEOUT, TimeUnit.SECONDS)
+            } catch (e: Exception) {
+                // The consequence of this is some current version schemas may not be available to later versions of the
+                // software. All workers will attempt to publish missing schemas so there is every change this would
+                // repair itself. If it didn't future versions of software might reject some evolved avro messages as it
+                // not know what they are.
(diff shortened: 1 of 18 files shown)


components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/AvroSchemaProcessor.kt
```diff
@@ -0,0 +1,87 @@
...
+import java.util.concurrent.TimeUnit
+
+internal class AvroSchemaProcessor(
+    private val coordinator: LifecycleCoordinator,
+    private val avroSchemaRegistry: AvroSchemaRegistry
+) : CompactedProcessor<Fingerprint, String> {
+
+    private companion object {
+        private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+        const val PUBLISH_TIMEOUT = 5L
```
simon-johnson-r3: I usually add _SECONDS on to these to you don't have to go and check and it's obvious if there's a mistake at the point of use, must have forgotten here

components/configuration/configuration-read-service-impl/src/test/kotlin/net/corda/configuration/read/impl/AvroSchemaProcessorTest.kt
```diff
@@ -0,0 +1,111 @@
...
+
+     @Test
+    fun `publishNewSchemas publishes schemas added during onSnapshot`() {
+         val processor = AvroSchemaProcessor(coordinator, avroSchemaRegistry)
+
+         processor.onSnapshot(mapOf(catSchemaFingerPrint to catSchemaJson))
+
+         processor.publishNewSchemas(publisher)
+
+         `when`(publisher.publish(any())).doAnswer {
```
simon-johnson-r3: Mockito's Kotlin support adds a `whenever` function to get round this being a keyword.

components/configuration/configuration-read-service-impl/src/test/kotlin/net/corda/configuration/read/impl/AvroSchemaProcessorTest.kt
```diff
@@ -0,0 +1,111 @@
...
+             assertThat(records.size).isEqualTo(1)
+             // record is cat
+             val record = records.single()
+             assertSoftly { a ->
+                 a.assertThat(record.key).isEqualTo(catSchemaFingerPrint)
+                 a.assertThat(record.value).isEqualTo(catSchemaJson)
+             }
+             null
+         }
+    }
```
simon-johnson-r3, line 112: can you add `verify(publisher).publish(any())` here?
otherwise if publish is never called the test would still pass

libs/schema-registry/schema-registry-impl/src/test/kotlin/net/corda/schema/registry/impl/AvroSchemaRegistryImplTest.kt
```diff
@@ -216,6 +221,65 @@ internal class AvroSchemaRegistryImplTest {
...
+
+        registry.initialiseSchemas(avroGeneratedMessages)
+
+        val schemasCount = registry.schemasByFingerprintSnapshot.size
+        assertThat(registry.schemasByFingerprintSnapshot).doesNotContainKey(newSchemaFingerprint)
+
+        registry.addSchemaOnly(newSchema)
+
+        SoftAssertions.assertSoftly {
+            assertThat(registry.schemasByFingerprintSnapshot).size().isEqualTo(schemasCount+1)
```
simon-johnson-r3: minor, but should we have spaces around operators?

libs/schema-registry/schema-registry-impl/src/test/kotlin/net/corda/schema/registry/impl/AvroSchemaRegistryImplTest.kt
```diff
@@ -216,6 +221,65 @@ internal class AvroSchemaRegistryImplTest {
...
+    fun `containsSchema returns true if schema exists`() {
+        val registry = AvroSchemaRegistryImpl()
+
+        registry.schemasByFingerprintSnapshot.keys.forEach {
+            assertThat(registry.containsSchema(it)).isTrue
+        }
+    }
+
+    @Test
+    fun `containsSchema returbs false if schema does not exists`() {
```
simon-johnson-r3: typos
returbs -> returns
exists -> exist

libs/schema-registry/schema-registry-impl/src/test/kotlin/net/corda/schema/registry/impl/AvroSchemaRegistryImplTest.kt
```diff
@@ -216,6 +221,65 @@ internal class AvroSchemaRegistryImplTest {
...
+               "namespace" : "foo",
+               "name" : "bar",
+               "fields" : [
+                  { "name" : "Name" , "type" : "string" },
+                  { "name" : "CoolFactor" , "type" : "int" }
+               ]
+            }
+        """.trimIndent())
+        val newSchemaFingerprint = Fingerprint(SchemaNormalization.parsingFingerprint("SHA-256", newSchema))
+
```
simon-johnson-r3, line 278: minor: double CR

libs/schema-registry/schema-registry/src/main/kotlin/net/corda/schema/registry/AvroSchemaRegistry.kt
```diff
@@ -85,6 +86,16 @@ interface AvroSchemaRegistry {
      * @return the class type of the object encoded in [bytes].
      */
     fun getClassType(bytes: ByteBuffer) : Class<*>
+
+    /**
+     * Returns true if the Registry has a schema with the given [Fingerprint].
```
simon-johnson-r3: this would be more consistent as
`@return true if the Registry has a schema with the given [Fingerprint].`

libs/schema-registry/schema-registry/src/main/kotlin/net/corda/schema/registry/AvroSchemaRegistry.kt
```diff
@@ -85,6 +86,16 @@ interface AvroSchemaRegistry {
...
+     */
+    fun containsSchema(fingerPrint: Fingerprint): Boolean
+
+    /**
+     * Current, point in time, view of all [Schema]s in the registry, keyed by [Fingerprint].
```
simon-johnson-r3: @return

Maybe `Current, point in time,` should be `Snapshot of all...` or similar. The reason I say this is I thought I knew what you meant, but I had to go and double check the implementation to be sure.

simon-johnson-r3 (changes requested):

components/configuration/configuration-read-service-impl/src/test/kotlin/net/corda/configuration/read/impl/AvroSchemaProcessorTest.kt
driessamyn: Thanks. Didn't know that, usually use the `do {` construct, and copied this from another example.

Changed now.

components/configuration/configuration-read-service-impl/src/test/kotlin/net/corda/configuration/read/impl/AvroSchemaProcessorTest.kt
driessamyn: Good catch. Thanks!

libs/schema-registry/schema-registry/src/main/kotlin/net/corda/schema/registry/AvroSchemaRegistry.kt
driessamyn: I was trying to capture that it is a copy of the map that is returned. But you're right, Snapshot is probably a better description. Changed now.

simon-johnson-r3 (dismissed):

fowlerrr (dismissed):

fowlerrr (dismissed):

fowlerrr (approved):

simon-johnson-r3 (approved):

mcgovc (approved):
LGTM - build changes

yift-r3 (approved):
LGTM (Gateway + Link manager)

PR #3491, opened 2023-03-28
CORE-12034 Make `DigitalSignature` an interface on public API

kyriathar:
- adds `DigitalSignature.WithKeyId` implementation and updates use sites
- adds tests asserting AMQP serializability for types `DigitalSignatureWithKey` and `DigitalSignatureWithKeyId`

api PR: [#1011](https://github.com/corda/corda-api/pull/1011)

### DIFF
--- a/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/DigitalSignatureWithKey.kt
+++ b/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/DigitalSignatureWithKey.kt
@@ -1,9 +1,10 @@
 package net.corda.crypto.core
 
+import net.corda.v5.base.types.OpaqueBytes
 import net.corda.v5.crypto.DigitalSignature
 import java.security.PublicKey
 
 class DigitalSignatureWithKey(
     val by: PublicKey,
     bytes: ByteArray
-) : DigitalSignature(bytes)
+) : DigitalSignature, OpaqueBytes(bytes)
--- a/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/DigitalSignatureWithKeyId.kt
+++ b/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/DigitalSignatureWithKeyId.kt
@@ -8,6 +8,5 @@ class DigitalSignatureWithKeyId(
     private val by: SecureHash,
     bytes: ByteArray
     ) : DigitalSignature.WithKeyId, OpaqueBytes(bytes) {
-    // TODO Not sure why Kotlin generated getter cannot override?
     override fun getBy() = by
 }
--- a/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/DigitalSignatureWithKeyId.kt
+++ b/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/DigitalSignatureWithKeyId.kt
@@ -1,0 +1,13 @@
+package net.corda.crypto.core
+
+import net.corda.v5.base.types.OpaqueBytes
+import net.corda.v5.crypto.DigitalSignature
+import net.corda.v5.crypto.SecureHash
+
+class DigitalSignatureWithKeyId(
+    private val by: SecureHash,
+    bytes: ByteArray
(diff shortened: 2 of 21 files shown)


libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/DigitalSignatureWithKey.kt
```diff
@@ -1,9 +1,10 @@
 package net.corda.crypto.core
 
+import net.corda.v5.base.types.OpaqueBytes
 import net.corda.v5.crypto.DigitalSignature
 import java.security.PublicKey
 
 class DigitalSignatureWithKey(
```
mnesbit, line 7: Needs equality and hash

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/DigitalSignatureWithKeyId.kt
```diff
@@ -0,0 +1,13 @@
+package net.corda.crypto.core
+
+import net.corda.v5.base.types.OpaqueBytes
+import net.corda.v5.crypto.DigitalSignature
+import net.corda.v5.crypto.SecureHash
+
+class DigitalSignatureWithKeyId(
```
mnesbit, line 7: Needs equality and hash

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/DigitalSignatureWithKeyId.kt
```diff
@@ -0,0 +1,13 @@
...
+    private val by: SecureHash,
+    bytes: ByteArray
+    ) : DigitalSignature.WithKeyId, OpaqueBytes(bytes) {
+    // TODO Not sure why Kotlin generated getter cannot override?
+    override fun getBy() = by
```
mnesbit, line 11: When doing Kotlin -> Java properties have to become explicit getters and setters.

mnesbit: I am worried that this class hides a lot of details in the innocent inheritance from OpaqueBytes. It isn't clear that this has value equality and hashcode and it isn't clear that measures are taken to make the getBytes() read-only/a copy

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/DigitalSignatureWithKey.kt
mnesbit: Does this concrete class have to be marked as @CordaSerializable to be AMQP serializable? Not sure we can inherit that from the interface?

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/DigitalSignatureWithKeyId.kt
mnesbit: We need to verify that this is AMQP serializable and doesn't require an AMQP custom serializer. I currently only see Kryo custom serializers

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/DigitalSignatureWithKey.kt
kyriathar: Added tests showing they are AMQP serializable without `CordaSerializable` so seems they do inherit it from interface.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/DigitalSignatureWithKeyId.kt
kyriathar: > I am worried that this class hides a lot of details in the innocent inheritance from OpaqueBytes. It isn't clear that this has value equality and hashcode

Yes indeed, they are offered through `ByteSequence` for all byte wrapper types extending it.

driessamyn (dismissed):
LGTM

driessamyn (approved):
LGTM

mnesbit (approved):
OK

PR #3493, opened 2023-03-28
CORE-9705 - Modify start registration handler

nikinagy:
- Modify existing validation for duplicate member names (refer for design docs).
- Add new validation for requests: if the serial number on the request is smaller than the current version of the requestor's MemberInfo, then decline the request.
- MGM should increase and define serial for the pending member info.


components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
```diff
@@ -103,17 +107,28 @@ internal class StartRegistrationHandler(
...
                 listOf(pendingMemberHoldingId)
             )
             validateRegistrationRequest(
                 existingMemberInfo is MembershipQueryResult.Success
-                        && (existingMemberInfo.payload.isEmpty()
-                        || !existingMemberInfo.payload.sortedBy { it.modifiedTime }.last().isActive)
-            ) { "The latest member info for given member is in 'Active' status or " +
-                    "there is a member with the same name." }
+            ) { "Looking up for existing member info (if there were any), failed due to a persistence exception." }
+            val queryResult = (existingMemberInfo as MembershipQueryResult.Success).payload
```
yift-r3: `getOrThrow` is shorter

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
```diff
@@ -103,17 +107,28 @@ internal class StartRegistrationHandler(
...
+            // The MemberX500Name is not a duplicate
+            validateRegistrationRequest(
+                queryResult.isEmpty() || registrationRequest.serial != 0L
+            ) { "Member already exists with the same X500 name." }
+            // Serial number on the request should be smaller than the current version of the requestor's MemberInfo
+            val activeOrSuspendedInfo = queryResult.find {
```
yift-r3: Maybe `lastOrNull` instead of `find`?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
```diff
@@ -103,17 +107,28 @@ internal class StartRegistrationHandler(
...
+                it.status == MEMBER_STATUS_ACTIVE || it.status == MEMBER_STATUS_SUSPENDED
+            }
+            validateRegistrationRequest(
+                activeOrSuspendedInfo == null || activeOrSuspendedInfo.serial <= registrationRequest.serial!!
```
yift-r3, line 124: We can save the serial in line 96 and avoid the `!!`

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
```diff
@@ -93,6 +93,10 @@ internal class StartRegistrationHandler(
                 }
             }
 
+            validateRegistrationRequest(registrationRequest.serial != null) {
+                "Serial on the registration request should not be null."
```
yift-r3, line 98: Do we have a unit test for this?

nikinagy: Yes it is called `declined if serial in the registration request is null` in the test class, starting from line 249

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
nikinagy: That would work if we would use require() or an if() to check the nullability, but since we use our validateRegistrationRequest() function for validations, kotlin cannot smart-cases and it still needs the '!!' :(

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
nikinagy: addressed in [a4f13c0](https://github.com/corda/corda-runtime-os/pull/3493/commits/a4f13c07026b5048faa9d33310b929bf217e4f34)

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
nikinagy: addressed in [a4f13c0](https://github.com/corda/corda-runtime-os/pull/3493/commits/a4f13c07026b5048faa9d33310b929bf217e4f34)

yift-r3 (approved):
LGTM

PR #3496, opened 2023-03-28
CORE-11255: Fail a flow when the current platform version is different than the initial one.

fowlerrr:
In order to avoid any issues with underlying changes during a platform upgrade we'll fail any outstanding flows, for now, when we detect a platform version change.  These flow can then be run again on the new platform version

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
@@ -12,6 +12,7 @@ import net.corda.flow.fiber.FlowContinuation
 import net.corda.flow.fiber.FlowLogicAndArgs
 import net.corda.flow.fiber.factory.FlowFiberFactory
 import net.corda.flow.pipeline.events.FlowEventContext
+import net.corda.flow.pipeline.exceptions.FlowFatalException
 import net.corda.flow.pipeline.factory.FlowFactory
 import net.corda.flow.pipeline.factory.FlowFiberExecutionContextFactory
 import net.corda.flow.pipeline.runner.FlowRunner
@@ -46,6 +47,12 @@ class FlowRunnerImpl @Activate constructor(
         context: FlowEventContext<Any>,
         flowContinuation: FlowContinuation
     ): FiberFuture {
+        if (context.checkpoint.initialPlatformVersion != platformInfoProvider.localWorkerPlatformVersion) {
+            throw FlowFatalException("Platform has been modified from version " +
+                    "${context.checkpoint.initialPlatformVersion} to version " +
+                    "${platformInfoProvider.localWorkerPlatformVersion}.  The flow must be restarted.")
+        }
+
         return when (val receivedEvent = context.inputEvent.payload) {
             is StartFlow -> startFlow(context, receivedEvent)
             is SessionEvent -> {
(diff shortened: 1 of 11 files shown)


simon-johnson-r3 (dismissed):

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
```diff
@@ -27,12 +29,20 @@ class FlowRunnerImpl @Activate constructor(
...
+    private val platformInfoProvider: PlatformInfoProvider,
 ) : FlowRunner {
     override fun runFlow(
         context: FlowEventContext<Any>,
         flowContinuation: FlowContinuation
     ): FiberFuture {
+        if (context.checkpoint.initialPlatformVersion != platformInfoProvider.localWorkerPlatformVersion) {
+            throw FlowFatalException("Platform has been modified from version " +
+                    "${context.checkpoint.initialPlatformVersion} to version " +
+                    "${platformInfoProvider.localWorkerPlatformVersion}.  The flow must be restarted.")
```
lankydan, line 53: Might need a more of an explanation of what to do in this scenario.

Also theres a double space 🤮

fowlerrr: _Always_ double space after a full stop!

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/sessions/CheckpointInitializerImplTest.kt
```diff
@@ -54,16 +56,17 @@ class CheckpointInitializerImplTest {
         checkpointInitializer.initialize(checkpoint, waitingFor, holdingIdentity.toCorda()) {
             startContext
         }
-        verify(checkpoint).initFlowState(any(), any())
+        verify(checkpoint).initFlowState(any(), any(), any())
```
JamesHR3: I think this test might need adjusting - it should really assert that `cpkFileHashes` is equal to some test set, and you should probably assert that the platform version is set correctly on the checkpoint object.

components/flow/flow-service/src/test/kotlin/net/corda/flow/state/FlowCheckpointImplTest.kt
```diff
@@ -110,6 +110,7 @@ class FlowCheckpointImplTest {
             flowId = "F1"
             flowState = newFlowState
             pipelineState = newPipelineState
+            initialPlatformVersion = 50000
         }
```
emilybowe, line 114: I would add a unit test here to check you can get `initialPlatformVersion` out of the checkpoint once it's set.

fowlerrr: Fair enough

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/sessions/CheckpointInitializerImplTest.kt
fowlerrr: Yes, but I ended up going away from that approach, so haven't touched the test now... 😬

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
fowlerrr: BTW, what more should we say other than "restart the flow"?

simon-johnson-r3 (dismissed):

driessamyn (commented):
LGTM (when API is merged)

simon-johnson-r3 (dismissed):

mcgovc (dismissed):
LGTM build file changes

PR #3497, opened 2023-03-28
CORE-11976 - Use session context properties for interop session identification

ac101m:
# Changes
- Add `isInteropSession` context property to to flow context properties map.
- Add `isInteropSession` field to flow mapper state avro schema.
- Remove hack on session ID and use the new field for interop session identification.

# Info
- Corresponding corda-api PR: https://github.com/corda/corda-api/pull/1013


libs/flows/flow-mapper-impl/src/integrationTest/kotlin/net/corda/flow/mapper/integration/FlowMapperIntegrationTest.kt
```diff
@@ -215,7 +215,7 @@ class FlowMapperIntegrationTest {
         val sessionEvent =
             buildSessionEvent(MessageDirection.INBOUND, inputKey, 3, SessionData(ByteBuffer.wrap("".toByteArray())))
         val flowMapperEvent = FlowMapperEvent(sessionEvent)
-        val flowMapperState = FlowMapperState("flowKey", null, FlowMapperStateType.OPEN)
+        val flowMapperState = FlowMapperState("flowKey", null, FlowMapperStateType.OPEN, false)
```
yarivus, line 218: why do we explicitly specify the `false` parameter if all functions have a default value of false ? 
This would immensely reduce the amount of changes in this PR

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionInitExecutor.kt
```diff
@@ -71,19 +78,19 @@ class SessionInitExecutor(
...
                         emptyList(),
-                        SessionConfirm(KeyValuePairList(emptyList()))
+                        SessionConfirm(KeyValuePairList(listOf(KeyValuePair(Constants.FLOW_SESSION_IS_INTEROP, "true"))))
                     )
                 )
             )
 
             return FlowMapperResult(
-                FlowMapperState(flowKey, null, FlowMapperStateType.OPEN),
+                FlowMapperState(flowKey, null, FlowMapperStateType.OPEN, true),
```
yarivus, line 87: this would be the only change in this PR for when it's true - see my prev comment

yarivus (changes requested):

libs/flows/flow-mapper-impl/src/integrationTest/kotlin/net/corda/flow/mapper/integration/FlowMapperIntegrationTest.kt
ac101m: I've used defaults where I can, but `FlowMapperState` is an avro type. We get this all-parameters constructor and that's it! It might be possible to use the generated builder to do this, but this would mean many more changed lines of code.

jmacmahonr3 (approved):
Looks alright to me

SaurabhR3 (approved):

PR #3505, opened 2023-03-29
CORE-9790 Vault support in crypto plugin

simon-johnson-r3:
Added support for Vault in the crypto initial config plugin. This involves removing the wrapping key and wrapping salt from the Config entirely. I believe these are used to create a root key for `EncryptionSecretsServiceImpl` only and are not required for Vault.

@dickon please let me know if I got this wrong at all.


simon-johnson-r3:
Ok, this is not the correct thing to be doing, it results in the following error in the db worker
```
net.corda.libs.configuration.validation.ConfigurationValidationException: Configuration failed to validate for key corda.cryptoLibrary at schema version null: $.hsmMap.SOFT.hsm.cfg.wrappingKeyMap.salt: is missing but it is required, $.hsmMap.SOFT.hsm.cfg.wrappingKeyMap.passphrase: is missing but it is required
```

PR #3506, opened 2023-03-29
CORE-12143 Internalize `OpaqueBytes` and `ByteSequence`

kyriathar:
- adds `OpaqueBytes`, `ByteSequence` and `OpaqueBytesSubSequence` ported from public API
- changes `ByteSequence`:: `abstract val bytes: ByteArray` to `abstract fun getBytes(): ByteArray` to make multiple inheritance from `ByteSequence` and Java types having `getBytes` method work.

api PR: [#1015](https://github.com/corda/corda-api/pull/1015)


dickon (approved):
LGTM

PR #3508, opened 2023-03-29
CORE-9085 - Fail MGM registration if trust root specified is not valid PEM certificate

dimosr:
### Changes
Changing the MGM registration logic, so that if any of the specified TLS trustroots is invalid then the registration fails. 

Given how easy it was, I added the same checks for session trust roots although it's out of scope of the associated bug.

### Testing
Tested by following the onboarding steps for an MGM up until registration where I used an invalid string for TLS trustroot. Registration was submitted successfully and checking the registration status returned the following:
```
{
  "registrationId": "836c3a0e-22aa-4d15-bac3-d00842c2d923",
  "registrationSent": "2023-03-29T14:16:26.049Z",
  "registrationUpdated": "2023-03-29T14:16:28.787Z",
  "registrationStatus": "INVALID",
  "memberInfoSubmitted": {
    "data": {
      "registrationProtocolVersion": "1",
      "corda.session.keys.0.id": "09D8C3F24172",
      "corda.ecdh.key.id": "C111168C006D",
      "corda.group.protocol.registration": "net.corda.membership.impl.registration.dynamic.member.DynamicMemberRegistrationService",
      "corda.group.protocol.synchronisation": "net.corda.membership.impl.synchronisation.MemberSynchronisationServiceImpl",
      "corda.group.protocol.p2p.mode": "Authenticated_Encryption",
      "corda.group.key.session.policy": "Combined",
      "corda.group.pki.session": "NoPKI",
      "corda.group.pki.tls": "Standard",
      "corda.group.tls.type": "OneWay",
      "corda.group.tls.version": "1.3",
      "corda.endpoints.0.connectionURL": "https://corda-p2p-gateway-worker.corda-cluster-a:8080",
      "corda.endpoints.0.protocolVersion": "1",
      "corda.group.trustroot.tls.0": "TLS CA Pem Certificate"
    }
  },
  "reason": "Onboarding MGM failed. Trust root certificate specified in registration context under key corda.group.trustroot.tls.0 was not a valid PEM certificate.",
  "serial": null
}
```

### DIFF
--- a/components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/mgm/MGMRegistrationContextValidator.kt
+++ b/components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/mgm/MGMRegistrationContextValidator.kt
@@ -82,6 +82,9 @@ internal class MGMRegistrationContextValidator(
             context.keys.filter { TRUSTSTORE_SESSION.format("[0-9]+").toRegex().matches(it) }.apply {
                 require(isNotEmpty()) { "No session trust store was provided." }
                 require(orderVerifier.isOrdered(this, 4)) { "Provided session trust stores are incorrectly numbered." }
+                this.map { key ->
+                    validateTrustrootCert(context[key]!!, key)
+                }
             }
         }
         context.keys.filter { TRUSTSTORE_TLS.format("[0-9]+").toRegex().matches(it) }.apply {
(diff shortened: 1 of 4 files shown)


components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/mgm/MGMRegistrationContextValidator.kt
```diff
@@ -101,6 +110,15 @@ internal class MGMRegistrationContextValidator(
             )
         }
     }
+
+    fun validateTrustrootCert(pemCert: String, key: String) {
+        try {
+            CertificateFactory.getInstance("X.509").generateCertificate(pemCert.byteInputStream())
```
yift-r3: 1. Do we want to verify that the certificate has not expired?
2. Should we create a single CertificateFactory instead of a new one for every verification?

dimosr: 1. Given it _sounds_ relatively simple, I think it's worth checking. Let me give it a try.
2. yes, good catch. I'll change that.

yift-r3 (approved):

PR #3510, opened 2023-03-29
CORE-11935: Use session preferred key

yift-r3:
Use the session preferred key (instead of the first one) to sign the membership package


components/membership/locally-hosted-identities-service-impl/src/main/kotlin/net/corda/membership/locally/hosted/identities/impl/PublicKeyFactory.kt
```diff
@@ -0,0 +1,22 @@
+package net.corda.membership.locally.hosted.identities.impl
+
+import net.corda.v5.base.exceptions.CordaRuntimeException
+import org.bouncycastle.asn1.x509.SubjectPublicKeyInfo
+import org.bouncycastle.openssl.PEMParser
+import org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter
+import java.security.PublicKey
+
+internal fun publicKeyFactory(pem: String): PublicKey {
```
dimosr: Seems like we have an identical method in `PublicKeyReader`. Should we pull this out into `:libs:utilities`?

components/membership/membership-p2p/src/main/kotlin/net/corda/membership/p2p/helpers/Verifier.kt
```diff
@@ -21,26 +21,43 @@ class Verifier(
...
+        signature: CryptoSignatureWithKey,
+        signatureSpecAvro: CryptoSignatureSpec,
+        data: ByteArray,
+    ) {
+        val key = keyEncodingService.decodePublicKey(signature.publicKey.array())
+        if (!acceptableKeys.contains(key)) {
+            throw IllegalArgumentException("The signature public key is not one of the acceptable keys.")
+        }
+        verify(
+            keyEncodingService.decodePublicKey(signature.publicKey.array()),
```
dimosr: shouldn't we reuse `key` here?

components/membership/synchronisation-impl/src/integrationTest/kotlin/net/corda/membership/impl/synchronisation/dummy/TestLocallyHostedIdentitiesService.kt
```diff
@@ -0,0 +1,60 @@
...
+import net.corda.virtualnode.HoldingIdentity
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.Reference
+import org.osgi.service.component.propertytypes.ServiceRanking
+import java.security.PublicKey
+import java.util.concurrent.ConcurrentHashMap
+
+interface TestLocallyHostedIdentitiesService : LocallyHostedIdentitiesService {
+    fun setPreferredSessionKey(id: HoldingIdentity, key: PublicKey)
```
dimosr: I think it would be a bit cleaner if we have the following method instead of `setPreferredSessionKey`:
```
fun setIdentityInfo(identity: HoldingIdentity, IdentityInfo)
```

dimosr (commented):
Looks good - only a few minor comments

components/membership/locally-hosted-identities-service-impl/src/main/kotlin/net/corda/membership/locally/hosted/identities/impl/LocallyHostedIdentitiesServiceImpl.kt
```diff
@@ -171,7 +173,9 @@ class LocallyHostedIdentitiesServiceImpl(
         return IdentityInfo(
             identity = this.holdingIdentity.toCorda(),
             tlsCertificates = this.tlsCertificates.toCertificates(),
-            preferredSessionKey = publicKeyFactory(this.preferredSessionKeyAndCert.sessionPublicKey),
+            preferredSessionKey = publicKeyFactory(
+                this.preferredSessionKeyAndCert.sessionPublicKey.reader(),
+            ) ?: throw CordaRuntimeException("Invalid session public key PEM"),
```
dimosr: What is the impact of this exception escaping here, is this going to halt the subscription from processing?

Given this is at the point we read (so "invalid" data must have already been written, which I suppose should be impossible given we perform validation when the vnode is onboarded), should we just log a warning here, ignore that record and continue?

libs/utilities/src/main/kotlin/net/corda/utilities/PublicKeyFactory.kt
dimosr, line 1: I'd suggest creating a `crypto` sub-package and putting this there.

libs/utilities/src/main/kotlin/net/corda/utilities/crypto/PublicKeyFactory.kt
dimosr, line 1: Judging from the build status, I think some OSGi metadata are missing for this package now.

components/membership/locally-hosted-identities-service-impl/src/main/kotlin/net/corda/membership/locally/hosted/identities/impl/LocallyHostedIdentitiesServiceImpl.kt
```diff
@@ -162,21 +162,23 @@ class LocallyHostedIdentitiesServiceImpl(
...
-            identity = this.holdingIdentity.toCorda(),
-            tlsCertificates = this.tlsCertificates.toCertificates(),
-            preferredSessionKey = publicKeyFactory(
-                this.preferredSessionKeyAndCert.sessionPublicKey.reader(),
-            ) ?: throw CordaRuntimeException("Invalid session public key PEM"),
-        )
+    private fun HostedIdentityEntry.toIdentityInfo(): IdentityInfo? {
+        return publicKeyFactory(
+            this.preferredSessionKeyAndCert.sessionPublicKey.reader(),
+        )?.let { preferredSessionKey ->
```
dimosr: Let's also add a warning log here in the unlikely case where we have an invalid key, e.g.
```
logger.warn("Hosted Identity entry for ${this.holdingIdentity.toCorda()} had invalid preferred session key (${this.preferredSessionKeyAndCert.sessionPublicKey}). Ignoring this record.")
```

dimosr (approved):

ronanbrowne (approved):
LGTM - build

PR #3511, opened 2023-03-29
CORE-9740 Change protocol name of non-validating notary plugin

relyafi:
### Summary

This PR changes the protocol name of the non-validating notary plugin to better reflect the package structure. Most importantly, changing from `net.corda` to `com.r3.corda` makes it clear this is part of CorDapp, not "platform" code.

Version is now also set explicitly in both the client and server plugins rather than relying on the default of 1. This is to make it clear to anyone reading or copying the code that versioning is supported and needs to be considered.

### Testing

Ensure that all automated tests continue to pass.


nkovacsx (approved):

filipesoliveira (approved):
LGTM

charlieR3 (approved):

PR #3515, opened 2023-03-30
CORE-12174: Upgrade Log4J 2.19.0 -> 2.20.0.

chrisr3:
This should also avoid a few ugly errors when running with the `SecurityManager` enabled.


mnesbit (approved):
OK

PR #3516, opened 2023-03-30
CORE-3819: Prevent Gradle trying to download new JDKs automatically.

chrisr3:
Later versions of Gradle require `org.gradle.java.installations.auto-download` to be set explicitly.


ronanbrowne (dismissed):
lgtm

ronanbrowne (approved):

PR #3521, opened 2023-03-30
CORE-11879 Add Enterprise Dependencies

simon-johnson-r3:
Corda 5 Enterprise gets runtime dependencies from `runtime-os` transitively. We cannot create runtime dependencies against Enterprise because it creates OSGi errors where there are duplicates.

The problem is that some workers already have runtime dependencies Enterprise addons needs and some don't. In this case the Vault addon uses Jackson and Caffeine, but not all workers depend on this already.

Before this change, the Enterprise deployment looked like this:
Screenshot 2023-03-30 at 16 28 43

After, it looks like this:
Screenshot 2023-03-30 at 19 14 12


driessamyn (approved):
LGTM

wzur-r3 (approved):
LGTM

PR #3522, opened 2023-03-30
CORE-10407: Resolve warnings related to Java modules.

chrisr3:
Open selected Java modules to prevent warnings about "illegal access". This makes us more "Java 17 friendly" without actually migrating to Java 17.

Also fix some general Gradle warnings about deprecated APIs, and assign correct "Automatic Module Name" to our new `kryo-serializers` artifact.

---

There is a curious issue in the OSGi tests for `:components:membership:registration-impl`, which I don't have time to diagnose. Up to now, the `Resolve` task has not been checking that every service the tests ask for actually exists within the framework, and there have indeed been a few holes. However, if I now _provide_ the missing `GroupParametersWriterService` component then suddenly the tests fail for no apparent reason?!

I have worked around this for now by telling the resolver not to worry about the `GroupParametersWriterService` because the framework will provide it. _This is a fib,_ but strangely the integration tests don't seem to care... :exploding_head:.

### DIFF
--- a/components/membership/registration-impl/build.gradle
+++ b/components/membership/registration-impl/build.gradle
@@ -51,9 +51,10 @@ dependencies {
     integrationTestImplementation project(':testing:db-message-bus-testkit')
 
     integrationTestRuntimeOnly project(':components:configuration:configuration-read-service-impl')
-    integrationTestRuntimeOnly project(':components:crypto:crypto-client-impl')
+    integrationTestRuntimeOnly project(':components:crypto:crypto-client-hsm-impl')
     integrationTestRuntimeOnly project(':components:db:db-connection-manager-impl')
     integrationTestRuntimeOnly project(':components:membership:membership-group-read-impl')
+    integrationTestRuntimeOnly project(':components:membership:membership-persistence-client-impl')
     integrationTestRuntimeOnly project(':libs:lifecycle:lifecycle-impl')
     integrationTestRuntimeOnly project(':libs:messaging:db-message-bus-impl')
     integrationTestRuntimeOnly project(':libs:messaging:messaging-impl')
--- a/applications/workers/workers-smoketest/build.gradle
+++ b/applications/workers/workers-smoketest/build.gradle
@@ -110,6 +110,8 @@ tasks.register('smokeTest', Test) {
     description = "Runs smoke tests."
     group = "verification"
 
+    dependsOn smokeTestResources
+
     testClassesDirs = project.sourceSets["smokeTest"].output.classesDirs
     classpath = project.sourceSets["smokeTest"].runtimeClasspath
 
@@ -129,9 +131,6 @@ tasks.register('smokeTest', Test) {
 
     systemProperty "combinedWorkerHealthHttp",
             project.getProperties().getOrDefault("combinedWorkerHealthHttp", combinedWorker ? "http://localhost:7004/" : null)
-}
 
-
-tasks.named('smokeTest', Test) {
-    dependsOn smokeTestResources
+    jvmArgs '--add-opens', 'java.base/java.lang.reflect=ALL-UNNAMED'
 }
--- a/build.gradle
+++ b/build.gradle
@@ -89,6 +89,7 @@ allprojects {
 
         doFirst {
             systemProperty 'java.io.tmpdir', buildDir.absolutePath
+            jvmArgs '--illegal-access=deny'
         }
     }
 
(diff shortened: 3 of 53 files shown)


vkolomeyko (approved):

mcgovc (approved):
LGTM - build changes

fowlerrr (approved):
LGTM

driessamyn:
Please don't merge this until the api breaking changes that are needed for today's deadline.

chrisr3:
> Please don't merge this until the api breaking changes that are needed for today's deadline.

This isn't about APIs, it's just Gradle and Bnd.

driessamyn:
> > Please don't merge this until the api breaking changes that are needed for today's deadline.
> 
> This isn't about APIs, it's just Gradle and Bnd.

I know, but those merges have priority. We don't want conflicts. See comms on the channel.

chrisr3:
> I know, but those merges have priority. We don't want conflicts. See comms on the channel.

What kind of 11:59:59.9999 hour kind of change could be conflicting with this PR?! Is someone trying to remove OSGi before Beta3?

components/membership/registration-impl/build.gradle
```diff
@@ -51,9 +51,10 @@ dependencies {
     integrationTestImplementation project(':testing:db-message-bus-testkit')
 
     integrationTestRuntimeOnly project(':components:configuration:configuration-read-service-impl')
-    integrationTestRuntimeOnly project(':components:crypto:crypto-client-impl')
+    integrationTestRuntimeOnly project(':components:crypto:crypto-client-hsm-impl')
     integrationTestRuntimeOnly project(':components:db:db-connection-manager-impl')
     integrationTestRuntimeOnly project(':components:membership:membership-group-read-impl')
+    integrationTestRuntimeOnly project(':components:membership:membership-persistence-client-impl')
```
yift-r3, line 57: We have the `components/membership/registration-impl/src/integrationTest/kotlin/net/corda/membership/impl/registration/dummy/TestMembershipPersistenceClientImpl.kt` to avoid adding this dependency.

fowlerrr (changes requested):
Just holding this until the "Big Merge" is done

driessamyn:
> > I know, but those merges have priority. We don't want conflicts. See comms on the channel.
> 
> What kind of 11:59:59.9999 hour kind of change could be conflicting with this PR?! Is someone trying to remove OSGi before Beta3?

It's a 53 file change. Not worth the risk. There is no reason for this to go in today, unlike the other 3 PRs that are lined up to go in shortly.

components/membership/registration-impl/build.gradle
chrisr3: > We have the `components/membership/registration-impl/src/integrationTest/kotlin/net/corda/membership/impl/registration/dummy/TestMembershipPersistenceClientImpl.kt` to avoid adding this dependency.

Unfortunately it's currently required to provide `MembershipQueryClientImpl`, and no alternative is available. This module clearly has issues, but I don't know what the "correct" fix is. The best I've been able to do is enable "service resolving" to verify that everything is present.

yift-r3: We can create a dummy `MembershipQueryClient` in the same way we do a dummy for the `MembershipPersistenceClient`. Having the actual client implementation without the service will cause a lot of pain...

fowlerrr (approved):
Unblocking.  Still good.

components/membership/registration-impl/build.gradle
chrisr3: You (as in "the @corda/corda-platform-network-team") can resolve this issue however you see fit, _in a subsequent PR_. This _particular_ PR concerns "resolving warnings related to Java modules", and it has only high-lighted an _existing_ issue with these OSGi tests at the very least. Ultimately, you should be able to remove these lines from your `test.bndrun` file:
```
-runsystemcapabilities:\
    osgi.service;objectClass:List<String>='net.corda.membership.groupparams.writer.service.GroupParametersWriterService';effective:=active
```
because they are falsely claiming that a `GroupParametersWriterService` exists at runtime, just to satisfy the `Resolve` task. This kind of nonsense should only be required if the `GroupParametersWriterService` is being registered dynamically, which seems very unlikely.

yift-r3: https://r3-cev.atlassian.net/browse/CORE-12212 created. Any other things we need to do?

yift-r3 (approved):
Ticket for additional work created

PR #3527, opened 2023-03-31
CORE-9790 Crypto Initial Config for Vault supported by plugin

simon-johnson-r3:
The initial config plugin can now be executed with parameters like the following
`corda-cli initial-config create-crypto-config -t VAULT -v cryptosecrets -ks salt -kp passphrase`
to produce a Config with secrets referencing Vault rather than using the internal encryption secrets service:
```
insert into config (config, is_deleted, schema_version_major, schema_version_minor, section, update_actor, update_ts, version) values ('{"busProcessors":{"flow":{"maxAttempts":3,"waitBetweenMills":[200]},"ops":{"maxAttempts":3,"waitBetweenMills":[200]},"registration":{"maxAttempts":3,"waitBetweenMills":[200]}},"cryptoConnectionFactory":{"expireAfterAccessMins":5,"maximumSize":3},"hsmMap":{"SOFT":{"hsm":{"capacity":-1,"categories":[{"category":"*","policy":"WRAPPED"}],"cfg":{"keyMap":{"cache":{"expireAfterAccessMins":60,"maximumSize":1000},"name":"CACHING"},"wrapping":{"name":"DEFAULT"},"wrappingKeyMap":{"cache":{"expireAfterAccessMins":60,"maximumSize":1000},"name":"CACHING","passphrase":{"configSecret":{"vaultKey":"passphrase","vaultPath":"cryptosecrets"}},"salt":{"configSecret":{"vaultKey":"salt","vaultPath":"cryptosecrets"}}}},"masterKeyPolicy":"UNIQUE","name":"SOFT","supportedSchemes":["CORDA.RSA","CORDA.ECDSA.SECP256R1","CORDA.ECDSA.SECP256K1","CORDA.EDDSA.ED25519","CORDA.X25519","CORDA.SM2","CORDA.GOST3410.GOST3411","CORDA.SPHINCS-256"]},"retry":{"attemptTimeoutMills":20000,"maxAttempts":3},"workerTopicSuffix":""}},"hsmService":{"downstreamMaxAttempts":3},"signingService":{"cache":{"expireAfterAccessMins":60,"maximumSize":10000}}}', false, 1, 0, 'corda.cryptoLibrary', 'init', '2023-03-31T10:04:01.120394Z', 0)
```

This has been tested against Corda 5 Enterprise.


tools/plugins/initial-config/src/main/kotlin/net/corda/cli/plugin/initialconfig/CryptoConfigSubcommand.kt
```diff
@@ -23,70 +25,71 @@ import java.util.Base64
...
+        description = ["Passphrase for the encrypting secrets service. Mandatory for CORDA type secrets service."]
     )
-    lateinit var passphrase: String
+    var passphrase: String? = null
 
     @CommandLine.Option(
         names = ["-ws", "--wrapping-salt"],
-        required = false,
-        description = ["Salt for the SOFT HSM root wrapping key."]
+        description = ["Salt for the SOFT HSM root wrapping key. Used only by CORDA type secrets service."]
```
dickon: No, the wrapping salt is an input to the crypto service, and is currently used in all cases and is not relevant to the Vault support.

tools/plugins/initial-config/src/main/kotlin/net/corda/cli/plugin/initialconfig/CryptoConfigSubcommand.kt
```diff
@@ -23,70 +25,71 @@ import java.util.Base64
...
     )
     var softHsmRootSalt: String? = null
 
     @CommandLine.Option(
         names = ["-wp", "--wrapping-passphrase"],
-        required = false,
-        description = ["Passphrase for the SOFT HSM root wrapping key."]
+        description = ["Passphrase for the SOFT HSM root wrapping key. Used only by CORDA type secrets service."]
```
dickon: Again, the wrapping passphrase is an input to the crypto service, and is currently used in all cases and is not relevant to the Vault support.

dickon (dismissed):
Since @simon-johnson-r3  is out I've taken on finishing tis PR. I rebased this work to the new release/os/5.0, and went back to keymap passpharse and vault not being optional. I've also arranged to throw the expect expection type when we fail validation of the command line options so that Picoli validates it.

tools/plugins/initial-config/src/main/kotlin/net/corda/cli/plugin/initialconfig/CryptoConfigSubcommand.kt
dickon: fixed

tools/plugins/initial-config/src/main/kotlin/net/corda/cli/plugin/initialconfig/CryptoConfigSubcommand.kt
dickon: fixed

josephzunigadaly (dismissed):

tools/plugins/initial-config/src/main/kotlin/net/corda/cli/plugin/initialconfig/CryptoConfigSubcommand.kt
```diff
@@ -114,8 +122,61 @@ class CryptoConfigSubcommand : Runnable {
...
+                { "'vaultPath' must be set for VAULT type secrets." })
+        }
+
+        val (passphraseSecretValue, saltSecretValue) = generateSecretValuesForType()
+
+        val wrappingPassphraseSecret =
+            secretsService.createValue(passphraseSecretValue, "corda-master-wrapping-key-passphrase").root()
+        val wrappingSaltSecret =
+            secretsService.createValue(saltSecretValue, "corda-master-wrapping-key-salt").root()
+        return Pair(wrappingPassphraseSecret, wrappingSaltSecret)
```
kyriathar: My understanding is, if it is vault configuration then `passphraseSecretValue` and `saltSecretValue` are the keys to the vault's secrets for passphrase and salt, but if not vault configuration then `passphraseSecretValue` and `saltSecretValue` are the actual passphrase and salt. Should we be mapping those things to same config keys like we do now? And if we do, do we have some check to make sure right ones are present based on the implementation injected at runtime? I.e. vault or not vault

tools/plugins/initial-config/src/main/kotlin/net/corda/cli/plugin/initialconfig/CryptoConfigSubcommand.kt
```diff
@@ -114,8 +122,61 @@ class CryptoConfigSubcommand : Runnable {
...
+                checkParamPassed(vaultPath)
+                { "'vaultPath' must be set for VAULT type secrets." })
+        }
+
+        val (passphraseSecretValue, saltSecretValue) = generateSecretValuesForType()
+
+        val wrappingPassphraseSecret =
+            secretsService.createValue(passphraseSecretValue, "corda-master-wrapping-key-passphrase").root()
+        val wrappingSaltSecret =
+            secretsService.createValue(saltSecretValue, "corda-master-wrapping-key-salt").root()
```
kyriathar: Should these names be present in the example outcome configuration mentioned on description? I think they are not mentioned there.

tools/plugins/initial-config/src/test/kotlin/net/corda/cli/plugin/initialconfig/TestInitialConfigPluginCrypto.kt
```diff
@@ -105,6 +115,36 @@ class TestInitialConfigPluginCrypto {
...
+        val json = outText.substring(expectedPrefix.length until (outJsonEnd + 3))
+        assertGeneratedJson(json) { it: Config, _: SmartConfigFactory ->
+            val key1 = it.getObjectList("wrappingKeys")[0]
+            assertThat(key1.getValue("salt").render()).doesNotContain("encryptedSecret")
+            assertThat(key1.getValue("passphrase").render()).doesNotContain("encryptedSecret")
+            assertThat(key1.getValue("salt").render()).contains("vaultKey")
+            assertThat(key1.getValue("salt").render()).contains("vaultPath")
+            assertThat(key1.getValue("passphrase").render()).contains("vaultKey")
+            assertThat(key1.getValue("passphrase").render()).contains("vaultPath")
+        }
```
kyriathar, line 145: I think worth validating configuration outcome against crypto schema.

tools/plugins/initial-config/src/main/kotlin/net/corda/cli/plugin/initialconfig/CryptoConfigSubcommand.kt
dickon: Oh you are right, this code is a bit messy. It does work, but only since it passes in the key in the plainText argument, and the key is ignored. I'm testing a rework of this.

dickon: Rework pushed.

tools/plugins/initial-config/src/main/kotlin/net/corda/cli/plugin/initialconfig/CryptoConfigSubcommand.kt
dickon: Rework pushed.

kyriathar (approved):
Looks OK as a first step. I think we should somehow check emitted crypto configuration validates against crypto schema somehow so we can catch that early. And also maybe we should distinguish somehow `passphrase` and `salt` placeholders in configuration between `VAULT` and `CORDA`?

PR #3528, opened 2023-03-31
CORE-11481 Added logs for flow start and flow fail

Cap-j:
Logs now show when flow fails and starts.


libs/lifecycle/lifecycle-impl/src/main/kotlin/net/corda/lifecycle/impl/LifecycleProcessor.kt
```diff
@@ -180,7 +180,7 @@ internal class LifecycleProcessor(
                 dependentComponents?.stopAll()
             }
             runUserEventHandler(event, coordinator)
-            closeManagedResources(emptySet())
```
pmck91: It looks like these files got pulled in from your last ticket! you can merge in an up to date runtime-os or rebase to it!

pmck91 (approved):
LGTM

PR #3529, opened 2023-03-31
CORE-12190 - Update default gateway configuration for timeouts

dimosr:
### Changes
For context on the change, see description on https://github.com/corda/corda-api/pull/1030.
Currently, defaults exist in corda-runtime-os so updating them in the same way as in corda-api.

There's a [separate ticket](https://r3-cev.atlassian.net/browse/CORE-9946) to remove default from corda-runtime-os.


williamvigorr3 (approved):
LGTM

PR #3533, opened 2023-03-31
CORE-12027 Additional input and reference state checking

relyafi:
### Summary

This change re-instates some additional platform checks that were performed as part of transaction verification in Corda 4. The following checks have been re-instated:

- Specifying the same input or reference state multiple times in the same transaction is now an error
- Similarly, specifying the same state as both an input and reference state in the same transaction is now an error

These checks have also been added to the uniqueness checker logic (in addition to transaction verification) as a defensive measure, in case this service is used outside of the UTXO ledger in future.

The check for the presence of a notary that is performed in Corda 4 has not been re-instated; a notary is now required on every UTXO ledger transaction, which was not the case in Corda 4. This is therefore enforced by the API which requires a notary to be set.

### Testing

This change breaks several automated test cases, as we are changing the behavior of both transaction verification and uniqueness checking. Tests have been updated accordingly and additional test cases added to address gaps in the existing tests.

### DIFF
--- a/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
+++ b/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
@@ -978,7 +978,7 @@ class FlowTests {
     }
 
     @Test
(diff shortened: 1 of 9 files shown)


blsemo (approved):
LGTM

vlajos (approved):
LGTM

PR #3534, opened 2023-04-03
CORE-12211 - fix windows build

driessamyn:
Reduce timestamp precision to ensure tests works with Windows Postgres driver.


driessamyn:
Green windows build: https://ci02.dev.r3.com/job/Corda5/job/Nightlys/job/Corda-Runtime-OS/job/Windows/job/PR-3534/3/

dickon (approved):
Bit unfortunate that we need to do this but okay, LGTM.

PR #3536, opened 2023-04-03
CORE-11759: Add support for Corda Prometheus PodMonitor

oscardalmau-r3:
We currently have an option exposed on the Corda Helm chart that adds Prometheus annotations to the workers pods. Promtail does not utilise these and, instead, the expectation is that Pod/ServiceMonitors are defined to identify endpoints to scrape. 

This PR adds support to create a PodMonitor. Unlike the annotations, this is disabled by default (as it would fail if the CRD is not defined in the environment).


charts/corda/values.yaml
```diff
@@ -39,6 +39,13 @@ metrics:
   # -- enable scraping of worker metrics through Prometheus annotations
   scrape: true
 
+# Pod monitor configuration for scraping metrics. Please note that the corresponding CRD must exist in the environment
+podMonitor:
```
davidcurrie: This should be under `metrics`

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -513,6 +513,31 @@ data:
...
+{{/*
+Pod Monitor creation
+*/}}
+{{- define "corda.podMonitor" -}}
+{{- if .Values.podMonitor.enabled }}
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: corda
```
davidcurrie: the name should take into account the Helm release

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -513,6 +513,31 @@ data:
...
+  {{- range $k, $v := .Values.podMonitor.labels }}
+    {{ $k }}: {{ $v | quote }}
+  {{- end }}
+spec:
+  podMetricsEndpoints:
+  - port: monitor
+  jobLabel: corda
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: corda
```
davidcurrie: this should use the `corda.name` function as we do when generating the label

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -513,6 +513,31 @@ data:
...
+metadata:
+  name: corda
+  labels:
+  {{- range $k, $v := .Values.podMonitor.labels }}
+    {{ $k }}: {{ $v | quote }}
+  {{- end }}
+spec:
+  podMetricsEndpoints:
+  - port: monitor
+  jobLabel: corda
```
davidcurrie: can't remember how this is used but should it also use the release name?

charts/corda/values.schema.json
```diff
@@ -188,6 +188,35 @@
...
+                    "default": false,
+                    "title": "enable pod monitor creation to identify endpoints to scrape",
+                    "examples": [
+                        false
+                    ]
+                },
+                "labels": {
+                    "type": "object",
+                    "default": {},
+                    "title": "labels that can be used so PodMonitor is discovered by Prometheus",
```
davidcurrie: be consistent about "pod monitor" or "PodMonitor"

charts/corda/values.schema.json
```diff
@@ -188,6 +188,35 @@
...
+                "enabled": {
+                    "type": "boolean",
+                    "default": false,
+                    "title": "enable pod monitor creation to identify endpoints to scrape",
+                    "examples": [
+                        false
+                    ]
+                },
+                "labels": {
+                    "type": "object",
```
davidcurrie, line 186: can we be more precise about the properties of this object?

davidcurrie (changes requested):

charts/corda/values.yaml
oscardalmau-r3: Done

charts/corda-lib/templates/_helpers.tpl
oscardalmau-r3: Done - sorry I forgot about this

charts/corda-lib/templates/_helpers.tpl
oscardalmau-r3: Done

charts/corda-lib/templates/_helpers.tpl
oscardalmau-r3: Done

charts/corda/values.schema.json
oscardalmau-r3: Done

charts/corda/values.schema.json
oscardalmau-r3: @davidcurrie I don't think we can. In the `non-functional-tests` we use `release` key, but Prometheus can be installed in different ways and `release` may not be an option always.

davidcurrie (approved):
👍

PR #3539, opened 2023-04-03
CORE-12070 - Adding a flow fiber cache (shared between virtual nodes)

conal-smith:
**Overview**
This flow fiber cache is a caffeine cache with key `FlowKey` (avro object, composed of `flowId`, `holding identity`) and hard reference values to flow fiber objects. 

**Flow Sandbox Cache Eviction**
Cache is shared between all virtual nodes, when sandboxes are evicted from the flow sandbox cache we also evict the flow fibers for this virtual node. Flow fibers have a hard reference to the sandbox (through `FlowLogicAndArgs`). The sandbox eviction listener provides a way to evict these flow fibers before the sandbox is put into the `expiryQueue`.

We null out `flowFiberExecutionContext` and `suspensionOutcome` on the flow fiber before caching, these are populated again in `FlowFiberImpl.resume`. Quasar itself populates some inherited fields in `unparkDeserialized` in the base `Fiber` class. 

**Invalidation on repartition event**
We invalidate the flow keys from the flow fiber cache for lost partitions. `FlowExecutorRebalanceListener.onPartitionLost` triggers removal of flow keys lost in a rebalance. Kafka API provides us with the partitions and states lost. This prevents flow resuming on a worker with an out-of-date cached flow fiber.

**Configuration**
For 5.1 this will be configurable via a flow cache configuration, for now configurable through system properties. Max size of 0 causes the cache to instantly remove entries when added. Defaults are max size of flow fibers 10000 and 600 seconds. 5.1 will also improve the size using a weight system as we already know the size of the serialized fiber ByteArray.
```
-Dnet.corda.flow.fiber.cache.maximumSize=10000
-Dnet.corda.flow.fiber.cache.expireAfterWriteSeconds=600
```

**Other**

**Weakly referenced fibers**
The values in the cache were configured as `weakValues()` but through testing we found the flow fibers were being garbage collected almost immediately with many cache misses. Removing `weakValues()` reinstated full cache hit rate. No longer using weak values for the cache.

**Manual testing**
Manually tested that after suspending, the flow restarts from the correct line of code after unparking.

**Testing**
On sandbox eviction, logs for flow fiber evicting holding identity can be seen [here](https://grafana.non-functional-tests.corda.cloud/explore?orgId=1&left=%7B%22datasource%22:%22loki%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki%22%7D,%22editorMode%22:%22builder%22,%22expr%22:%22%7Bnamespace%3D%5C%22performance-conal-perf-tests-20230508170702%5C%22%7D%20%7C%3D%20%60Flow%20fiber%20cache%20removing%20holdingIdentity%60%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221683565983144%22,%22to%22:%221683567320112%22%7D%7D). 

On repartition, logs for flow fiber evicting flow keys lost in repartition event can be seen [here](https://grafana.non-functional-tests.corda.cloud/explore?orgId=1&left=%7B%22datasource%22:%22loki%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki%22%7D,%22editorMode%22:%22builder%22,%22expr%22:%22%7Bnamespace%3D%5C%22resiliency-conal-perf-tests-20230508171037%5C%22%7D%20%7C%3D%20%60flow%20fibers%20from%20flow%20fiber%20cache%60%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221683567875661%22,%22to%22:%221683568926339%22%7D%7D).


components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/factory/FlowFiberFactoryImpl.kt
```diff
@@ -52,15 +64,32 @@ class FlowFiberFactoryImpl : FlowFiberFactory {
...
-                    flowFiberExecutionContext.flowCheckpoint.serializedFiber.array(),
-                    FlowFiberImpl::class.java
-                )
+                getFromCacheOrDeserialize(flowFiberExecutionContext)
             }!!
 
         return FiberFuture(fiber, fiber.resume(flowFiberExecutionContext, suspensionOutcome, currentScheduler))
     }
 
+    private fun getFromCacheOrDeserialize(flowFiberExecutionContext: FlowFiberExecutionContext): FlowFiberImpl {
```
sprintengine, line 72: Unfortunately this is not unit testable due to mockito's inability to mock FlowFiberImpl, and difficulties with OSGI when trying to test using a real FlowFiberImpl object:

```
co/paralleluniverse/fibers/suspend/SuspendExecution
java.lang.NoClassDefFoundError: co/paralleluniverse/fibers/suspend/SuspendExecution
	at net.corda.flow.fiber.factory.FlowFiberFactoryImplTest.<init>(FlowFiberFactoryImplTest.kt:56)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at org.junit.platform.commons.util.ReflectionUtils.newInstance(ReflectionUtils.java:552)
	at org.junit.jupiter.engine.execution.ConstructorInvocation.proceed(ConstructorInvocation.java:56)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.api.extension.InvocationInterceptor.interceptTestClassConstructor(InvocationInterceptor.java:73)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.lambda$invoke$0(InterceptingExecutableInvoker.java:93)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:106)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.proceed(InvocationInterceptorChain.java:64)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.chainAndInvoke(InvocationInterceptorChain.java:45)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.invoke(InvocationInterceptorChain.java:37)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.invoke(InterceptingExecutableInvoker.java:92)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.invoke(InterceptingExecutableInvoker.java:62)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.invokeTestClassConstructor(ClassBasedTestDescriptor.java:363)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.instantiateTestClass(ClassBasedTestDescriptor.java:310)
	at org.junit.jupiter.engine.descriptor.ClassTestDescriptor.instantiateTestClass(ClassTestDescriptor.java:79)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.instantiateAndPostProcessTestInstance(ClassBasedTestDescriptor.java:286)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$testInstancesProvider$4(ClassBasedTestDescriptor.java:278)
	at java.base/java.util.Optional.orElseGet(Optional.java:369)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$testInstancesProvider$5(ClassBasedTestDescriptor.java:277)
	at org.junit.jupiter.engine.execution.TestInstancesProvider.getTestInstances(TestInstancesProvider.java:31)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$prepare$0(TestMethodTestDescriptor.java:105)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.prepare(TestMethodTestDescriptor.java:104)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.prepare(TestMethodTestDescriptor.java:68)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$prepare$2(NodeTestTask.java:123)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.prepare(NodeTestTask.java:123)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:90)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:41)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$6(NodeTestTask.java:155)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:141)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$9(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:138)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:95)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:41)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$6(NodeTestTask.java:155)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:141)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$9(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:138)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:95)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.submit(SameThreadHierarchicalTestExecutorService.java:35)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutor.execute(HierarchicalTestExecutor.java:57)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestEngine.execute(HierarchicalTestEngine.java:54)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:107)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:88)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.lambda$execute$0(EngineExecutionOrchestrator.java:54)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.withInterceptedStreams(EngineExecutionOrchestrator.java:67)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:52)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:114)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:86)
	at org.junit.platform.launcher.core.DefaultLauncherSession$DelegatingLauncher.execute(DefaultLauncherSession.java:86)
	at org.junit.platform.launcher.core.SessionPerRequestLauncher.execute(SessionPerRequestLauncher.java:53)
	at org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestClassProcessor$CollectAllTestClassesExecutor.processAllTestClasses(JUnitPlatformTestClassProcessor.java:99)
	at org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestClassProcessor$CollectAllTestClassesExecutor.access$000(JUnitPlatformTestClassProcessor.java:79)
	at org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestClassProcessor.stop(JUnitPlatformTestClassProcessor.java:75)
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.stop(SuiteTestClassProcessor.java:62)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:33)
	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:94)
	at com.sun.proxy.$Proxy5.stop(Unknown Source)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker$3.run(TestWorker.java:193)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.executeAndMaintainThreadName(TestWorker.java:129)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:100)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:60)
	at org.gradle.process.internal.worker.child.ActionExecutionWorker.execute(ActionExecutionWorker.java:56)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:113)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:65)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.run(GradleWorkerMain.java:69)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.main(GradleWorkerMain.java:74)
Caused by: java.lang.ClassNotFoundException: co.paralleluniverse.fibers.suspend.SuspendExecution
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	... 86 more
```

applications/workers/release/flow-worker/build.gradle
```diff
@@ -12,6 +12,10 @@ osgiRun {
     )
 }
 
+quasar {
+    cacheLocations = [ 'FLOW/*' ]
+}
```
chrisr3: This configures Quasar to cache instrumented byte-code in memory. The point is that Quasar only needs to instrument a flow _once_, no matter how many sandboxes you create with it.

I don't see how this relates to your `FlowFiberCache`.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/factory/FlowFiberFactoryImpl.kt
chrisr3: The `SuspendExecution` class is found here:
```gradle
dependencies {
    testRuntimeOnly "co.paralleluniverse:quasar-core-osgi:$quasarVersion:framework-extension"
}
```
However, anything using Quasar should be tested inside an OSGi framework.

---

Hmm, I am now confused!? The `co.paralleluniverse.fibers.suspend` package is _clearly_ inside this framework extension:
```shell
$ jar tfv quasar-core-osgi-0.9.0_r3-SNAPSHOT-framework-extension.jar | grep suspend
     0 Wed Apr 05 16:57:40 BST 2023 co/paralleluniverse/fibers/suspend/
  1097 Wed Apr 05 16:57:40 BST 2023 co/paralleluniverse/fibers/suspend/Instrumented.class
  1174 Wed Apr 05 16:57:40 BST 2023 co/paralleluniverse/fibers/suspend/RuntimeSuspendExecution.class
  2113 Wed Apr 05 16:57:40 BST 2023 co/paralleluniverse/fibers/suspend/StackOps.class
  1007 Wed Apr 05 16:57:40 BST 2023 co/paralleluniverse/fibers/suspend/SuspendExecution.class
   220 Wed Apr 05 16:57:40 BST 2023 co/paralleluniverse/fibers/suspend/package-info.class
```
which _by definition_ of being a framework extension means that these classes _must_ all be available as soon as the framework starts! Which in turn means that there should not be any reason for the other `quasar-core-osgi` bundle to be installed before `SuspendExecution` becomes available for class-loading.

Oh dear, I need to think about this a bit more!

applications/workers/release/flow-worker/build.gradle
sprintengine: The intention was to identify any new memory leak introduced by this change. Since the main release branch has a memory leak upon sandbox eviction, and since your changes fix this memory leak, a test run with your changes will expose if the fiber caching has any memory leak. I added a change to null out the flowFiberExecutionContext which hopefully avoids lots of lingering objects. 

If it all looks good I'll undo your changes before merging.

chrisr3: > The intention was to identify any new memory leak introduced by this change.

I don't see the connection. Quasar's own caching has nothing to do with either the sandbox or OSGi.

sprintengine: I was under the impression [this branch](https://github.com/corda/corda-runtime-os/tree/chrisr3-very-unsecure) contained changes that fix the memory leak with sandbox cache eviction so I just copied them over. Is that the wrong branch?

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberCacheImpl.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    private val maximumSize = java.lang.Long.getLong("net.corda.flow.fiber.cache.maximumSize", 10)
+    private val expireAfterWriteSeconds = java.lang.Long.getLong("net.corda.flow.fiber.cache.expireAfterWriteSeconds", 60)
+
+    private val cache: Cache<FlowFiberCacheKey, Any> = CacheFactoryImpl().build(
+        "flow-fiber-cache",
+        Caffeine.newBuilder()
+            .maximumSize(maximumSize)
+            .weakValues()
+            .expireAfterWrite(Duration.ofSeconds(expireAfterWriteSeconds))
```
jujoramos: Isn't 60 seconds _**too short**_?. Maybe I'm reading this wrong, but if we don't manage to complete a flow execution from start to finish within 60 seconds, does that mean the entry will be automatically removed from the cache?.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/factory/FlowFiberFactoryImpl.kt
```diff
@@ -52,15 +64,32 @@ class FlowFiberFactoryImpl : FlowFiberFactory {
...
+            ) as FlowFiberImpl?
+        } catch (e: Exception) {
+            // shouldn't really be possible to get in here
+            logger.warn("Failure casting flow fiber from checkpoint for flow ${flowFiberExecutionContext.flowCheckpoint.flowId}. " +
+                    "Removing and skipping cache.")
+            flowFiberCache.remove(
+                FlowFiberCacheKey(flowFiberExecutionContext.flowCheckpoint.holdingIdentity, flowFiberExecutionContext.flowCheckpoint.flowId)
+            )
+            null
+        }
```
jujoramos: We're making too strong an assumption here: an exception (other than `ClassCastException`) might be thrown for whatever reason and the actual cause will be hard to troubleshoot due to the "misleading" log message and the fact that the original exception (or the stack trace) is not included in log the message itself.

applications/workers/release/flow-worker/build.gradle
chrisr3: The `chrisr3-very-unecure` branch _does_ contain the fixes for the memory leaks, but I added that _particular_ change separately just to see whether it improved performance.

sprintengine: Ah ok, thanks for clarifying, I'll pull out the quasar caching from this branch.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/factory/FlowFiberFactoryImpl.kt
sprintengine: Yeah that's true, the caffeine cache itself could throw, I'll make a separate catch block 👍

chrisr3:
> Due to issues with mockito and mocking quasar's Fiber class, the cache itself holds type Any, and we safely cast to FlowFiberImpl in the FlowFiberFactoryImpl.

:warning: **Please don't even _try_ mocking Quasar Fibers. Just use `TestOSGi` tasks to test with _actual_ Quasar.** :warning:.

It sounds like you've already compromised your implementation to accommodate Mockito, which is itself a Bad Sign. But the bigger issue is that _Quasar itself doesn't work unless it is allowed to instrument the byte-code!_ And since these "mocks" do no such thing, IMO that makes these tests meaningless.

---

As an aside, I was under the impression that our flows were not CPU-bound, and that any performance problems were due to missing / unaccounted for wall time. In which case, how is all this extra complexity intended to help?

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberCacheImpl.kt
sprintengine: It won't be an entire flow, just the next part of the flow after it unparks and it runs until the next suspension point.

I think part of investigation running with soak tests will be trying various sizes and expiry times and seeing the performance tradeoff with memory footprint. I think the cache will struggle when we fire lots of load at it, entries will probably get evicted before the flow can be resumed due to high number of initial checkpoints and other events being processed.

Maybe there's a way we could prioritise the cache for in-flight flows, i.e. don't evict in-flight fibers, or when the cache is full, we don't use the cache.

applications/workers/release/flow-worker/build.gradle
```diff
@@ -11,7 +11,6 @@ osgiRun {
         'org.osgi.framework.security': 'osgi'
     )
 }
-
```
chrisr3: Accidental change here?

conal-smith:
> It sounds like you've already compromised your implementation to accommodate Mockito

I've changed the cache to hold FlowFiberImpl instead of "Any" which was at the time of writing the comment.

I've ran several different types of tests on the aws cluster and identified a couple issues which I resolved. Some references can be seen in the description links.

I'm preparing to look into testOSGi tests and make sure all of this code has good test coverage. It would be good to get any feedback on these changes while I do that.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
```diff
@@ -0,0 +1,50 @@
...
+import net.corda.flow.fiber.FlowFiberImpl
+import net.corda.flow.fiber.cache.FlowFiberCache
+import net.corda.flow.fiber.cache.FlowFiberCacheKey
+import net.corda.virtualnode.HoldingIdentity
+
+@Suppress("unused")
+@Component(service = [FlowFiberCache::class])
+class FlowFiberCacheImpl @Activate constructor() : FlowFiberCache {
+
+    private val maximumSize = java.lang.Long.getLong("net.corda.flow.fiber.cache.maximumSize", 10000)
```
sprintengine: Not entirely sure about these configurations, probably should be made configurable via config read service, etc. And if size is 0 effectively disables the cache.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberImpl.kt
```diff
@@ -158,6 +158,13 @@ class FlowFiberImpl(
         }
     }
 
+    private fun Fiber<*>.prepareForCaching(): FlowFiberImpl {
```
jujoramos, line 165: Would be helpful to have some docs here explaining why this is needed.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
jujoramos: +1

libs/virtual-node/sandbox-group-context/src/main/kotlin/net/corda/sandboxgroupcontext/SandboxGroupContextPreRemovalCallback.kt
```diff
@@ -0,0 +1,19 @@
...
+    /**
+     * Performs a pre-removal action with the sandbox's holding identity.
+     *
+     * This function is called automatically before a [SandboxGroupContext] is removed.
+     * The `holdingIdentity` parameter provides information about the holding identity who's
+     * sandbox is being removed.
+     *
+     * @param holdingIdentity the holding identity of the sandbox being removed
+     */
+    fun preSandboxRemoval(holdingIdentity: HoldingIdentity)
```
jujoramos: Shouldn't the parameter be of type `VirtualNodeContext` instead as that's what `SandboxGroupContextCache` knows?.

sprintengine: Yeah that sounds better idea

components/flow/flow-service/build.gradle
```diff
@@ -42,6 +42,7 @@ dependencies {
 
     implementation platform("net.corda:corda-api:$cordaApiVersion")
 
+    implementation project(":libs:cache:cache-caffeine")
```
JamesHR3: Would you mind putting this with the other `project` dependencies above?

components/flow/flow-service/src/main/java/net/corda/flow/fiber/cache/package-info.java
```diff
@@ -0,0 +1,4 @@
+@Export
+package net.corda.flow.fiber.cache;
```
JamesHR3, line 2: Does this need exporting outside this bundle? I'm wondering if anything external ever sees this package?

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/FlowFiberCache.kt
```diff
@@ -0,0 +1,34 @@
...
+import net.corda.flow.fiber.FlowFiberImpl
+import net.corda.virtualnode.HoldingIdentity
+
+/**
+ * Cache for flow fibers.
+ */
+interface FlowFiberCache {
+    /**
+     * Put a flow fiber into the cache keyed by the given flowId.
+     */
```
JamesHR3: Might need to change the KDoc for this and other methods to indicate that the key is `FlowFiberCacheKey` (which I assume is a combination of flow ID and holding identity).

There might already be a `FlowKey` type that you can use that is exactly that? Worth double checking.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/FlowFiberCacheKey.kt
```diff
@@ -0,0 +1,11 @@
+package net.corda.flow.fiber.cache
+
+import net.corda.virtualnode.HoldingIdentity
+
+/**
+ * Flow fiber cache is keyed by the holding identity and flow ID.
+ */
+data class FlowFiberCacheKey(
```
JamesHR3: Worth double checking there isn't already a type that serves this purpose.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
```diff
@@ -0,0 +1,50 @@
...
+    private val expireAfterWriteSeconds = java.lang.Long.getLong("net.corda.flow.fiber.cache.expireAfterWriteSeconds", 600)
```
JamesHR3: We should at least make these string constants a `const val` in a companion object.

They should be made configurable though, yes. We can do that in a follow up for 5.1.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheEvictionServiceImpl.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.osgi.service.component.annotations.Reference
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+@Suppress("unused")
+@Component(service = [FlowFiberCacheEvictionService::class])
+class FlowFiberCacheEvictionServiceImpl @Activate constructor(
+    @Reference(service = FlowFiberCache::class)
+    private val flowFiberCache: FlowFiberCache
+) : FlowFiberCacheEvictionService {
```
JamesHR3: What do you need this service for? Doesn't the `FlowFiberCache` do the same thing anyway and have the same sandbox injection properties?

components/flow/flow-service/src/main/java/net/corda/flow/fiber/cache/package-info.java
sprintengine: Think left over from previous iteration, don't think I need it any more since all encapsulated within flow service

edit- think it is needed

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/FlowFiberCache.kt
sprintengine: FlowKey is a avro object so I don't want that one.

Yeah good spot I'll update these docs

Edit - I actually think I'll change my mind on that we can use `FlowKey` based on most usage coming from avro types anyway

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheEvictionServiceImpl.kt
sprintengine: Yeah a bit of a pointless abstraction, think I'll get rid of it

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/factory/FlowFiberFactoryImpl.kt
```diff
@@ -11,14 +11,25 @@ import net.corda.flow.fiber.FlowLogicAndArgs
...
 import net.corda.flow.pipeline.exceptions.FlowFatalException
 import net.corda.metrics.CordaMetrics
+import org.osgi.service.component.annotations.Activate
 import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Deactivate
+import org.osgi.service.component.annotations.Reference
+import org.slf4j.LoggerFactory
 import java.util.UUID
 import java.util.concurrent.ExecutorService
+import net.corda.flow.fiber.cache.FlowFiberCache
```
chrisr3: Oops...

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventPipelineImpl.kt
```diff
@@ -21,6 +21,7 @@ import org.slf4j.LoggerFactory
 import java.nio.ByteBuffer
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
+import net.corda.flow.fiber.cache.FlowFiberCache
```
chrisr3: Oops...

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextCacheImpl.kt
```diff
@@ -15,6 +15,7 @@ import java.time.Duration
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentMap
+import net.corda.sandboxgroupcontext.SandboxGroupContextPreRemovalCallback
```
chrisr3: Oops...

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextCache.kt
```diff
@@ -11,6 +12,7 @@ interface SandboxGroupContextCache : AutoCloseable {
     fun remove(virtualNodeContext: VirtualNodeContext): CompletableFuture<*>?
     fun get(
         virtualNodeContext: VirtualNodeContext,
+        preSandboxRemovalCallback: SandboxGroupContextPreRemovalCallback? = null,
```
chrisr3: My general distaste for Kotlin functions having default parameters aside, I don't think your new callback parameter sits well in this API. All other sandbox "clean-up" operations are installed as instances of `AutoCloseable`  during the `createFunction` action. Why can't your callback just be another one of those? Do (`FLOW`) sandboxes really need _two_ "clean-up" mechanisms?

Could whatever this callback is doing (for `FLOW` sandboxes) be absorbed into `FlowSandboxGroupContextImpl`?

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextCacheImpl.kt
```diff
@@ -42,6 +43,7 @@ internal class SandboxGroupContextCacheImpl private constructor(
      * invoke [CloseableSandboxGroupContext.close] on cache eviction when all strong references are gone.
      */
     private class SandboxGroupContextWrapper(
+        val preRemovalCallback: SandboxGroupContextPreRemovalCallback?,
```
chrisr3: Yuk. This code was already complicated, and now it has a new callback too? And just for the sake of `FLOW` sandboxes?

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
```diff
@@ -58,6 +58,7 @@ import java.util.SortedMap
 import java.util.TreeMap
 import java.util.UUID
 import java.util.concurrent.CompletableFuture
+import net.corda.sandboxgroupcontext.SandboxGroupContextPreRemovalCallback
```
chrisr3: Oops.

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
```diff
@@ -165,7 +167,15 @@ class SandboxGroupContextServiceImpl @Activate constructor(
         virtualNodeContext: VirtualNodeContext,
         initializer: SandboxGroupContextInitializer
     ): SandboxGroupContext {
-        return cache.get(virtualNodeContext) { vnc ->
+        return getOrCreate(virtualNodeContext, null, initializer)
```
chrisr3: This callback is getting everywhere :facepalm:

chrisr3 (commented):
I do not like `SandboxGroupContextPreRemovalCallback`.

A `SandboxGroupContext` is destroyed when it is closed, which is why the initialiser function returns an `AutoCloseable`. If the FLOW sandboxes have some extra cleaning to do before they're destroyed then that's fine, but that cleaning should be confined to `FlowSandboxServiceImpl` / `FlowSandboxGroupContextImpl` and not affect the underlying `SandboxGroupContextServiceImpl` or `SandboxGroupContextCacheImpl` components.

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextCacheImpl.kt
```diff
@@ -60,8 +62,15 @@ internal class SandboxGroupContextCacheImpl private constructor(
             // is not referenced anymore).
             .removalListener { key, context, cause ->
                 purgeExpiryQueue()
-                (context?.wrappedSandboxGroupContext as? AutoCloseable)?.also { autoCloseable ->
-                    toBeClosed += ToBeClosed(key!!, context.completion, autoCloseable, context, expiryQueue)
+
+                context?.let { ctx ->
+                    key?.let {
+                        ctx.preRemovalCallback?.preSandboxRemoval(key)
+                    }
```
chrisr3: The (`FLOW`) sandbox is only being evicted from the cache here, and is possibly still being used. So starting to "clean it up" now is probably a _Really Bad Idea_ :tm:. In fact, this is _precisely_ why we wrap the context in a `ToBeClosed` instance in the first place.

`ToBeClosed` extends `WeakReference`, which allows the JVM to tell us when all strong references to the `wrappedSandboxGroupContext` have been garbage-collected. And _then_ we can close it safely.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/factory/FlowFiberFactoryImpl.kt
sprintengine: Not quite sure what you're referring to here, the ordering of these imports? 

Generally I don't reformat existing classes as PR footprints quickly grow and makes it harder to review. It also seems to be reordering most of these imports. Maybe that's an issue with my intelliJ settings as of late

chrisr3: Yes, that new `import` should be with the other `import`s from the same package. I thought we were trying to keep them sorted? (IntelliJ doesn't appear to be either co-operative or consistent though.)

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextCacheImpl.kt
sprintengine: The cached flow fiber has a strong reference to the sandbox (flowLogic), so there is a circular dependency preventing the sandboxes in the expiry queue from closing. That's why I have to perform this "pre" close

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
sprintengine: Yeah you're right, this approach is pretty ugly. So we've removed this callback and made the initializer return a "SandboxCloseable" which has a pre-close step and a close step. We call pre-close first before creating ToBeClosed and adding into expiry queue. How do you feel about this change?

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextCacheImpl.kt
sprintengine: I've made a few updates. Unfortunately as it stands verification/entity sandboxes have to provide empty implementation for  "preClose"

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
chrisr3: IMO you should not need to change the underlying sandbox code _at all_. The `AutoClosable`s are executed _in order_, so why not simply put your extra step first?

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextCacheImpl.kt
chrisr3: Then how about _not_ having a strong reference? Weak references are available...

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/factory/FlowEventPipelineFactoryImplTest.kt
```diff
@@ -54,9 +54,10 @@ class FlowEventPipelineFactoryImplTest {
         flowGlobalPostProcessor,
         flowCheckpointFactory,
         mock(),
+        mock(),
         listOf(flowEventHandler),
         listOf(flowWaitingForHandler),
-        listOf(flowRequestHandler)
+        listOf(flowRequestHandler),
```
chrisr3: Why are people adding superfluous commas these days?

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/factory/FlowEventPipelineFactoryImplTest.kt
```diff
@@ -68,6 +69,7 @@ class FlowEventPipelineFactoryImplTest {
             flowRunner,
             flowGlobalPostProcessor,
             flowEventContext,
+            mock(),
```
chrisr3, line 74: Another unnecessary change :exploding_head:

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextCacheImpl.kt
```diff
@@ -60,8 +61,9 @@ internal class SandboxGroupContextCacheImpl private constructor(
             // is not referenced anymore).
             .removalListener { key, context, cause ->
                 purgeExpiryQueue()
-                (context?.wrappedSandboxGroupContext as? AutoCloseable)?.also { autoCloseable ->
-                    toBeClosed += ToBeClosed(key!!, context.completion, autoCloseable, context, expiryQueue)
+                (context?.wrappedSandboxGroupContext as? SandboxCloseable)?.also { sandboxCloseable ->
+                    sandboxCloseable.preClose(key!!)
```
chrisr3: This is just an accident waiting to happen. This "preClose" is being executed while the sandbox is likely still to be in use, and somebody _**will**_ eventually put something in a `preClose` handler that will break that sandbox.

If the problem is a strong reference somewhere then the solution is _not_ to have a strong reference.

JamesHR3: I'm not sure it's that simple because I think the reference here is to a class loaded out the sandbox (the flow itself), and the point is that we don't want to let go of that reference while both the sandbox and fiber are cached.

It might be that we're missing something here that would make this possible. @mnesbit do you have any insight here that might help?

chrisr3: I believe I already suggested using a `WeakReference`...?

But the idea of adding this _very_ dangerous callback for purely flow-sandbox-specific reasons scares the :hankey:  out of me.

JamesHR3: We don't hold a strong reference directly though, I think it's referenced through the fact that `FlowLogicAndArgs` holds a class that we've loaded out the sandbox and has sandbox services injected into it. I don't think we can introduce a weak reference at the point we'd need it.

We already know that a weak reference to the fiber itself doesn't work (the cache no longer does anything as it holds the only reference), and presumably a weak reference to `FlowLogicAndArgs` or the underlying `Flow` also falls into the same problem.

chrisr3:
```kotlin
override fun preClose(virtualNodeContext: VirtualNodeContext) {
    flowFiberCache.remove(virtualNodeContext.holdingIdentity.toAvro())
}
```
One obvious question: If all `preClose()` does is remove the flow from the `FlowFiberCache`, why do you need to do anything at all? Doesn't this step happen automatically when the flow completes (either successfully or unsuccessfully) anyway?

Could the `FlowFiberCache` be a "Concurrent Weak Value Map"? `Caffeine.weakValues()` is "A Thing".

JamesHR3: Weak values don't help because there's no other reference to the fibers when they're not running, so they just get removed from the cache straight away.

You can't just wait for the flow to finish either because if the sandbox gets unloaded and reloaded we'll attempt to serialize the fiber using the new sandbox's Kryo instance, which will fail because the classes were loaded originally in the old sandbox. (There's also issues around rebalancing although that's addressed by also evicting from the fiber cache when the partition for that flow is rebalanced.)

mnesbit: I agree this is ugly, but I am not sure we can do much petter. I would NOT call it preClose as that is hinting at too much re-usability. Perhaps we call it onInvalidate, since it is done when sandbox cache evicts the item.

chrisr3: > I agree this is ugly, but I am not sure we can do much better.

I am _convinced_ there must be a better way, and have dusted off my Evil Hat, Evil Cloak and Evil Boots! :supervillain:

(They weren't _that_ dusty...)

chrisr3: I think a better solution is modifying `SandboxGroupContextCache` etc to support "eviction listeners", and then have `FlowSandboxService` register a handler. I can put the "eviction listener" support together quickly in a separate PR - assuming all this is for 5.1?

mnesbit: The callback pointers may well prevent GC. I think we will have to draw out the pointer graph to make sure it works

chrisr3: The eviction listeners are "in". And the callback belongs to `FlowSandboxServiceImpl` and so should not prevent GC.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
```diff
@@ -0,0 +1,61 @@
...
+    override fun get(key: FlowKey): FlowFiberImpl? {
+        return cache.getIfPresent(key)
+    }
+
+    override fun remove(key: FlowKey) {
+        cache.invalidate(key)
+    }
+
+    override fun remove(keys: Collection<FlowKey>) {
+        logger.info("Removing ${keys.size} flow fibers from flow fiber cache: ${keys.joinToString()}")
```
sprintengine: info just for a few tests

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/sandbox/impl/FlowSandboxServiceImpl.kt
```diff
@@ -65,6 +69,7 @@ class FlowSandboxServiceImpl @Activate constructor(
 
     private fun onEviction(vnc: VirtualNodeContext) {
         logger.debug("Sandbox {} has been evicted", vnc)
+        flowFiberCache.remove(vnc.holdingIdentity.toAvro())
```
sprintengine, line 72: Logs showing this evicting fiber cache and then sandbox cache:

```
17:11:24.994 [SecManagerForkJoinPool-worker-12] INFO  net.corda.flow.fiber.cache.impl.FlowFiberCacheImpl {} - Flow fiber cache removing holdingIdentity {"x500Name": "CN=R3, O=Alice-2-d3be0124-8a17-47ca-bc6e-389043e216df, L=London, C=GB", "groupId": "42e60349-dc87-4b9a-a861-d9f57c96f960"}
17:11:24.994 [SecManagerForkJoinPool-worker-12] INFO  net.corda.sandboxgroupcontext.service.impl.SandboxGroupContextCacheImpl {} - Evicting flow sandbox for CN=R3, O=Alice-2-d3be0124-8a17-47ca-bc6e-389043e216df, L=London, C=GB [SIZE]
```

components/flow/flow-service/test.bndrun
```diff
@@ -22,6 +23,7 @@
     co.paralleluniverse.quasar.verbose=false
 
 -runrequires: \
+    bnd.identity;id='co.paralleluniverse.quasar-core',\
```
chrisr3: Why do you need to add this? I would have thought that Quasar would already have been included _automatically_.

The `-runrequires` are the "seed" requirements for Bnd's `Resolve` task. The full list of bundles to load at runtime is written to `build/resolved-test.bndrun`, and should already contain:
```
co.paralleluniverse.quasar-core;version='[0.9.0,0.9.1)',\
co.paralleluniverse.quasar-core.framework.extension;version='[0.9.0,0.9.1)',\
```
You should be able to see Quasar's internal logging by changing:
```
-runproperties: \
    co.paralleluniverse.quasar.verbose=false
```
to `true` :wink:.

components/flow/flow-service/test.bndrun
```diff
@@ -10,7 +10,8 @@
 -runvm: \
     -Djava.util.logging.config.file=${.}/logging.properties,\
     -Djava.io.tmpdir=${task.temporaryDir},\
-    --illegal-access=deny
+    --illegal-access=deny,\
+    --add-opens=java.base/java.lang=ALL-UNNAMED
```
chrisr3: Other `test.bndrun` files have written this as
```
--add-opens, 'java.base/java.lang=ALL-UNNAMED',\
```

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/context/OutputAssertions.kt
```diff
@@ -80,6 +81,12 @@ interface OutputAssertions {
     fun noEntityRequestSent()
 
     fun flowKilledStatus(flowTerminatedReason: String)
+
+    fun expectFlowFiberCacheContainsKey(holdingId: HoldingIdentity, flowId: String)
+
+    fun expectFlowFiberCacheDoesNotContain(holdingId: HoldingIdentity, flowId: String)
```
jujoramos: Nitpick: use `expectFlowFiberCacheDoesNotContainKey` as `expectFlowFiberCacheContainsKey`, or remove the `Key` suffix from both methods.

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/context/OutputAssertions.kt
```diff
@@ -80,6 +81,12 @@ interface OutputAssertions {
...
+
+    fun expectFlowFiberCacheOperations(holdingId: HoldingIdentity, flowId: String, expected: List<FlowFiberCacheOperation>)
```
jujoramos: Same as above, would be better to have the method names consistent (`expectFlowFiberCacheOperationsForKey`?).

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/fakes/FakeFlowFiberCache.kt
```diff
@@ -0,0 +1,67 @@
+package net.corda.flow.testing.fakes
+
+import net.corda.data.flow.FlowKey
+import net.corda.data.identity.HoldingIdentity
+import net.corda.flow.fiber.FlowFiberImpl
+import net.corda.flow.fiber.cache.FlowFiberCache
+import org.osgi.service.component.annotations.Component
+
+@Component(service = [FlowFiberCache::class, FakeFlowFiberCache::class])
+class FakeFlowFiberCache : FlowFiberCache {
```
jujoramos: Can't we use the real instance of `FlowFiberCacheImpl` instead of a stub for the tests?, we wouldn't be really testing the cache otherwise. I might be missing some integration issues (with OSGi or quasar probably?), so feel free to ignore this comment if that's the case.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/FlowFiberCache.kt
```diff
@@ -0,0 +1,35 @@
...
+import net.corda.data.flow.FlowKey
+import net.corda.data.identity.HoldingIdentity
+import net.corda.flow.fiber.FlowFiberImpl
+
+/**
+ * Cache for flow fibers.
+ */
+interface FlowFiberCache {
+    /**
+     * Put a flow fiber into the cache keyed by the given flowId.
```
jujoramos: Nitpick: the key is not the `flowId` but the `flowId` + `holdingId`, right?. Same for the other KDoc within this interface.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
```diff
@@ -0,0 +1,61 @@
...
+//        logger.debug { "Removing ${keys.size} flow fibers from flow fiber cache: ${keys.joinToString()}" }
+        cache.invalidateAll(keys)
+        cache.cleanUp()
```
jujoramos, line 51: Do we need the `cleanUp()` here?, isn't the actual clean up automatically handled by Caffeine asynchronously when needed?.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    override fun remove(holdingIdentity: HoldingIdentity) {
+        logger.info("Flow fiber cache removing holdingIdentity $holdingIdentity")
+//        logger.debug { "Flow fiber cache removing holdingIdentity $holdingIdentity" }
+        val keysToInvalidate = cache.asMap().keys.filter { holdingIdentity == it.identity }
+        cache.invalidateAll(keysToInvalidate)
+        cache.cleanUp()
```
jujoramos, line 58: Do we need the `cleanUp()` here?, isn't the actual clean up automatically handled by Caffeine asynchronously when needed?.

components/flow/flow-service/test.bndrun
sprintengine: Thanks, you're right!

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
chrisr3: Personally, I would prefer _not_ to wait for Caffeine to pull its finger out... :wink:

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/fakes/FakeFlowFiberCache.kt
JamesHR3: Think I'm inclined to agree here. Looks like you've added this for extra tracking, but I'd be tempted to just assert what you expect the state to be after running some (or all) of a flow and make sure that's correct. You might need to break up some acceptance tests in order to do that effectively (multiple `when`/`then` blocks), but that shouldn't be too hard.

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/context/FlowServiceTestContext.kt
```diff
@@ -391,12 +406,17 @@ class FlowServiceTestContext @Activate constructor(
...
             sessionInitiatingIdentity,
-            sessionInitiatedIdentity
+            sessionInitiatedIdentity,
+            flowFiberCache
         )
         assertions.add(assertionsCapture)
         outputAssertions(assertionsCapture)
     }
 
+    override fun resetFlowFiberCache() {
```
JamesHR3, line 417: Is this going to need to be called after each test and if so is it possible to do this automatically? When is `clearTestRuns` called for example?

I think without the fake you could still do this but you'd need to know what holding identities to clean up - possibly you can do this by looking up all holding identities available to the test and clearing them all out?

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
sprintengine: I don't think we want to leave flow fibers hanging around with hard references to sandboxes, stalling sandbox cache eviction, waiting for caffeine to do it

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/context/FlowServiceTestContext.kt
sprintengine: I think what you're looking for is `resetTestContext()`, I already added it there, but that does remind me I should be able to remove explicitly calling it in some of the tests' beforeEach.

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/fakes/FakeFlowFiberCache.kt
sprintengine: I got rid of the extra tracking, it's probably not really necessary.

JamesHR3 (approved):

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/ExternalEventAcceptanceTest.kt
```diff
@@ -464,13 +465,20 @@ class ExternalEventAcceptanceTest : FlowServiceTestBase() {
                 )
         }
 
+        then {
+            expectOutputForFlow(FLOW_ID1) {
+                flowFiberCacheContainsKey(ALICE_HOLDING_IDENTITY, REQUEST_ID1)
+            }
+        }
+
         `when` {
```
chrisr3, line 474: Doesn't `mockito-kotlin` have a `whenever` function to use instead of `when`? Seeing as `when` is a reserved word in Kotlin.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberImpl.kt
```diff
@@ -158,6 +158,17 @@ class FlowFiberImpl(
         }
     }
 
+    /**
+     * Prepare the fiber for caching by removing unnecessary transient fields. These fields will be set
+     * on the fiber after it is selected from the cache and resumed in [resume].
```
chrisr3, line 163: If the fields are marked as `transient` then why do you need to `null` them out? Kryo doesn't serialize `transient` fields unless explicitly configured to. And AFAIK we _don't_ configure it to.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
```diff
@@ -0,0 +1,60 @@
...
+import net.corda.data.flow.FlowKey
+import net.corda.data.identity.HoldingIdentity
+import net.corda.flow.fiber.FlowFiberImpl
+import net.corda.flow.fiber.cache.FlowFiberCache
+import net.corda.utilities.debug
+import org.slf4j.LoggerFactory
+
+@Suppress("unused")
+@Component(service = [FlowFiberCache::class])
+class FlowFiberCacheImpl @Activate constructor() : FlowFiberCache {
```
chrisr3, line 18: Do you need the `@Activate constructor` bit?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
```diff
@@ -33,6 +33,7 @@ import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
 import org.slf4j.LoggerFactory
 import java.time.Instant
+import net.corda.flow.fiber.cache.FlowFiberCache
```
chrisr3, line 36: IntelliJ has put this `import` in the wrong place.

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImplTest.kt
```diff
@@ -162,6 +222,7 @@ class FlowEventExceptionProcessorImplTest {
 
         verify(flowCheckpoint).waitingFor = WaitingFor(net.corda.data.flow.state.waiting.Wakeup())
         verify(flowCheckpoint).setPendingPlatformError(FlowProcessingExceptionTypes.PLATFORM_ERROR, error.message)
+        verify(flowFiberCache, times(0)).remove(any<List<FlowKey>>())
```
chrisr3, line 225: `times(0)` -> `never()`

components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowExecutorRebalanceListener.kt
```diff
@@ -0,0 +1,36 @@
...
+    override fun onPartitionLost(states: Map<String, Checkpoint>) {
+        flowWakeUpScheduler.onPartitionLost(states)
+        removeCachedFlowFibersForHoldingIdentity(states.values.map { it.flowState.flowStartContext.statusKey }.toSet())
+    }
+
+    override fun onPostCommit(updatedStates: Map<String, Checkpoint?>) {
+        flowWakeUpScheduler.onPostCommit(updatedStates)
+    }
+
+    private fun removeCachedFlowFibersForHoldingIdentity(toBeRemoved: Set<FlowKey>) {
```
rick-r3, line 33: I would argue the name of this function is misleading (it's not holding identities, but flows), but low priority.  Rename in 5.1 etc.

components/flow/flow-service/test.bndrun
```diff
@@ -10,7 +10,8 @@
 -runvm: \
     -Djava.util.logging.config.file=${.}/logging.properties,\
     -Djava.io.tmpdir=${task.temporaryDir},\
-    --illegal-access=deny
+    --illegal-access=deny,\
+    --add-opens, 'java.base/java.lang=ALL-UNNAMED'
```
chrisr3, line 15: Does this mean that the fiber cache works and needs no extra integration tests?

chrisr3 (commented):
Just a few thoughts.

rick-r3 (approved):
LGTM.  See one cosmetic comment about function name.

jujoramos (approved):

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberImpl.kt
sprintengine: This isn't anything to do with the kryo serialization, we null out these fields on the fiber before we put it into the cache, the serialized form is [also present](https://github.com/corda/corda-runtime-os/blob/14c45a343541b94fd94147265587d8738b815d8c/components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberImpl.kt#L139) in the FlowSuspended object, and it is put into the checkpoint.

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/ExternalEventAcceptanceTest.kt
sprintengine: This is `when` from FlowServiceTestBase, part of the integration tests.

chrisr3: > This is `when` from FlowServiceTestBase, part of the integration tests.

So _we_ wrote a function whose name clashes with a Kotlin reserved word?!!? :exploding_head:

components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowExecutorRebalanceListener.kt
sprintengine: Yeah good spot, I will fix this in 5.1

PR #3542, opened 2023-04-03
CORE-11948: State synchronization enhancements in the state and event message pattern

JamesHR3:
## Problem Description
The state and event message pattern uses two message bus consumers to consume from two topics: the event consumer from the event topic, and the state consumer from the corresponding state topic. When a partition is first assigned to the event consumer, it is paused there and the corresponding state partition is manually assigned to the state consumer. The state consumer must then read all the states from that partition until the current state is held in memory, after which the event partition can be unpaused and event processing can resume.

This process was previously slow for two reasons. On startup, if there was any previous activity on the topics, the pattern would have all event partitions paused and so no events could be processed. The pattern would then alternate between calling the state consumer poll and the event consumer poll. As all event partitions were paused, this incurred a mandatory `EVENT_POLL_TIMEOUT` pause on every iteration as no data would be fetched (in this case 100ms). Once one partition was in sync, the state consumer would still be assigned to that partition, and so some of the data read back would be information already stored in the in-memory state anyway.

## Solution
This PR introduces two enhancements to try and address this. Firstly, the `StateAndEventConsumer` now keeps track of all partitions currently being synchronized, and all partitions that are currently in sync. By doing this, the event consumer `poll` can be avoided if it has been invoked recently and there are no currently in-sync partitions. Secondly, the state consumer is now unassigned partitions when they are in sync, which allows polls of the state consumer to only return data needed to update the in-memory state.

## Other changes
- The logic around assigning, pausing and removing partitions from the state and event consumers has been moved from `StateAndEventRebalanceListener` to `StateAndEventConsumer`. That keeps the logic around managing the two consumers better encapsulated in the `StateAndEventConsumer` wrapper class.
- A bug was identified where the topic prefix was not correctly stripped from the `CordaTopicPartition` structure when the consumer rebalance listener was invoked from the Kafka message bus implementation. This caused a bug with the refactored code that was previously hidden.


libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
```diff
@@ -372,11 +372,11 @@ class CordaKafkaConsumerImpl<K : Any, V : Any>(
     ) {
         when {
             listener != null -> {
-                consumer.subscribe(topics, listener.toKafkaListener())
+                consumer.subscribe(topics, listener.toKafkaListener(config.topicPrefix))
             }
 
             defaultListener != null -> {
-                consumer.subscribe(topics, defaultListener?.toKafkaListener())
+                consumer.subscribe(topics, defaultListener?.toKafkaListener(config.topicPrefix))
```
jujoramos, line 365: Nitpick: this statement will only execute if `defaultListener` is not `null`, so you probably don't need `defaultListener?` here?.

JamesHR3:
Resiliency test run here: https://ci02.dev.r3.com/job/Corda5/job/corda-resiliency-tests/job/jamesh-core-11948-test/2/. The failures here are related to known problems in the crypto worker and membership workers, and issues around flows taking too long to complete over rebalancing appear to have gone.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/StateAndEventConsumer.kt
```diff
@@ -20,6 +22,34 @@ interface StateAndEventConsumer<K : Any, S : Any, E : Any> : AutoCloseable {
...
+    /**
+     * Reset the event consumer to its last committed position.
+     *
+     * Should be invoked if processing of a batch of events fails to ensure that they are polled again.
+     */
+    fun resetEventOffsetPosition()
+
     /**
      * Get the in memory state value for a given [key]
      */
```
jujoramos, line 55: Would it make sense to remove direct access to the `eventConsumer` and `stateConsumer` here (see https://r3-cev.atlassian.net/browse/CORE-9569)?. Just wondering if it's possible without a lot of rework of the `PR`, though, feel free to ignore the suggestion.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/StateAndEventConsumerImpl.kt
```diff
@@ -88,42 +178,55 @@ internal class StateAndEventConsumerImpl<K : Any, S : Any, E : Any>(
...
-        eventConsumer.resume(partitionsSynced)
-
-        stateAndEventListener?.let { listener ->
-            for (partition in partitionsSynced) {
-                listener.onPartitionSynced(getStatesForPartition(partition.partition))
+    override fun pollEvents(): List<CordaConsumerRecord<K, E>> {
+        return when {
+            inSyncPartitions.isNotEmpty() -> {
+                eventConsumer.poll(EVENT_POLL_TIMEOUT).also {
+                    log.debug { "Received ${it.size} events to process" }
```
jujoramos: The actual size is generally not useful enough when debugging issues, can we use `log.debug { "Received events(keys: ${events.joinToString { it.key.toString() }}, size: ${events.size})" }` or something like that instead?. Same comment for the other branches of the `when` statement.

jujoramos (dismissed):
LGTM, just some minor comments.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
JamesHR3: Annoyingly that doesn't work because `defaultListener` is declared as `var`, and so the compiler realizes that it could be changed by a different thread from one line to the next.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/StateAndEventConsumer.kt
JamesHR3: I did think about this but it's a fair bit of refactoring still and I didn't want to do too much more in this PR.

For the event consumer, there's a single usage to subscribe (which isn't too bad), but then there's a call to `producer.sendRecordOffsetsToTransaction` which takes the event consumer as a parameter, and refactoring that is probably a lot of work.

For the state consumer, there's a single usage in the `StateAndEventRebalanceListener` to work out what partitions need to be created in the current state map. This is actually probably a bit dodgy, but I think a more serious refactor would be required to address this sensibly.

Longer term I suspect we should split the pattern up a bit more as `StateAndEventConsumerImpl` is becoming too big - but that can be addressed at some later point.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/StateAndEventConsumerImpl.kt
JamesHR3: Good call - thanks. I'll update.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/StateAndEventConsumer.kt
jujoramos: Yep, sounds good, thanks 👍

jujoramos (approved):

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
```diff
@@ -34,6 +29,9 @@ import net.corda.schema.Schemas.getStateAndEventDLQTopic
 import net.corda.schema.Schemas.getStateAndEventStateTopic
 import net.corda.utilities.debug
 import org.slf4j.LoggerFactory
+import java.nio.ByteBuffer
+import java.time.Clock
+import java.util.*
```
sprintengine, line 34: I think you can replace this * import with UUID

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/StateAndEventConsumerImpl.kt
```diff
@@ -46,10 +53,94 @@ internal class StateAndEventConsumerImpl<K : Any, S : Any, E : Any>(
...
+            val end = endOffsets[statePartition] ?: 0L
+            beginning < end
+        }
+
+        log.debug { "The following partitions need to sync: $needsSync. The following partitions are in sync: $inSync" }
+
+        // Out of sync partitions need assigning to the state consumer to bring us into sync.
+        partitionsToSync.addAll(needsSync)
+        eventConsumer.pause(needsSync)
+        stateConsumer.seekToBeginning(needsSync.map { it.toStatePartition() })
```
sprintengine, line 80: What if, instead of starting seeking the partition from the beginning what if we started from the beginningOffset for this partition which is already calculated up above? (using api `fun seek(partition: CordaTopicPartition, offset: Long)`)

Would that mean we avoid traversing over records we already have?

edit - when onPartitionsAssigned is called (by rebalance listener) it'll only pass the new partitions so we won't have already read any states for those partitions, do probably no gain here - docs for ConsumerRebalanceListener 
> "Params:
partitions – The list of partitions that are now assigned to the consumer (previously owned partitions will NOT be included, i.e. this list will only include newly added partitions)"

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/StateAndEventConsumerImpl.kt
```diff
@@ -46,10 +53,94 @@ internal class StateAndEventConsumerImpl<K : Any, S : Any, E : Any>(
     private val initialProcessorTimeout = maxPollInterval / 4
 
     private val currentStates = partitionState.currentStates
-    private val partitionsToSync = partitionState.partitionsToSync
+    private val partitionsToSync: MutableSet<CordaTopicPartition> = ConcurrentHashMap<CordaTopicPartition, Unit>().keySet(Unit)
+    private val inSyncPartitions: MutableSet<CordaTopicPartition> = ConcurrentHashMap<CordaTopicPartition, Unit>().keySet(Unit)
```
sprintengine, line 57: Minor - I think `ConcurrentHashMap.newKeySet()` will get you the thread safe set in slightly less characters

```
    private val partitionsToSync: MutableSet<CordaTopicPartition> = ConcurrentHashMap.newKeySet()
    private val inSyncPartitions: MutableSet<CordaTopicPartition> = ConcurrentHashMap.newKeySet()
```

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/StateAndEventConsumerImpl.kt
```diff
@@ -46,10 +53,94 @@ internal class StateAndEventConsumerImpl<K : Any, S : Any, E : Any>(
...
+    private fun onPartitionsSynchronized(partitions: Set<CordaTopicPartition>) {
+        // Remove the assignment from the state consumer. There's no need to read back from the state topic, as from
+        // now on the pattern will rely on the in-memory state.
+        log.info(
+            "$partitions are now in sync. Resuming event feed. " +
+                    "Current in sync: $inSyncPartitions, current to be synced: $partitionsToSync"
+        )
+        updateStateConsumerAssignment(partitions, StatePartitionOperation.REMOVE)
+
+        eventConsumer.resume(partitions)
```
sprintengine, line 101: Just something I'm not entirely clear on, by resuming these partitions (so polls will not fetch from paused partitions) do we actually consume events from in-sync partitions while the other state partitions are sync'ing? or does `pollAndUpdateStates` effectively blocks processing batches of events until all state partitions are in sync?

conal-smith (approved):

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/StateAndEventConsumerImpl.kt
JamesHR3: We'll process events from any unpaused partitions, so we can start processing once any partition is in sync rather than waiting for all of them.

PR #3543, opened 2023-04-03
CORE-9725 Workers create sandboxes from CPK hashes passed through external events from flow checkpoint

ben-millar:
This PR satisfies [CORE-9725](https://r3-cev.atlassian.net/browse/CORE-9725).
Associated `corda-api` PR: https://github.com/corda/corda-api/pull/1040

#### Context:
Full context can be found in the ticket linked above, but in short: Today, sandboxes for external workers are being created based on the information stored in the `CpiInfoReadService`. To ensure consistency, we should instead create the sandbox from the same set of CPKs present in the flow checkpoint at the time the external event is dispatched.

#### Fix:
To achieve the above, CPK file checksums have been added to the `ExternalEventContext.ContextProperties` K:V map with keys `"corda.cpk.hash.0" -> "corda.cpk.hash.n"`, which are then extracted in the appropriate request processors and passed through to the respective SandboxServiceFactory. We then use the `CpkReadService` to retrieve the CPK metadata which is used to get/create the sandboxes for external workers.

#### Changlog:
- `EntitySandboxService` now expects a `Set<SecureHash>` of the CPK file hashes needed to initialise a checkpoint, and the factory has been updated accordingly.
- `ExternalEventRequestHandler` modified to add CPK file hashes from the origin checkpoint to external events.
- `PersistenceRequestProcessor` and `EntityMessageProcessor` modified to pull CPK file hashes from incoming events and passing those through to `EntitySandboxService::get`
- `getClass` extension function in `EntityMessageProcessor` which previously extended `EntitySandboxService` now extends `SandboxGroupContext`; this removes the need to pass through a holdingIdentity as the context is implicit from the call.
- `PersistenceServiceInternal` updated with new `getClass` function signature.
- A new utility function, `.toKeyValuePairList(prefix: String?)` has been added to KeyValueStore. This allows us to convert a generic `Iterable` to a KeyValuePairList. Keys are generated from the passed prefix, or the `simpleName` of `<T>` if no prefix is passed.
- A new file, `CpkTestUtils.kt`, has been added to `testing:test-utilities` which extends `CpiInfoReadService` to allow us to easily retrieve a set of CPK filehashes given a `VirtualNodeInfo` object.
- Unit tests and integration tests have been added or updated for all changes.

[CORE-9725]: https://r3-cev.atlassian.net/browse/CORE-9725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gradle.properties
```diff
@@ -44,7 +44,7 @@ bouncycastleVersion=1.72
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.0.0.xx-SNAPSHOT to pick up maven local published copy
 #cordaApiVersion=5.0.0.xxx-SNAPSHOT
-cordaApiVersion=5.0.0.743-beta+
+cordaApiVersion=5.0.0.743-alpha-1680620535349
```
ben-millar: This will be changed prior to merging, once the API changes are finalised.

ben-millar:
Blocked by: https://github.com/corda/corda-api/pull/1040

mcgovc (approved):
LGTM - build file changes

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
```diff
@@ -216,12 +230,13 @@ class PersistenceExceptionTests {
...
                 virtualNode.sandboxGroupContextComponent,
-                cpiInfoReadService,
+                cpkReadService,
                 virtualNodeInfoReadService,
                 dbConnectionManager
             )
 
-        val sandboxOne = entitySandboxService.get(virtualNodeInfoOne.holdingIdentity)
+        val cpkFileHashes = cpiInfoReadService.getCpkFileHashes(virtualNodeInfoOne)
+        val sandboxOne = entitySandboxService.get(virtualNodeInfoOne.holdingIdentity, cpkFileHashes)
```
emilybowe, line 239: The test mentions two sandboxes and I can only find sandbox and sandboxOne so maybe rename to sandbox and additionalSandbox / sandbox and sandboxTwo or something like that ?

If it was sandbox and sandboxTwo it would match up better with cpkFileHashesTwo etc

driessamyn (approved):
LGTM from DB sandbox pov.

emilybowe (approved):
LGTM

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
ben-millar: You're right, this actually confused me when I was refactoring the tests so I should've fixed it. If I have any other updates to make to the PR, I'll fix this.

PR #3549, opened 2023-04-04
CORE-11436 ledger persistence exceptions

blsemo:
Treat any exceptions except PersistenceException as platform exceptions (non-retryable)

### DIFF
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/processor/PersistenceRequestProcessor.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/processor/PersistenceRequestProcessor.kt
@@ -3,7 +3,6 @@ package net.corda.ledger.persistence.processor
 import net.corda.crypto.core.parseSecureHash
 import net.corda.data.ledger.persistence.LedgerPersistenceRequest
 import net.corda.ledger.persistence.common.InconsistentLedgerStateException
-import net.corda.ledger.common.data.transaction.InvalidTransactionMetadataException
 import net.corda.ledger.persistence.common.UnsupportedLedgerTypeException
 import net.corda.ledger.persistence.common.UnsupportedRequestTypeException
 import net.corda.messaging.api.processor.DurableProcessor
@@ -15,7 +14,7 @@ import net.corda.utilities.withMDC
 import net.corda.v5.application.flows.FlowContextPropertyKeys.CPK_FILE_CHECKSUM
 import net.corda.virtualnode.toCorda
 import org.slf4j.LoggerFactory
-import java.lang.IllegalArgumentException
+import javax.persistence.PersistenceException
 
 /**
  * Handles incoming requests, typically from the flow worker, and sends responses.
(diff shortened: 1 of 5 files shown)


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/processor/PersistenceRequestProcessor.kt
```diff
@@ -52,9 +53,14 @@ class PersistenceRequestProcessor(
                                     responseFactory.fatalErrorResponse(request.flowExternalEventContext, e)
                                 }
 
-                                else -> {
+                                is PersistenceException -> {
                                     responseFactory.errorResponse(request.flowExternalEventContext, e)
```
mnesbit: Perhaps we should check https://docs.jboss.org/hibernate/jpa/2.1/api/javax/persistence/EntityTransaction.html#getRollbackOnly() as the not done type exceptions e.g LockTimeout seem to leave the transaction open and could be retried locally and the others are aborted

blsemo: I don't understand this comment? This is not within a database transaction - this is formatting the AVRO response to send back to the flow worker.
If we send back a `transientErrorResponse`, the flow worker will retry with some back-off, which might make sense if it was a potentially transient database error, but not if some transaction pre-condition failed.

vlajos (dismissed):
LGTM

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/processor/PersistenceRequestProcessor.kt
```diff
@@ -52,9 +53,14 @@ class PersistenceRequestProcessor(
                                     responseFactory.fatalErrorResponse(request.flowExternalEventContext, e)
                                 }
 
-                                else -> {
+                                is PersistenceException -> {
```
lankydan: I think this is in the wrong place.

I think it would be better to switch the behaviour of `ResponseFactory.errorResponse`.

```kotlin
override fun errorResponse(externalEventContext : ExternalEventContext, exception: Exception) = when (exception) {
        is CpkNotAvailableException, is VirtualNodeException ->
            transientErrorResponse(externalEventContext, exception)
        is NotSerializableException ->
            platformErrorResponse(externalEventContext, exception)
        is KafkaMessageSizeException, is NullParameterException ->
            fatalErrorResponse(externalEventContext, exception)
        // added persistence exception
        is PersistenceException -> transientErrorResponse(externalEventContext, exception)
        // switched to platform
        else -> platformErrorResponse(externalEventContext, exception)
    }
```

This has the knock-on effect of changing the entity persistence behaviour, but I think both the ledger and entity persistence code should have the same behaviour here anyway.

The problem with the current code, is that if we have `CpkNotAvailableException`s etc, these will now be platform errors rather than transient. These were purposely made transient errors iirc because of timing issues with distributing CPKs in E2E tests.

There is an argument that the `else` block should be a FATAL error, but I'm ok with leaving as a platform error for now since its safer option to default to imo.

driessamyn (approved):
As discussed on Slack. 
There is a risk that this changes existing behaviour for the non-ledger persistence API. 
This may be ok I think but we need to watch out for regressions.

lankydan (approved):

PR #3550, opened 2023-04-04
CORE-9973 Flow mapper received session event for session which does not exist

mbrkic-r3:
Session initiation is triggered from within the flow engine after a session is interacted with for the first time, i.e from the client code calling send or receive. For calls to send, we will send the SessionInit and the SessionData messages at the same time. This could result in the scenario of the flow mapper receiving the data first and sending back an error.  

This will be fixed by changing the session manager getMessagesToSend function. If there is a SessionInit message in the send queue then only send the SessionInit message, ignore other pending messages. Otherwise send all messages in the queue as normal. After the SessionInit is acknowledged by the other side it will be automatically removed from the send queue which will free up other messages to be sent.


mbrkic-r3:
Issue was resolved with https://github.com/corda/corda-runtime-os/pull/3577

PR #3555, opened 2023-04-04
CORE-10583 Block receipt of messages in the Link Manager from suspended members

charlieR3:
Update the inbound and outbound processors to verify that the two parties can communicate with other based. The rules for allowed communication are:

* The MGM can send to members of any status.
* The MGM can receive from members of any status and also members it does not know about yet.
* Suspended members cannot send or receive and messages
* Active members can communicate with active members.

Session messages should allow for sessions to be created with suspended members, but messages would be dropped later when processing data messages.

Requires changes from https://github.com/corda/corda-api/pull/1037


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
```diff
@@ -0,0 +1,83 @@
...
+    fun validate(source: HoldingIdentity, destination: HoldingIdentity) {
+        val sourceMemberInfo = getMemberInfo(source, source.x500Name)
+        val destinationMemberInfo = getMemberInfo(source, destination.x500Name)
+
+        if (!canMessage(sourceMemberInfo, destinationMemberInfo)
+            || !canMessage(destinationMemberInfo, sourceMemberInfo)
+        ) {
+            throw InvalidNetworkStatusForMessaging(
+                "network membership status for the source identity is (${sourceMemberInfo?.status}) and the " +
+                        "source's network membership status is (${destinationMemberInfo?.status})."
```
williamvigorr3: `destination's`

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
```diff
@@ -0,0 +1,83 @@
...
+     * Validates whether [source] can participate in messaging with [destination].
+     *
+     * @throws InvalidNetworkStatusForMessaging if validation failed
+     */
+    fun validate(source: HoldingIdentity, destination: HoldingIdentity) {
+        val sourceMemberInfo = getMemberInfo(source, source.x500Name)
+        val destinationMemberInfo = getMemberInfo(source, destination.x500Name)
+
+        if (!canMessage(sourceMemberInfo, destinationMemberInfo)
+            || !canMessage(destinationMemberInfo, sourceMemberInfo)
```
williamvigorr3, line 55: As we are making this symmetric. Should we make `canMessage` symmetric itself?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
```diff
@@ -0,0 +1,83 @@
...
+class NetworkMessagingValidator(private val membershipGroupReaderProvider: MembershipGroupReaderProvider) {
+
+    private companion object {
+        val logger: Logger = LoggerFactory.getLogger(this::class.java)
+    }
+
+    /**
+     * Validates whether [source] can participate in messaging with [destination].
+     *
+     * @throws InvalidNetworkStatusForMessaging if validation failed
```
williamvigorr3: This might be personal opinion but I think it would be better if this function returned an error, instead of throwing an exception. Otherwise we can end up in a the situation where the code becomes difficult to read/understand  because it is difficult to work out all the places an exception could be thrown (as they propagate up the stack). In this case it is not that bad as we only throw `InvalidNetworkStatusForMessaging` in one place.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -37,10 +39,12 @@ import net.corda.utilities.time.Clock
 import net.corda.virtualnode.toCorda
 import org.slf4j.LoggerFactory
 
+@Suppress("LongParameterList")
 internal class InboundMessageProcessor(
     private val sessionManager: SessionManager,
     private val groupPolicyProvider: GroupPolicyProvider,
     private val membershipGroupReaderProvider: MembershipGroupReaderProvider,
+    private val networkMessagingValidator: NetworkMessagingValidator,
```
williamvigorr3: Very much a nit: We could default here: `private val networkMessagingValidator = NetworkMessagingValidator(groupPolicyProvider)` (probably need to move this argument to be last). Assume this needs to be an argument because we want to inject a mock for unit testing purposes.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
```diff
@@ -42,6 +44,7 @@ internal class OutboundMessageProcessor(
     private val linkManagerHostingMap: LinkManagerHostingMap,
     private val groupPolicyProvider: GroupPolicyProvider,
     private val membershipGroupReaderProvider: MembershipGroupReaderProvider,
+    private val networkMessagingValidator: NetworkMessagingValidator,
```
williamvigorr3: Nit: again we could default here: private val networkMessagingValidator = NetworkMessagingValidator(groupPolicyProvider)

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
```diff
@@ -0,0 +1,83 @@
...
+     * Return true if [party] can message [otherParty].
+     *
+     * If a [party] is not in the member list (i.e. null), or they are not active, they can only message [otherParty]
+     * if [otherParty] is an MGM.
+     * If [party] is an MGM, they can message [otherParty] in any status.
+     * Otherwise, a party can message if [otherParty] is in active status.
+     */
+    private fun canMessage(party: MemberInfo?, otherParty: MemberInfo?) = when {
+        party?.isMgm == true -> true
+        party == null || !party.isActive -> otherParty?.isMgm == true
```
williamvigorr3, line 110: Not sure this `party == null` is correct (what if `otherParty` is suspended and in the member list at this point)?

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidatorTest.kt
```diff
@@ -0,0 +1,130 @@
...
+
+        // MGM member
+        private val mgmHoldingId = mockMember("mgm") {
+            on { isActive } doReturn true
+            on { mgmProvidedContext } doReturn mgmMgmProvidedContext
+        }
+
+        @JvmStatic
+        fun allowedMessagingGroups() = listOf(
+            TestConfig(activeMemberHoldingId, activeMemberHoldingId, true),
```
williamvigorr3, line 64: I think we should also test the cases where the members are not in the member list i.e. three more cases.

williamvigorr3 (commented):
I left some comments.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
charlieR3: If we're trying to check if `party` can message `otherParty` and `party` is null, the only scenario where this would be valid is if `otherParty` is an MGM and `party` is trying to send the initial registration before their own member info is in the memberlist. That's why this check says messaging for a null `party` is only valid if `otherParty` is an MGM. 

If `otherParty` is suspended and in the member list at this point for some reason, messaging will not be allowed because `otherParty` would not be an MGM (since we don't suspend an MGM) so `canMessage` would return false which I think is correct. I don't think `party == null & otherParty.status == SUSPENDED` should return true.

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidatorTest.kt
charlieR3: That should be covered by `registeringMemberHoldingId`. I've called it `registeringMemberHoldingId` because the only valid case I can think of where a member info might not be in the member list but messaging should still be allowed is when a member tries to contact an MGM for initial registration (or possible sync if they lose their local copy of the member list)

This test class should cover active member, suspended member, `null` member, and MGM.

charlieR3: Maybe it's not clear, but `mockMember` sets the member info to null by default so when mocking `registeringMemberHoldingId`, no member info is created.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
charlieR3: Good point, thanks!

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
charlieR3: So I understand properly, do you mean change this so that instead of `!canMessage(sourceMemberInfo, destinationMemberInfo) || !canMessage(destinationMemberInfo, sourceMemberInfo)` it is `!canMessage(sourceMemberInfo, destinationMemberInfo)` and calling it once checks both to and from paths? So instead of it checking one direction and being called twice, we call it once and it checks both directions?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
williamvigorr3: I misread that as `party == null || (!party.isActive -> otherParty?.isMgm == true)` 🤦 .

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
williamvigorr3: Yes but maybe it is clearer as is.

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessorTest.kt
```diff
@@ -358,6 +371,63 @@ class InboundMessageProcessorTest {
                 .hasSize(1)
                 .contains("Received message with SessionId = Session for which there is no active session. The message was discarded.")
         }
+
+        @Test
+        fun `AuthenticatedDataMessage with Inbound session which fails membership messaging validation will drop messages`() {
```
williamvigorr3, line 376: Should we test an `Unathenticated` message fails validation.

charlieR3: Yeah, I'll add that, thanks

charlieR3: Added

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
charlieR3: I'll leave it as is then if that's OK

williamvigorr3 (dismissed):
LGTM

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
```diff
@@ -0,0 +1,144 @@
...
+import net.corda.membership.lib.MemberInfoExtension.Companion.isMgm
+import net.corda.membership.lib.MemberInfoExtension.Companion.status
+import net.corda.membership.read.MembershipGroupReaderProvider
+import net.corda.v5.base.types.MemberX500Name
+import net.corda.v5.membership.MemberInfo
+import net.corda.virtualnode.HoldingIdentity
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+class NetworkMessagingValidator(private val membershipGroupReaderProvider: MembershipGroupReaderProvider) {
```
yift-r3: internal?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
```diff
@@ -0,0 +1,144 @@
...
+     * If [party] is an MGM, they can message [otherParty] in any status.
+     * Otherwise, a party can message if [otherParty] is in active status.
+     */
+    private fun canMessage(party: MemberInfo?, otherParty: MemberInfo?) = when {
+        party?.isMgm == true -> true
+        party == null || !party.isActive -> otherParty?.isMgm == true
+        else -> otherParty?.isActive == true
+    }
+
+    private sealed class ValidationDirection {
```
yift-r3: This can be enum

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
```diff
@@ -0,0 +1,144 @@
...
+        object Inbound : ValidationDirection()
+        object Outbound : ValidationDirection()
+    }
+
+}
+
+
+sealed class NetworkStatusValidationResult {
```
yift-r3: We can use the `Either<Unit, String>` or just `String?` instead

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
```diff
@@ -0,0 +1,144 @@
...
+    ): T? = invokeIfValid(source, destination, ValidationDirection.Inbound, func)
+
+    /**
+     * Validates the input by calling [validate] and invokes [func] if validation succeeds.
+     * If validation fails, no exception is thrown and [func] is not invoked.
+     * Validation is performed based on the network member list visible to [source].
+     *
+     * @return result of [func] or null if validation failed.
+     */
+    fun <T> invokeIfValidOutbound(
```
yift-r3: Is this used?

yift-r3 (commented):
Maybe @williamvigorr3 should have another look. Nothing significant from my side

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
```diff
@@ -0,0 +1,144 @@
...
+        source: HoldingIdentity,
+        destination: HoldingIdentity
+    ) = validate(source, destination, ValidationDirection.Outbound)
+
+    private fun validate(
+        source: HoldingIdentity,
+        destination: HoldingIdentity,
+        validationDirection: ValidationDirection
+    ): NetworkStatusValidationResult {
+        val memberListViewOwner = when(validationDirection) {
```
williamvigorr3: This is an observation (nothing to do here). This makes a sense to me. If we are handling an inbound message then `destination` is hosted by a vnode in this Link Managers cluster, if we are handling an outbound message then `source`  is hosted by a vnode in this link managers cluster. So doing a lookup this way is correct and the code before 1a8e6d9 was wrong.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
```diff
@@ -128,10 +131,15 @@ internal class OutboundMessageProcessor(
...
         }
         return if (source.groupId != destination.groupId) {
-            return "group IDs do not match"
+            "group IDs do not match"
         } else if (!linkManagerHostingMap.isHostedLocally(cordaSource)) {
-            return "source ID is not locally hosted"
-        } else null
+            "source ID is not locally hosted"
+        } else {
+            when(val result = networkMessagingValidator.validateOutbound(source.toCorda(), destination.toCorda())) {
```
williamvigorr3: I think (after changes to the `NetworkMessagingValidator`) we should also do a `validateInbound` if we are doing a loop back (i.e. both the source and destination are locally hosted). 

Probably it is easiest to check the hosting map here and discard the message if the destination is locally hosted and `validateInbound` fails.

williamvigorr3 (commented):
I left a comment (should be quick to change, assuming you agree).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
charlieR3: Good catch. I must have just added it to mirror the inbound logic. I'll remove since it's currently not needed.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
charlieR3: Replaced with `Either<Unit, String>`

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
charlieR3: Added in https://github.com/corda/corda-runtime-os/pull/3555/commits/b9f3ad4e57ee8cd08cfda9ef6e255b7cd3c95c5a

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
charlieR3: Removed in https://github.com/corda/corda-runtime-os/pull/3555/commits/b9f3ad4e57ee8cd08cfda9ef6e255b7cd3c95c5a

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
charlieR3: Replaced in https://github.com/corda/corda-runtime-os/pull/3555/commits/b9f3ad4e57ee8cd08cfda9ef6e255b7cd3c95c5a

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
charlieR3: Updated in https://github.com/corda/corda-runtime-os/pull/3555/commits/b9f3ad4e57ee8cd08cfda9ef6e255b7cd3c95c5a

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
charlieR3: Updated in https://github.com/corda/corda-runtime-os/pull/3555/commits/b9f3ad4e57ee8cd08cfda9ef6e255b7cd3c95c5a

yift-r3 (dismissed):
LGTM

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidator.kt
williamvigorr3: This doesn't make the `InboundMessageProcessor` very easy to read.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
williamvigorr3: I think this needs to do the validation in both directions (for looped back messages). I realise what I said in the second paragraph was maybe a bit confusing. So we always do a `validateOutbound` and do an extra `validateInbound` if the destination is also locally hosted (as well as the source).

charlieR3: I think that should be addressed by https://github.com/corda/corda-runtime-os/pull/3555/commits/4a15bcd557ed6822fed2d99dbbb8f9334a90b40a now

williamvigorr3 (dismissed):
LGTM

williamvigorr3 (approved):
LGTM

dimosr (approved):

PR #3558, opened 2023-04-04
CORE-9564 - send error messages eagerly to counterparties when flow fails

pmck91:
send error messages eagerly to counterparties when flow fails


components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/helper/FailedRequestHelper.kt
```diff
@@ -0,0 +1,14 @@
+package net.corda.flow.pipeline.handlers.requests.helper
+
+import net.corda.data.flow.state.session.SessionStateType
+import net.corda.flow.pipeline.sessions.FlowSessionManager
+import net.corda.flow.state.FlowCheckpoint
+
+fun getSessionsToError(checkpoint: FlowCheckpoint, sessionIds: List<String>, flowSessionManager: FlowSessionManager): List<String> {
```
LWogan: makes sense to split this out to a new function. Might be better to add it to the flowSessionManager instead of a new utils class. That or change/add  api in FlowSessionManager to take multiple statuses 
```
fun getSessionsWithStatus(
        checkpoint: FlowCheckpoint,
        sessionIds: List<String>,
        status: Set<SessionStateType>
    ): List<SessionState>```

LWogan (approved):
LGTM

PR #3559, opened 2023-04-05
CORE-11706 CORE-11710 CORE-12341 Fix notary composite key threshold and use correct checking function

relyafi:
### Summary

This PR makes two changes to resolve issues with notary composite keys and checking:

- `NotaryInfoConverter` now correctly sets the threshold on the composite key for the notary service to 1 (addressing [CORE-11706](https://r3-cev.atlassian.net/browse/CORE-11706). This reflects the fact that a signature by any one of the notary virtual nodes constitutes a valid signature for the notary service. This behavior is not testable directly due to the current artificial restriction of a 1-1 mapping between notary virtual node and notary service, so a new integration test has been written to verify this behavior works as intended.
- Notary key checking now uses `isKeyFulfilledBy` instead of `isKeyInSet` (addressing [CORE-11710](https://r3-cev.atlassian.net/browse/CORE-11710)). This was always the correct API call to use, but was not possible due to the issue addressed in the first point.

As part of this change, the redundant `LayeredContextImpl` class that was only used by tests has also been removed (addressing [CORE-12341](https://r3-cev.atlassian.net/browse/CORE-12341))

### Testing

Existing automated test scenarios have been updated to reflect the change in behavior. New integration test written to verify the notary composite key behavior that is not currently reachable in a real system. All automated tests confirmed to pass.

[CORE-11706]: https://r3-cev.atlassian.net/browse/CORE-11706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CORE-11710]: https://r3-cev.atlassian.net/browse/CORE-11710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CORE-12341]: https://r3-cev.atlassian.net/browse/CORE-12341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImplTest.kt
+++ b/components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImplTest.kt
@@ -54,7 +54,7 @@ internal class UtxoSignedTransactionImplTest: UtxoLedgerTest() {
     }
 
     @Test
-    fun `verifyAttachedNotarySignature throws on unnotarized transaction`() {
+    fun `verifyNotarySignatureAttached throws on unnotarized transaction`() {
         Assertions.assertThatThrownBy { signedTransaction.verifyAttachedNotarySignature() }.isInstanceOf(
             TransactionSignatureException::class.java)
             .hasMessageContaining("There are no notary")
@@ -62,8 +62,7 @@ internal class UtxoSignedTransactionImplTest: UtxoLedgerTest() {
     }
 
     @Test
-    @Disabled("Composite key validation does not look correct at the moment.")
-    fun `verifyAttachedNotarySignature does not throw on notarized transaction`() {
+    fun `verifyNotarySignature does not throw on notarized transaction`() {
         val sig = getSignatureWithMetadataExample(notaryNode1PublicKey)
         signedTransaction = signedTransaction.addSignature(sig)
         assertDoesNotThrow {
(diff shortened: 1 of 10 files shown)


components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImplTest.kt
```diff
@@ -54,16 +54,15 @@ internal class UtxoSignedTransactionImplTest: UtxoLedgerTest() {
     }
 
     @Test
-    fun `verifyAttachedNotarySignature throws on unnotarized transaction`() {
+    fun `verifyNotarySignatureAttached throws on unnotarized transaction`() {
```
vlajos: The old test name looks the correct.

relyafi: Thanks - yeah this was a merge gone wrong, will fix.

vlajos (dismissed):
LGTM

libs/membership/membership-impl/build.gradle
```diff
@@ -5,6 +5,15 @@ plugins {
 
 description 'Membership implementation'
 
+kotlin {
+    target {
+        compilations.integrationTest {
+            // Needed to expose internal LayeredContextImpl class to this task
```
charlieR3: I'm not sure why `LayeredContextImpl` even exists to be honest. It might have been something we used in the past. 
Typically in tests we just create a context map that the `LayeredPropertyMapMocks` will accept which is basically:
```
private class TestContext(map: LayeredPropertyMap) : LayeredPropertyMap by map
```
It's a bit easier than changing the build to expose internal classes. I'll raise a ticket to remove `LayeredContextImpl` but it's not a high priority. We can either remove this then, or you can use a test context in this PR and remove this.

charlieR3: Right now it looks like `LayeredContextImpl` is only used by one other test so it doesn't make sense to keep it in the implementation package. (but again, not really your fault so I can raise a ticket for that)

charlieR3 (dismissed):

libs/membership/membership-impl/build.gradle
charlieR3: https://r3-cev.atlassian.net/browse/CORE-12341

blsemo (dismissed):
LGTM

libs/membership/membership-impl/build.gradle
relyafi: Removed as part of this PR as discussed verbally.

charlieR3 (approved):

PR #3560, opened 2023-04-05
Not for mering - verifying PR gate 

ronanbrowne:


Jenkinsfile
```diff
@@ -16,3 +16,11 @@ cordaPipeline(
     // allow publishing an installer to a download site
     publishToDownloadSiteTask: ':tools:plugins:publish',
     )
+
+
+//dummy data fake key purpusly triggering scan
```
ronanbrowne, line 21: fake data to verify if our security scanners pick up

PR #3562, opened 2023-04-05
CORE-12316: Upgrade to Kotlin 1.8.20.

chrisr3:
Upgrade Kotlin 1.8.10 -> 1.8.20.

Kotlin 1.8.20 and Mockito do not appear to want to play "nicely" with each other, for reasons unknown. But something seems to be breaking "equality matching" when mocks are involved.

Some of these mock-heavy tests - e.g. _Mock Hibernate?!_ - may be more useful if rewritten as integration tests with a _real_ database.

### DIFF
--- a/components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
+++ b/components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
@@ -18,6 +18,7 @@ import net.corda.virtualnode.toCorda
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 import org.mockito.kotlin.any
+import org.mockito.kotlin.anyVararg
 import org.mockito.kotlin.argumentCaptor
 import org.mockito.kotlin.doReturn
 import org.mockito.kotlin.eq
@@ -62,10 +63,9 @@ class QueryRegistrationRequestsHandlerTest {
     private val order = mock<Order>()
     private val query = mock<CriteriaQuery<RegistrationRequestEntity>> {
         on { from(RegistrationRequestEntity::class.java) } doReturn root
-        on { select(root) } doReturn mock
-        on { where() } doReturn mock
-        on { where(any()) } doReturn mock
-        on { orderBy(order) } doReturn mock
+        on { select(any()) } doReturn mock
+        on { where(anyVararg()) } doReturn mock
+        on { orderBy(anyVararg<Order>()) } doReturn mock
     }
     private val keyValuePairListDeserializer = mock<CordaAvroDeserializer<KeyValuePairList>> {
         on { deserialize(any()) } doReturn KeyValuePairList(emptyList())
@@ -104,9 +104,7 @@ class QueryRegistrationRequestsHandlerTest {
         on { jpaEntitiesRegistry } doReturn jpaEntitiesRegistry
         on { cordaAvroSerializationFactory } doReturn serializationFactory
     }
-    private val context = mock<MembershipRequestContext> {
-        on { holdingIdentity } doReturn holdingIdentity
-    }
+    private val context = MembershipRequestContext(Instant.ofEpochSecond(0), null, holdingIdentity)
 
     private val handler = QueryRegistrationRequestsHandler(service)
 
--- a/components/virtual-node/virtual-node-write-service-impl/build.gradle
+++ b/components/virtual-node/virtual-node-write-service-impl/build.gradle
@@ -52,4 +52,5 @@ dependencies {
 
     testRuntimeOnly "org.hsqldb:hsqldb:$hsqldbVersion"
     testRuntimeOnly "org.postgresql:postgresql:$postgresDriverVersion"
+    testRuntimeOnly "org.slf4j:slf4j-simple:$slf4jVersion"
 }
--- a/components/virtual-node/virtual-node-write-service-impl/src/test/kotlin/net/corda/virtualnode/write/db/impl/tests/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandlerTest.kt
+++ b/components/virtual-node/virtual-node-write-service-impl/src/test/kotlin/net/corda/virtualnode/write/db/impl/tests/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandlerTest.kt
@@ -47,18 +47,18 @@ class VirtualNodeUpgradeOperationHandlerTest {
     private val entityTransaction = mock<EntityTransaction>()
     private val em = mock<EntityManager>()
     private val entityManagerFactory = mock<EntityManagerFactory>()
-    private val migrationUtility = mock<MigrationUtility>() {
+    private val migrationUtility = mock<MigrationUtility> {
         whenever(it.areChangesetsDeployedOnVault(any(), any(), any())).thenReturn(false)
     }
     private val vnodeId = "123456789011"
 
     private val mockChangelog1 = mock<CpkDbChangeLog> { changelog ->
         whenever(changelog.id).thenReturn(CpkDbChangeLogIdentifier(SecureHashImpl("SHA-256","abc".toByteArray()), "cpk1"))
-        whenever(changelog.content).thenReturn( "dog.xml")
+        whenever(changelog.content).thenReturn("dog.xml")
     }
     private val mockChangelog2 = mock<CpkDbChangeLog> { changelog ->
         whenever(changelog.id).thenReturn(CpkDbChangeLogIdentifier(SecureHashImpl("SHA-256","abc".toByteArray()),"cpk1"))
-        whenever(changelog.content).thenReturn( "cat.xml")
+        whenever(changelog.content).thenReturn("cat.xml")
     }
     private val cpkDbChangelogs = listOf(mockChangelog1, mockChangelog2)
     private val mockCpkDbChangeLogRepository = mock<CpkDbChangeLogRepository> {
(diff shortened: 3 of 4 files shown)


components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
```diff
@@ -104,9 +107,7 @@ class QueryRegistrationRequestsHandlerTest {
         on { jpaEntitiesRegistry } doReturn jpaEntitiesRegistry
         on { cordaAvroSerializationFactory } doReturn serializationFactory
     }
-    private val context = mock<MembershipRequestContext> {
-        on { holdingIdentity } doReturn holdingIdentity
-    }
+    private val context = MembershipRequestContext(Instant.now(), null, holdingIdentity)
```
yift-r3: `Instant.now()` -> `Instant.ofEpochSecond(30)` We want the unit test to be deterministic.

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
```diff
@@ -62,10 +63,12 @@ class QueryRegistrationRequestsHandlerTest {
     private val order = mock<Order>()
     private val query = mock<CriteriaQuery<RegistrationRequestEntity>> {
         on { from(RegistrationRequestEntity::class.java) } doReturn root
-        on { select(root) } doReturn mock
-        on { where() } doReturn mock
-        on { where(any()) } doReturn mock
-        on { orderBy(order) } doReturn mock
+        on { select(any()) } doReturn mock
```
yift-r3, line 66: `select(root)` works with 1.8.20

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
```diff
@@ -62,10 +63,12 @@ class QueryRegistrationRequestsHandlerTest {
...
+        on { where(anyVararg()) } doReturn mock
+        // We shouldn't need to invoke anyVararg() multiple times.
+        // This looks like a Kotlin / Mockito bug to me!
+        on { where(anyVararg(), anyVararg()) } doReturn mock
+        on { orderBy(anyVararg<Order>()) } doReturn mock
```
yift-r3, line 68: `on { orderBy(order) } doReturn mock` works

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
```diff
@@ -62,10 +63,12 @@ class QueryRegistrationRequestsHandlerTest {
...
         on { from(RegistrationRequestEntity::class.java) } doReturn root
-        on { select(root) } doReturn mock
-        on { where() } doReturn mock
-        on { where(any()) } doReturn mock
-        on { orderBy(order) } doReturn mock
+        on { select(any()) } doReturn mock
+        on { where(anyVararg()) } doReturn mock
+        // We shouldn't need to invoke anyVararg() multiple times.
+        // This looks like a Kotlin / Mockito bug to me!
+        on { where(anyVararg(), anyVararg()) } doReturn mock
```
yift-r3: I think we can remove this line (the one above it covered it and the tests are passing without it)

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
```diff
@@ -165,7 +166,7 @@ class QueryRegistrationRequestsHandlerTest {
     fun `invoke queries with correct predicates if X500 name specified in request`() {
         whenever(actualQuery.resultList).doReturn(emptyList())
         val captor = argumentCaptor<Predicate>()
-        whenever(query.where(captor.capture())).thenReturn(query)
+        whenever(query.where(captor.capture(), captor.capture())).thenReturn(query)
```
yift-r3: Do we need to duplicate the capture? The tests pass when we capture the predicate only once.

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
chrisr3: Do you actually _care_ what the timestamp value is? You don't appear to be using it.

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
chrisr3: Maybe it does, but Kotlin 1.8.20 and Mockito aren't playing nicely together. There seems to be an issue with "equality matching" involving mocks, otherwise I wouldn't have touched this test in the first place. I think using `any()` is safer.

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
chrisr3: Really? Have you tried?

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
chrisr3: Yes. Kotlin 1.8.20 and Mockito aren't playing nicely with each other, hence this change in the first place.

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
yift-r3: `any` will make the test pass if we change the code to select another root, making the test less valuable.  Since root works, we should use it. (If not, we should use `eq(root)` if that won't work, we should use  `argThat { it == root }`).

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
yift-r3: I have.

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
yift-r3: Because I don't care, 30 should be good as now or any other number, and `now` will make the test non-deterministic...

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
yift-r3: I understand the cause. But if it works without the duplication (and on my machine it did) why not remove the duplication?

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
chrisr3: No. The `CriteriaQuery.select()` API should _never_ return `null` - see [here](https://docs.oracle.com/javaee/7/api/javax/persistence/criteria/CriteriaQuery.html#select-javax.persistence.criteria.Selection-), which says that it always returns the modified query. Do you think Hibernate would ever return `null` here?

With `select(root)`, if someone were to pass a different `root` then this test just throws an NPE. And speaking as someone who spent _all day_ yesterday wondering why this test was throwing NPEs _for no good reason_, and who ended up _single-stepping through Mockito code_ to understand what was going on, I can state *categorically* that an NPE is a hideously unhelpful and uninformative kind of "failure". There was no "value" in that experience whatsoever.

```kotlin
on { select(any()) } doReturn mock
```
is a faithful generic mock for the `CriteriaQuery.select()` fluent API. If `query` were intended to be anything other than a generic `CriteriaQuery` mock then it would do more than just return the `mock` object each time.

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
chrisr3: OK, I was originally working with Mockito 5.2.0 here, but subsequently reverted to Mockito 4.11.0 due to other problems. This _particular_ garbage must have been required by Mockito 5.2.0, so I am _delighted_ to remove it.

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
chrisr3: Maybe it does in your _particular_ case, but actual the fluent API is:
```java
CriteriaQuery<T> orderBy(Order... o)
```
which means that `anyVararg<Order>()` is correct. See my description of the time _wasted_ yesterday debugging NPEs from APIs which should _never_ return `null` under _any_ circumstances.

---

It is probably better to use:
```kotlin
verify(query).orderBy(order)
```
to guarantee that `orderBy()` receives `order` as its parameter.

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
yift-r3: It's a unit test. We are only testing that our code is working the way we expect it to work. The integration with Hibernate is tested in the integration test. 

If we will change the test to always return a mock instead of only when we call it with the correct root, and then we will change the implementation to call with another root, the test must fail. Otherwise, it will be a lot less valuable. So, yes, NPE is better than no failure.

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
chrisr3: This is clearly another relic from my aborted attempt to upgrade to Mockito 5.2.0, so it can go too.

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
chrisr3: I spent most of yesterday debugging inexplicable NPEs from this test, so _believe me_ when I say that there **was no value whatsoever in that experience!**

I will spare anybody else from going through that again.

yift-r3: I disagree, but it's not important enough to argue about.

yift-r3 (approved):

JamesHR3 (approved):

mcgovc (approved):
LGTM - build changes

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
chrisr3: I'm not a Mockito expert, but it seems to me that if you want to verify that `query.select()` is invoked exactly once with argument `root` then you would use:
```kotlin
verify(query).select(root)
```
And seeing that (some of) these tests do _indeed_ verify the `query` mock in this way, this should already address any concerns you may have about `query.select()` ever being invoked with a different root. Or maybe even use:
```kotlin
verify(query, never()).select(not(eq(root)))
````
if you want to be _extremely_ paranoid. But regardless, a failed `verify()` will give a far more meaningful error than some random NPE.

PR #3565, opened 2023-04-05
CORE-10576 Trigger member list distribution when suspending and activating a member

williamvigorr3:
Trigger member list distribution from the rest worker after updating the database when suspending or activating a member. Additionally change the db transactions to remove and add a notary from the group parameters if a member is suspended/activated. Also changed the distribution protocol inside DistributeMemberInfoAction to only send their own member info to a suspended member (so information is not leaked about the state of the network).  Also changed the `Member` and MGM side of the Synchronisation service to reflect this. Requires API changes: https://github.com/corda/corda-api/pull/1041


libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/InternalGroupParametersUtils.kt
```diff
@@ -0,0 +1,31 @@
...
+import java.nio.ByteBuffer
+
+fun InternalGroupParameters.toPersistentGroupParameters(
+    owner: HoldingIdentity,
+    keyEncodingService: KeyEncodingService
+): PersistentGroupParameters {
+    val signed = this as? SignedGroupParameters
+    return PersistentGroupParameters(
+        owner.toAvro(),
+        net.corda.data.membership.SignedGroupParameters(
```
nikinagy: Maybe we can introduce a type alias for this and call it "SignedGroupParametersAvro" to make the code a bit prettier? But it's just an idea, we can leave it as is if we prefer that way

components/membership/membership-persistence-client/src/main/kotlin/net/corda/membership/persistence/client/MembershipPersistenceClient.kt
```diff
@@ -283,7 +283,7 @@ interface MembershipPersistenceClient : Lifecycle {
         memberX500Name: MemberX500Name,
         serialNumber: Long?,
         reason: String?,
-    ): MembershipPersistenceResult<PersistentMemberInfo>
+    ): MembershipPersistenceResult<Pair<PersistentMemberInfo, InternalGroupParameters?>>
```
nikinagy: Modify the kdocs's return part as well

components/membership/membership-persistence-client/src/main/kotlin/net/corda/membership/persistence/client/MembershipPersistenceClient.kt
```diff
@@ -300,7 +300,7 @@ interface MembershipPersistenceClient : Lifecycle {
...
```
nikinagy: Modify the kdocs's return part as well

components/membership/membership-client-impl/src/main/kotlin/net/corda/membership/impl/client/MGMResourceClientImpl.kt
```diff
@@ -573,17 +588,38 @@ class MGMResourceClientImpl @Activate constructor(
             return mgm to memberShortHash
         }
 
-        private fun publishMemberInfo(memberInfo: PersistentMemberInfo, mgm: String, member: String) {
+        private fun publishMemberInfoAndRequestDistribution(
```
nikinagy: I am not a fan of the name as here we can publish group params too, a bit misleading.

components/membership/membership-client-impl/src/main/kotlin/net/corda/membership/impl/client/MGMResourceClientImpl.kt
```diff
@@ -573,17 +588,38 @@ class MGMResourceClientImpl @Activate constructor(
...
+            memberInfo: PersistentMemberInfo,
+            groupParameters: InternalGroupParameters?,
+            memberX500Name: MemberX500Name,
+            memberShortHash: String,
+            mgmHoldingIdentity: HoldingIdentity,
+            mgmShortHash: String,
+        ) {
+            val serialNumber = memberInfoFactory.create(memberInfo).serial
+            val publisher = coordinator.getManagedResource<Publisher>(PUBLISHER_RESOURCE_NAME)
```
nikinagy, line 604: is there a good reason why we moved these out from the try-catch?

nikinagy: ok don't mind this comment I think its because we are just catching CordaMessageAPIIntermittentException anyway.

components/membership/membership-client-impl/src/test/kotlin/net/corda/membership/impl/client/MGMResourceClientTest.kt
```diff
@@ -1578,7 +1618,7 @@ class MGMResourceClientTest {
             whenever(publisher.publish(any())).doReturn(listOf(mockFuture))
             val mockMemberInfo = mock<PersistentMemberInfo>()
             whenever(membershipPersistenceClient.activateMember(eq(holdingIdentity), eq(memberName), any(), any()))
-                .doReturn(MembershipPersistenceResult.Success(mockMemberInfo))
+                .doReturn(MembershipPersistenceResult.Success(mockMemberInfo to null))
```
nikinagy: Can we add a test where the group params won't be null?

components/membership/membership-persistence-client-impl/src/test/kotlin/net/corda/membership/impl/persistence/client/MembershipPersistenceClientImplTest.kt
```diff
@@ -1182,7 +1182,7 @@ class MembershipPersistenceClientImplTest {
 
         @Test
         fun `suspendMember returns the correct result`() {
-            mockPersistenceResponse(SuspendMemberResponse(mock()))
+            mockPersistenceResponse(SuspendMemberResponse(mock(), mock()))
```
nikinagy: maybe verify that groupParamatersFactory.create has been called?

components/membership/membership-persistence-service-impl/src/integrationTest/kotlin/net/corda/membership/impl/persistence/service/MembershipPersistenceTest.kt
```diff
@@ -1404,7 +1404,7 @@ class MembershipPersistenceTest {
 
         val suspended1 = membershipPersistenceClientWrapper.activateMember(
             viewOwningHoldingIdentity, member1, 1L, "test-reason"
-        ).getOrThrow()
+        ).getOrThrow().first
```
nikinagy, line 1465: can we have a test for notary too? and see what is returned with the group params

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/ActivateMemberHandler.kt
```diff
@@ -1,24 +1,59 @@
...
-
     override fun invoke(context: MembershipRequestContext, request: ActivateMember): ActivateMemberResponse {
-        val updatedMemberInfo = changeMemberStatus(
-            context,
-            request.activatedMember,
-            request.serialNumber,
-            MEMBER_STATUS_SUSPENDED,
-            MEMBER_STATUS_ACTIVE
-        )
+        val (updatedMemberInfo, updatedGroupParameters) = transaction(context.holdingIdentity.toCorda().shortHash) { em ->
```
nikinagy, line 38: Just a nit, but adding log messages to handlers would make life easier in the long run if something goes wrong.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/SuspendMemberHandler.kt
```diff
@@ -1,24 +1,140 @@
...
+                em,
+                request.suspendedMember,
+                context.holdingIdentity.groupId,
+                request.serialNumber,
+                MEMBER_STATUS_ACTIVE
+            )
+            val currentMgmContext = keyValuePairListDeserializer.deserialize(currentMemberInfo.mgmContext)
+                ?: throw MembershipPersistenceException("Failed to deserialize the MGM-provided context.")
+            PersistentMemberInfo(
+                HoldingIdentity(request.suspendedMember, context.holdingIdentity.groupId),
```
nikinagy: I am a bit lost here: why the view owning member is the suspended member? Same goes in ActivateMemberHandler. What are we even doing with these PersistentMemberInfos here? it seems unused

nikinagy: I think here and in the ActivateMemberHandler these were left by mistake

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/ActivateMemberHandlerTest.kt
```diff
@@ -87,12 +89,20 @@ class ActivateMemberHandlerTest {
             createAvroSerializer<KeyValuePairList>(any())
         } doReturn keyValuePairListSerializer
     }
+    private val memberInfo = mock<MemberInfo> {
```
nikinagy, line 92: test with notary too

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/SuspendMemberHandlerTest.kt
```diff
@@ -87,12 +89,21 @@ class SuspendMemberHandlerTest {
             createAvroSerializer<KeyValuePairList>(any())
         } doReturn keyValuePairListSerializer
     }
+    private val memberInfo = mock<MemberInfo> {
```
nikinagy, line 92: test with notary too and always check the test coverage on unit tests

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoActionHandler.kt
```diff
@@ -110,12 +110,20 @@ class DistributeMemberInfoActionHandler(
             }
         }
 
+        val allActiveMembers = groupReader.lookup()
+        val allActiveMembersExcludingMgm = allActiveMembers.filterNot { it.isMgm }
+        //If the updated member is suspended then we only send it's own member info to itself (so it can tell it has been suspended).
+        val membersToDistributeToUpdatedMember = if (updatedMemberInfo.status == MEMBER_STATUS_SUSPENDED) {
+            listOf(updatedMemberInfo)
```
nikinagy, line 117: I am a bit a worried if this will work with the hash checks in place. I think the suspended member will keep asking for syncs from MGM, because the hash is not matching the expected hash.

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/GroupParametersNotaryUpdater.kt
```diff
@@ -0,0 +1,209 @@
...
+        const val NOTARY_SERVICE_KEYS_PREFIX = "$NOTARY_SERVICE_KEY_PREFIX.keys"
+        const val NOTARY_SERVICE_KEYS_KEY = "$NOTARY_SERVICE_KEYS_PREFIX.%s"
+
+        const val EPOCH_KEY = "corda.epoch"
+        const val MODIFIED_TIME_KEY = "corda.modifiedTime"
+        val notaryServiceRegex = NOTARY_SERVICE_NAME_KEY.format("([0-9]+)").toRegex()
+    }
+    private val logger = LoggerFactory.getLogger(this::class.java)
+
+    fun addNewNotaryService(
```
charlieR3, line 31: Should `addNewNotaryService` be removed from `GroupParametersUtils` now after being moved here?

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/GroupParametersNotaryUpdater.kt
```diff
@@ -0,0 +1,209 @@
...
+            filterNot { listOf(EPOCH_KEY, MODIFIED_TIME_KEY).contains(it.key) }
+                .map { KeyValuePair(it.key, it.value) } + listOf(
+                KeyValuePair(EPOCH_KEY, newEpoch.toString()),
+                KeyValuePair(MODIFIED_TIME_KEY, clock.instant().toString())
+            )
+        }
+        return newEpoch to KeyValuePairList(parametersWithUpdatedEpoch + newService)
+    }
+
+    fun updateExistingNotaryService(
```
charlieR3, line 73: Same here, should `updateExistingNotaryService` be removed from `GroupParametersUtils` now after being moved here?

libs/membership/membership-common/src/test/kotlin/net/corda/membership/lib/GroupParametersNotaryUpdaterTest.kt
```diff
@@ -0,0 +1,247 @@
...
+            KeyValuePair(String.format(NOTARY_SERVICE_KEYS_KEY, 5, 0), NOTARY_KEY_A),
+            KeyValuePair(String.format(NOTARY_SERVICE_KEYS_KEY, 5, 1), "test-key"),
+            KeyValuePair(String.format(NOTARY_SERVICE_NAME_KEY, 5), NOTARY_SERVICE_A),
+            KeyValuePair(String.format(NOTARY_SERVICE_PROTOCOL_KEY, 5), NOTARY_PROTOCOL_A),
+            KeyValuePair(String.format(NOTARY_SERVICE_PROTOCOL_VERSIONS_KEY, 5, 0), "1"),
+            KeyValuePair(String.format(NOTARY_SERVICE_PROTOCOL_VERSIONS_KEY, 5, 1), "3"),
+        )
+    }
+
+    //Todo test exceptions.
```
charlieR3, line 109: TODO comment. Does it need to be removed or are there tests remaining to add?

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/MGMRestResource.kt
```diff
@@ -461,9 +461,9 @@ interface MGMRestResource : RestResource {
      * @param holdingIdentityShortHash The holding identity ID of the MGM of the membership group.
      * @param activationParams Parameters for activating a member. See [SuspensionActivationParameters] for more details.
      */
-//    @HttpPOST(
-//        path = "{holdingIdentityShortHash}/activate"
-//    )
+    @HttpPOST(
```
charlieR3, line 464: Does this need an update to swaggerBaseline?

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/BaseSuspensionActivationHandler.kt
```diff
@@ -14,25 +15,96 @@ import net.corda.membership.lib.MemberInfoExtension.Companion.STATUS
...
+            LockModeType.PESSIMISTIC_WRITE
+        ) ?: throw MembershipPersistenceException("Member '$memberName' does not exist.")
+        expectedSerial?.let {
+            require(member.serialNumber == it) {
+                throw InvalidEntityUpdateException(
+                    "The provided serial number '$expectedSerial' does not match the current version '${member.serialNumber}' of " +
+                        "MemberInfo for member '$memberName'.")
+            }
+        }
+        require(member.status == expectedStatus) {
```
charlieR3: Maybe if the member is already in the expected status we can just skip the update instead of throwing an exception

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/SuspendMemberHandler.kt
```diff
@@ -1,24 +1,140 @@
...
+            val currentMemberInfo = findMember(
+                em,
+                request.suspendedMember,
+                context.holdingIdentity.groupId,
+                request.serialNumber,
+                MEMBER_STATUS_ACTIVE
+            )
+            val currentMgmContext = keyValuePairListDeserializer.deserialize(currentMemberInfo.mgmContext)
+                ?: throw MembershipPersistenceException("Failed to deserialize the MGM-provided context.")
+            PersistentMemberInfo(
```
charlieR3: Unused?

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/SuspendMemberHandler.kt
```diff
@@ -1,24 +1,140 @@
...
+
+            val updatedMemberInfo = updateStatus(
+                em,
+                request.suspendedMember,
+                context.holdingIdentity,
+                currentMemberInfo,
+                currentMgmContext,
+                MEMBER_STATUS_SUSPENDED
+            )
+            val updatedGroupParameters = if (memberInfoFactory.create(updatedMemberInfo).notaryDetails != null) {
```
charlieR3: There's an `isNotary` function is MemberInfoExtension which could be used to check if it's a notary without the additional notary details parsing if the notary details aren't needed at this point.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/SuspendMemberHandler.kt
```diff
@@ -1,24 +1,140 @@
...
+            .resultList.map {
+                memberInfoFactory.create(
+                    keyValuePairListDeserializer.deserializeKeyValuePairList(it.memberContext).toSortedMap(),
+                    keyValuePairListDeserializer.deserializeKeyValuePairList(it.mgmContext).toSortedMap(),
+                )
+            }
+        val otherMembersOfSameNotaryService = members.filter {
+            it.notaryDetails?.serviceName.toString() == notaryServiceName &&
+                    it.name != notaryInfo.name &&
+                    it.status == MEMBER_STATUS_ACTIVE
```
charlieR3: Could the status be part of the DB query for member infos to reduce the number of member infos that need to be parsed?

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/SuspendMemberHandler.kt
```diff
@@ -1,24 +1,140 @@
...
+        }
+        if (notaryServiceNumber == null) {
+            logger.warn("Notary `$holdingIdentity` has not updated the group parameters ")
+            return null
+        }
+        val memberQueryBuilder = criteriaBuilder.createQuery(MemberInfoEntity::class.java)
+        val memberQuery = memberQueryBuilder.select(memberQueryBuilder.from(MemberInfoEntity::class.java))
+        val members = em.createQuery(memberQuery)
+            .setLockMode(LockModeType.PESSIMISTIC_WRITE)
+            .resultList.map {
```
charlieR3, line 127: Maybe something to look at during large network support (so I don't mind not blocking the PR on this) but `resultStream` could be better for querying the entire member list if it's a very large list.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/SuspendMemberHandler.kt
charlieR3: Maybe even the name != notaryInfo.name could be added to the DB query. That way we could probably have `.mapNotNull { it.notaryDetails }.filter {serviceName.toString() == notaryServiceName }`

charlieR3: These are all kind of optimizations thinking ahead to large network support and trying to reduce the amount of times we need to parse `notaryDetails`

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
```diff
@@ -52,84 +56,77 @@ internal class AddNotaryToGroupParametersHandler(
...
+        val notaryServiceNumber = parametersMap.entries.firstOrNull { it.value == notaryServiceName }?.run {
+            notaryServiceRegex.find(key)?.groups?.get(1)?.value?.toIntOrNull()
+        }
+        val (epoch, groupParameters) = if (notaryServiceNumber != null) {
+            // Add notary to existing notary service, or update notary with rotated keys
+            val memberQueryBuilder = criteriaBuilder.createQuery(MemberInfoEntity::class.java)
+            val memberQuery = memberQueryBuilder.select(memberQueryBuilder.from(MemberInfoEntity::class.java))
+            val members = em.createQuery(memberQuery)
+                .setLockMode(LockModeType.PESSIMISTIC_WRITE)
+                .resultList.map {
```
charlieR3, line 98: Same comment here in that we could stream the results of querying the full member list, which could be quite large. But I'm willing to cover it later if necessary.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/ActivateMemberHandler.kt
```diff
@@ -1,24 +1,59 @@
...
+            val currentMemberInfo = findMember(
+                em,
+                request.activatedMember,
+                context.holdingIdentity.groupId,
+                request.serialNumber,
+                MEMBER_STATUS_SUSPENDED
+            )
+            val currentMgmContext = keyValuePairListDeserializer.deserialize(currentMemberInfo.mgmContext)
+                ?: throw MembershipPersistenceException("Failed to deserialize the MGM-provided context.")
+            PersistentMemberInfo(
```
charlieR3: Unused?

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/ActivateMemberHandler.kt
```diff
@@ -1,24 +1,59 @@
...
+
+            val updatedMemberInfo = updateStatus(
+                em,
+                request.activatedMember,
+                context.holdingIdentity,
+                currentMemberInfo,
+                currentMgmContext,
+                MEMBER_STATUS_ACTIVE
+            )
+            val updatedGroupParameters = if (memberInfoFactory.create(updatedMemberInfo).notaryDetails != null) {
```
charlieR3:
```suggestion
            val updatedGroupParameters = if (memberInfoFactory.create(updatedMemberInfo).isNotary()) {
```

components/membership/membership-client-impl/src/main/kotlin/net/corda/membership/impl/client/MGMResourceClientImpl.kt
```diff
@@ -573,17 +588,38 @@ class MGMResourceClientImpl @Activate constructor(
...
+            val recordForGroupParameters = groupParameters?.let {
```
charlieR3, line 605: This seems incorrect. Is this publishing the group parameters for another member? Other members should receive the updated group parameters through the data distribution protocol and the MGM shouldn't attempt to write the members view of the group parameters to kafka (especially since they probably won't be on the same cluster)

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/GroupParametersNotaryUpdater.kt
williamvigorr3: I am going to delete the `GroupParametersUtils`.

libs/membership/membership-common/src/test/kotlin/net/corda/membership/lib/GroupParametersNotaryUpdaterTest.kt
williamvigorr3: I need to add the tests and probably rework the `AddNotaryToGroupParametersHandler` to be a purer unit test (coverage of the exceptions will move from their to here).

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
williamvigorr3: I was thinking about this I am not sure we need to query all the members here (I decided not to change the behaviour as a part of this PR). As we want to union the set of notary versions for each member of the notary service. We could just union with the set of versions already in the group parameters, each time we update it (as union is associative), instead of recomputing the whole thing. Unless I missed something.

The same is not true for the `SuspendMemberHandler` (we need to recalculate the union of all the other not suspended members if we suspend a member of the notary service).

components/membership/membership-client-impl/src/main/kotlin/net/corda/membership/impl/client/MGMResourceClientImpl.kt
williamvigorr3: Good spot. This should publish the MGM's view of the group parameters not the members.

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoActionHandler.kt
williamvigorr3: 😟

components/membership/membership-client-impl/src/main/kotlin/net/corda/membership/impl/client/MGMResourceClientImpl.kt
charlieR3: You might be able to use `GroupParametersWriterService` for that if it reduces duplication

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
charlieR3: That's a good point. I think that could work for adding a notary but not removing as you said

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/MGMRestResource.kt
williamvigorr3: Yes I forgot to do that.

components/membership/membership-client-impl/src/main/kotlin/net/corda/membership/impl/client/MGMResourceClientImpl.kt
williamvigorr3: I renamed this `publishSuspensionActivationRecords`.

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/InternalGroupParametersUtils.kt
williamvigorr3: I imported it as `AvroSignedGroupParameters`.

components/membership/membership-client-impl/src/main/kotlin/net/corda/membership/impl/client/MGMResourceClientImpl.kt
williamvigorr3: Yes.

components/membership/membership-client-impl/src/test/kotlin/net/corda/membership/impl/client/MGMResourceClientTest.kt
williamvigorr3: I have added `suspendMember should publish the updated group parameters, the updated member info and request distribution` and `activateMember should publish the updated group parameters, the updated member info and request distribution`.

components/membership/membership-persistence-client-impl/src/test/kotlin/net/corda/membership/impl/persistence/client/MembershipPersistenceClientImplTest.kt
williamvigorr3: I split this test (and the same for activateMember) into two cases (one where the group parameters is null, one where it is not). I added the verifies as suggested.

components/membership/membership-client-impl/src/main/kotlin/net/corda/membership/impl/client/MGMResourceClientImpl.kt
williamvigorr3: I fixed this. I didn't use the `GroupParametersWriterService` as so we can ignore errors if writing to Kafka failed.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/SuspendMemberHandler.kt
williamvigorr3: Ideally we could check if the member was a notary in the query (but we can't because the member context has been serialised). 

I think even with the optimisations you have suggested we could potentially be querying most of the member list here in ActivateMamberHandler and in AddNotaryToGroupParametersHandler (most members will likely be `ACTIVE` and `name != notaryInfo.name` will reduce the number of members by one).

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/SuspendMemberHandler.kt
williamvigorr3: Yes I deleted this now.

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoActionHandler.kt
nikinagy: I am still worried this won't work as we imagined. Did we double-checked this?

williamvigorr3: I am going to do that now (when I run the e2e tests).

nikinagy: I think it will settle eventually by the synchronisation protocol due to the fact the hashes won't match when the member receives the data package. So it will just do an extra round of sync.

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
```diff
@@ -333,9 +336,15 @@ class MemberSynchronisationServiceImpl internal constructor(
                         mgm,
                     )
                 } else {
-                    val knownMembers = groupReader.lookup().filter { !it.isMgm }.associateBy { it.id }
-                    val allMembers = knownMembers + updateMembersInfo
-                    val expectedHash = merkleTreeGenerator.generateTree(allMembers.values).root
+                    val knownMembers = groupReader.lookup(MembershipStatusFilter.ACTIVE_OR_SUSPENDED)
+                        .filter { !it.isMgm }.associateBy { it.id }
+                    val updatedViewOwner = updateMembersInfo[viewOwningMember.shortHash.value]
```
dimosr, line 341: Is it safer to perform a lookup on `knownMembers` as well if not found in `updateMembersInfo` to capture cases where we might be sent other members info, but we might still be suspended?

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MgmSynchronisationServiceImpl.kt
```diff
@@ -213,16 +216,20 @@ class MgmSynchronisationServiceImpl internal constructor(
             val mgmInfo = groupReader.lookup(mgmName)
                 ?: throw CordaRuntimeException("MGM $mgmName $IDENTITY_EX_MESSAGE")
             val requesterName = MemberX500Name.parse(requester.x500Name)
-            val requesterInfo = groupReader.lookup(requesterName)
+            val requesterInfo = groupReader.lookup(requesterName, filter = MembershipStatusFilter.ACTIVE_OR_SUSPENDED)
                 ?: throw CordaRuntimeException("Requester $requesterName $IDENTITY_EX_MESSAGE")
             // we don't want to include the MGM in the data package since MGM information comes from the group policy
-            val allMembers = groupReader.lookup().filterNot { it.holdingIdentity == mgm.toCorda() }
+            val allMembersActiveMemberExcludingMgm = groupReader.lookup().filterNot { it.holdingIdentity == mgm.toCorda() }
```
dimosr: Don't we want to send both ACTIVE and SUSPENDED members here now? Seems like this lookup will only return ACTIVE ones by default.

williamvigorr3:
List of things I think needs to be done (in my opinion):
1. Add coverage of notary suspension/activation to `SuspendMemberHandlerTest` and `ActivateMemberHandlerTest`.
2. Add more test coverage to MembershipPersistenceTest (there is some coverage of notary suspension I could add more (although these tests are expensive and this will be covered by unit tests) I need to test notary activations.
3. Unit test changes to the MgmSynchronisation service.

Things which could be done in a follow up (in my opinion):
4. Tidy SuspensionActivationEntityOperationsTest (assuming there is enough coverage here).
5. Move coverage of exceptions from the  `AddNotaryToGroupParameterHandlerTest` to the `GroupParametersNotaryUpdaterTest`. To make purer unit tests.

dimosr (commented):
I only had a look at the changes in the sync process, posting a couple of questions

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
charlieR3: Posted on slack, but will comment here too for visibility. I think we don't have a case currently where a member should be sent other member infos while still suspended so I would say the way it's implemented here is OK.

charlieR3: Although, thinking again, double checking against the member list would protect against messages outside of what is expected like replays or some mistake on the MGM side so I'm thinking now it could be wise to check against known members also for our own suspended status.

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/SuspensionActivationEntityOperationsTest.kt
```diff
@@ -0,0 +1,382 @@
...
+        whenever(keyValuePairListDeserializer.deserialize(contextBytes)).doReturn(null)
+
+        assertThrows<MembershipPersistenceException> {
+            handler.updateStatus(em, knownX500Name.toString(), mock { on {groupId} doReturn knownGroupId }, mockEntity, mock(), "")
+        }.apply {
+            assertThat(this.message).contains("Failed to deserialize")
+        }
+    }
+
+//    @Test
```
charlieR3: Tests should be uncommented or removed before merging

charlieR3 (commented):
No additional comments (other than one minor one about commented tests)

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/SuspendMemberHandler.kt
williamvigorr3: I remove this.

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MgmSynchronisationServiceImpl.kt
williamvigorr3: Yes. I did this in b0391ac.

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/SuspensionActivationEntityOperationsTest.kt
williamvigorr3: I will remove this for now, we need to cover some of this in `SuspendMemberHandlerTest` / `ActivateMemberHandlerTest`.

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
dimosr: [Discussed offline with everyone and agreed to do that change, but as a follow-up along with all the other pending comments in order to get this functionality in for the next RC/HC]

dimosr (dismissed):

nikinagy (dismissed):
LGTM

charlieR3 (dismissed):

processors/rest-processor/src/integrationTest/resources/swaggerBaseline.json
```diff
@@ -5088,6 +5170,29 @@
...
+            "nullable" : true,
+            "example" : 0
+          },
+          "x500Name" : {
+            "type" : "string",
+            "nullable" : false,
+            "example" : "string"
+          }
+        },
+        "description" : "Parameters for activating a suspended member."
```
vkolomeyko:
```suggestion
        "description" : "Parameters for activating or suspending a member."
```

vkolomeyko (dismissed):
LGTM

Minor comment posted

processors/rest-processor/src/integrationTest/resources/swaggerBaseline.json
williamvigorr3: I will fix that (once I get the merge lock).

williamvigorr3: I change it too `Parameters for suspending or activating a member.` to be consistent with the KDOC on `SuspensionActivationParameters`.

dimosr (approved):

vkolomeyko (approved):
LGTM

PR #3566, opened 2023-04-05
CORE-12169: Configuration service should be up only when it's ready

yift-r3:
1. The `ConfigReadServiceEventHandler` follows two subscriptions; it should be marked as up only when the last one is UP.
2. Increase timeouts of some of the tests.

### DIFF
--- a/components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/ConfigReadServiceEventHandler.kt
+++ b/components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/ConfigReadServiceEventHandler.kt
@@ -106,9 +106,11 @@ internal class ConfigReadServiceEventHandler(
             }
 
             is RegistrationStatusChangeEvent -> {
-                // Only registration is on the subscription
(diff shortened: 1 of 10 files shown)


components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/ConfigReadServiceEventHandler.kt
```diff
@@ -106,9 +106,11 @@ internal class ConfigReadServiceEventHandler(
             }
 
             is RegistrationStatusChangeEvent -> {
-                // Only registration is on the subscription
+                // Only registration is on the subscriptions
```
charlieR3: That sounds like there's one registration covering multiple subscriptions
```suggestion
                // Only set LifecycleStatus.UP for coordinator after the config subscription is UP
```

charlieR3 (dismissed):
One nit-picky comment on the wording of the comment but looks good

JamesHR3 (approved):

charlieR3 (approved):

PR #3567, opened 2023-04-05
CORE-12178 - Document rules for X.500 names

dimosr:
Documenting current platform rules for X.500 names of virtual nodes.

Rules were retrieved from the `MemberX500Name` type.

### DIFF
--- a/wiki/X500-Name-rules.md
+++ b/wiki/X500-Name-rules.md
@@ -2,12 +2,13 @@
 
 Each virtual node within an application network is uniquely identified by its X.500 distinguished name. 
 The rules for a valid X.500 name in Corda are the following:
+* The string specified must use the grammar defined in RFC 1779 or RFC 2253.
 * The only supported attributes are CN, OU, O, L, ST, C.
 * Attributes cannot be duplicated and must have a single value.
 * The attributes O, L, C are mandatory.
 * The Organization attribute (O) cannot be blank and must be less than 128 characters.
 * The Locality attribute (L) cannot be blank and must be less than 64 characters.
-* The Country attribute (C) cannot be blank and must be an ISO 3166-1 2-letter country code.
+* The Country attribute (C) cannot be blank and must be an ISO 3166-1 2-letter country code or "ZZ" to indicate unspecified country.
 * If specified, the State attribute (ST) cannot be blank and must be less than 64 characters.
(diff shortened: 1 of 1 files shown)


wiki/X500-Name-rules.md
```diff
@@ -0,0 +1,13 @@
+# Rules for X500 names of virtual nodes
+
+Each virtual node within an application network is uniquely identified by its X.500 distinguished name. 
+The rules for a valid X.500 name in Corda are the following:
+* The only supported attributes are CN, OU, O, L, ST, C.
+* Attributes cannot be duplicated and must have a single value.
+* The attributes O, L, C are mandatory.
+* The Organization attribute (O) cannot be blank and must be less than 128 characters.
+* The Locality attribute (L) cannot be blank and must be less than 64 characters.
+* The Country attribute (C) cannot be blank and must be an ISO 3166-1 2-letter country code.
```
charlieR3: I notice we can also allow `private static final String UNSPECIFIED_COUNTRY = "ZZ";` as an option for country code. I'm not sure what it's usage is for though and if we want to document that.

charlieR3 (commented):
LGTM.

One question on the PR and one more here:
Should we also mention that it must be a valid `javax.security.auth.x500.X500Principal` since we parse it as one to get the name attributes?

wiki/X500-Name-rules.md
dimosr: good catch. I checked with Matthew and that value is intentional so I'll document it.

dimosr:
> Should we also mention that it must be a valid javax.security.auth.x500.X500Principal since we parse it as one to get the name attributes?

I think it would be ideal if we don't refer to code and rules are self-explanatory. So, instead of referring to X500Principal, it might be better to say that the string must be valid as defined in RFC 1779 or RFC 2253. Looking at the docs of X500Principal, this is what this really checks in this context.

charlieR3:
> I think it would be ideal if we don't refer to code and rules are self-explanatory. So, instead of referring to X500Principal, it might be better to say that the string must be valid as defined in RFC 1779 or RFC 2253. Looking at the docs of X500Principal, this is what this really checks in this context.

Much better idea. Thanks 👍

charlieR3 (approved):
LGTM

PR #3568, opened 2023-04-05
CORE-12293 - Disable re-registration

nikinagy:
We need to make sure that the uncompleted feature of re-registration would not be available for users in the GA release.

### DIFF
--- a/components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
+++ b/components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
@@ -26,6 +26,7 @@ import net.corda.data.membership.p2p.MembershipRegistrationRequest
 import net.corda.data.membership.p2p.UnauthenticatedRegistrationRequest
 import net.corda.data.membership.p2p.UnauthenticatedRegistrationRequestHeader
 import net.corda.data.p2p.app.AppMessage
+import net.corda.data.p2p.app.MembershipStatusFilter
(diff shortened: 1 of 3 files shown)


components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
```diff
@@ -313,8 +314,13 @@ class DynamicMemberRegistrationService @Activate constructor(
                 val mgm = groupReader.lookup().firstOrNull { it.isMgm }
                     ?: throw IllegalArgumentException("Failed to look up MGM information.")
 
+                val currentInfo = groupReader.lookup(member.x500Name, MembershipStatusFilter.ACTIVE_OR_SUSPENDED)
+                require(currentInfo == null) {
+                    "Re-registration is disabled."
```
charlieR3: nit picky but `not supported` might be better than `disabled` since disabled might imply that is a configuration that can be enabled.

charlieR3 (dismissed):
LGTM. One nit pick comment

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
nikinagy: Fair point. Addressed in [7bbbdd6](https://github.com/corda/corda-runtime-os/pull/3568/commits/7bbbdd6274bea53258e71f795612b07cb563aa8a)

charlieR3 (approved):

PR #3571, opened 2023-04-06
CORE-11852 Removal of client notarization request signing and verification

relyafi:
### Summary

Previously, a client virtual node requesting notarization would sign notarization requests as part of the notary client plugin logic, with the notary virtual node verifying the request signature as part of the notary server plugin logic.

This checking is redundant, as the P2P layer has already verified the sender identity by the time the notary virtual node processes a notarization request. Therefore, this change removes this checking.

### Testing

Some automated tests are impacted by this change and have been updated. Confirmed all tests pass.


blsemo (approved):
LGTM

nkovacsx (approved):

PR #3573, opened 2023-04-06
CORE-4541 - Immediately throw if session status is close or error on send/receive 

pmck91:
Calling send or receive on a `FlowSession` that has been closed should throw an error to the calling flow.

We want to eagerly throw an exception if a sessions status is CLOSED or ERROR to stop 

For more information see https://r3holdco.slack.com/archives/C04HJ0SJ8SH/p1680705429585929


LWogan:
There are some additional APIs for sending and receiving messages in FlowMessagingImpl that will need these changes applied to. Otherwise looks ok at a first glance, bar the failing smoke test.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FlowMessagingImpl.kt
```diff
@@ -25,9 +25,10 @@ import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
 import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
 import java.util.UUID
+import net.corda.data.flow.state.session.SessionStateType
 
 @Suppress("TooManyFunctions")
-@Component(service = [ FlowMessaging::class, UsedByFlow::class ], scope = PROTOTYPE)
+@Component(service = [FlowMessaging::class, UsedByFlow::class], scope = PROTOTYPE)
```
LWogan, line 31: Need some Unit tests for this one

LWogan:
Changes look good to me! just need some unit tests. You already mentioned you're planning on extracting out the validation method. That makes sense to me.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FlowMessagingImpl.kt
pmck91: Added now

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
```diff
@@ -244,7 +244,7 @@ class FlowTests {
     @Test
     fun `Platform Error - user code receives platform errors`() {
```
LWogan: This test originally was used to trigger the FlowEngine to throw a FlowPlatformException. This is then caught in the FlowEventExceptionProcessorImpl which calls checkpoint.setPendingPlatformError(PLATFORM_ERROR, exception.message).

This wil trigger some logic to then throw an exception in the fiber when it resumes.

Before your code change, a convenient way of trigger this logic evidently was to call send on a closed session and have the flow engine throw it. 

So I think your new test is valid and should be kept. The new test should have a new name. 

I think we should keep the existing test and just change the throwPlatformError method to do something else to throw a FlowPlatformException in the flow engine. We'll need to look around for uses of "throw FlowPlatformException" to see which is easily replicable from this test.

LWogan (approved):
LGTM

PR #3574, opened 2023-04-06
CORE-12253: use tenant ID consistently when dealing with wrapping table

dickon:
Always use the tenantId passed in, and take care when it is not passed in to be consistent about which tenant, and therefore which database, is used for storing the wrapping keys. This fixes a bug where we generate a wrapping key, store it in the cache and cluster DB, and then later expect to find it in the vndoe db, and so fail. That bug was a regression introduced last week, since before that point we always stored wrapping keys in the cluster database, which was problematic. I believe this fixes regressions seen by @jujoramos  and @williamvigorr3 but at this point that's not proven.


driessamyn (approved):
LGTM

PR #3576, opened 2023-04-07
CORE-8962 flow mapper absorbs messages when in state closing

emilybowe:
**Description:**
When a session has completed the Flow Mapper holds onto its state and sets it to CLOSING. It now ignores any incoming messages. This means if an ACK message is lost that one party cannot complete its close and will log the session terminated in error while the other party will record it completed successfully. This change will send automatic ACKS in that situation to prevent this. This PR also introduces a new state, ERROR. 

**Links:**
Ticket -> [CORE-8962: Flow Mapper swallows all messages it receives when it is in state CLOSING](https://r3-cev.atlassian.net/browse/CORE-8962)

Api pr -> [CORE-8962: api change for flow mapper absorbs messages (https://github.com/corda/corda-api/pull/1020)

//update api number before merge


components/flow/flow-mapper-service/src/main/kotlin/net/corda/session/mapper/service/executor/FlowMapperMessageProcessor.kt
```diff
@@ -34,40 +31,28 @@ class FlowMapperMessageProcessor(
...
-        return if (!isExpiredSessionEvent(value) && isValidState(state, value)) {
+        return if (!isExpiredSessionEvent(value)) {
             val executor = flowMapperEventExecutorFactory.create(key, value, state, flowConfig)
             val result = executor.execute()
             StateAndEventProcessor.Response(result.flowMapperState, result.outputEvents)
         } else {
-            logger.debug { "Ignoring event (isExpiredSessionEvent: ${isExpiredSessionEvent(value)}, " +
-                    "isValidState: ${isValidState(state, value)})" }
+            logger.debug {
+                errorMsg.format("ignored", isExpiredSessionEvent(value), state)
```
LWogan: I think isExpiredSessionEvent(value) in this case will always be true so can avoid double processing by either just hardcoding true, or assigning the result of the method to a val and using that in the conditional check and in the else.

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/FlowMapperEventExecutorFactoryImpl.kt
```diff
@@ -43,8 +45,26 @@ class FlowMapperEventExecutorFactoryImpl @Activate constructor(
                 val eventPayload = sessionEvent.payload
                 if (eventPayload is SessionInit) {
                     SessionInitExecutor(eventKey, sessionEvent, eventPayload, state, sessionEventSerializer, flowConfig)
```
LWogan: We should change the if ... else if ... else to a `when`. and maybe extract that additional when to another function to avoid nested whens which would be an eye sore

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionErrorExecutor.kt
```diff
@@ -0,0 +1,103 @@
...
+                log.warn("FlowMapperState with null status. Key: $eventKey, Event: $sessionEvent.")
+                FlowMapperResult(null, listOf())
+            }
+            FlowMapperStateType.ERROR -> {
+                log.warn(errorMsg + "Ignoring event.")
+                FlowMapperResult(flowMapperState, listOf())
+            }
+            FlowMapperStateType.OPEN -> {
+                log.warn(errorMsg + "Forwarding event.")
+                flowMapperState.status = FlowMapperStateType.ERROR
```
LWogan, line 62: Long winded comment incoming. In short I think this change is OK :D

I can see why you'd change the state from OPEN to ERROR. We have received an ERROR. This does raise an interesting point though

At the moment the only thing that sets the Flow Mapper state is the flow engine that created it. Up to now this was done by creating a new session via the first SessionInit event going out or coming in(sets to OPEN), and ending a session/flow and sending the Mapper a ScheduleCleanup event (sets to CLOSING),  Now we are adding an additional case where we send out a SessionError and this triggers the mapper state to go to ERROR.

Additionally this logic means if a counterparty sends an ERROR to us, it will also change the mapper status to be ERROR. I don't think this change would cause problems right now. But we might in the future change the flow engine to examine the session error and let it decide if it is in fact an error or not. If the flow engine decided to say "hey this isn't an error I care about, I'm going to ignore it" but the mapper has closed the gate and stopped letting messages through we could be in trouble. The flow engine could be willing to accept more messages but the mapper has already been set to ERROR. However, that use case does not exist yet. All errors are legit errors which is why this logic is fine.

If we changed this to ignore SessionErrors being receive from the other side it would mean wed have to add logic to the flow engine to send a message to the mapper after receiving an error to tell it to set to ERROR. We do something similar right now in the global post processor to schedule mapper cleanup. But yeah, for now, I think always setting the mapper state to ERROR works.

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionErrorExecutor.kt
```diff
@@ -0,0 +1,103 @@
...
+                if (messageDirection == MessageDirection.OUTBOUND) {
+                    FlowMapperResult(
+                        flowMapperState, listOf(
+                            Record(
+                                Schemas.P2P.P2P_OUT_TOPIC, sessionId, appMessageFactory(
```
LWogan: I think we have this logic to generate a record for p2p in two places, so it would be good to extract it out to a common method shared between the 2 classes.

LWogan (commented):
I've done a first pass. And I think it looks correct. I'll need to go back over it again though in more depth. 

I do think we're missing a small piece though. The flow mapper only holds onto states for a certain period of time after they error or close. We have logic in the FlowGlobalPostProcessor.getFlowMapperSessionCleanupEvents to send a ScheduleCleanup message to the mapper when a session errors or closes. This message also sets the mapper state to be CLOSING.  

So I think this scenario will happen when we receive a SessionError from a counterparty:
- mapper receives error, sets status to be ERROR
- mapper sends error to flow engine.
- flow engine changes it session state status to be ERROR
- FlowGlobalPostProcessor sees a session state with status ERROR
- FlowGlobalPostProcessor sends a ScheduleCleanup() to mapper
- mapper receives ScheduleCleanup
- mapper sets status to be CLOSING and schedules cleanup
- counterparty sends a normal session message, and mapper responds with an ACK.

 I think you need to make a change to make sure that the state stays in ERROR when the mapper receives a ScheduleCleanup message (if it is already in state ERROR)

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/FlowMapperEventExecutorFactoryImpl.kt
```diff
@@ -38,21 +40,49 @@ class FlowMapperEventExecutorFactoryImpl @Activate constructor(
...
+                        SessionInitExecutor(
+                            eventKey,
+                            flowMapperEventPayload,
+                            sessionPayload,
+                            state,
+                            sessionEventSerializer,
+                            flowConfig
+                        )
+                    }
+                    is SessionError -> {
```
LWogan, line 56: I think we're missing a unit test for SessionErrorExecutor in FlowMapperEventExecutorFactoryImplTest

components/flow/flow-mapper-service/src/main/kotlin/net/corda/session/mapper/service/executor/FlowMapperMessageProcessor.kt
mbrkic-r3: Message expects event as the second argument, not expiration status

mbrkic-r3: Changed code to send `event` as the second argument

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionErrorExecutor.kt
mbrkic-r3: Refactored duplicate code into function "createP2PRecord"

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/FlowMapperEventExecutorFactoryImpl.kt
mbrkic-r3: Added missing test

mbrkic-r3:
> I've done a first pass. And I think it looks correct. I'll need to go back over it again though in more depth.
> 
> I do think we're missing a small piece though. The flow mapper only holds onto states for a certain period of time after they error or close. We have logic in the FlowGlobalPostProcessor.getFlowMapperSessionCleanupEvents to send a ScheduleCleanup message to the mapper when a session errors or closes. This message also sets the mapper state to be CLOSING.
> 
> So I think this scenario will happen when we receive a SessionError from a counterparty:
> 
> * mapper receives error, sets status to be ERROR
> * mapper sends error to flow engine.
> * flow engine changes it session state status to be ERROR
> * FlowGlobalPostProcessor sees a session state with status ERROR
> * FlowGlobalPostProcessor sends a ScheduleCleanup() to mapper
> * mapper receives ScheduleCleanup
> * mapper sets status to be CLOSING and schedules cleanup
> * counterparty sends a normal session message, and mapper responds with an ACK.
> 
> I think you need to make a change to make sure that the state stays in ERROR when the mapper receives a ScheduleCleanup message (if it is already in state ERROR)

Updated `FlowMapper` state in `ScheduleCleanupEventExecutor` to `CLOSING` only if it was not `ERROR`

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionErrorExecutor.kt
```diff
@@ -0,0 +1,94 @@
...
+                                    ExceptionEnvelope(
+                                        "FlowMapper-SessionError",
+                                        "Received SessionError with sessionId $sessionId"
+                                    )
+                                ),
+                                instant,
+                                sessionEventSerializer,
+                                appMessageFactory,
+                                flowConfig,
+                                0
```
mbrkic-r3: Would `sessionEvent.receivedSequenceNum` have value 0 here so we don't need to send it?

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionEventExecutor.kt
```diff
@@ -38,54 +39,91 @@ class SessionEventExecutor(
...
+                            ExceptionEnvelope(
+                                "FlowMapper-SessionExpired",
+                                "Tried to process session event for expired session with sessionId $sessionId"
+                            )
+                        ),
+                        instant,
+                        sessionEventSerializer,
+                        appMessageFactory,
+                        flowConfig,
+                        0
```
mbrkic-r3, line 70: Would `sessionEvent.receivedSequenceNum` have value 0 here so we don't need to send it?

LWogan: I think 0 is ok. This scenario is for when a session is expired so we cant set it sensibly. The SessionError will trigger the error logic on the other side anyway so the field wont be used.

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionEventExecutor.kt
```diff
@@ -38,54 +39,91 @@ class SessionEventExecutor(
...
+                FlowMapperResult(null, listOf())
+            }
+            FlowMapperStateType.CLOSING -> {
+                if (messageDirection == MessageDirection.OUTBOUND) {
+                    log.warn("Attempted to send a message but flow mapper state is in CLOSING. Session ID: ${sessionEvent.sessionId}")
+                    FlowMapperResult(flowMapperState, listOf())
+                } else {
+                    val outputRecord =
+                        createP2PRecord(
+                            sessionEvent,
```
LWogan: I think this needs to be changed. The SessionEvent is used to carry the ack info on the receivedSequenceNum and outOfOrderSequenceNums fields. If we copy the event they sent us back to them, it will not give valid ack info. We need to set outOfOrderSequenceNums = emptylist and we need to set sessionEvent.receivedSequenceNum = sessionEvent.sequenceNum

LWogan: we should also set sessionEvent.sequenceNum = null as messages with SessionAck as a payload are not allowed have sequence numbers

mbrkic-r3: `emptyList` and for `outOfOrderSequenceNums` and `null` for `sequenceNum` are already used in `createP2PRecord`

LWogan: ah i see. ok then you just need to update how receivedSequenceNumber is set

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/FlowMapperHelper.kt
```diff
@@ -76,3 +77,41 @@ fun generateAppMessage(
...
+    return Record(
+        Schemas.P2P.P2P_OUT_TOPIC, sessionId, appMessageFactory(
+            SessionEvent(
+                MessageDirection.OUTBOUND,
+                instant,
+                sessionId,
+                null,
+                sessionEvent.initiatingIdentity,
+                sessionEvent.initiatedIdentity,
+                receivedSequenceNumber,
```
LWogan, line 110: should be set to sessionEvent.sequenceNumber

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionEventExecutor.kt
mbrkic-r3: Set `sessionEvent.sequenceNum` as default value for `receivedSequenceNumber` in `createP2PRecord`

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/FlowMapperHelper.kt
mbrkic-r3: Should be fixed by the change mentioned in the comment above

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionErrorExecutor.kt
LWogan: not a big deal. but the outbound message is one which originated from within the flow engine. so we can set  receivedSequenceNum as sessionEvent.receivedSequenceNum here

LWogan (dismissed):
These changes look good to me!

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionErrorExecutor.kt
mbrkic-r3: modified

LWogan (dismissed):
These changes look good to me. This handles the infinite ack problem we uncovered late on.

mnesbit (dismissed):
OK

thiagoviana (approved):
Re-Approving it.

PR #3578, opened 2023-04-07
CORE-11477 Revisit crypto configuration

kyriathar:
- aligns with having removed unused `cryptoConnectionFactory` configuration
- aligns with having removed unused `hsmService` configuration
- aligns with having replaced `hsmMap` configuration which was used to hold multiple HSM configurations with `hsm` (as we only support soft hsm)
- aligns with having generalized caching configurations
- aligns with having generalized retrying configurations

api PR: [#1044](https://github.com/corda/corda-api/pull/1044)


mnesbit (dismissed):
OK

driessamyn (dismissed):
LGTM

libs/configuration/configuration-validation/src/integrationTest/kotlin/net/corda/libs/configuration/validation/integration/ConfigurationValidationIntegrationTest.kt
```diff
@@ -82,7 +82,7 @@ class ConfigurationValidationIntegrationTest {
     @ParameterizedTest(name = "verify that a sensible default is created for config section: {0}")
     @ValueSource(strings = [
         MESSAGING_CONFIG,
-        CRYPTO_CONFIG,
+//        CRYPTO_CONFIG, TODO Crypto defaulted values needs reviewing
```
dickon, line 85: Why is this taken out? There's no reason that these tests shouldn't still work if you've made the correct updates to the JSON schema for the config.

kyriathar: Good catch so, I think this used to work because when defaulted before it would get empty value for `hsmMap` (`"default": {}`). But now in its place we have `hsm` which currently has no default value. In a follow up PR I am going to be removing unused properties from `hsm` config so I am going to address this there.

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/HSMServiceTests.kt
```diff
@@ -86,6 +87,7 @@ class HSMServiceTests {
         assertHSMAssociation(association2, tenantId, CryptoConsts.Categories.LEDGER)
     }
 
+    @Disabled
```
dickon, line 90: Please remove or fix this in a follow up PR.

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/HSMServiceTests.kt
```diff
@@ -114,6 +116,7 @@ class HSMServiceTests {
         )
     }
 
+    @Disabled
```
dickon, line 119: Please remove or fix this in a follow up PR.

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/HSMServiceTests.kt
```diff
@@ -150,6 +153,7 @@ class HSMServiceTests {
         assertThat(usage3).anyMatch { it.hsmId == CUSTOM2_HSM_ID && it.usages == 1 }
     }
 
+    @Disabled
```
dickon, line 156: Please remove or fix this in a follow up PR.

mnesbit (approved):
OK

dickon (approved):
I think this disables too many tests for now without justification and for the benefit of this work package. @kyriathar has stated he intends to fix them in follow up PR, and that now needs to be done before Kyriakos is moved on to other work if we are not to suffer an important  reduction in test coverage.

All the sample JSON configuration in CryptoConfigUtils.kt is now wrong since it has been not been updated to the new structure.  Ideally `CryptoConfigUtils.createDefaultCryptoConfig` would be deleted and the sample JSON would be deleted as well, but I found the sample JSON important for the maintainaibility of that file. 

Normally I'd have rejected this, but unfortunately the API change has already gone in so we need to get the matching changes in quickly otherwise we block other work that uses API changes.  So, I'll approve for now since it will still work and is still shippable.

libs/configuration/configuration-validation/src/integrationTest/kotlin/net/corda/libs/configuration/validation/integration/ConfigurationValidationIntegrationTest.kt
kyriathar: Actually had another look at it, empty default seems to be there for `hsm` configuration as well. Not really sure why it was not working, trying putting in back in [#3593](https://github.com/corda/corda-runtime-os/pull/3593)

kyriathar: Sorry, yes I think what I wrote above originally should be about right, so now we don't provide defaults for these required fields while required there, whereas previously their defaults were again required but hidden under some optional property.

PR #3579, opened 2023-04-08
CORE-12407: Fix lingering OSGi issues in the worker applications.

chrisr3:
- Identify OSGi fragment bundles via `BundleRevision.getTypes()` flags rather than by searching bundle headers.
- Fix `loadClass` and `getResources` methods for `OsgiClassLoader` to support multiple bundles correctly.
- Ensure uncaught lifecycle exceptions are logged properly instead of just printed to console.
- Replace `OsgiDelegatedClassLoader` with `BundleWiring.getClassLoader()`, and delete `OsgiDelegatedClassLoader`.


libs/lifecycle/lifecycle-impl/src/main/kotlin/net/corda/lifecycle/impl/LifecycleCoordinatorSchedulerFactoryImpl.kt
```diff
@@ -7,9 +7,16 @@ import net.corda.lifecycle.LifecycleCoordinatorSchedulerFactory
 import org.osgi.service.component.annotations.Activate
 import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Deactivate
+import org.slf4j.LoggerFactory
 
 @Component(service = [ LifecycleCoordinatorSchedulerFactory::class ])
 class LifecycleCoordinatorSchedulerFactoryImpl @Activate constructor(): LifecycleCoordinatorSchedulerFactory {
+    private val logger = LoggerFactory.getLogger(this::class.java)
+
+    private fun handleUncaughtException(thread: Thread, exception: Throwable) {
```
mnesbit, line 16: Any chance we can get the lifecycle object name?

chrisr3: It's the "uncaught exception handler", so where would I get that from? All I have to work with is the thread and the uncaught exception, and it's better for that to get logged properly than just sent to the console. I'm hoping you can infer the context from the thread name, and anything else in the Log4J logger context. And from the exception stack trace, of course.

So short answer: "No".

mnesbit (approved):
OK

libs/lifecycle/lifecycle-impl/src/main/kotlin/net/corda/lifecycle/impl/LifecycleCoordinatorSchedulerFactoryImpl.kt
mnesbit: The task that was being executed ideally

chrisr3: > The task that was being executed ideally

If the "task name" (`LifecycleCoordinator.name`?) is contained in either the logger context, the thread or the exception then it's logged already. But I cannot change the definition of what the [UncaughtExceptionHandler](https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.UncaughtExceptionHandler.html) is.

However, considering that the default ultimate `UncaughtExceptionHandler` is effectively:
```java
if (!(e instanceof ThreadDeath)) {
    System.err.print("Exception in thread \"" + t.getName() + "\" ");
    e.printStackTrace(System.err);
}
```
I still think this is an improvement.

driessamyn (approved):
LGTM

PR #3580, opened 2023-04-09
CORE-12378: implement support methods to reduce duplication in combined worker logic between OS and ENT versions of job

ronanbrowne:
Following on from https://github.com/corda/corda-shared-build-pipeline-steps/pull/708 and https://github.com/corda/corda5-enterprise/pull/34

Implement the support methods migrated to the Jenkins Shared library project to reduce duplication between this combined worker job and the Enterprise implementation

### DIFF
--- a/.ci/e2eTests/JenkinsfileCombinedWorker
+++ b/.ci/e2eTests/JenkinsfileCombinedWorker
@@ -1,6 +1,8 @@
-@Library('corda-shared-build-pipeline-steps@5.0.1') _
+@Library('corda-shared-build-pipeline-steps@CORE-12378-C5-Ent-pipeline-support') _
 
 import groovy.transform.Field
+import com.r3.build.utils.PipelineUtils
+import com.r3.build.utils.GitUtils
 
 @Field
 String postgresHost = 'localhost'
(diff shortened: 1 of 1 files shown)


.ci/e2eTests/JenkinsfileCombinedWorker
```diff
@@ -1,6 +1,8 @@
-@Library('corda-shared-build-pipeline-steps@5.0.1') _
+@Library('corda-shared-build-pipeline-steps@CORE-12378-C5-Ent-pipeline-support') _
```
ronanbrowne: temp while testing

.ci/e2eTests/JenkinsfileCombinedWorker
```diff
@@ -70,52 +64,17 @@ pipeline {
...
-                                echo "${error.getMessage()}\nPort forwarding Postgres has not been set up yet, retrying"
-                                retry(5) {
-                                    sleep(time: 5, unit: "SECONDS")
-                                    sh 'psql --quiet --tuples-only -c \'select \''
-                                }
-                            }
-                            sh 'createdb -w "${DATABASE}"'
-                        }
-                    }
+                script {
```
yarivus, line 76: why does this need a script { } scope ?

.ci/e2eTests/JenkinsfileCombinedWorker
```diff
@@ -143,16 +102,8 @@ pipeline {
...
-                    try {
-                       timeout(time: 3, unit: 'MINUTES') {
-                            checkConnection()
-                       }
-                    } catch(err) {
-                        // If a timeout is reached fail the build as we want the calling job to result in a failure
-                        // without this the calling job will result in a aborted status
-                        error 'Could not connect to the Combined Worker in a 3 minute window'
-                    }
+                script {
```
yarivus, line 105: same comment as prev

yarivus (changes requested):

.ci/e2eTests/JenkinsfileCombinedWorker
ronanbrowne: @yarivus you can not call a method on a object in a Jenkins declarative scrip pipeline , such as 

`pipelineUtils.createPostgresDatabase(postgresPort, postgresHost, postgresDb, postgresCredentialsId)
`

it simply will not compile without being in a script block, we did not previously need this as we used a different approach, we were simply calling a method which already existed in this class so no object declaration was needed then needed, and thus no script block

.ci/e2eTests/JenkinsfileCombinedWorker
ronanbrowne: see answer above

yarivus (approved):
LGTM

PR #3588, opened 2023-04-11
CORE-9481 Removed filename from chunk and refactored CPI writer to store filename in properties

ciaran238:
The chunk object does not need a filename property as its not used in all cases, it makes more sense to move it to the generic properties map so it can be added or not as required
[Link](https://github.com/corda/corda-api/pull/1093) to the associated api changes
[Old API Change PR here](https://github.com/corda/corda-api/pull/1046)


components/chunking/chunk-db-write-impl/src/main/kotlin/net/corda/chunking/db/impl/persistence/database/DatabaseChunkPersistence.kt
```diff
@@ -72,7 +72,7 @@ class DatabaseChunkPersistence(private val entityManagerFactory: EntityManagerFa
             // Persist this chunk.
             val chunkEntity = ChunkEntity(
                 chunk.requestId,
-                chunk.fileName,
+                chunk.properties.items.find {it.key == "FileName"}?.value,
```
mnesbit: No magic constants please!!! Define a constant for this property in the chunking api module and use it everywhere the property is read/written. FileName is also a bit ambiguous, so perhaps ChunkFileName ?

ben-millar: Also, what happens if we can't find this item? Is it okay for this to be `null`, or should there be a default value?

ben-millar: Discussed my question offline; it's fine for the `filename` not to exist, as not all chunked objects are files. However, I'm not sure if the `ChunkEntity` object should still have a `filename` property after this change, maybe @LWogan can clarify.

LWogan: Yeah, it looks like there's some more refactoring to do. The fileName field should be removed from ChunkEntity

components/chunking/chunk-db-write-impl/src/test/kotlin/net/corda/chunking/db/impl/validation/AssembleFromChunksTest.kt
```diff
@@ -24,10 +25,16 @@ import java.util.UUID
 
 internal class AssembleFromChunksTest {
     lateinit var fs: FileSystem
+    companion object {
+        const val expectedFileName = "some.cpi"
 
+    }
+    private var properties: MutableMap<String, String?>? = null
```
LWogan: if we know we're going to always set this before every test  then we don't need to make it nullable. you can use this instead

`private lateinit var properties: MutableMap<String, String?>`

then you won't need to the !!. Generally we want to avoid !! as much as possible.

libs/chunking/chunking-core/src/test/kotlin/net/corda/chunking/ChunkWritingTest.kt
```diff
@@ -75,8 +76,10 @@ class ChunkWritingTest {
     @Test
     fun `chunk file name is set correctly`() {
```
LWogan, line 77: This test will execute the code that youve added to set properties to a new mutable map created within the ChunkWriterImpl. We would need a similar test that makes sure that when we pass in a mutable map that already has some data in it, that data doesnt get erased, and that the filename does still get added. However a different PR comment ive made is suggesting we remove filename from the api so this point is moot!

libs/chunking/chunking-core/src/main/kotlin/net/corda/chunking/impl/ChunkWriterImpl.kt
```diff
@@ -39,11 +39,16 @@ internal class ChunkWriterImpl(
     //add extra overhead to avoid message bus level chunking
     val chunkSize = maxAllowedMessageSize - APP_LEVEL_CHUNK_MESSAGE_OVERHEAD
 
-    override fun write(fileName: String, inputStream: InputStream): ChunkWriter.Request {
+    override fun write(fileName: String?, inputStream: InputStream): ChunkWriter.Request {
```
LWogan: I'm not sure if its correct to make fileName nullable here. Thinking about this a bit more. It looks like the ChunkReader doesn't actually have a concept of the chunks having a file name. The filename logic lives in the function passed to the chunk reader factory for onComplete in the CPIValidator:

```
// Set up chunk reader.  If onComplete is never called, we've failed.
    val reader = chunkReaderFactory.create(cacheDir).apply {
        onComplete {  tempPathOfBinary: Path, fileChecksum: SecureHash, properties: Map<String, String?>? ->
            fileName = properties?.get(CHUNK_FILENAME_KEY)
            tempPath = tempPathOfBinary
            checksum = fileChecksum
            localProperties = properties
        }
    }

    // Now read chunks, and create CPI on local disk
    chunkPersistence.forEachChunk(requestId, reader::read)

    return with(fileName) {
        if (this == null) {
            throw ValidationException("Did not combine all chunks to produce file", requestId)
        }

        FileInfo(this, tempPath, checksum, localProperties)
    }
```

    
If thats the case then I think we can do the same on the write side too. We should pass filename to the factory inside the properties map rather than via the API `fun write(fileName: String?, inputStream: InputStream)` and then writing to the map after.

I think we should change the CpiWriter API to `fun write(inputStream: InputStream)`. This should simplify the code in the CpiWriterImpl too.

LWogan (commented):
Mostly looks good to me. I've added a comment about removing fileName from the CPIWriter API. I think this would make sense to completely remove the concept of fileName from the reader and writer.

components/chunking/chunk-db-write-impl/src/integrationTest/kotlin/net/corda/chunking/db/impl/tests/DatabaseChunkPersistenceTest.kt
```diff
@@ -138,19 +139,21 @@ internal class DatabaseChunkPersistenceTest {
             it.write(mockCpkContent)
         }
 
+        val properties= mutableMapOf<String,String?>()
```
LWogan: minor comment but the formatting is off here. missing space between 'properties' and '='. theres another further down too

"cntrl+alt+l" is your friend for formatting

libs/chunking/chunking-core/src/main/kotlin/net/corda/chunking/impl/ChunkWriterImpl.kt
```diff
@@ -22,11 +22,10 @@ import java.util.UUID
 internal class ChunkWriterImpl(
     maxAllowedMessageSize: Int,
     private val chunkBuilderService: ChunkBuilderService,
-    private val properties: Map<String, String?>? = null,
+    private var properties: MutableMap<String, String?>? = null,
```
LWogan: this can be `val` and `Map` rather than var and MutableMap now

LWogan (commented):
One more small change, Otherwise this looks good to me.

thiagoviana (dismissed):

components/chunking/chunk-db-write-impl/src/test/kotlin/net/corda/chunking/db/impl/validation/AssembleFromChunksTest.kt
```diff
@@ -24,10 +25,16 @@ import java.util.UUID
...
     lateinit var fs: FileSystem
+    companion object {
+        const val expectedFileName = "some.cpi"
 
+    }
+    private lateinit var properties: MutableMap<String, String?>
     @BeforeEach
     fun beforeEach() {
         fs = Jimfs.newFileSystem()
+        properties = mutableMapOf()
```
driessamyn: This can be initialised as a normal val property outside the `@BeforeEach`

components/chunking/chunk-db-write-impl/src/test/kotlin/net/corda/chunking/db/impl/validation/AssembleFromChunksTest.kt
```diff
@@ -24,10 +25,16 @@ import java.util.UUID
 
 internal class AssembleFromChunksTest {
     lateinit var fs: FileSystem
+    companion object {
+        const val expectedFileName = "some.cpi"
 
+    }
+    private lateinit var properties: MutableMap<String, String?>
```
driessamyn: Minor: formatting is a bit odd as the properties are separatd by the companion object.

libs/chunking/chunking-datamodel/src/main/kotlin/net/corda/chunking/datamodel/ChunkEntity.kt
driessamyn, line 1: There needs to be a DB change related to this.
Is this planning to go in before 5.0 ?

components/chunking/chunk-db-write-impl/src/integrationTest/kotlin/net/corda/chunking/db/impl/tests/DatabaseChunkPersistenceTest.kt
```diff
@@ -138,19 +138,21 @@ internal class DatabaseChunkPersistenceTest {
             it.write(mockCpkContent)
         }
 
+        val properties = mutableMapOf<String, String?>()
+        properties[Constants.CHUNK_FILENAME_KEY] = someFile
```
driessamyn: Does this need to be a mutable Map?

components/chunking/chunk-db-write-impl/src/integrationTest/kotlin/net/corda/chunking/db/impl/tests/RecreateBinaryTest.kt
```diff
@@ -82,13 +83,15 @@ class RecreateBinaryTest {
         val chunkSize = loremIpsum.length / divisor
 
         val chunks = mutableListOf<Chunk>()
-        val writer = ChunkWriterFactory.create(chunkSize + APP_LEVEL_CHUNK_MESSAGE_OVERHEAD).apply {
+        val properties= mutableMapOf<String,String?>()
+        properties[CHUNK_FILENAME_KEY] = someFile
```
driessamyn: As above, this can be an immutable map.

components/virtual-node/cpk-write-service-impl/src/main/kotlin/net/corda/cpk/write/impl/CpkWriteServiceImpl.kt
```diff
@@ -226,16 +227,18 @@ class CpkWriteServiceImpl @Activate constructor(
     private fun CpkChunksPublisher.chunkAndPublishCpk(cpkFile: CpkFile) {
         logger.debug { "Publishing CPK ${cpkFile.fileChecksum}" }
         val cpkChecksum = cpkFile.fileChecksum
+        val properties= mutableMapOf<String,String?>()
+        properties[CHUNK_FILENAME_KEY] = cpkChecksum.toFileName()
```
driessamyn: As above, can be immutable.

components/virtual-node/virtual-node-rest-maintenance-impl/src/main/kotlin/net/corda/libs/virtualnode/maintenance/rest/impl/v1/VirtualNodeMaintenanceRestResourceImpl.kt
```diff
@@ -133,10 +141,12 @@ class VirtualNodeMaintenanceRestResourceImpl @Activate constructor(
         if (!isRunning) throw IllegalStateException(
             "${this.javaClass.simpleName} is not running! Its status is: ${lifecycleCoordinator.status}"
         )
-        val cpiUploadRequestId = tryWithExceptionHandling(logger,"Force CPI upload") {
+        val properties = mutableMapOf(PropertyKeys.FORCE_UPLOAD to true.toString())
+        properties[CHUNK_FILENAME_KEY] = upload.fileName
```
driessamyn: As above.

libs/chunking/chunking-core/src/main/kotlin/net/corda/chunking/ChunkWriterFactory.kt
```diff
@@ -13,6 +13,6 @@ object ChunkWriterFactory {
             throw CordaRuntimeException("Cannot write chunks larger than $MAX_DB_CHUNK_SIZE because it will exceed the db table definition")
         }
 
-        return ChunkWriterImpl(maxAllowedMessageSize, ChunkBuilderServiceImpl(), properties)
+        return ChunkWriterImpl(maxAllowedMessageSize, ChunkBuilderServiceImpl(), properties?.toMutableMap())
```
driessamyn: Does this need to be a mutable map?

libs/chunking/chunking-core/src/test/kotlin/net/corda/chunking/ChunkReadingTest.kt
```diff
@@ -49,22 +50,24 @@ class ChunkReadingTest {
 
     @Test
     fun `can read in order chunks`() {
+        val ourFileName = randomFileName()
+        val properties = mutableMapOf<String, String?>()
+        properties[CHUNK_FILENAME_KEY] = ourFileName
```
driessamyn: As above.

driessamyn (dismissed):
Apart from the comments about the use of a mutable map, the DB change worries me slightly as this is not reflected in the api repo and DB changes will be harder post-GA, so we should be clear about when this is in scope for.

libs/chunking/chunking-datamodel/src/main/kotlin/net/corda/chunking/datamodel/ChunkEntity.kt
ciaran238: Its tagged for 5.1, but do you think it should go in 5.0?

driessamyn: > Its tagged for 5.1, but do you think it should go in 5.0?

That's a question for @thiagoviana and team. But I would carefully consider it as removing a DB field is not easy in an upgrade scenario, so I would consider removing it now.
But either way, this PR should not get merged until the DB Entity reflects what's defined in the API.

BTW. Given @thiagoviana approved this PR, I assume it was in scope for 5.0.

LWogan: This ticket was a leftover ticket from the chunking/rocks stuff a while back. It has fallen into a loose grouping tickets in an epic that was moved from 5.0 to 5.1. I've been giving Ciaran a hand with this isolated ticket. I think given the db change, getting this in this week for 5.0 makes sense. It is nearly complete, he's made that schema change. Looks like theres some unit test updates and a merge from release to do. so it should be ok for 5.0

thiagoviana: Agreed @LWogan. Applying this DB change now will avoid some extra work in the future.

libs/chunking/chunking-core/src/test/kotlin/net/corda/chunking/ChunkWritingTest.kt
```diff
@@ -48,13 +49,13 @@ class ChunkWritingTest {
 
     @Test
     fun `simple chunking`() {
-        val chunkWriter = ChunkWriterFactory.create(1 * MB)
+        val chunkWriter = ChunkWriterFactory.create(1 * MB, mutableMapOf())
```
driessamyn: Minor, but this can be `emptyMap()`

libs/chunking/chunking-core/src/test/kotlin/net/corda/chunking/ChunkWritingTest.kt
```diff
@@ -67,20 +68,27 @@ class ChunkWritingTest {
...
         assertThat(0).isEqualTo(lastChunkSize)
     }
 
     @Test
     fun `chunk file name is set correctly`() {
         val fileName = randomFileName()
-        val writer = ChunkWriterFactory.create(1 * MB).apply {
-            onChunk { assertThat(it.fileName).isEqualTo(fileName.toString()) }
+        val properties = mutableMapOf<String, String?>()
+        properties[CHUNK_FILENAME_KEY] = fileName
```
driessamyn: This one can also be changed to regular immutable map.

libs/chunking/chunking-datamodel/src/main/kotlin/net/corda/chunking/datamodel/ChunkEntity.kt
driessamyn: Thanks. Makes sense.

Removed the blocking review, but will still need to be re-approved once API is merged and version set to `-beta`.

LWogan (dismissed):
These changes look good to me. lets try get this merged for friday.

edit: looks like you missed some of the mutable maps still. otherwise it looks ok.

LWogan (dismissed):
LGTM. You've resolved the merge conflict test failures caused by the previous API merge yesterday.

lankydan (approved):

driessamyn (approved):
LGTM (packaging)

PR #3590, opened 2023-04-11
CORE-12375: Remove combined worker and notary cpb external publishing

knguyen-r3:
* Remove reusable code block in build.gradle files which controls publishing to an internal staging area in S3, which is later promoted to maven Staging / Maven central

### DIFF
--- a/applications/workers/release/combined-worker/build.gradle
+++ b/applications/workers/release/combined-worker/build.gradle
@@ -6,8 +6,6 @@ plugins {
 }
 
 ext {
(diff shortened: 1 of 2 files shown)


ronanbrowne:
@driessamyn / @relyafi FYI, minor change here ensures these modules will not get picked up for publication to maven central on external release.

cc @abellagonzalo / @tlawson3 also for CSDE visibility

ronanbrowne (approved):
LGTM

PR #3592, opened 2023-04-11
CORE-12408 Passphrase location moved

davidcurrie:
When the passphrase moved from `workers.db` to `config.encryption` I must have missed this reference.


JasonSinclair95 (approved):
LGTM

PR #3596, opened 2023-04-12
CORE-12386: Avoid duplications in link manager replay manager

yift-r3:
This change will cancel any previous replay task before recreating a new one.


dimosr (commented):
Code looks relatively benign to me, but I'd like to understand a few things:
* a description of what was the issue seen in the larger context. Was some upstream client invoking `addForReplay` repeatedly or something else? 
* How was this tested?

yift-r3:
* The link manager kept sending the same message over and over again.
* It was easy to test; I just added a log message on the `MembershipP2PProcessor` and saw a lot of messages sent once and received twice. After the fix, each message was received only once.

dimosr (approved):
Given the existing, known issues around the replays caused by this, happy to get this fix in since it will make things better. 

But, it's worth understanding long term what is the actual reason the link manager kept replaying the message, given `addForReplay` seems to only be called once from `DeliveryTracker` when partitions are synced or new state entries are committed, which should be mutually exclusive.

PR #3597, opened 2023-04-12
CORE-9790: refactor helm bootstrap to reduce repetition

dickon:
Previously we had 6 instances of a pattern where we run corda-cli to make some SQL, then run another container to execute it. This was hard to maintain, so used a new named template to go from 6 to 1 copy of the declarations. 

To verify this produces the exact same output, I used `helm template` and then diffed the output against the old run. The differences are only in the random elements.

This exercise has revealed various unnecessary consistencies in the whitespace, quoting and naming. For now, I've got conditionals that preserve the inconsistencies, and we should make this big change, then remove the inconsistencies in later work.

### DIFF
--- a/charts/corda-lib/templates/_bootstrap.tpl
+++ b/charts/corda-lib/templates/_bootstrap.tpl
@@ -41,7 +41,7 @@ spec:
             - mountPath: /tmp/working_dir
               name: working-volume
       initContainers:
-        {{- include "corda.generateAndExecuteSql" ( dict "name" "db" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "RBAC_DB_USER" "schema" "RBAC" "quoteUser" "true" "namePostfix" "schemas") | nindent 8 }}
+        {{- include "corda.generateAndExecuteSql" ( dict "name" "db" "Values" .Values "Chart" .Chart "Release" .Release "schema" "RBAC" "quoteUser" "true" "namePostfix" "schemas") | nindent 8 }}
         {{- include "corda.generateAndExecuteSql" ( dict "name" "rbac" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "RBAC_DB_USER" "schema" "RBAC" "quoteUser" "true") | nindent 8 }}
         {{- include "corda.generateAndExecuteSql" ( dict "name" "vnodes" "longName" "virtual-nodes" "admin" "true" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "DB_CLUSTER") | nindent 8 }}
         {{- include "corda.generateAndExecuteSql" ( dict "name" "crypto" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "CRYPTO_DB_USER" "quoteUser" "true" "quotePassword" "false" "schema" "CRYPTO") | nindent 8 }}                
--- a/charts/corda-lib/templates/_bootstrap.tpl
+++ b/charts/corda-lib/templates/_bootstrap.tpl
@@ -66,8 +66,8 @@ spec:
               name: working-volume
           env:
           {{- include "corda.bootstrapClusterDbEnv" . | nindent 12 }}
-        {{- include "corda.bootstrapInitialConfigGenerateAndApply" ( dict "name" "rbac" "clusterDb" "false" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "RBAC_DB_USER" "schema" "RBAC" "quoteUser" "true") | nindent 8 }}
-        {{- include "corda.bootstrapInitialConfigGenerateAndApply" ( dict "name" "vnodes" "longName" "virtual-nodes" "admin" "true" "clusterDb" "true" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "DB_CLUSTER") | nindent 8 }}
+        {{- include "corda.bootstrapInitialConfigGenerateAndApply" ( dict "name" "rbac" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "RBAC_DB_USER" "schema" "RBAC" "quoteUser" "true") | nindent 8 }}
+        {{- include "corda.bootstrapInitialConfigGenerateAndApply" ( dict "name" "vnodes" "longName" "virtual-nodes" "admin" "true" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "DB_CLUSTER") | nindent 8 }}
         {{- include "corda.bootstrapInitialConfigGenerateAndApply" ( dict "name" "crypto" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "CRYPTO_DB_USER" "quoteUser" "true" "quotePassword" "false" "schema" "CRYPTO") | nindent 8 }}                
         {{- include "corda.bootstrapInitialConfigGenerateAndApply" ( dict "name" "rpc"  "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "REST_API_ADMIN" "quoteUser" "true" "quotePassword" "false" "schema" "RBAC"  "searchPath" "RBAC" "command" "create-user-config" "namePostfix" "admin" "sqlFile" "rbac-config.sql") | nindent 8 }}
         - name: create-db-users-and-grant
@@ -467,15 +467,18 @@ a second init container to execute the output SQL to the relevant database
     {{- if or (eq .name "rbac") (eq .name "crypto")  (eq .name "vnodes") -}}
        {{- "\n    " -}} {{- /* legacy whitespace compliance */ -}}
     {{- end -}}    
+
     {{- /* TODO remove this special case, it is just that the old template has these declarations later */ -}}
     {{- if not (eq .name "rpc") -}}
       {{ include "corda.bootstrapCliEnv" . | nindent 4 -}}{{- /* set JAVA_TOOL_OPTIONS, CONSOLE_LOG*, CORDA_CLI_HOME_DIR */ -}}
     {{- end -}}
+
     {{- if or (eq .name "rbac") (eq .name "vnodes") }}
     {{ include "corda.rbacDbUserEnv" . | nindent 4 }}
     {{- end -}}
-    {{- /* TODO remove this and use the name instead */ -}}
-    {{- if eq .clusterDb "true" -}}
+    
+
+    {{- if eq .name "vnodes" -}}
       {{ include "corda.clusterDbEnv" . | nindent 4 -}}
     {{- end -}}
     {{- if eq .name "rpc" -}}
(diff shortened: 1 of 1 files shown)


charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -507,7 +359,7 @@ serviceAccountName: {{ . }}
 {{- end }}
 
 {{/*
-Boostrap Corda CLI environment variables
+Bootstrap Corda CLI environment variables
```
davidcurrie, line 362: That must have been me - I've lost count of the number of times I've typed boostrap!

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -519,3 +371,103 @@ Boostrap Corda CLI environment variables
...
+
+
+{{/*
+Bootstrap declaration to declare an initial container for running corda-cli initial-config, then
+a second init container to execute the output SQL to the relevant database
+*/}}
+
+{{- define "corda.generateAndExecuteSql" -}}
+{{- /* define 2 init containers, which run in sequence. First run corda-cli initial-config to generate some SQL, storing in a persistent volume called working-volume. Second is a postgres image which mounts the same persistent volume and executes the SQL. */ -}}  
+- name: create-{{- if not (eq .name "db") -}}initial-{{- end -}}{{ .name }}-{{ .namePostfix | default "db-config" }}
```
davidcurrie, line 383: there's nothing special about these container names - we should just be consistent

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -519,3 +371,103 @@ Boostrap Corda CLI environment variables
...
+  image: {{ include "corda.bootstrapCliImage" . }}
+  imagePullPolicy: {{ .Values.imagePullPolicy }}
+  {{- include "corda.bootstrapResources" . | nindent 2 }}
+  {{- include "corda.containerSecurityContext" . | nindent 2 }}
+  {{- if eq .name "db" }}
+  args: [ 'database', 'spec', '-g', 'config:{{ .Values.db.cluster.schema }},rbac:{{ .Values.bootstrap.db.rbac.schema }},crypto:{{ .Values.bootstrap.db.crypto.schema }}', '-c', '-l', '/tmp/working_dir', '--jdbc-url', 'jdbc:{{ include "corda.clusterDbType" . }}://{{ required "A db host is required" .Values.db.cluster.host }}:{{ include "corda.clusterDbPort" . }}/{{ include "corda.clusterDbName" . }}', '-u', $(PGUSER), '-p', $(PGPASSWORD) ]  
+  {{- else }}
+  args: [ 'initial-config', '{{ .subCommand | default "create-db-config" }}',{{ " " -}}
+  
+         {{- /* request admin access in some cases, onl when the optional admin argument to this function (named tempalte) is specified as true */ -}}
```
davidcurrie:
```suggestion
         {{- /* request admin access in some cases, only when the optional admin argument to this function (named template) is specified as true */ -}}
```

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -519,3 +371,103 @@ Boostrap Corda CLI environment variables
...
+         {{- if eq (.admin | default "false") "true" -}} '-a',{{ " " -}}{{- end -}}
```
davidcurrie: you can use actual boolean values rather than strings.

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -519,3 +371,103 @@ Boostrap Corda CLI environment variables
...
+         
+         {{- if (and (not (eq .name "db")) (not (eq .subCommand "create-crypto-config"))) -}}
+           {{- /* specify DB user - note that the quotes being optional is to preserve output while refactory, we can always safely quote */ -}}
```
davidcurrie, line 397: what's your preference for handling this? merge this PR and then have another to bring everything into line? our propensity for squashing commits means this subtlety is lost otherwise

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -519,3 +371,103 @@ Boostrap Corda CLI environment variables
...
+           {{- "'-u'" -}},{{ " " -}}{{- if .quoteUser }}'{{- end -}} $({{ .environmentVariablePrefix -}}_USERNAME){{- if .quoteUser }}'{{- end -}},
+         
+           {{- /* specify DB password - note again that the quotes being optional is to preserve output while refactory, we can always safely quote */ -}}   
+           {{- " '-p'" -}}, {{- if .quotePassword }} '{{- else -}} {{ " " -}}{{- end -}}$({{ .environmentVariablePrefix -}}_PASSWORD){{- if .quotePassword }}'{{- end -}},
+         {{- end -}}           
+         
+         {{- if and (not (eq .name "rpc")) (not (eq .subCommand "create-crypto-config")) (not (eq .name "db")) -}}
```
davidcurrie: we're already not in the `db` leg at this point; it might be preferable to say what names/commands we are passing these variables for

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -519,3 +371,103 @@ Boostrap Corda CLI environment variables
...
+             {{- " '--name'" -}}, 'corda-{{ .longName | default .name }}', 
+             {{- " '--jdbc-url'" -}}, 'jdbc:{{ include "corda.clusterDbType" . }}://{{ required "A db host is required" .Values.db.cluster.host }}:{{ include "corda.clusterDbPort" . }}/{{ include "corda.clusterDbName" . }}{{- if .schema }}?currentSchema={{.schema }}{{- end -}}', 
+             {{- " '--jdbc-pool-max-size'" -}}, {{ .Values.bootstrap.db.rbac.dbConnectionPool.maxSize | quote }}, {{- " " -}}
+         {{- end -}}         
+         
+         {{- if (and (not (eq .name "rpc")) (not (eq .name "db"))) -}}
```
davidcurrie: we're already not in the `db` leg at this point

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -519,3 +371,103 @@ Boostrap Corda CLI environment variables
...
+           {{- "'-u'" -}},{{ " " -}}{{- if .quoteUser }}'{{- end -}} $({{ .environmentVariablePrefix -}}_USERNAME){{- if .quoteUser }}'{{- end -}},
+         
+           {{- /* specify DB password - note again that the quotes being optional is to preserve output while refactory, we can always safely quote */ -}}   
+           {{- " '-p'" -}}, {{- if .quotePassword }} '{{- else -}} {{ " " -}}{{- end -}}$({{ .environmentVariablePrefix -}}_PASSWORD){{- if .quotePassword }}'{{- end -}},
+         {{- end -}}           
+         
+         {{- if and (not (eq .name "rpc")) (not (eq .subCommand "create-crypto-config")) (not (eq .name "db")) -}}
+             {{- " '--name'" -}}, 'corda-{{ .longName | default .name }}', 
+             {{- " '--jdbc-url'" -}}, 'jdbc:{{ include "corda.clusterDbType" . }}://{{ required "A db host is required" .Values.db.cluster.host }}:{{ include "corda.clusterDbPort" . }}/{{ include "corda.clusterDbName" . }}{{- if .schema }}?currentSchema={{.schema }}{{- end -}}', 
+             {{- " '--jdbc-pool-max-size'" -}}, {{ .Values.bootstrap.db.rbac.dbConnectionPool.maxSize | quote }}, {{- " " -}}
```
davidcurrie: the pool size needs to be parameterised (currently it's always using the value for `rbac`)

davidcurrie (changes requested):

charts/corda-lib/templates/_bootstrap.tpl
dickon: Yes, I propose to clean this up in a later change.

charts/corda-lib/templates/_bootstrap.tpl
dickon: Yes

charts/corda-lib/templates/_bootstrap.tpl
dickon: fixed locally

charts/corda-lib/templates/_bootstrap.tpl
dickon: fixed locally

charts/corda-lib/templates/_bootstrap.tpl
dickon: I wonder if the baseline has a bug here, since `.Values.bootstrap.db.rbac.dbConnectionPool.maxSize` is used when setting up vondes initial config.

```
       - name: create-initial-vnodes-db-config
          image: {{ include "corda.bootstrapCliImage" . }}
          imagePullPolicy: {{ .Values.imagePullPolicy }}
          {{- include "corda.bootstrapResources" . | nindent 10 }}
          {{- include "corda.containerSecurityContext" . | nindent 10 }}
          args: [ 'initial-config', 'create-db-config', '-a', '-u', $(DB_CLUSTER_USERNAME), '-p', $(DB_CLUSTER_PASSWORD), '--name', 'corda-virtual-nodes', '--jdbc-url', 'jdbc:{{ include "corda.clusterDbType" . }}://{{ required "A db host is required" .Values.db.cluster.host }}:{{ include "corda.clusterDbPort" . }}/{{ include "corda.clusterDbName" . }}', '--jdbc-pool-max-size', {{ .Values.bootstrap.db.rbac.dbConnectionPool.maxSize | quote }},
                 '--salt', "$(SALT)", '--passphrase', "$(PASSPHRASE)", 
                 '-l', '/tmp/working_dir']

```

dickon: In release/os/5.0 we have:

So:

* .Values.bootstrap.db.rbac.dbConnectionPool.maxSize is used for create-initial-rbac-db-config
* .Values.bootstrap.db.crypto.dbConnectionPool.maxSize  is used for create-initial-crypto-db-config
* .Values.bootstrap.db.rbac.dbConnectionPool.maxSize is used for create-initial-vnodes-db-config
          
I think I'd still like to preserve behaviour so should replicate this for now and fix it in a separete PR.

charts/corda-lib/templates/_bootstrap.tpl
dickon: fixed

charts/corda-lib/templates/_bootstrap.tpl
dickon: pushed

charts/corda-lib/templates/_bootstrap.tpl
dickon: resolved, preserving what we have right now even though I don't know if it is correct.

davidcurrie: @simon-johnson-r3 - are the virtual nodes picking up these pool size properties from the config? if so, we should expose this as a separate property rather than arbitrarily reusing the RBAC one

davidcurrie (approved):
👍

PR #3598, opened 2023-04-12
CORE-12422 - Flow External Messaging - Dev Testing Issues

filipesoliveira:
Updated the class `ChunkDbWriterFactoryImpl` so that the correct configuration validator is created. Added the `external-channels.json` file to the smoke tests so that if the wrong validator is created, the smoke test will fail.

Changed the template replacement code to use a single $ prefix.

Added the map of the `externalMessagingRouteConfig` field of the Avro object in the .toAvro() method on net.corda.virtualnode.VirtualNodeInfo.


owenstanford (approved):
LGTM

kyriathar (approved):

PR #3599, opened 2023-04-12
CORE-12189: Update VNode creation test utilities

josephzunigadaly:
E2eCluster.createVirtualNode:
- Change timeout to 1 minute
- Use vnode create request ID to poll status and read holding ID from that response

ClusterBuilder.awaitVirtualNodeOperationStatusCheck:
- Change timeout to 1 minute
- Remove nullable from return type


vkolomeyko (approved):
LGTM

PR #3581, opened 2023-04-11
CORE-10585 Add E2E test utilities to support member suspension testing

charlieR3:
This adds the e2e test utility functions used in the e2e test repo for testing membership suspension and activation.

This includes a refactor to make utility functions operate in the context of `ClusterInfo` rather than passing the `ClusterInfo` as a parameters. This allows for more readable tests, especially for multi cluster tests.

E2E test repo update: https://github.com/corda/corda-e2e-tests/pull/59

PR #3601, opened 2023-04-12
CORE-10585 Multi cluster test support in E2E test utils and bug fixes for some multi cluster functionality

charlieR3:
Follows on from https://github.com/corda/corda-runtime-os/pull/3581 to add support for multi cluster testing. 

All of the E2E tests currently target a default cluster specified by calling the `endpoint` function in every test using cluster builder. That one cluster URI used by all tests is set by setting `restEndpointUrl` in the project properties, or else by defaulting to `localhost:8888`. To support multi cluster testing we need to be able to specify multiple cluster endpoints and to do so using the existing method of specifying environment variables for cluster A, B, and C would mean having multiple places to set the URI for different tests. In this PR I've removed that duplication by removing the `restEndpointUrl` properties. `E2E_CLUSTER_B_REST_HOST` and `E2E_CLUSTER_B_REST_PORT` should be used instead. A util property `DEFAULT_CLUSTER` has been added also set to cluster B and tests which were previously using the same URI now target this cluster.

Functions which were required for multi cluster testing have been updated to be called in the context of a `ClusterInfo` so we can easily target different clusters. Defaults have been added for these functions so that they call the default cluster so existing tests for single cluster testing do not need to be updated.

Fixed issues found while testing: 
* Logging in `NetworkMessagingValidator` was invalid so it was corrected.
* Updated distribution handler to send to active or suspended since the member could have been suspended and we would still need to send them a reduced member list.
* Updated the approval handler when sending registration status to send to active or suspended since suspended members can still register.
* Updated start registration handler to build the pending member info early and publish it to kafka since certain failures trigger the decline handler which was failing because no pending member info existed when sending registration status.
* Updated start registration handler to set registration status for pending member since on initial registration an active version would not exist.
* Allow the MGM to lookup suspended or active members during sync.

Added helpers for testing:
* Disable certificate revocation checks (for when setting up TLS certs using the fake CA)
* Check for existing certificate chain so we can conditionally upload code signing certificates on each cluster.
* Conditionally upload code signing cert (so we can run tests in isolation without needing to run the certificate upload test each time)
* Migrate some existing functions to be cluster context aware (focused on functions necessary for multi cluster testing)

Also updated the testing deploy script to log in to helm once instead of once per cluster. Once per cluster often caused the script to fail to deploy two out of three of the clusters.


components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerTest.kt
dimosr, line 1: I think there's a couple of new behaviours introduced in `SessionManagerImpl` that are not unit tested here now. Although, I acknowledge this class is not the easiest one to unit test and we might want to get this in before Friday (since it has a couple of bug fixes along with the tests). So, happy for this testing to be added as a follow-up, if needed.

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/MemberInfoExtension.kt
```diff
@@ -209,8 +208,8 @@ class MemberInfoExtension {
          * It is preferable to always store this in the member context to avoid the repeated calculation.
          */
         @JvmStatic
-        val MemberInfo.sessionKeysHash: Collection<SecureHash>
-            get() = memberProvidedContext.parseSet<SecureHash>(SESSION_KEYS_HASH).let { storedKeys ->
+        val MemberInfo.sessionKeyHashes: Collection<SecureHash>
```
dimosr, line 207: can you shed some light on what is the motivation behind this change and the one for ledger key hashes?

Did it not really work with the previous constants and it was actually calculating the hashes on the fly?

dimosr (commented):
Code changes look reasonable - a couple of comments/questions.

We should also get a green run of https://github.com/corda/corda-e2e-tests/pull/62 with the alpha version produced by this one before merging, given we are so close to the release deadline.

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/MemberInfoExtension.kt
charlieR3: Yes, it wasn't working and was calculating hashes on the fly despite the hashes being present. This is only since the session key was changed to a list of session keys so it's a fairly recent regression. It didn't cause any failures since the hashes were calculated when missing.

The `parseSet` functionality for the layered property map relies on the key passed in to be the key up to the index value (i.e. `corda.ledger.keys` rather than `corda.ledger.keys.%s.hash`, and `corda.session.keys` rather than `corda.session.keys.%s.hash`). This code previously tried to parse the set using the incorrect key (`corda.session.keys.%s.hash`) for session key hashes so no hashes were found even though they were present.

For the ledger keys I just removed `LEDGER_KEY_HASHES` constant because it's a duplicate of `LEDGER_KEYS` and actually was not ledger key hash specific.

dimosr: that makes sense - thanks for explaining

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/delivery/InMemorySessionReplayer.kt
```diff
@@ -96,16 +96,47 @@ internal class InMemorySessionReplayer(
...
+                        "The message was not replayed."
+            )
+            if(destinationMemberInfo.serial > messageReplay.sessionCounterparties.serial) {
+                logger.warn(
+                    "Replay message is attempting to use ${messageReplay.sessionCounterparties.counterpartyId}" +
+                            " with serial number ${messageReplay.sessionCounterparties.serial}, but the current serial " +
+                            "number ${destinationMemberInfo.serial} is higher so message will never send. " +
+                            "Dropping message replay."
+                )
+                removeMessageFromReplay(messageId, messageReplay.sessionCounterparties)
```
williamvigorr3, line 138: This is an observation (nothing needs to be done). I can see why this was a (potential) problem before 🤦. As we might end up with tasks which get scheduled but never complete. In theory this could cause the Link Manager to slow down (possibly so much that messages are never delivered). If we end up with a lot of tasks which get scheduled forever.

I stared a the code for a while and calling `removeMessageFromReplay` from a thread from the `executorService` in the ReplaySchedular is safe.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -477,6 +477,20 @@ internal class SessionManagerImpl(
                 "${sessionCounterparties.counterpartyId}"
         )
 
+        val responderMemberInfo = membershipGroupReaderProvider.lookup(
+            sessionCounterparties.ourId,
+            sessionCounterparties.counterpartyId,
+            filter
+        )
+        if (responderMemberInfo != null && responderMemberInfo.serial > sessionCounterparties.serial) {
```
williamvigorr3, line 485: We should cover this in a unit test (in a follow up ticket).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/delivery/InMemorySessionReplayer.kt
```diff
@@ -96,16 +96,47 @@ internal class InMemorySessionReplayer(
...
+                        " for session with ID ${messageReplay.sessionId}" +
+                        " between ${messageReplay.sessionCounterparties.ourId}" +
+                        " and peer ${messageReplay.sessionCounterparties.counterpartyId} with " +
+                        "serial ${messageReplay.sessionCounterparties.serial} " +
+                        "which is not in the members map. Member was found but with serial " +
+                        "${destinationMemberInfo.serial}. " +
+                        "The message was not replayed."
+            )
+            if(destinationMemberInfo.serial > messageReplay.sessionCounterparties.serial) {
+                logger.warn(
```
williamvigorr3, line 132: If this can happen (which sounds likely) maybe we should log at a trace level and not at a warning level. This could be done in a follow up Jira.

williamvigorr3 (commented):
I looked at the P2P changes + suspension (implementation).

dickon (approved):
LGTM from crypto side

seanbrereton (approved):
LGTM(build only)

williamvigorr3 (approved):
LGTM

dimosr (approved):

PR #3604, opened 2023-04-12
CORE-12422 Fixed external messaging topic issue

owenstanford:
This change fixes an issue that prevented a flow from seeing newly created, external messaging topics.


filipesoliveira (approved):
LGTM

PR #3606, opened 2023-04-13
CORE-12423: Adjust aprove registration to keep the member status

yift-r3:
Make sure that a suspended member status will not get overwritten.

### DIFF
--- a/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/UpdateMemberAndRegistrationRequestToApprovedHandler.kt
+++ b/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/UpdateMemberAndRegistrationRequestToApprovedHandler.kt
@@ -16,6 +16,7 @@ import net.corda.membership.impl.persistence.service.handler.RegistrationStatusH
 import net.corda.membership.impl.persistence.service.handler.RegistrationStatusHelper.toStatus
 import net.corda.membership.lib.exceptions.MembershipPersistenceException
 import net.corda.membership.lib.MemberInfoExtension.Companion.MEMBER_STATUS_ACTIVE
+import net.corda.membership.lib.MemberInfoExtension.Companion.MEMBER_STATUS_SUSPENDED
 import net.corda.membership.lib.MemberInfoExtension.Companion.STATUS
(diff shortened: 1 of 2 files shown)


charlieR3 (dismissed):

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/UpdateMemberAndRegistrationRequestToApprovedHandler.kt
```diff
@@ -106,4 +99,22 @@ internal class UpdateMemberAndRegistrationRequestToApprovedHandler(
             )
         }
     }
+
+    private fun processMgmContext(
```
dimosr: isn't the status at this point always going to be `PENDING` given we query for the pending record (effectively always switching it to `ACTIVE`)?

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/UpdateMemberAndRegistrationRequestToApprovedHandlerTest.kt
```diff
@@ -229,6 +230,90 @@ class UpdateMemberAndRegistrationRequestToApprovedHandlerTest {
         )
     }
 
+    @Test
+    fun `invoke will keep the state if it was suspended`() {
```
dimosr, line 234: Do we have a test that exercises the scenario where no entity is returned for the query that searches for non-pending entries (ACTIVE or SUSPENDED)?

dimosr (commented):
Logic looks good to me now, just a comment on the tests

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/UpdateMemberAndRegistrationRequestToApprovedHandlerTest.kt
yift-r3: In the test above it (`invoke will update the MGM context members`) the search is returning null

dimosr (approved):

PR #3610, opened 2023-04-13
CORE-9790: helm corda-lib support for Vault, and tidy up

dickon:
Support Hashicorp Vault for config secrets if it is configured, while preserving the Corda-native configruation system for non-Vault users. Tested for both Corda Open Source, and with https://github.com/corda/corda5-enterprise/pull/38 for Corda Enterprise. The enterprise/Vault scenario requires careful handling of secrets.

Add a sequence number prefix to each init container, so that they are listed in the right order and the ordering requirement is clear, since this initially confused me. This seems like a nice usability benefit for operators.

Add a short script to deploy to kubernetes in one go, respecting the prereqs chart and the requirement to delete namespaces.

In my earlier change I was careful to preserve all the idiosyncancies of the Helm output, so I knew that fiddly refactoring would have no impact on the Kubernetes resources that helm specifies. Now that's merged, we no longer need to preserve the exact whitespace and inconsistent use of quoting, so that allows the Helm chart to be simplified.

### DIFF
--- a/charts/corda-lib/templates/_bootstrap.tpl
+++ b/charts/corda-lib/templates/_bootstrap.tpl
@@ -379,7 +379,7 @@ a second init container to execute the output SQL to the relevant database
 
 {{- define "corda.generateAndExecuteSql" -}}
 {{- /* define 2 init containers, which run in sequence. First run corda-cli initial-config to generate some SQL, storing in a persistent volume called working-volume. Second is a postgres image which mounts the same persistent volume and executes the SQL. */ -}}  
-- name: {{ printf "%02d" .sequenceNumber }}-create-{{ .name }}
+- name: {{ printf "%02d-create-%s" .sequenceNumber .name }}
   image: {{ include "corda.bootstrapCliImage" . }}
   imagePullPolicy: {{ .Values.imagePullPolicy }}
   {{- include "corda.bootstrapResources" . | nindent 2 }}
@@ -456,7 +456,7 @@ a second init container to execute the output SQL to the relevant database
     {{- if eq .environmentVariablePrefix "CRYPTO_DB_USER" -}}
       {{- include "corda.cryptoDbUserEnv" . | nindent 4 -}}
     {{- end }}
-- name: {{ printf "%02d" (add .sequenceNumber 1)}}-apply-{{ .name }}
+- name: {{ printf "%02d-apply-%s" (add .sequenceNumber 1) .name }}
   image: {{ include "corda.bootstrapDbClientImage" . }}
   imagePullPolicy: {{ .Values.imagePullPolicy }}
   {{- include "corda.bootstrapResources" . | nindent 2 }}
--- a/charts/corda-lib/templates/_bootstrap.tpl
+++ b/charts/corda-lib/templates/_bootstrap.tpl
@@ -72,24 +72,6 @@ spec:
           {{ include "corda.cryptoDbUserEnv" . | nindent 12 }}
           {{- include "corda.clusterDbEnv" . | nindent 12 }}
         {{- include "corda.generateAndExecuteSql" ( dict "name" "crypto-config" "subCommand" "create-crypto-config" "Values" .Values "Chart" .Chart "Release" .Release "schema" "CRYPTO" "namePostfix" "worker-config" "sqlFile" "crypto-config.sql" "sequenceNumber" 12) | nindent 8 }}
-        {{- if (((.Values).config).vault).url }} 
-        - name: 14-apply-db-secrets-to-vault
-          image: curlimages/curl:8.00.1
-          imagePullPolicy: {{ .Values.imagePullPolicy }}
-          {{- include "corda.bootstrapResources" . | nindent 10 }}
-          {{- include "corda.containerSecurityContext" . | nindent 10 }}
-          args: [ '--retry', '5', '--retry-max-time', '30', '--retry-all-errors', '--header', 'X-Vault-Token: {{ .Values.config.vault.token }}', '--request', 'POST', '--data', '{ "data": {"crypto": "$(CRYPTO_DB_USER_PASSWORD)", "vnodes": "$(DB_CLUSTER_PASSWORD)", "rbac": "$(RBAC_DB_USER_PASSWORD)"} }', '{{ .Values.config.vault.url }}/v1/secret/data/dbsecrets', '--verbose']
-          env:
-          {{ include "corda.rbacDbUserEnv" . | nindent 12 }}
-          {{ include "corda.cryptoDbUserEnv" . | nindent 12 }}
-          {{- include "corda.clusterDbEnv" . | nindent 12 }}
-        - name: 15-apply-crypto-secrets-to-vault
-          image: curlimages/curl:8.00.1
-          imagePullPolicy: {{ .Values.imagePullPolicy }}
-          {{- include "corda.bootstrapResources" . | nindent 10 }}
-          {{- include "corda.containerSecurityContext" . | nindent 10 }}
-          args: [ '--retry', '5', '--retry-max-time', '30', '--retry-all-errors', '--header', 'X-Vault-Token: {{ .Values.config.vault.token }}', '--request', 'POST', '--data', '{ "data": {"passphrase": "6SIgbxSKt6WpBcSifBUy4geKOGyyVDMIELQ/uj8k2P4=", "salt": "URSKJ/t6xNd/hu3XmQ6q57BVq3R0DNgQn75Hz/2SPf0="} }', '{{ .Values.config.vault.url }}/v1/secret/data/cryptosecrets', '--verbose']
-        {{- end }}
       volumes:
         - name: working-volume
           emptyDir: {}
--- a/charts/corda-lib/templates/_bootstrap.tpl
+++ b/charts/corda-lib/templates/_bootstrap.tpl
@@ -45,7 +45,7 @@ spec:
         {{- include "corda.generateAndExecuteSql" ( dict "name" "rbac" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "RBAC_DB_USER" "schema" "RBAC" "sequenceNumber" 3) | nindent 8 }}
         {{- include "corda.generateAndExecuteSql" ( dict "name" "vnodes" "longName" "virtual-nodes" "dbName" "rbac" "admin" "true" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "DB_CLUSTER" "sequenceNumber" 5) | nindent 8 }}
         {{- include "corda.generateAndExecuteSql" ( dict "name" "crypto" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "CRYPTO_DB_USER"  "schema" "CRYPTO" "sequenceNumber" 7) | nindent 8 }}                
-        {{- include "corda.generateAndExecuteSql" ( dict "name" "rpc"  "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "REST_API_ADMIN"  "schema" "RBAC"  "searchPath" "RBAC" "subCommand" "create-user-config" "namePostfix" "admin" "sqlFile" "rbac-config.sql" "sequenceNumber" 9) | nindent 8 }}
+        {{- include "corda.generateAndExecuteSql" ( dict "name" "rest"  "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "REST_API_ADMIN"  "schema" "RBAC"  "searchPath" "RBAC" "subCommand" "create-user-config" "namePostfix" "admin" "sqlFile" "rbac-config.sql" "sequenceNumber" 9) | nindent 8 }}
         - name: 11-create-db-users-and-grant
           image: {{ include "corda.bootstrapDbClientImage" . }}
           imagePullPolicy: {{ .Values.imagePullPolicy }}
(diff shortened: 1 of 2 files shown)


charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -408,9 +408,12 @@ a second init container to execute the output SQL to the relevant database
          {{- end -}}         
          
          {{- if not (eq .name "rpc") -}}
-             {{- "'--salt'" -}}, "$(SALT)", '--passphrase', "$(PASSPHRASE)", 
+           {{- if and (hasKey .Values "config") (hasKey .Values.config "vault") (hasKey .Values.config.vault "url") -}}
```
davidcurrie: I believe this is dangerous as it relies on you being on a version of Helm that is built with a version of go that supported lazy evaluation in templates. There's a hack to handle the chained nulls which is to introduce parentheses e.g. `if ((.Values.config).vault).url`

davidcurrie (changes requested):

charts/corda-lib/templates/_bootstrap.tpl
dickon: Okay, I didn't know that was maybe unsupported sometimes. Fixed.

kubernetes-deploy.sh
```diff
@@ -0,0 +1,26 @@
+#! /bin/zsh
```
simon-johnson-r3: This is a mac specific script, I'm not sure how popular it will be to commit it to git to the OS repo. It means someone has to maintain it always, and there's no test for that. And we'd probably want equivalents for Linux and Windows. I guess at the very least it would have to be named appropriately for the platform.

I wonder if this is one of those cases where whilst clearly useful, it's probably easier to maintain these some other way outside the repo for internal developer use then get into the wider discussion?

I note the upcoming hackathon has "Tools for Corda" as one of its topics, perhaps this would be a good time to assess stuff like this?

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -71,8 +71,25 @@ spec:
           {{ include "corda.rbacDbUserEnv" . | nindent 12 }}
           {{ include "corda.cryptoDbUserEnv" . | nindent 12 }}
           {{- include "corda.clusterDbEnv" . | nindent 12 }}
-        {{- include "corda.generateAndExecuteSql" ( dict "name" "crypto" "subCommand" "create-crypto-config" "Values" .Values "Chart" .Chart "Release" .Release "quoteUser" "true" "quotePassword" "false" "schema" "CRYPTO" "namePostfix" "worker-config" "sqlFile" "crypto-config.sql") | nindent 8 }}
-
+        {{- include "corda.generateAndExecuteSql" ( dict "name" "crypto-config" "subCommand" "create-crypto-config" "Values" .Values "Chart" .Chart "Release" .Release "schema" "CRYPTO" "namePostfix" "worker-config" "sqlFile" "crypto-config.sql" "sequenceNumber" 12) | nindent 8 }}
+        {{- if (((.Values).config).vault).url }}
```
simon-johnson-r3: I assume the brackets are necessary here?

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -41,12 +41,12 @@ spec:
...
-        {{- include "corda.generateAndExecuteSql" ( dict "name" "vnodes" "longName" "virtual-nodes" "dbName" "rbac" "admin" "true" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "DB_CLUSTER") | nindent 8 }}
-        {{- include "corda.generateAndExecuteSql" ( dict "name" "crypto" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "CRYPTO_DB_USER" "quoteUser" "true" "quotePassword" "false" "schema" "CRYPTO") | nindent 8 }}                
-        {{- include "corda.generateAndExecuteSql" ( dict "name" "rpc"  "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "REST_API_ADMIN" "quoteUser" "true" "quotePassword" "false" "schema" "RBAC"  "searchPath" "RBAC" "subCommand" "create-user-config" "namePostfix" "admin" "sqlFile" "rbac-config.sql") | nindent 8 }}
-        - name: create-db-users-and-grant
+        {{- include "corda.generateAndExecuteSql" ( dict "name" "db" "Values" .Values "Chart" .Chart "Release" .Release "schema" "RBAC" "namePostfix" "schemas" "sequenceNumber" 1) | nindent 8 }}
```
simon-johnson-r3, line 44: Sequence numbers might be inconvenient to maintain as inserting steps will mean changing multiple lines.
I see we already have name, do these really add anything?

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -71,8 +71,25 @@ spec:
           {{ include "corda.rbacDbUserEnv" . | nindent 12 }}
           {{ include "corda.cryptoDbUserEnv" . | nindent 12 }}
           {{- include "corda.clusterDbEnv" . | nindent 12 }}
-        {{- include "corda.generateAndExecuteSql" ( dict "name" "crypto" "subCommand" "create-crypto-config" "Values" .Values "Chart" .Chart "Release" .Release "quoteUser" "true" "quotePassword" "false" "schema" "CRYPTO" "namePostfix" "worker-config" "sqlFile" "crypto-config.sql") | nindent 8 }}
-
+        {{- include "corda.generateAndExecuteSql" ( dict "name" "crypto-config" "subCommand" "create-crypto-config" "Values" .Values "Chart" .Chart "Release" .Release "schema" "CRYPTO" "namePostfix" "worker-config" "sqlFile" "crypto-config.sql" "sequenceNumber" 12) | nindent 8 }}
+        {{- if (((.Values).config).vault).url }} 
+        - name: 14-apply-db-secrets-to-vault
```
simon-johnson-r3: These 2 curl commands were in my PoC but we agreed to remove them and instead have them pushed by the CI pipeline. They should be deleted from here.

simon-johnson-r3 (changes requested):

charts/corda-lib/templates/_bootstrap.tpl
dickon: We should still be able to support efficient local installation of Enterprise Corda, and in particular the Vault support.

charts/corda-lib/templates/_bootstrap.tpl
dickon: Without sequence numbers, it is not clear that there's a strict ordering with sequential dependencies  on these init containers. That confused me for a bit. Both `k9s` and `kubectl describe pod` sort the init containers by name, which obscures the sequential ordering and makes it look like there are no dependencies.

With the sequence numbers, I think it is more clear in `k9s` and the other tools which following the sorting convention what's happening:

Screenshot 2023-04-13 at 17 09 15

and if something does go wrong you can tell at a glance which it is.

I do agree there's a little overhead in maintaining the numbering, but since operating the helm chart and looking at the current status is much more common than authoring it we should optimise for reading.

charts/corda-lib/templates/_bootstrap.tpl
dickon: They are. @davidcurrie believes some versions of helm do not have short-circuit evaluation logic, but for an obscure reason this structure stops Helm printing an error if vault is not mentioined in the Values.

davidcurrie: Obscure it certainly is. See https://stackoverflow.com/questions/59795596/how-to-make-nested-variables-optional-in-helm for the details.

charts/corda-lib/templates/_bootstrap.tpl
simon-johnson-r3: This is as per https://r3holdco.slack.com/archives/C0527PCBKTR/p1681214437759929 if you want to deviate from this plan, you should probably take it to that channel again.

dickon: This is gone now.

simon-johnson-r3 (dismissed):

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -41,12 +41,12 @@ spec:
...
+        {{- include "corda.generateAndExecuteSql" ( dict "name" "rbac" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "RBAC_DB_USER" "schema" "RBAC" "sequenceNumber" 3) | nindent 8 }}
+        {{- include "corda.generateAndExecuteSql" ( dict "name" "vnodes" "longName" "virtual-nodes" "dbName" "rbac" "admin" "true" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "DB_CLUSTER" "sequenceNumber" 5) | nindent 8 }}
+        {{- include "corda.generateAndExecuteSql" ( dict "name" "crypto" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "CRYPTO_DB_USER"  "schema" "CRYPTO" "sequenceNumber" 7) | nindent 8 }}                
+        {{- include "corda.generateAndExecuteSql" ( dict "name" "rpc"  "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "REST_API_ADMIN"  "schema" "RBAC"  "searchPath" "RBAC" "subCommand" "create-user-config" "namePostfix" "admin" "sqlFile" "rbac-config.sql" "sequenceNumber" 9) | nindent 8 }}
```
davidcurrie: not your doing but this should probably be called `rest` now rather than `rpc`

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -380,7 +379,7 @@ a second init container to execute the output SQL to the relevant database
 
 {{- define "corda.generateAndExecuteSql" -}}
 {{- /* define 2 init containers, which run in sequence. First run corda-cli initial-config to generate some SQL, storing in a persistent volume called working-volume. Second is a postgres image which mounts the same persistent volume and executes the SQL. */ -}}  
-- name: create-{{- if not (eq .name "db") -}}initial-{{- end -}}{{ .name }}-{{ .namePostfix | default "db-config" }}
+- name: {{ printf "%02d" .sequenceNumber }}-create-{{ .name }}
```
davidcurrie: nit
```suggestion
- name: {{ printf "%02d-create-%s" .sequenceNumber .name }}
```

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -447,19 +452,11 @@ a second init container to execute the output SQL to the relevant database
...
       {{- include "corda.cryptoDbUserEnv" . | nindent 4 -}}
-    {{- end -}}
-    
-    {{- /* TODO remove this special case, it is just that the old template has these declarations later */ -}}
-    {{- if (eq .name "rpc") -}}
-      {{- "\n    " -}} {{- /* legacy whitespace compliance */ -}}
-      {{ include "corda.bootstrapCliEnv" . | nindent 4  -}} {{- /* JAVA_TOOL_OPTIONS, CONSOLE_LOG*, CORDA_CLI_HOME_DIR */ -}}
     {{- end }}
-- name: apply-{{- if not (eq .name "db") -}}initial-{{- end -}}{{ .name }}-{{ .namePostfix | default "db-config" }}
+- name: {{ printf "%02d" (add .sequenceNumber 1)}}-apply-{{ .name }}
```
davidcurrie:
```suggestion
- name: {{ printf "%02d-apply-%s" (add .sequenceNumber 1) .name }}
```

kubernetes-deploy.sh
```diff
@@ -0,0 +1,26 @@
...
+kubectl delete namespace corda || echo corda namesapce already deleted?
+kubectl create namespace corda
+
+# Usual prereqs
+helm install prereqs -n corda oci://registry-1.docker.io/corda/corda-dev-prereqs --timeout 10m --wait
+
+# Corda
+helm dependency build charts/corda
+helm upgrade --install corda -n corda charts/corda \
+  --values ../corda-runtime-os/values-prereqs.yaml \
```
davidcurrie: if the script has to be run from the root of `corda-runtime-os`, why the path?

kubernetes-deploy.sh
```diff
@@ -0,0 +1,26 @@
...
+# Currently the init containers are not repeatble, so ensure we start from scratch each time
+kubectl delete namespace corda || echo corda namesapce already deleted?
+kubectl create namespace corda
+
+# Usual prereqs
+helm install prereqs -n corda oci://registry-1.docker.io/corda/corda-dev-prereqs --timeout 10m --wait
+
+# Corda
+helm dependency build charts/corda
+helm upgrade --install corda -n corda charts/corda \
```
davidcurrie: nit: shouldn't matter given we're cleaning out the namespace anyway but just to make a point...
```suggestion
helm install corda -n corda charts/corda \
```

kubernetes-deploy.sh
```diff
@@ -0,0 +1,26 @@
+#! /bin/bash
+
+# Set up Corda on Kubernetes; must be run from the root of corda-runtime-os
+
+set -e
+
+./gradlew publishOSGiImage
```
davidcurrie, line 10: This is only going to work for Docker Desktop users. Do we want to try and detect if they're using minikube and do the `eval $(minikube docker-env)` if so? Not doing it is a common reason for failed deployments for minikube users. If nothing else, maybe it should be in the script but commented out?

kubernetes-deploy.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Currently the init containers are not repeatble, so ensure we start from scratch each time
```
davidcurrie:
```suggestion
# Currently the init containers are not repeatable, so ensure we start from scratch each time
```

davidcurrie (commented):
nothing all that substantive...

charts/corda-lib/templates/_bootstrap.tpl
dickon: changed

charts/corda-lib/templates/_bootstrap.tpl
dickon: done

kubernetes-deploy.sh
dickon: I wondered about that (inherited that logic from Simon). Maybe it's because it gives a nicer error if the charts cannot be found but that's tenuous. Simplified.

kubernetes-deploy.sh
dickon: done

kubernetes-deploy.sh
dickon: done

kubernetes-deploy.sh
dickon: Added commented out

charts/corda-lib/templates/_bootstrap.tpl
dickon: done

kubernetes-deploy.sh
dickon: done

davidcurrie (approved):
LGTM

PR #3611, opened 2023-04-13
CORE-12304 Cascade Intermittent Unavailability Caused by DB Worker Restart

kyriathar:
It seems there was a synchronization issue between `ConfigurationReadService` and Configuration reconciler. `ConfigurationReadService` holds a cache for caching the Configuration on Kafka. That cache is used by the Configuration reconciler to find the delta between Configuration on Kafka and Configuration in the database and then update Kafka with that delta. After adding Avro schema registry updating mechanism through Kafka as part of `ConfigurationReadService` bootstrapping process, it means `ConfigurationReadService` is now ready at a later point in time. So now, on DB worker restart the Configuration reconciler reader which is `ConfigurationReadService` has not yet read the Kafka Configuration, however marked as `UP`, which leads to the reconciler understand an empty Configuration on Kafka and therefore running a complete reconciliation with the database.

- Signals `ConfigurationReadService` as `UP` only after its `ConfigProcessor.onSnapshot` has run to make sure its cache has been populated with Kafka Configuration before reconciler uses it

### DIFF
--- a/components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/ConfigReadServiceEventHandler.kt
+++ b/components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/ConfigReadServiceEventHandler.kt
@@ -86,7 +86,7 @@ internal class ConfigReadServiceEventHandler(
                 setupConfigSubscription(coordinator)
             }
 
-            is ConfigOnSnapshotRun -> {
+            is InitialConfigRead -> {
                 // Only set LifecycleStatus.UP for coordinator after the initial config is read from Kafka.
                 // This is to make sure config reconciler will not start reconciling before its
                 // kafka reader (config read service) has read the initial config from Kafka, otherwise
--- a/components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/ConfigReadServiceEventHandler.kt
+++ b/components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/ConfigReadServiceEventHandler.kt
@@ -89,9 +89,10 @@ internal class ConfigReadServiceEventHandler(
                 }
                 publishAvroSchemas()
                 if (event.isInitialConfig) {
-                    // Only set LifecycleStatus.UP for coordinator after the config subscription onSnapshot
-                    // has run (initial config). This is to make sure config reconciler will not start
-                    // reconciling before its kafka reader (config read service) has read the initial config from Kafka.
+                    // Only set LifecycleStatus.UP for coordinator after the initial config is read from Kafka.
+                    // This is to make sure config reconciler will not start reconciling before its
+                    // kafka reader (config read service) has read the initial config from Kafka, otherwise
+                    // the config reconciler understands empty Kafka and tries to reconcile the full DB config.
                     coordinator.updateStatus(LifecycleStatus.UP)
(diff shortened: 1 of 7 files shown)


components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/ConfigReadServiceEventHandler.kt
```diff
@@ -148,7 +145,6 @@ internal class ConfigReadServiceEventHandler(
             avroSchemaProcessor,
             configMerger.getMessagingConfig(config, null)
         )
-        avroSubReg = coordinator.followStatusChangesByName(setOf(sub.subscriptionName))
```
simon-johnson-r3: by removing these, this component's UP and DOWN status is no longer tethered to the UP and DOWN status of the subscriptions, I'm not sure this is what you really want is it?

kyriathar: Good catch, yeah no longer needed the `UP` for it, as we now intend to mark config read service as `UP` after `ConfigProcessor.onSnapshot` has run, but the `DOWN`s/ `ERROR`s are now disconnected indeed.

kyriathar: Brought back following the subscriptions to follow their `DOWN`s/ `ERROR`s

libs/lifecycle/lifecycle-test-impl/src/main/kotlin/net/corda/lifecycle/test/impl/LifecycleTest.kt
```diff
@@ -228,17 +228,21 @@ class LifecycleTest<T : Lifecycle>(
      * @param dependency the coordinator name of the dependency to toggle
      * @param verificationWhenDown can be used to assert expectations for once the dependency is down
      * @param verificationWhenUp can be used to assert expectations for once the dependency is back up
+     * @param onUppingDependency additional actions after [dependency] going up
      */
     fun toggleDependency(
         dependency: LifecycleCoordinatorName,
         verificationWhenDown: () -> Unit = {},
         verificationWhenUp: () -> Unit = {},
+        onUppingDependency: () -> Unit = {}
```
fowlerrr: I think `onDependencyUp` would be clearer, though I'm not sure we need yet-another-lambda here.  Probably better to repurpose `verificationWhenUp` and change the names to something that means "do verification and/or other tasks".  No idea what that name would be though... 🤔

fowlerrr (changes requested):
I think the bulk of the change is okay but I'm not comfortable with that change to `LifecycleTest`.  I think we need to keep the lambdas from getting out of control

libs/lifecycle/lifecycle-test-impl/src/main/kotlin/net/corda/lifecycle/test/impl/LifecycleTest.kt
kyriathar: Renamed as proposed and also introduced `onDependencyDown` as discussed offline

fowlerrr (approved):

kyriathar:
Closing as we already have the above behavior through `CompactedSubscriptionImpl`

PR #3612, opened 2023-04-13
CORE-12465 - Add missing override to CpiMetadataEntityKey

driessamyn:


libs/virtual-node/cpi-datamodel/src/main/kotlin/net/corda/libs/cpi/datamodel/entities/CpiMetadataEntity.kt
```diff
@@ -188,7 +188,25 @@ class CpiMetadataEntityKey(
     private val name: String,
     private val version: String,
     private val signerSummaryHash: String,
-) : Serializable
+) : Serializable {
```
chrisr3, line 191: Just a thought but would a `data` class work here too?

kyriathar (approved):

libs/virtual-node/cpi-datamodel/src/main/kotlin/net/corda/libs/cpi/datamodel/entities/CpiMetadataEntity.kt
kyriathar: I think we didn't want data class for entities but then this isn't really so I think Chris is right.

driessamyn: I think in this instance it doesn't matter, but following the guidelines with set ourselves (https://github.com/corda/platform-eng-design/blob/master/core/corda-5/corda-5.0/standards/jpa.md#kotlin-data-classes), it's easier to reason with a "don't use data classes" rule.

PR #3613, opened 2023-04-13
CORE-12445 Flow External Messaging - Support route configuration migration during virtual node upgrade

filipesoliveira:
Added support to route configuration migration during virtual node upgrade. 
Added tests to ensure the upgrade works as expected.

### DIFF
--- a/components/virtual-node/virtual-node-write-service-impl/src/test/kotlin/net/corda/virtualnode/write/db/impl/tests/writer/asyncoperation/handlers/CreateVirtualNodeOperationHandlerTest.kt
+++ b/components/virtual-node/virtual-node-write-service-impl/src/test/kotlin/net/corda/virtualnode/write/db/impl/tests/writer/asyncoperation/handlers/CreateVirtualNodeOperationHandlerTest.kt
@@ -15,6 +15,7 @@ import net.corda.messaging.api.publisher.Publisher
 import net.corda.messaging.api.records.Record
 import net.corda.schema.Schemas.VirtualNode.VIRTUAL_NODE_OPERATION_STATUS_TOPIC
 import net.corda.v5.membership.MemberInfo
+import net.corda.virtualnode.HoldingIdentity
 import net.corda.virtualnode.write.db.impl.tests.ALICE_HOLDING_ID1
 import net.corda.virtualnode.write.db.impl.tests.CPI_CHECKSUM1
 import net.corda.virtualnode.write.db.impl.tests.CPI_IDENTIFIER1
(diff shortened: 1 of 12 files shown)


libs/external-messaging/src/main/kotlin/net/corda/libs/external/messaging/ExternalMessagingRouteConfigGenerator.kt
```diff
@@ -3,7 +3,18 @@ package net.corda.libs.external.messaging
...
 import net.corda.virtualnode.HoldingIdentity
+import net.corda.virtualnode.VirtualNodeInfo
 
 interface ExternalMessagingRouteConfigGenerator {
-    fun generateConfig(holdingId: HoldingIdentity, cpiId: CpiIdentifier, cpks: Collection<CpkMetadata>): String
+    /**
+     * The method generates the configuration for external messaging.
+     * N.B.: Please keep in mind that any historical configuration is lost when this method is called.
+     */
+    fun generateConfig(holdingId: HoldingIdentity, cpiId: CpiIdentifier, cpks: Collection<CpkMetadata>): String?
```
owenstanford: Perhaps we should name the functions inline with the config they are generating e.g. `generateNewConfig` `generateUpgradeConfig`, to avoid confusion.

components/virtual-node/virtual-node-write-service-impl/src/test/kotlin/net/corda/virtualnode/write/db/impl/tests/writer/asyncoperation/handlers/CreateVirtualNodeOperationHandlerTest.kt
```diff
@@ -59,7 +62,7 @@ class CreateVirtualNodeOperationHandlerTest {
     }
 
     private val externalMessagingRouteConfigGenerator = mock<ExternalMessagingRouteConfigGenerator>().apply {
-        whenever(generateConfig(any(),any(),any())).thenReturn(externalMessagingRouteConfig)
+        whenever(generateConfig(isA<HoldingIdentity>(),any(),any())).thenReturn(externalMessagingRouteConfig)
```
owenstanford: See comment below, I think if we rename the methods it will avoid the need for isA

libs/external-messaging/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
owenstanford, line 1: what is this used for?

filipesoliveira: Mockito requires this file to be able to mock classes marked as final. All Kotlin classes are final by default.

filipesoliveira: Ref: https://www.baeldung.com/mockito-final

owenstanford (approved):
LGTM

PR #3617, opened 2023-04-13
CORE-8956 check notary whitelisted

vlajos:
* CORE-8956 check if notary is whitelisted. It gets triggered in both sides of finality and also on tx creation time.
* Fix notary comparison in `verifyInputNotaries()`. (Public keys are not checked anymore, to support key rotation.)

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
@@ -45,7 +45,8 @@ class UtxoLedgerTransactionVerificationServiceImpl @Activate constructor(
     @Suspendable
     override fun verify(transaction: UtxoLedgerTransaction) {
 
-        verifyNotary(transaction)
+        val signedGroupParameters = fetchAndVerifySignedGroupParameters(transaction)
+        verifyNotaryAllowed(transaction, signedGroupParameters)
 
         val verificationResult = externalEventExecutor.execute(
             TransactionVerificationExternalEventFactory::class.java,
@@ -66,13 +67,7 @@ class UtxoLedgerTransactionVerificationServiceImpl @Activate constructor(
     }
 
     @Suspendable
-    private fun verifyNotary(transaction: UtxoLedgerTransaction) {
-        val signedGroupParameters = fetchSignedGroupParameters(transaction)
-        verifyNotaryAllowed(transaction, signedGroupParameters)
-    }
-
-    @Suspendable
-    private fun fetchSignedGroupParameters(transaction: UtxoLedgerTransaction): SignedGroupParameters {
+    private fun fetchAndVerifySignedGroupParameters(transaction: UtxoLedgerTransaction): SignedGroupParameters {
         val membershipGroupParametersHashString =
             (transaction.metadata as TransactionMetadataInternal).getMembershipGroupParametersHash()
         requireNotNull(membershipGroupParametersHashString) {
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
@@ -11,6 +11,7 @@ import net.corda.ledger.utxo.flow.impl.persistence.UtxoLedgerGroupParametersPers
 import net.corda.ledger.utxo.flow.impl.transaction.verifier.external.events.TransactionVerificationExternalEventFactory
 import net.corda.ledger.utxo.flow.impl.transaction.verifier.external.events.TransactionVerificationParameters
 import net.corda.ledger.utxo.transaction.verifier.SignedGroupParametersVerifier
+import net.corda.membership.lib.SignedGroupParameters
 import net.corda.sandbox.type.SandboxConstants.CORDA_SYSTEM_SERVICE
 import net.corda.sandbox.type.UsedByFlow
 import net.corda.v5.application.serialization.SerializationService
@@ -24,8 +25,8 @@ import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
 import java.nio.ByteBuffer
 
 @Component(
-    service = [ UtxoLedgerTransactionVerificationService::class, UsedByFlow::class ],
-    property = [ CORDA_SYSTEM_SERVICE ],
+    service = [UtxoLedgerTransactionVerificationService::class, UsedByFlow::class],
+    property = [CORDA_SYSTEM_SERVICE],
     scope = PROTOTYPE
 )
 class UtxoLedgerTransactionVerificationServiceImpl @Activate constructor(
@@ -43,24 +44,9 @@ class UtxoLedgerTransactionVerificationServiceImpl @Activate constructor(
 
     @Suspendable
     override fun verify(transaction: UtxoLedgerTransaction) {
-        val membershipGroupParametersHashString =
-            (transaction.metadata as TransactionMetadataInternal).getMembershipGroupParametersHash()
-        // todo CORE-8956, CORE-8958 requireNotNull
-        if (membershipGroupParametersHashString != null) {
-            val currentGroupParameters = currentGroupParametersService.get()
-            val signedGroupParameters =
-                if (currentGroupParameters.hash.toString() == membershipGroupParametersHashString) {
-                    currentGroupParameters
-                } else {
-                    val membershipGroupParametersHash = parseSecureHash(membershipGroupParametersHashString)
-                    utxoLedgerGroupParametersPersistenceService.find(membershipGroupParametersHash)
-                }
-            signedGroupParametersVerifier.verify(
-                transaction,
-                signedGroupParameters,
-                currentGroupParametersService.getMgmKeys()
-            )
-        }
+
+        verifyNotary(transaction)
+
         val verificationResult = externalEventExecutor.execute(
             TransactionVerificationExternalEventFactory::class.java,
             TransactionVerificationParameters(
@@ -77,6 +63,37 @@ class UtxoLedgerTransactionVerificationServiceImpl @Activate constructor(
                 verificationResult.errorMessage
             )
         }
+    }
+
+    private fun verifyNotary(transaction: UtxoLedgerTransaction) {
+        val signedGroupParameters = fetchSignedGroupParameters(transaction)
+        verifyNotaryAllowed(transaction, signedGroupParameters)
+    }
+
+    private fun fetchSignedGroupParameters(transaction: UtxoLedgerTransaction): SignedGroupParameters {
+        val membershipGroupParametersHashString =
+            (transaction.metadata as TransactionMetadataInternal).getMembershipGroupParametersHash()
+        requireNotNull(membershipGroupParametersHashString) {
+            "Membership group parameters hash cannot be found in the transaction metadata."
+        }
+
+        val currentGroupParameters = currentGroupParametersService.get()
+        val signedGroupParameters =
+            if (currentGroupParameters.hash.toString() == membershipGroupParametersHashString) {
+                currentGroupParameters
+            } else {
+                val membershipGroupParametersHash = parseSecureHash(membershipGroupParametersHashString)
+                utxoLedgerGroupParametersPersistenceService.find(membershipGroupParametersHash)
+            }
+        requireNotNull(signedGroupParameters) {
+            "Signed group parameters $membershipGroupParametersHashString related to the transaction ${transaction.id} cannot be accessed."
+        }
+        signedGroupParametersVerifier.verify(
+            transaction,
+            signedGroupParameters,
+            currentGroupParametersService.getMgmKeys()
+        )
+        return signedGroupParameters
     }
 
     private fun UtxoLedgerTransaction.toContainer() =
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
@@ -65,11 +65,13 @@ class UtxoLedgerTransactionVerificationServiceImpl @Activate constructor(
         }
     }
 
+    @Suspendable
     private fun verifyNotary(transaction: UtxoLedgerTransaction) {
         val signedGroupParameters = fetchSignedGroupParameters(transaction)
         verifyNotaryAllowed(transaction, signedGroupParameters)
     }
 
+    @Suspendable
     private fun fetchSignedGroupParameters(transaction: UtxoLedgerTransaction): SignedGroupParameters {
         val membershipGroupParametersHashString =
             (transaction.metadata as TransactionMetadataInternal).getMembershipGroupParametersHash()
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
@@ -1,5 +1,6 @@
 package net.corda.ledger.utxo.flow.impl.transaction.verifier
 
+import net.corda.crypto.cipher.suite.publicKeyId
 import net.corda.ledger.common.data.transaction.TransactionMetadataInternal
 import net.corda.membership.lib.SignedGroupParameters
 import net.corda.v5.ledger.utxo.transaction.UtxoLedgerTransaction
@@ -24,9 +25,10 @@ fun verifyNotaryAllowed(transaction: UtxoLedgerTransaction, signedGroupParameter
     }
 
     val notaryCandidate = checkNotNull(allowedNotaries.singleOrNull { it.name == transaction.notaryName }) {
-        "Notary of the transaction is not listed in the available notaries."
+        "Notary of the transaction (${transaction.notaryName}) is not listed in the available notaries."
     }
     check(notaryCandidate.publicKey == transaction.notaryKey) {
-        "Notary key of the transaction is not matching against the related notary in Signed Group Parameters."
+        "Notary key of the transaction (${transaction.notaryKey.publicKeyId()} is not matching against " +
+                "the related notary (${notaryCandidate.publicKey.publicKeyId()} in Signed Group Parameters."
     }
 }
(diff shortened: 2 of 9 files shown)


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
```diff
@@ -0,0 +1,24 @@
+package net.corda.ledger.utxo.flow.impl.transaction.verifier
+
+import net.corda.ledger.common.data.transaction.TransactionMetadataInternal
+import net.corda.membership.lib.SignedGroupParameters
+import net.corda.v5.ledger.utxo.transaction.UtxoLedgerTransaction
+
+fun verifyNotaryAllowed(transaction: UtxoLedgerTransaction, signedGroupParameters: SignedGroupParameters) {
```
relyafi, line 18: It feels overkill to create a separate file for one function. Why not have it as part of the service impl class, as we do for many other functions?

If it's because this will then break the tests, then you could use `internal` scope with the `@VisibleForTesting` annotation.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
```diff
@@ -79,6 +65,37 @@ class UtxoLedgerTransactionVerificationServiceImpl @Activate constructor(
         }
     }
 
+    private fun verifyNotary(transaction: UtxoLedgerTransaction) {
+        val signedGroupParameters = fetchSignedGroupParameters(transaction)
+        verifyNotaryAllowed(transaction, signedGroupParameters)
+    }
```
relyafi: I think this is adding little value, and is misleading. You could (and arguably should) verify the group parameters, irrespective of notary verification. The fact that these parameters are an input to verifying the notary is incidental.

I would just inline the body of this function in `verify`.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
```diff
@@ -0,0 +1,24 @@
...
+    val txNotaryPublicKey = transaction.notaryKey
+    val txNotaryName = transaction.notaryName
```
relyafi: Redundant variables as only used once and the intent is fairly clear from the existing attributes.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    val txGroupParametersHash = (transaction.metadata as TransactionMetadataInternal).getMembershipGroupParametersHash()
+    check(txGroupParametersHash == signedGroupParameters.hash.toString()) {
+        "Membership group parameters (${signedGroupParameters.hash}) is not the one associated to the transaction " +
+                " in its metadata ($txGroupParametersHash)."
+    }
+
+    checkNotNull(
+        allowedNotaries.firstOrNull { it.publicKey == txNotaryPublicKey }
+            ?: allowedNotaries.firstOrNull { it.name == txNotaryName }
```
relyafi: Why do we need to fall back to checking on name? Shouldn't there always be a key match?

In the key rotation case we would be validating against historic parameters anyway.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
```diff
@@ -79,6 +65,37 @@ class UtxoLedgerTransactionVerificationServiceImpl @Activate constructor(
...
+            "Membership group parameters hash cannot be found in the transaction metadata."
+        }
+
+        val currentGroupParameters = currentGroupParametersService.get()
+        val signedGroupParameters =
+            if (currentGroupParameters.hash.toString() == membershipGroupParametersHashString) {
+                currentGroupParameters
+            } else {
+                val membershipGroupParametersHash = parseSecureHash(membershipGroupParametersHashString)
+                utxoLedgerGroupParametersPersistenceService.find(membershipGroupParametersHash)
```
mnesbit, line 83: Is this call suspending? As the code here isn't marked with @Suspendable?

relyafi (commented):
A few comments. I also think it would be good (if possible) to write a higher level integration or smoke test, given the importance of this check.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
blsemo: Should not both of these conditions be met? I would expect that the key id of the notary in the GroupParameters from the time when the transaction was created matches the key id on the transaction, and the name definitely has to match.

blsemo (commented):
The code looks fine structurally, but I think the check is wrong.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
vlajos: Yes. Thanks. fixed.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
vlajos: At the moment it is quite simple.
I suspect we will need something more complicated for the inputs/references notary checks.
Probably a service since we'll need to fetch some data from the database too.
Anyway I wanted to keep it separated and simple for now. 8958 will improve this part.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
vlajos: I expect some more meat in this one too. (And probably a rename... :)
Something like verifyNotariesTransitively? Not too sure yet. 8958 should tackle this part too.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
vlajos: https://r3-cev.atlassian.net/browse/CORE-8958's second point defines the notary verification this way.
Is not that correct?

vlajos: https://r3-cev.atlassian.net/browse/CORE-8958's second point defines the notary verification this way.
Is not that correct?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
vlajos: I wanted to make the check shorter, but also cleaner.
Trivial to inline them if you think, it just made it worse.

vlajos:
> A few comments. I also think it would be good (if possible) to write a higher level integration or smoke test, given the importance of this check.

The main Utxo flow smoke test walks through on the happy path.
But I think you are right, we could add some negative checks to the e2e repo.
I have opened this to add more tests:
https://r3-cev.atlassian.net/browse/CORE-12471

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
relyafi: Hmm, ok. Personally, I think it's better to introduce / abstract when we need to, rather than doing so speculatively ahead of time. We don't know when we'll get round to doing that, or how things will evolve by the time we do.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
relyafi: I mean to me, `transaction.notaryKey` and `transaction.NotaryName` are just as clear inline. We're saving a handful of characters where we reference them by doing these, at the expense of adding 2 redundant lines of code.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
relyafi: That's true, although I'm trying to remember the rationale behind that. Was it that retrieving the name from group parameters was expensive, so you'd rather just compare on the keys we have to hand for efficiency?

If we trust the signed blob, then I'd have thought just comparing on one or the other would be ok. And based on the conversation with Matthew, maybe name would be enough.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
mnesbit: The flip side of that is will this ever be run in verify sandbox? If so it won't work?

vlajos: AFAIK verification sandbox does not have too much support for anything db related.
So that would be a big rework anyway.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
vlajos: You are right. I have just updated it. Yesterday was a bit longer than ideal. :)

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
mnesbit: Service is injected only into flow sandbox, so this is safe and I am fretting over nothing

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
mnesbit: where do we check alignment of the notary key with the named notary data in group parameters?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
lankydan: I agree with @relyafi

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
```diff
@@ -79,6 +65,39 @@ class UtxoLedgerTransactionVerificationServiceImpl @Activate constructor(
...
     }
 
+    @Suspendable
+    private fun verifyNotary(transaction: UtxoLedgerTransaction) {
+        val signedGroupParameters = fetchSignedGroupParameters(transaction)
+        verifyNotaryAllowed(transaction, signedGroupParameters)
+    }
+
+    @Suspendable
+    private fun fetchSignedGroupParameters(transaction: UtxoLedgerTransaction): SignedGroupParameters {
```
lankydan: can we call this `fetchAndVerifySignedGroupParameters`

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
vlajos: Thanks. I have just inlined it.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
vlajos: Thanks. I have renamed it.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
vlajos: This function is supposed to do that.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
vlajos: We have reworked this as agreed on call.
I added some kdocs to explain it too.

vlajos: We have reworked this as agreed on call.
I added some kdocs to explain it too.

lankydan (dismissed):
Ok from my perspective.

libs/ledger/ledger-utxo-transaction-verifier/src/test/kotlin/net/corda/ledger/utxo/transaction/verifier/UtxoLedgerTransactionVerifierTest.kt
```diff
@@ -132,32 +134,38 @@ class UtxoLedgerTransactionVerifierTest {
...
-        whenever(inputTransactionState.notaryKey).thenReturn(anotherPublicKeyExample)
         whenever(referenceTransactionState.notaryName).thenReturn(anotherNotaryX500Name)
-        whenever(referenceTransactionState.notaryKey).thenReturn(anotherPublicKeyExample)
         assertThatThrownBy { verifier.verify() }
             .isExactlyInstanceOf(IllegalStateException::class.java)
             .hasMessageContaining("Input and reference states' notaries need to be the same as the UtxoLedgerTransaction's notary")
     }
 
+    @Test
+    fun `does not throw when input and reference states don't have the same notary keys passed into the verification (but the names are still the same)`() {
```
relyafi:
```suggestion
    fun `does not throw when input and reference states have different notary keys passed into verification (names are the same)`() {
```

Means you don't need the long line limit suppression.

relyafi (dismissed):
One trivial comment, otherwise LGTM

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
```diff
@@ -0,0 +1,32 @@
...
+    val allowedNotaries = signedGroupParameters.notaries
+
+    val txGroupParametersHash = (transaction.metadata as TransactionMetadataInternal).getMembershipGroupParametersHash()
+    check(txGroupParametersHash == signedGroupParameters.hash.toString()) {
+        "Membership group parameters (${signedGroupParameters.hash}) is not the one associated to the transaction " +
+                " in its metadata ($txGroupParametersHash)."
+    }
+
+    val notaryCandidate = checkNotNull(allowedNotaries.singleOrNull { it.name == transaction.notaryName }) {
+        "Notary of the transaction is not listed in the available notaries."
```
mnesbit: Would be nice to include name of the not found notary in the error message

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
```diff
@@ -0,0 +1,32 @@
...
+    }
+    check(notaryCandidate.publicKey == transaction.notaryKey) {
+        "Notary key of the transaction is not matching against the related notary in Signed Group Parameters."
```
mnesbit: Worth including the keys, or more compactly the hash id of the keys in the error message to help fault finding

mnesbit (dismissed):
OK

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
vlajos: Thanks. Updated.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoNotaryVerifier.kt
vlajos: Thanks. Updated.

libs/ledger/ledger-utxo-transaction-verifier/src/test/kotlin/net/corda/ledger/utxo/transaction/verifier/UtxoLedgerTransactionVerifierTest.kt
vlajos: Thanks. Updated.

relyafi (approved):
LGTM

PR #3618, opened 2023-04-13
CORE-12441 Disable object referencing in AMQP serialization

vlajos:
Make serialization more stable as a default to support storing use cases.
On longer term, messaging and storing serialization needs to be reworked properly. (https://r3-cev.atlassian.net/browse/CORE-12472)

### DIFF
--- a/components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/AMQPSerializationProvider.kt
+++ b/components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/AMQPSerializationProvider.kt
@@ -1,6 +1,6 @@
 package net.corda.sandbox.serialization.amqp
 
-import net.corda.internal.serialization.AMQP_P2P_CONTEXT
+import net.corda.internal.serialization.AMQP_STORAGE_CONTEXT
 import net.corda.internal.serialization.SerializationServiceImpl
 import net.corda.internal.serialization.amqp.DeserializationInput
 import net.corda.internal.serialization.amqp.SerializationOutput
@@ -77,7 +77,7 @@ class AMQPSerializationProvider @Activate constructor(
         val serializationService = SerializationServiceImpl(
             serializationOutput,
             deserializationInput,
-            AMQP_P2P_CONTEXT.withSandboxGroup(context.sandboxGroup)
+            AMQP_STORAGE_CONTEXT.withSandboxGroup(context.sandboxGroup) //todo double check in CORE-12472
         )
 
         context.putObjectByKey(AMQP_SERIALIZATION_SERVICE, serializationService)
--- a/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
+++ b/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
@@ -17,7 +17,7 @@ import net.corda.serialization.SerializationContext
 val AMQP_STORAGE_CONTEXT = SerializationContextImpl(
         amqpMagic,
         emptyMap(),
-        true,
+        true, // todo should not this be false? CORE-12472
         SerializationContext.UseCase.Storage,
         null,
         AlwaysAcceptEncodingAllowList
--- a/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
+++ b/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
@@ -17,7 +17,7 @@ import net.corda.serialization.SerializationContext
 val AMQP_STORAGE_CONTEXT = SerializationContextImpl(
         amqpMagic,
         emptyMap(),
-        true, // todo should not this be false? CORE-12472
+        false, // todo double check in CORE-12472
         SerializationContext.UseCase.Storage,
         null,
         AlwaysAcceptEncodingAllowList
--- a/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
+++ b/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
@@ -15,10 +15,10 @@ import net.corda.serialization.SerializationContext
  */
 
 val AMQP_STORAGE_CONTEXT = SerializationContextImpl(
-        preferredSerializationVersion = amqpMagic,
-        properties = emptyMap(),
-        objectReferencesEnabled = false, // todo double check in CORE-12472
-        useCase = SerializationContext.UseCase.Storage,
-        encoding = null,
-        encodingAllowList = AlwaysAcceptEncodingAllowList
+    preferredSerializationVersion = amqpMagic,
+    properties = emptyMap(),
+    objectReferencesEnabled = false,
+    useCase = SerializationContext.UseCase.Storage,
+    encoding = null,
+    encodingAllowList = AlwaysAcceptEncodingAllowList
 )
--- a/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
+++ b/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
@@ -15,10 +15,10 @@ import net.corda.serialization.SerializationContext
  */
 
 val AMQP_STORAGE_CONTEXT = SerializationContextImpl(
-        amqpMagic,
-        emptyMap(),
-        false, // todo double check in CORE-12472
-        SerializationContext.UseCase.Storage,
-        null,
-        AlwaysAcceptEncodingAllowList
+        preferredSerializationVersion = amqpMagic,
+        properties = emptyMap(),
+        objectReferencesEnabled = false, // todo double check in CORE-12472
+        useCase = SerializationContext.UseCase.Storage,
+        encoding = null,
+        encodingAllowList = AlwaysAcceptEncodingAllowList
 )
--- a/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/SharedContexts.kt
+++ b/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/SharedContexts.kt
@@ -9,7 +9,7 @@ import net.corda.serialization.SerializationEncoding
 val AMQP_P2P_CONTEXT = SerializationContextImpl(
         amqpMagic,
         emptyMap(),
-        true,
+        false, //todo revert this to true in CORE-12472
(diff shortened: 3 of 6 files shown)


blsemo (commented):
It looks ok to me, but I am not the expert here...

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
```diff
@@ -17,7 +17,7 @@ import net.corda.serialization.SerializationContext
 val AMQP_STORAGE_CONTEXT = SerializationContextImpl(
```
mnesbit, line 17: First comment is can we have named parameters when defining these contexts as this is hard to read without them.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
```diff
@@ -17,7 +17,7 @@ import net.corda.serialization.SerializationContext
...
         amqpMagic,
         emptyMap(),
-        true,
+        true, // todo should not this be false? CORE-12472
```
mnesbit: objectReferencesEnabled should always have been false

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/SharedContexts.kt
```diff
@@ -9,7 +9,7 @@ import net.corda.serialization.SerializationEncoding
 val AMQP_P2P_CONTEXT = SerializationContextImpl(
         amqpMagic,
         emptyMap(),
-        true,
+        false, //todo revert this to true in CORE-12472
```
mnesbit: No this should be left as true, but instead we should change the context used in https://github.com/corda/corda-runtime-os/blob/2e0db8d10ee12640beeb89fe8e6de4d2339002d8/components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/AMQPSerializationProvider.kt#L80 to the AMQP_STORAGE_CONTEXT

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
vlajos: Yes. I have just updated it.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
vlajos: Thanks. I have updated it.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/SharedContexts.kt
vlajos: Thanks. I have updated it.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
```diff
@@ -15,10 +15,10 @@ import net.corda.serialization.SerializationContext
...
 val AMQP_STORAGE_CONTEXT = SerializationContextImpl(
-        amqpMagic,
-        emptyMap(),
-        true,
-        SerializationContext.UseCase.Storage,
-        null,
-        AlwaysAcceptEncodingAllowList
+        preferredSerializationVersion = amqpMagic,
+        properties = emptyMap(),
+        objectReferencesEnabled = false, // todo double check in CORE-12472
```
mnesbit: yep this is correct no need for todo

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/SharedContexts.kt
```diff
@@ -7,11 +7,11 @@ import net.corda.serialization.SerializationContext
...
 
 val AMQP_P2P_CONTEXT = SerializationContextImpl(
-        amqpMagic,
-        emptyMap(),
-        true,
-        SerializationContext.UseCase.P2P,
-        null
+        preferredSerializationVersion = amqpMagic,
+        properties = emptyMap(),
+        objectReferencesEnabled = true, //todo double check in CORE-12472
```
mnesbit: yes this is correct no need for todo

components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/AMQPSerializationProvider.kt
```diff
@@ -77,7 +77,7 @@ class AMQPSerializationProvider @Activate constructor(
         val serializationService = SerializationServiceImpl(
             serializationOutput,
             deserializationInput,
-            AMQP_P2P_CONTEXT.withSandboxGroup(context.sandboxGroup)
+            AMQP_STORAGE_CONTEXT.withSandboxGroup(context.sandboxGroup) //todo double check in CORE-12472
```
mnesbit, line 80: Note the long term fix is going to be either: 1. We allow people to inject a serialization factory rather than a serializer from which they can then request a flavour of serializer from by name, or 2. we make two service interfaces for serialization so we can differentiate and inject right one at right point

vlajos: There are some traces of the second approach already. But it does not seem to work.
(`P2pSerializationService` and `StorageSerializationService` interfaces)

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/SerializationOutput.kt
```diff
@@ -149,11 +149,13 @@ open class SerializationOutput constructor(
         }
 
         val retrievedRefCount = objectHistory[obj]
-        if (retrievedRefCount == null) {
+        if (!context.objectReferencesEnabled || retrievedRefCount == null) {
```
rick-r3: I would put this on the line above where you retrieve the reference count and just not do the retrieve if references are not enabled.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/SerializationOutput.kt
```diff
@@ -149,11 +149,13 @@ open class SerializationOutput constructor(
...
             serializer.writeObject(obj, data, type, this, context, debugIndent)
             // Important to do it after serialization such that dependent object will have preceding reference numbers
             // assigned to them first as they will be first read from the stream on receiving end.
             // Skip for primitive types as they are too small and overhead of referencing them will be much higher than their content
+
+            // todo CORE-12472 is this needed if referencing is disabled?
```
rick-r3: I think it is fine to leave this, and if the flag was to change during a call to serialise, it is perfectly reasonable to reference back to something serialised while the flag was off.  (Not that it currently does not change during a serialise, but it could if you went with annotations / interface approach to switching in the future).

rick-r3 (commented):
Seems somewhat obfuscated to be switching on the P2P vs. STORAGE.  Likewise in the tests it's not obvious which is being used.  But not up to me and I don't know if this achieves what is desired.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
lankydan: Can we use named arguments in these constructors?

components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/AMQPSerializationProvider.kt
lankydan: Yeah, those were a carry over from C4, but they were created for this sort of purpose, but only the P2P context was used. Explains why we're seeing this problem is coming up now, because the same P2P serializer that I first created with the flow worker is being used everywhere.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/ServerContexts.kt
lankydan: Saw that Matthew already made the same comment and you addressed.

components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/AMQPSerializationProvider.kt
lankydan: Just to double check with this, even though I think we're fine with this anyway.

Is there going to be an issue in the future where we reintroduce the P2P context back into our code? E.g. they have existing P2P messages that are serialized using the STORAGE context and then we try and desrialize them using the P2P context?

Also when do we plan on putting in the long term fix? I'd say this is a bug imo so I think we should still try do this for 5.0 as a bug fix. As @mnesbit said, I think we should have injectable `P2pSerializationService` and `StorageSerializationService` which we can inject as needed into our services. This would also not be a difficult solution to implement.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/SerializationOutput.kt
vlajos: @rick-r3 I do not fully get this. Depending on the retrieve may want to end in the same branch. Or you say it is better to repeat that branch than having the map lookup when not needed?

vlajos: I have updated it to avoid the lookup when not needed.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/SerializationOutput.kt
vlajos: Thanks. I have removed the todo.

josephzunigadaly (approved):

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/SerializationOutput.kt
rick-r3: You've implemented what I intended

kyriathar (approved):
ok (sandbox & packaging)

PR #3619, opened 2023-04-14
CORE-12454: Fix retry mechanism

yift-r3:
This pull request will:
1. Replace the state of the registration request with a list of registration request states (since the key is now the short hash of the holding identity and not the registration request ID).
2. Add a timestamp of when to invoke the next re-registration (since the resistance tests will kill the worker over and over again, so a constant time will just keep waiting).
3. Will not mark any state as DLQ (the state is now per holding identity and not request ID).

Tested using the non-functional resistance tests.

The API pull request is in https://github.com/corda/corda-api/pull/1049


dimosr:
Given I am off for a week, someone else needs to take a more thorough look at this if you want to get it merged soon to get the resiliency test suite green.

But, I had a quick look and a high-level comment I had is we can still retain keying per registration ID. But, in order to do that we'd need a separate state+event topic, i.e. on state+event for the `MemberOpsAsyncProcessor` keyed by holding identity short hash which ensures requests for a vnode are processed sequentially, and a separate state+event topic keyed by registration ID that holds all the registration requests under processing to be replayed. We can have just a `StateAndEventListener` on the second one that sends events back to the first one for retries - it's possible there's no need for events there. It's also possible at that point that the first one can be a simple topic and `MemberOpsAsyncProcessor` could be converted to a durable processor that processes these events and atomically produces state events for the second topic.

This might help in terms of scalability limitations, i.e. if we bundle all states for a vnode under a single record we are subject to the limits of Kafka per record (1MB by default). To be fair, this is relatively large and applies only to requests under processing, but still it would be nice if we can avoid this because it could cause a serious issue if it happens. Another point where it could help is with the processing logic - I think we might be able to simplify significantly the logic that combines states on `MemberOpsAsyncProcessor` and the logic that combines timers on `CommandsRetryManager`.

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/MemberOpsServiceImpl.kt
```diff
@@ -82,7 +83,8 @@ class MemberOpsServiceImpl @Activate constructor(
         private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
         const val RPC_GROUP_NAME = "membership.ops.rpc"
         const val RPC_CLIENT_NAME = "membership.ops.rpc"
-        const val ASYNC_GROUP_NAME = "membership.ops.async"
+        const val ASYNC_COMMANDS_GROUP_NAME = "membership.ops.async.commands"
+        const val ASYNC_RETRIES_GROUP_NAME = "membership.ops.async.reties"
```
charlieR3:
```suggestion
        const val ASYNC_RETRIES_GROUP_NAME = "membership.ops.async.retries"
```

charlieR3 (dismissed):
LGTM. One typo that could be updated when bumping the API. Otherwise it looks good.

charlieR3 (approved):

PR #3620, opened 2023-04-14
CORE-3819: Replace deprecated "dependency substitution" Gradle APIs.

chrisr3:
Replace deprecated API which has been removed in Gradle 8.x.


ronanbrowne (approved):
lgtm

PR #3622, opened 2023-04-14
CORE-12498 Vault named query parsing escape special characters

lankydan:
`?` and `::` need to be used in Postgres JSON queries; however, they
cause parameter issues when creating the native query and execution
errors when executing them.

To bypass these errors, these characters must be escaped in the executed
query.

We have decided to keep our queries tidy, so the escaping has been hidden
within our parsing code. This means that developers write `?` and `::` 
into their query strings, and we translate them into `\\?\\?` and `\\:\\:` 
respectively.


nkovacsx (dismissed):

blsemo (approved):
PR looks fine to me, but the description needs updating, it now claims the opposite of what you implemented.

PR #3623, opened 2023-04-14
CORE-12504 Support missing JSON field keyword

lankydan:
Support `->` which was missing from the JSON expression parsing.


blsemo (approved):
LGTM

PR #3624, opened 2023-04-14
CORE-12505: fix SSO Docker Hub issue when building worker images

ronanbrowne:
`setCredentialsOnBaseImage` is not set on the base image resulting in the Docker Hub rate limit imposed on users despite setting the needed credentials.

This fix ensures users who have set valid creds for `DOCKER_HUB_USERNAME` and `DOCKER_HUB_PASSWORD` environment variables are not subject to Docker Hubs rate limit of 100 pulls per 6 hours per IP For anonymous users.


ronanbrowne:
@dickon FYI this resolves the issue you flagged earlier.

yarivus (approved):
LGTM

PR #3626, opened 2023-04-17
CORE-12499 Enforce the usage of the latest group parameters when finalizing transactions.

vlajos:
CORE-12499 Enforce the usage of the latest group parameters when finalizing transactions.

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
@@ -130,6 +130,7 @@ class UtxoLedgerTransactionVerificationServiceImpl @Activate constructor(
             "Membership group parameters hash cannot be found in the transaction metadata."
         }
     }
+
     private fun transactionVerificationFlowTimer(): Timer {
         return CordaMetrics.Metric.Ledger.TransactionVerificationFlowTime
             .builder()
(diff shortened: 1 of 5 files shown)


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
```diff
@@ -67,12 +67,23 @@ class UtxoLedgerTransactionVerificationServiceImpl @Activate constructor(
...
+        verifyCurrenGroupParametersUsed(transaction)
+        verify(transaction)
+    }
+
+    private fun verifyCurrenGroupParametersUsed(transaction: UtxoLedgerTransaction){
+        check(
+            transaction.getMembershipGroupParametersHash() ==
+                    currentGroupParametersService.get().hash.toString()
+        ) {
+            "Transactions can be created only with the latest membership group parameters."
```
blsemo, line 89: Hmm - that would cause problems when GroupParameters are changing - either a race conditions when they are changing while a flow is in flight (and between clusters this takes a finite amount of time), or a time window where not transactions can be finalized if e.g. the distribution of new group paramters is not atomically to all members.
Consider e.g. a  use case where members need to accept new group paramters?
We can totally can do this, but it's not just a technical detail, it's a design/usability decision.

blsemo (commented):
I am not comfortable rushing in this change at this point - it has the potential to change the pattern how group parameters can be updated, and needs to be thought through in detail, and in conjunctions with the design for GroupParameter updates.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/UtxoLedgerTransactionVerificationServiceImpl.kt
vlajos: I thought the point of the requirement is to fail similar finality flows where the initiator and responder sides use different versions.

mnesbit: Yes that is my expectation right now. The kinds of controls people want to put in Group Parameters e.g. notary operational changes and controls over issuers do require that we are on the same page. I would rather trigger a flow error than a ledger inconsistency.

blsemo (approved):
Ok

PR #3627, opened 2023-04-17
CORE-12169 - Remove MembershipP2PIntegrationTest

dimosr:
After discussion with the team, agreed to remove `MembershipP2PIntegrationTest`, since all associated functionality is now exercised as part of our e2e test suite.


yift-r3 (approved):

chrisr3:
> After discussion with the team, agreed to remove `MembershipP2PIntegrationTest`, since all associated functionality is now exercised as part of our e2e test suite.

Local integration tests are easier to run and debug than E2E tests in Jenkins, so I suspect you're going to make your own lives more difficult by doing this.

dimosr:
> Local integration tests are easier to run and debug than E2E tests in Jenkins, so I suspect you're going to make your own lives more difficult by doing this.

I don't think so. It has been discussed internally within the team and we collectively concluded that 1. OSGi integration tests are generally quite hard to troubleshoot and maintain (arguably more than e2e tests, at least in our team) 2. specifically these tests engage a lot of faking/mocking, which leads to much lower fidelity than the e2e tests. From past experience, failures on OSGi integration tests (like this one) were predominantly noisy and didn't indicate a real regression. They are also quite slow and became considerably slower due to some recent changes, so their value-add is considerably lower (close to zero) than the overhead required to maintain them at this point.

chrisr3:
> I don't think so. It has been discussed internally within the team and we collectively concluded that 1. OSGi integration tests are generally quite hard to troubleshoot and maintain (arguably more than e2e tests, at least in our team) 2. specifically these tests engage a lot of faking/mocking, which leads to much lower fidelity than the e2e tests. From past experience, failures on OSGi integration tests (like this one) were predominantly noisy and didn't indicate a real regression. They are also quite slow and became considerably slower due to some recent changes, so their value-add is considerably lower (close to zero) than the overhead required to maintain them at this point.

When you say "faking/mocking", do you mean "Mockito"? Because I would definitely suggest _not_ doing that. The point of the `TestOSGi` tasks is to put a subset of _real_ components inside an OSGi framework, so that you are confident that they work correctly together in their own right _before_ assembling the final application for E2E testing.

The E2E tests also use a full OSGi container of course, so you're not saving yourselves any effort there. But now your own code is no longer isolated, and you have to contend with the side-effects of everyone else's code too.

I would also consider the E2E tests to be historically flaky, and I really don't see how deleting the _integration tests_ is going to fix that.

dimosr:
> When you say "faking/mocking", do you mean "Mockito"? Because I would definitely suggest _not_ doing that. The point of the `TestOSGi` tasks is to put a subset of _real_ components inside an OSGi framework, so that you are confident that they work correctly together in their own right _before_ assembling the final application for E2E testing.

No, they are not using Mockito. They are using fake implementations of some services that belong into different modules from the one being tested.

> The E2E tests also use a full OSGi container of course, so you're not saving yourselves any effort there. But now your own code is no longer isolated, and you have to contend with the side-effects of everyone else's code too.

We are not trying to remove the use of OSGi entirely with this. We are removing a test that does not provide any extra value at the moment on top of the combination of existing e2e tests and unit tests for the corresponding components within this quite small module and at the same time adds a maintenance & build time tax for everyone. 

> I would also consider the E2E tests to be historically flaky, and I really don't see how deleting the _integration tests_ is going to fix that.

It's not the goal of this change to fix any flakiness of the E2E tests, as explained above.

charlieR3 (approved):

PR #3631, opened 2023-04-17
CORE-11665: Change LiquibaseDiffCheckFailedException error handling to return Bad Request

josephzunigadaly:
* Move `LiquibaseDiffCheckFailedException` info `virtual-node-common` so it is visible to `virtual-node-rest-service-impl`
* Add `LiquibaseDiffCheckFailedException` to list of exception `VirtualNodeRestResourceImpl.handleFailure` understands, return Bad Request HTTP status


vkolomeyko (approved):
LGTM

PR #3634, opened 2023-04-17
CORE-12352: Add suspend and activate to the e2e APIs

yift-r3:
Add a few helper methods to help test the membership suspension from within a flow.

the tests pull request is in https://github.com/corda/corda-e2e-tests/pull/72


testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ConfigTestUtils.kt
```diff
@@ -38,8 +38,11 @@ fun getConfig(section: String): JsonNode {
  * The currently installed schema and configuration versions are automatically obtained from the running system
  * before updating.
  */
-fun updateConfig(config: String, section: String) {
+fun updateConfig(config: String, section: String, clusterInfo: ClusterInfo? = null) {
     return cluster {
+        if (clusterInfo != null) {
+            init(clusterInfo)
```
williamvigorr3: Does this have any effect (the same is true in all the other places)? init calls `endpoint` which sets the `client` in the `ClusterBuilder` but we are calling `endpoint` on line 46 (thus resetting the client).

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ConfigTestUtils.kt
```diff
@@ -38,14 +38,8 @@ fun getConfig(section: String): JsonNode {
...
  */
-fun updateConfig(config: String, section: String) {
-    return cluster {
-        endpoint(
-            CLUSTER_URI,
-            USERNAME,
-            PASSWORD
-        )
-
+fun updateConfig(config: String, section: String, clusterInfo: ClusterInfo = ClusterBInfo) {
```
williamvigorr3, line 41: Defaulting to clusterB is a bit confusing. Also I think this change also means that the system property `restEndpointUrl` is not read and used anymore (here but is still used in other places). Even though it defaults to the same thing in the smoke test Gradle file (I think).

yift-r3: Most of the utilities are using cluster B by default implicitly. I changed some of them (the ones we need for the suspension) to be explicit while trying to keep the signature similar (so I won't need to change everything).

williamvigorr3: Maybe we should raise a Jira to tidy this up (set explicitly everywhere to cluster B + remove duplication between `Utils.kt` and `ClusterInfo.kt`).

I am a bit worried that this basically breaks the `"restEndpointUrl"` which is set by `applications/workers/workers-smoketest/build.gradle` if someone decided to set this to something else it wouldn't get propagated to the places you changed, this could cause a bug/some confusion.

yift-r3: https://r3-cev.atlassian.net/browse/CORE-13116 created

williamvigorr3 (approved):
LGTM

PR #3635, opened 2023-04-17
CORE-12505: ensure non R3 users can build docker images correctly

ronanbrowne:
If environment variables or Gradle properties for Docker hub login are not set, default to an empty string for these. This is required as we later evaluate these properties and Gradle expects _something_ to be set here in all cases.

This also ensures non-R3 users can build worker images, as these users will not have these values set, in this use case simply pull the image anonymously with no creds set.

Also, Add some javaDoc style comments to the method in q for clarity


buildSrc/src/main/groovy/DeployableContainerBuilder.groovy
```diff
@@ -323,21 +325,26 @@ abstract class DeployableContainerBuilder extends DefaultTask {
...
+    */
     private JibContainerBuilder setCredentialsOnBaseImage(JibContainerBuilder builder) {
         def baseImage = RegistryImage.named("${baseImageName.get()}:${baseImageTag.get()}")
         if ((registryUsername.get() != null && !registryUsername.get().isEmpty()) && baseImageName.get().contains("software.r3.com")) {
             logger.info("Authenticating against Artifactory for base image resolution")
             baseImage.addCredential(registryUsername.get(), registryPassword.get())
             builder = Jib.from(baseImage)
-        }
-        else if ((dockerHubUsername.get() != null && !dockerHubUsername.get().isEmpty()) && (dockerHubPassword.get() != null && !dockerHubPassword.get().isEmpty())){
+        } else if ((dockerHubUsername.get() != null && !dockerHubUsername.get().isEmpty()) && (dockerHubPassword.get() != null && !dockerHubPassword.get().isEmpty())){
```
yarivus:
```suggestion
        } else if ((dockerHubUsername.get() != null && !dockerHubUsername.get().isEmpty()) &&
                   (dockerHubPassword.get() != null && !dockerHubPassword.get().isEmpty())) {
```

yarivus (changes requested):

yarivus (dismissed):
LGTM

ronanbrowne:
Follow on change to use the same pattern for not just docker hub but Artifactory values, which an external user will also not have set

yarivus (approved):
LGTM

PR #3637, opened 2023-04-17
CORE-12619: Also include custom AMQP serializers inside PERSISTENCE sandboxes.

chrisr3:
We need to be able to serialize / deserialize `ContractState` and `StateAndRef` in all sandbox group types, which means that their custom AMQP serializers also need to be available for `PERSISTENCE` sandboxes.

Custom serializers which do not use sandbox services can become "platform" services.


components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/filtered/serializer/amqp/FilteredTransactionSerializer.kt
```diff
@@ -28,11 +28,15 @@ class FilteredTransactionSerializer @Activate constructor(
     private val merkleTreeProvider: MerkleTreeProvider
 ) : BaseProxySerializer<FilteredTransaction, FilteredTransactionProxy>(), UsedByFlow {
 
-    override val proxyType = FilteredTransactionProxy::class.java
+    override val proxyType
```
lankydan, line 31: Whats the benefit of using `get()` here?

lankydan (commented):
Change is fine left one comment.

components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/filtered/serializer/amqp/FilteredTransactionSerializer.kt
chrisr3: A `val` property like this:
```kotlin
override val proxyType = FilteredTransactionProxy::class.java
```
is compiled to something like this:
```java
private final Class<FilteredTransactionProxy> _proxyType = FilteredTransactionProxy.class;

public Class<FilteredTransactionProxy> getProxyType() {
    return _proxyType;
}
```
We don't really need that `_proxyType` field - all we need is the getter:
```java
public Class<FilteredTransactionProxy> getProxyType() {
    return FilteredTransactionProxy.class;
}
```

vlajos (approved):
LGTM

lankydan (approved):

wzur-r3 (approved):
LGTM

driessamyn (approved):
LGTM (for packaging)

PR #3640, opened 2023-04-18
CORE-11984: Upgrade version of Nimbus library to `10.7.2`

vkolomeyko:
It is used in REST SSO functionality. Integration tests pass, and also I have manually tested Azure AD SSO with Combined Worker locally.

### DIFF
--- a/libs/rest/rest-server-impl/src/main/kotlin/net/corda/rest/server/impl/security/provider/bearer/azuread/AzureAdIssuerJWSKeySelector.kt
+++ b/libs/rest/rest-server-impl/src/main/kotlin/net/corda/rest/server/impl/security/provider/bearer/azuread/AzureAdIssuerJWSKeySelector.kt
@@ -1,8 +1,7 @@
 package net.corda.rest.server.impl.security.provider.bearer.azuread
 
 import com.nimbusds.jose.JWSHeader
-import com.nimbusds.jose.jwk.source.DefaultJWKSetCache
-import com.nimbusds.jose.jwk.source.RemoteJWKSet
+import com.nimbusds.jose.jwk.source.JWKSourceBuilder
 import com.nimbusds.jose.proc.BadJOSEException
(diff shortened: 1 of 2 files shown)


libs/rest/rest-server-impl/src/main/kotlin/net/corda/rest/server/impl/security/provider/bearer/azuread/AzureAdIssuerJWSKeySelector.kt
```diff
@@ -21,8 +20,8 @@ internal class AzureAdIssuerJWSKeySelector(
 ) : JWTClaimsSetAwareJWSKeySelector<SecurityContext> {
     companion object {
         private const val ISS_CLAIM = "iss"
-        private const val DEFAULT_REFRESH_TIME_MINUTES = 5L
-        private val DEFAULT_JWK_SET_CACHE_LIFESPAN = TimeUnit.MINUTES.toMillis(5L)
+        private val DEFAULT_JWK_CACHE_REFRESH = TimeUnit.MINUTES.toMillis(4L)
```
driessamyn, line 23: Is this change from 5 to 4 minutes intentional?

driessamyn (commented):
Looking ok, just checking if that change from 5 to 4 minutes was intentional?
I'm guessing it doesn't make much difference either way.

libs/rest/rest-server-impl/src/main/kotlin/net/corda/rest/server/impl/security/provider/bearer/azuread/AzureAdIssuerJWSKeySelector.kt
vkolomeyko: Yes - it is. The two cannot be matching as there is also `refreshAhead` interval set by default to 30 seconds.

driessamyn (approved):

josephzunigadaly (approved):

PR #3644, opened 2023-04-18
CORE-10584 Add docs for member suspension

williamvigorr3:
Docs update for member suspension.

### DIFF
--- a/wiki/Member-Suspension.md
+++ b/wiki/Member-Suspension.md
@@ -5,7 +5,7 @@ It is not possible to suspend the MGM.
 
 ## Set variables to be used by other commands
 
-Set the `API_URL`, to the URL of the REST worker.
+Set the `API_URL` to the URL of the REST worker.
 This may vary depending on where you have deployed your cluster and how you have forwarded the ports.
 Set `MGM_HOLDING_ID` to the short hash of the MGM's Holding Identity.
 Set `MEMBER_X500_NAME` to the X500 name of the member being suspended or re-activated.
@@ -39,13 +39,15 @@ $AUTH_INFO = [Convert]::ToBase64String([Text.Encoding]::ASCII.GetBytes(("admin:a
 ## Querying for Members
 
 The member lookup REST endpoint can be used to query for all members with a certain status.
+It is possible to query for members with status `ACTIVE` or `SUSPENDED`, by default only `ACTIVE` members are shown.
+Multiple statuses can be queried for together.
 For example to query for all members with status `SUSPENDED`:
 
 <details>
 <summary>Bash</summary>
 
 ```bash
-curl --insecure -u admin:admin -X 'GET' "$API_URL/members/$MGM_HOLDING_ID?statuses=SUSPENDED"
+curl --insecure -u admin:admin "$API_URL/members/$MGM_HOLDING_ID?statuses=SUSPENDED"
 ```
 
 </details>
@@ -61,6 +63,7 @@ Invoke-RestMethod -SkipCertificateCheck  -Headers @{Authorization=("Basic {0}" -
 </details>
 
 From this the serial number of each member's member information can be extracted from the `corda.serial` field inside the `mgmContext`.
+This is used in the following APIs.
 Only the MGM can query for `SUSPENDED` members.
 
 ## Suspending a Member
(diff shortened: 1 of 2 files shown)


wiki/Member-Suspension.md
```diff
@@ -0,0 +1,121 @@
+It is possible to temporarily suspend a member of the group. 
+Once a member has been suspended, flow communication between it and other members of the group will be blocked.
+A suspended member performing a member lookup will not see updates for the other members, apart from the MGM.
+It is not possible to suspend the MGM.
+
+## Set variables to be used by other commands
+
+Set the `API_URL`, to the URL of the REST worker.
```
charlieR3:
```suggestion
Set the `API_URL` to the URL of the REST worker.
```

wiki/Member-Suspension.md
```diff
@@ -0,0 +1,121 @@
...
+## Querying for Members
+
+The member lookup REST endpoint can be used to query for all members with a certain status.
+For example to query for all members with status `SUSPENDED`:
+
+<details>
+<summary>Bash</summary>
+
+```bash
+curl --insecure -u admin:admin -X 'GET' "$API_URL/members/$MGM_HOLDING_ID?statuses=SUSPENDED"
```
charlieR3: I think the default method is GET so you could optionally leave out `-X 'GET' `

wiki/Member-Suspension.md
```diff
@@ -0,0 +1,121 @@
...
+$API_URL = "https://$REST_HOST`:$REST_PORT/api/v1"
+$MGM_HOLDING_ID = "<MGM Holding ID>"
+$MEMBER_X500_NAME = "<Member X500 Name>"
+$AUTH_INFO = [Convert]::ToBase64String([Text.Encoding]::ASCII.GetBytes(("admin:admin" -f $username,$password)))
+```
+</details>
+
+## Querying for Members
+
+The member lookup REST endpoint can be used to query for all members with a certain status.
```
charlieR3, line 41: I would add here the list of possible statuses e.g. ACTIVE & SUSPENDED, and a note that by default, only active members are returned. 
Also worth mentioning that multiple statuses can be submitted together.

wiki/Member-Suspension.md
```diff
@@ -0,0 +1,121 @@
...
+<summary>PowerShell</summary>
+
+```powershell
+Invoke-RestMethod -SkipCertificateCheck  -Headers @{Authorization=("Basic {0}" -f $AUTH_INFO)}`
+   -Method Get -Uri $API_URL/members/$MGM_HOLDING_ID`?statuses=SUSPENDED
+```
+
+</details>
+
+From this the serial number of each member's member information can be extracted from the `corda.serial` field inside the `mgmContext`.
```
charlieR3, line 65: ` for use in the following APIs` maybe? It might not be clear at this point why a serial number is relevant.

wiki/Member-Suspension.md
```diff
@@ -0,0 +1,121 @@
...
+Invoke-RestMethod -SkipCertificateCheck  -Headers @{Authorization=("Basic {0}" -f $AUTH_INFO)}`
+  -Method Post -Uri $API_URL/mgm/$MGM_HOLDING_ID/suspend -Body (ConvertTo-Json -Depth 1 @{
+  x500Name = $MEMBER_X500_NAME; serialNumber = <SERIAL NUMBER>})
+```
+</details>
+
+Where `<SERIAL NUMBER>` is the current serial number of the member information.
+If the serial number doesn't match, then the REST method will return a 503.
+This can happen if another process has updated the member information, before the suspension operation.
+The operator can, then re-query the member lookup REST endpoint and decide if they still want to proceed with the operation.
```
charlieR3, line 97: It could be good to mention also that submitting the serial number is optional, what will happen if it's not submitted (i.e. the latest member info found is updated), and that submitting the serial number is safer to avoid suspending based on outdated information.

wiki/Member-Suspension.md
```diff
@@ -0,0 +1,121 @@
...
+Invoke-RestMethod -SkipCertificateCheck  -Headers @{Authorization=("Basic {0}" -f $AUTH_INFO)}`
+  -Method Post -Uri $API_URL/mgm/$MGM_HOLDING_ID/activate -Body (ConvertTo-Json -Depth 1 @{
+  x500Name = $MEMBER_X500_NAME; serialNumber = <SERIAL NUMBER>})
+```
+</details>
+
+Where `<SERIAL NUMBER>` is the current serial number of the member information.
+If the serial number doesn't match, then the REST method will return a 503.
+This works in the same way as for [Suspending a Member](#suspending-a-Member).
+Once a member has been re-activated it flow communication between it and other members can resume.
```
charlieR3: And probably here would be good to mention the serial number is optional also.

wiki/Member-Suspension.md
williamvigorr3: I left that out on purpose but will add it in with a warning.

charlieR3 (approved):
LGTM

PR #3647, opened 2023-04-18
CORE-12181 Documentation for security policy format

mbrkic-r3:
Added documentation for security policy format.


components/security-manager/README.md
```diff
@@ -1,4 +1,43 @@
 # Security Manager
 
 See https://engineering.r3.com/product-areas/corda-platform/node-and-packaging/docs/security-manager/ for more
```
mbrkic-r3: @nadinequinr3 This URL is not reachable anymore, do you maybe know where was this moved to?

nadinequinr3: The content in that old internal knowledgebase was migrated to Confluence before my time. As far as I know, the author of each page was responsible for moving it.

mbrkic-r3: I've removed this README as I couldn't find anything in wiki/confluence

thiagoviana (approved):

kyriathar (approved):

PR #3648, opened 2023-04-18
CORE-12562 lazy deserialization of StateAndRefs

vlajos:
* CORE-12562 Introduce LazyStateAndRefImpl to delay the deserialization of related contract states for performance reasons and also to deserialize them first in the verification sandbox.
* CORE-12562 Centralize UtxoTransactionOutputDto classes in ledger-utxo-data


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerStateQueryService.kt
```diff
@@ -31,4 +33,15 @@ interface UtxoLedgerStateQueryService {
...
+    /**
+     * Fetch serialized information from db worker needed to resolve [StateRef]s to [StateAndRef]s
+     *
+     * @param stateRefs The [StateRef]s to be resolved.
+     * @return [UtxoTransactionOutputDto] for each [StateAndRef]s.
+     *
+     * @throws CordaPersistenceException if an error happens during resolve operation.
+     */
+    @Suspendable
+    fun lazyResolveStateAndRefs(stateRefs: Iterable<StateRef>): List<LazyStateAndRefImpl<*>>
```
blsemo: I think this function is wrong - `resolveStateRefs` should always return lazy StateAndRefs. The impl shouldn't be on the interface at all.

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/state/LazyStateAndRefImpl.kt
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Determines whether the specified object is equal to the current object.
+     *
+     * @param other The object to compare with the current object.
+     * @return Returns true if the specified object is equal to the current object; otherwise, false.
+     */
+    override fun equals(other: Any?): Boolean {
+        return this === other
+                || other != null
+                && other is LazyStateAndRefImpl<*>
```
blsemo:
```suggestion
                && other is StateAndRef<*>
```

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/state/serializer/amqp/StateAndRefSerializer.kt
```diff
@@ -15,6 +15,8 @@ import org.osgi.service.component.annotations.Activate
 import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
 
+// todo try to remove?
```
blsemo: No, we can't remove those. They are required for visibility of internal classes to the serialization framework.

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/transaction/UtxoTransactionOutputDto.kt
```diff
@@ -0,0 +1,57 @@
...
+                contractState as T,
+                info.notaryName,
+                info.notaryKey,
+                info.getEncumbranceGroup()
+            ),
+            ref = StateRef(parseSecureHash(transactionId), leafIndex)
+        )
+    }
+
+    fun <T : ContractState> toLazyStateAndRefImpl(serializationService: SerializationService): LazyStateAndRefImpl<T> =
```
blsemo: This can probably just return a `StateAndRef`?

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/transaction/UtxoTransactionOutputDto.kt
```diff
@@ -0,0 +1,57 @@
...
+
+        return true
+    }
+
+    override fun hashCode(): Int = Objects.hash(transactionId, leafIndex, info, data)
+
+    @Suppress("UNCHECKED_CAST")
+    fun <T : ContractState> toStateAndRef(serializationService: SerializationService): StateAndRef<T> {
+        val info = serializationService.deserialize<UtxoOutputInfoComponent>(info)
+        val contractState = serializationService.deserialize<ContractState>(data)
```
blsemo: Can we put a `try...catch` around the deserialization of the contract state and add as much information as possible from the info before rethrowing?

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/state/serializer/amqp/StateAndRefSerializer.kt
blsemo: Maybe you could rewrite it to make the container one of the DTOs, and deserialize into a lazy StateAndRefImpl?

blsemo (commented):
I think the overall logic looks fine.
But I was hoping to hide more of it - basically have `StateAndRef` everywhere, the impls should just be details. 
E.g. the ledger service should always just resolve to lazy StateAndRefs. The `toStateRef` function could be internal to the LazyStateAndRefImpl.

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/state/LazyStateAndRefImpl.kt
vlajos: Do you mean, we should treat all `StateAndRef`s equally they resolve the same `StateRef` regardless of the impl?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerStateQueryService.kt
vlajos: Thanks. Updated.

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/transaction/UtxoTransactionOutputDto.kt
vlajos: If it does not return the impl, the impl's serializer will need a cast.

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/transaction/UtxoTransactionOutputDto.kt
vlajos: Good idea. I have updated it.

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/state/LazyStateAndRefImpl.kt
blsemo: Yes - that should be transparent to the user.

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/state/serializer/amqp/EncumbranceGroupSerializer.kt
blsemo, line 1: We might still need this one - EncumbranceGroup is a part of the `UtxoOutputInfo`, if I remember correctly.

vlajos:
```
data class UtxoOutputInfoComponent(
    val encumbrance: String?,
    val encumbranceGroupSize: Int?,
...
```
Only its properties are there. The class itself is not used there.

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/state/LazyStateAndRefImpl.kt
vlajos: OK. Updated.

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/state/serializer/amqp/StateAndRefSerializer.kt
vlajos: I ended up removing these serializers since (I hope) they won't be used anymore.

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/transaction/UtxoTransactionOutputDto.kt
vlajos: > If it does not return the impl, the impl's serializer will need a cast.

This is not true anymore assuming the serializers are not needed.
On the other hand this class does not explicit return type.

blsemo (approved):
LGTM

ronanbrowne (approved):
Lgtm - build

kyriathar (commented):
LGTM (sandboxing)

PR #3649, opened 2023-04-18
CORE-12539 - Postgres driver update

driessamyn:


driessamyn:
I raised this a self-contained change as the DB  driver upgrade could obviously have performance implications.
FYI @JamesHR3 , @davidcurrie , @jujoramos for awareness.

vkolomeyko (approved):
LGTM

PR #3650, opened 2023-04-18
CORE-12539 - Caffeine lib update

driessamyn:


driessamyn:
I raised this a self-contained change as the cache upgrade could have performance implications.
FYI @JamesHR3 , @davidcurrie , @jujoramos for awareness.

vkolomeyko (approved):
LGTM

PR #3652, opened 2023-04-18
CORE-12539 - testing and doc libs upgrades

driessamyn:
Update of:

- Dokka
- Junit
- AssertJ
- Mockito

The mockito update triggered a bigger change than expected. [Since v5, mockito-inline is the default](https://github.com/mockito/mockito/releases/tag/v5.0.0), so we no longer need the mockito-inline library.
The flip side is that there seems to be a problem with Mockito in OSGi, so mockito was replaced with "manual stubs" in some OSGi tests.


yift-r3 (dismissed):

applications/workers/worker-common/build.gradle
```diff
@@ -42,7 +42,10 @@ dependencies {
     runtimeOnly project(':libs:lifecycle:lifecycle-impl')
     runtimeOnly "org.apache.commons:commons-lang3:$commonsLangVersion"
 
-    testImplementation "org.mockito.kotlin:mockito-kotlin:$mockitoKotlinVersion"
+    testImplementation "org.mockito:mockito-core:$mockitoVersion"
+    testImplementation("org.mockito.kotlin:mockito-kotlin:$mockitoKotlinVersion") {
+        exclude group: 'mockito-core'
```
vkolomeyko: What is going to happen if we remove this `exclude` instruction? (especially given that we include `mockito-core` in the line above)

yift-r3: I think this force the `mockito-core` to be the `mockitoVersion` regardless of the one in `mockito-kotlin`.

driessamyn: @yift-r3 made this change, but I think what this says is to use the mockito-core version we have declared, not the one that mockito-kotlin is asking for.

vkolomeyko: But `mockitoVersion` we define in our properties is `5.3.0`, whereas the one that comes from `mockito-kotlin` is only `4.5.1` as per: https://central.sonatype.com/artifact/org.mockito.kotlin/mockito-kotlin/4.1.0/dependencies.
I.e. the version of `mockito-core` that we explicitly specify is greater than the one comes from `mockito-kotlin`, therefore Gradle should do automatic resolution and choose `5.3.0`.
My point is that this `exclude` instruction may not be necessary and can be removed. WDYT?

driessamyn: You might be right. I've tried to remove them and it compiles locally, so just trying to run the tests now: https://github.com/corda/corda-runtime-os/pull/3663

yift-r3: I think it supposes to protect us for the future (i.e. if we will upgrade `mockito-kotlin` to a version that uses another version of `mockito-core`). But if it works without it I'm OK.

vkolomeyko: Whichever is easier, I do not have a strong opinion here.

The other option could be to use Gradle's constraint feature: https://docs.gradle.org/current/userguide/dependency_constraints.html
if we want to override the version of `mockito-core` which `mockito-kotlin` is pulling in.

driessamyn: I tried on the other branch and applied the change.
I think it should work as long as the kotlin library keeps being compatible with the core.
We should find out quickly if that stops working after a future upgrade, so I suggest we leave it till then as it's a slightly simpler config

wzur-r3 (approved):
LGTM for Gradle build files

yift-r3 (approved):

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
```diff
@@ -163,7 +163,7 @@ class QueryRegistrationRequestsHandlerTest {
     fun `invoke queries with correct predicates if X500 name specified in request`() {
         whenever(actualQuery.resultList).doReturn(emptyList())
         val captor = argumentCaptor<Predicate>()
-        whenever(query.where(captor.capture())).thenReturn(query)
+        whenever(query.where(captor.capture(), any())).thenReturn(query)
```
fowlerrr, line 166: Was this the wrong check?  Are we expecting two predicates in the where clause?

fowlerrr (approved):

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandlerTest.kt
yift-r3: We are expecting two predicates. I'm not sure why it passed before. It probably has something to do with Mockito handling of var args.

vkolomeyko (approved):
LGTM

PR #3654, opened 2023-04-19
CORE-9633 `DigitalSignatureMetadata.properties` should be deterministically serialized

kyriathar:
`DigitalSignatureMetadata.properties` which is of type `Map<String, String>` is getting serialized as part of `DigitalSignatureMetadata` and gets signed alongside actual payload. It can be that two `DigitalSignatureMetadata` may be equal meaning their `DigitalSignatureMetadata.properties` are equal but `DigitalSignatureMetadata.properties` iteration order might be different leading to different serializations and therefore different signatures.

- Adds a custom AMQP serializer for `DigitalSignatureMetadata` to convert `DigitalSignatureMetadata.properties` to a `SortedMap` before serializing it, thus making equal maps but with different iteration orders over their entries serialize deterministically.


mnesbit (approved):
OK

vlajos (approved):
LGTM

fowlerrr (approved):

PR #3656, opened 2023-04-19
CORE-12440: Improve Flow Tracing Capabilities

jujoramos:
Automatically add the 'clientRequestId' to the PlatformProperties so
the original request Id used by the client is always propagated across
workers for an entire flow execution.

- Include 'clientRequestId' within the MDC used by
  'EntityMessageProcessor'.
- Include 'clientRequestId' within the MDC used by 
  'PersistenceRequestProcessor'.
- Include 'clientRequestId' within the MDC used by
  'VerificationRequestProcessor'.
- Include 'clientRequestId' and 'externalEventId' within the MDC used
  by 'CryptoFlowOpsBusProcessor'.
- Update flow external event manager so it registers under INFO log
  level whenever external events are sent or received.

### DIFF
--- a/components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
+++ b/components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
@@ -72,7 +72,7 @@ class CryptoFlowOpsBusProcessor(
 
             try {
                 if (Instant.now() >= expireAt) {
-                    logger.warn("Event ${request.request::class.java} for tenant ${request.context.tenantId} " +
+                    logger.warn("Event ${request.request::class.java.name} for tenant ${request.context.tenantId} " +
                             "is no longer valid, expired at $expireAt")
                     externalEventResponseFactory.transientError(
                         request.flowExternalEventContext,
@@ -84,7 +84,7 @@ class CryptoFlowOpsBusProcessor(
                     }
 
                     if (Instant.now() >= expireAt) {
-                        logger.warn("Event ${request.request::class.java} for tenant ${request.context.tenantId} " +
+                        logger.warn("Event ${request.request::class.java.name} for tenant ${request.context.tenantId} " +
                                 "is no longer valid, expired at $expireAt")
                         externalEventResponseFactory.transientError(
                             request.flowExternalEventContext,
@@ -99,7 +99,7 @@ class CryptoFlowOpsBusProcessor(
                 }
             } catch (throwable: Throwable) {
                 logger.error(
-                    "Failed to handle ${request.request::class.java} for tenant ${request.context.tenantId}",
+                    "Failed to handle ${request.request::class.java.name} for tenant ${request.context.tenantId}",
                     throwable
                 )
                 externalEventResponseFactory.platformError(request.flowExternalEventContext, throwable)
--- a/components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
+++ b/components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
@@ -22,7 +22,8 @@ import net.corda.messaging.api.processor.DurableProcessor
 import net.corda.messaging.api.records.Record
 import net.corda.utilities.MDC_CLIENT_ID
 import net.corda.utilities.MDC_EXTERNAL_EVENT_ID
-import net.corda.utilities.debug
+import net.corda.utilities.MDC_FLOW_ID
+import net.corda.utilities.trace
 import net.corda.utilities.withMDC
 import org.slf4j.LoggerFactory
 import java.time.Instant
@@ -46,8 +47,10 @@ class CryptoFlowOpsBusProcessor(
         BackoffStrategy.createBackoff(config.maxAttempts, config.waitBetweenMills)
     )
 
-    override fun onNext(events: List<Record<String, FlowOpsRequest>>): List<Record<*, *>> =
-        events.mapNotNull { onNext(it) }
+    override fun onNext(events: List<Record<String, FlowOpsRequest>>): List<Record<*, *>> {
+        logger.trace { "onNext processing messages ${events.joinToString(",") { it.key }}" }
+        return events.mapNotNull { onNext(it) }
+    }
 
     private fun onNext(event: Record<String, FlowOpsRequest>): Record<*, *>? {
         val request = event.value
@@ -56,65 +59,50 @@ class CryptoFlowOpsBusProcessor(
             return null // cannot send any error back as have no idea where to send to
         }
 
-        val expireAt = getRequestExpireAt(request)
-        val flowId = request.flowExternalEventContext.flowId
-        val requestId = request.flowExternalEventContext.requestId
-        val clientRequestId = request.flowExternalEventContext.contextProperties.toMap()[MDC_CLIENT_ID] ?: ""
+        logger.info("Handling ${request.request::class.java.name} for tenant ${request.context.tenantId}")
 
+        val expireAt = getRequestExpireAt(request)
+        val clientRequestId = request.flowExternalEventContext.contextProperties.toMap()[MDC_CLIENT_ID] ?: ""
         val mdc = mapOf(
+            MDC_FLOW_ID to request.flowExternalEventContext.flowId,
             MDC_CLIENT_ID to clientRequestId,
-            MDC_EXTERNAL_EVENT_ID to requestId
+            MDC_EXTERNAL_EVENT_ID to request.flowExternalEventContext.requestId
         )
 
         return withMDC(mdc) {
-            if (Instant.now() >= expireAt) {
-                logger.warn(
-                    "Event ${request.request::class.java} for tenant ${request.context.tenantId} is no longer valid, " +
-                            "expired at $expireAt { requestId: $requestId, key: $flowId }"
-                )
-
-                return@withMDC externalEventResponseFactory.transientError(
-                    request.flowExternalEventContext,
-                    ExceptionEnvelope("Expired", "Expired at $expireAt")
-                )
-            }
-
             try {
-                logger.info(
-                    "Handling ${request.request::class.java.name} for tenant ${request.context.tenantId} " +
-                            "{ requestId: $requestId, key: $flowId }"
-                )
-                val response = executor.executeWithRetry {
-                    handleRequest(request.request, request.context)
-                }
                 if (Instant.now() >= expireAt) {
-                    logger.error(
-                        "Event ${request.request::class.java} for tenant ${request.context.tenantId} is no longer valid, " +
-                                "expired at $expireAt { requestId: $requestId, key: $flowId }"
-                    )
-
-                    return@withMDC externalEventResponseFactory.transientError(
+                    logger.warn("Event ${request.request::class.java} for tenant ${request.context.tenantId} " +
+                            "is no longer valid, expired at $expireAt")
+                    externalEventResponseFactory.transientError(
                         request.flowExternalEventContext,
                         ExceptionEnvelope("Expired", "Expired at $expireAt")
                     )
-                }
-                val result = externalEventResponseFactory.success(
-                    request.flowExternalEventContext,
-                    FlowOpsResponse(createResponseContext(request), response, null)
-                )
-                logger.debug {
-                    "Handled ${request.request::class.java.name} for tenant ${request.context.tenantId} " +
-                            "{ requestId: $requestId, key: $flowId }"
-                }
-                result
-            } catch (t: Throwable) {
-                logger.error(
-                    "Failed to handle ${request.request::class.java} for tenant ${request.context.tenantId} " +
-                            "{ requestId: $requestId, key: $flowId }",
-                    t
-                )
+                } else {
+                    val response = executor.executeWithRetry {
+                        handleRequest(request.request, request.context)
+                    }
 
-                return@withMDC externalEventResponseFactory.platformError(request.flowExternalEventContext, t)
+                    if (Instant.now() >= expireAt) {
+                        logger.warn("Event ${request.request::class.java} for tenant ${request.context.tenantId} " +
+                                "is no longer valid, expired at $expireAt")
+                        externalEventResponseFactory.transientError(
+                            request.flowExternalEventContext,
+                            ExceptionEnvelope("Expired", "Expired at $expireAt")
+                        )
+                    } else {
+                        externalEventResponseFactory.success(
+                            request.flowExternalEventContext,
+                            FlowOpsResponse(createResponseContext(request), response, null)
+                        )
+                    }
+                }
+            } catch (throwable: Throwable) {
+                logger.error(
+                    "Failed to handle ${request.request::class.java} for tenant ${request.context.tenantId}",
+                    throwable
+                )
+                externalEventResponseFactory.platformError(request.flowExternalEventContext, throwable)
             }
         }
     }
--- a/components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
+++ b/components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
@@ -7,6 +7,7 @@ import net.corda.crypto.config.impl.toCryptoConfig
 import net.corda.crypto.flow.CryptoFlowOpsTransformer
 import net.corda.crypto.impl.retrying.BackoffStrategy
 import net.corda.crypto.impl.retrying.CryptoRetryingExecutor
+import net.corda.crypto.impl.toMap
 import net.corda.data.ExceptionEnvelope
 import net.corda.data.KeyValuePairList
 import net.corda.data.crypto.wire.CryptoRequestContext
@@ -19,7 +20,10 @@ import net.corda.data.crypto.wire.ops.flow.queries.FilterMyKeysFlowQuery
 import net.corda.flow.external.events.responses.factory.ExternalEventResponseFactory
 import net.corda.messaging.api.processor.DurableProcessor
 import net.corda.messaging.api.records.Record
+import net.corda.utilities.MDC_CLIENT_ID
+import net.corda.utilities.MDC_EXTERNAL_EVENT_ID
 import net.corda.utilities.debug
+import net.corda.utilities.withMDC
 import org.slf4j.LoggerFactory
 import java.time.Instant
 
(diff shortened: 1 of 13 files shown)


components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
```diff
@@ -52,54 +56,66 @@ class CryptoFlowOpsBusProcessor(
...
+        val flowId = request.flowExternalEventContext.flowId
+        val requestId = request.flowExternalEventContext.requestId
+        val clientRequestId = request.flowExternalEventContext.contextProperties.toMap()[MDC_CLIENT_ID] ?: ""
+
+        val mdc = mapOf(
+            MDC_CLIENT_ID to clientRequestId,
+            MDC_EXTERNAL_EVENT_ID to requestId
+        )
+
+        return withMDC(mdc) {
```
JamesHR3, line 70: I wonder if this return actually makes this code harder to read, particularly with the nested returns that then bubble up to the `withMDC` level rather than exiting this function.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
```diff
@@ -52,54 +56,66 @@ class CryptoFlowOpsBusProcessor(
...
+            try {
+                logger.info(
+                    "Handling ${request.request::class.java.name} for tenant ${request.context.tenantId} " +
+                            "{ requestId: $requestId, key: $flowId }"
+                )
+                val response = executor.executeWithRetry {
+                    handleRequest(request.request, request.context)
+                }
+                if (Instant.now() >= expireAt) {
+                    logger.error(
```
JamesHR3: I think this log should be a `warn` rather than `error`, particularly as it's transient. (Appreciate this isn't something you've changed though.)

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
```diff
@@ -118,6 +134,7 @@ class CryptoFlowOpsBusProcessor(
                     tenantId = context.tenantId,
                     candidateKeys = request.keys
                 )
+
```
JamesHR3: Did this autoformat? Might be worth getting rid of the whitespace here.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowMDCServiceImpl.kt
```diff
@@ -8,19 +8,19 @@ import net.corda.data.flow.state.checkpoint.Checkpoint
...
-    private companion object {
-        const val MDC_FLOW_ID = "flow_id"
-        const val MDC_CLIENT_ID = "client_id"
-        const val MDC_VNODE_ID = "vnode_id"
-        const val MDC_SESSION_EVENT_ID = "session_event_id"
-        const val MDC_EXTERNAL_EVENT_ID = "external_event_id"
+    companion object {
+        const val MDC_FLOW_ID = "flow.id"
+        const val MDC_VNODE_ID = "vnode.id"
+        const val MDC_SESSION_EVENT_ID = "session.event.id"
```
JamesHR3: Should these constants also be in `net.corda.utilities`? Feels like we have a split of where these constants live.

JamesHR3 (commented):
Couple of thoughts.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowMDCServiceImpl.kt
jujoramos: I wasn't sure whether these ones are only pertinent to flows or not, that's why I didn't move them to `net.corda.utilities` directly. Makes sense to have them all in the same place, though, I'll make the change.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
jujoramos: 👍

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
jujoramos: 👍

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
jujoramos: Done

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
```diff
@@ -46,8 +47,10 @@ class CryptoFlowOpsBusProcessor(
         BackoffStrategy.createBackoff(config.maxAttempts, config.waitBetweenMills)
     )
 
-    override fun onNext(events: List<Record<String, FlowOpsRequest>>): List<Record<*, *>> =
-        events.mapNotNull { onNext(it) }
+    override fun onNext(events: List<Record<String, FlowOpsRequest>>): List<Record<*, *>> {
```
jujoramos, line 50: RequestId and FlowId are now part of the MDC context for all logged messages, so I've removed `{ requestId: $requestId, key: $flowId }` from the message string itself.

conal-smith (dismissed):
lgtm

vlajos (dismissed):
LGTM.

components/flow/flow-rest-resource-service-impl/src/main/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImpl.kt
```diff
@@ -166,7 +167,10 @@ class FlowRestResourceImpl @Activate constructor(
 
         // TODO Platform properties to be populated correctly, for now a fixed 'account zero' is the only property
         // This is a placeholder which indicates access to everything, see CORE-6076
-        val flowContextPlatformProperties = mapOf("corda.account" to "account-zero")
+        val flowContextPlatformProperties = mapOf(
```
fowlerrr, line 170: Very minor but may adjust the comment above?  "is the only property" isn't correct anymore.  Not worth waiting for a build if we want to get his in though.

seanbrereton (dismissed):
LGTM (gradle files only)

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
```diff
@@ -78,7 +78,8 @@ class ExternalEventManagerImpl(
         externalEventResponse: ExternalEventResponse
     ): ExternalEventState {
         val requestId = externalEventResponse.requestId
-        log.debug { "Processing received external event response with id $requestId" }
+        log.info("Processing response for external event with id '{}'", requestId)
```
fowlerrr: I'm not sure about raising the level of these.  We were asked to reduce the volume of logging above INFO level.

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
```diff
@@ -187,6 +197,8 @@ class ExternalEventManagerImpl(
         val eventToSend = externalEventState.eventToSend
         eventToSend.timestamp = instant
         externalEventState.sendTimestamp = instant.plusMillis(config.getLong(FlowConfig.EXTERNAL_EVENT_MESSAGE_RESEND_WINDOW))
+        log.info("Dispatching external event with id '{}' to '{}'", externalEventState.requestId, eventToSend.topic)
```
fowlerrr: Same here.  Are we introducing a lot more logging now?

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
jujoramos: I might have misunderstood @mnesbit 's comment the other day, but I think the idea is to be able to track down whenever an event leaves/enters any particular worker so yeah, we deliberately increased the level of this message.

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
jujoramos: Same as above, yes.

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
driessamyn: I agree with @fowlerrr . 
This feels to me like a tracing thing, not a logging thing.

components/flow/flow-rest-resource-service-impl/src/main/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImpl.kt
jujoramos: 👍

components/flow/flow-rest-resource-service-impl/src/test/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImplTest.kt
```diff
@@ -190,9 +191,9 @@ class FlowRestResourceImplTest {
 
         val flowRestResource = createFlowRestResource()
 
-        assertThrows<ResourceNotFoundException> {
+        assertThatThrownBy {
```
dickon: Is there an issue with the previous Jupiter form of `assertThrows`? I don't really mind either way but without knowing any better I'd probably have used the slightly more compact variant we had before.

(I ask for the sake of general curiousity rather than any objection here).

dickon (commented):
Nice. LGTM. I'd not looked much at MDC before but having consistent attributes we can pull out should be handy for difficult issues we might hit in the future.

components/flow/flow-rest-resource-service-impl/src/test/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImplTest.kt
jujoramos: No issues at all, just a matter of preference I guess. I had to add an extra assertion so went straight for `assertj`.

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
jujoramos: Just spoke with @JamesHR3 and @mnesbit regarding this: the amount of logging will certainly be increased across the board but it is still something we need in order to trace the entire flow execution across workers. We'll likely need to investigate tools and/or other formats in the future to reduce the amount of space required, but for the time being logging is what we have.

kyriathar (dismissed):
LGTM (crypto files)

components/flow/flow-rest-resource-service-impl/src/test/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImplTest.kt
fowlerrr: I prefer what you've put in.  It's slightly more readable, to me.

fowlerrr (dismissed):

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
driessamyn: Worth looping in @davidcurrie as it will have implications on the deployed cluster.

components/flow/flow-rest-resource-service-impl/src/test/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImplTest.kt
driessamyn, line 1: The majority of change in this file are stylistic changes that have nothing to do with the purpose of this PR.
I think they should be undone. 
`AssertThrows` is consistently used across the codebase, so I don't think we need to change it for the sake of it.
It just makes reviewing the PR harder, which is already hard to review due to the source branch being on a fork.

driessamyn (changes requested):

libs/utilities/src/main/kotlin/net/corda/utilities/LogUtils.kt
```diff
@@ -7,6 +7,15 @@ import java.util.Collections
...
 import org.slf4j.MDC
 
+/**
+ * Common MDC properties used across corda.
+ */
+const val MDC_CLIENT_ID = "corda.client.id"
+const val MDC_FLOW_ID = "flow.id"
+const val MDC_VNODE_ID = "vnode.id"
+const val MDC_SESSION_EVENT_ID = "session.event.id"
+const val MDC_EXTERNAL_EVENT_ID = "corda.external.event.id"
```
driessamyn, line 22: Should these be corda-api if we expect users to be able to filter/search for these?

components/flow/flow-rest-resource-service-impl/src/test/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImplTest.kt
jujoramos: I'd argue that `assertj` is also consistently used across the code base 🤷‍♂️:

```
➜  corda-runtime-os git:(feature/CORE-12440) ✗ grep -I -R "org.assertj.core.api.Assertions" | sort | wc -l
     837
➜  corda-runtime-os git:(feature/CORE-12440) ✗ grep -I -R "org.junit.jupiter.api.Assertions" | sort | wc -l
     759
```

I've reverted the change, anyways.

---

Aside from that, we should probably choose a single testing library and go for it to prevent these frictions (same applies for the "branch on main repo" vs "branch on personal fork" approaches).

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
davidcurrie: Does it need to be at a level that we have enabled by default though vs modifying the level when we want to collect this information?

components/flow/flow-rest-resource-service-impl/src/test/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImplTest.kt
driessamyn: I agree both are used. My point is there's no point in changing just for the sake of preference, unless you're going to suggest standardizing on one and make the change across the board. Otherwise it's just bike shedding and that's expensive when half a dozen people need to review those changes.

> Aside from that, we should probably choose a single testing library and go for it to prevent these frictions

This may have been a good candidate to look at for the dev friction event. But this is only successful if it ends up removing one of the usages, not change the balance marginally from one to the other.

> same applies for the "branch on main repo" vs "branch on personal fork" approaches

I don't know anybody else in the team who uses personal forks.
I can't really see any advantages either.

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
jujoramos: Yes, the ask was to have this logged under the default `log-level` so any issues can be analysed straight away (otherwise we would need to tell customers to change the `log-level` and reproduce the problem).

components/flow/flow-rest-resource-service-impl/src/test/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImplTest.kt
jujoramos: 👍

davidcurrie: > I don't know anybody else in the team who uses personal forks.

I do

> I can't really see any advantages either.

We wouldn't have 401 branches in `corda-runtime-os` for starters...

I'm intrigued to know what your review process is such that a PR from a fork is harder. It's a pattern well-supported in the GitHub UI, the GitHub CLI, and in IntelliJ.

jujoramos: > I can't really see any advantages either.

The first one that comes to mind is having a "clean" repository with only active branches, specifically for `develop` and released version support (such as `support/5.0`, `support/5.1`, `support/6.X`, whatever).

jujoramos: Ah, @davidcurrie beat me to it.

libs/utilities/src/main/kotlin/net/corda/utilities/LogUtils.kt
davidcurrie: fwiw: PS were asking recently if there was a document that covered these MDC keys

components/flow/flow-rest-resource-service-impl/src/test/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImplTest.kt
driessamyn: We don't have `develop` or `support` branches.
I think branch clean-up is a completely separate issue.

> I'm intrigued to know what your review process is such that a PR from a fork is harder.

`git fetch` doesn't get the forked branches, so checking out the forked branch is just more hassle.

driessamyn (dismissed):
approved for packaging team.

libs/utilities/src/main/kotlin/net/corda/utilities/LogUtils.kt
jujoramos: Maybe in the future as part of another `PR` (probably in `5.0.1`)?. I want to get this merged before the deadline for `5.0.0`, and I would put extra thought regarding the actual names used within the MDC if we are gonna make this part of the public api.

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
davidcurrie: This is a point of contention between myself and @mnesbit. If we insist on abusing logging in this manner then we should at least update the standards that @fowlerrr documented [here](https://github.com/corda/platform-eng-design/blob/master/core/corda-5/corda-5.0/standards/logging.md) to justify it.

libs/utilities/src/main/kotlin/net/corda/utilities/LogUtils.kt
davidcurrie: Just because it's not in `corda-api`, don't assume that people won't come to rely on it.

jujoramos: > Just because it's not in corda-api, don't assume that people won't come to rely on it.

Agreed, but having these constants defined within `corda-api` has stronger implications. I'm not saying that we shouldn't do it, just that we shouldn't do it now as part of this `PR`.

components/flow/flow-rest-resource-service-impl/src/test/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImplTest.kt
jujoramos: These changes have been reverted, resolving this discussion now.

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
mnesbit: I agree that we are likely logging too much into what is essentially a human readable stream. The problem is these events are BAU critical. Do we have a high volume tracing/logging mechanism?

seanbrereton (dismissed):
LGTM (build files)

libs/utilities/src/main/kotlin/net/corda/utilities/LogUtils.kt
davidcurrie: I really just meant: if you think the names should change, now's the time to change them wherever they might be defined.

components/flow/flow-rest-resource-service-impl/src/test/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImplTest.kt
davidcurrie:
```
gh pr checkout 3656
```

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
davidcurrie: > Do we have a high volume tracing/logging mechanism?

No - because we didn't prioritise that for 5.0. I still think there's a distinction between what we're looking for here in terms of performance tuning (which is definitely in the realm of tracing), and what I think you're suggesting a customer wants in terms of understanding the state of a flow execution (which I'd suggest should be a part of the API).

davidcurrie: As a compromise, perhaps we could put these flow events on a separate logger? That would make it easier to enable/disable them en masse.

jujoramos: @davidcurrie : I've added an extra [commit](https://github.com/corda/corda-runtime-os/pull/3656/commits/90eabe1a6f5d0f446c1df5d9df0ce9ffbcd08aa9) with some changes, flow tracing is enabled by default but it can be optionally removed by customers through the following configuration:
```
<MarkerFilter marker="FLOW_TRACING" onMatch="DENY" onMismatch="NEUTRAL"/>
```

Sounds reasonable?.

davidcurrie: Thanks, @jujoramos. I had no idea that markers were even a thing! That looks reasonable (given where we are in the release).

davidcurrie (approved):

seanbrereton (approved):

driessamyn (approved):

PR #3657, opened 2023-04-19
CORE-11167 Pre-install checks for Corda Cluster

jacob-scott3141:
## Description
This PR adds a new plugin for the Corda CLI, called preinstall, which performs pre-install checks for the Corda platform. The plugin provides several subcommands, such as check-limits, check-postgres, check-kafka, and run-all.

## Motivation
The pre-install plugin is designed to help in the deployment of Corda to ensure that the Corda environment is properly configured and set up before installing the platform. By running the preinstall checks, users can more easily identify and fix potential issues with resource limitations, Postgres connectivity / DB issues, and Kafka cluster connectivity / broker issues.

## Usage
The easiest way to test the preinstall plugin is using the CLI-Plugin-Host. To do this, first compile the plugin using Gradle:

`./gradlew clean :tools:plugins:preinstall:build`

Then, copy the build files from corda-runtime-os to corda-cli-plugin-host:

`cp ./corda-runtime-os/tools/plugins/preinstall/build/libs/preinstall-cli-plugin-*.jar ./corda-cli-plugin-host/build/plugins/`

Finally, run the plugin from the CLI plugin host with:

`./gradlew run --args="preinstall <subcommand> [options] <path>"`

Replace `<subcommand>` with the desired subcommand, `<path>` with the path to the YAML file containing the relevant configuration settings, and `[options]` with any optional flags supported by the subcommand. See the` README.md` for more info.

For example, to run all pre-install checks, use the run-all subcommand:

`./gradlew run --args="preinstall run-all -n corda /path/to/config.yaml"`


tools/plugins/preinstall/README.md
```diff
@@ -0,0 +1,76 @@
...
+	  -d, --debug             Show information about kafka config creation for debugging 
+                              purposes
+	  -f, --file=<truststoreLocation>
+                              The file location of the truststore if TLS is enabled
+	  -n, --namespace=<namespace>
+                              The namespace in which to look for the Kafka secrets if TLS or 
+                              SASL is enabled
+	  -r, --replicas=<replicaCount>
+                              The replica count of the Kafka cluster, if you want to check the 
+                              number of brokers are correctly configured
```
davidcurrie: were you unable to get this from the admin API then?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,151 @@
...
+class CheckKafka : Runnable, PluginContext() {
+
+    @Parameters(index = "0", description = ["The yaml file containing the Kafka, SASL, and TLS configurations"])
+    lateinit var path: String
+
+    @Option(names = ["-n", "--namespace"], description = ["The namespace in which to look for the Kafka secrets if TLS or " +
+            "SASL is enabled"])
+    var namespace: String? = null
+
+    @Option(names = ["-f", "--file"], description = ["The file location of the truststore if TLS is enabled"])
```
davidcurrie: Are we not able to retrieve the truststore from the secret given in the config?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,151 @@
...
+                    || yaml.kafka.tls.truststore.password?.valueFrom?.secretKeyRef?.name != null
+                ) {
+                    props["ssl.truststore.password"] = truststorePassword
+                }
+            }
+        } else if (yaml.kafka.sasl.enabled) {
+            props["security.protocol"] = "SASL_PLAINTEXT"
+            props["sasl.mechanism"] = yaml.kafka.sasl.mechanism
+            props["sasl.jaas.config"] = "org.apache.kafka.common.security.scram.ScramLoginModule required username=" +
+                    "\"$saslUsername\" password=\"$saslPassword\" ;"
```
davidcurrie: still need the option for `PLAIN` here

tools/plugins/preinstall/README.md
davidcurrie: actually, when I then see how this is used, I'm confused about what this is meant to represent - is it the number of replicas that have been specified for the Corda topics? We should be taking that from the bootstrap properties (and it doesn't make sense to check if we're not creating the topics).

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,151 @@
...
+        var saslUsername = ""
+        var saslPassword = ""
+        var truststorePassword = ""
+
+        if (yaml.kafka.sasl.enabled) {
+            try {
+                    saslUsername = getCredentialOrSecret(yaml.kafka.sasl.username!!, namespace) ?: return
+                    saslPassword = getCredentialOrSecret(yaml.kafka.sasl.password!!, namespace) ?: return
+
+            } catch (_: NullPointerException) {
```
davidcurrie: not nice - put an explicit null check up front

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,151 @@
...
+                    props["sasl.jaas.config"] =
+                        "org.apache.kafka.common.security.plain.ScramLoginModule required username=" +
+                                "\"$saslUsername\" password=\"$saslPassword\" ;"
+                }
+            } else {
+                props["security.protocol"] = "SSL"
+            }
+            yaml.kafka.tls.truststore?.valueFrom?.secretKeyRef?.name?.let {
+                props["ssl.truststore.location"] = truststoreLocation ?: run {
+                    log("If SSL is enabled, you must provide the location of the truststore file with -f.", ERROR)
```
davidcurrie: it would be better to propagate an exception here than have to handle the nulls

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
```diff
@@ -0,0 +1,108 @@
...
+            return false
+        }
+
+        return limit >= request
+    }
+
+    // use the checkResource function to check each individual resource
+    private fun checkResources(resources: ResourceConfig, name: String): Boolean {
+        val requests: ResourceValues = resources.requests
+        val limits: ResourceValues = resources.limits
```
davidcurrie: what if they're not set? or the limits are set and not the resources or vice-versa?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
```diff
@@ -0,0 +1,108 @@
...
+        log("Requests: \n\t memory - ${requests.memory}\n\t cpu - ${requests.cpu}", INFO)
+        log("Limits: \n\t memory - ${limits.memory}\n\t cpu - ${limits.cpu}", INFO)
+
+        val check = checkResource(requests.memory, limits.memory) and checkResource(requests.cpu, limits.cpu)
+
+        if (check) {
+            log("Resource requests for $name are appropriate and are under the set limits\n", INFO)
+        }
+        else {
+            log("Resource requests for $name have been exceeded!\n", ERROR)
```
davidcurrie: it would be preferable to be explicit about exactly where the problem lies here e.g. `Memory request ${requests.memory} exceeds limit ${limits.memory} for $name`

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
```diff
@@ -0,0 +1,108 @@
...
+
+        yaml.bootstrap?.let { check = check && checkResources(it.resources, "bootstrap") }
+        yaml.workers?.db?.let { check = check && checkResources(it.resources, "DB") }
+        yaml.workers?.flow?.let { check = check && checkResources(it.resources, "flow") }
+        yaml.workers?.membership?.let { check = check && checkResources(it.resources, "membership") }
+        yaml.workers?.rest?.let { check = check && checkResources(it.resources, "rest") }
+        yaml.workers?.p2pLinkManager?.let { check = check && checkResources(it.resources, "P2P link manager") }
+        yaml.workers?.p2pGateway?.let { check = check && checkResources(it.resources, "P2P gateway") }
+
+        if (check) {println("[INFO] All resource requests are appropriate and are under the set limits.")}
```
davidcurrie: it would be good if the return code for the command reflected whether the checks passed or not

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
```diff
@@ -0,0 +1,108 @@
...
+        val limits: ResourceValues = resources.limits
+
+        log("${name.uppercase()}:", INFO)
+        log("Requests: \n\t memory - ${requests.memory}\n\t cpu - ${requests.cpu}", INFO)
+        log("Limits: \n\t memory - ${limits.memory}\n\t cpu - ${limits.cpu}", INFO)
+
+        val check = checkResource(requests.memory, limits.memory) and checkResource(requests.cpu, limits.cpu)
+
+        if (check) {
+            log("Resource requests for $name are appropriate and are under the set limits\n", INFO)
```
davidcurrie: we don't really check they are "appropriate" at the moment

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckPostgres.kt
```diff
@@ -0,0 +1,53 @@
...
+class CheckPostgres : Runnable, PluginContext(){
+
+    @Parameters(index = "0", description = ["The yaml file containing either the username and password value, " +
+            "or valueFrom.secretKeyRef.key fields for Postgres"])
+    lateinit var path: String
+
+    @Option(names = ["-n", "--namespace"], description = ["The namespace in which to look for the secrets if there are any"])
+    var namespace: String? = null
+
+    @Option(names = ["-v", "--verbose"], description = ["Display additional information when connecting to postgres"])
```
davidcurrie: We should probably consistently use `PostgreSQL`

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckPostgres.kt
```diff
@@ -0,0 +1,53 @@
...
+    @Option(names = ["-d", "--debug"], description = ["Show extra information while connecting to Postgres for debugging purposes"])
+    var debug: Boolean = false
+
+    override fun run() {
+        register(verbose, debug)
+
+        val yaml: DB = parseYaml<DB>(path) ?: return
+
+        val username: String = getCredentialOrSecret(yaml.db.cluster.username, namespace) ?: return
+        val password: String = getCredentialOrSecret(yaml.db.cluster.password, namespace) ?: return
```
davidcurrie: potentially there also the bootstrap DB credentials to check

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,267 @@
...
+    private companion object {
+        private val logger: Logger = LoggerFactory.getLogger(this::class.java)
+    }
+
+    override fun start() {
+        logger.info("starting preinstall plugin")
+    }
+
+    override fun stop() {
+        logger.info("stopping preinstall plugin")
```
davidcurrie: elsewhere I've lowered the priority of these as they're not very useful `info` messages

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,267 @@
...
+
+        // used for logging
+        fun log(s: String, level: Int) {
+            when (level) {
+                ERROR -> println("[ERROR] $s")
+                INFO -> if (verbose) println("[INFO] $s")
+                DEBUG -> if (debug) println("[DEBUG] $s")
+                WARN -> println("[WARN] $s")
+            }
+        }
```
davidcurrie: why not actually use a logging library as we do in the other plugins?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,267 @@
...
+
+        // parse a yaml file, and return an object of type T or null if there was an error
+        inline fun <reified T> parseYaml(path: String): T? {
+            log("Working Directory = ${System.getProperty("user.dir")}\n", INFO)
```
davidcurrie: what's this used for?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,267 @@
...
+
+            val file = File(path)
+
+            if(!file.isFile) {
+                log("File does not exist", ERROR)
+                return null
```
davidcurrie: use exceptions

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,267 @@
...
+            }
+
+            return try {
+                val mapper: ObjectMapper = YAMLMapper()
+                mapper.readValue(file, T::class.java)
+            } catch ( e: ValueInstantiationException) {
+                log("Could not parse the YAML file at $path: ${e.message}", ERROR)
+                null
```
davidcurrie: exceptions

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,267 @@
...
+                if (secretKey == null)  {
+                    log("Username secret key could not be parsed.", ERROR)
+                    return null
+                }
+                if (secretName == null) {
+                    log("Username secret name could not be parsed.", ERROR)
+                    return null
+                }
+                val encoded = getSecret(secretName, secretKey, namespace) ?: run{
+                    log("Username secret could not be found in namespace $namespace.", ERROR)
```
davidcurrie: the log messages say `Username` but it won't always be a username

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,267 @@
...
+        fun getCredentialOrSecret(values: SecretValues, namespace: String?): String? {
+            val secretKey: String? = values.valueFrom?.secretKeyRef?.key
+            val secretName: String? = values.valueFrom?.secretKeyRef?.name
+            var credential: String? = values.value
+
+            credential = credential ?: run {
+                if (namespace == null) {
+                    log("No namespace has been specified. If the username is supposed to be in a secret, " +
+                            "specify a namespace with -n or --namespace.", ERROR)
+                    return null
```
davidcurrie: we should default to using whatever namespace the kubernetes context is currently pointing at

tools/plugins/preinstall/src/test/kotlin/net/corda/cli/plugins/preinstall/CheckSecretsTests.kt
```diff
@@ -0,0 +1,60 @@
...
+import net.corda.cli.plugins.preinstall.PreInstallPlugin.PluginContext
+import net.corda.cli.plugins.preinstall.PreInstallPlugin.SecretValues
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Nested
+import picocli.CommandLine
+
+
+@CommandLine.Command(name = "sub-command", description = ["Example subcommand."])
```
davidcurrie: ?!?

tools/plugins/preinstall/README.md
jacob-scott3141: In the extension I thought about checking that the number of brokers was configured correctly - so I was thinking for example the replicaCount from prereqs-eks.yaml could be used to make sure the number of brokers is greater than the replica count

davidcurrie: but you connect to Kafka and ask it how many brokers there are - can't you just use that?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
jacob-scott3141: Yes, I think I can use ssl.truststore.certificates as an alternative to ssl.truststore.location

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
jacob-scott3141: Is Plain ever used in the case where TLS is not enabled? I thought it was written this way to prevent this

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
jacob-scott3141: This was useful for passing a relative path for file parsing

davidcurrie: Perhaps more useful would be just to output the absolute path of the file that you're loading.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
davidcurrie: Written where? The combination doesn't make great sense (sending credentials in plain across non-TLS connection) but I don't think Kafka prevents it.

jacob-scott3141: The bootstrap template here https://github.com/corda/corda-runtime-os/blob/release/os/5.0/charts/corda-lib/templates/_bootstrap.tpl#L184

davidcurrie: Either that was a response to a different comment or it's the wrong line!

jacob-scott3141: Apologies, updated the link just now

davidcurrie: yeah - I think that's just wrong!

jacob-scott3141: I just assumed it was written to prevent the use of SASL Plain without TLS

davidcurrie: if we did want to prevent it, it would be better to explicitly fail on that combination rather than just specifying the wrong login module...

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -97,35 +97,51 @@ class PreInstallPlugin(wrapper: PluginWrapper) : Plugin(wrapper) {
...
+                        .withMasterUrl(url)
+                        .build()
+                    KubernetesClientBuilder().withConfig(kubeConfig).build()
+                } else {
+                    KubernetesClientBuilder().build()
+                }
+
+                // if no namespace is set, will try and use the default. Worst case scenario, if no default is set, it will use "default".
+                val kubeNamespace = namespace ?: run {
+                    client.namespace ?: "default"
```
davidcurrie: That's not quite what I meant... I meant use the namespace that the provided Kubernetes context is pointing at. Typically that's what you get if you don't specify a namespace at all.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/RunAll.kt
```diff
@@ -0,0 +1,65 @@
...
+
+import picocli.CommandLine
+
+@CommandLine.Command(name = "run-all", description = ["Runs all preinstall checks."])
+class RunAll : Runnable {
+
+    @CommandLine.Parameters(index = "0", description = ["The yaml file containing all configurations"])
+    lateinit var path: String
+
+    @CommandLine.Option(names = ["-n", "--namespace"], description = ["The namespace in which to look for both the Postgres " +
```
yarivus: I think breaking the string and concat using '+' is less readable ...
Either leave as 1 line or better break the annotation to parameter on each newline

yarivus: Fix other occurrences

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,322 @@
...
+            }
+        }
+
+        // parse a yaml file, and return an object of type T or null if there was an error
+        inline fun <reified T> parseYaml(path: String): T? {
+            log("Working Directory = ${System.getProperty("user.dir")}\n", INFO)
+
+            val file = File(path)
+
+            if(!file.isFile) {
```
yarivus:
```suggestion
            if (!file.isFile) {
```

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,176 @@
...
+import java.util.Properties
+import java.util.concurrent.ExecutionException
+
+@CommandLine.Command(name = "check-kafka", description = ["Check that Kafka is up and that the credentials work."])
+class CheckKafka : Runnable, PluginContext() {
+
+    @Parameters(index = "0", description = ["The yaml file containing the Kafka, SASL, and TLS configurations"])
+    lateinit var path: String
+
+    @Option(names = ["-n", "--namespace"], description = ["The namespace in which to look for the Kafka secrets if TLS or " +
```
yarivus: see other comment of mine regarding breaking the strings and concat ... fix everywhere please

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,176 @@
...
+    fun connect(client: KafkaAdmin) {
+        val nodes: Collection<Node>? = client.getNodes()
+        val clusterID = client.getDescriptionID()
+
+        if (nodes.isNullOrEmpty()) {
+            log("There are no brokers in the kafka cluster.", ERROR)
+            return
+        }
+
+        println("[INFO] Kafka client connected to cluster with ID ${clusterID}.")
```
yarivus: use logger ?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckPostgres.kt
```diff
@@ -0,0 +1,56 @@
...
+import picocli.CommandLine
+import picocli.CommandLine.Option
+import picocli.CommandLine.Parameters
+import java.sql.DriverManager
+import java.sql.SQLException
+import net.corda.cli.plugins.preinstall.PreInstallPlugin.DB
+import net.corda.cli.plugins.preinstall.PreInstallPlugin.PluginContext
+
+@CommandLine.Command(name = "check-postgres", description = ["Check that the PostgreSQL DB is up and that the credentials work."])
+class CheckPostgres : Runnable, PluginContext(){
```
yarivus:
```suggestion
class CheckPostgres : Runnable, PluginContext() {
```

tools/plugins/preinstall/src/test/resources/KafkaTest0.yaml
yarivus, line 1: maybe better rename the yaml file to a name which indicates what it's testing for.
For example: KafkaTest0.yaml => kafka-test-sasl-plain.yaml

yarivus (changes requested):

tools/plugins/preinstall/README.md
jacob-scott3141: I get what you meant now, the changes should be in place

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
jacob-scott3141: This is done now

yarivus (dismissed):
LGTM

yarivus (dismissed):
LGTM

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
```diff
@@ -71,9 +71,9 @@ class CheckLimits : Callable<Int>, PluginContext() {
         val requests: ResourceValues = resources.requests
         val limits: ResourceValues = resources.limits
 
-        log("${name.uppercase()}:", INFO)
-        log("Requests: \n\t memory - ${requests.memory}\n\t cpu - ${requests.cpu}", INFO)
-        log("Limits: \n\t memory - ${limits.memory}\n\t cpu - ${limits.cpu}", INFO)
+        getLogger().info("${name.uppercase()}:")
```
yarivus: wouldn't it be easier to get the logger once instead of each log line ?

yarivus (commented):
see comment / question

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,273 @@
...
+    lateinit var path: String
+
+    @Option(
+        names = ["-n", "--namespace"],
+        description = ["The namespace in which to look for both the Postgres and Kafka secrets"]
+    )
+    var namespace: String? = null
+
+    @Option(
+        names = ["-u", "--url"],
```
davidcurrie: I really wouldn't try and support this option. I can't imagine a scenario where the URL is going to be enough on its own (you're always going to need some credentials as well at a minimum) and rather than trying to expose all of the client configuration options I'd just require them to have set up the Kubernetes context correctly in advance of calling the CLI.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,273 @@
...
+        description = ["The kubernetes cluster URL (if the preinstall is being called from outside the cluster)"]
+    )
+    var url: String? = null
+
+    @Option(
+        names = ["-f", "--file"],
```
davidcurrie: Likewise, I'd keep it simple and not support this option either. We're expecting to be able to retrieve secrets from Kubernetes so why not the same for the truststore?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,273 @@
...
+        description = ["The file location of the truststore for Kafka"]
+    )
+    var truststoreLocation: String? = null
+
+    @Option(
+        names = ["-t", "--timeout"],
+        description = ["The timeout in milliseconds for testing the kafka connection - defaults to 3000"]
```
davidcurrie:
```suggestion
        description = ["The timeout in milliseconds for testing the Kafka connection - defaults to 3000"]
```

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,273 @@
...
+    )
+    var timeout: Int = 3000
+
+    @Option(
+        names = ["-m", "--max-idle"],
+        description = ["The maximum ms a connection can be idle for while testing the kafka connection - defaults to 5000"]
```
davidcurrie: Is this something that we really need to expose?
```suggestion
        description = ["The maximum milliseconds a connection can be idle for while testing the Kafka connection - defaults to 5000"]
```

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,273 @@
...
+    class BrokerException(message: String) : Exception(message)
+
+    open class KafkaAdmin(props: Properties) {
+        private val admin: AdminClient?
+
+        init {
+            admin = AdminClient.create(props)
+        }
+
+        open fun getNodes(): Collection<Node>? {
```
davidcurrie: nit: a friendlier signature would be to always return at least an empty collection

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    class KafkaProperties(private val yaml: Kafka) {
+        var saslUsername: String? = null
+        var saslPassword: String? = null
+        var truststorePassword: String? = null
+        var truststoreLocation: String? = null
+        var truststoreFile: String? = null
+        var timeout: Int = 3000
+        var connectionsMaxIdleMs: Int = 5000
```
davidcurrie: There's a slightly strange split of responsibilities here. Either have this class do all the work of extracting the properties from YAML, or have all that work done outside and pass in the extracted values.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,273 @@
...
+                    } else {
+                        props["sasl.jaas.config"] =
+                            "org.apache.kafka.common.security.scram.ScramLoginModule required username=" +
+                                    "\"${saslUsername}\" password=\"${saslPassword}\" ;"
+                    }
+                } else {
+                    props["security.protocol"] = "SSL"
+                }
+
+                props["ssl.truststore.type"] = yaml.kafka.tls.truststore!!.type.uppercase()
```
davidcurrie: I suspect the `!!` here is indicative of that strange split as I'm guessing you've already verified somewhere else that the truststore is not `null`.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,273 @@
...
+                } else {
+                    throw TruststoreNotFoundException(
+                        "If SSL is enabled, you must provide either a truststore.valueFrom.secretKeyRef entry, " +
+                                "or specify the location of a truststore file with -f."
+                    )
+                }
+
+            } else if (yaml.kafka.sasl.enabled) {
+                props["security.protocol"] = "SASL_PLAINTEXT"
+                props["sasl.mechanism"] = yaml.kafka.sasl.mechanism
```
davidcurrie: As previously discussed, if you don't want to support `PLAIN` here (which is a reasonable stance) then we should flag that as an error.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,273 @@
...
+        if (nodes.isNullOrEmpty()) {
+            report.addEntry(ReportEntry("Kafka cluster has brokers", false))
+            return
+        }
+
+        getLogger().info("Kafka client connected to cluster with ID ${clusterID}.")
+        getLogger().info("Number of brokers: ${nodes.size}")
+        replicas?.let {
+            if (nodes.size < it) {
+                throw BrokerException("Number of brokers (${nodes.size}) is less than replica count.")
```
davidcurrie: why is this not just a failed report?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,273 @@
...
+            }
+            try {
+                saslUsername = getCredentialOrSecret(yaml.kafka.sasl.username, namespace, url)
+                saslPassword = getCredentialOrSecret(yaml.kafka.sasl.password, namespace, url)
+            } catch (e: Exception) {
+                report.addEntry(ReportEntry("Get SASL credentials", false, e))
+                getLogger().error(report.failingTests())
+                return 1
+            }
+        }
```
davidcurrie: I'd be tempted to factor this differently into three separate checks - one for the mechanism, one for username, and one for password.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,273 @@
...
+            return props
+        }
+    }
+
+    // Connect to kafka using the properties assembled earlier
+    fun connect(client: KafkaAdmin, replicas: Int?) {
+        val nodes: Collection<Node>? = client.getNodes()
+        val clusterID = client.getDescriptionID()
+
+        report.addEntry(ReportEntry("Connect to Kafka cluster using client", true))
```
davidcurrie, line 125: At least as far as it is possible, I'd expect to always see the same set of entries in the results and just different combinations of pass/fail.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
```diff
@@ -0,0 +1,125 @@
...
+import net.corda.cli.plugins.preinstall.PreInstallPlugin.PluginContext
+import picocli.CommandLine
+import picocli.CommandLine.Parameters
+import picocli.CommandLine.Option
+import java.util.concurrent.Callable
+
+@CommandLine.Command(name = "check-limits", description = ["Check the resource limits have been assigned correctly."])
+class CheckLimits : Callable<Int>, PluginContext() {
+
+    @Parameters(index = "0", description = ["YAML file containing resource limit overrides for the corda install"])
```
davidcurrie:
```suggestion
    @Parameters(index = "0", description = ["YAML file containing resource limit overrides for the Corda install"])
```

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
```diff
@@ -0,0 +1,125 @@
...
+    var verbose: Boolean = false
+
+    @Option(names = ["-d", "--debug"], description = ["Show information about limit calculation for debugging purposes"])
+    var debug: Boolean = false
+
+    class ResourceLimitsExceededException(message: String) : Exception(message)
+
+    private var hasCheckedLimits = false
+
+    // split resource into a digit portion and a unit portion
```
davidcurrie, line 27: and... the function does a bit more than that!

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
davidcurrie: if this is to provide context for the next two lines then they should really all be one log entry

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckPostgres.kt
```diff
@@ -0,0 +1,91 @@
...
+import net.corda.cli.plugins.preinstall.PreInstallPlugin.PluginContext
+import java.util.concurrent.Callable
+
+@CommandLine.Command(name = "check-postgres", description = ["Check that the PostgreSQL DB is up and that the credentials work."]
+)
+class CheckPostgres : Callable<Int>, PluginContext() {
+
+    @Parameters(
+        index = "0",
+        description = ["YAML file containing either the username and password value, or valueFrom.secretKeyRef.key fields for PostgreSQL"]
```
davidcurrie: It's not really just the `key` fields though. Maybe something like "YAML file containing the username and password values for PostgreSQL either as values or as secret references"

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckPostgres.kt
```diff
@@ -0,0 +1,91 @@
...
+            getLogger().error(report.failingTests())
+            return 1
+        }
+
+        val postgresUrl = "jdbc:postgresql://${yaml.db.cluster.host}:${yaml.db.cluster.port}/postgres"
+
+        try {
+            Class.forName("org.postgresql.Driver")
+            val connection = DriverManager.getConnection(postgresUrl, username, password)
+            if (connection.isValid(0)) {
```
davidcurrie: if `isValid` returns `false`, what does that mean? should we be creating a report entry for that?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    override fun stop() {
+        logger.debug("stopping preinstall plugin")
+    }
+
+    @Extension
+    @CommandLine.Command(name = "preinstall",
+        subcommands = [CheckLimits::class, CheckPostgres::class, CheckKafka::class, RunAll::class],
+        description = ["Preinstall checks for corda."])
```
davidcurrie:
```suggestion
        description = ["Preinstall checks for Corda."])
```

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,321 @@
...
+        }
+
+        fun getLogger(): Logger {
+            return logger
+        }
+
+        fun register(verbose: Boolean, debug: Boolean) {
+            this.verbose = verbose
+            this.debug = debug
+        }
```
davidcurrie: There seems to be something a bit circular here. The subcommands set these into the context so that they can retrieve them again?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,321 @@
...
+                throw FileNotFoundException("Could not read file at $path.")
+            }
+
+            val mapper: ObjectMapper = YAMLMapper()
+            return mapper.readValue(file, T::class.java)
+        }
+
+        // get the credentials (.value) or credentials from a secret (.valueFrom.secretKeyRef...) from a SecretValues
+        // object, and a namespace (if the credential is in a secret)
+        fun getCredentialOrSecret(values: SecretValues, namespace: String?, url: String?): String {
```
davidcurrie: nit: I'd just call this `getCredential` or, if you wanted to be really explicit, `getCredentialFromValueOrSecret` but it always returns a credential

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,321 @@
...
+            val secretKey: String? = values.valueFrom?.secretKeyRef?.key
+            val secretName: String? = values.valueFrom?.secretKeyRef?.name
+            var credential: String? = values.value
+
+            credential = credential ?: run {
```
davidcurrie: logic is the wrong way round - the secret takes precedence (if the name is not null/empty)

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,321 @@
...
+        private fun getSecret(secretName: String, secretKey: String, namespace: String?, url: String?): String? {
+            return try {
+                val client: KubernetesClient = if (url != null) {
+                    val kubeConfig: Config = ConfigBuilder()
+                        .withMasterUrl(url)
+                        .build()
+                    KubernetesClientBuilder().withConfig(kubeConfig).build()
+                } else {
+                    KubernetesClientBuilder().build()
+                }
```
davidcurrie: Is there a downside to continually rebuilding the client?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,321 @@
...
+                secret.data[secretKey]
+            } catch (e: KubernetesClientException) {
+                throw SecretException("Could not read secret $secretName with key $secretKey.", e)
+            }
+        }
+    }
+
+    class ReportEntry(private var check: String, var result: Boolean, var reason: Exception? = null) {
+        override fun toString(): String {
+            return "$check: ${if (result) "PASSED" else "FAILED"}"
```
davidcurrie: are we losing important information about why it failed by not including the exception message?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,321 @@
...
+        }
+    }
+
+    class Report (private var report: MutableList<ReportEntry> = mutableListOf()) {
```
davidcurrie: calling both the class and variable report is confusing - maybe `entries` for that latter?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,321 @@
...
+
+        fun addEntries(reportEntries: Report) {
+            report.addAll(reportEntries.getEntries())
+        }
+
+        // Fails (returns 1) if any of the report entries have failed. If no entries are found, the report passes.
+        fun testsPassed(): Int {
+            val result = report.reduceOrNull { acc, entry -> ReportEntry("", acc.result && entry.result) } ?: return 0
+
+            return if (result.result) 0 else 1
```
davidcurrie: I think you just want something like `if (report.all { it.result } 0 else 1`

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,321 @@
...
+        fun addEntries(entries: List<ReportEntry>) {
+            report.addAll(entries)
+        }
+
+        fun addEntries(reportEntries: Report) {
+            report.addAll(reportEntries.getEntries())
+        }
+
+        // Fails (returns 1) if any of the report entries have failed. If no entries are found, the report passes.
+        fun testsPassed(): Int {
```
davidcurrie: this is conflating a report concern (did everything pass) and a command concern (return code) - I'd just return a boolean here and then convert to a return code in the command

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,321 @@
...
+        fun failingTests(): String {
+            var acc = ""
+            for(entry in report) {
+                if (!entry.result) {
+                    val message = entry.reason?.message ?: "(No message provided)"
+                    val cause = entry.reason?.cause ?: "(No cause provided)"
+                    acc += "$entry \n\t - $message \n\t - $cause \n"
+                }
+            }
+            return acc
```
davidcurrie: this should be able to use a (suitably modified) `entry.toString` then you can have something like:

```
return report.filter { !it.result }.joinToString(separator = "\n")
```

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,321 @@
...
+        }
+
+        override fun toString(): String {
+            var acc = ""
+            for (entry in report) {
+                acc += "$entry\n"
+            }
+            return acc
```
davidcurrie:
```
report.joinToString(separator = "\n")
```

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,321 @@
...
+        @JsonProperty("limits")
+        val limits: ResourceValues
+    )
+
+    data class ResourceValues(
+        @JsonProperty("memory")
+        val memory: String,
+        @JsonProperty("cpu")
+        val cpu: String
+    )
```
davidcurrie: I think there need to be a few more `?` in here...

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/RunAll.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    @CommandLine.Parameters(
+        index = "0",
+        description = ["YAML file containing all configurations"]
+    )
+    lateinit var path: String
+
+    @CommandLine.Option(
+        names = ["-n", "--namespace"],
+        description = ["The namespace in which to look for both the Postgres and Kafka secrets"]
```
davidcurrie:
```suggestion
        description = ["The namespace in which to look for both the PostgreSQL and Kafka secrets"]
```

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/RunAll.kt
```diff
@@ -0,0 +1,95 @@
...
+        if (debug) { kafkaArgs.add("-d") }
+        namespace?.let{ kafkaArgs.add("-n$it") }
+        url?.let{ kafkaArgs.add("-n$it") }
+        truststoreLocation?.let{ kafkaArgs.add("-f$it") }
+        kafkaArgs.add("-t$timeout")
+        kafkaArgs.add("-m$maxIdleMs")
+
+        CommandLine(limitsCMD).execute(path, *limitsArgs.toTypedArray())
+        CommandLine(postgresCMD).execute(path, *postgresArgs.toTypedArray())
+        CommandLine(kafkaCMD).execute(path, *kafkaArgs.toTypedArray())
```
davidcurrie: does this gain you something over just executing `call` directly?

tools/plugins/preinstall/src/test/kotlin/net/corda/cli/plugins/preinstall/TestSubCommands.kt
```diff
@@ -0,0 +1,175 @@
...
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+import picocli.CommandLine
+
+
+class TestSubCommands {
+    @Test
+    fun testFileParsing() {
```
davidcurrie: Preference should be one test per method - so `testFileParsingResources`, `testFileParsingPostgreSQL`, ...

tools/plugins/preinstall/src/test/kotlin/net/corda/cli/plugins/preinstall/TestSubCommands.kt
```diff
@@ -0,0 +1,175 @@
...
+        var path = "./src/test/resources/LimitsTestUnderLimits.yaml"
+        val limits = CheckLimits()
+        CommandLine(limits).execute(path)
```
davidcurrie: Worth checking the return codes?

tools/plugins/preinstall/src/test/kotlin/net/corda/cli/plugins/preinstall/TestSubCommands.kt
```diff
@@ -0,0 +1,175 @@
...
+            val mockAdmin = mock<CheckKafka.KafkaAdmin>()
+            val nodes: Collection<Node> = listOf(Node(0, "localhost", 9092), Node(1, "localhost", 9093))
+            whenever(mockAdmin.getDescriptionID()).thenReturn("ClusterID")
+            whenever(mockAdmin.getNodes()).thenReturn(nodes)
+
+            val ck = CheckKafka()
+            ck.register(verbose=true, debug=false)
+            ck.connect(mockAdmin, 2)
+
+            assertTrue( ck.report.toString().contains("Connect to Kafka cluster using client: PASSED") )
```
davidcurrie, line 207: how about a test for failing to connect to Kafka?

davidcurrie (changes requested):

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/RunAll.kt
jacob-scott3141: `execute` allows you to pass the PicoCLI command line arguments to the subcommand, I don't believe you can do that by running `call`

yarivus (commented):
LGTM - but still pending comments

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
jacob-scott3141: This is optional, as it is capable of fetching the truststore from a secret. Do you think it's not worth supporting a truststore location? I originally added it to match the functionality of the helm charts.

davidcurrie: The Helm chart itself always takes the truststore as the location of an existing secret. I'd stop there given you have the logic to handle that.

yarivus (dismissed):
LGTM

tools/plugins/preinstall/src/test/kotlin/net/corda/cli/plugins/preinstall/TestSubCommands.kt
jacob-scott3141: I've added a return code for the limits file parsing, but in this case postgres and kafka will always return zero here because there is no access to a cluster or postgres / kafka instances

yarivus (dismissed):
LGTM

yarivus (dismissed):
LGTM

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,265 @@
...
+
+    @Option(
+        names = ["-n", "--namespace"],
+        description = ["The namespace in which to look for both the Postgres and Kafka secrets"]
+    )
+    var namespace: String? = null
+
+    @Option(
+        names = ["-t", "--timeout"],
+        description = ["The timeout in milliseconds for testing the kafka connection - defaults to 3000"]
```
davidcurrie:
```suggestion
        description = ["The timeout in milliseconds for testing the Kafka connection - defaults to 3000"]
```

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,265 @@
...
+
+    @Parameters(
+        index = "0",
+        description = ["YAML file containing the Kafka, SASL, and TLS configurations"]
+    )
+    lateinit var path: String
+
+    @Option(
+        names = ["-n", "--namespace"],
+        description = ["The namespace in which to look for both the Postgres and Kafka secrets"]
```
davidcurrie: I'm trying (and probably failing) to be consistent about using `PostgreSQL`, at least in external facing documentation:
```suggestion
        description = ["The namespace in which to look for both the PostgreSQL and Kafka secrets"]
```

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,265 @@
...
+                ?.get()
+                ?: listOf()
+        }
+
+        open fun getDescriptionID(): String? {
+            return admin?.describeCluster()?.clusterId()?.get()
+        }
+    }
+
+    class KafkaProperties(private val bootstrapServers: String) {
```
davidcurrie: This should now just be a Kotlin data class.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,265 @@
...
+        class SaslPlainWithoutTlsException(message: String) : Exception(message)
+
+        fun getKafkaProperties(): Properties {
+            val props = Properties()
+            props["bootstrap.servers"] = bootstrapServers
+            props["request.timeout.ms"] = timeout
+            props["connections.max.idle.ms"] = 5000
+
+            val tlsEnabled = !(truststoreFile == null && truststorePassword == null)
+            val saslEnabled = !(saslUsername == null && saslPassword == null)
```
davidcurrie: Given you've previously worked out whether TLS and SASL are enabled, it would be better to keep hold of that information rather than attempting to reverse engineer it again here.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,265 @@
...
+                props["sasl.jaas.config"] =
+                    "org.apache.kafka.common.security.scram.ScramLoginModule required username=" +
+                            "\"${saslUsername}\" password=\"${saslPassword}\" ;"
+            }
+            return props
+        }
+    }
+
+    // Connect to kafka using the properties assembled earlier
+    fun connect(client: KafkaAdmin, replicas: Int?) {
```
davidcurrie: would prefer a more meaningful name given what the method does e.g. `checkConnectionAndSufficientBroker`

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,265 @@
...
+        }
+
+        val creds = KafkaProperties(yaml.kafka.bootstrapServers)
+        saslUsername?.let{ creds.saslUsername = it }
+        saslPassword?.let{ creds.saslPassword = it }
+        saslMechanism?.let{ creds.saslMechanism = it }
+        truststorePassword?.let{ creds.truststorePassword = it }
+        truststoreFile?.let{ creds.truststoreFile = it }
+        truststoreType?.let{ creds.truststoreType = it }
+        creds.timeout = timeout
```
davidcurrie: The `let`s are a bit redundant given you can pass through null values. I suspect the `KafkaProperties` abstraction is now just getting in the way - you'd be better off just building the Kafka properties as you go through the YAML.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,265 @@
...
+            } catch (e: Exception) {
+                report.addEntry(ReportEntry("Get SASL credentials", false, e))
+                logger.error(report.failingTests())
+                return 1
+            }
+        }
+
+        if (yaml.kafka.tls.enabled) {
+            if (!checkCredential(yaml.kafka.tls.truststore,
+                    TruststoreNotFoundException("If SSL is enabled, you must provide entries for kafka.tls.truststore."))) {
```
davidcurrie: This isn't true - if the broker's certs would be trusted by the JVM (as they are for MSK, for example), you don't need to provide a truststore

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
```diff
@@ -0,0 +1,265 @@
...
+
+        var truststorePassword: String? = null
+        var truststoreFile: String? = null
+        var truststoreType: String? = null
+
+        if (yaml.kafka.sasl.enabled) {
+            if (!checkCredential(yaml.kafka.sasl.username, SASLCredentialException("If SASL is enabled, you must provide a username.")) ||
+                !checkCredential(yaml.kafka.sasl.password, SASLCredentialException("If SASL is enabled, you must provide a password.")) ||
+                !checkCredential(yaml.kafka.sasl.mechanism, SASLCredentialException("If SASL is enabled, you must provide a mechanism."))) {
+                return 1
```
davidcurrie: here and elsewhere: do we need to return here or could we continue on to perform other verifications? nothing more frustrating than having to fix one problem at a time

yarivus (dismissed):
LGTM

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
```diff
@@ -0,0 +1,135 @@
...
+        try {
+            yaml = parseYaml<Configurations>(path)
+            report.addEntry(ReportEntry("Parse resource properties from YAML", true))
+        } catch (e: Exception) {
+            report.addEntry(ReportEntry("Parse resource properties from YAML", false, e))
+            logger.error(report.failingTests())
+            return 1
+        }
+
+        yaml.resources?.let { checkResources(it, "resources") }
```
davidcurrie: we don't really care that `resources` on its own is valid - what we care about is that, for each type of worker, the combination of the worker-specific resources and the defaults specified in `resources` is valid

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckPostgres.kt
```diff
@@ -0,0 +1,85 @@
...
+            report.addEntry(PreInstallPlugin.ReportEntry("Parse PostgreSQL properties from YAML", false, e))
+            logger.error(report.failingTests())
+            return 1
+        }
+
+        // Get DB credentials from values or secrets
+        val username: String
+        val password: String
+        val dbName = yaml.bootstrap?.db?.cluster?.username ?: yaml.db.cluster.username
+        val dbPass = yaml.bootstrap?.db?.cluster?.password ?: yaml.db.cluster.password
```
davidcurrie: We should verify both sets of credentials here. Those that are in `db.cluster` and, if they are different, those in `bootstrap.db.cluster`.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckPostgres.kt
```diff
@@ -0,0 +1,85 @@
...
+            password = getCredential(dbPass, namespace)
+            report.addEntry(PreInstallPlugin.ReportEntry("Get PostgreSQL credentials", true))
+        } catch (e: Exception) {
+            report.addEntry(PreInstallPlugin.ReportEntry("Get PostgreSQL credentials", false, e))
+            logger.error(report.failingTests())
+            return 1
+        }
+
+        // Create the URL using DB host and port
+        val postgresUrl = "jdbc:postgresql://${yaml.db.cluster.host}:${yaml.db.cluster.port}/postgres"
```
davidcurrie: the database name (on the end of the URL) should also come from the YAML

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        // get the credentials (.value) or credentials from a secret (.valueFrom.secretKeyRef...) from a SecretValues
+        // object, and a namespace (if the credential is in a secret)
+        fun getCredential(values: SecretValues, namespace: String?): String {
+            val secretKey: String? = values.valueFrom?.secretKeyRef?.key
+            val secretName: String? = values.valueFrom?.secretKeyRef?.name
+            val credential: String? = values.value
+
+            if (secretKey.isNullOrEmpty() || secretName.isNullOrEmpty())  {
```
davidcurrie: although it is later enforced by the chart, we should check that if you have a `secretName` then they should have specified a `secretKey`

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,316 @@
...
+                    client.secrets().withName(secretName).get()
+                }
+                secret.data[secretKey]
+            } catch (e: KubernetesClientException) {
+                throw SecretException("Could not read secret $secretName with key $secretKey.", e)
+            }
+        }
+    }
+
+    class ReportEntry(private var check: String, var result: Boolean, var reason: Exception? = null) {
```
davidcurrie, line 110:
```suggestion
    class ReportEntry(private var check: String, var result: boolean, var reason: Exception? = null) {
```

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,316 @@
...
+        override fun toString(): String {
+            var entry = "$check: ${if (result) "PASSED" else "FAILED"}"
+            if (!result) {
+                val message = reason?.message ?: "(No message provided)"
+                val cause = reason?.cause ?: "(No cause provided)"
+                entry += "\n\t - $message \n\t - $cause"
```
davidcurrie: If they're not provided I just wouldn't add them to the message

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/RunAll.kt
davidcurrie: but you could just set them directly on the `Callable` could you not? e.g. `limitsCMD.path = path`

davidcurrie (changes requested):

yarivus (dismissed):
LGTM

yarivus (dismissed):
LGTM

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
```diff
@@ -64,29 +67,44 @@ class CheckLimits : Callable<Int>, PluginContext() {
...
 
         try {
-            if (requests.memory != null || limits.memory != null) {
-                if (requests.memory == null || limits.memory == null) {
+            if (requests?.memory == null) {
+                requests?.memory = defaultRequests?.memory
+            }
+            if (limits?.memory == null) {
+                limits?.memory = defaultLimits?.memory
+            }
```
yarivus, line 82: I don't follow this logic ... if `requests?.memory` can be null ... then why assign it the `defaultRequests?` which are assigned null as well ?!
IOW why do you need `defaultRequests` set to null at all if you used `ResourceValues?`

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
```diff
@@ -64,29 +67,44 @@ class CheckLimits : Callable<Int>, PluginContext() {
...
-                checkResource(requests.memory, limits.memory)
+                checkResource(requests.memory!!, limits.memory!!)
+            }
+
+            if (requests?.cpu == null) {
+                requests?.cpu = defaultRequests?.cpu
+            }
+            if (limits?.cpu == null) {
+                limits?.cpu = defaultLimits?.cpu
             }
```
yarivus: same comment as before

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckPostgres.kt
```diff
@@ -58,20 +78,10 @@ class CheckPostgres : Callable<Int>, PluginContext() {
...
-                    "Connect to PostgreSQL",
-                    false,
-                    Exception("Connection to PostgreSQL DB timed out.")
-                ))
-            }
-        } catch(e: SQLException) {
-            report.addEntry(PreInstallPlugin.ReportEntry("Connect to PostgreSQL", false, e))
+        connect(postgresUrl, username, password)
+
+        if (bootstrapUsername != null && bootstrapUsername != username && bootstrapPassword != null) {
```
yarivus, line 84: a comment is needed why you choose to call `connect()` again w/ different params.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -71,7 +71,7 @@ class PreInstallPlugin : Plugin() {
             val credential: String? = values.value
 
             if (secretKey.isNullOrEmpty() || secretName.isNullOrEmpty())  {
-                if (!credential.isNullOrEmpty()) {
+                if (!credential.isNullOrEmpty() && secretName.isNullOrEmpty()) {
```
yarivus: Maybe this condition should be not nested ?!
You want to return the credential when: `!credential.isNullOrEmpty() && secretName.isNullOrEmpty() && !secretKey.isNullOrEmpty()` 
At least that's what the exception msg says ;)

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckKafka.kt
jacob-scott3141: I've kept the KafkaProperties for now; for me it helps with readability to separate the parsing of the yaml and fetching of secrets from the building of the properties, but if you would prefer the properties to assemble them together I can make that change

yarivus (changes requested):

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
jacob-scott3141: The ResourceValue object can be null, in which case it is assigned `defaultRequests` - if the ResourceValue object itself is not not null, it can still have null fields for memory or cpu, in which case the default is used for that field

yarivus (dismissed):
LGTM

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,316 @@
...
+        @JsonProperty("p2pLinkManager")
+        val p2pLinkManager: Resources?,
+        @JsonProperty("p2pGateway")
+        val p2pGateway: Resources?
+    )
+
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    data class Resources(
+        @JsonProperty("resources")
+        val resources: ResourceConfig
```
davidcurrie: Should this be nullable?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
```diff
@@ -0,0 +1,159 @@
...
+            defaultRequests = it.requests
+        }
+
+        yaml.bootstrap?.let { checkResources(it.resources, "bootstrap") }
+        yaml.workers?.db?.let { checkResources(it.resources, "DB") }
+        yaml.workers?.flow?.let { checkResources(it.resources, "flow") }
+        yaml.workers?.membership?.let { checkResources(it.resources, "membership") }
+        yaml.workers?.rest?.let { checkResources(it.resources, "rest") }
+        yaml.workers?.p2pLinkManager?.let { checkResources(it.resources, "P2P link manager") }
+        yaml.workers?.p2pGateway?.let { checkResources(it.resources, "P2P gateway") }
```
davidcurrie: This means we only check the resources if there is some configuration specified for the bootstrap/worker whereas we always want to perform the checks. Currently, it won't fail if I specify invalid or no defaults but I don't override any of the worker properties. Worth adding a test for this.

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckPostgres.kt
```diff
@@ -0,0 +1,96 @@
...
+        // Create the URL using DB host and port
+        val postgresUrl = "jdbc:postgresql://${yaml.db.cluster.host}:${yaml.db.cluster.port}/${yaml.db.cluster.database}"
+        report.addEntry(PreInstallPlugin.ReportEntry("Create PostgreSQL URL with DB host", true))
+
+        // Try connecting to the DB URL using supplied credentials
+        connect(postgresUrl, username, password)
+
+        // If the bootstrap credentials exist, try connecting to the DB URL using them
+        if (bootstrapUsername != null && bootstrapUsername != username && bootstrapPassword != null) {
+            connect(postgresUrl, bootstrapUsername!!, bootstrapPassword!!)
```
davidcurrie: Could you get something in the report entry that lets the user know this is for the bootstrap credentials so they know where to look when it fails?

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/PreInstallPlugin.kt
```diff
@@ -0,0 +1,316 @@
...
+
+            if (secretKey.isNullOrEmpty() || secretName.isNullOrEmpty())  {
+                if (!credential.isNullOrEmpty() && secretName.isNullOrEmpty() && secretKey.isNullOrEmpty()) {
+                    return credential
+                }
+                throw SecretException("No value $credential and no secret $secretName with key $secretKey could be parsed.")
+            }
+            val encoded = getSecret(secretName, secretKey, namespace) ?: run {
+                throw SecretException("Secret $secretName has no key $secretKey.")
+            }
```
davidcurrie: The logic here is still not quite right - if I have a `secretName` then I must have a `secretKey`

davidcurrie (changes requested):

tools/plugins/preinstall/src/main/kotlin/net/corda/cli/plugins/preinstall/CheckLimits.kt
jacob-scott3141: Just to double check - is it supposed to fail if defaults aren't supplied?

davidcurrie: The question is not whether defaults are supplied or not, it is whether or not bootstrap and each worker have valid values (taking into account both the specific and default values). Previously I'd suggested this is where it might be useful to have different levels of severity but let's not get into that now. Let's just say that valid means that both requests and limits have been specified and limits >= requests.

yarivus (approved):
LGTM

davidcurrie (approved):
LGTM - thanks for persevering @jacob-scott3141 !

dickon:
Just a friendly reminder that we generally squash all our PRs and prefer not to have authors like "root <root@LDNW-95ZJMN3>".

PR #3658, opened 2023-04-19
CORE-9295 Removed writeAtomicTransactionRecord function as Atomic Transaction will be inserted with the schema creation

ciaran238:
The "Atomic Transaction" will be inserted with the schema creation (This change is in the corda-api), therefore the code for writing atomic transaction is being removed.
[Link](https://github.com/corda/corda-api/pull/1094) to related API changes


mbrkic-r3:
Looking at Jira ticket, I think that proper solution is to insert “Atomic Transaction” with the schema creation, but maybe I'm missing something. I've added comment to Jira ticket.

libs/messaging/db-message-bus-impl/src/integrationTest/kotlin/net/corda/messagebus/db/persistence/DBAccessIntegrationTest.kt
```diff
@@ -419,4 +420,14 @@ class DBAccessIntegrationTest {
...
         assertThat(dbAccess.getTopicPartitionMapFor(topic)).isEqualTo(expectedResult)
     }
+    @Test
+    fun `Atomic Transaction has been committed`() {
+        val result =
+            query(
+                TransactionRecordEntry::class.java,
+                "from transaction_record where transactionId = '${ATOMIC_TRANSACTION.transactionId}'"
+            ).single()
+        assertThat(result.transactionId == ATOMIC_TRANSACTION.transactionId)
```
mbrkic-r3: It should be `assertThat(result.transactionId).isEqualTo(ATOMIC_TRANSACTION.transactionId)`, similar below

fowlerrr (dismissed):

LWogan (approved):
LGTM but i am not a lead!

JamesHR3 (approved):

PR #3661, opened 2023-04-19
CORE-11532: Add a back off to crypto retries

fowlerrr:
This should ensure that we actually give ourselves time for a problem to resolve before actually retrying.

### DIFF
--- a/components/crypto/crypto-component-core-impl/src/main/kotlin/net/corda/crypto/component/impl/RetryUtils.kt
+++ b/components/crypto/crypto-component-core-impl/src/main/kotlin/net/corda/crypto/component/impl/RetryUtils.kt
@@ -1,6 +1,8 @@
 package net.corda.crypto.component.impl
 
 import org.slf4j.Logger
+import kotlin.math.pow
+import kotlin.math.roundToLong
 
 inline fun <R> retry(numRetries: Int, logger: Logger, block: () -> R): R {
     var firstException: Exception? = null
@@ -11,6 +13,7 @@ inline fun <R> retry(numRetries: Int, logger: Logger, block: () -> R): R {
             logger.warn("Exception occurred in retry block (invocation: ${i+1}, retries left: ${numRetries-i}): $e")
             if (firstException == null)
                 firstException = e
+            Thread.sleep((1000 * 1.0.pow(i+1)).roundToLong())
         }
     }
     throw firstException!!


dickon (approved):
LGTM

components/crypto/crypto-component-core-impl/src/main/kotlin/net/corda/crypto/component/impl/RetryUtils.kt
```diff
@@ -11,6 +13,7 @@ inline fun <R> retry(numRetries: Int, logger: Logger, block: () -> R): R {
             logger.warn("Exception occurred in retry block (invocation: ${i+1}, retries left: ${numRetries-i}): $e")
             if (firstException == null)
                 firstException = e
+            Thread.sleep((1000 * 1.0.pow(i+1)).roundToLong())
```
driessamyn, line 16: Should be ok for now, but I wonder if the initial sleep is a bit long. In case of a DNS resolution, for example, it could be a glitch.

fowlerrr: Going to merge, but what would you think is a decent initial sleep?

driessamyn: Y. No problem, it was a fly-by comment.
I'm thinking the first retry should maybe be after 50ms or so, then the 1s backoff after that is probably ok.

Maybe the the code is actually something like this:

```
            if (firstException == null) {
                firstException = e
               Thread.sleep(50)
           } else {
                Thread.sleep((1000 * 1.0.pow(i+1)).roundToLong())
          }
```

I.e. if it's a single glitch, we've retried straight away, but we're still patient after that.

On the other hand, I wonder actually if it'd be worth looking into something like https://github.com/resilience4j/resilience4j and have a defined and consistent strategy across the code-base to handle retries and backoffs etc.

driessamyn: @Omar-awad may have an opinion on ^

fowlerrr: I like consistency.  👍

PR #3662, opened 2023-04-19
CORE-3819: Reconfigure repositories for Corda dependencies.

chrisr3:
We no longer need the `kotlin-stdlib-jdkx-osgi` bundles. However, we can still experiment with "Kotlin as a CorDapp".


ronanbrowne (approved):
LGTM

PR #3664, opened 2023-04-20
CORE-12769 Move the JSON factory create to try-catch

nkovacsx:
### Problem overview

If the user decides to throw an exception in their `ContractStateVaultJsonFactory` implementation, or return `null` (Java only) then our logic will fill as the calling of the `create()` function is not part of the try-catch block.

### Solution

Move the `ContractStateVaultJsonFactory.create()` call to the try catch block. This way we can't log the JSON in the warn log because we might not even have a JSON. It would be possible to split this into two and have two try-catches. One for `JsonProcessingException` and one for `Exception` in general, it's probably not worth it though.

### DIFF
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
@@ -245,16 +245,15 @@ class UtxoPersistenceServiceImpl(
     ): String {
         val jsonMap = factoryStorage.getFactoriesForClass(contractState).associate {
 
-            val jsonToParse = @Suppress("unchecked_cast")
-            (it as ContractStateVaultJsonFactory<ContractState>)
-                .create(contractState, jsonMarshallingService)
-                .ifBlank { "{}" } // Default to "{}" if the provided factory returns empty string to avoid exception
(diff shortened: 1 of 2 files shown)


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
```diff
@@ -245,16 +245,15 @@ class UtxoPersistenceServiceImpl(
...
+                val jsonToParse = @Suppress("unchecked_cast")
+                (it as ContractStateVaultJsonFactory<ContractState>)
+                    .create(contractState, jsonMarshallingService)
+                    .ifBlank { "{}" } // Default to "{}" if the provided factory returns empty string to avoid exception
                 jsonMarshallingService.parse(jsonToParse, Any::class.java)
             } catch (e: Exception) {
-                log.warn("Error while processing factory for class: ${it.stateType.name}. " +
-                        "JSON that could not be processed: $jsonToParse. Defaulting to empty JSON.")
+                // We can't log the JSON string here because the logic might have failed before we have a JSON
+                log.warn("Error while processing factory for class: ${it.stateType.name}. Defaulting to empty JSON.")
```
lankydan, line 255: We could have two levels of try catch, one around the `create` and one around `parse`.

lankydan (approved):

PR #3669, opened 2023-04-20
CORE-11504 Detailed ledger logs for combined worker

relyafi:
### Summary

This PR enhances the default log4j configuration for the combined worker to capture TRACE and DEBUG log messages relating to ledger functionality. These are output to a separate file (`corda-ledger-detailed.log`) to avoid polluting the main log file and ato aid debugging ledger functionality by allowing these messages to be seen without the general noise of system activity. INFO level ledger related log messages continue to appear in the main log file in addition to the detailed log file.

This change is limited to the combined worker because it is anticipated that this logging would be most useful to a developer, who is likely to be testing against the combined worker. Separate workers are likely to be deployed in a containerized Corda cluster, where outputting logs to file will be of limited value. Such deployments will need to make their own decisions about what logging levels to enable. 

### Testing

Started up the combined worker, and ran the `smokeTest` command. The tests of particular interest are `UtxoLedgerTests` and `ConsensualLedgerTests`. Confirmed that after running these tests, TRACE and DEBUG messages appeared in the new log file, and were not observed in the primary log file (`corda.log`). INFO level ledger log messages appeared in both the detailed and primary log file. Example log file attached.

[corda-ledger-detailed.log](https://github.com/corda/corda-runtime-os/files/11286302/corda-ledger-detailed.log)


vlajos (dismissed):

kyriathar:
I had a quick look at the attached log file and I can see some WARN and INFO loggings as well, is that intended?

relyafi:
> I had a quick look at the attached log file and I can see some WARN and INFO loggings as well, is that intended?

@kyriathar I think that makes sense, because we effectively are logging at trace level _and above_. So while that duplicates some messages from the main log, it means you can look in the ledger log file and get a comprehensive trace of ledger activity without the noise of the other WARN / INFO logs.

relyafi:
@kyriathar actually, it looks like ledger INFO messages were not appearing in the main log file, I've just pushed an extra change to resolve. Thanks - good spot!

kyriathar (approved):

PR #3670, opened 2023-04-20
CORE-7184 AMQP custom serialization asserts for private CPK types

kyriathar:
- Adds tests to assert the below cases:
  - custom serializer targeting private CPK type fails.
  - custom serializer targeting public CPK type but whose proxy type references private CPK type fails.
  - custom serializer targeting public CPK type and uses a public CPK proxy type for the private CPK type succeeds.


josephzunigadaly (approved):

driessamyn (approved):
LGTM

PR #3671, opened 2023-04-20
CORE-10330: Remove spurious Kotlin dependency from Corda APIs.

chrisr3:
Update our Gradle scripts now that the lingering (_and out-dated!_) Kotlin dependency has been removed from our API.


wzur-r3 (approved):
LGTM

josephzunigadaly (approved):

mnesbit (approved):
OK

components/ledger/ledger-common-flow-api/build.gradle
```diff
@@ -8,8 +8,9 @@ description 'Corda ledger common - flow api'
 dependencies {
     compileOnly 'org.osgi:osgi.annotation'
 
-    implementation platform("net.corda:corda-api:$cordaApiVersion")
-    implementation 'net.corda:corda-ledger-common'
-    implementation 'net.corda:corda-crypto'
-    implementation project(':libs:ledger:ledger-common-data')
+    api platform("net.corda:corda-api:$cordaApiVersion")
```
blsemo, line 11: Not that I necessarily disagree, but why did you need to change this from `implementation` to `api`?

chrisr3: Because this bundle is providing an API, which means that everything you need to compile against this API also needs to be present on the compile classpath.

This is what `api` does. As `implementation` dependencies, they would only be included on the runtime classpath.

PR #3673, opened 2023-04-20
CORE-10426 : Sync with release

SaurabhR3:
Sync with release


ac101m (commented):
At a glance, not seeing anything that should cause issues with the flow mapper. Should note though that the interop smoke test is currently disabled, so any failures _won't_ show up on Jenkins. I will have to check the branch locally. Approval to follow.

P.S. Typo in title, should be "sync" not "synch"!

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionEventExecutor.kt
```diff
@@ -117,12 +116,17 @@ class SessionEventExecutor(
...
-                        appMessageFactory(sessionEvent, sessionEventSerializer, flowConfig)
-                    )
+                    if (!flowMapperState.isInteropSession) {
+                        Record(
+                            outputTopic,
+                            sessionEvent.sessionId,
+                            appMessageFactory(sessionEvent, sessionEventSerializer, flowConfig)
+                        )
+                    } else {
+                        Record(outputTopic, sessionEvent.sessionId, FlowMapperEvent(sessionEvent))
```
szymonsztuka, line 128: I think this is correct, only for OUTBOUND case we need to produce different record type for interop message. The newly introduced other cases `FlowMapperStateType.CLOSING`, ` FlowMapperStateType.ERROR` don't produce any normal output (message), so we don't need to produce anything for interop as well.

ac101m: > I will have to check the branch locally. Approval to follow.

Looks like there has been a breakage. I will have to take a closer look at the changes!

ac101m:
> I will have to check the branch locally. Approval to follow.

Looks like there have been some breakages, will have to investigate more closely.

ac101m (approved):
LGTM, as my issues appear to be local.

Should fix the spelling mistake in the title though, as that will end up as part of the commit history.

szymonsztuka (approved):

PR #3674, opened 2023-04-20
CORE-12788 do not throw a platform exception while already processing the result of a platform exception.

LWogan:
- Only validate that counterparty exists for sessions which are not in a terminated state
- Only update the checkpoint session state via `putSessionState()` when the checkpoint is not marked for deletion 

I think this pr highlights again that the way we handle session state updates is a bit confusing and requires refactoring. The `FlowCheckpoint` wrapper object manages the state of avro `Checkpoint`s. We deep copy the `FlowState` and assign sessionStates to a map. The `Flowcheckpoint` has a method for getting and updating the sessionStates to manage them and apply some validation to their updates.
However, The `SessionManager` updates the `SessionState`  objects in-situ. This means the flow engine can call `flowCheckpoint.getSessionState()`, perform an operation on it such as `updatedState = sessionManager.getMessagestoSend(sessionState)`, then decide not to call `flowCheckpoint.putSessionState(updatedState)`, but it will still get saved to the topic when `flowCheckpoint.toAvro()` is called. I will raise a JIRA to address this. 

As a consequence of the above point, this means the logic to verify that a couterparty exists will not be executed on all wakeups for sessions which have sent no data. This is because the getMessagesToSend() has updated the timestamps on SessionEvents based on the messageSesendWindow despite the messages not being sent (due to counterparty not being found), This may not be an issue as the counterparty validation window is set to be more than 2x larger than the resend window, but it may cause issues in the future.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowGlobalPostProcessorImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowGlobalPostProcessorImpl.kt
@@ -1,5 +1,6 @@
 package net.corda.flow.pipeline.impl
 
+import java.time.Instant
 import net.corda.data.flow.FlowKey
 import net.corda.data.flow.event.mapper.FlowMapperEvent
 import net.corda.data.flow.event.mapper.ScheduleCleanup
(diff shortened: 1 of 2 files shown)


components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowGlobalPostProcessorImpl.kt
```diff
@@ -97,11 +104,16 @@ class FlowGlobalPostProcessorImpl @Activate constructor(
             ).plusMillis(timeoutWindow)
 
             if (expiryTime < now) {
-                context.checkpoint.putSessionState(sessionManager.errorSession(sessionState))
                 val msg = "[${context.checkpoint.holdingIdentity.x500Name}] has failed to create a flow with counterparty: " +
                         "[${counterparty}] as the recipient doesn't exist in the network."
-                log.debug(msg)
-                throw FlowPlatformException(msg)
+                sessionManager.errorSession(sessionState)
```
LWogan, line 113: This code is confusing because im trying to obey the "rules" of SessionState management...

ben-millar (dismissed):
These changes make a lot of sense, good spot. I'll take a look at the `putSessionState` confusion.

ben-millar (approved):
LGTM

PR #3675, opened 2023-04-20
CORE-11704 Drop capabilities and use read-only FS

davidcurrie:
Drops all capabilities for containers. Sets root filesystem as read-only. Mounts `/tmp` and `/work` as writeable and sets `--workspace-dir` and `--temp-dir` on Corda workers). Worked out the required combination of quotes to allow us to pass in the SASL JAAS config directly as a Corda messaging argument rather than using the JVM JAAS config. This allowed the init container on the workers to be removed and ensures that the JAAS configuration is only usable by the Corda Kafka connections. Updated the worker helper so that the JAAS config is not logged (given it contains the credentials) but modified the behaviour so that at least some indication that the argument has been provided is output.

With a suitable set of overrides (e.g. to increase the number of replicas and set resource limits/requests) the YAML generated by the chart now gets a clean bill of health from Polaris.


davidcurrie:
Created separate volume mount for `--workspace-dir` at the suggestion of @mnesbit. Adding @corda/packaging to reviewers as a consequence of the changes to the worker arguments.

driessamyn (approved):
LGTM (packaging)

CornelR3 (approved):
LGTM

PR #3676, opened 2023-04-20
CORE-12685: Use the correct key for dynamic network

yift-r3:
In a dynamic network, the session keys raw PEMs are in `PARTY_SESSION_KEYS_PEM`. In a static network, they are in `PARTY_SESSION_KEYS`. This should work with both of them.


vlajos (approved):
It looks good to me.

charlieR3 (approved):
This should resolve QA's failures, but I think we should look at having the static and dynamic context's in sync with each other. I'll raise a separate ticket for that.

Edit: Raised ticket https://r3-cev.atlassian.net/browse/CORE-12852

PR #3677, opened 2023-04-21
CORE-12845: Upgrade to Corda Gradle plugins 7.0.3.

chrisr3:
Upgrade to Corda Gradle plugins 7.0.3.

Also only perform some Java-specific configuration if we apply the `java` plugin.


ronanbrowne (approved):
LGTM

PR #3679, opened 2023-04-21
CORE-12857 Allow resetting of `VaultNamedParameterizedQuery` parameters

lankydan:
Allow `VaultNamedParameterizedQuery`'s `offset`, `limit` and `parameters` to be set multiple times.

This keeps it following the same behaviour as `NamedParameterizedQuery`.

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/UtxoLedgerServiceImpl.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/UtxoLedgerServiceImpl.kt
@@ -140,7 +140,15 @@ class UtxoLedgerServiceImpl @Activate constructor(
 
     @Suspendable
     override fun <R> query(queryName: String, resultClass: Class<R>): VaultNamedParameterizedQuery<R> {
-        return VaultNamedParameterizedQueryImpl(queryName, externalEventExecutor, serializationService, resultClass)
+        return VaultNamedParameterizedQueryImpl(
+            queryName,
+            externalEventExecutor,
+            serializationService,
+            parameters = mutableMapOf(),
+            limit = Int.MAX_VALUE,
+            offset = 0,
+            resultClass
+        )
     }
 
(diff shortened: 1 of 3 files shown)


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/UtxoLedgerServiceImpl.kt
```diff
@@ -140,7 +140,15 @@ class UtxoLedgerServiceImpl @Activate constructor(
 
     @Suspendable
     override fun <R> query(queryName: String, resultClass: Class<R>): VaultNamedParameterizedQuery<R> {
-        return VaultNamedParameterizedQueryImpl(queryName, externalEventExecutor, serializationService, resultClass)
+        return VaultNamedParameterizedQueryImpl(
+            queryName,
+            externalEventExecutor,
+            serializationService,
+            parameters = mutableMapOf(),
+            limit = Int.MAX_VALUE,
```
nkovacsx, line 148: I don't necessarily agree with this, this feels a bit dangerous to me. Let's say the user doesn't provide this by accident and the result set they return contains hundreds of thousands of records. This way it could easily kill their flow right?

I think either setting this to a lower amount and displaying a warning for them to set it explicitly or keeping it as it was before would be a better approach. Silently setting this to `MAX_VALUE` could be dangerous imo.

lankydan: While I think you have a fair point, this is the behaviour we have for entity queries as well. So either we keep what I've done, or we change it there as well. Because of time I'd rather not change the other stuff.

blsemo: I think it's ok to match the other queries. Not paging at all is a valid default - I don't think that any other random default limit would be better.

blsemo (approved):
LGTM

PR #3680, opened 2023-04-21
CORE-12808: Add new metrics for batch sizes, poll timings and commit timings to state and event pattern

JamesHR3:
Add new metrics to the state and event message pattern. This should be useful to help measure flow performance, by understanding where most time is spent during a single loop of polling for events, processing those events, and writing the resulting messages back to Kafka.

The added metrics are:
- State poll timer, indicating how much time was spent in the state poll
- Event poll timer, indicating how much time was spent in the event poll
- Batch size, indicating how many events were processed by the pattern in a single batch before poll was called again
- Commit time, indicating how long it took to write all output messages back to the bus, along with the new offsets on the event topic

Additionally, some cleanup around existing metrics has been done to ensure string constants are consistent across the patterns library.


conal-smith (approved):
LGTM

PR #3683, opened 2023-04-21
CORE-9070: use corda micrometer fork

ronanbrowne:
Use our published fork of `micrometer`, which repackages this [RC 1.11.0 ](https://github.com/micrometer-metrics/micrometer/releases/tag/v1.11.0-RC1,)as 1.11.0 under `net.corda.micrometer`. 

We use this as a recent contribution to the micrometer repo By @driessamyn has added OSGi support, however, this has not been released in a GA version as yet.

This should be replaced with the real micrometre artifacts once micrometre 1.11.0 GA is released.

### DIFF
--- a/libs/metrics/build.gradle
+++ b/libs/metrics/build.gradle
@@ -15,7 +15,7 @@ dependencies {
     implementation 'org.jetbrains.kotlin:kotlin-osgi-bundle'
     implementation 'org.slf4j:slf4j-api'
 
-    api ("io.micrometer:micrometer-core:$micrometerVersion") {
+    api ("net.corda.micrometer:micrometer-core:$micrometerVersion") {
         // we don't need these in classpath, so excluding them to reduce dependencies.
         exclude group: 'org.latencyutils', module: 'LatencyUtils'
     }
--- a/applications/workers/worker-common/build.gradle
+++ b/applications/workers/worker-common/build.gradle
@@ -38,12 +38,12 @@ dependencies {
         }
     }
     implementation "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"
-    implementation ("io.micrometer:micrometer-core:0.1.0-SNAPSHOT") {
+    implementation ("net.corda.micrometer:micrometer-core:$micrometerVersion") {
         // we don't need these in classpath, so excluding them to reduce dependencies.
         exclude group: 'org.hdrhistogram', module: 'HdrHistogram'
         exclude group: 'org.latencyutils', module: 'LatencyUtils'
     }
-    implementation "io.micrometer:micrometer-registry-prometheus:0.1.0-SNAPSHOT"
+    implementation "net.corda.micrometer:micrometer-registry-prometheus:$micrometerVersion"
 
     runtimeOnly project(':libs:lifecycle:lifecycle-impl')
     runtimeOnly "org.apache.commons:commons-lang3:$commonsLangVersion"
--- a/build.gradle
+++ b/build.gradle
@@ -206,11 +206,11 @@ allprojects {
             exclusiveContent {
                 forRepository {
                     maven {
-                        url "$artifactoryContextUrl/corda-dependencies-dev"
+                        url "$artifactoryContextUrl/corda-dependencies"
                     }
                 }
                 filter {
-                    includeGroup 'io.micrometer'
+                    includeGroup 'net.corda.micrometer'
                 }
             }
         }
--- a/build.gradle
+++ b/build.gradle
@@ -201,8 +201,8 @@ allprojects {
                 }
             }
 
-            // NOTE: this needs to be removed once Micrometer contains OSGi metadata
-            //   https://github.com/micrometer-metrics/micrometer/pull/3457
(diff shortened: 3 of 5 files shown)


driessamyn:
Thanks for doing this, @ronanbrowne 

I think this can be removed now too: https://github.com/corda/corda-runtime-os/blob/dcaf4a3901e04a7066a17b80e4b27289147375ff/build.gradle#L204

ronanbrowne:
> Thanks for doing this, @ronanbrowne
> 
> I think this can be removed now too:
> 
> https://github.com/corda/corda-runtime-os/blob/dcaf4a3901e04a7066a17b80e4b27289147375ff/build.gradle#L204


The next lines declare a Maven repo entry for "corda-dependencies"  ( i'v updated this from corda-dependecies-dev to corda-dependecies for this non-snapshot versions location) 

perhaps i am misunderstanding, but In this context the comment is still valid as once the "real" 1.11.0 is out we can revert to that and remove this?

edit: although i can update the comment to more accurately reflect current reality

libs/metrics/build.gradle
```diff
@@ -15,7 +15,7 @@ dependencies {
     implementation 'org.jetbrains.kotlin:kotlin-osgi-bundle'
     implementation 'org.slf4j:slf4j-api'
 
-    api ("io.micrometer:micrometer-core:$micrometerVersion") {
+    api ("net.corda.micrometer:micrometer-core:$micrometerVersion") {
```
yarivus, line 18: why not have the `net.corda.micrometer` as a gradle property ?

yarivus (changes requested):

driessamyn:
> > Thanks for doing this, @ronanbrowne
> > I think this can be removed now too:
> > https://github.com/corda/corda-runtime-os/blob/dcaf4a3901e04a7066a17b80e4b27289147375ff/build.gradle#L204
> 
> The next lines declare a Maven repo entry for "corda-dependencies" ( i'v updated this from corda-dependecies-dev to corda-dependecies for this non-snapshot versions location)
> 
> perhaps i am misunderstanding, but In this context the comment is still valid as once the "real" 1.11.0 is out we can revert to that and remove this?
> 
> edit: although i can update the comment to more accurately reflect current reality

Yes, you're right, but I think this was only needed since it was consuming from the `-dev` repo. I think it should now be able to resolve this without the `exclusiveContent` block.
It doesn't harm to have it there though.

libs/metrics/build.gradle
ronanbrowne: @yarivus While that's certainly possible, this will change once more and then our need for this is over, so it's an extra property to be maintained with minimal benefit - It feels like we are creating more code unnecessarily to externalize this. unlike something like a version, we will maintain long-term and want in a centralized place.

yarivus (approved):
LGTM

ronanbrowne:
> > > Thanks for doing this, @ronanbrowne
> > > I think this can be removed now too:
> > > https://github.com/corda/corda-runtime-os/blob/dcaf4a3901e04a7066a17b80e4b27289147375ff/build.gradle#L204
> > 
> > 
> > The next lines declare a Maven repo entry for "corda-dependencies" ( i'v updated this from corda-dependecies-dev to corda-dependecies for this non-snapshot versions location)
> > perhaps i am misunderstanding, but In this context the comment is still valid as once the "real" 1.11.0 is out we can revert to that and remove this?
> > edit: although i can update the comment to more accurately reflect current reality
> 
> Yes, you're right, but I think this was only needed since it was consuming from the `-dev` repo. I think it should now be able to resolve this without the `exclusiveContent` block. It doesn't harm to have it there though.

Removing the block breaks compilation as we don't have a repository block declared further up for corda-dependencies (only corda-dpendecies-dev)  i could add this as a normal dependency block but I've left this as to minimize dependencies we are unknowingly pulling in from this repo, I've updated comments however for clarity

driessamyn:
> > > > Thanks for doing this, @ronanbrowne
> > > > I think this can be removed now too:
> > > > https://github.com/corda/corda-runtime-os/blob/dcaf4a3901e04a7066a17b80e4b27289147375ff/build.gradle#L204
> > > 
> > > 
> > > The next lines declare a Maven repo entry for "corda-dependencies" ( i'v updated this from corda-dependecies-dev to corda-dependecies for this non-snapshot versions location)
> > > perhaps i am misunderstanding, but In this context the comment is still valid as once the "real" 1.11.0 is out we can revert to that and remove this?
> > > edit: although i can update the comment to more accurately reflect current reality
> > 
> > 
> > Yes, you're right, but I think this was only needed since it was consuming from the `-dev` repo. I think it should now be able to resolve this without the `exclusiveContent` block. It doesn't harm to have it there though.
> 
> Removing the block breaks compilation as we don't have a repository block declared further up for corda-dependencies (only corda-dpendecies-dev) i could add this as a normal dependency block but I've left this as to minimize dependencies we are unknowingly pulling in from this repo, I've updated comments however for clarity

Makes sense. I thought we already had it.
Thanks

driessamyn (approved):
LGTM

PR #3684, opened 2023-04-21
CORE-12866 Fix vault named queries join

lankydan:
Join on `transaction_id` and `leaf_idx` for the vault named query in `VaultNamedQueryExecutorImpl` because they will return unique rows between the columns.

Note, it isn’t actually unique because of the group_idx column; however, we include a WHERE group_idx in (3, 8) which circumvents this anyway.


vlajos (approved):

MrMatthewLayton (approved):

blsemo (approved):
LGTM

PR #3685, opened 2023-04-21
CORE-12867 `LazyStateAndRefImpl` serialization errors

lankydan:
Include `LazyStateAndRefImplSerializer` in the persistence sandbox.

Add `LazyStateAndRefImplKryoSerializer` to serialize `LazyStateAndRefImpl` within flows. Without this an error is thrown when trying to serialize its `lazy` field.

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/state/serializer/kryo/LazyStateAndRefImplKryoSerializer.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/state/serializer/kryo/LazyStateAndRefImplKryoSerializer.kt
@@ -1,0 +1,60 @@
+package net.corda.ledger.utxo.flow.impl.state.serializer.kryo
+
+import net.corda.ledger.utxo.data.state.LazyStateAndRefImpl
+import net.corda.ledger.utxo.data.transaction.UtxoTransactionOutputDto
+import net.corda.sandbox.type.SandboxConstants.CORDA_UNINJECTABLE_SERVICE
+import net.corda.sandbox.type.UsedByFlow
+import net.corda.serialization.checkpoint.CheckpointInput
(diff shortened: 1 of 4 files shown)


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/state/serializer/kryo/LazyStateAndRefImplKryoSerializer.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    override val type: Class<LazyStateAndRefImpl<*>> get() = LazyStateAndRefImpl::class.java
+
+    override fun write(output: CheckpointOutput, obj: LazyStateAndRefImpl<*>) {
+        output.writeClassAndObject(obj.serializedStateAndRef)
+        val delegate = obj::stateAndRef
+            .also { property -> property.isAccessible = true }
+            .getDelegate()
+        @Suppress("UNCHECKED_CAST")
+        if ((delegate as Lazy<StateAndRef<*>>).isInitialized()) {
```
vlajos, line 41: Could not ` || deserializedStateAndRef !=null` be added here too?

vlajos (approved):
I mentioned an optimization idea.
Otherwise it looks good to me.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/state/serializer/kryo/LazyStateAndRefImplKryoSerializer.kt
lankydan: I think I can actually make `deserializedStateAndRef` private because I didn't end up using it in the end. Would rather do that in a follow up PR though if possible.

blsemo (approved):
LGTM

PR #3686, opened 2023-04-21
CORE-11154 Remove `PublicKeyHash`

kyriathar:
- removes `PublicKeyHash` API and replaces its use sites with `SecureHash`
- `PublicKeyHash.id` uses (i.e. short public key id)  are replaced by `ShortHash`


libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/PublicKeyHashUtils.kt
```diff
@@ -12,4 +13,16 @@ fun ByteArray.sha256Bytes(): ByteArray = messageDigestSha256().digest(this)
 
 fun PublicKey.sha256Bytes(): ByteArray = messageDigestSha256().digest(encoded)
 
-fun PublicKey.toStringShort(): String = "DL" + EncodingUtils.toBase58(sha256Bytes())
\ No newline at end of file
+fun PublicKey.toStringShort(): String = "DL" + EncodingUtils.toBase58(sha256Bytes())
```
mnesbit: If nothing uses this I would be keen to kill of base58, which is a C4 legacy, especially as we use different type of short id everywhere else!

kyriathar: It was only used in one logging place. Replaced that with short key id.

mnesbit (approved):
OK

charlieR3 (approved):

driessamyn (approved):
LGTM

dickon (commented):
Having looked at the changes, I think this is making the code worse. The size of the PR makes it clear that hashing public keys is something we do a lot, and mostly this is making the code that used to use PublicKeyHash harder to maintain.

I will talk to @kyriathar since this surprised me and originally I did think it was a good idea.

dickon (changes requested):
Need to review if this is really helping us; we're removing a  utiltiy  class and spreadhing the knowledge around instead, and that may not be best.

dickon:
I will discuss my concern on #corda-platform-crypto

dickon (approved):
Discussed on slack today on [#corda-platformI](https://r3holdco.slack.com/archives/C04FKKK1ANT/p1682347573963409) 

To summarwsie, ideally, we'd be using platform digest service, and allowing the Corda operator configure all the digest types used. That's not possible right now. So, consensus is the single extension method to `PublicKey` is therefore a cleaner way to handle the hard coding of SHA256.

It is unfortunate the number of test cases that specify SHA-256 with this change. Test quality could be improved with a shared test helper that knows we use SHA-256. PublicKeyHash had that secondary purpose, but the loss of that needn't block merging this.

PR #3688, opened 2023-04-24
CORE-11366 Subsequent SessionInit messages now get passed through to retry

Cap-j:
Outbound SessionInit messages are no longer ignored if the first SessionInit message received no ACK.


libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionInitExecutor.kt
```diff
@@ -37,7 +37,7 @@ class SessionInitExecutor(
         } else {
             //duplicate
             log.debug { "Duplicate SessionInit event received. Key: $eventKey, Event: $sessionEvent" }
-            FlowMapperResult(flowMapperState, emptyList())
+            FlowMapperResult(flowMapperState, listOf(Record(outputTopic, sessionEvent, sessionInit)))
```
LWogan: This will work for OUTBOUND messages but will not work for INBOUND messages. 

There is some logic in the getSessionInitOutputs() that you need to preserve for both INBOUND and OUTBOUND messages.

The SessionInitExecutor handles INBOUND SessionInit events received from the other party to be passed to the Flow Engine next. 
As well as handling OUTBOUND SessionInit events sent from the Flow Engine to the counter party.

For SessionInit events sent OUTBOUND, the Flow Engine will send the message with the Record key set to the SessionID.  This is the same Record key that the mapper should use to send the message to the P2P topic (p2p.out). And then P2P will send it to the other party, and they will receive it on p2p.in and then its passed to their Flow Mapper (via the flow.mapper.event topic). This is why you don't need to change the record key for OUTBOUND. We always send messages to the other side with the SessionID as the key.

But when the other party receives the message in the Flow Mapper, it will now to be marked as INBOUND. This is done in the FlowP2PFilterProcessor which reads from p2p.in topic and passes the message to the flow mapper topic. The message will still have the key set to SessionID. If it was a duplicate message (detected by the FlowMapper state not being null), this code change would allow the duplicate message to be sent to the flow engine. The duplicate SessionInit message received would end up creating a new flow with the key set the sessionId. The mapper  should changes INBOUND message key to be a flowId, instead of SessionId, and keep that flowId saved to the FlowMapper state to be able to forward future messages (such as SessionDatas) to the correct key (flowId) for the flow engine to process. 

I think you need an extra condition in the else block, to continue to drop duplicate INBOUND messages and to only allow duplicate OUTBOUND messages through. The reason why we don't care about duplicate INBOUNDS is because the messages at this point, are on kafka topics which are reliable. We know they wont get lost and so we are sure the Flow Engine has already received it. But for OUTBOUND messages they go through the P2P layer which sends messages across the internet (if the other person is on a different kafka cluster, if they're on the same cluster it just forwards it directly to the p2p.in topic).

LWogan: Separately there is an additional line of code you'll need to execute for duplicate SessionInits for the OUTBOUND use case which i forgot to tell ye about

`sessionInit.flowId = null`

The Flow Engine sets this field on the SessionInit OUTBOUND messages to tell the Flow Mapper what flowId to forward the messages received to when the other side responds.  But we do not want to send our flowId to the other side as it is considered private data. The other side should only know about the SessionId used to communicate on, not the flowId. So we set it to null while saving it to the FlowMapper state. For duplicate messages OUTBOUND weve already saved it, so just need to set it to null.

This is a little different to the SessionInit INBOUND case where the Flow Engine doesnt have a flowId yet as the flow has not been run yet, so the flow mapper generates the FlowId for the flow engine to receive messages on.

LWogan (commented):
Not far off. Just need to make a change to continue to ignore duplicate INBOUND messages. Also a small change to set SessionInit.flowId = null for duplicate OUTBOUND messages,

libs/flows/flow-mapper-impl/src/test/kotlin/net/corda/flow/mapper/impl/executor/SessionInitExecutorTest.kt
```diff
@@ -100,4 +100,37 @@ class SessionInitExecutorTest {
...
+        val retrySessionInit = SessionInit("info", "1", emptyKeyValuePairList(), emptyKeyValuePairList(), emptyKeyValuePairList(), null)
+
+        val payloadOutbound = buildSessionEvent(MessageDirection.OUTBOUND, "sessionId-INITIATED", 1, retrySessionInit)
+        val payloadInbound = buildSessionEvent(MessageDirection.INBOUND, "sessionId-INITIATED", 2, retrySessionInit)
+
+        val flowMapperState = FlowMapperState()
+        flowMapperState.setStatus(FlowMapperStateType.OPEN)
+
+        val resultOutbound = SessionInitExecutor(
+            "sessionId-INITIATED",
```
LWogan: The format of the sessionIds don't matter too much in the context of this test but just so you know:
- the **initiating** party's sessionId will have the format '{sessionId}'
- the initiated party's sessionId will have the format '{sessionId}-INITIATED'

The only party which will send an OUTBOUND SessionInit is the **Initiating** party. So in keeping with the rest of the test this would be set to "sessionId" here.

libs/flows/flow-mapper-impl/src/test/kotlin/net/corda/flow/mapper/impl/executor/SessionInitExecutorTest.kt
```diff
@@ -100,4 +100,37 @@ class SessionInitExecutorTest {
         assertThat(state).isNotNull
         assertThat(outboundEvents).isEmpty()
     }
+    @Test
+    fun `Subsequent SessionInit messages get passed through if no ACK received from first message`(){
```
LWogan: This test is doing 2 separate things. We have an inbound test and and outbound test. Lets separate them out into two different unit tests.

libs/flows/flow-mapper-impl/src/test/kotlin/net/corda/flow/mapper/impl/executor/SessionInitExecutorTest.kt
```diff
@@ -100,4 +100,37 @@ class SessionInitExecutorTest {
         assertThat(state).isNotNull
         assertThat(outboundEvents).isEmpty()
     }
+    @Test
```
LWogan, line 104: missing a new line between lines 102 and 103

LWogan (commented):
These changes look correct to me. Lets add a new integration test to FlowMapperIntegrationTest which sends 2 session inits within a single test to validate that when 2 SessionInits are sent OUTBOUND in succession that the second message results in the SessionInit being resent as part of the output events.

Cap-j:
Decided not to add integration tests for this change

LWogan (approved):
LGTM

PR #3689, opened 2023-04-24
CORE-12364 cleanup nonexistent state

emilybowe:
When a flow is started by the REST  api a start flow event is generated by the FlowRestResourceImpl  and it is passed to the Flow Mapper with the key set to the FlowKey (converted to string). This allows the flow mapper to map the client requested flow to a FlowId. When a flow fails or finishes  the flow engine will schedule a cleanup event for this key on the mapper.

For flows which are triggered by session initiation (from receiving  a SessionInit) then this key will not exist. Therefore we should not send this cleanup event in FlowFinishedRequestHandler and FlowFailedRequestHandler when the flow was started in this way.


components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/FlowFailedRequestHandler.kt
```diff
@@ -68,10 +67,9 @@ class FlowFailedRequestHandler @Activate constructor(
             request.exception.message ?: request.exception.javaClass.name
         )
         val flowCleanupTime = context.config.getLong(PROCESSING_FLOW_CLEANUP_TIME)
-        val expiryTime = Instant.now().plusMillis(flowCleanupTime).toEpochMilli()
+        Instant.now().plusMillis(flowCleanupTime).toEpochMilli()
```
ben-millar: I'm guessing you intended to remove this entire line (The `java.time.Instant` import can also be removed in that case)

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/FlowFinishedRequestHandler.kt
```diff
@@ -45,10 +44,9 @@ class FlowFinishedRequestHandler @Activate constructor(
         val status = flowMessageFactory.createFlowCompleteStatusMessage(checkpoint, request.result)
 
         val flowCleanupTime = context.config.getLong(PROCESSING_FLOW_CLEANUP_TIME)
-        val expiryTime = Instant.now().plusMillis(flowCleanupTime).toEpochMilli()
+        Instant.now().plusMillis(flowCleanupTime).toEpochMilli()
```
ben-millar: Same thing here. Also, we probably don't need to pull the `flowCleanupTime` from the configs if it's not being used.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/FlowFailedRequestHandler.kt
emilybowe: closing as now this will stay but replaced with if/else block

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/FlowFinishedRequestHandler.kt
emilybowe: closing as now this will stay but replaced with if/else block

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/FlowFailedRequestHandler.kt
```diff
@@ -67,12 +68,25 @@ class FlowFailedRequestHandler @Activate constructor(
...
-        val records = listOf(
-            flowRecordFactory.createFlowStatusRecord(status),
-            flowRecordFactory.createFlowMapperEventRecord(checkpoint.flowKey.toString(), ScheduleCleanup(expiryTime))
-        )
+
+        //When a flow is started by the REST api, a FlowKey is sent to the Flow mapper, so we send a cleanup event for this key.
+        //Flows triggered by SessionInit don't have this FlowKey, so we do not send a cleanup event.
+        val records = if (checkpoint.flowStartContext.initiatorType == FlowInitiatorType.RPC) {
+            val flowCleanupTime = context.config.getLong(PROCESSING_FLOW_CLEANUP_TIME)
+            Instant.now().plusMillis(flowCleanupTime).toEpochMilli()
```
mbrkic-r3: This seems to be left from the previous commit and can be deleted now

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/FlowFinishedRequestHandler.kt
```diff
@@ -39,18 +40,23 @@ class FlowFinishedRequestHandler @Activate constructor(
...
-            flowRecordFactory.createFlowStatusRecord(status),
-            flowRecordFactory.createFlowMapperEventRecord(checkpoint.flowKey.toString(), ScheduleCleanup(expiryTime))
-        )
-
+        //When a flow is started by the REST api, a FlowKey is sent to the Flow mapper, so we send a cleanup event for this key.
+        //Flows triggered by SessionInit don't have this FlowKey, so we do not send a cleanup event.
+        val records = if (checkpoint.flowStartContext.initiatorType == FlowInitiatorType.RPC) {
+            val flowCleanupTime = context.config.getLong(PROCESSING_FLOW_CLEANUP_TIME)
+            val expiryTime = Instant.now().plusMillis(flowCleanupTime).toEpochMilli()
+            Instant.now().plusMillis(flowCleanupTime).toEpochMilli()
```
mbrkic-r3: This seems to be left from the previous commit and can be deleted now

ben-millar (dismissed):
The code changes make sense to me. 

In the acceptance tests, I would expect that we look for Session Cleanup events in all cases **except** for when we're a flow that has been initiated over RPC, but it's not clear to me if that's what the logic says. It seems like we're removing that expectation in **all** cases.

If you're happy that the logic is right and the tests pass, LGTM!

emilybowe:
> The code changes make sense to me.
> 
> In the acceptance tests, I would expect that we look for Session Cleanup events in all cases **except** for when we're a flow that has been initiated over RPC, but it's not clear to me if that's what the logic says. It seems like we're removing that expectation in **all** cases.
> 
> If you're happy that the logic is right and the tests pass, LGTM!

@ben-millar thanks - yeah that makes sense to me
I think for the tests that are startFlowEventReceived we expect a cleanup event, for the sessionInitEventReceived we don't expect a cleanup event.

LWogan:
>   for the sessionInitEventReceived we don't expect a cleanup event.

Its a little more subtle than this. An initiated session (via P2P) will have a flow mapper state that needs cleaning up too, as there will be a state associated with the key sessionId. The code to cleanup that state already exists in FlowGlobalPostProcessorImpl.getFlowMapperSessionCleanupEvents. So before this change there would have been 2 cleanup events scheduled. One for checkpoint.flowKey (invalid) and one for sessionState.sessionId (valid). After this change the acceptance test for sessionInitEventReceived  should only expect 1 cleanup event and it to have the key set to sessionId.

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/FlowFinishedAcceptanceTest.kt
```diff
@@ -98,7 +98,6 @@ class FlowFinishedAcceptanceTest : FlowServiceTestBase() {
             expectOutputForFlow(FLOW_ID1) {
                 nullStateRecord()
                 flowStatus(FlowStates.COMPLETED, result = DONE)
-                scheduleFlowMapperCleanupEvents(FlowKey(INITIATED_SESSION_ID_1, BOB_HOLDING_IDENTITY).toString())
```
LWogan, line 101: I think there should be a cleanup event for this one for the sessionId. It was missing in the test before your changes too, so lets add it in now. Its generated in the FlowGlobalPostProcessor.

scheduleFlowMapperCleanupEvents(INITIATED_SESSION_ID_1)

The removal of the cleanup for FlowKey(INITIATED_SESSION_ID_1, BOB_HOLDING_IDENTITY).toString() is valid though!

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/helper/FlowRecords.kt
```diff
@@ -0,0 +1,39 @@
...
+import net.corda.messaging.api.records.Record
+import net.corda.schema.configuration.FlowConfig
+import org.apache.avro.specific.SpecificRecordBase
+import java.time.Instant
+
+    /**
+     * When a flow is started by the REST api, a FlowKey is sent to the Flow mapper, so we send a cleanup event for this key.
+     *
+     * Flows triggered by SessionInit don't have this FlowKey, so we do not send a cleanup event
+     */
```
LWogan: The indentation for this is a bit off after removing the outer class

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/helper/FlowRecords.kt
```diff
@@ -0,0 +1,39 @@
...
+    fun getRecords(
+        flowRecordFactory: FlowRecordFactory,
+        context: FlowEventContext<Any>,
+        status: FlowStatus
+    ): List<Record<out Any, out SpecificRecordBase>> {
```
LWogan: The return type of this is a little funky. When we have multiple different types of records with different key and value types we usually define them as List<Record<*, *>>

LWogan (commented):
These changes look correct! a few minor comments.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/helper/FlowRecords.kt
ben-millar: I think it's probably appropriate to use `List<Record<*,*>>` here, but this did prompt me to do some reading which was pretty interesting.

`Any` is determined at compile-time, which allows us to access methods and properties of the object at compile-time, whereas `*` is determined using runtime reflection which is maybe less type safe (If it was a parameter this would matter, but it makes no difference as a returned value).

The `out` modifier makes the return types covariant, meaning that we can now substitute in any subclasses of the listed types.

Maybe every day is a learning day, as @thiagoviana says 😁

thiagoviana: Interesting. Adding something: in a list the * is not just "any" type, but type (a fixed one), which you don't know what is.

- List<*> can contain any type of objects as long as the are all of the same type.
- List<Any> can contain any object, all in the same list.

https://kotlinlang.org/docs/generics.html#star-projections

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/FlowFinishedAcceptanceTest.kt
emilybowe: Ah, right -
so I did notice this after our call when I was finishing the integration tests and I noticed the test failing on the osgi debugger asserting it was an empty list. 
I checked the release and noticed we never checked for it before either (but we do in the failed acceptance test). I thought maybe there was a difference in how we closed between a flow finish close and a flow failed close.
I checked with @JamesHR3 and after a lot of digging it seems that because in the acceptance tests the fiber and quasar are mocked in a manner of speaking, those sessions are left alive and not cleaned up. There's a cleanup event in the flow failed because this:

```
        try {
            val sessionIds = checkpoint.sessions.map { it.sessionId }
            val sessionToError = sessionIds - flowSessionManager.getSessionsWithStatuses(
                checkpoint,
                sessionIds,
                setOf(SessionStateType.ERROR, SessionStateType.CLOSED)
            ).map { it.sessionId }.toSet()

            flowSessionManager.sendErrorMessages(
                checkpoint,
                sessionToError,
                request.exception,
                Instant.now()
            )
        } catch (e: FlowSessionStateException) {
            throw FlowFatalException(e.message, e)
        }
```

sets the state to ERROR. However in flow finished in order to get to CLOSE quite a but more would be needed. 
For next steps James said we shouldn't be checking in flow finished for the cleanup it should already be in sub flow finished instead.

LWogan: I see now that the test itself is skipping lots of steps in the session protocol. I agree the test should not expect any scheduled cleanups.

I had assumed it was sending and receiving the expected session protocol messages that are required to reach the FlowFinished result,  similar to the test following it in that class. 

This test says we receive a SessionInit and then suspends with FlowFinished. This isn't possible in the real flow engine as I THINK you have to have closed all your active sessions to be able to return with FlowFinished. So yeah the SessionState in this case is still set to CONFIRMED when the flow suspends with FlowFinished. This means the global post processor wont schedule any cleanups for the session.
There might also be more to it too that im missing. I'd have to dig deeper into the acceptance test framework to refresh myself

LWogan (approved):
LGTM

PR #3691, opened 2023-04-24
CORE-12880 merkle proof execution time

dickon:
Reduce test time from 22 seconds to 0.5 seconds by:
- Using a cheap and non-cryptographic hash function for Merkle tree tests.
- Test only up to 11 leafs rather than 15, since it gets slow due to the exponential growth of the test time. 

Both changes have about the same contribution to speeding up the tests.

If anyone can demonstrate any regressions caught by testing up to 15 but not at 11 then maybe we should move the larger tree tests to an e2e test run periodically on the release branch. But short of a demonstratble problem it feels like a waste of electricity.

### DIFF
--- a/libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
+++ b/libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
@@ -15,7 +15,6 @@ import net.corda.v5.crypto.SecureHash
 import net.corda.v5.crypto.extensions.merkle.MerkleTreeHashDigestProvider
 import net.corda.v5.crypto.merkle.MerkleProof
 import net.corda.v5.crypto.merkle.MerkleProofType
-import org.jetbrains.annotations.NotNull
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertFalse
 import org.junit.jupiter.api.Assertions.assertNotEquals
--- a/libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
+++ b/libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
@@ -273,6 +273,8 @@ class MerkleTreeTest {
         runMerkelProofTest(treeSize)
     }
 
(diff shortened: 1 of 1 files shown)


libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
```diff
@@ -63,18 +82,18 @@ class MerkleTreeTest {
         }
 
         @JvmStatic
-        fun merkleProofTestSizes(): List<Int> = (1 until 16).toList()
+        fun merkleProofTestSizes(): List<Int> = (1 until 12).toList()
```
mnesbit, line 86: Sadly I did have bugs that only appeared when we got to larger trees with several layers. However, I would be fine with testing 1 until 12 here and having a second disabled test  for 12 until 16 that is meant to be run if anybody messes with the merkle multiproof code

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
```diff
@@ -36,7 +37,25 @@ class MerkleTreeTest {
...
+            private val ZERO_BYTE = ByteArray(1) { 0 }
+            private val ONE_BYTE = ByteArray(1) { 1 }
+            private fun hash(b: ByteArray): SecureHash {
+                var acc = 0L
+                for (c in b) acc += c.toLong()
+                return SecureHashImpl("byteadd", acc.toByteArray())
+            }
 
+            override fun getDigestAlgorithmName(): DigestAlgorithmName = DigestAlgorithmName("add")
+            override fun leafNonce(index: Int): ByteArray? = null
```
mnesbit, line 51: This won't exercise the nonce code paths, which may be fine so long as that is covered in at least some tests

dickon: That was my thinking, yes.

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
dickon: @mnesbit  done

driessamyn (approved):
LGTM

PR #3692, opened 2023-04-24
CORE-12881: crypto ops client test share fixtures

dickon:
Allow retries to be configurable for CryptoOpsClientComponent, but not fixed.

The motivation for this is simply to significantly decreate test execution time, saving about 5 seconds on my machine, i.e. double the speed of the test. The absolute time saving should be more on Jenkins. See https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4449697918/Unit+Test+Execution+Time+findings

On my machine, CryptoOpsClientComponentTests takes 7.6s now. That seems to be mostly test set up. Baseline 5.1 tests take 39.7s.

### DIFF
--- a/components/crypto/crypto-client-impl/src/main/kotlin/net/corda/crypto/client/impl/CryptoOpsClientComponent.kt
+++ b/components/crypto/crypto-client-impl/src/main/kotlin/net/corda/crypto/client/impl/CryptoOpsClientComponent.kt
@@ -62,13 +62,14 @@ class CryptoOpsClientComponent @Activate constructor(
     ),
     configKeys = setOf(MESSAGING_CONFIG, CRYPTO_CONFIG)
 ), CryptoOpsClient, CryptoOpsProxyClient {
+    @Suppress("LongParameterList")
     constructor(
         coordinatorFactory: LifecycleCoordinatorFactory,
         publisherFactory: PublisherFactory,
         schemeMetadata: CipherSchemeMetadata,
         configurationReadService: ConfigurationReadService,
         digestService: PlatformDigestService,
-        retriesLimit: Int,
+        retriesLimit: Int = 3,
     ) : this(
         coordinatorFactory,
         publisherFactory,
--- a/components/crypto/crypto-client-impl/src/main/kotlin/net/corda/crypto/client/impl/CryptoOpsClientComponent.kt
+++ b/components/crypto/crypto-client-impl/src/main/kotlin/net/corda/crypto/client/impl/CryptoOpsClientComponent.kt
@@ -50,7 +50,8 @@ class CryptoOpsClientComponent @Activate constructor(
     @Reference(service = ConfigurationReadService::class)
     configurationReadService: ConfigurationReadService,
     @Reference(service = PlatformDigestService::class)
-    private val digestService: PlatformDigestService
+    private val digestService: PlatformDigestService,
+    private val rpcRetries: Int = 3,
 ) : AbstractConfigurableComponent<CryptoOpsClientComponent.Impl>(
     coordinatorFactory = coordinatorFactory,
     myName = LifecycleCoordinatorName.forComponent<CryptoOpsClient>(),
@@ -68,7 +69,7 @@ class CryptoOpsClientComponent @Activate constructor(
     }
 
     override fun createActiveImpl(event: ConfigChangedEvent): Impl =
-        Impl(publisherFactory, schemeMetadata, digestService, event)
+        Impl(publisherFactory, schemeMetadata, digestService, event, rpcRetries)
 
     override fun getSupportedSchemes(tenantId: String, category: String): List<String> =
         impl.ops.getSupportedSchemes(tenantId, category)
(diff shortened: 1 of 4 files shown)


components/crypto/crypto-client-impl/src/test/kotlin/net/corda/crypto/client/impl/CryptoOpsClientComponentTests.kt
```diff
@@ -79,59 +77,53 @@ import kotlin.test.assertFalse
 import kotlin.test.assertNotSame
 import kotlin.test.assertTrue
 
+val schemeMetadata = CipherSchemeMetadataImpl()
```
fowlerrr: Would this be better as a static on the test class?

components/crypto/crypto-client-impl/src/test/kotlin/net/corda/crypto/client/impl/CryptoOpsClientComponentTests.kt
```diff
@@ -79,59 +77,53 @@ import kotlin.test.assertFalse
...
-    private lateinit var configurationReadService: TestConfigurationReadService
-    private lateinit var publisherFactory: PublisherFactory
-    private lateinit var component: CryptoOpsClientComponent
-
-    @BeforeEach
-    fun setup() {
-        knownTenantId = toHex(UUID.randomUUID().toString().toByteArray().sha256Bytes()).take(12)
-        knownAlias = UUID.randomUUID().toString()
-        knownOperationContext = mapOf(
+        private val knownTenantId = toHex(UUID.randomUUID().toString().toByteArray().sha256Bytes()).take(12)
```
fowlerrr: I don't know how expensive it is in real terms but could we factor this out to avoid a `sha256Bytes` for each test?  Or do we need those tenant Ids to be unique per test?

components/crypto/crypto-client-impl/src/main/kotlin/net/corda/crypto/client/impl/CryptoOpsClientComponent.kt
```diff
@@ -62,13 +62,31 @@ class CryptoOpsClientComponent @Activate constructor(
     ),
     configKeys = setOf(MESSAGING_CONFIG, CRYPTO_CONFIG)
 ), CryptoOpsClient, CryptoOpsProxyClient {
+    constructor(
+        coordinatorFactory: LifecycleCoordinatorFactory,
+        publisherFactory: PublisherFactory,
+        schemeMetadata: CipherSchemeMetadata,
+        configurationReadService: ConfigurationReadService,
+        digestService: PlatformDigestService,
+        retriesLimit: Int,
```
fowlerrr: I think you could simply set the default `= 3` here, right?

components/crypto/crypto-client-impl/src/main/kotlin/net/corda/crypto/client/impl/CryptoOpsClientComponent.kt
```diff
@@ -194,7 +212,8 @@ class CryptoOpsClientComponent @Activate constructor(
         publisherFactory: PublisherFactory,
         schemeMetadata: CipherSchemeMetadata,
         digestService: PlatformDigestService,
-        event: ConfigChangedEvent
+        event: ConfigChangedEvent,
+        retries: Int = 3,
```
fowlerrr, line 217: And, arguably, get rid of the default here.  Though... meh

fowlerrr (commented):
Tiny things...

components/crypto/crypto-client-impl/src/main/kotlin/net/corda/crypto/client/impl/CryptoOpsClientComponent.kt
dickon: That would make it amibiguous with the primary constructor.

components/crypto/crypto-client-impl/src/test/kotlin/net/corda/crypto/client/impl/CryptoOpsClientComponentTests.kt
dickon: I think it'd be more standard for Corda and possibly for Kotlin to do so, so I've made the change.

(my instincts are more to have declarations be at as high a level as possible, and mostly immutable but if I do that my code will be harded to read for people used to Java industry standard OOP)

components/crypto/crypto-client-impl/src/main/kotlin/net/corda/crypto/client/impl/CryptoOpsClientComponent.kt
dickon: but that's not necessariy a problem, change made.

components/crypto/crypto-client-impl/src/test/kotlin/net/corda/crypto/client/impl/CryptoOpsClientComponentTests.kt
dickon: This is static, so I wouldn't expect it to get repeated for each test? 

Maybe I could hard code a speciifc sha256-length bit pattern here.

components/crypto/crypto-client-impl/src/main/kotlin/net/corda/crypto/client/impl/CryptoOpsClientComponent.kt
dickon: I don't really understand what the concern is for reasonable defaults like this.

fowlerrr (approved):

PR #3693, opened 2023-04-24
CORE-12579 Update notary validations for re-registration

YashNabar:
Updates validations against existing notary services (for enforcing a single notary virtual node under each notary service, and for ensuring that a registering member's name is not already in use as a notary service name) to work with re-registration. Also adds checks while approving members and adding notary to group parameters to prevent race conditions between manual and automatic registration approvals.
Enables notary re-registration e2e test in https://github.com/corda/corda-e2e-tests/pull/83.


components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
```diff
@@ -78,6 +78,15 @@ internal class AddNotaryToGroupParametersHandler(
...
             val (epoch, groupParameters) = if (notaryServiceNumber != null) {
+                // Enforces a single virtual node under a notary service, until support for multiple virtual nodes per
+                // notary service is added.
+                // There is also a check in StartRegistrationHandler, this additional check prevents race
+                // conditions between manual and automatic registration approvals.
+                require(false) {
+                    throw MembershipPersistenceException(
+                        "Cannot add notary to group parameters - notary service '$notaryServiceName' already exists."
+                    )
+                }
```
charlieR3:
```suggestion
                    throw MembershipPersistenceException(
                        "Cannot add notary to group parameters - notary service '$notaryServiceName' already exists."
                    )
```

components/membership/membership-persistence-service-impl/src/test/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandlerTest.kt
```diff
@@ -281,6 +282,7 @@ class AddNotaryToGroupParametersHandlerTest {
     }
 
     @Test
+    @Disabled("Until support is added for multiple notary virtual nodes per notary service")
```
charlieR3: Can we link all these disabled tests to a specific jira ticket and add a comment (or update the deiscription) so the ticket has a reminder to remove this annotation?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
```diff
@@ -259,15 +260,33 @@ internal class StartRegistrationHandler(
...
+
+    /**
+     * Check for name clashes with existing notary services, to ensure that a registering member's name is not
+     * already in use as a notary service name, and that a particular notary service may have only a single virtual
+     * node listed under it (while allowing re-registrations).
+     */
+    private fun checkAgainstExistingNotaryServices(
+        mgmHoldingId: HoldingIdentity, notary: MemberNotaryDetails?, registeringMemberName: String
+    ) {
+        membershipQueryClient.queryRegistrationRequests(mgmHoldingId, statuses = listOf(RegistrationStatus.APPROVED)).getOrThrow()
```
nikinagy: I wonder if there is a way to reduce the amount of data, I am a bit worried we will end up processing huge loads of data in a large network here.

nikinagy: Maybe we could re-use the limit to get the latest entry for each member and that would reduce the amount of data to 1 per each member.

charlieR3: Yeah, I agree. We could easily end up with a very large number of registrations so we certainly don't want to loop over them all during registrations, nor should we send them all over kafka.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
```diff
@@ -259,15 +260,33 @@ internal class StartRegistrationHandler(
...
-            membershipGroupReaderProvider.getGroupReader(mgmHoldingId).groupParameters?.let { groupParameters ->
-                validateRegistrationRequest(groupParameters.notaries.none { it.name == member.name }) {
-                    "Registering member's name '${member.name}' is already in use as a notary service name."
-                }
-                validateRegistrationRequest(groupParameters.notaries.none { it.name == notary.serviceName }) {
-                    "Notary service '${notary.serviceName}' already exists."
-                }
-            } ?: throw MembershipRegistrationException("Could not read group parameters of the membership group '${member.groupId}'.")
         }
+        checkAgainstExistingNotaryServices(mgmHoldingId, notaryDetails, member.name.toString())
```
charlieR3: I'm not sure if `validateRoleInformation` if the correct place for this. It's validating the name, right? Not the role?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
YashNabar: Revised in https://github.com/corda/corda-runtime-os/pull/3693/commits/833a2db0a83d240cbcab813cd0abc11f5bda362f

buildSrc/build.gradle
```diff
@@ -37,6 +37,11 @@ repositories {
                 username = findProperty('cordaArtifactoryUsername') ?: System.getenv('CORDA_ARTIFACTORY_USERNAME')
                 password = findProperty('cordaArtifactoryPassword') ?: System.getenv('CORDA_ARTIFACTORY_PASSWORD')
             }
+            metadataSources {
```
charlieR3: Just double checking - is this a change you mentioned that needed to be cherry picked from the release branch?

charlieR3 (approved):

buildSrc/build.gradle
YashNabar: Yes that's the one (ES-160 [560dd54](https://github.com/corda/corda-runtime-os/pull/3693/commits/560dd541afef7100a2bca8aff6aa32818a680008))

charlieR3 (approved):

PR #3695, opened 2023-04-24
CORE-11488: Make the database messaging bus implementation stable during database access issues

yift-r3:
This should make the DB implementation of the message bus keep working even if the database is not accessible.

### DIFF
--- a/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/persistence/DBAccess.kt
+++ b/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/persistence/DBAccess.kt
@@ -8,13 +8,15 @@ import net.corda.messagebus.db.datamodel.TopicRecordEntry
 import net.corda.messagebus.db.datamodel.TransactionRecordEntry
 import net.corda.messagebus.db.datamodel.TransactionState
 import net.corda.messaging.api.exception.CordaMessageAPIFatalException
+import net.corda.messaging.api.exception.CordaMessageAPIIntermittentException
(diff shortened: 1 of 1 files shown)


charlieR3 (commented):
Looks good to me. Does it need some unit testing also? Best to wait for @fowlerrr 's review also since he's more familiar than I am with this part of the codebase.

libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/persistence/DBAccess.kt
```diff
@@ -380,27 +341,35 @@ class DBAccess(
...
     }
 
-    private fun <T : Exception> Exception.isCausedBy(exceptionType: Class<T>): Boolean {
+    private inline fun <reified T : Exception> Exception.isCausedBy(message: String? = null): Boolean {
         var currentCause = this.cause
         while (currentCause != null) {
-            if (exceptionType.isAssignableFrom(currentCause::class.java)) {
-                return true
+            if (currentCause is T) {
+                return message == null || currentCause.message?.contains(message) ?: false
```
fowlerrr, line 357: I don't exactly remember why I had `isAssignableFrom` but probably because I couldn't guarantee which exceptions I'd get.  And this lets you catch all transient ones.  I'm not sure getting rid of that is a good idea.  At least keep an eye on it if we see some other transient exception that we're not specifically catering for.

fowlerrr (approved):
Not sure about getting rid of `isAssignableFrom` but otherwise this is fine.  Just keep an eye on it for errors.

PR #3697, opened 2023-04-24
CORE-12218: Prep work for allowing `JsonObject` as a method result

vkolomeyko:
This work will ensure that when `JsonObject` is included into return type of REST method and if its `escapedJson` content is valid, it will be include into the output JSON payload verbatim, i.e. without escaping. Please, see Jira description why this will be important.

**Note**: `FlowRestResource` has not been modified in this PR, but will be modified in PRs to follow to take advantage of this new functionality. This separation is required as modification to `FlowRestResource` may be considered as a breaking change to OpenAPI.

This PR also includes some re-jig of Json marshalling code used in REST.

### DIFF
--- a/libs/rest/json-serialization/src/main/kotlin/net/corda/rest/json/serialization/JsonObjectMarshalling.kt
+++ b/libs/rest/json-serialization/src/main/kotlin/net/corda/rest/json/serialization/JsonObjectMarshalling.kt
@@ -1,4 +1,4 @@
-package net.corda.rest.json
+package net.corda.rest.json.serialization
 
 import com.fasterxml.jackson.core.JsonGenerator
 import com.fasterxml.jackson.core.JsonParser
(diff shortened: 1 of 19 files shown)


libs/rest/json-serialization/src/main/kotlin/net/corda/rest/json/serialization/JsonObjectMarshalling.kt
```diff
@@ -0,0 +1,45 @@
...
+        // Check if `escapedJson` is a valid JSON content
+        val actualJsonNode: JsonNode? = try {
+            mapper.readTree(obj.escapedJson)
+        } catch (ex: Exception) {
+            // Assuming not a valid Json then
+            null
+        }
+
+        if (actualJsonNode != null) {
+            generator.writeTree(actualJsonNode)
```
vkolomeyko, line 30: The "meat" of the change, which will write valid JSON content verbatim.

filipesoliveira (approved):
LGTM

PR #3702, opened 2023-04-25
CORE-12539 - Update to latest version of jackson and mockito

driessamyn:


libs/rest/rest-server-impl/src/integrationTest/kotlin/net/corda/rest/server/impl/RestServerRequestsTest.kt
```diff
@@ -360,7 +365,7 @@ class RestServerRequestsTest : RestServerTestBase() {
         val timeCallResponse = client.call(POST, WebRequest<Any>("health/timecall", """ { "time": { "time": "$time" } } """), userName, password)
 
         assertEquals(HttpStatus.SC_OK, timeCallResponse.responseStatus)
-        assertEquals("2020-01-01T11:00Z[UTC]", timeCallResponse.body)
+        assertEquals("2020-01-01T11:00Z", timeCallResponse.body)
```
driessamyn, line 368: This change is triggered by an Jackson update. See https://github.com/FasterXML/jackson-modules-java8/issues/204 and https://github.com/FasterXML/jackson-modules-java8/pull/267

I think this change is ok as the new format is ISO-8601 compatible.

@mnesbit , do you agree this change is ok?

mnesbit: yes it is fine and indeed probably more canonical

mnesbit (approved):
OK

gradle.properties
```diff
@@ -56,7 +56,7 @@ felixSecurityVersion=2.8.3
 guavaVersion=30.1.1-jre
 hibernateVersion=5.6.15.Final
 hikariCpVersion=5.0.1
-jacksonVersion=2.14.2
+jacksonVersion=2.15.0
```
vkolomeyko, line 59: Probably as a separate PR, but it would be worth upgrading SnakeYaml version to eliminate reported C1 vulnerability.

vkolomeyko (approved):

ronanbrowne (approved):
lgtm

PR #3704, opened 2023-04-25
CORE-12908: Rename :libs:flows:utils to :libs:flows:flow-utils.

chrisr3:
Gradle uses the module name as the published artifact ID, so `:libs:flows:utils` would be published uninformatively as just `utils.jar`. Rename the module to `flow-utils` so that the artifact becomes `flow-utils.jar` instead.


mcgovc (approved):
LGTM

fowlerrr (approved):
Not sure why DB approval was need but LGTM

PR #3705, opened 2023-04-25
CORE-12561: Close the server before opening a new one if the same host and port are used

yift-r3:
Close the server before opening a new one if the new one has the same port and host as the old one.

### DIFF
--- a/components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/ReconfigurableHttpServer.kt
+++ b/components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/ReconfigurableHttpServer.kt
@@ -7,7 +7,6 @@ import net.corda.lifecycle.domino.logic.ConfigurationChangeHandler
 import net.corda.lifecycle.domino.logic.LifecycleWithDominoTile
 import net.corda.lifecycle.domino.logic.util.ResourcesHolder
 import net.corda.p2p.gateway.messaging.GatewayConfiguration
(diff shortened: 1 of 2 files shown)


components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/ReconfigurableHttpServer.kt
```diff
@@ -27,14 +26,19 @@ internal class ReconfigurableHttpServer(
     private val dynamicCertificateSubjectStore: DynamicCertificateSubjectStore,
 ) : LifecycleWithDominoTile {
 
+    private data class ServerKey(
```
williamvigorr3, line 29: Is it worth adding a unit test that the first server gets closed before the second one is opened if we change the endpoint. i.e. add a test which would fail if this bug was reintroduced.

williamvigorr3 (dismissed):
It makes sense that we need to key the map by (port, host) if we want to prevent multiple concurrent servers trying to bind on the same port, host.

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/ReconfigurableHttpServer.kt
yift-r3: I've added an integration test.

williamvigorr3 (approved):
LGTM

PR #3706, opened 2023-04-25
CORE-12862 Unique notary service and MGM group ids on smoke test runs

relyafi:
Previously, the notary service name was hard-coded to a fixed string. This meant that repeated runs of the smoke tests against the same database would re-use the notary service identity. This had the unintended effect of adding more notary virtual nodes to the existing service identity, rather than provisioning a new service identity on each test run.

Aside from this currently being an unrealistic setup, composite keys (used to aggregate the keys of notary virtual nodes) have a maximum number of leaves. Repeated test runs could eventually fail due to hitting this limit. To address this, this PR generates a new UUID for the notary service on each test run.

Making this change also revealed that in many cases, the MGM group id is also reused between test runs. This led to some flows failing due to the fact that the above change would result in multiple notary services on a network, which some CorDapps cannot handle. This change therefore also makes a similar change to the MGM group id, ensuring a new id is used on each test run.

See https://github.com/corda/corda-e2e-tests/pull/74 for the corresponding E2E test change.

### DIFF
--- a/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/AmqpSerializationTests.kt
+++ b/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/AmqpSerializationTests.kt
@@ -2,7 +2,6 @@ package net.corda.applications.workers.smoketest.flow
 
 import net.corda.applications.workers.smoketest.TEST_CPB_LOCATION
 import net.corda.applications.workers.smoketest.TEST_CPI_NAME
-import net.corda.e2etest.utilities.GROUP_ID
 import net.corda.e2etest.utilities.RPC_FLOW_STATUS_SUCCESS
 import net.corda.e2etest.utilities.awaitRpcFlowFinished
 import net.corda.e2etest.utilities.conditionallyUploadCordaPackage
@@ -22,9 +21,10 @@ class AmqpSerializationTests {
         private const val AmqpSerializationTestFlow = "net.cordapp.testing.smoketests.flow.AmqpSerializationTestFlow"
 
         private val testRunUniqueId = UUID.randomUUID()
+        private val groupId = UUID.randomUUID().toString()
         private val cpiName = "${TEST_CPI_NAME}_$testRunUniqueId"
         private val bobX500 = "CN=Bob-${testRunUniqueId}, OU=Application, O=R3, L=London, C=GB"
-        private var bobHoldingId: String = getHoldingIdShortHash(bobX500, GROUP_ID)
+        private var bobHoldingId: String = getHoldingIdShortHash(bobX500, groupId)
         private val staticMemberList = listOf(
             bobX500,
         )
@@ -36,7 +36,7 @@ class AmqpSerializationTests {
             conditionallyUploadCordaPackage(
                 cpiName,
                 TEST_CPB_LOCATION,
-                GROUP_ID,
+                groupId,
                 staticMemberList
             )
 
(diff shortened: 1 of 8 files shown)


applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/AmqpSerializationTests.kt
```diff
@@ -22,9 +21,10 @@ class AmqpSerializationTests {
         private const val AmqpSerializationTestFlow = "net.cordapp.testing.smoketests.flow.AmqpSerializationTestFlow"
 
         private val testRunUniqueId = UUID.randomUUID()
+        private val groupId = UUID.randomUUID().toString()
```
vlajos, line 24: Was there any reason to not reuse `testRunUniqueId`?

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/Utils.kt
```diff
@@ -16,7 +16,6 @@ import java.util.concurrent.TimeoutException
 
 const val USERNAME = "admin"
 val PASSWORD = System.getenv("INITIAL_ADMIN_USER_PASSWORD") ?: "admin"
-const val GROUP_ID = "7c5d6948-e17b-44e7-9d1c-fa4a3f667cad"
 val testRunUniqueId: UUID = UUID.randomUUID()
```
vlajos, line 19: It seems that there were some previous attempts to centralize these `testRunUniqueId`s.
Maybe re/using instead of all the other uuid generations would be a bit cleaner.

vlajos (approved):
LGTM. Only a question/idea which is not very significant.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/Utils.kt
vlajos: At least if it is not intentional for some tests to have an isolated groupId

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/AmqpSerializationTests.kt
relyafi: Yeah, my main concern is that it is used for other things. For example, it can factor into the CPI package name. Whilst I don't think re-using again would not be a problem as such, I thought it would be unrealistic - in practice we'd always want a distinct CPI name to group id. So I thought best to keep it separate so it's more realistic, and to avoid introducing any unintended issues.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/Utils.kt
relyafi: See previous comment. Arguably we could centralise `groupId` here as has been done for `testRunUniqueId`, but imo it just obfuscates things from the perspective of a reader of the test.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/AmqpSerializationTests.kt
vlajos: OK. Thanks for checking.

charlieR3 (approved):

PR #3708, opened 2023-04-25
CORE-12714 Improve Avro schema registration logging

simon-johnson-r3:
Logging was displaying `null` in all case because incorrect variable was being logged. Also added a test to exercise this path through the code.


driessamyn (approved):
LGTM

fowlerrr (approved):

PR #3711, opened 2023-04-25
CORE-4126 Refactoring KeyValueStore to prevent Quasar ASM from erroring classloader

ben-millar:
This is a follow-up from https://github.com/corda/corda-runtime-os/pull/3698

Further investigation into the issue by @chrisr3 determined that the root cause of the issue described in [CORE-4126](https://r3-cev.atlassian.net/browse/CORE-4126) lies in how the ASM library, used by Quasar for bytecode instrumentation, builds the `StackMapTable` for the `KeyValuePairList::setValue()` extension function defined in `KeyValueStore` in such a way as to pull in superclass dependencies not present in the original bytecode.

Given these findings, I think it makes sense to implement a finer-grained solution rather than to throw a blanket `@QuasarIgnorePackage` over the affected package and call it a day.

Further details can be found in the original ticket.

[CORE-4126]: https://r3-cev.atlassian.net/browse/CORE-4126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/flows/flow-utils/build.gradle
+++ b/libs/flows/flow-utils/build.gradle
@@ -5,7 +5,6 @@ plugins {
 
 dependencies {
     compileOnly 'org.osgi:osgi.annotation'
-    compileOnly "co.paralleluniverse:quasar-osgi-annotations:$quasarVersion"
 
     implementation platform("net.corda:corda-api:$cordaApiVersion")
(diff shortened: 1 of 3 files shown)


mcgovc (approved):
LGTM

PR #3712, opened 2023-04-25
CORE-11959: make lifecycle being down cause retries in crypto flow ops

dickon:
To that end, make the error thrown by AbstractComponent when invoking operations on components in state DOWN or ERROR be a specific type, since it was IllegalStateException and we don't know if it is appropriate to retry this.

I believe there are situations where errors may not be recoverable, and we will rely on the operator, often via Kubernets, to restart the worker, e.g. because the health check is returning negative. However, if we error out from the CryptoOpsClient into CryptoFlowOpsBusProcessor, then a platform error will be reported to the flow. If we wait, maybe we can proceed either after the service is restarted (e.g. by Kubernetes) or if the lifecycle goes UP again some other way. For instance, if we get a UnknownHostException from Kafka and put CryptoOpsClient down, then without this change we will error out a bunch of flows. With this change, we will wait until the service is restarted and then maybe succeed (if the TTL hasn't expired), or potentially we can have some additional retry logic that marks the CryptoOpsClient UP again when Kafka comes back. 

Long term, we may remove the `AbstractComponent` abstraction, and that takes work in each component that uses it, so this is a short term small fix to improve robustness in 5.0.

### DIFF
--- a/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/AbstractComponentNotReadyException.kt
+++ b/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/AbstractComponentNotReadyException.kt
@@ -1,0 +1,3 @@
+package net.corda.crypto.core
+
+class AbstractComponentNotReadyException(message: String) : IllegalStateException(message)
--- a/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/CryptoExceptionsUtils.kt
+++ b/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/CryptoExceptionsUtils.kt
@@ -17,6 +17,10 @@ val RETRYABLE_EXCEPTIONS = setOf(
 
 fun Throwable.isRecoverable(): Boolean =
     when (this) {
+        // AbstractComponentNotReadyException is thrown when lifecycle is goes away from up on an abstract component, 
+        // e.g. due to a transient or permanent error; we retry since it may come up later, and if it stays down
+        // long enough we will get restarted and this code may have more luck in the new process
+        is AbstractComponentNotReadyException -> true
         is TimeoutException -> true
         is CryptoException -> isRecoverable
         else -> when {
--- a/components/crypto/crypto-component-core-impl/src/main/kotlin/net/corda/crypto/component/impl/AbstractComponent.kt
+++ b/components/crypto/crypto-component-core-impl/src/main/kotlin/net/corda/crypto/component/impl/AbstractComponent.kt
@@ -1,5 +1,6 @@
 package net.corda.crypto.component.impl
 
+import net.corda.crypto.core.AbstractComponentNotReadyException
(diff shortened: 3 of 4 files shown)


libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/AbstractComponentNotReadyException.kt
```diff
@@ -0,0 +1,3 @@
+package net.corda.crypto.core
+
+class AbstractComponentNotReadyException(message: String) : IllegalStateException(message)
```
fowlerrr, line 3: If you can maybe a doc for this?  Though it's fairly obvious so I wouldn't hold a PR up for it.

fowlerrr (approved):
LGTM.  Just the one comment.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/CryptoExceptionsUtils.kt
```diff
@@ -17,6 +17,10 @@ val RETRYABLE_EXCEPTIONS = setOf(
 
 fun Throwable.isRecoverable(): Boolean =
     when (this) {
+        // AbstractComponentNotReadyException is thrown when lifecycle is goes away from up on an abstract component, 
+        // e.g. due to a transient or permanent error; we retry since it may come up later, and if it stays down
+        // long enough we will get restarted and this code may have more luck in the new process
+        is AbstractComponentNotReadyException -> true
```
kyriathar, line 23: What if the component is down for good. It means we will keep retrying for ever? Also this changes the API for all other `AbstractComponent`s as well where this, I understand, is targeting `CryptoOpsClient`?

kyriathar:
I think we already have lifecycle dependency between `CryptoFlowOpsBusProcessor` and `CryptoOpsClient` in place.

The service creating `CryptoFlowOpsBusProcessor` is `CryptoFlowOpsBusServiceImpl`. `CryptoFlowOpsBusServiceImpl` is a lifecycle depending on `CryptoOpsClient` meaning it should wait for it to go UP and then proceed with creating the `CryptoFlowOpsBusProcessor`.

Shouldn't we rely on our lifecycle framework for this? With this extra retrying it feels we are by-passing it.

dickon:
> I think we already have lifecycle dependency between `CryptoFlowOpsBusProcessor` and `CryptoOpsClient` in place.
> 
> The service creating `CryptoFlowOpsBusProcessor` is `CryptoFlowOpsBusServiceImpl`. `CryptoFlowOpsBusServiceImpl` is a lifecycle depending on `CryptoOpsClient` meaning it should wait for it to go UP and then proceed with creating the `CryptoFlowOpsBusProcessor`.
> 
> Shouldn't we rely on our lifecycle framework for this? With this extra retrying it feels we are by-passing it.

That's all correct for when bootstrapping and works fine. The trouble is if Kafka is completely removed to the point that its DNS is no longer even present and then put back; we don't recover.

PR #3713, opened 2023-04-25
CORE-12891 Add ordering to custom query and multiply limit/offset

nkovacsx:
### Overview

For an extended description on the bug and the proposed solution please check the ticket: https://r3-cev.atlassian.net/browse/CORE-12891

### Test plan 

Created a lot of new e2e test scenarios in the e2e test repo and all of them are passing with this bug fix:
https://github.com/corda/corda-e2e-tests/pull/75
Screenshot 2023-04-25 at 16 31 15


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -98,7 +98,8 @@ class VaultNamedQueryExecutorImpl(
                                 "AND visible_states.leaf_idx = tc.leaf_idx " +
                             "$whereJson " +
                             "AND tc.group_idx IN (:groupIndices) " +
-                            "AND visible_states.created <= :$TIMESTAMP_LIMIT_PARAM_NAME",
+                            "AND visible_states.created <= :$TIMESTAMP_LIMIT_PARAM_NAME " +
+                            "ORDER BY tc.created, tc.transaction_id, tc.leaf_idx, tc.group_idx",
```
blsemo, line 102: Do we need to sort by tc.created and tc.transaction? That only makes a difference if two transaction have been created at exactly the same time, doesn't it?

blsemo (dismissed):
Looks ok to me, just one question about the ordering.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
lankydan: I think you might be right in saying that we don't _need_ it, is there any downside to having it? I feel more confident that there won't be issues if we also order by `transaction_id`.

That said, with the `leaf_idx` and `group_idx` ordering, even if the `transaction_id`s interleave by not sorting, we'd still have pairs of output state and output info records, so everything should continue to work.

mnesbit: I think we should leave it in. I did wonder if we could push it to the end, since it is only a tie breaker, but then stuff would be interleaved a bit oddly. Also timestamp granularity is often several milliseconds, so duplicates is inevitable.

mnesbit: The real issue with paging is that we aren't snapshotting the result set in any way, nor are we able to do as-at type queries currently

lankydan (dismissed):

lankydan (approved):

PR #3715, opened 2023-04-26
CORE-12777: Retry to update the registration status if it is not available at first

yift-r3:
Retry asynchronous persistence requests to update the registration request status when the request can not be found.

For testing, I ran the non-functional tests a few times.


components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/UpdateMemberAndRegistrationRequestToApprovedHandler.kt
```diff
@@ -90,7 +91,7 @@ internal class UpdateMemberAndRegistrationRequestToApprovedHandler(
                 RegistrationRequestEntity::class.java,
                 request.registrationId,
                 LockModeType.PESSIMISTIC_WRITE,
-            ) ?: throw MembershipPersistenceException("Could not find registration request: ${request.registrationId}")
+            ) ?: throw RecoverableException("Could not find registration request: ${request.registrationId}")
```
dimosr: is this strictly needed for this method? 

Seems like this is only called from the approval handler and at the point this is executed it's guaranteed the registration must have been persisted (e.g. the start registration handler persists it synchronously)

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/RecoverableException.kt
```diff
@@ -0,0 +1,5 @@
+package net.corda.membership.impl.persistence.service
+
+import net.corda.membership.lib.exceptions.MembershipPersistenceException
+
+internal class RecoverableException(message: String) : MembershipPersistenceException(message)
```
dimosr, line 8: worth adding some kdocs here on the purpose & expectations of this exception, since it's something generic that will probably be adopted elsewhere gradually.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/UpdateRegistrationRequestStatusHandler.kt
```diff
@@ -18,7 +18,7 @@ internal class UpdateRegistrationRequestStatusHandler(
                 RegistrationRequestEntity::class.java,
                 request.registrationId,
                 LockModeType.PESSIMISTIC_WRITE
-            ) ?: throw MembershipPersistenceException("Could not find registration request: ${request.registrationId}")
+            ) ?: throw RecoverableException("Could not find registration request: ${request.registrationId}")
```
dimosr: worth adding a comment here explaining why this is recoverable for future readers (i.e. because initial persistence of the registration request on the member side can be asynchronous in some cases) as looking locally only at the code here it wouldn't be very obvious.

dimosr (approved):

PR #3718, opened 2023-04-26
CORE-12539: Remove unwanted transitive dependency from jackson-core-2.15.0.

chrisr3:
Prevent spurious transitive dependency of `jackson-core` from polluting our applications.

Note that this only removes the unwanted dependency from each application jar as that application is assembled.

### DIFF
--- a/buildSrc/src/main/groovy/corda.common-app.gradle
+++ b/buildSrc/src/main/groovy/corda.common-app.gradle
@@ -51,6 +51,9 @@ configurations {
         attributes { attrs ->
             attrs.attribute(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE, objects.named(LibraryElements, LibraryElements.JAR))
         }
+
+        // Remove unwanted transitive dependency from jackson-core-2.15.0
+        exclude group: 'ch.randelshofer', module: 'fastdoubleparser'
     }
     applyDependencySubstitution(systemPackages)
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/corda.common-app.gradle
```diff
@@ -51,6 +51,9 @@ configurations {
         attributes { attrs ->
             attrs.attribute(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE, objects.named(LibraryElements, LibraryElements.JAR))
         }
+
+        // Remove unwanted transitive dependency from jackson-core-2.15.0
+        exclude group: 'ch.randelshofer', module: 'fastdoubleparser'
```
driessamyn, line 56: Does this mean 

https://github.com/corda/corda-runtime-os/blob/e89057bf06af7d9bd04930001cea2649752484e1/testing/cpbs/sandbox-security-manager-one/build.gradle#L23 
and 
https://github.com/corda/corda-runtime-os/blob/e89057bf06af7d9bd04930001cea2649752484e1/testing/cpbs/sandbox-security-manager-two/build.gradle#L24

can be removed?

chrisr3: No, because those are CPK projects and do not apply the `corda.common-app` plugin.

driessamyn (approved):
LGTM

mcgovc (approved):
LGTM

PR #3720, opened 2023-04-26
CORE-13030: Add a batch publish API

JamesHR3:
### Problem description
The flow REST resource can show some thread contention when lots of clients are trying to publish start flow events simultaneously. This occurs as the publisher in the REST resource is transactional to ensure that the start flow and status events are published together. However, by enabling transactionality, the publisher only allows a single thread to publish its records at once, resulting in a bottleneck.

### Solution
To address this, a new API is added to the publisher that allows requests from multiple threads to be grouped together in a single batch when transactionality is enabled. This works by adding the requests to a queue, and then waiting to acquire a lock. When the lock is acquired, the full queue is drained and all records are published in a single transaction. This means a thread may publish records from multiple threads at once.

### Impact
This change shows an improved start flow latency when running many clients in the soak tests. Previously the median time was 800ms to respond to a start flow request with 96 clients. With this change, response times are reduced to an order of 80ms.

### Other changes
- Adjust the state and event processing time metric to record the time to process a (sub-)batch rather than an individual event. This is much easier to interpret when looking at metrics coming out the cluster.


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaPublisherImpl.kt
```diff
@@ -34,10 +37,16 @@ internal class CordaPublisherImpl(
 
     private companion object {
         private val log: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+
+        private const val QUEUE_SIZE = 200
```
sprintengine, line 41: What should we do about making things like this configurable at this stage, is it considered a breaking change to config APIs to add an optional messaging configuration like this?

Similar question about flow fiber cache size, currently I have system property which doesn't feel quite correct, though this is also used for repeated logons cache.

https://github.com/corda/corda-runtime-os/blob/conal/CORE-12070-add-fiber-caching/components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/FlowFiberCacheImpl.kt#L14

JamesHR3: Adding configuration isn't a breaking change. I think providing the ability to tune the fiber cache would make a lot of sense and we probably should make that configurable. Less sure how important it is to make this configurable - it's memory vs client contention in this case, but the memory concern is likely not that great and you don't get more than 200 REST client threads anyway, so it probably doesn't matter that this is set like this.

jujoramos: Was about to ask the same question, @conalsmith-r3 beat me to it 👍 .

jujoramos (dismissed):
LGTM

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaPublisherImpl.kt
rick-r3: 200 is probably fine for the time being as it is not particularly expensive to have an array that size and unless there are more than 200 threads trying to concurrently publish via a publisher, it won't be exceeded.  

It is worth thinking though how it might be plumbed through in the future.  e.g. if the threads in the REST worker became configurable, you'd want the REST worker to also be able to pass in a value here through one of the constructor args.  Doesn't seem like you'd want a global setting.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaPublisherImpl.kt
```diff
@@ -73,6 +82,35 @@ internal class CordaPublisherImpl(
...
+    override fun batchPublish(records: List<Record<*, *>>): CompletableFuture<Unit> {
+        if (!config.transactional) {
+            throw CordaMessageAPIFatalException(
+                "Cannot use the batch publish API unless the publisher is transactional. Client id: ${config.clientId}"
+            )
+        }
+        val batch = Batch(records, CompletableFuture())
+        queue.add(batch)
+        lock.withLock {
+            val batches = mutableListOf<Batch>()
```
rick-r3, line 94: Yes, I think we continue investigations into more optimal implementations in a follow up.  File a JIRA?

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/publisher/Publisher.kt
```diff
@@ -40,4 +40,21 @@ interface Publisher : Resource {
...
      */
     fun publish(records: List<Record<*, *>>): List<CompletableFuture<Unit>>
+
+    /**
+     * Publish a list of records as part of a batch.
+     *
+     * This API should be used for transactional publishers that are heavily contended across multiple threads. The API
+     * will attempt to group the supplied records with other lists of records provided by other threads, and commit them
+     * within a single Kafka transaction. This improves performance by reducing the number of transactions a thread has
+     * to wait for before its records get committed.
```
rick-r3, line 50: I would have said it also leads to bigger batches of operations inside the transactions which is more efficient for throughput, but I wouldn't necessarily go to the effort of making just this change unless you need to change something else.

conal-smith (approved):

rick-r3 (approved):

PR #3724, opened 2023-04-26
CORE-12707: Add support to automatically remove secrets that are only…

oscardalmau-r3:
… used for bootsraping


charts/corda-lib/templates/_helpers.tpl
```diff
@@ -491,6 +493,9 @@ metadata:
   annotations:
     "helm.sh/hook-weight": "-1"
     "helm.sh/hook": pre-install
+{{- if $options.cleanup | default false }}
```
davidcurrie: on the basis, less is more, I still think you can get away with
```suggestion
{{- if $options.cleanup }}   
```

davidcurrie (approved):
LGTM

PR #3725, opened 2023-04-26
CORE-10324: Validate SAN when creating a CSR

yift-r3:
This pull request will verify the hostname when creating a CSR.


dimosr (approved):
LGTM

PR #3727, opened 2023-04-26
CORE-13041: Forbid fragment bundles inside sandboxes.

chrisr3:
Adjust the security policy so that sandboxes may neither contain fragment bundles nor use other fragment bundles.

Also ensure that all sandboxes apply the same security policy regarding sockets and URLs. (And remove some weird unicode characters which were displayed by IntelliJ?!)


driessamyn (approved):
LGTM

PR #3734, opened 2023-04-27
CORE-12553 Corda CLI package `create-cpi` command outputs stacktrace if no truststore

kyriathar:
- makes error message user-friendly by not including stack trace
- slightly modifies error message to be consistent with equivalent message in `verify` command
- removes duplicate test

**Error message** at failing to verify CPB **now looks like**: "_Error verifying CPB: ${error message}_"


driessamyn (approved):
LGTM

PR #3736, opened 2023-04-27
CORE-8835: split resources for bootstrap and workers

JasonSinclair95:
the resources field is global and can be overridden under the bootstrap section but, if not set, a short lived job will end up getting the same cpu and memory requests and limits  as an actual long lived worker pod. 

now bootstrap can only get its resources from Values.bootstrap.resources and can not default to the workers values


oscardalmau-r3 (approved):
LGTM

PR #3739, opened 2023-04-27
CORE-9004: Snakeyaml waiver removed

nargas-ritu:
Snakeyaml version is upgraded which removes the vulnerability. Hence the waiver is removed from the file.


driessamyn (approved):
LGTM

PR #3740, opened 2023-04-27
CORE-13096: Fix flaky SandboxGroupContextTest.

chrisr3:
Invoke `System.gc()` during `eventually {}` processing to ensure weak references are processed.

Also convert all `double` metric values to `int` since we cannot have 2.334 cache hits / misses!?


jujoramos (approved):
👍

driessamyn (approved):
LGTM

PR #3742, opened 2023-04-27
CORE-7793 - Failed serialization when producing Kafka records should send the Flow to the DLQ

pmck91:
To fix this potential infinite loop, we can return null for un-serializable records, and push an explanation sting to the DLQ as we cant push the data that failed to serialize. 

For other uses of the serializer they may expect the exception, so we only return null in the case a throwOnSerializationError is true.

We have decided that the api for corda's avro serializers should be moved from corda-api to runtime-os
I have ported CordaAvroSerializer, CordaAvroDeserializer, CordaAvroSerializationFactory. 

I have added a Boolean field (defaults to true to match current behaviour) 'throwOnSerializationError ' to the CordaProducerBuilder. this will allow us to decide if we wish to maintain current behaviour and throw when a serialization error occurs, or swallow it and allow it be handled by the onError handler only. 

Most of the files in this PR are just import changes. 
The following are files of interest:

- CordaSubscriptionFactory
- StateAndEventBuilder
- StateAndEventSubscriptionImpl
- CordaAvroSerializer
- CordaAvroSerializerImpl
- CordaAvroDeserializer
- CordaAvroSerializationFactory
- KafkaCordaProducerBuilderImpl


libs/serialization/serialization-avro/src/main/java/net/corda/serialization/package-info.java
```diff
@@ -0,0 +1,4 @@
+@Export
+package net.corda.serialization;
```
LWogan: We should move this new lib to a different package other than net.corda.serialization;.  Maybe something with the word avro in it would do?

In OSGI we cant have 2 bundles  export the same package. If we did then  OSGI wouldnt know which bundle to wire to, to get the classes for an imported package. OSGI doesn't look inside the packages for specific classes when wiring together bundles. it just wires import package to export package at startup so it would have a 50/50 chance of getting it right.

Serialization-internal has  already taken net.corda.serialization
https://github.com/corda/corda-runtime-os/blob/release/os/5.0/libs/serialization/serialization-internal/src/main/java/net/corda/serialization/package-info.java

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
```diff
@@ -29,7 +29,7 @@ internal class AddNotaryToGroupParametersHandler(
     }
 
     private val keyValuePairListSerializer: CordaAvroSerializer<KeyValuePairList> =
-        cordaAvroSerializationFactory.createAvroSerializer {
+        cordaAvroSerializationFactory.createAvroSerializer(false) {
```
LWogan: I think this is the only one youve set it to false for. is it an easter egg to see if we're paying attention :D

I think lets just leave it as the default "true" and leave it up to this ticket to modify the logic to handle the error case better https://r3-cev.atlassian.net/browse/CORE-9673

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/serialization/CordaAvroSerializerImpl.kt
```diff
@@ -1,13 +1,16 @@
...
 
-import net.corda.data.CordaAvroSerializer
+import java.util.function.Consumer
+import net.corda.avro.serialization.CordaAvroSerializer
 import net.corda.schema.registry.AvroSchemaRegistry
 import org.apache.kafka.common.serialization.Serializer
 import org.apache.kafka.common.serialization.StringSerializer
 import org.slf4j.LoggerFactory
 
 class CordaAvroSerializerImpl<T : Any>(
```
LWogan, line 18: we should put some kdocs and @param tags on this class to help explain the logic

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/serialization/CordaAvroSerializerImpl.kt
```diff
@@ -23,9 +26,21 @@ class CordaAvroSerializerImpl<T : Any>(
                 try {
                     schemaRegistry.serialize(data).array()
                 } catch (ex: Throwable) {
-                    log.error("Failed to serialize instance of class type ${data::class.java.name} containing " +
-                            "$data", ex)
-                    throw ex
+                    // We don't want to throw as that would mean the entire poll (with possibly
```
LWogan: i think move this comment up to the class level. might need a little rewrite. it reads as if this class only handles message pattern serialization, might confuse someone not familiar with the message patterns, consumer and produce loop and that use case etc

libs/messaging/kafka-message-bus-impl/src/test/kotlin/net/corda/messagebus/kafka/serialization/CordaAvroSerializerImplTest.kt
```diff
@@ -14,7 +14,7 @@ class CordaAvroSerializerImplTest {
 
     private val topic = "topic"
     private val avroSchemaRegistry : AvroSchemaRegistry = mock()
-    private val cordaAvroSerializer = CordaAvroSerializerImpl<Any>(avroSchemaRegistry)
+    private val cordaAvroSerializer = CordaAvroSerializerImpl<Any>(avroSchemaRegistry, true, null)
 
     data class SerializeTester(val contents: String = "test contents")
```
LWogan, line 21: You mentioned you're going to add a unit test, just in case anyone else adds a comment!

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/builder/CordaProducerBuilder.kt
```diff
@@ -13,8 +13,15 @@ interface CordaProducerBuilder {
      * Generate kafka producer with given properties.
      * @param producerConfig The mandatory config for setting up producers
      * @param messageBusConfig Configuration for connecting to the message bus and controlling its behaviour.
+     * @param onSerializationError a callback to receive messages that fail to serialize.  In the producer feed
```
LWogan: missing kdoc for throwOnError. 

kdoc for onSerializationError might need a little altering as it has the old description about it being used to decide whether to return null

libs/serialization/serialization-avro/build.gradle
```diff
@@ -0,0 +1,15 @@
...
+    id 'corda.common-library'
+    id 'corda.common-publishing'
+}
+
+description 'AVRO Serializer and Deserializer'
+
+dependencies {
+    compileOnly "org.osgi:osgi.annotation"
+    compileOnly "org.osgi:org.osgi.service.component.annotations"
+    implementation platform("net.corda:corda-api:$cordaApiVersion")
```
LWogan, line 12: i dont think you need

libs/serialization/serialization-avro/src/main/kotlin/net/corda/avro/serialization/CordaAvroSerializationFactory.kt
```diff
@@ -0,0 +1,22 @@
...
+import java.util.function.Consumer
+
+/**
+ * Defines the interface for message bus deserialization. The underlying mechanism may differ.
+ */
+interface CordaAvroSerializationFactory {
+
+    /**
+     * Create the {@link CordaAvroSerializer} for use in Avro/message bus serialization.
+     */
```
LWogan, line 15: They were missing already before but we should add some @param and @return  kdocs to explain clearly the new functionality and how it relates to the flags. not sure why we didnt have them before already. 

I know it hasnt changed since the move from corda-api but please add some for the createAvroDeserializer method too.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/serialization/CordaAvroSerializerImpl.kt
```diff
@@ -23,9 +26,21 @@ class CordaAvroSerializerImpl<T : Any>(
...
+                    // many records) would fail, and keep failing. So we'll just callback to note the serialize
+                    // and return a null.  This will mean the record gets treated as 'deleted' in the processors
+                    val message = "Failed to serialize instance of class type ${data::class.java.name} containing " +
```
LWogan: this should fit on one line. i think detekt is configured to 140 line lengths. intellij defaults to 120

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
```diff
@@ -122,7 +122,10 @@ internal class StateAndEventSubscriptionImpl<K : Any, S : Any, E : Any>(
             attempts++
             try {
                 deadLetterRecords = mutableListOf()
-                nullableProducer = builder.createProducer(config)
+                nullableProducer = builder.createProducer(config, false) { data ->
```
LWogan: we should add this dead letter records lambda to the same place in  EventLogSubscriptionImpl

LWogan (commented):
Looks good. Glad to move this out of corda-api given we dont have any intentions on letting customers implement their own avro serializers. These API changes will allow us to better handle these serialization error use cases!

Given this is mostly complete I think we should apply the same changes youve made in CordaAvroSerializerImpl  to the CordaDBAvroSerializerImpl. Should be quick enough and this will keep the logic in our two bus implementations consistent. 

This would allow us to write a db message bus integration test easily to test the dead letter functionality too _I think_. We have some similar tests for the deserialization dead letter logic in testing:message-pattern:integrationTests DurableSubscriptionIntegrationTest & StateAndEventSubscriptionIntegrationTest. We could try out writing a processor that outputs non avro/string/bytearray type objects and see if it will fail serialization and trigger the DLQ.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
pmck91: I was searching for this and couldnt find it!

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/builder/StateAndEventBuilder.kt
```diff
@@ -8,7 +8,12 @@ import net.corda.messaging.subscription.consumer.listener.StateAndEventConsumerR
 
 @Suppress("LongParameterList")
 interface StateAndEventBuilder {
-    fun createProducer(config: ResolvedSubscriptionConfig): CordaProducer
+    fun createProducer(
+        config: ResolvedSubscriptionConfig,
+        throwOnError: Boolean = true,
```
simon-johnson-r3: From this layer of abstraction, and those down except the serializer itself I think `throwOnError` is misleading. I had to go and look what sort of errors it would throw on, and then it wasn't quite clear whether I would get `onSerializationError` called either way, which required another follow through of the code.

I suggest renaming to `throwOnSerializationError` explaining why that is important to control (as per @LWogan's comment above about this information being in the wrong place).

Then I would document that `onSerializationError` is called no matter what `throwOnSerializtionError` is set to.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/serialization/CordaAvroSerializerImpl.kt
```diff
@@ -23,9 +26,21 @@ class CordaAvroSerializerImpl<T : Any>(
...
+                                "$data"
+
+                    if(throwOnError) {
+                        log.error(message, ex)
+                        runOnErrorLambda(message, onError)
```
simon-johnson-r3: I can't see a compelling reason not to move `runOnErrorLambda` above the `if`?
Then you can remove the helper method entirely as it's only a one liner anyway.

libs/serialization/serialization-avro/src/main/kotlin/net/corda/avro/serialization/CordaAvroSerializationFactory.kt
```diff
@@ -0,0 +1,22 @@
...
+    fun <T : Any> createAvroSerializer(throwOnError: Boolean = true, onError: Consumer<ByteArray>? = null): CordaAvroSerializer<T>
+
+    /**
+     * Create the {@link CordaAvroDeserializer} for use in Avro/message bus serialization.
+     */
+    fun <T : Any> createAvroDeserializer(
+        onError: Consumer<ByteArray>,
```
simon-johnson-r3: Small point, and not related to this change, but I find it unusual that when I'm creating a deserializer, the first parameter is not the class but the error consumer.

I also find the use of Consumer<> here a little idiosyncratic when this is really an error callback. Consumer has a semantic meaning that is not really appropriate in this use case. Perhaps this is because it was moved from a Java api and some Java functional interface to represent the callback was required then? Either way I think this will confuse future maintainers and accepting a Kotlin lambda would be better and consistent with the other code here.

simon-johnson-r3 (changes requested):

libs/serialization/serialization-avro/src/main/kotlin/net/corda/avro/serialization/CordaAvroSerializationFactory.kt
pmck91: Ive swapped it to a regular lambda now

applications/workers/release/rest-worker/build.gradle
```diff
@@ -70,6 +70,7 @@ dependencies {
     e2eTestImplementation project(':libs:virtual-node:cpi-upload-endpoints')
     e2eTestImplementation project(':libs:virtual-node:virtual-node-endpoints')
     e2eTestImplementation project(':libs:lifecycle:lifecycle')
+    e2eTestImplementation project(':libs:serialization:serialization-avro')
```
vkolomeyko: Why do we need this extra "implementation" dependency when no source files been modified which use classes from that module?

seanbrereton (dismissed):
LGTM(build files only)

libs/serialization/serialization-avro/src/main/kotlin/net/corda/avro/serialization/CordaAvroSerializationFactory.kt
simon-johnson-r3: Am I correct in thinking you switched it on createAvroSerializer but not createAvroDeserializer?

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/serialization/CordaAvroDeserializerImpl.kt
```diff
@@ -11,7 +11,14 @@ import org.apache.kafka.common.serialization.Deserializer
 import org.apache.kafka.common.serialization.StringDeserializer
 import org.slf4j.LoggerFactory
 
-class CordaAvroDeserializerImpl<T : Any>(
+/**
+ * Corda avro serializer impl
+ *
+ * @param T Type to deserialize
+ * @property schemaRegistry the Avro-based Schemas
+ * @property onError lambda to be run on deserialization error
```
LWogan, line 19: expectedClass missing

libs/serialization/serialization-avro/src/main/kotlin/net/corda/avro/serialization/CordaAvroSerializationFactory.kt
pmck91: missed it! added now

simon-johnson-r3 (dismissed):

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/builder/CordaProducerBuilder.kt
LWogan: This is still a bit off
```
@param onSerializationError a callback to receive messages that fail to serialize. In the producer feed
these will show up as records with a null value, which means they should be removed from any maps.
```

should be something like:
`@param onSerializationError a callback to execute when serialization fails 
`
theres other stuff that uses the serializer, not just producers, and this flag doesnt flag whether we return null anymore

LWogan (commented):
Looking good! Almost there!

applications/workers/release/rest-worker/build.gradle
pmck91: without it the task :applications:workers:release:rest-worker:compileE2eTestKotlin fails as it cant find the supertypes which were previously supplied by corda api

```
Supertypes of the following classes cannot be resolved. Please make sure you have the required dependencies in the classpath:
    class net.corda.messagebus.kafka.serialization.CordaAvroSerializationFactoryImpl, unresolved supertypes: net.corda.avro.serialization.CordaAvroSerializationFactory
```

vkolomeyko: But if this is the case should module containing `CordaAvroSerializationFactoryImpl` have dependency on module containing `CordaAvroSerializationFactory` as `api` dependency?
If transitive dependency expressed that way this will eliminate the need to put `implementation project(':libs:serialization:serialization-avro')` in many modules.

Please see my suggestion: https://github.com/corda/corda-runtime-os/pull/3742/files#r1185240153

libs/messaging/kafka-message-bus-impl/build.gradle
```diff
@@ -16,6 +16,7 @@ dependencies {
     implementation project(":libs:configuration:configuration-merger")
     implementation project(":libs:metrics")
     implementation project(":libs:utilities")
+    implementation project(":libs:serialization:serialization-avro")
```
vkolomeyko:
```suggestion
    api project(":libs:serialization:serialization-avro")
```
Due to `CordaAvroSerializationFactoryImpl` using `CordaAvroSerializationFactory` as a super-type.

components/chunking/chunk-db-write-impl/build.gradle
```diff
@@ -46,6 +46,7 @@ dependencies {
     implementation project(':libs:membership:network-info')
     implementation project(':components:chunking:chunk-db-write')
     implementation project(':components:membership:certificates-service')
+    implementation project(":libs:serialization:serialization-avro")
```
vkolomeyko: Do we need this line?

components/flow/flow-p2p-filter-service/build.gradle
```diff
@@ -13,6 +13,7 @@ dependencies {
     implementation project(":libs:messaging:messaging")
     implementation project(":libs:flows:session-manager")
     implementation project(':libs:utilities')
+    implementation project(":libs:serialization:serialization-avro")
```
vkolomeyko, line 16: Do we need this line?

components/membership/membership-service-impl/build.gradle
```diff
@@ -27,6 +27,7 @@ dependencies {
     implementation project(":components:membership:registration")
     implementation project(":components:virtual-node:virtual-node-info-read-service")
     implementation project(":components:crypto:crypto-client")
+    implementation project(":libs:serialization:serialization-avro")
```
vkolomeyko, line 30: Do we need this line, provided that no source code been changed?

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/builder/CordaProducerBuilder.kt
```diff
@@ -13,8 +13,14 @@ interface CordaProducerBuilder {
      * Generate kafka producer with given properties.
      * @param producerConfig The mandatory config for setting up producers
      * @param messageBusConfig Configuration for connecting to the message bus and controlling its behaviour.
+     * @param throwOnSerializationError throw exception on error or return null defaults to true
```
LWogan: can remove throwOnSerializationError now

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/builder/StateAndEventBuilder.kt
```diff
@@ -8,7 +8,19 @@ import net.corda.messaging.subscription.consumer.listener.StateAndEventConsumerR
 
 @Suppress("LongParameterList")
 interface StateAndEventBuilder {
-    fun createProducer(config: ResolvedSubscriptionConfig): CordaProducer
+    /**
+     * Create producer
+     *
+     * @param config subscription configuration for the messaging layer.
+     * @param throwOnSerializationError throw exception on serialization error, or return null. defaults to throw
```
LWogan: need to remove throwOnSerializationError kdoc

libs/serialization/serialization-avro/src/main/kotlin/net/corda/avro/serialization/CordaAvroSerializationFactory.kt
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Defines the interface for message bus deserialization. The underlying mechanism may differ.
+ */
+interface CordaAvroSerializationFactory {
+
+    /**
+     * Create the {@link CordaAvroSerializer} for use in Avro/message bus serialization.
+     *
+     * @param T the type to serialize
+     * @param throwOnSerializationError throw an exception on serialization failure, or return null (defaults to true)
```
LWogan: need to remove throwOnSerializationError kdoc

LWogan (commented):
couple kdocs to remove, and we need to merge the other pr into this one. Other than that this is good to go

components/flow/flow-p2p-filter-service/build.gradle
LWogan: there is code in this lib using this in flow p2p filter

components/membership/membership-service-impl/build.gradle
LWogan: this one is used too

components/chunking/chunk-db-write-impl/build.gradle
LWogan: this one could be changed to integrationTestImplementation i think

LWogan (dismissed):
LGTM

thiagoviana (dismissed):

charlieR3 (approved):
LGTM (network team files)

mag1c-48 (approved):
LGTM - build only

simon-johnson-r3 (approved):

PR #3743, opened 2023-04-27
CORE-11656 Add validations to `DigestServiceImpl.parseSecureHash`

kyriathar:
- adds validations on secure hash parsing; that 1. digest algorithm is supported and 2. the hex value of the hash is of the expected length:
  - adds `PlatformDigestService.parseSecureHash` to delegate for platform digest algorithms validations
  - adds custom algorithms validations in `DigestServiceImpl.parseSecureHash`

### DIFF
--- a/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
+++ b/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
@@ -38,18 +38,21 @@ fun parseDigestAlgoName(algoNameAndHexString: String): String {
     return algoNameAndHexString.substring(0, idx)
 }
 
-fun parseSecureHash(algoNameAndHexString: String, expectedLength: Int? = null): SecureHash {
+fun parseHexString(algoNameAndHexString: String): String {
+    val idx = algoNameAndHexString.indexOf(DELIMITER)
+    if (idx == -1) {
+        throw IllegalArgumentException("Provided string: $algoNameAndHexString should be of format algorithm:hexadecimal")
+    }
+    return algoNameAndHexString.substring(idx + 1)
+}
+
+fun parseSecureHash(algoNameAndHexString: String): SecureHash {
     val idx = algoNameAndHexString.indexOf(DELIMITER)
     return if (idx == -1) {
         throw IllegalArgumentException("Provided string: $algoNameAndHexString should be of format algorithm:hexadecimal")
     } else {
         val algorithm = algoNameAndHexString.substring(0, idx)
         val value = algoNameAndHexString.substring(idx + 1)
-        expectedLength?.let {
-            require(it == value.length) {
-                "Required hex string length: $it for algo: \"$algorithm\" is not met by the provided hex string: \"$value\""
-            }
-        }
         val data = ByteArrays.parseAsHex(value)
         SecureHashImpl(algorithm, data)
     }
--- a/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
+++ b/libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
@@ -30,7 +30,7 @@ class SecureHashImpl(
     override fun toString() = "$algorithm$DELIMITER${toHexString()}"
 }
 
-fun parseDigestAlgoName(algoNameAndHexString: String): String {
+fun parseSecureHashAlgoName(algoNameAndHexString: String): String {
     val idx = algoNameAndHexString.indexOf(DELIMITER)
     if (idx == -1) {
         throw IllegalArgumentException("Provided string: $algoNameAndHexString should be of format algorithm:hexadecimal")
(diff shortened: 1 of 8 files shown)


libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
```diff
@@ -30,13 +30,26 @@ class SecureHashImpl(
...
+    if (idx == -1) {
+        throw IllegalArgumentException("Provided string: $algoNameAndHexString should be of format algorithm:hexadecimal")
+    }
+    return algoNameAndHexString.substring(0, idx)
+}
+
+fun parseSecureHash(algoNameAndHexString: String, expectedLength: Int? = null): SecureHash {
     val idx = algoNameAndHexString.indexOf(DELIMITER)
     return if (idx == -1) {
         throw IllegalArgumentException("Provided string: $algoNameAndHexString should be of format algorithm:hexadecimal")
```
driessamyn: Should this not re-use `parseDigestAlgoName` now that it exists?

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
```diff
@@ -30,13 +30,26 @@ class SecureHashImpl(
...
-fun parseSecureHash(algoNameAndHexString: String): SecureHash {
+fun parseDigestAlgoName(algoNameAndHexString: String): String {
+    val idx = algoNameAndHexString.indexOf(DELIMITER)
+    if (idx == -1) {
+        throw IllegalArgumentException("Provided string: $algoNameAndHexString should be of format algorithm:hexadecimal")
+    }
+    return algoNameAndHexString.substring(0, idx)
+}
+
+fun parseSecureHash(algoNameAndHexString: String, expectedLength: Int? = null): SecureHash {
```
driessamyn: Shouldn't expected length be defaulted to the length for the given algoName for the ones we know off?

kyriathar: We are not aware of digest algorithms at this point, it's the `DigestService` that manages them/ is aware of them. That's why I left it null, it wasn't there at all. This is an extra check if the caller wants it to be more restrictive, so thought it would be useful to be there.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
kyriathar: I think the way it currently is we have the index and we grab substrings based on that. It means `parseDigestAlgoName` would have to return index as well or we make two functions `parseDigestAlgoName` and `parseDigestHexString` but I believe the way it currently is looks ok. Happy to change it though.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
driessamyn: What I meant was that internally we can validate the length if it's a algorithm we know. But I'm not sure about the usefulness of it as why care about the length? It either parses or not.

I would prefer not to have it if it's not used, I don't like "just in case" functionality, it's an unnecessary tax. This is internal API, so we can add it once it becomes useful. YAGNI.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
driessamyn: It's just that we have 4 lines of identical code. And one of the functions is just 4 lines.

> It means parseDigestAlgoName would have to return index as well

I don't think so. Once you know the `algoName` you know the index for the hash is `algoName.size()`

kyriathar: Done

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
kyriathar: Reverted

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SecureHashImpl.kt
```diff
@@ -30,16 +30,23 @@ class SecureHashImpl(
     override fun toString() = "$algorithm$DELIMITER${toHexString()}"
 }
 
-fun parseSecureHash(algoNameAndHexString: String): SecureHash {
+fun parseSecureHashAlgoName(algoNameAndHexString: String): String {
```
dickon, line 33: Better to spell out names in full, i.e. Algorithm rather than Algo, so `parseSecureHashAlgorithmName`. This is becasue if you abbreviate people have to remember which abbreviation you picked.

Not a blocking comment

dickon (approved):
LGTM.

I did wonder if there's such a thing as a hash algorithm which produces a variable length output, but I'm not aware of any so no point in making this more complex and the length validation is useful in real cases.

driessamyn (approved):

PR #3744, opened 2023-04-27
CORE-12706: improve accuracy in notes

JasonSinclair95:
The NOTES output on installation of the Corda Helm chart do not always reflect reality

### DIFF
--- a/charts/corda/templates/NOTES.txt
+++ b/charts/corda/templates/NOTES.txt
@@ -1,10 +1,22 @@
+{{- if .Values.bootstrap.db.enabled }}
 1. Extract the username and password for the REST API admin created during bootstrap:
 kubectl get secret {{ include "corda.restApiAdminUsernameSecretName" . }} -o go-template='{{ `{{` }} .data.{{ include "corda.restApiAdminSecretUsernameKey" . }} | base64decode {{ `}}` }}'
 kubectl get secret {{ include "corda.restApiAdminPasswordSecretName" . }} -o go-template='{{ `{{` }} .data.{{ include "corda.restApiAdminSecretPasswordKey" . }} | base64decode {{ `}}` }}'
+{{- end }}
+
+{{- if .Values.workers.rest.ingress.hosts }}
+
+2. The API endpoint definition can then be accessed via: 
+{{- range .Values.workers.rest.ingress.hosts }}
+https://{{ . }}/api/v1/swagger
+{{- end }}
+{{- else }}
 
 2. Expose the API endpoint on localhost by running this command:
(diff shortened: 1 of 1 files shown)


charts/corda/templates/NOTES.txt
```diff
@@ -1,10 +1,22 @@
+{{- if .Values.bootstrap.db.enabled }}
 1. Extract the username and password for the REST API admin created during bootstrap:
 kubectl get secret {{ include "corda.restApiAdminUsernameSecretName" . }} -o go-template='{{ `{{` }} .data.{{ include "corda.restApiAdminSecretUsernameKey" . }} | base64decode {{ `}}` }}'
 kubectl get secret {{ include "corda.restApiAdminPasswordSecretName" . }} -o go-template='{{ `{{` }} .data.{{ include "corda.restApiAdminSecretPasswordKey" . }} | base64decode {{ `}}` }}'
+{{- end }}
```
davidcurrie, line 7: I'd add an `else` here that says something like "1. The REST API admin username and password were specified when the configuration database was initialized." if only so that your bullet points don't then start at 2!

charts/corda/templates/NOTES.txt
```diff
@@ -1,10 +1,22 @@
...
+
+{{- if .Values.workers.rest.ingress.hosts }}
+
```
davidcurrie, line 10: is the empty line needed here and after the `else` given you have one before the `if`?

charts/corda/templates/NOTES.txt
```diff
@@ -1,10 +1,22 @@
...
+{{- else }}
 
 2. Expose the API endpoint on localhost by running this command:
 kubectl port-forward --namespace {{ .Release.Namespace }} deployment/{{ include "corda.fullname" . }}-rest-worker 8888 &
 
 3. The API endpoint definition can then be accessed via: https://localhost:8888/api/v1/swagger
 
-For more information, see the Corda 5 documentation at docs.r3.com.
+{{- end }}
+
```
davidcurrie, line 23: similarly, you can probably drop the empty line before `end` given you have one after...

davidcurrie (changes requested):

charts/corda/templates/NOTES.txt
JasonSinclair95: yes otherwise the notes have no space divider

charts/corda/templates/NOTES.txt
JasonSinclair95: same thing here, the notes will have no line dividing them

charts/corda/templates/NOTES.txt
davidcurrie: doesn't the one before the `if` count?

davidcurrie (approved):
🚀

PR #3745, opened 2023-04-27
CORE-12619: Force test to fail without custom StateAndRef AMQP serializer.

chrisr3:
Modify test scenario so that it actually _requires_ a custom AMQP serializer for `StateAndRefImpl`. We do this by dropping our `StateAndRef` into a generic `Bucket`, and then serializing / deserializing that instead.


dickon (approved):
LGTM (with limited understanding)

driessamyn (approved):
LGTM

PR #3746, opened 2023-04-27
CORE-12804: Propogate the error from the persistante service

yift-r3:
This should propagate the error type from the DB worker to the persistence client and support the `INVALID_ENTITY_UPDATE` <-> `InvalidEntityUpdateException` mapping.

The API pull request is in https://github.com/corda/corda-api/pull/1076


libs/utilities/src/main/kotlin/net/corda/utilities/Either.kt
```diff
@@ -2,9 +2,16 @@ package net.corda.utilities
...
     data class Left<T>(
-        val a: T
+        val a: T,
     ) : Either<T, Nothing>
     data class Right<S>(
-        val b: S
+        val b: S,
     ) : Either<Nothing, S>
+
+    fun <U> mapRight(mapper: (S) -> U): Either<T, U> {
```
williamvigorr3, line 11: Nit: I think `mapIfRight` would be slightly clearer?

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/MembershipPersistenceRPCProcessor.kt
```diff
@@ -35,7 +37,11 @@ internal class MembershipPersistenceRPCProcessor(
         } catch (e: Exception) {
             val error = "Exception thrown while processing membership persistence request: ${e.message}"
             logger.warn(error)
-            PersistenceFailedResponse(error)
+            val kind = when (e) {
+                is InvalidEntityUpdateException -> ErrorKind.INVALID_ENTITY_UPDATE
```
williamvigorr3, line 41: So I think a `PessimisticLockException` could also be thrown by the `ActivateMemberHandler`/`SuspendMemberHandler` and if that happens we want to send the same response back to the REST worker. For example that could happen if re-registration happened at the same time as suspension.

This is different to registration, we want the user to retry the REST request in this case.

williamvigorr3 (commented):
I left a couple of comments. Maybe this is just my personal opinion but I would favour us moving away from using exceptions for error handling (we shouldn't do that now though). As they can propagate through multiple layers of the stack and it makes the code hard to reason about (probably at the expensive of more lines of code).

libs/utilities/src/main/kotlin/net/corda/utilities/Either.kt
yift-r3: Arrow has [`map`](https://old.arrow-kt.io/docs/apidocs/arrow-core/arrow.core/-either/map.html) and [`mapLeft`](https://old.arrow-kt.io/docs/apidocs/arrow-core/arrow.core/-either/map-left.html). Scala [`map`](https://www.scala-lang.org/api/2.13.6/scala/util/Either.html#map[B1](f:B=%3EB1):scala.util.Either[A,B1]) is also right only. Rust has [`map_laft`](https://docs.rs/either/latest/either/enum.Either.html#method.map_left) and [`map_right`](https://docs.rs/either/latest/either/enum.Either.html#method.map_right). I don't see a `mapIf...` is used in Either monad.  Maybe it should be `mapLeft` and a simple `map` to map the right side?

williamvigorr3: I don't mind (do what you think is best).

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/SuspensionActivationEntityOperations.kt
```diff
@@ -69,21 +70,27 @@ internal class SuspensionActivationEntityOperations(
...
-                newStatus,
-                now,
-                currentMemberInfo.memberContext,
-                currentMemberInfo.memberSignatureKey,
-                currentMemberInfo.memberSignatureContent,
-                currentMemberInfo.memberSignatureSpec,
-                serializedMgmContext,
-                updatedSerial
+        try {
+            em.merge(
```
williamvigorr3: According to the EntityManager Java doc only `em.find` can throw a `PessimisticLockException`. I also think we should do the same thing if the GroupParameters conflict (although only inside `ActivateMemberHandler` and `SuspendMemberHandler`).

yift-r3: Yes, sorry, that is a lot more reasonable.

williamvigorr3: I think you missed the `GroupParameters` part out.

williamvigorr3 (dismissed):
LGTM

dimosr (approved):

PR #3747, opened 2023-04-27
ES-175: remove external publishing for simulator and it's dependencies

ronanbrowne:
Use of the `simulator` module in CSDE brought in multiple runtime dependencies which over the course of development required us to make several corda-runtime-os jars publicly available. 

With the retirement of simulator, this should be corrected.

- Remove `releasable = true` extensions block from these modules, this property signals to our internal publishing gradle plugin to pick up these modules for external publishing.
- Remove `corda.javadoc-generation` from these modules also, which is a BuildSrc plugin that controls JavaDoc generation this was only added as a requirement of Maven Central upload, now that we are no longer externally releasing these modules this can also be removed, this will have an added benefit of less build activity needed on CI runs.

### DIFF
--- a/libs/serialization/serialization-encoding/build.gradle
+++ b/libs/serialization/serialization-encoding/build.gradle
@@ -4,10 +4,6 @@ plugins {
     id 'corda.javadoc-generation'
 }
 
-ext {
-    releasable = true
-}
-
 description 'Corda Serialization encoding'
 
 configurations {
--- a/libs/configuration/configuration-core/build.gradle
+++ b/libs/configuration/configuration-core/build.gradle
@@ -1,11 +1,6 @@
 plugins {
     id 'corda.common-library'
     id 'corda.common-publishing'
-    id 'corda.javadoc-generation'
-}
-
-ext {
-    releasable = true
 }
 
 description 'Configuration Library'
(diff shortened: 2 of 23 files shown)


ronanbrowne:
@driessamyn FYI

yarivus (approved):
LGTM ...
I find it confusing that all the libs/*/* belong to simulator implicitly ... 
Or ... maybe they don't and you've removed them as a side effect ?

ronanbrowne:
> LGTM ... I find it confusing that all the libs/_/_ belong to simulator implicitly ... Or ... maybe they don't and you've removed them as a side effect ?

These were on the transitive dependency graph of the runtime simulator. Due to how dependencies were handle within the simulator module

driessamyn (approved):
LGTM

libs/serialization/serialization-encoding/build.gradle
```diff
@@ -4,10 +4,6 @@ plugins {
     id 'corda.javadoc-generation'
```
charlieR3, line 4: Out of the network team review scope but just noticed every other build file which has `releasable` removed also removed `id 'corda.javadoc-generation'`. Should it have been removed here also?

charlieR3 (approved):
LGTM for network team ownership. (left one comment on a file outside that scope)

libs/serialization/serialization-encoding/build.gradle
ronanbrowne: Thanks @charlieR3 I missed that, It can be yes however having it there has no functional impact for the core purpose of this story (removing publishing). 

I will make a follow-up PR to remove this as I want to get this in to unblock a release.

PR #3755, opened 2023-04-28
CORE-13137 Additional txn serialization test cases

relyafi:
This PR adds additional test cases to confirm the correct behavior of our transaction serialization, specifically:

- `WireTransaction` can be serialized within a sandbox, supplementing existing integration tests that do this outside of a sandbox context.
- `ConsensualLedgerTransaction` cannot be serialized, as it never makes sense to do this.


components/virtual-node/sandbox-group-context-service/src/integrationTest/kotlin/net/corda/sandboxgroupcontext/test/AMQPSerializationTest.kt
```diff
@@ -64,6 +70,18 @@ class AMQPSerializationTest {
...
+    @InjectService(timeout = TIMEOUT_MILLIS)
+    lateinit var merkleTreeProvider: MerkleTreeProvider
+
+    @InjectService(timeout = TIMEOUT_MILLIS)
+    lateinit var digestService: DigestService
+
+    @InjectService(timeout = TIMEOUT_MILLIS)
+    lateinit var jsonMarshallingService: JsonMarshallingService
+
+    @InjectService(timeout = TIMEOUT_MILLIS)
```
vlajos, line 82: If I remember right
`jsonValidator = sandboxGroupContext.getSandboxSingletonService()`
is the correct way to get access to the sandbox specific services.
https://github.com/corda/corda-runtime-os/blob/release/os/5.0/testing/ledger/ledger-common-base-integration-test/src/main/kotlin/net/corda/ledger/common/integration/test/CommonLedgerIntegrationTest.kt#L95

Also that one needs one less services to create an example:
```
wireTransaction = wireTransactionFactory.createExample(jsonMarshallingService, jsonValidator)
```

components/virtual-node/sandbox-group-context-service/src/integrationTest/kotlin/net/corda/sandboxgroupcontext/test/AMQPSerializationTest.kt
```diff
@@ -200,6 +218,38 @@ class AMQPSerializationTest {
         }
     }
 
+    @ParameterizedTest
+    @EnumSource(SandboxGroupType::class)
+    @DisplayName("Serialize & Deserialize WireTransaction: {0}")
+    fun testSerializeAndDeserializeWireTransaction(sandboxGroupType: SandboxGroupType) {
```
vlajos, line 224: `WireTransactionSerializerTest` does something similar as well.
But that one uses different `SerializatioFactory`s from the same `SandboxGroupContext`.
So this new one may have a better isolation.

relyafi: Yeah, I think there is overlap with `WireTransactionAMQPSerializerTest` too, but I don't think either of those are OSGi tests running in real sandboxes, whereas the new test does (and will run in each type of sandbox)?

vlajos: If the old one is worse from all aspects, probably it could be removed.

components/virtual-node/sandbox-group-context-service/src/integrationTest/kotlin/net/corda/sandboxgroupcontext/test/AMQPSerializationTest.kt
relyafi: Unfortunately it appears that using `getSandboxSingletonService` doesn't work, it causes the following error:

```
java.lang.IllegalStateException: net.corda.v5.application.marshalling.JsonMarshallingService service missing from sandbox
	at net.corda.sandboxgroupcontext.SandboxGroupContextExtensionsKt.getSandboxSingletonService(SandboxGroupContextExtensions.kt:41)
	at net.corda.sandboxgroupcontext.test.AMQPSerializationTest.testSerializeAndDeserializeWireTransaction$lambda$10(AMQPSerializationTest.kt:250)
	at net.corda.sandboxgroupcontext.test.VirtualNodeService.withSandbox(VirtualNodeService.kt:115)
	at net.corda.sandboxgroupcontext.test.AMQPSerializationTest.testSerializeAndDeserializeWireTransaction(AMQPSerializationTest.kt:219)
```

I'm guessing this is because this is running in a real sandbox, and these are not loaded into the context

components/virtual-node/sandbox-group-context-service/src/integrationTest/kotlin/net/corda/sandboxgroupcontext/test/AMQPSerializationTest.kt
relyafi: I've removed the old integration test, but have left the other (unit) test in place as that's possibly still useful as a fast check.

vlajos: Thanks.

vlajos (dismissed):

vlajos (approved):
LGTM

seanbrereton (approved):
LGTM (build file)

driessamyn (approved):

components/virtual-node/sandbox-group-context-service/src/integrationTest/kotlin/net/corda/sandboxgroupcontext/test/AMQPSerializationTest.kt
chrisr3: @vlajos is correct - `@InjectService` can only inject _platform_ services, and _cannot_ work for sandbox services.

FWIW I didn't intend for this particular OSGi test to start testing whether _every_ class can be AMQP-serialized, and particularly so for modules which already have OSGi tests of their own.

PR #3757, opened 2023-04-28
CORE-11162: Remove source and destination from inbound unauthenticated messages

yift-r3:
API pull request in https://github.com/corda/corda-api/pull/1079.

This pull request split the unauthenticated messages into inbound (without source and destination) and outbound (with source and destination).

The link manager will convert the outbound to inbound, and the Gateway will only be aware of the inbound messages.

The registration request handler will get the MGM from the request header and the member from the registration context (and not from the message header).

It was tested by registering using a single cluster and multiple clusters.

### DIFF
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
@@ -16,7 +16,7 @@ import net.corda.data.p2p.MessageAck
 import net.corda.data.p2p.SessionPartitions
 import net.corda.data.p2p.app.AppMessage
 import net.corda.data.p2p.app.AuthenticatedMessage
-import net.corda.data.p2p.app.UnauthenticatedMessage
+import net.corda.data.p2p.app.InboundUnauthenticatedMessage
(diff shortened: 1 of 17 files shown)


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -309,9 +295,9 @@ internal class InboundMessageProcessor(
         }
     }
 
-    private fun recordInboundMessagesMetric(message: UnauthenticatedMessage) {
+    private fun recordInboundMessagesMetric(message: InboundUnauthenticatedMessage) {
         message.header.let {
-            recordInboundMessagesMetric(it.source.x500Name, it.destination.x500Name, it.source.groupId,
+            recordInboundMessagesMetric("***", "***", "***",
```
dimosr: should we make these fields optional (source, dest, group)  and avoid emitting any tags for unauthenticated messages completely?

dimosr (commented):
LGTM - just a minor comment on metrics

dimosr (dismissed):

vkolomeyko (approved):
LGTM

dimosr (approved):

PR #3758, opened 2023-04-28
CORE-12539 - Jetbrains annotations version update

driessamyn:


vkolomeyko (approved):
LGTM

It is quiet a big step!

driessamyn:
> LGTM
> 
> It is quiet a big step!

Yes, but I had a look through the history. It's mostly new annotations that have been added since, so figured it should be ok.

PR #3762, opened 2023-04-28
CORE-13157 - Upgrade dependencies used by p2p

dimosr:
Upgrade `com.lmax:disruptor` and `io.netty:netty-codec-http`.


driessamyn (approved):
LGTM

PR #3763, opened 2023-04-28
CORE-12543: REST 3rd party dependencies, part 3

vkolomeyko:
- Add a comment on `jettyVersion`
- Upgrade `jcip-annotations`

### DIFF
--- a/libs/rest/rest-server-impl/build.gradle
+++ b/libs/rest/rest-server-impl/build.gradle
@@ -32,7 +32,6 @@ dependencies {
         }
     }
     implementation "com.nimbusds:oauth2-oidc-sdk:$nimbusVersion"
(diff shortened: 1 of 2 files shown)


libs/rest/rest-server-impl/build.gradle
```diff
@@ -32,7 +32,7 @@ dependencies {
         }
     }
     implementation "com.nimbusds:oauth2-oidc-sdk:$nimbusVersion"
-    implementation "org.apache.servicemix.bundles:org.apache.servicemix.bundles.jcip-annotations:1.0_1"
+    implementation "org.apache.servicemix.bundles:org.apache.servicemix.bundles.jcip-annotations:1.0_2"
```
driessamyn: Should we move this in variable in the properties file too?

filipesoliveira (approved):

driessamyn (approved):
LGTM

PR #3764, opened 2023-04-28
CORE-13158 Remove MGM CPB

charlieR3:
Removes the sample MGM CPB since a CPB is no longer required to build a CPI. Removing the CPB means that a network operator will not need to check out source code to build an MGM CPB, nor do we need to consider publishing an MGM CPB artifact. This also simplifies the network set up.


yift-r3 (approved):

ronanbrowne (approved):
LGTM

PR #3766, opened 2023-04-28
CORE-12501 JsonLayout -> JsonTemplateLayout

davidcurrie:
Switches JSON format logging from using the deprecated `JSONLayout` to `JsonTemplateLayout` with the `JsonLayout` template. On the whole, the output should be identical but the way in which exceptions are serialized changes which we hope will address issues with the use of reflection and the security manager.

The PR is made infinitely more complicated by the fact that the `log4j-layout-template-json` dependency brings with it a second `META-INF/org/apache/logging/log4j/core/config/plugins/Log4j2Plugins.dat` file. Normally Log4J would expect to read both files from their respective JARs but when we combine them into a single worker JAR, it only sees one. The `mergeLog4j2Plugins` task is therefore added to merge together the two files into one using Log4J's `PluginCache` class.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -11,6 +11,7 @@ def artifactoryContextUrl = constants.getProperty('artifactoryContextUrl')
 def jibCoreVersion = constants.getProperty('jibCoreVersion')
 def internalPublishVersion = constants.getProperty('internalPublishVersion')
 def gradleEnterpriseVersion = constants.getProperty('gradleEnterpriseVersion')
+def log4jVersion = constants.getProperty('log4jVersion')
 
 dependencies {
     implementation "biz.aQute.bnd:biz.aQute.bnd.gradle:$bndVersion"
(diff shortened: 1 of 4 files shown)


yarivus (approved):
LGTM

simon-johnson-r3 (approved):
I've tested this fixes CORE-12974
With
```
    override fun call(requestBody: ClientRequestBody): String {
        try {
            throw CordaRuntimeException("some exception message")
        } catch (ex: Exception) {
            log.info("failed: ", ex)
        }
```
I now see in the JSON log:
```
{
    "instant": {
        "epochSecond": 1683022035,
        "nanoOfSecond": 37000000
    },
    "thread": "single-threaded-scheduled-executor-1",
    "level": "INFO",
    "loggerName": "com.r3.corda.testing.enterprise.smoketests.MessageFlow",
    "message": "failed: ",
    "thrown": {
        "message": "some exception message",
        "name": "net.corda.v5.base.exceptions.CordaRuntimeException",
        "extendedStackTrace": [{
            "class": "com.r3.corda.testing.enterprise.smoketests.MessageFlow",
            "method": "call",
            "file": "MessageFlows.kt",
            "line": 36
```
whereas before you saw no log. So it seems to have resolved this problem.

CornelR3 (approved):
LGTM

PR #3768, opened 2023-04-28
CORE-9483 Enabling sending of primitive types across sessions

ben-millar:
This PR addresses [CORE-9483](https://r3-cev.atlassian.net/browse/CORE-9483).

## Background:
Due to some errors observed in Corda 4 when serialising primitive types in flow sessions, the function `requireBoxedType(type: Class<*>) { require(!type.isPrimitive) }` was added to a number of classes to assert that a passed parameter was **not** a primitive type. Today, this function is causing failures when we try to pass integers through FlowMessaging as Kotlin autoboxing is converting the **boxed** `java.lang.Integer`/`kotlin.Int` to the primitive `int` which then fails the `requireBoxedType()` check.

## Findings:
When we remove the `requireBoxedType()` check, the sending of primitive types fails for two reasons:

### Autoboxing
It's important to note that, unlike Java, Kotlin has no primitive types; instead, all primitive types (e.g., boolean, float, int) are represented by boxed classes. When Kotlin code is compiled to JVM bytecode, all Kotlin variables are compiled to Java variables and it is up to the compiler to determine whether they become the boxed (class) or unboxed (primitive) equivalent.

To send data between flows, we use the `send()`, `receive()`, and `sendAndReceive()` functions on the FlowSession class (or equivalent functions in FlowMessaging). The `send(payload: Any)` function accepts a payload of `Any` type, serialises it using an AMQP serialiser, and sends the resulting bytes to the designated counterparty.

To receive data, we use the `receive(receiveType: Class<R>)` function, which takes a type token `receiveType` representing the expected class of the received data. The function waits for a send event from the counterparty flow and attempts to deserialise the received payload to the specified class type.

The issue here is that the `Any` parameter in `send()` will cause passed arguments to be implicit cast to their boxed representation at runtime, meaning that a passed `int` will be cast to a `Java.lang.Integer` before being serialised and sent. The generic type parameter `Class<R>` in `send()` does not do such casting, and so upon receiving the serialised `Java.lang.Integer` at runtime we attempt to deserialise it as a primitive `int` causing the following exception:
```
java.io.NotSerializableException: Internal deserialization failure: java.lang.ClassCastException: Cannot cast java.lang.Integer to int 
```
[java.io.NotSerializableException.txt](https://github.com/corda/corda-runtime-os/files/11450822/java.io.NotSerializableException.txt)

### Kryo Serialization / Class Loading
When we call `receive()` on a flow, we persist a `FlowIORequest` into the flow fiber and suspend the fiber, waiting for the expected message to be received. This suspension was triggering a Kryo serialisation/deserialisation of the flow which was tripping up when trying to load the `int` 'class' from the system class loader as there is no class to load for an `int` primitive, resulting in the following:
```
com.esotericsoftware.kryo.KryoException: net.corda.sandbox.SandboxException: Class int was not from a bundle, and could not be found in the system classloader.
```
[KryoException- net.corda.sandbox.SandboxException.txt](https://github.com/corda/corda-runtime-os/files/11450817/KryoException-.net.corda.sandbox.SandboxException.txt)

### Autoboxing
We can assume that any primitive value will be autoboxed before serialization, and deserialize it as such on receipt with the following check:
```
return if (receiveType.isPrimitive) {
    deserializeReceivedPayload(received, receiveType.kotlin.javaObjectType)
} else {
    deserializeReceivedPayload(received, receiveType)
}
```
> (Note: We could call the deserialiser with `receiveType.kotlin.javaObjectType` in all cases – it would simply be a no-op on boxed classes – but I felt this conditional check made the intent more clear).

This check makes sure we pass the class type of the boxed representation to the deserialiser. The return type, `R`, will ensure that the returned value is implicit cast to whichever representation the `receive()` function was called with, either boxed or primitive. 

We could also solve this by modifying the API of `send()` and `sendAndReceive()` to prevent autoboxing, but this solution doesn't require any modifications to the API.

### Kryo Serialization / Class Loading
Flow class loading is delegated through the `GroupSandboxImpl`; I've modified this delegator so that when it is called with a primitive, e.g. `int` or `float`, it will directly return the Java `Class` wrapper for that primitive from a mapping:
```
private val primitiveTypes = mapOf(
    "boolean" to Boolean::class.javaPrimitiveType,
    "byte" to Byte::class.javaPrimitiveType,
    "short" to Short::class.javaPrimitiveType,
    ...
)
```
### Miscellanious
Special handling was also added to `DeserializationInput` for `char` types, as they go through a widening conversion to an `Integer` type in-transit. More info on this can be found in comments below.

## Testing
As of now, I've written unit tests for AMQP serialisation/deserialisation of primitive types, and I've performed manual E2E testing of flows with primitive types. I'll add automated E2E and integration testing of the new behaviour.

[CORE-9483]: https://r3-cev.atlassian.net/browse/CORE-9483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/PersistenceServiceImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/PersistenceServiceImpl.kt
@@ -1,6 +1,13 @@
 package net.corda.flow.application.persistence
 
-import net.corda.flow.application.persistence.external.events.*
+import net.corda.flow.application.persistence.external.events.FindExternalEventFactory
+import net.corda.flow.application.persistence.external.events.FindParameters
+import net.corda.flow.application.persistence.external.events.MergeExternalEventFactory
+import net.corda.flow.application.persistence.external.events.MergeParameters
+import net.corda.flow.application.persistence.external.events.PersistExternalEventFactory
+import net.corda.flow.application.persistence.external.events.PersistParameters
+import net.corda.flow.application.persistence.external.events.RemoveExternalEventFactory
+import net.corda.flow.application.persistence.external.events.RemoveParameters
 import net.corda.flow.application.persistence.query.PagedQueryFactory
 import net.corda.flow.external.events.executor.ExternalEventExecutor
 import net.corda.sandbox.type.UsedByFlow
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/PersistenceServiceImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/PersistenceServiceImpl.kt
@@ -1,14 +1,6 @@
 package net.corda.flow.application.persistence
 
-import java.nio.ByteBuffer
-import net.corda.flow.application.persistence.external.events.FindExternalEventFactory
-import net.corda.flow.application.persistence.external.events.FindParameters
-import net.corda.flow.application.persistence.external.events.MergeExternalEventFactory
-import net.corda.flow.application.persistence.external.events.MergeParameters
-import net.corda.flow.application.persistence.external.events.PersistExternalEventFactory
-import net.corda.flow.application.persistence.external.events.PersistParameters
-import net.corda.flow.application.persistence.external.events.RemoveExternalEventFactory
-import net.corda.flow.application.persistence.external.events.RemoveParameters
+import net.corda.flow.application.persistence.external.events.*
 import net.corda.flow.application.persistence.query.PagedQueryFactory
 import net.corda.flow.external.events.executor.ExternalEventExecutor
 import net.corda.sandbox.type.UsedByFlow
@@ -22,6 +14,7 @@ import org.osgi.service.component.annotations.Activate
 import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
 import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
+import java.nio.ByteBuffer
 
 @Suppress("TooManyFunctions")
 @Component(service = [ PersistenceService::class, UsedByFlow::class ], scope = PROTOTYPE)
@@ -36,7 +29,6 @@ class PersistenceServiceImpl @Activate constructor(
 
     @Suspendable
     override fun <R : Any> find(entityClass: Class<R>, primaryKey: Any): R? {
-        requireBoxedType(entityClass)
         return wrapWithPersistenceException {
             externalEventExecutor.execute(
                 FindExternalEventFactory::class.java,
@@ -47,7 +39,6 @@ class PersistenceServiceImpl @Activate constructor(
 
     @Suspendable
     override fun <R : Any> find(entityClass: Class<R>, primaryKeys: List<*>): List<R> {
-        requireBoxedType(entityClass)
         return wrapWithPersistenceException {
             externalEventExecutor.execute(
                 FindExternalEventFactory::class.java,
@@ -58,13 +49,11 @@ class PersistenceServiceImpl @Activate constructor(
 
     @Suspendable
     override fun <R : Any> findAll(entityClass: Class<R>): PagedQuery<R> {
-        requireBoxedType(entityClass)
         return pagedQueryFactory.createPagedFindQuery(entityClass)
     }
 
     @Suspendable
     override fun <R : Any> merge(entity: R): R? {
-        requireBoxedType(entity.javaClass)
         return wrapWithPersistenceException {
             externalEventExecutor.execute(
                 MergeExternalEventFactory::class.java,
@@ -76,7 +65,6 @@ class PersistenceServiceImpl @Activate constructor(
     @Suspendable
     override fun <R : Any> merge(entities: List<R>): List<R> {
         return if (entities.isNotEmpty()) {
-            requireBoxedType(entities)
             val mergedEntities = wrapWithPersistenceException {
                 externalEventExecutor.execute(
                     MergeExternalEventFactory::class.java,
@@ -96,7 +84,6 @@ class PersistenceServiceImpl @Activate constructor(
 
     @Suspendable
     override fun persist(entity: Any) {
-        requireBoxedType(entity.javaClass)
         wrapWithPersistenceException {
             externalEventExecutor.execute(
                 PersistExternalEventFactory::class.java,
@@ -108,7 +95,6 @@ class PersistenceServiceImpl @Activate constructor(
     @Suspendable
     override fun persist(entities: List<*>) {
         if (entities.isNotEmpty()) {
-            requireBoxedType(entities)
             wrapWithPersistenceException {
                 externalEventExecutor.execute(
                     PersistExternalEventFactory::class.java,
@@ -120,7 +106,6 @@ class PersistenceServiceImpl @Activate constructor(
 
     @Suspendable
     override fun remove(entity: Any) {
-        requireBoxedType(entity.javaClass)
         wrapWithPersistenceException {
             externalEventExecutor.execute(
                 RemoveExternalEventFactory::class.java,
@@ -132,7 +117,6 @@ class PersistenceServiceImpl @Activate constructor(
     @Suspendable
     override fun remove(entities: List<*>) {
         if (entities.isNotEmpty()) {
-            requireBoxedType(entities)
             wrapWithPersistenceException {
                 externalEventExecutor.execute(
                     RemoveExternalEventFactory::class.java,
@@ -147,23 +131,7 @@ class PersistenceServiceImpl @Activate constructor(
         queryName: String,
         entityClass: Class<T>
     ): ParameterizedQuery<T> {
-        requireBoxedType(entityClass)
         return pagedQueryFactory.createNamedParameterizedQuery(queryName, entityClass)
-    }
-
-    /**
-     * Required to prevent class cast exceptions during AMQP serialization of primitive types.
-     */
-    private fun requireBoxedType(type: Class<*>) {
-        require(!type.isPrimitive) { "Cannot perform persistence operation on primitive type ${type.name}" }
-    }
-
-    private fun requireBoxedType(objects: List<*>) {
-        for (obj in objects) {
-            if (obj != null) {
-                requireBoxedType(obj.javaClass)
-            }
-        }
     }
 
     private fun serialize(payload: Any): ByteBuffer {
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FlowMessagingImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FlowMessagingImpl.kt
@@ -7,6 +7,7 @@ import net.corda.flow.application.serialization.SerializationServiceInternal
 import net.corda.flow.application.sessions.FlowSessionInternal
 import net.corda.flow.application.sessions.SessionInfo
 import net.corda.flow.application.sessions.factory.FlowSessionFactory
+import net.corda.flow.application.sessions.utils.SessionUtils.verifySessionStatusNotErrorOrClose
 import net.corda.flow.application.versioning.impl.sessions.VersionReceivingFlowSession
 import net.corda.flow.application.versioning.impl.sessions.VersionSendingFlowSession
 import net.corda.flow.fiber.FlowFiber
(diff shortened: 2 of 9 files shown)


components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
```diff
@@ -100,21 +100,33 @@ class FlowSessionImpl(
     @Suspendable
     override fun <R : Any> sendAndReceive(receiveType: Class<R>, payload: Any): R {
         verifySessionStatusNotErrorOrClose(sourceSessionId, flowFiberService)
-        requireBoxedType(receiveType)
         val request = FlowIORequest.SendAndReceive(mapOf(getSessionInfo() to serialize(payload)))
         val received = fiber.suspend(request)
         setSessionConfirmed()
-        return deserializeReceivedPayload(received, receiveType)
+
+        // We could call `.kotlin.javaObjectType` on non-primitive types too, and it would have no effect,
```
lankydan: Extract this into a method and remove the comment imo.

lankydan (commented):
Seems reasonable.

Probably not one for 5.0 because this change looks backwards compatible.

Also worth @chrisr3 or @josephzunigadaly having a quick look due to serialization/sandbox change.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
ben-millar: Makes sense, I'll do that

ben-millar:
Currently, one of the new unit tests I added to test `char` serialisation is failing. After much hair pulling I've traced the root cause down to the Data library in Java ([DOCS](https://qpid.apache.org/releases/qpid-proton-j-0.33.8/api/org/apache/qpid/proton/codec/Data.html)). The `putChar()` and `getChar()` functions in this library use integer values (32-bit) rather than chars (16-bit); this widening is done to allow for negative values representing EOF characters, but this is causing a cast to fail in [DeserializationInput::doReadObject()](https://github.com/corda/corda-runtime-os/blob/release/os/5.1/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/DeserializationInput.kt#L145)

```
Internal deserialization failure: java.lang.ClassCastException: Cannot cast java.lang.Integer to java.lang.Character
```

Currently taking a look at the best way to fix this.

chrisr3:
Derp, is someone suggesting that AMQP has been unable to handle primitive Java types for the past 7 years? :thinking:

`net.corda.internal.serialization.amqp.standard.AMQPPrimitiveSerializer` anyone?

chrisr3:
> It's important to note that, unlike Java, Kotlin has no primitive types; instead, all primitive types (e.g., boolean, float, int) are represented by boxed classes.

Not quite. It's actually like this:
- `Int` -> `int`
- `Int?` -> `java.lang.Integer`
- `Short` -> `short`
- `Short?` -> `java.lang.Short`
- `Byte` -> `byte`
- `Byte?` -> `java.lang.Byte`
- `Char` -> `char`
- `Char?` -> `java.lang.Character`
- `Float` -> `float`
- `Float?` -> `java.lang.Float`
- `Double` -> `double`
- `Double?` -> `java.lang.Double`

The point is that Java primitives can _never_ be `null`, and so are represented in Kotlin as the non-nullable versions. However, if `null` is allowed then Kotlin must use the boxed type instead.

chrisr3 (changes requested):
AMQP serialization has _always_ supported Java primitive types - see [AMQPPrimitiveSerializer](https://github.com/corda/corda-runtime-os/blob/release/os/5.1/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/standard/AMQPPrimitiveSerializer.kt). Consequently, there should not be any need to touch `DeserializationInput`.

Seriously, what's going on?

ben-millar:
> AMQP serialization has _always_ supported Java primitive types - see [AMQPPrimitiveSerializer](https://github.com/corda/corda-runtime-os/blob/release/os/5.1/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/standard/AMQPPrimitiveSerializer.kt). Consequently, there should not be any need to touch `DeserializationInput`.
> 
> Seriously, what's going on?

@chrisr3 
AMQP serialization has always supported Java primitive types (Not so much 'handle' as return essentially untouched); the purpose of this ticket is to enable our **flow sessions** to handle Java primitive types. 

_However_, 

this work uncovered the fact that our implementation of `DeserializationInput` could not handle `char` types, because they were being cast to integers by the `AMQPPrimitiveSerializer` (for valid reasons).

The functional changes to DeserializationInput simply add handling (narrowing conversion back to `char`) for this ([HERE](https://github.com/corda/corda-runtime-os/pull/3768/files#diff-021102d7da23c7b21bd4e30bdfa4027ab6a6b9466f9a0c7f309cbb8c0894fc0eR333)) and, while doing that, I split out the `readObject()` function so it wouldn't need `@Suppress("NestedBlockDepth", "ComplexMethod")` annotations to pass the linter.

chrisr3:
> this work uncovered the fact that our implementation of `DeserializationInput` could not handle `char` types, because they were being cast to integers by the `AMQPPrimitiveSerializer` (for valid reasons).

I am _deeply sceptical_ of any changes to AMQP's handing of primitive types. Our `SerializationOutputTests` class may not _explicitly_ test `char` atm, but that's easily fixed:
```diff
diff --git a/libs/serialization/serialization-amqp/src/test/kotlin/net/corda/internal/serialization/amqp/SerializationOutputTests.kt b/libs/serialization/serialization-amqp/src/test/kotlin/net/corda/internal/serialization/amqp/SerializationOutputTests.kt
index a442695e42..fc93e21cda 100644
--- a/libs/serialization/serialization-amqp/src/test/kotlin/net/corda/internal/serialization/amqp/SerializationOutputTests.kt
+++ b/libs/serialization/serialization-amqp/src/test/kotlin/net/corda/internal/serialization/amqp/SerializationOutputTests.kt
@@ -100,6 +100,9 @@ class SerializationOutputTests {
     @CordaSerializable
     data class TestBoolean(val b: Boolean)
 
+    @CordaSerializable
+    data class TestChar(val c: Char)
+
     @CordaSerializable
     interface FooInterface {
         val pub: Int
@@ -332,6 +335,12 @@ class SerializationOutputTests {
         serdes(obj)
     }
 
+    @Test
+    fun `test char`() {
+        val obj = TestChar('?')
+        serdes(obj)
+    }
+
     @Test
     fun `test foo implements`() {
         val obj = FooImplements("Hello World!", 123)
```
And it passes, which suggests to me that `AMQPPrimitiveSerializer` is working correctly.

To confirm that `TestChar` really _is_ using a primitive type:
```
  private final char c;
    descriptor: C
    flags: (0x0012) ACC_PRIVATE, ACC_FINAL

  public net.corda.internal.serialization.amqp.SerializationOutputTests$TestChar(char);
    descriptor: (C)V
    flags: (0x0001) ACC_PUBLIC
    Code:
      stack=2, locals=2, args_size=2
         0: aload_0
         1: invokespecial #10                 // Method java/lang/Object."<init>":()V
         4: aload_0
         5: iload_1
         6: putfield      #13                 // Field c:C
         9: return
      LineNumberTable:
        line 103: 0
        line 104: 4
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0      10     0  this   Lnet/corda/internal/serialization/amqp/SerializationOutputTests$TestChar;
            0      10     1     c   C
    MethodParameters:
      Name                           Flags
      c
```

chrisr3:
> The functional changes to DeserializationInput simply add handling (narrowing conversion back to `char`) for this ([HERE](https://github.com/corda/corda-runtime-os/pull/3768/files#diff-021102d7da23c7b21bd4e30bdfa4027ab6a6b9466f9a0c7f309cbb8c0894fc0eR333))

Are you trying to serialize _naked_ primitive types with AMQP? I don't believe we have **ever** supported doing that - @mnesbit?

ben-millar:
> > The functional changes to DeserializationInput simply add handling (narrowing conversion back to `char`) for this ([HERE](https://github.com/corda/corda-runtime-os/pull/3768/files#diff-021102d7da23c7b21bd4e30bdfa4027ab6a6b9466f9a0c7f309cbb8c0894fc0eR333))
> 
> Are you trying to serialize _naked_ primitive types with AMQP? I don't believe we have **ever** supported doing that - @mnesbit?

Actually, you've uncovered a good point. So, honest-to-goodness primitive types are handled as of today [HERE](https://github.com/corda/corda-runtime-os/blob/release/os/5.1/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/DeserializationInput.kt#L214) – we simply return them... Or so it may seem. In actuality, the fact that the `obj` parameter is of type `Any` will cast any primitive objects to their boxed equivalent, so this will never be hit.

ben-millar:
> ```
>@CordaSerializable
>data class TestChar(val c: Char)
>
>@Test
>fun `test char`() {
>    val obj = TestChar('?')
>    serdes(obj)
>}
>```

Re: this test, I mentioned it offline (but it's worth noting here for future reference), by wrapping the `char` in a dataclass we're invoking the `ComposableObjectSerializer` rather than the `AMQPPrimitiveSerializer`. This, in turn, invokes the `ComposableTypePropertySerializer` on the `char` which delegates to `AMQPCharProperty[Write/Read]Strategy` [HERE](https://github.com/corda/corda-runtime-os/blob/release/os/5.1/libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/ComposableTypePropertySerializer.kt#L221) which do the same sort of casting as I've added for the failing case.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/PersistenceServiceImpl.kt
```diff
@@ -36,7 +36,6 @@ class PersistenceServiceImpl @Activate constructor(
 
     @Suspendable
     override fun <R : Any> find(entityClass: Class<R>, primaryKey: Any): R? {
-        requireBoxedType(entityClass)
```
chrisr3, line 39: But how would you declare _any_ native Java or Kotlin type to be a JPA `@Entity` in the first place? :confused::exploding_head:

I'd have thought something like:
```kotlin
if (!entityClass.isAnnotationPresent(Entity::class.java)) {
    throw IllegalArgumentException("$entityClass is not a JPA entity")
}
```
would be more appropriate, possibly with
```kotlin
requireCordaSerializable(entityClass)
```
thrown in.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FlowMessagingImpl.kt
```diff
@@ -55,8 +55,6 @@ class FlowMessagingImpl @Activate constructor(
 
     @Suspendable
     override fun <R : Any> receiveAll(receiveType: Class<out R>, sessions: Set<FlowSession>): List<R> {
-        requireBoxedType(receiveType)
```
chrisr3, line 58: I'd actually prefer
```kotlin
requireCordaSerializable(receiveType)
```
here.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
```diff
@@ -100,21 +100,23 @@ class FlowSessionImpl(
     @Suspendable
     override fun <R : Any> sendAndReceive(receiveType: Class<R>, payload: Any): R {
         verifySessionStatusNotErrorOrClose(sourceSessionId, flowFiberService)
-        requireBoxedType(receiveType)
```
chrisr3, line 103: I'd prefer
```kotlin
requireCordaSerializable(receiveType)
requireCordaSerializable(payload::class.java)
```
here.

---

Or maybe not - see comment.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FlowMessagingImpl.kt
```diff
@@ -149,7 +145,6 @@ class FlowMessagingImpl @Activate constructor(
             return
         }
         val sessionPayload = payloadsPerSession.map { (session, payload) ->
-            requireBoxedType(payload::class.java)
```
chrisr3, line 152: Or maybe
```kotlin
requireCordaSerializable(payload::class.java)
```
?

chrisr3 (commented):
I think this PR has a deeper problem.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FlowMessagingImpl.kt
chrisr3: Actually, maybe I wouldn't after all. The problem is that `receiveType` may be an `interface` type, e.g. some kind of marker interface. The type of the actual _object_ received could be:
```kotlin
@CordaSerializable
data class Response(val message: String) : Marker
```
which _is_ `@CordaSerializable`, but `interface Marker` wouldn't know that :crying_cat_face:.

PR #3770, opened 2023-04-30
CORE-13168: Check whether the Kotlin metadata is compatible with the library.

chrisr3:
The Kotlin metadata library returns `null` if the class data it was asked to read is incompatible. Check for this explicitly and return an informative error message as required.

Ultimately, supporting multiple versions of Kotlin will require supporting multiple versions of the metadata library too.


seanbrereton (approved):
LGTM (gradle file only)

mnesbit (approved):
OK

driessamyn (approved):
LGTM

PR #3774, opened 2023-05-02
CORE-11807 stop and close should not be called together. Only close is required when wanting to stop and close.

LWogan:
the implementation of close is will post and internal stop event and then remove the coordinator from the registry.
stop() posts an internal stop event. 

Given that close and stop both post stop events there is no need to call both as this causes warnings. there is also no need to post DOWN as this is done already as part of close.

Additionall there is a race condition within the coordinator close logic. close() will post events to its own processor to process in order on a different thread, such as CloseCoordinator(). Then the coordinator will remove itself from the registry. This means it will usually remove itself from the registry before it processed the CloseCoordinator event fully. When it tries to update its status in the registry in CloseCoordinator handler it will fail and log a warning. Moved the registry removal to the CloseCoordinator handler


thiagoviana (dismissed):

ben-millar (dismissed):

ben-millar (approved):

ben-millar:
LGTM pending 🟢

emilybowe (approved):
LGTM

PR #3776, opened 2023-05-02
CORE-13137: Fix incorrect AMQP serialization test for WireTransaction.

chrisr3:
`@InjectService` can only inject a platform service. You must therefore acquire a reference to a sandbox service from the `SandboxGroupContext`, as otherwise the test becomes meaningless.

Note that `WireTransactionSerializer` requires a `WireTransactionFactory`, which only exists in `FLOW` and `PERSISTENCE` sandboxes. As does the `JsonValidator` service, as it happens. Therefore there is no point declaring `WireTransactionSerializer` as `UsedForVerification`, which means `testSerializeAndDeserializeWireTransaction` should only execute for `FLOW` and `PERSISTENCE` sandboxes.

The fact that this test used to pass for `VERIFICATION` sandboxes demonstrates my earlier point about incorrect usage of `@InjectService` making tests meaningless.

---
It appears that `VERIFICATION` sandboxes also require the following services:
- `WireTransactionFactory`
- `WireTransactionSerializer`
- `JsonMarshallingService`
- `JsonValidator`

as otherwise `VerificationRequestProcessorTest` fails. Declare all of these components as `UsedByVerification` so that `testSerializeAndDeserializeWireTransaction` can also pass for all sandbox types.

_It is critical that sandboxes use their own instance of `JsonMarshallingService` so that CPKs can be unloaded without leaking memory._ So declare these references as `PROTOTYPE_REQUIRED` to prevent them accidentally being satisfied by the platform's `JsonMarshallingService`.

### DIFF
--- a/components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/filtered/factory/FilteredTransactionFactoryImpl.kt
+++ b/components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/filtered/factory/FilteredTransactionFactoryImpl.kt
@@ -17,6 +17,7 @@ import net.corda.v5.serialization.SingletonSerializeAsToken
 import org.osgi.service.component.annotations.Activate
 import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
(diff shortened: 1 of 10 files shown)


vlajos (approved):
LGTM

driessamyn (approved):
LGTM

PR #3777, opened 2023-05-02
CORE-13256 Remove evolvability tags from transaction output.

blsemo:
The tags were added to UtxoOutputInfo and Consensual transactions, but they are completely redundant as they are part of the AMQP serialized data.


mnesbit:
They were intended to be a separate component group for partial merkle tree support, not for evolution

blsemo:
Can't do that - we will need the information for other reasons.

PR #3780, opened 2023-05-03
CORE-12911 Use only class bundle name for evolved classes

simon-johnson-r3:
Ultimately Corda 5 should use `Corda-CPK-Cordapp-Name` from the CPK metadata to identify a class type as being the same as another version of that class. See CORE-13259. This will be written in the evolvable tag. This change goes part way to supporting that in the evolvable tag.

Right now CorDapp developers cannot set this metadata, the gradle plugin always sets it to bundle symbolic name, so offering full support for `Corda-CPK-Cordapp-Name` is not yet possible. For Corda 5.0 we only use the class's bundle symbolic name to find sandboxes. Future versions of the platform will still write the class's bundle name and therefore be compatible with platforms running on 5.0.

We also write bundle symbolic name as `Corda-CPK-Cordapp-Name` in 5.0 meaning future versions of the platform reading that back will get a sensible answer - this is the same as the proposed default in future versions in fact.

### DIFF
--- a/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
+++ b/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
@@ -43,13 +43,14 @@ internal class SandboxGroupImpl(
         }
     }
 
-    override val metadata: SortedMap<Bundle, CpkMetadata> =
-        unmodifiableSortedMap(cpkSandboxes.associateTo(TreeMap()) { cpk ->
-            cpk.mainBundle to cpk.cpkMetadata
-        })
+    override val metadata: SortedMap<Bundle, CpkMetadata> = unmodifiableSortedMap(cpkSandboxes.associateTo(TreeMap()) { cpk ->
+        cpk.mainBundle to cpk.cpkMetadata
+    })
 
     override fun loadClassFromPublicBundles(className: String): Class<*>? {
-        val clazz = publicSandboxes.flatMap(Sandbox::publicBundles).filterNot(Bundle::isFragment)
+        val clazz = publicSandboxes
+            .flatMap(Sandbox::publicBundles)
+            .filterNot(Bundle::isFragment)
             .mapNotNullTo(LinkedHashSet()) { bundle ->
                 try {
                     bundle.loadClass(className)
@@ -74,9 +75,8 @@ internal class SandboxGroupImpl(
                 null
             }
         }.singleOrNull()
-            ?: throw SandboxException("Class $className was not found in any sandbox in the sandbox group.").withSuppressed(
-                suppressed
-            )
+            ?: throw SandboxException("Class $className was not found in any sandbox in the sandbox group.")
+                .withSuppressed(suppressed)
     }
 
     override fun <T : Any> loadClassFromMainBundles(className: String, type: Class<T>): Class<out T> {
--- a/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
+++ b/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
@@ -43,14 +43,13 @@ internal class SandboxGroupImpl(
         }
     }
 
-    override val metadata: SortedMap<Bundle, CpkMetadata> = unmodifiableSortedMap(cpkSandboxes.associateTo(TreeMap()) { cpk ->
-        cpk.mainBundle to cpk.cpkMetadata
-    })
+    override val metadata: SortedMap<Bundle, CpkMetadata> =
+        unmodifiableSortedMap(cpkSandboxes.associateTo(TreeMap()) { cpk ->
+            cpk.mainBundle to cpk.cpkMetadata
+        })
 
     override fun loadClassFromPublicBundles(className: String): Class<*>? {
-        val clazz = publicSandboxes
-            .flatMap(Sandbox::publicBundles)
-            .filterNot(Bundle::isFragment)
+        val clazz = publicSandboxes.flatMap(Sandbox::publicBundles).filterNot(Bundle::isFragment)
             .mapNotNullTo(LinkedHashSet()) { bundle ->
                 try {
                     bundle.loadClass(className)
@@ -75,8 +74,9 @@ internal class SandboxGroupImpl(
                 null
             }
         }.singleOrNull()
-            ?: throw SandboxException("Class $className was not found in any sandbox in the sandbox group.")
-                .withSuppressed(suppressed)
+            ?: throw SandboxException("Class $className was not found in any sandbox in the sandbox group.").withSuppressed(
+                suppressed
+            )
     }
 
     override fun <T : Any> loadClassFromMainBundles(className: String, type: Class<T>): Class<out T> {
@@ -110,20 +110,23 @@ internal class SandboxGroupImpl(
             }
 
             ClassType.CpkSandboxClass -> {
-                val sandbox = when (classTag) {
+                val (sandbox, bundleName) = when (classTag) {
                     is StaticTag -> cpkSandboxes.find { sandbox -> sandbox.cpkMetadata.fileChecksum == classTag.cpkFileHash }
+                        ?.let { Pair(it, classTag.classBundleName) }
+
                     is EvolvableTag -> {
                         cpkSandboxes.find {
-                            it.cpkMetadata.cpkId.signerSummaryHash == classTag.cpkSignerSummaryHash
-                                    && (it.cpkMetadata.cpkId.name == classTag.cordaCpkCordappName || // CPK given names match or
-                                        it.mainBundle.symbolicName == classTag.classBundleName) // symbolic names of class bundle match
-                        }
+                            it.cpkMetadata.cpkId.signerSummaryHash == classTag.cpkSignerSummaryHash &&
+                                    (it.cpkMetadata.cpkId.name == classTag.cordaCpkCordappName || // CPK given names match or
+                                    it.mainBundle.symbolicName == classTag.classBundleName) // symbolic names of class bundle match
+                        }?.let { Pair(it, it.mainBundle.symbolicName) } // only load evolvable classes from the main bundle
                     }
                 } ?: throw SandboxException(
                     "Class tag $serialisedClassTag did not match any sandbox in the sandbox group."
                 )
-                sandbox.loadClass(className, sandbox.mainBundle.symbolicName) ?: throw SandboxException(
-                    "Class $className could not be loaded from bundle ${classTag.classBundleName} in sandbox ${sandbox.id}."
+
+                sandbox.loadClass(className, bundleName) ?: throw SandboxException(
+                    "Class $className could not be loaded from bundle $bundleName in sandbox ${sandbox.id}."
                 )
             }
 
@@ -131,8 +134,7 @@ internal class SandboxGroupImpl(
                 publicSandboxes.asSequence().mapNotNull { publicSandbox ->
                     publicSandbox.loadClass(className, classTag.classBundleName)
                 }.firstOrNull() ?: throw SandboxException(
-                    "Class $className from bundle ${classTag.classBundleName} could not be loaded from any of the public " +
-                            "sandboxes."
+                    "Class $className from bundle ${classTag.classBundleName} could not be loaded from any of the public " + "sandboxes."
                 )
             }
         }
@@ -150,18 +152,15 @@ internal class SandboxGroupImpl(
      * bundle.
      */
     private fun getClassTag(klass: Class<*>, isStaticTag: Boolean): String {
-        val bundle = bundleUtils.getBundle(klass)
-            ?: return classTagFactory.createSerialisedTag(isStaticTag, null, null)
+        val bundle = bundleUtils.getBundle(klass) ?: return classTagFactory.createSerialisedTag(isStaticTag, null, null)
 
-        val publicSandbox =
-            publicSandboxes.find { sandbox -> sandbox.containsBundle(bundle) }
+        val publicSandbox = publicSandboxes.find { sandbox -> sandbox.containsBundle(bundle) }
         if (publicSandbox != null) {
             return classTagFactory.createSerialisedTag(isStaticTag, bundle, null)
         }
 
-        val cpkSandbox =
-            cpkSandboxes.find { sandbox -> sandbox.containsBundle(bundle) }
-                ?: throw SandboxException("Bundle ${bundle.symbolicName} was not found in the sandbox group or in a public sandbox.")
+        val cpkSandbox = cpkSandboxes.find { sandbox -> sandbox.containsBundle(bundle) }
+            ?: throw SandboxException("Bundle ${bundle.symbolicName} was not found in the sandbox group or in a public sandbox.")
         if (bundle in cpkSandbox.privateBundles && !isStaticTag) {
             throw SandboxException("Attempted to create evolvable class tag for cpk private bundle ${bundle.symbolicName}.")
         }
--- a/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
+++ b/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
@@ -152,15 +152,18 @@ internal class SandboxGroupImpl(
      * bundle.
      */
     private fun getClassTag(klass: Class<*>, isStaticTag: Boolean): String {
-        val bundle = bundleUtils.getBundle(klass) ?: return classTagFactory.createSerialisedTag(isStaticTag, null, null)
+        val bundle = bundleUtils.getBundle(klass)
+            ?: return classTagFactory.createSerialisedTag(isStaticTag, null, null)
 
-        val publicSandbox = publicSandboxes.find { sandbox -> sandbox.containsBundle(bundle) }
+        val publicSandbox =
+            publicSandboxes.find { sandbox -> sandbox.containsBundle(bundle) }
         if (publicSandbox != null) {
             return classTagFactory.createSerialisedTag(isStaticTag, bundle, null)
         }
 
-        val cpkSandbox = cpkSandboxes.find { sandbox -> sandbox.containsBundle(bundle) }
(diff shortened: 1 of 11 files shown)


libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -113,19 +113,12 @@ internal class SandboxGroupImpl(
...
                             it.cpkMetadata.cpkId.signerSummaryHash == classTag.cpkSignerSummaryHash
-                                    && it.mainBundle.symbolicName == classTag.mainBundleName
-                        }
-                        sandbox?.let {
-                            if (classTag.classBundleName != it.mainBundle.symbolicName) {
-                                throw SandboxException(
-                                    "Attempted to load class $className with an evolvable class tag from cpk private bundle " +
-                                            "${classTag.classBundleName}."
-                                )
-                            } else {
```
simon-johnson-r3, line 126: This is the code with the most risk associated.

Here we used to look up main bundle symbolic name, but now we don't support that in the evolvable tag. Instead we use only class bundle symbolic name - this is the symbolic name obtained directly from OSGi and as line 121 suggests, in a working system never differed from the main bundle anyway and we'd throw if it did.

Assuming I have the logic right here, this then should be logically identical.

The reason I do this is so that future version evolvable tags read back by Corda 5.0 software will still work, but will be limited to supporting evolving the class in the same CPK.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/classtag/ClassTag.kt
```diff
@@ -33,13 +33,17 @@ internal abstract class StaticTag : ClassTag() {
...
+ * @property cordaCpkCordappName The name given to the CPK by the CorDapp developer which uniquely identifies it. This will
+ * be used for continuity purposes, the assumption being that a developer expects Corda to treat a class from a CPK with this
+ * cordaCpkCordappName as the same class type as any other with the same cordaCpkCordappName - generally to identify it between
+ * serialization and deserialization.
  * @property cpkSignerSummaryHash A summary hash of the hashes of the public keys that signed the CPK the class is from.
  */
 internal abstract class EvolvableTag : ClassTag() {
-    abstract val mainBundleName: String
+    abstract val cordaCpkCordappName: String
     abstract val cpkSignerSummaryHash: SecureHash?
```
simon-johnson-r3, line 47: Whilst this looks like a simple variable rename, please keep in mind this is a different property with conceptually different data in it. We no longer store the main bundle symbolic name in the tag. Instead we store `Corda-CPK-Cordapp-Name` although note we cannot get at that until the plugin changes as part of CORE-13259.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -113,19 +113,12 @@ internal class SandboxGroupImpl(
...
-                                it
-                            }
+                                    && it.mainBundle.symbolicName == classTag.classBundleName
```
mnesbit: I think this is the nubbins of the problem. We should be matching cpkMetatda against Corda-CPK-Cordapp-Name from the evolvable tag. This happens to be same as bundle symbolic name, but it is merely meant to be a cpk identifier. The main bundle of this cpk is then the appropriate place to lookup the class (which may fail if they haven't evolved things right).
Where this needs tweaking is that for system classloader and platform bundles we can't match on cpkMetdata as they aren't cpk's, so they would have to be spotted early based upon the evolvable tag magic values and then we look for bundles in our platform.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/classtag/ClassTagFactoryImpl.kt
```diff
@@ -37,10 +37,13 @@ internal class ClassTagFactoryImpl : ClassTagFactory {
             if (isStaticTag) {
                 StaticTagImplV1(ClassType.CpkSandboxClass, bundleName, cpkSandbox.cpkMetadata.fileChecksum)
             } else {
-                val mainBundleName = cpkSandbox.mainBundle.symbolicName
+                // In later versions of Corda 5, cordaCpkCordappName will be populated from Corda-CPK-Cordapp-Name
+                // from the CPK metadata, and this will be modifiable by the CorDapp developer. Currently Corda-CPK-Cordapp-Name
+                // is always populated by the build system with the CPK symbolic bundle name, so we do that here too.
+                val cordaCpkCordappName = cpkSandbox.mainBundle.symbolicName
```
josephzunigadaly: We have `Corda-CPK-Cordapp-Name` in our CPK manifest if we want to wire it up. It would mean raising an API ticket to change Avro objects. I'm not sure how easy it will be to change Avro types after GA.

josephzunigadaly (dismissed):

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
simon-johnson-r3: Discussed offline. Here we will go one step further and implement the actual `Corda-CPK-Cordapp-Name` storing logic and put the proper test in here. This way we don't have to revisit on later versions except for the gradle plugin.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
kyriathar: I think this was added to make sure an evolvable tag for a class comes from a main bundle and not a private/ lib one. It happens during deserialization and is the symmetric check of https://github.com/corda/corda-runtime-os/blob/bb8e6380a09f340bd48f4752542a5def1b534d7b/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt#L167 which happens during serialization. Should we be adding it back using the new property? So with the current state we are checking on serialization with above check but not have this symmetric check on deserialization.

simon-johnson-r3: We want evolvable classes to be identified as being the same type entirely by the `Corda-CPK-Cordapp-Name` of the CPK in which they were declared. End users are then free to move classes around bundles (for example if they want to refactor their app and it involves renaming the subproject) but as long as they keep `Corda-CPK-Cordapp-Name` the same in the original and new bundle we identify that class as being the same type in both version.

This means it is legitimate that `classTag.classBundleName != it.mainBundle.symbolicName` for example if class A was previously defined in CPK `myCpk` (symbolic bundle name `com.some-org.myCpk`) which was then renamed to `myCpk2` (symbolic bundle name `com.some-org.myCpk2`). As long as the app developer sets `Corda-CPK-Cordapp-Name` in their project such that `myCpk` and `myCpk2` have the same value, this will be fine. However we can see here the serialized class type would have a `classTag.classBundleName` of `com.some-org.myCpk` but the class is now declared in `com.some-org.myCpk2` which are different, so we can not throw.

Note that `Corda-CPK-Cordapp-Name` cannot be set by the user yet, that's not supported in the gradle plugin, that is scheduled for 5.1. Currently it defaults to bundle symbolic name.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/classtag/ClassTagFactoryImpl.kt
simon-johnson-r3: It's already in the attributes map of the manifest, I have wired it up as part of @mnesbit comment above. Please have a look over the change. Note that the Avro message doesn't need changing because the name field in the `CpkIdentifier` is now switched to `Corda-CPK-Cordapp-Name` instead of bundle symbolic name, so the field was already there but being used differently. Apparently this was the original intent which got lost somewhere along the way.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -110,38 +110,31 @@ internal class SandboxGroupImpl(
...
-                                )
-                            } else {
-                                it
-                            }
-                        }
+                        cpkSandboxes.find {
+                            it.cpkMetadata.cpkId.signerSummaryHash == classTag.cpkSignerSummaryHash &&
+                                    (it.cpkMetadata.cpkId.name == classTag.cordaCpkCordappName || // CPK given names match or
+                                    it.mainBundle.symbolicName == classTag.classBundleName) // symbolic names of class bundle match
+                        }?.let { Pair(it, it.mainBundle.symbolicName) } // only load evolvable classes from the main bundle
```
simon-johnson-r3, line 122: the removed code, from line 120 above, checked the type was in the main bundle, and was considered necessary (thanks @kyriathar for pointing this out) which means here we must return the sandbox and the name of a bundle - for evolvable tags it's always the main bundle, which means if we somehow serialized a type from outside the main bundle, line 128 below would fail - this now replaces the old check and throw

for static tags the bundle is classTag.classBundleName which is the same as before

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -110,38 +110,31 @@ internal class SandboxGroupImpl(
...
                 )
             }
 
             ClassType.PublicSandboxClass -> {
                 publicSandboxes.asSequence().mapNotNull { publicSandbox ->
                     publicSandbox.loadClass(className, classTag.classBundleName)
                 }.firstOrNull() ?: throw SandboxException(
-                    "Class $className from bundle ${classTag.classBundleName} could not be loaded from any of the public " +
-                            "sandboxes."
+                    "Class $className from bundle ${classTag.classBundleName} could not be loaded from any of the public " + "sandboxes."
```
josephzunigadaly: It's minor but we could remove this `+ "sandboxes."` string concatenation. It only existed for the line break which we remove here.

josephzunigadaly (dismissed):

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
simon-johnson-r3: good spot, auto format did this, I'll fix

testing/cpbs/packaging-verification-app-v1/src/main/kotlin/com/r3/corda/testing/packagingverification/ReceiveTransferFlow.kt
```diff
@@ -16,6 +17,7 @@ class ReceiveTransferFlow : ResponderFlow {
     @CordaInject
     lateinit var utxoLedgerService: UtxoLedgerService
 
+    @Suspendable
```
simon-johnson-r3, line 20: Spotted this missing in the logs during testing.

testing/cpbs/packaging-verification-app-v1/src/main/kotlin/com/r3/corda/testing/packagingverification/ReportStatesFlow.kt
```diff
@@ -5,11 +5,13 @@ import net.corda.v5.application.flows.ClientStartableFlow
 import net.corda.v5.application.flows.CordaInject
 import net.corda.v5.ledger.utxo.UtxoLedgerService
 import com.r3.corda.testing.packagingverification.contract.SimpleState
+import net.corda.v5.base.annotations.Suspendable
 
 class ReportStatesFlow : ClientStartableFlow {
     @CordaInject
     lateinit var utxoLedgerService: UtxoLedgerService
 
+    @Suspendable
```
simon-johnson-r3, line 14: Spotted this missing in the logs during testing.

simon-johnson-r3:
I've logged some usage of the SandboxGroupImpl whilst executing transactions, I can see the evolvable tag being used, for example
```
Sandbox getClass
className: com.r3.corda.testing.packagingverification.contract.SimpleState
EvolvableTag
bundleName: net.corda.packaging-verification-contract-v1
Sandbox getClass
className: com.r3.corda.testing.packagingverification.contract.TransferCommand
EvolvableTag
bundleName: net.corda.packaging-verification-contract-v1
```
and also static tags in use (presumably something like Kryo deserialization during flow resume)
```
Sandbox getClass
className: com.r3.corda.testing.packagingverification.TransferStatesFlow
StaticTag
bundleName: net.corda.packaging-verification-app-v1
```
So it all appears to work. I can't 100% confirm the sandboxes are matched by Corda-CPK-Cordapp-Name for evolvable tags rather than bundle symbolic name because they are always identical in 5.0.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -75,8 +74,9 @@ internal class SandboxGroupImpl(
                 null
             }
         }.singleOrNull()
-            ?: throw SandboxException("Class $className was not found in any sandbox in the sandbox group.")
-                .withSuppressed(suppressed)
+            ?: throw SandboxException("Class $className was not found in any sandbox in the sandbox group.").withSuppressed(
+                suppressed
+            )
```
kyriathar: I think these are all only formatting changes, however it will overwrite git message, could we revert please?

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -159,18 +152,15 @@ internal class SandboxGroupImpl(
...
+        val publicSandbox = publicSandboxes.find { sandbox -> sandbox.containsBundle(bundle) }
         if (publicSandbox != null) {
             return classTagFactory.createSerialisedTag(isStaticTag, bundle, null)
         }
 
-        val cpkSandbox =
-            cpkSandboxes.find { sandbox -> sandbox.containsBundle(bundle) }
-                ?: throw SandboxException("Bundle ${bundle.symbolicName} was not found in the sandbox group or in a public sandbox.")
+        val cpkSandbox = cpkSandboxes.find { sandbox -> sandbox.containsBundle(bundle) }
+            ?: throw SandboxException("Bundle ${bundle.symbolicName} was not found in the sandbox group or in a public sandbox.")
```
kyriathar: Same as above formatting changes, could we revert please.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -110,38 +110,31 @@ internal class SandboxGroupImpl(
...
-                                            "${classTag.classBundleName}."
-                                )
-                            } else {
-                                it
-                            }
-                        }
+                        cpkSandboxes.find {
+                            it.cpkMetadata.cpkId.signerSummaryHash == classTag.cpkSignerSummaryHash &&
+                                    (it.cpkMetadata.cpkId.name == classTag.cordaCpkCordappName || // CPK given names match or
+                                    it.mainBundle.symbolicName == classTag.classBundleName) // symbolic names of class bundle match
```
kyriathar, line 121: I thought we were replacing this check with trying to load it from main bundle below, so I think we should remove this now. (`it.mainBundle.symbolicName == classTag.classBundleName`)

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
simon-johnson-r3: yes I will do that, but the wider problem is we can never use an auto formatting tool, including the one built into the IDE which I used here to format the functions below, without destroying any idiosyncratic formatting of the original author - even if that isn't consistent with the rest of the file as the IDE suggests here

this is an open source project visible to our customers and potential customers and we should really have auto formatting everywhere IMHO, but there is an action in progress to look into this already

simon-johnson-r3: Also worth noting the "blame" message in the IDE is unaffected by these changes, except that it increments the +n changed suffix, but the original author is preserved.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
simon-johnson-r3: This was added at the request of @mnesbit I can remove as long as he's happy with that. It was added when it was not clear to either of us that the platform types went through other path of:
```
ClassType.PublicSandboxClass -> {
                publicSandboxes.asSequence().mapNotNull { publicSandbox ->
                    publicSandbox.loadClass(className, classTag.classBundleName)
```
so it probably can be removed now.

mnesbit: I think this is needed for when we have old and new nodes interoperating.

mnesbit: but we should check this as I could well be wrong

mnesbit: ok. I think you guys are much more on top of this than me. So yes go with
```
cpkSandboxes.find {
                            it.cpkMetadata.cpkId.signerSummaryHash == classTag.cpkSignerSummaryHash &&
                                    it.cpkMetadata.cpkId.name == classTag.cordaCpkCordappName // CPK given names match
}
```

simon-johnson-r3: This code here represents old nodes now, so I think as `classTag.cordaCpkCordappName` is always written to the tag, we can probably remove it. I'll remove and test.

kyriathar (approved):
LGTM

PR #3781, opened 2023-05-03
CORE-10329 Support for separate member and registration context in internal registration messaging

charlieR3:
API PR: https://github.com/corda/corda-api/pull/1084

This change introduces an additional context map to the internal network registration request. This is because the original context map is signed over and the signature becomes invalid if the map is modified so whatever is submitted in that context map became part of the member's member context map. It's preferred to not include the pre-auth token in the member info distributed across the network. To support this and other properties in the future which should not be distributed, an additional context map has been added.

This new context map is also signed. To avoid duplication, data and a signature over that data was pulled out in to it's own schema and membership schemas were updated where the same structure is already in use.


dimosr (commented):
LGTM - only a comment on the API PR that would require some refactoring on the names here

dimosr (dismissed):

dimosr (approved):

PR #3783, opened 2023-05-03
CORE-11707 Fix lookup in set of `PublicKey`s between `BCECPublicKey` and non `BCECPublicKey`

kyriathar:


kyriathar:
Closing as this is not an issue after all. We should rather be making sure we use same `java.security.Provider`.

PR #3784, opened 2023-05-03
CORE-8714 Update authentication protocol

williamvigorr3:
Update the authentication protocol to send the protocol mode(s) as a part of the handshake messages instead of the hello messages. Changed the session manager to fail session negotiation if protocol mode negotiation fails.

Needs api change: https://github.com/corda/corda-api/pull/1083

### DIFF
--- a/components/gateway/src/test/kotlin/net/corda/p2p/gateway/messaging/internal/InboundMessageHandlerTest.kt
+++ b/components/gateway/src/test/kotlin/net/corda/p2p/gateway/messaging/internal/InboundMessageHandlerTest.kt
@@ -24,7 +24,6 @@ import net.corda.data.p2p.crypto.CommonHeader
 import net.corda.data.p2p.crypto.InitiatorHandshakeMessage
 import net.corda.data.p2p.crypto.InitiatorHelloMessage
 import net.corda.data.p2p.crypto.MessageType
-import net.corda.data.p2p.crypto.ProtocolMode
 import net.corda.data.p2p.crypto.ResponderHandshakeMessage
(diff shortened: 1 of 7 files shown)


dimosr (dismissed):
LGTM.

I assume what's missing is removing the supported modes from the responder constructor and performing the group policy lookup inside the `validatePeerHandshakeMessage` (e.g. with a lambda function passed in as a parameter that will use the responder public key hash included in the encrypted extensions), which as you explained in the standup can be done as a follow-up and will not bring any further changes to the authentication protocol so it can also be done after Friday's cutoff.

williamvigorr3:
> LGTM.
> 
> I assume what's missing is removing the supported modes from the responder constructor and performing the group policy lookup inside the `validatePeerHandshakeMessage` (e.g. with a lambda function passed in as a parameter that will use the responder public key hash included in the encrypted extensions), which as you explained in the standup can be done as a follow-up and will not bring any further changes to the authentication protocol so it can also be done after Friday's cutoff.

Yes I also need to do the same for `certificateCheckMode` i.e. remove it from the constructor (as we don't know what it should be until we lookup the group policy). The session certificates are already sent as a part of the encrypted extensions, so no API changes are needed for that. Do you want me to raise a follow up Jira?

dimosr (approved):

dimosr:
> Yes I also need to do the same for `certificateCheckMode` i.e. remove it from the constructor (as we don't know what it should be until we lookup the group policy). The session certificates are already sent as a part of the encrypted extensions, so no API changes are needed for that. Do you want me to raise a follow up Jira?

Good point. We should probably keep this ticket open and follow up with the remaining changes soon.

PR #3785, opened 2023-05-03
CORE-13289: Allow Kryo serialization mocks to use any Java class.

chrisr3:
Modify the Kryo test-kit to handle any `java.*` class, for more realistic serialization testing.

Also open the `java.time.chrono` and `java.time.zone` packages for our workers so that CorDapps can use these classes safely.


seanbrereton (approved):
LGTM(build files)

driessamyn (approved):
LGTM

PR #3786, opened 2023-05-03
CORE-12094: Remove the call to stop the DbReconcilerReader lifecycle on db errors

fowlerrr:
We use the DbConnectionMananger to let us know the database is down.  And we already rethrow the exception, so we don't need to handle this in any other way.


kyriathar (approved):
LGTM

josephzunigadaly (approved):

driessamyn (approved):
LGTM

PR #3787, opened 2023-05-03
CORE-12443: Avoid creating unused secrets

oscardalmau-r3:
The Corda Helm chart creates the `-rbac-db` , `-crypto-db`, and `-rest-api-admin` secrets even when there is no bootstrapping enabled to consume them. This is potentially confusing, not least for the last of these where the Helm chart notes will indicate that this secret contains the API credentials when the API is actually using those stored in the database.

### DIFF
--- a/charts/corda/templates/secrets.yaml
+++ b/charts/corda/templates/secrets.yaml
@@ -26,6 +26,7 @@
     ( dict "salt" ( dict "generate" 32 ) "passphrase" ( dict "generate" 32 ) )
   )
 }}
+{{- if .Values.bootstrap.db.enabled }}
 {{- include "corda.secret"
   ( list
     $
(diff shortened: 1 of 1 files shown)


charts/corda/templates/secrets.yaml
```diff
@@ -55,6 +58,9 @@
     ( dict "username" ( dict "required" true ) "password" ( dict "generate" 12 ) )
   )
 }}
+{{ else if .Values.bootstrap.rbac.enabled }}
+{{- fail "please set credentials for restApiAdmin secret or disable bootstrap RBAC" }}
```
davidcurrie: I'd word this slightly differently:
```suggestion
{{- fail "credentials for restApiAdmin must be provided to bootstrap RBAC if DB bootstrap is disabled" }}
```

davidcurrie (dismissed):
Suggested a tweak to the words but otherwise the logic looks sound

davidcurrie (approved):
👍

PR #3788, opened 2023-05-03
CORE-12311 Upsert users and fail on errors

davidcurrie:
Check whether the RBAC and crypto users exist before trying to create them. Add `-v ON_ERROR_STOP=1` to `psql` commands to fail on error.

### DIFF
--- a/charts/corda-lib/templates/_bootstrap.tpl
+++ b/charts/corda-lib/templates/_bootstrap.tpl
@@ -32,7 +32,7 @@ spec:
           {{- include "corda.bootstrapResources" . | nindent 10 }}
           {{- include "corda.containerSecurityContext" . | nindent 10 }}
           command:
-            - /bin/sh
+            - /bin/bash
             - -e
             - -c
           args: ["echo", "'DB Bootstrapped'"]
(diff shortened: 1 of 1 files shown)


charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -51,18 +51,20 @@ spec:
...
+                DO \$\$ BEGIN IF EXISTS (SELECT FROM pg_catalog.pg_roles WHERE rolname = '$RBAC_DB_USER_USERNAME') THEN RAISE NOTICE 'Role "$RBAC_DB_USER_USERNAME" already exists'; ELSE CREATE USER "$RBAC_DB_USER_USERNAME" WITH ENCRYPTED PASSWORD '$RBAC_DB_USER_PASSWORD'; END IF; END \$\$;
+                GRANT USAGE ON SCHEMA {{ .Values.bootstrap.db.rbac.schema }} TO "$RBAC_DB_USER_USERNAME";
+                GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA {{ .Values.bootstrap.db.rbac.schema }} TO "$RBAC_DB_USER_USERNAME";
+                DO \$\$ BEGIN IF EXISTS (SELECT FROM pg_catalog.pg_roles WHERE rolname = '$CRYPTO_DB_USER_USERNAME') THEN RAISE NOTICE 'Role "$CRYPTO_DB_USER_USERNAME" already exists'; ELSE CREATE USER "$CRYPTO_DB_USER_USERNAME" WITH ENCRYPTED PASSWORD '$CRYPTO_DB_USER_PASSWORD'; END IF; END \$\$;
+                GRANT USAGE ON SCHEMA {{ .Values.bootstrap.db.crypto.schema }} TO "$CRYPTO_DB_USER_USERNAME";
+                GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA {{ .Values.bootstrap.db.crypto.schema }} TO "$CRYPTO_DB_USER_USERNAME";
+              SQL
```
davidcurrie, line 67: Most of this change is just collapsing the multiple `psql` calls into one. The only changes within the SQL itself are the two `DO $$ BEGIN IF EXISTS...` around the `CREATE USER` calls, and the addition of the `-v ON_ERROR_STOP=1` option. The user of the HEREDOC requires the switch from `sh` to `bash`.

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -80,6 +82,7 @@ spec:
       {{- include "corda.bootstrapNodeSelector" . | nindent 6 }}
 
       restartPolicy: Never
+  backoffLimit: 0
```
davidcurrie, line 85: Given we know that some of the init containers will fail if retried, set the limit to zero. This is also consistent with the other bootstrap jobs.

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -470,7 +473,7 @@ a second init container to execute the output SQL to the relevant database
...
-  command: [ 'sh', '-c',{{- if eq .name "db" }} 'for f in /tmp/*.sql; do psql -h {{ required "A db host is required" .Values.db.cluster.host }} -p {{ include "corda.clusterDbPort" . }} -f "$f" --dbname {{ include "corda.clusterDbName" . }}; done'{{- else }} 'psql -h {{ required "A db host is required" .Values.db.cluster.host }} -p {{ include "corda.clusterDbPort" . }} -f /tmp/{{ .sqlFile | default "db-config.sql" }} --dbname "dbname={{ include "corda.clusterDbName" . }} options=--search_path={{ .searchPath | default .Values.db.cluster.schema }}"' {{- end }} ]
+  command: [ 'sh', '-c', '-e',{{- if eq .name "db" }} 'for f in /tmp/*.sql; do psql -v ON_ERROR_STOP=1 -h {{ required "A db host is required" .Values.db.cluster.host }} -p {{ include "corda.clusterDbPort" . }} -f "$f" --dbname {{ include "corda.clusterDbName" . }}; done'{{- else }} 'psql -v ON_ERROR_STOP=1 -h {{ required "A db host is required" .Values.db.cluster.host }} -p {{ include "corda.clusterDbPort" . }} -f /tmp/{{ .sqlFile | default "db-config.sql" }} --dbname "dbname={{ include "corda.clusterDbName" . }} options=--search_path={{ .searchPath | default .Values.db.cluster.schema }}"' {{- end }} ]
```
davidcurrie, line 476: Addition of `-e` on the shell and `-v ON_ERROR_STOP=1` on the `psql` commands to ensure we fail the job if the SQL fails.

oscardalmau-r3 (approved):
LGTM

PR #3789, opened 2023-05-03
CORE-13235 Uniqueness checker in-flight double spend test

relyafi:
### Summary

This PR adds an additional integration test case to verify that "in-flight" double spends cannot occur within the uniqueness checker, i.e. by attempting to spend the same state twice in parallel. In a real system, this would be when multiple DB workers are operating at the same time and different transactions are attempting to spend the same state in parallel, with their requests being processed by different workers.

Reproducing this scenario in practice is very difficult, so the integration test is performed at the backing store level, using multiple backing store instances and executing each in a separate thread. This allows for simultaneous database access. The test is also repeated 10 times to increase the chance of exposing any issues, as reproducing this issue reliably is difficult.

### Testing

To prove that the test will in fact detect double spends (which should never happen), the `JPABackingStoreEntities` code was intentionally broken to alter the SQL that checks for a blank consuming transaction id as part of the update expression. Confirmed that the test fails when this is done.


vlajos (approved):
Not too familiar with this part of the code, but generally it looks good to me.

nkovacsx (approved):

PR #3791, opened 2023-05-03
CORE-12902: Return BadRequest when attempting to change vnode status while DB migration is running

josephzunigadaly:
Add `javax.persistence.RollbackException` to the list of exceptions that map to `BadRequest`


vkolomeyko (approved):
LGTM

PR #3792, opened 2023-05-03
CORE-12555 Interop Facades as Corda service

szymonsztuka:
The aim of this PR is to demonstrate a working Client Proxy WEFT and Server Dispatcher WEFT from within Cordapps and a sandbox.
No changes to security policy files are needed. Three  occurrences of programmatic AccessController.doPrivileged(PrivilegedExceptionAction { ... }) around Weft operation have been added.
The actual simplified Cordapp API is not important here, as the proper API is subject of other work items.


components/security-manager/src/main/resources/basic_security.policy
```diff
@@ -28,6 +29,7 @@ ALLOW {
 
 (net.corda.internal.serialization.amqp.CustomSerializerPermission "FLOW")
 
+(java.lang.reflect.ReflectPermission "suppressAccessChecks" "")
```
chrisr3: There is no point having a `SecurityManager` if we then drive a bus through its security policy :cry:.

components/security-manager/src/main/resources/basic_security.policy
```diff
@@ -55,7 +57,6 @@ DENY {
 (java.lang.management.ManagementPermission "control" "")
 (java.lang.management.ManagementPermission "monitor" "")
 (java.lang.RuntimePermission "createClassLoader" "")
-(java.lang.RuntimePermission "getClassLoader" "")
```
chrisr3: This applies to _all_ `FLOW` sandboxes, not just yours.

components/security-manager/src/main/resources/basic_security.policy
```diff
@@ -265,7 +266,6 @@ DENY {
 (java.lang.RuntimePermission "getFileSystemAttributes" "")
 (java.lang.RuntimePermission "readFileDescriptor" "")
 (java.lang.RuntimePermission "writeFileDescriptor" "")
-(java.lang.RuntimePermission "accessDeclaredMembers" "")
```
chrisr3: Same here. We don't want everyone's CorDapps to be accessing private elements via Java reflection.

components/security-manager/src/main/resources/high_security.policy
```diff
@@ -56,7 +59,6 @@ DENY {
 (java.lang.management.ManagementPermission "control" "")
 (java.lang.management.ManagementPermission "monitor" "")
 (java.lang.RuntimePermission "createClassLoader" "")
-(java.lang.RuntimePermission "getClassLoader" "")
```
chrisr3: Similarly, we don't want to nerf security for `VERIFICATION` sandboxes either.

chrisr3 (commented):
I don't think you can compromise the security policy over this. The sandboxes contain untrusted third-party code and so they _must_ remain secure.

---
The correct way to escape a sandbox's security restrictions is to use a Corda service which invokes `AccessController.doPrivileged()`, as (e.g.) `JsonMarshallingServiceImpl` does.

The `SecurityManager` applies _all_ of the policy rules whose conditions match _any_ function on the current call-stack. Invoking `doPrivileged()` temporarily truncates the view of that call-stack to just the function invoking `doPrivileged()`, which allows `SecurityManager` to "forget" about any sandbox functions the current function may have been invoked from. We obviously only want to do this for as short a time as possible.

components/security-manager/src/main/resources/basic_security.policy
szymonsztuka: As you have advised me on other channels, I've removed all changes to `policy` files (both additions to ALLOW and removals from DENY), using `AccessController.doPrivileged(PrivilegedExceptionAction { ... })` for three operation `FacadeReaders.JSON.read`, accessing ` boundInterface.declaredMethods` in the class `InterfaceBindingContext`, and getting a ClassLoader in `FacadeProxies` (commits: https://github.com/corda/corda-runtime-os/pull/3792/commits/2237f1b1baed0b8826daae686f513397fa17f075 and https://github.com/corda/corda-runtime-os/pull/3792/commits/4626d9cd1d79952cbd5757593328c6c0ea8a234c).

yarivus (commented):
Reviewed - didn't follow rationale of the changes ...
Some of them looked like hard-coded test-code

yarivus (approved):
LGTM

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FacadeServiceImpl.kt
```diff
@@ -0,0 +1,101 @@
...
+        val facadeResponse = dispatcher.invoke(facadeRequest)
+        return jsonMarshallingService.format(facadeResponse)
+    }
+
+    @Suppress("UNUSED_PARAMETER")
+    private fun facadeLookup(facadeId: String): Facade {
+        return try {
+            AccessController.doPrivileged(PrivilegedExceptionAction {
+                FacadeReaders.JSON.read(
+                    """{
```
szymonsztuka, line 63: This configuration is intentionally hard-coded, as the proper solution will be addressed in next work items.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/proxies/FacadeClientProxy.kt
```diff
@@ -33,9 +36,13 @@ object FacadeProxies {
...
+            AccessController.doPrivileged(PrivilegedExceptionAction { interfaceType.classLoader })
+        } catch (e: PrivilegedActionException) {
+            throw e.exception
+        }
         return Proxy.newProxyInstance(
-            interfaceType.classLoader,
+            classLoader,
             arrayOf(interfaceType),
             proxy
         ) as T
```
chrisr3: Maybe better:
```kotlin
return try 
    AccessController.doPrivileged(PrivilegedExceptionAction {
        Proxy.newProxyInstance(
            interfaceType.classLoader,
            arrayOf(interfaceType),
            proxy
        ) as T
    })
} catch (e: PrivilegedActionException) {
    throw e.exception
}
```
because I would expect creating dynamic proxies to require privileges as well:

[SecurityException](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/SecurityException.html) - if a security manager, s, is present and any of the following conditions is met:
- the given loader is null and the caller's class loader is not null and the invocation of [s.checkPermission](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/SecurityManager.html#checkPermission(java.security.Permission)) with RuntimePermission("getClassLoader") permission denies access;
- for each proxy interface, intf, the caller's class loader is not the same as or an ancestor of the class loader for intf and invocation of [s.checkPackageAccess()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/SecurityManager.html#checkPackageAccess(java.lang.String)) denies access to intf;
- any of the given proxy interfaces is non-public and the caller class is not in the same [runtime package](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Package.html) as the non-public interface and the invocation of [s.checkPermission](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/SecurityManager.html#checkPermission(java.security.Permission)) with ReflectPermission("newProxyInPackage.{package name}") permission denies access.

chrisr3 (commented):
Not needing to modify the security policy is "good enough", although I think [Proxy.newProxyInstance](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/reflect/Proxy.html#newProxyInstance(java.lang.ClassLoader,java.lang.Class%5B%5D,java.lang.reflect.InvocationHandler)) also requires privileges.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/proxies/FacadeClientProxy.kt
szymonsztuka: I've applied your suggestion in commit  daaadc381d4640c384a1e08233bcc138ae507492.

yarivus (commented):
thx @szymonsztuka

PR #3795, opened 2023-05-04
CORE-3819: Remove unused OSGi dependencies from Notary CPK.

chrisr3:
These OSGi dependencies are not used and do not belong here. They need to be removed.


yarivus (approved):
LGTM

PR #3799, opened 2023-05-04
CORE-13456: POC one, no RPC

yift-r3:


components/membership/membership-persistence-client-impl/src/main/kotlin/net/corda/membership/impl/persistence/client/RequestSender.kt
```diff
@@ -0,0 +1,85 @@
...
+import net.corda.messaging.api.records.Record
+import net.corda.messaging.api.subscription.config.SubscriptionConfig
+import net.corda.messaging.api.subscription.factory.SubscriptionFactory
+import net.corda.schema.Schemas
+import java.util.UUID
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.Future
+
+internal class RequestSender(
```
williamvigorr3, line 23: Lets not forget to unit test this eventually.

components/membership/membership-persistence-client-impl/src/main/kotlin/net/corda/membership/impl/persistence/client/RequestSender.kt
```diff
@@ -0,0 +1,85 @@
...
+                    value = request,
+                ),
+            ),
+        ).forEach {
+            it.join()
+        }
+        return future
+    }
+
+    private inner class Processor : PubSubProcessor<String, MembershipPersistenceResponse> {
```
williamvigorr3, line 83: We could use a `DurableProcessor` instead WDYT? 

On one hand If the a REST worker or Member worker process dies (and we start a new processor to replace it) then we want to start from the end of the topic (so a PubSubProcessor makes sense in that case). 

If however we lose communication with Kafka for some reason (I don't know if this can happen in reality), we could end up not completing futures (as when we start a new `PubSubProcessor` it will start from the end of the topic, I think).

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/MembershipPersistenceRPCProcessor.kt
```diff
@@ -2,28 +2,46 @@ package net.corda.membership.impl.persistence.service
...
+import net.corda.membership.impl.persistence.service.handler.HandlerFactories
 import net.corda.membership.lib.exceptions.InvalidEntityUpdateException
-import net.corda.messaging.api.processor.RPCResponderProcessor
+import net.corda.messaging.api.processor.DurableProcessor
+import net.corda.messaging.api.records.Record
+import net.corda.schema.Schemas.Membership.MEMBERSHIP_DB_RPC_RESPONSE_TOPIC
 import org.slf4j.LoggerFactory
-import java.util.concurrent.CompletableFuture
 
 internal class MembershipPersistenceRPCProcessor(
```
williamvigorr3, line 16: We should eventually rename this.

williamvigorr3 (commented):
I didn't look at the tests yet.

PR #3800, opened 2023-05-04
CORE-3819: Fix publishing for non-R3 users.

chrisr3:
The task which creates a Kotlin "sources" jar is called `kotlinSourcesJar`, and we cannot include both this and Java's `sourcesJar` task in a single `MavenPublication`.

Only include the `sourcesJar` task if the publication doesn't already includes an artifact with the `sources` classifier.

### DIFF
--- a/buildSrc/src/main/groovy/corda.common-publishing.gradle
+++ b/buildSrc/src/main/groovy/corda.common-publishing.gradle
@@ -1,3 +1,5 @@
+import groovy.transform.CompileStatic
+
 // plugin to cater for R3 vs Non R3 users building code base. R3 employees will leverage internal plugins non
 // R3 users will use standard Maven publishing conventions as provided by the Maven-publish gradle plugin
 if (System.getenv('CORDA_ARTIFACTORY_USERNAME') != null || project.hasProperty('cordaArtifactoryUsername')) {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/corda.common-publishing.gradle
```diff
@@ -15,11 +17,13 @@ if (System.getenv('CORDA_ARTIFACTORY_USERNAME') != null || project.hasProperty('
...
-                            from components.findByName('cordapp') ?: components.findByName('kotlin') ?: components.java
+                            from findSoftwareComponent(components).get()
 
-                            try {
-                                artifact tasks.named('sourcesJar', Jar)
-                            } catch (UnknownTaskException ignored) {
+                            if (artifacts.matching { it.classifier == 'sources' }.isEmpty()) {
+                                try {
+                                    artifact tasks.named('sourcesJar', Jar)
+                                } catch (UnknownTaskException ignored) {
```
yarivus, line 25: I think that even if you chose to ignore - it's best to log it (for debugging)

yarivus (commented):
see my comment

buildSrc/src/main/groovy/corda.common-publishing.gradle
chrisr3: You misunderstand. The exception would be thrown because the task is not there, in which case the correct thing to do is "nothing at all".

Gradle (being "Gradle" :hankey:) doesn't provide a suitable API to detect whether the task exists. I am therefore forced to use an API which throws an exception instead, and then do "nothing at all" in the `catch` block.

Logging the exception is pointless. This is just `if - else` with exceptions because Gradle leaves me with no other choice.

ronanbrowne (approved):
LGTM

PR #3801, opened 2023-05-04
CORE-12581: run signing service direct from flow operations processor rather than going via RPC, in order to eliminate problems with lack of robustness in crypto to crypto RPC calls

dickon:
Perform flow crypto operations direct from the thread that picks up the Kafka topic, rather than crossing Kafka twice more.

Internally, the crypto worker implements a signing service, which provides operation to sign and verify signatures, as well as other operations such as creating and retrieve asymmetric keys. This exposed to Corda via 2 bus processor, each watching a Kakfa topic. One topic and bus processor is for flow operations, and the other for RPC operations. Up to this point, the bus  processor for flow operations would handle requests by posting them by  being RPC client, so posting to the RPC topic and waiting for a partition owned by that instance of the crypto processor. The trouble is that this is unreliable, since Kafka topics can and are rebalanced at any point, and there was no retry code.

This pattern of messages crossing kafka twice in each direction was done to allow, in the future, different instances of the crypto processor to handle a subset of keys. By posting crypto flow operations back to Kafka, any arrangement we have to make sure that a crypto processor instance with access to a specific key would extend to flow operations as well. However, we now think that for the foreseeable future all crypto processor instances will be able to access any key, and therefore this sophistication is not necessary, or, due to its inherent complexity and performance overheads, desirable. 

Previously, we had 2 separate OSGi+Lifecycle service classes that subscribe to Kafka. Since they both now need to access the signing service, and since the signing service is not itself OSGi, that means that we would have to construct the signing service implementation class. It's also a lot of code, and in particular code using AbstractConfigurableComponent. Instead, we have Crypto Processor itself receive dynamic config then construct the Signing Service and the Kafka subscriptions.  This is much simpler, plus now the crypto processor itself actually does something more than report the overall health of the crypto processor as ready once the dependencies start. Rejected design alternatives were to create the signing service twice, each with its own cache, or make the signing service be OSGi+Lifecycle itself, which makes it harder to test.

Make the crypto service bus implementations exported, so that crypto processor can import them. Perhaps we could alter move the service bus implementations into crypto processor, but that can be done in a separate PR to keep this reviewable and minimise changes if this is backported.

I notice that FilterMyKeysFlowQuery avro message is unused, so have added a separate JIRA [CORE-13554](https://r3-cev.atlassian.net/browse/CORE-13554) to address this.

Rework the crypto flows operations bus processor unit tests. There was a lot of repetition of the machinery to actually run the flows,  mixed up with the assertions for specific tests which made it pretty hard to maintain, and the code all had to be touched since the underyling code is the signing service now rather than the crypto ops client.

[CORE-13554]: https://r3-cev.atlassian.net/browse/CORE-13554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -79,66 +117,72 @@ class CryptoProcessorImpl @Activate constructor(
...
     @Suppress("ComplexMethod", "NestedBlockDepth")
     private fun eventHandler(event: LifecycleEvent, coordinator: LifecycleCoordinator) {
-        logger.info("Crypto processor received event $event.")
+        logger.trace("Crypto processor received event $event.")
         when (event) {
             is StartEvent -> {
-                // Nothing to do
+                logger.trace("Crypto processor starting")
+                registration?.close()
+                registration = coordinator.followStatusChangesByName(dependentComponents.coordinatorNames)
```
fowlerrr, line 163: I mean, I wouldn't be doing my job if I didn't mention that you can use managed resources here...

https://github.com/corda/corda-runtime-os/wiki/Lifecycle-Management#using-managed-resources-to-automate-closing-of-resources

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -175,16 +220,72 @@ class CryptoProcessorImpl @Activate constructor(
...
+            config = cryptoConfig.signingService()
+        )
+
+        // make both the processors
+        val flowOpsProcessor = CryptoFlowOpsBusProcessor(signingService, externalEventResponseFactory, event)
+        val rpcOpsProcessor = CryptoOpsBusProcessor(signingService, cryptoConfig)
+
+        // now make both the subscriptions
+        val messagingConfig = event.config.getConfig(MESSAGING_CONFIG)
+        val flowGroupName = "crypto.ops.flow"
```
fowlerrr, line 254: Is there a place to define all of these group names, etc?  Just here feels weird.  Even constants in the class would be okay, though I feel like some must be used elsewhere.

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/bus/CryptoOpsBusProcessorTests.kt
```diff
@@ -29,77 +29,6 @@ import kotlin.test.assertEquals
...
-//import java.time.Instant
-//import java.util.UUID
-//import java.util.concurrent.CompletableFuture
-//import java.util.concurrent.ExecutionException
-//import kotlin.test.assertEquals
-//import kotlin.test.assertFailsWith
-//import kotlin.test.assertNotNull
-//import kotlin.test.assertNull
-//import kotlin.test.assertTrue
-//
```
fowlerrr: Er, delete?

fowlerrr: Or, are you still writitng up these tests?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
```diff
@@ -26,12 +32,14 @@ import net.corda.utilities.MDC_FLOW_ID
 import net.corda.utilities.trace
 import net.corda.utilities.withMDC
 import org.slf4j.LoggerFactory
+import java.nio.ByteBuffer
 import java.time.Instant
+import kotlin.math.sign
 
 class CryptoFlowOpsBusProcessor(
-    private val cryptoOpsClient: CryptoOpsProxyClient,
+    private val signingService: SigningServiceImpl,
```
fowlerrr: It's a bit weird to use the impl instead of the interface here.  Kind of makes you wonder _why_ the interface?

fowlerrr (commented):
A few comments to start.  I'm still looking at the main change of moving it all into the BusProcessor, which looks nice and tidy, but I don't know enough yet to know if it works.

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/bus/CryptoOpsBusProcessorTests.kt
dickon: Changes to this file reverted. It seems I accidentally deleted a bunch of those tests a few weeks ago, so they need putting back, but it's not relevant to this change so I'll move that to a separate branch.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
dickon: I do often wonder why we have the interface. But in this case it is deliberate, the reason is that I want to use a variant of an entry point that does not log errors and backtraces to the console.

Maybe that should be in the interface but it doesn't really need to be.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
dickon: Maybe there should be; I don't know of a central dictionary of all the Kafka-related names.  Currently this string is defined in the CryptoFlowOpsBuseServiceImpl, so I made the smallest change I could and moved it here.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
dickon: Glancing at that page, it seems the the relevant object here is the crypto processor, which does have lifecycle, e.g. it exposes a health check. And it started when the crypto worker or combined worker starts, now needs to obtain the crypto and messaging config, then stays around for the duration of that process. So while  this is all still a little bit opaqure to me, it seems okay as it is though it may be worth a quick chat on this. Essentially this PR completely changes what the CryptoProcessor class does, so we may as well get it as clean as possible.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
dickon: Thinking more about this, @fowlerrr  is right; so I want to get a group decision on whether suppressing the error logging is important enough to justify the silent parameter, then either add it in to the interface or delete it.

fowlerrr: I don't think you need `SigningServiceImpl` for that, though.  You can just use `SigningService` here.

Regarding `onNextSilent`, I don't really like it.  I get the idea, and I understand that logging errors during a test can be difficult to parse, but you're creating two different (albeit nearly identical) paths: one for test and one for production.  I don't like that as a rule and this sets a precedent I'd rather avoid.

dickon: The alternatives would be:
1.  Add a `silent` flag to `DurableProcessor.onNext`, in the interface?
2. Some kind of contextual stuff, e.g. thread local storage
3. give up and just log the error and backtrace

dickon: I gave up. The right solution might be to dynamically disable logging at the right time, as discussed here https://stackoverflow.com/questions/28807545/slf4j-or-logback-turn-off-logging-for-1-unit-test-or-1-thread but that requires a non-public API and maybe a new slf4j.

fowlerrr (dismissed):

kyriathar:
Could we please shorten the PR description or at least the squashed text so that it gets merged with the least possible text so it's more quickly readable?

kyriathar:
Also if we are to mention on why this relay used to be in place, maybe worth catching up with @mnesbit to double check. 

>This pattern of messages crossing kafka twice in each direction was done to allow, in the future, different instances of the crypto processor to handle a subset of keys. By posting crypto flow operations back to Kafka, any arrangement we have to make sure that a crypto processor instance with access to a specific key would extend to flow operations as well. However, we now think that for the foreseeable future all crypto processor instances will be able to access any key, and therefore this sophistication is not necessary, or, due to its inherent complexity and performance overheads, desirable.

I believe the above implies it was in place to support multiple HSMs? But in these flow crypto requests I can't see anything implying potential routing to different HSMs i.e. routing to different crypto workers. @mnesbit could help us here.

dickon:
> Also if we are to mention on why this relay used to be in place, maybe worth catching up with @mnesbit to double check.

I've already discussed this with @mnesbit at length and I believe he wants this change made.

>  Could we please shorten the PR description or at least the squashed text so that it gets merged with the least possible text so it's more quickly readable?

I have eliminated some text, since I went a different direction after review with Ryan. However, I think everything in the PR description is relevant, and I am following our policy is to encourage detailed commit messages.

fowlerrr (dismissed):

mcgovc (approved):
LGTM build file changes

mcgovc (approved):
LGTM build file changes

fowlerrr (approved):

PR #3807, opened 2023-05-04
CORE-5711 - Kryo Throw Exception when try to deserialize immutable collections such as EntrySet

thiagoviana:
The [java.util.LinkedHashMap.LinkedEntrySet] has a problem when the default Kryo is deserializing immutable collections: an exception is thrown due to the fact that the methods add() and addAll() were not properly implemented (Take look at the  public interface Map<K, V> method to see that it does not support these operations).

To solve the problem, a new serialiser was created, where we are granting that, when reading, the object is of the same type and the also that order is being respected.


mbrkic-r3:
Looks good, you'll just need to replace wildcard import with fully qualified imports to make `detekt` happy

mbrkic-r3 (dismissed):
LGTM

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/serializers/EntrySetSerializer.kt
```diff
@@ -0,0 +1,41 @@
+package net.corda.kryoserialization.serializers
+
+import com.esotericsoftware.kryo.Kryo
+import com.esotericsoftware.kryo.Serializer
+import com.esotericsoftware.kryo.io.Input
+import com.esotericsoftware.kryo.io.Output
+
+/**
+ * The [LinkedEntrySet] have a problem when the default Kryo is
```
kyriathar:
```suggestion
 * The [java.util.LinkedHashMap.LinkedEntrySet] has a problem when the default Kryo is
```

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/serializers/EntrySetSerializer.kt
```diff
@@ -0,0 +1,41 @@
...
+ * deserializing immutable collections: an exception is thrown.
```
kyriathar: I believe "an exception is thrown" does not say much about the problem.

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/serializers/EntrySetSerializer.kt
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Taking a look at HashMap#keySet() method, it’s possible to see that it does
```
kyriathar: Where is `HashMap#keySet()` used in this code flow?

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/serializers/EntrySetSerializer.kt
```diff
@@ -0,0 +1,41 @@
...
+ * not support the add or addAll operations.
+ *
+ * We've added our own custom serializer in order to ensure that Set<Map.Entry<K, V>>
+ * are being serialized properly.
+ *
+ * Test Case: mapOf("1" to "a", "2" to null).entries
```
kyriathar: Maybe not needed here

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/serializers/EntrySetSerializer.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+internal object EntrySetSerializer : Serializer<Set<Map.Entry<*,*>>>() {
```
kyriathar: I believe this class should be named `LinkedHashSetSerializer` to be more intuitive and match others serializers targeting other types under `LinkedHashMap` that we already have.

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/serializers/EntrySetSerializer.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    // Create a dummy object to get the LinkedHashMap$LinkedEntrySet from it
+    val serializedType: Class<out Set<Map.Entry<*,*>>> = mapOf(Any() to Any(), Any() to Any()).entries::class.java
```
kyriathar:
```suggestion
    val serializedType: Class<out Set<Map.Entry<*,*>>> = linkedMapOf(Any() to Any(), Any() to Any()).entries::class.java
```

kyriathar: I think makes it more intuitive

libs/serialization/serialization-kryo/src/test/kotlin/net/corda/kryoserialization/serializers/EntrySetSerializerTest.kt
```diff
@@ -0,0 +1,36 @@
...
+import net.corda.kryoserialization.DefaultKryoCustomizer
+import org.assertj.core.api.Assertions
+import org.junit.jupiter.api.Test
+import org.mockito.kotlin.mock
+
+internal class EntrySetSerializerTest {
+
+    @Test
+    fun `EntrySet serializer returns correct value`() {
+        val map = mapOf("1" to "a", "2" to null)
```
kyriathar:
```suggestion
        val map = linkedMapOf("1" to "a", "2" to null)
```

kyriathar (commented):
Some tidy up comments

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/serializers/EntrySetSerializer.kt
thiagoviana: I think so. I will remove it. 
_obs.: It seems that we have comments like this one in other places._

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/serializers/EntrySetSerializer.kt
```diff
@@ -0,0 +1,41 @@
...
+    override fun write(kryo: Kryo, output: Output?, obj: Set<Map.Entry<*, *>>) {
+        // HashSet is already supported
+        kryo.writeClassAndObject(output, obj.toHashSet())
```
kyriathar: Having a second thought about this. We need to preserve the order of insertion of the `LinkedEntrySet` through serialization and deserialization therefore I think we should rather `obj.toList()` and further below read this as list. Then we should extend our test to assert same order too and not just the set. @mbrkic-r3 could help us here.

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/serializers/LinkedHashSetSerializer.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * The [java.util.LinkedHashMap.LinkedEntrySet] has a problem when the default Kryo is
+ * deserializing immutable collections: an exception is thrown due to the fact that
+ * the methods add() and addAll() were not implemented.
+ *
+ * Taking a look at HashMap#keySet() method, it’s possible to see that it does
+ * not support the add or addAll operations.
+ */
+internal object LinkedHashSetSerializer : Serializer<Set<Map.Entry<*,*>>>() {
```
kyriathar: Actually it should be `LinkedEntrySetSerializer`, sorry about this

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/serializers/EntrySetSerializer.kt
mbrkic-r3: You are right, great catch!

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/DefaultKryoCustomizer.kt
```diff
@@ -11,19 +11,20 @@ import com.esotericsoftware.kryo.serializers.CompatibleFieldSerializer
...
 import net.corda.kryoserialization.serializers.CertPathSerializer
-import net.corda.kryoserialization.serializers.ClassSerializer
 import net.corda.kryoserialization.serializers.CordaClosureSerializer
 import net.corda.kryoserialization.serializers.IteratorSerializer
-import net.corda.kryoserialization.serializers.LazyMappedListSerializer
 import net.corda.kryoserialization.serializers.LinkedHashMapEntrySerializer
 import net.corda.kryoserialization.serializers.LinkedHashMapIteratorSerializer
 import net.corda.kryoserialization.serializers.LinkedListItrSerializer
-import net.corda.kryoserialization.serializers.LoggerSerializer
 import net.corda.kryoserialization.serializers.NonSerializableSerializer
```
chrisr3: Someone has a curious view of the alphabet... :thinking:

chrisr3 (commented):
How does this new serializer differ from the existing `LinkedHashMapEntrySerializer`?

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/DefaultKryoCustomizer.kt
thiagoviana: I've got some IDE auto formatting problems. I hope not anymore ! 🤪

LWogan: you should still fix it for this pr too. looks like this pr has the issue still

mbrkic-r3:
> How does this new serializer differ from the existing `LinkedHashMapEntrySerializer`?

This serializer is for `LinkedEntrySet`, mentioned existing one is for `Entry`

libs/serialization/serialization-kryo/src/test/kotlin/net/corda/kryoserialization/serializers/LinkedEntrySetSerializerTest.kt
```diff
@@ -0,0 +1,70 @@
...
+
+        val output = Output(3000)
+        kryo.writeClassAndObject(output, entries)
+        output.close()
+
+        val input = Input(output.buffer)
+        val testedEntries = kryo.readClassAndObject(input)
+        input.close()
+
+        Assertions.assertThat(testedEntries).isEqualTo(entries)
```
kyriathar: Here again I think we comparing sets, maybe they would need to be converted to lists first. Not really sure what we achieve with this test. It's not the randomness that makes the initial map, rather the iteration through the `HashMap` values we had before. I think it would be enough on the above test to also convert the two sets to lists and assert they are equals too.

kyriathar (dismissed):
I think the tests might need reviewing but don't want to block the PR. And some imports needs re-ordering as Chris mentioned.

libs/serialization/serialization-kryo/src/test/kotlin/net/corda/kryoserialization/serializers/LinkedEntrySetSerializerTest.kt
kyriathar: Perhaps a more deterministic test that we always know for a linked hash map iteration is different to it's equivalent hash map (could use a small enough map (that does not re-index) to always reproduce that behavior).

thiagoviana: Yes sure, isEqual to checks also the order. But I was missing that one of them was actually a list. I am sending an update now. Thank you!

kyriathar (approved):
LGTM

PR #3808, opened 2023-05-05
CORE-7793 add integration test and handling for serialization errors to db bus producers

LWogan:
Add new flag to decide whether or not a producer should throw an error when serialization fails. The only use case we care about not flagging errors is in durable subscriptions where we handle errors using dead letter queues.

Add integration test to validate the DLQ logic.

### DIFF
--- a/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/producer/CordaTransactionalDBProducerImpl.kt
+++ b/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/producer/CordaTransactionalDBProducerImpl.kt
@@ -26,6 +26,7 @@ class CordaTransactionalDBProducerImpl(
     private val dbAccess: DBAccess,
     private val writeOffsets: WriteOffsets,
     private val headerSerializer: MessageHeaderSerializer,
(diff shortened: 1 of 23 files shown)


libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/producer/CordaTransactionalDBProducerImpl.kt
```diff
@@ -17,15 +19,14 @@ import net.corda.messagebus.db.util.WriteOffsets
...
-import kotlin.math.abs
 
 @Suppress("TooManyFunctions")
 class CordaTransactionalDBProducerImpl(
     private val serializer: CordaAvroSerializer<Any>,
     private val dbAccess: DBAccess,
     private val writeOffsets: WriteOffsets,
-    private val headerSerializer: MessageHeaderSerializer
+    private val headerSerializer: MessageHeaderSerializer,
+    private val throwOnSerializationError: Boolean,
```
pmck91: trailing comma

pmck91 (approved):
LGTM, theres a few trailing commas in class and method signatures

PR #3809, opened 2023-05-05
CORE-13373: Add "eviction listeners" to SandboxGroupContextCache.

chrisr3:
Allow sandboxing components to register "eviction listeners" which are invoked when sandboxes are evicted from the `SandboxGroupContextCache`.

Also simplify the cache handling, now that `SandboxGroupContextCache` can be resized _without_ recreating it.


driessamyn (approved):
LGTM

PR #3810, opened 2023-05-05
CORE-13288: Map errors correctly from the persistance client

yift-r3:
The persistence client mapped the error results wrong, which caused all errors to be logged as: `Query returned unexpected payload.` which is not helpful. This change fixes that issue.

See https://ci02.dev.r3.com/job/Corda5/job/corda-runtime-os-combined-worker-e2e-tests/job/release%252Fos%252F5.0/2470/artifact/workerLogs.txt for an exmple

### DIFF
--- a/components/membership/membership-persistence-client-impl/src/main/kotlin/net/corda/membership/impl/persistence/client/MembershipPersistenceClientImpl.kt
+++ b/components/membership/membership-persistence-client-impl/src/main/kotlin/net/corda/membership/impl/persistence/client/MembershipPersistenceClientImpl.kt
@@ -36,7 +36,6 @@ import net.corda.data.membership.db.response.command.PersistApprovalRuleResponse
 import net.corda.data.membership.db.response.command.PersistGroupParametersResponse
 import net.corda.data.membership.db.response.command.RevokePreAuthTokenResponse
 import net.corda.data.membership.db.response.command.SuspendMemberResponse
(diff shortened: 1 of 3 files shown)


components/membership/membership-persistence-client-impl/src/main/kotlin/net/corda/membership/impl/persistence/client/MembershipPersistenceClientImpl.kt
```diff
@@ -445,15 +444,13 @@ class MembershipPersistenceClientImpl(
     private fun nullToUnitConvertor(payload: Any?): Either<Unit, String> {
         return when (payload) {
             null -> Either.Left(Unit)
-            is PersistenceFailedResponse -> Either.Right(payload.errorMessage)
```
charlieR3, line 448: Should we still handle `PersistenceFailedResponse` here? If not it falls under "Unexpected response" but it's a known response type.

charlieR3 (approved):
LGTM - one minor comment but I'm OK with merging as is.

components/membership/membership-persistence-client-impl/src/main/kotlin/net/corda/membership/impl/persistence/client/MembershipPersistenceClientImpl.kt
yift-r3: We handle the PersistenceFailedResponse in the send of the operation implementation

PR #3818, opened 2023-05-08
ES-358: remove hard coded combined worker version from combined worker e2e test start up

ronanbrowne:
When reading the combined worker from the build dir to start the worker process, don't hard code the version, as this leads to issues when creating new branches with different versions. This directory will only ever contain 1 version of the combined worker so just wildcard the path


mcgovc (approved):
LGTM

yarivus (approved):
LGTM

PR #3819, opened 2023-05-08
CORE-13134 Triggering platform incompatability exception, creating error checkpoint and raising platform exception

ben-millar:
### Context
When we initiate a flow, a call is made to the `FlowProtocolStore` to determine which protocol versions are supported for the requested flow, and these are passed along to the counterparty. Upon receiving the SessionInit, the counterparty will also make a call to the `FlowProtocolStore` passing in the supported versions from the initiating flow, and the protocol store will select and return the latest mutually supported protocol.

### Problem
When no mutually supported protocol can be found, a `FlowFatalException` is thrown as the flow cannot be initiated. However, this exception is being thrown before the counterparty flow is able to initialize their checkpoint and create the session successfully, meaning it blows up before it has any way to communicate that error back to the initiating flow.

What this means is that the initiating flow will hang, waiting for a response, until it times out.

### Solution
To fix this, I've modified the logic such that the counterparty checkpoint will be created regardless, and the `FlowProtocolStore` check is now made in a `runCatching` block, meaning the result and any exceptions are caught in a `Result<T>` object (this can be treated like a `Future<T>`).

Once the checkpoint has been initialised successfully we enqueue either a `SessionConfirm` or a `SessionError` onto the SessionManager, which will be communicated back to the initiating flow meaning any errors will be propagated and handled through the normal pathways.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
@@ -123,7 +123,8 @@ class SessionEventHandler @Activate constructor(
                 result.isFailure -> sendErrorMessage(
                     context,
                     sessionId,
-                    initiatedFlowNameAndProtocolResult.exceptionOrNull()!!
+                    initiatedFlowNameAndProtocolResult.exceptionOrNull() ?:
+                    FlowFatalException("Failed to create initiated checkpoint for session: $sessionId.")
                 )
             }
         }
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
@@ -168,7 +168,7 @@ class SessionEventHandler @Activate constructor(
                 .setIdentity(initiatedIdentity)
                 .setCpiId(sessionInit.cpiId)
                 .setInitiatedBy(initiatingIdentity)
-                .setFlowClassName(initiatedFlowNameAndProtocolResult?.getOrNull()?.flowClassName ?: "INVALID PROTOCOL")
+                .setFlowClassName(initiatedFlowNameAndProtocolResult?.getOrNull()?.flowClassName ?: "Invalid protocol")
                 .setContextPlatformProperties(keyValuePairListOf(mapOf(MDC_CLIENT_ID to sessionId)))
                 .setCreatedTimestamp(Instant.now())
                 .build()
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
@@ -1,5 +1,6 @@
 package net.corda.flow.pipeline.handlers.events
 
+import java.time.Instant
 import net.corda.data.KeyValuePairList
 import net.corda.data.flow.FlowInitiatorType
 import net.corda.data.flow.FlowKey
@@ -31,7 +32,6 @@ import org.osgi.service.component.annotations.Activate
 import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
 import org.slf4j.LoggerFactory
-import java.time.Instant
 
 @Component(service = [FlowEventHandler::class])
 class SessionEventHandler @Activate constructor(
@@ -106,6 +106,7 @@ class SessionEventHandler @Activate constructor(
         val holdingIdentity = initiatedIdentity.toCorda()
         val (requestedProtocolName, initiatorVersionsSupported) = getProtocolInfo(sessionInit, sessionEvent)
         var initiatedFlowNameAndProtocol: FlowAndProtocolVersion? = null
+        var initiatedFlowNameAndProtocolResult: Result<FlowAndProtocolVersion>? = null
 
         checkpointInitializer.initialize(
             context.checkpoint,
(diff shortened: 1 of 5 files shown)


libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionErrorExecutor.kt
```diff
@@ -29,17 +29,21 @@ class SessionErrorExecutor(
...
+
+        private const val IGNORING = "ignoring"
+        private const val FORWARDING = "forwarding"
     }
 
     private val messageDirection = sessionEvent.messageDirection
     private val outputTopic = getSessionEventOutputTopic(messageDirection)
 
-    private val errorMsg = "Flow mapper received error event from counterparty for session which does not exist. " +
+    private val defaultMsg = "Received error event while in $flowMapperState, {} event. Key: $eventKey, Event: $sessionEvent"
```
LWogan: I think we should include the message direction in the error messages as they can be OUTBOUND or INBOUND which is key to understanding the source of the message. This can be retrieved from sessionEvent.messageDirection.

Other than that I'm happy with these changes!

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
```diff
@@ -115,37 +151,33 @@ class SessionEventHandler @Activate constructor(
             val protocolStore = try {
                 flowSandboxService.get(holdingIdentity, it).protocolStore
             } catch (e: Exception) {
-                // We assume that all sandbox creation failures are transient. This likely isn't true, but to handle
-                // it properly will need some changes to the exception handling to get the context elsewhere. Transient here
-                // will get the right failure eventually, so this is fine for now.
```
emilybowe, line 120: is this no longer a concern?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
```diff
@@ -115,37 +151,33 @@ class SessionEventHandler @Activate constructor(
...
                 .setCreatedTimestamp(Instant.now())
                 .build()
         }
 
-        //set initial session state, so it can be found when trying to send the confirmation message
-        context.checkpoint.putSessionState(initialSessionState)
-        context.flowMetrics.flowStarted()
-
-        sendConfirmMessage(initiatedFlowNameAndProtocol, requestedProtocolName, initiatorVersionsSupported, context, sessionId)
+        return initiatedFlowNameAndProtocolResult!!
```
emilybowe, line 177: if null this will just return "Invalid protocol" is that right?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/service/InitiateFlowRequestService.kt
```diff
@@ -96,9 +96,9 @@ class InitiateFlowRequestService @Activate constructor(
             }
         )
 
-        val sessionsNotInitiatedIds = sessionsNotInitiated.map { it.sessionId }
+        val sessionsNotInitiatedIds = sessionsNotInitiated.map { it.sessionId }.toSet()
         initiatingFlowStackItem.sessions
```
emilybowe, line 100: .toSet() to enforce uniqueness ?

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionErrorExecutor.kt
```diff
@@ -50,15 +55,15 @@ class SessionErrorExecutor(
         val sessionId = sessionEvent.sessionId
         return when (flowMapperState.status) {
             null -> {
-                log.warn("FlowMapperState with null status. Key: $eventKey, Event: $sessionEvent.")
+                log.debug("FlowMapperState with null status. Key: {}, Event: {}.", eventKey, sessionEvent)
```
emilybowe, line 58: why were some log.warns changed to debug level?

emilybowe (approved):
LGTM

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
ben-millar: So, what this returns is a `Result<FlowAndProtocolVersion>` which is like a wrapper on a call that may throw an exception.

So rather than having something like:
```
try {
    val result = doSomething()
} catch (e) {
    doSomethingElse(e)
}
```
and having to handle the result or exception immediately, we can us a runCatch block like so:
```
val result: Result<T> = runCatching { 
    doSomething()
}
```
and then we can do stuff with this `result` object like:
```
if (result.isSuccess()) {
    print(result.get())
} else if (result.isFailure()) {
    print(result.exception())
}
```

But the reason why I'm calling `!!` is because I initially declare the `Result<T>?` as a nullable type because it's not initialised until after the runCatch block runs, but by the end of the function I know it won't be null so I'm casting it back to a non-nullable `Result<T>` before returning it.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/service/InitiateFlowRequestService.kt
ben-millar: It's for efficiency because, in our filter expression, we're checking if `it.sessionId` is in `sessionsNotInitiatedIds` which is O(n) for a list and O(1) for a set. Similarly I've replaced the `map` function with `forEach` because they do the same thing, except `map` returns a copy of the modified list (which we just discard)

libs/flows/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionErrorExecutor.kt
ben-millar: That's just to line up with the logging level we use for other similar warnings (On Lorcan's suggestion :grin:)

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
ben-millar: Actually, I just removed it in refactoring and forgot to re-add it 😅 But actually I'm not sure if it's needed; if this causes issues we can track it as a bug in the future.

PR #3820, opened 2023-05-08
CORE-13419 Increase some retry intervals in the e2e tests

charlieR3:
Increase the intervals between retries when asserting for expected results from membership onboarding REST APIs. This is for both the e2e tests in the runtime-os and in the separate e2e test repo.

The default wait time between attempts for `eventually` is 100ms and for `assertWithRetry` it is 500ms. We don't expect registration to finish that fast so we can increase the wait time to reduce traffic during the tests.


vkolomeyko (approved):
LGTM

yift-r3 (approved):

PR #3830, opened 2023-05-09
CORE-13519 Exclude certificate OIDs and serialize X500 attributes deterministically on calculating signer summary hash 

kyriathar:
On calculating signer summary hash:
- filters x500 attributes to only keep attributes in "CN", "OU", "O", "L", "ST", "C"
- sorts x500 attributes in order of  "CN", "OU", "O", "L", "ST", "C" to deterministically serialize them regardless their passed in order


driessamyn (approved):
LGTM (approval from Marcello and James)

PR #3833, opened 2023-05-09
CORE-12208 Fix logic creating synthetic aliases

szymonsztuka:
Fix missing stop condition when adding synthetic aliases (was creating recursively Alias Alias...), fix interop integration tests.


ac101m (approved):
LGTM

PR #3834, opened 2023-05-09
CORE-8714 Set the Protocol and PKI mode after the InitiatorHandshake instead of the InitiatorHello

williamvigorr3:
We only know the initiator's identity after we have received the `InitiatorHandshake` message. The previous behaviour was looking for any identity in the same group and then using that to look up the group policy. This will cause problems once updating the parts of the group policy used by the Link Manager is supported. e.g. Reregistering and updating the protocol mode.

To do this I added a new step into the responders authentication protocol `VALIDATED_ENCRYPTED_EXTENSIONS`. This step comes after validating the `InititaorHandshake` message. I moved selecting and checking for a valid protocol mode and   validating the certificate from validating the initiator handshake message, to this new stage.

I left in the group policy lookup when processing the `InitiatorHello` message, this is only used to lookup the network type (Corda 5 or Corda 4). We could consider hard coding it to Corda 5

### DIFF
--- a/libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocolResponder.kt
+++ b/libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocolResponder.kt
@@ -31,6 +31,9 @@ import java.security.PublicKey
 import java.security.spec.X509EncodedKeySpec
 import java.time.Instant
 import javax.crypto.AEADBadTagException
+import net.corda.data.p2p.crypto.internal.InitiatorEncryptedExtensions
+import net.corda.data.p2p.gateway.certificates.RevocationCheckRequest
+import net.corda.data.p2p.gateway.certificates.RevocationCheckResponse
 import kotlin.math.min
 
 /**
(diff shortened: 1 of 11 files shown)


libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocolResponder.kt
```diff
@@ -207,21 +210,24 @@ class AuthenticationProtocolResponder(
         }
     }
 
-    private fun validateCertificate(
-        initiatorHandshakePayload: InitiatorHandshakePayload,
+    fun validateEncryptedExtensions(
```
dimosr, line 217: There's a bit of docs at the top about the methods to be called, worth updating it with the new method.

dimosr (commented):
LGTM - just a minor comment on docs

libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocolResponder.kt
williamvigorr3: I updated this and added doc for the new method.

dimosr (approved):

PR #3837, opened 2023-05-10
ES-379 Investigate use of unstable tags to skip certain tests on the main CI using Junit tags for e2e tests

ronanbrowne:
mark `MultiClusterDynamicNetworkTest`, `SessionCertificateTest`, `SingleClusterDynamicNetworkTest`, and `StaticNetworkTest` as _Unstable_ with junit Tags, this means they will be skipped by main CI / PR gates a separate "catch-all" job will be implemented to execute these.

Companion PR  https://github.com/corda/corda-shared-build-pipeline-steps/pull/766

### DIFF
--- a/applications/workers/release/rest-worker/src/e2eTest/kotlin/net/corda/applications/workers/rest/SingleClusterDynamicNetworkTest.kt
+++ b/applications/workers/release/rest-worker/src/e2eTest/kotlin/net/corda/applications/workers/rest/SingleClusterDynamicNetworkTest.kt
@@ -13,6 +13,7 @@ import net.corda.applications.workers.rest.utils.getMemberName
 import net.corda.applications.workers.rest.utils.onboardMembers
 import net.corda.applications.workers.rest.utils.onboardMgm
 import net.corda.data.identity.HoldingIdentity
+import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Tag
 import org.junit.jupiter.api.Test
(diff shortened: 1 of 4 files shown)


applications/workers/release/rest-worker/src/e2eTest/kotlin/net/corda/applications/workers/rest/SingleClusterDynamicNetworkTest.kt
```diff
@@ -42,7 +42,6 @@ class SingleClusterDynamicNetworkTest {
     (as this one will cover that use case as well)
     To run it locally while disabled follow the instruction in resources/RunP2PTest.md:
      */
-    @Disabled("Is disabled and can be run manually until CORE-6079 is complete.")
```
charlieR3: This test is actually disabled for a reason other than flakiness so this is expected to fail consistently rather than just being unstable. Best to leave this annotation here. The other test in this class is the flaky one and is one of the tests we want the unstable tag for.

charlieR3 (dismissed):

ronanbrowne:
Need to look into this more before merging, I see a SessionCertificateTest test executing which I would not expect on PR

ronanbrowne:
On further testing annotation needed on test level not class level for expected results

charlieR3 (approved):

vkolomeyko (approved):
LGTM

PR #3838, opened 2023-05-10
CORE-13373: Fix flakiness in sandbox cache eviction tests.

chrisr3:
Caffiene must be running the `removalListener` asynchronously, even after an explicit "remove". Wait until the sandbox has been added to the "expiry" queue before verifying the mock listener.


fenryka (approved):
LGTM

PR #3839, opened 2023-05-10
NOTICK Recurring merge from main onto interop - API version 1

szymonsztuka:
No conflicts, Corda version change from 5.0.0 to 5.1.0. In our case it's from `5.0.0-INTEROP` to `5.1.0-INTEROP`


.ci/e2eTests/JenkinsfileCombinedWorker
```diff
@@ -85,7 +85,7 @@ pipeline {
         } 			
         stage('start combined worker') {       
             environment {
-                JAR_PATH = "${env.WORKSPACE}/applications/workers/release/combined-worker/build/bin/corda-combined-worker-5.0.0-INTEROP.0-*.jar"
+                JAR_PATH = "${env.WORKSPACE}/applications/workers/release/combined-worker/build/bin/corda-combined-worker-5.1.0-INTEROP.0-*.jar"
```
szymonsztuka, line 88: This is a difference to main, in the may the string `corda-combined-worker-5.0.0.0-*.jar` was renames to `corda-combined-worker-*.jar`, however I think we need to mantain "-INTEROP" part of the label.

PR #3841, opened 2023-05-10
CORE-9673  Updating error handling for serialize

pmck91:
various places in the code expect serialization errors to return null. This isn't possible unless a null string is passed in, every other case will throw a CordaRuntimeException. 

This change will wrap all of the cases where null is expected in a try catch to throw the same exception type, as well as pass more information about what caused the error. In the other cases the system will catch the RuntimeException in the same way it currently does.


components/membership/membership-p2p/src/main/kotlin/net/corda/membership/p2p/helpers/P2pRecordsFactory.kt
```diff
@@ -59,10 +59,14 @@ class P2pRecordsFactory(
         id: String = UUID.randomUUID().toString(),
         filter: MembershipStatusFilter = MembershipStatusFilter.ACTIVE,
     ): Record<String, AppMessage> {
-        val data = cordaAvroSerializationFactory.createAvroSerializer<T> {
-            logger.warn("Serialization failed")
-        }.serialize(content)
-            ?: throw CordaRuntimeException("Could not serialize $content")
+        val data = try {
```
LWogan: it would be nice if we could move this pattern to a helper method shareable. perhaps in SerializationUtils.kt in libs:utilies?

this is some psuedo code i have not tested. Just copying an idea i saw in PersistenceUtils.kt in flow-service project

inline fun <T> wrapWithNullErrorHandling(message: String, function: () -> T?): T {
    val x = try {
        function()
    } catch (e: CordaRuntimeException) {
        throw CordaRuntimeException(message, e)
    } ?: throw CordaRuntimeException(message)
    return x
}

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
```diff
@@ -46,9 +47,13 @@ internal class AddNotaryToGroupParametersHandler(
...
     private fun serializeProperties(context: KeyValuePairList): ByteArray {
-        return keyValuePairListSerializer.serialize(context) ?: throw MembershipPersistenceException(
-            "Failed to serialize key value pair list."
-        )
+        return try {
+            keyValuePairListSerializer.serialize(context) ?: throw MembershipPersistenceException(
+                "Failed to serialize key value pair list."
+            )
+        } catch (ex: CordaRuntimeException) {
+            throw MembershipPersistenceException("Failed to serialize key value pair list.", ex)
```
thiagoviana: "Failed to serialize key value pair list."
 
Shouldn't we put this into constants / strings file?
Is there already a place for them?

components/membership/membership-p2p/src/main/kotlin/net/corda/membership/p2p/helpers/P2pRecordsFactory.kt
pmck91: great idea ill do that now

libs/utilities/src/main/kotlin/net/corda/utilities/serialization/SerializationUtils.kt
```diff
@@ -27,3 +28,22 @@ inline fun <reified T : Any> SerializationService.deserialize(serializedBytes: S
...
+ *
+ * @param T The return type expected
+ * @param E The exception type to throw
+ * @param message The message to pass tot the exception on failure
+ * @param exClazz The exception class
+ * @param f the block to wrap
+ * @return type T
+ */
+inline fun <reified T: Any, E: Exception>wrapWithNullErrorHandling(message: String, exClazz: Class<E>, f: () -> T?): T = try {
+    f()?: throw exClazz.getDeclaredConstructor().newInstance(message)
```
mbrkic-r3: Does serialization also return `null` in case of errors? If yes, why it's not changed to always throw exception? In that case checking for `null` value wouldn't be needed here.

Instead of providing message and class for wrapping exception, why not providing a factory instead (lambda)?

pmck91: no, it always throws, the only time you get null is if you pass a null value. I didn't change that behaviour because it made sense to me. A null didn't fail to serialize because of an error, but because nulls don't need serialization. I think theres use cases where you may want a null value, like the dead letter queue perhaps? 

This is the code that returns null:

`return when (data) {
            null -> null
            else -> serialize(data)
 }`

there are some usages that take the null and dont throw an exception, MerkleTreeGenerator seems to just accept a possible null and then filter it out of the results.

Can you elaborate more on the factory?

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
pmck91: There doesnt seem to be a standardised place, I can see @LWogan created a constants file in the chunking package that has some error messages, perhaps each module could have something similar to this?

LWogan: in R3 I've noticed it hasn't been standard practice to have constant classes. I've put them in occasionally when appropriate.

If you want to do it here, you could add them to the  SerializationUtils.kt file  (maybe?)  given its already a dependency at each call site already. I dont have a strong opinion on it.

libs/utilities/src/main/kotlin/net/corda/utilities/serialization/SerializationUtils.kt
mbrkic-r3: Something like 
`wrapWithNullErrorHandling(exception: (cause: Exception) -> Exception, f: () -> T?)`

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
mbrkic-r3: Too me it seems that having inconsistent approach across the code base is a bigger issue at the moment. We might do a lots of work doing this with every PR and then later we might decide to have a different approach company wide. I would leave it for now until we have clear guidelines company wide.

libs/utilities/src/main/kotlin/net/corda/utilities/serialization/SerializationUtils.kt
mbrkic-r3: So calling it would look like:
```
wrapWithNullErrorHandling(
   { MembershipPersistenceException("Failed to serialize key value pair list.", it) }
) {
    keyValuePairListSerializer.serialize(context)
}
```

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
thiagoviana: I do agree with @mbrkic-r3. Get back to this in amore consistent way, defining something system-wide.

I'd say that even Util classes should be avoided when possible, once they are an easy path to not find/create a proper place for our code to live in.

libs/utilities/src/main/kotlin/net/corda/utilities/serialization/SerializationUtils.kt
pmck91: Nice, Ive made that change now, works great

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/AddNotaryToGroupParametersHandler.kt
pmck91: I agree too, maybe a ticket can be created to investigate standardization of error messages per module?

mbrkic-r3 (approved):
LGTM

charlieR3 (approved):

thiagoviana (approved):

PR #3842, opened 2023-05-10
CORE-13562: Use pool for the entety factory

yift-r3:


components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/HandlerFactories.kt
```diff
@@ -103,6 +107,11 @@ internal class HandlerFactories(
     }
 
     fun handle(request: MembershipPersistenceRequest): Any? {
-        return getHandler(request.request::class.java).invoke(request.context, request.request)
+        val start = System.currentTimeMillis()
+        return getHandler(
+            request.request::class.java
+        ).invoke(request.context, request.request).also {
+            logger.info("QQQ operation took: ${System.currentTimeMillis() - start}")
+        }
```
chrisr3, line 115: `System.currentTimeMillis()` is not reliable for interval timing - especially short intervals. You should use `System.nanoTime()` instead.

PR #3843, opened 2023-05-10
ES-371/Update license to Apache

seanbrereton:
As part of the C5 build process a license file is injected into the pom.xml of a produced Jar this license file is specified as part of the gradle.properties of each repo the current license properties are not correct / outdated and should be updated pre GA


ronanbrowne:
Discussed with @sachingokhale and @marcellofarina75 OKd from product for entering post freeze - needed for correct Pom generation pre release

ronanbrowne (approved):

PR #3844, opened 2023-05-10
CORE-9961: Add flow validation in pipeline for flow start operational status

Tom-Fitzpatrick:
This adds validation to `StartFlowEventHandler` to check that a virtualNode's `flowStartOperationalStatus` is ACTIVE. If not, a `FlowMarkedForKill` exception is thrown and the flow is killed. 

A test was added to `StartFlowTest` to cover this scenario.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/StartFlowEventHandler.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/StartFlowEventHandler.kt
@@ -31,16 +31,6 @@ class StartFlowEventHandler @Activate constructor(
     override fun preProcess(context: FlowEventContext<StartFlow>): FlowEventContext<StartFlow> {
         log.info("Flow [${context.checkpoint.flowId}] started")
 
-        checkpointInitializer.initialize(
-            context.checkpoint,
-            WaitingFor(WaitingForStartFlow),
-            context.inputEventPayload.startContext.identity.toCorda(),
-        ) {
-            context.inputEventPayload.startContext
-        }
-
-        context.flowMetrics.flowStarted()
-
         val holdingIdentity = context.inputEventPayload.startContext.identity.toCorda()
         val virtualNodeInfo = virtualNodeInfoReadService.get(holdingIdentity)
 
@@ -50,6 +40,16 @@ class StartFlowEventHandler @Activate constructor(
                         "shortHash ${holdingIdentity.shortHash}"
             )
         }
+
+        checkpointInitializer.initialize(
+            context.checkpoint,
+            WaitingFor(WaitingForStartFlow),
+            context.inputEventPayload.startContext.identity.toCorda(),
+        ) {
+            context.inputEventPayload.startContext
+        }
+
+        context.flowMetrics.flowStarted()
 
         return context
     }
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/StartFlowEventHandler.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/StartFlowEventHandler.kt
@@ -31,16 +31,6 @@ class StartFlowEventHandler @Activate constructor(
     override fun preProcess(context: FlowEventContext<StartFlow>): FlowEventContext<StartFlow> {
         log.info("Flow [${context.checkpoint.flowId}] started")
 
-        val holdingIdentity = context.inputEventPayload.startContext.identity.toCorda()
-        val virtualNodeInfo = virtualNodeInfoReadService.get(holdingIdentity)
-
-        if (virtualNodeInfo?.flowStartOperationalStatus == OperationalStatus.INACTIVE) {
-            throw FlowMarkedForKillException(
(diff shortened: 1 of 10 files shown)


josephzunigadaly (dismissed):

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/StartFlowEventHandler.kt
```diff
@@ -25,13 +30,25 @@ class StartFlowEventHandler @Activate constructor(
...
             WaitingFor(WaitingForStartFlow),
             context.inputEventPayload.startContext.identity.toCorda(),
         ) {
             context.inputEventPayload.startContext
         }
+
+        val holdingIdentity = context.inputEventPayload.startContext.identity.toCorda()
+        val virtualNodeInfo = virtualNodeInfoReadService.get(holdingIdentity)
+
+        if (virtualNodeInfo?.flowStartOperationalStatus == OperationalStatus.INACTIVE) {
```
LWogan, line 37: This looks like the same logic as what's in the virtualNodeFlowOperationalChecks() step of FlowEventPipelineImpl. This step is executed every time a flow event is processed.

I don't think we also need it in StartFlowEventHandler.kt?

LWogan: I misread it. flowStartOperationalStatus  versus flowOperationalStatus.

I can see why you'd put it in StartFlowEventHandler.kt. But I feel like we should move this validation into the virtualNodeFlowOperationalChecks() step instead of the start flow handler. As thats what its for. That way we can keep all the checks in one place. 

I'm not familiar with what other operational checks need to be done, or how complex the conditionals will get. Perhaps you could have a simple condition/switch off the input payload type for now, and in the future if that gets too complex we could factory some event operational service based on the input payload instead.

Tom-Fitzpatrick: So we chatted a little offline about this and the decision was to leave this check in the `StartFlowEventHandler` as we can actually validate before the checkpoint is created and save on creating it at all when it will be killed anyway. Also, there will likely be some more operational checks added in the future that won't fit into `virtualNodeFlowOperationalChecks` either, so putting this one there at the moment wouldn't be worth it only for the sake of consistency with those checks. 

So I have moved the validation to before we create the flow checkpoint in `StartFlowEventHandler`, and added some logic to be able to create a `FlowStatusRecord` in `FlowEventExceptionProcessorImpl` without a checkpoint.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
```diff
@@ -259,6 +265,29 @@ class FlowEventExceptionProcessorImpl @Activate constructor(
...
+            is StartFlow -> {
+                val status = FlowStatus().apply{
+                    key = inputPayload.startContext.statusKey
+                    flowStatus = FlowStates.KILLED
+                    this.flowId = flowId
+                    processingTerminatedReason = message
+                }
+                listOf(flowRecordFactory.createFlowStatusRecord(status))
+            }
+            else -> createFlowKilledStatusRecord(
```
Tom-Fitzpatrick, line 285: If the InputEvent payload is a StartFlow, we take the FlowKey from it and use it to build the FlowKilledStatusRecord. For other types, we're defaulting to the previous approach, which will result in a blank FlowStatus when the `FlowMarkedForKillException` was thrown before the flow checkpoint is initialized.

I looked into adding handling for other types, for example, when the InputEvent payload is a SessionInit, but there was no flow key available on the object, so I wasn't able to create a valid FlowKilledStatusRecord.

I'm not sure if there is a better approach for these scenarios now or maybe we should create a follow up task to look at handling these other cases better.

josephzunigadaly (approved):

vkolomeyko (approved):
LGTM

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
ben-millar: It could be worth creating a follow up task to look into whether or not we could make this more descriptive in other (non StartFlow) cases.

ben-millar (approved):
```
  _    ___ _____ __  __ 
 | |  / __|_   _|  \/  |
 | |_| (_ | | | | |\/| |
 |____\___| |_| |_|  |_|
```

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
Tom-Fitzpatrick: I created a ticket here to capture this: https://r3-cev.atlassian.net/browse/CORE-15263

Let me know if this isn't the right place or more info is needed. 

@ben-millar @LWogan @thiagoviana

PR #3845, opened 2023-05-10
CORE-13593: Register Kryo's CordaClosureSerializer correctly to serialize Java lambdas.

chrisr3:
We have not registered `CordaClosureSerializer` correctly, and so Kryo never attempts to use it for classes like:
```kotlin
class UtxoReceiveFinalityFlow(
    private val session: FlowSession,
    private val validator: UtxoTransactionValidator
) : SubFlow<UtxoSignedTransaction> {
```
where `validator` has a lambda value at runtime.

Register `CordaClosureSerializer` correctly. Also remove an unnecessary field from the `parkAndCustomSerialize` lambda.


rick-r3:
image
This is the original problem.... if the `ClosureSerializer` was not correctly registered, the `ReplaceableObjectKryo` when it writes the class for a lambda would actually cause a `FieldSerializer` for the lambda class to be built, even though it wasn't going to use the serializer because `ReplaceableObjectKryo` would intercept because lambdas are "replaceable".

rick-r3 (approved):

PR #3852, opened 2023-05-11
CORE-10100: AWS KMS Integration - copy JCE Provider classes

anien:
This is the first step of creating an internal tool (modified `corda-cli`) to allow signing CPx with the AWS KMS key.
JCE Provider (https://github.com/aws-samples/aws-kms-jce) contains helpful classes needed to interact with AWS KMS.

Changes:
* JCE Java classes are copied to runtime-os repo
* Lombok dependencies are removed from JCE Java classes

### DIFF
--- a/tools/plugins/package/src/main/java/net/corda/cli/plugins/packaging/aws/kms/KmsKeyStore.java
+++ b/tools/plugins/package/src/main/java/net/corda/cli/plugins/packaging/aws/kms/KmsKeyStore.java
@@ -1,0 +1,139 @@
+package software.amazon.awssdk.services.kms.jce.provider;
+
+import software.amazon.awssdk.services.kms.jce.provider.ec.KmsECKeyFactory;
+import software.amazon.awssdk.services.kms.jce.provider.rsa.KmsRSAKeyFactory;
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+import software.amazon.awssdk.services.kms.KmsClient;
+import software.amazon.awssdk.services.kms.model.AliasListEntry;
+import software.amazon.awssdk.services.kms.model.DescribeKeyResponse;
+import software.amazon.awssdk.services.kms.model.ListAliasesRequest;
+import software.amazon.awssdk.services.kms.model.ListAliasesResponse;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.security.*;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateException;
+import java.util.*;
+import java.util.stream.Collectors;
+
+@RequiredArgsConstructor
+public class KmsKeyStore extends KeyStoreSpi {
+
+    @NonNull
+    private final KmsClient kmsClient;
+
+    @Override
+    public Enumeration<String> engineAliases() {
+        return Collections.enumeration(getAliases().stream().map(AliasListEntry::aliasName).collect(Collectors.toSet()));
+    }
+
+    @Override
+    public boolean engineContainsAlias(String alias) {
+        String prefixedAlias = getPrefixedAlias(alias);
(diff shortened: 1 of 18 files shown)


tools/plugins/package/src/main/java/net/corda/cli/plugins/packaging/aws/kms/KmsKeyStore.java
```diff
@@ -0,0 +1,151 @@
...
+        return getAliases().stream().filter(e -> e.aliasName().equals(prefixedAlias)).findAny().isPresent();
+    }
+
+    @Override
+    public int engineSize() {
+        return getAliases().size();
+    }
+
+    @Override
+    public Key engineGetKey(String alias, char[] chars) throws NoSuchAlgorithmException, UnrecoverableKeyException {
```
driessamyn, line 57: It looks like this is not used, which sounds correct as we never want to extract the key from the keystore. So maybe we should throw a NotSupported exception here instead of keeping the implementation. I realise that this would divert the code from the original, but it would prevent somebody accidentally making a change that would involve extracting the key.

dickon:
~Is there a reason to use Java rather than Kotlin for this?~ nevermind, the upstream code is Java

dickon:
How much did we change in the AWS KMS SPI code? I'd generally rather we fork and try to merge back any changes we need when we can't use the AWS github code as is, and only copy the code into our repo if there's no other choice and we've agreed it within the company.

In this case the upstream code is MIT no-attribution licensed so there's not a GPL contamination risk.

driessamyn:
> How much did we change in the AWS KMS SPI code? I'd generally rather we fork and try to merge back any changes we need when we can't use the AWS github code as is, and only copy the code into our repo if there's no other choice and we've agreed it within the company.
> 
> In this case the upstream code is MIT no-attribution licensed so there's not a GPL contamination risk.

The source repo is explicitly marked as an example, so I don't think forking makes much sense. I would, for example, prefer we remove everything we don't need (not necessarily from this PR, but once we productionise this).
We have already too many things like Quasar where we've become accidental maintainers even though we only use a small part.
(BTW. I suggested to Anna we copy/paste the code)

anien:
> How much did we change in the AWS KMS SPI code? I'd generally rather we fork and try to merge back any changes we need when we can't use the AWS github code as is, and only copy the code into our repo if there's no other choice and we've agreed it within the company.
> 
> In this case the upstream code is MIT no-attribution licensed so there's not a GPL contamination risk.

As Dries mentioned, I asked first how to approach using this repo and copying it here was suggested. 

Re changes, I deliberately made first commit just after copying the repo, so you can check what all has changed.
But in general I have only removed Lombok dependency - so had to create few getters, constructors, deal with `@NonNull` (add Jetbrains annotation plus `Objects.requireNonNull` checks), add one builder, add throw exceptions.

dickon:
Oh, since it is an example we are deriving from that's a bit different and maybe this PR is the right thing to do. As to Dries' point of having stuff we don't need, probably yes in the long run although I'm fine with working in small steps and merging with them if the idea is to end up with something as simple as possible.

dickon (approved):
LGTM

kyriathar:
Should we be adding this code under a dedicated module to be better contained conceptually instead of adding it in `:tools:plugins:package`?

driessamyn:
> Should we be adding this code under a dedicated module to be better contained conceptually instead of adding it in `:tools:plugins:package`?

Let's keep it in here until there is a good reason to move it.
It's already in a different package, and we really need to reduce our approx 330 modules not increase.

tools/plugins/package/src/main/java/net/corda/cli/plugins/packaging/aws/kms/KmsKeyStore.java
anien: If I understand the code correctly, it never retrieves the key from the keystore/KMS. It creates a `Kms[EC/RSA]PrivateKey` object that holds only `keyId`.

PR #3858, opened 2023-05-12
CORE-13225: Use an atomic reference for the Publisher in CpiInfoWriterComponentImpl

fowlerrr:
This should avoid concurrency issues when using a Publisher around the time it's recreated.


driessamyn (approved):
LGTM
As discussed and captured in CORE-13689, we should look at handling this in a generic way, but I agree with applying this local change for now.

PR #3861, opened 2023-05-12
CORE-13373: Fix flaky sandbox cache eviction test.

chrisr3:
Caffeine fires its `removalListener` asynchronously, so we need to wait for it.


mnesbit (approved):
OK

PR #3862, opened 2023-05-12
CORE-13593: Restore default registration of SerializedLambda serializer.

chrisr3:
Ensure that Kryo never needs to create a serializer for `SerializedLambda` "on the fly".


rick-r3 (approved):
On the basis this makes yesterday's fix "smaller", approved.

PR #3863, opened 2023-05-12
CORE-10453 : Remote alias look up service 

SaurabhR3:
Info for this change :

Interop client flow will inject or use RemoteAliasLookUpService to fetch the required interop information.
RemoteAliasLookUpService will be loaded at the time of instantiation for interop worker/ flow worker.
Internally RemoteAliasLookUpService will use the cache of AliasIdentityInfo.
Cache will contain the data of alias group info with holding identity short hash as a key.
RemoteAliasLookUpService will only fetch the alias group info for specific short hash and it will have cache entry for that short hash.
Interop client flow can query its own cached entry.

https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4448321990/WIP+Alias+Identities+at+Flow+Level

Changes:

1. Created injectable RemoteAliasLookUpService
2. Implemented hardcoded RemoteAliasLookUpService
3. Injected the service in FacadeInvocationFlow for logging purpose


components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpServiceImpl.kt
```diff
@@ -0,0 +1,34 @@
...
+import net.corda.v5.application.interop.InteropGroupInfo
+import net.corda.v5.application.interop.RemoteAliasLookUpService
+import net.corda.v5.serialization.SingletonSerializeAsToken
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
+
+@Component(service = [RemoteAliasLookUpService::class, UsedByFlow::class], scope = PROTOTYPE)
+class RemoteAliasLookUpServiceImpl @Activate constructor() : RemoteAliasLookUpService, UsedByFlow,
+    SingletonSerializeAsToken {
```
yarivus:
```suggestion
class RemoteAliasLookUpServiceImpl @Activate constructor() :
    RemoteAliasLookUpService, UsedByFlow, SingletonSerializeAsToken {
```
I think it's more readable to have all the inherited in a single newline.

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
```diff
@@ -4,6 +4,8 @@ import net.corda.v5.application.flows.ClientStartableFlow
 import net.corda.v5.application.flows.CordaInject
 import net.corda.v5.application.flows.InitiatingFlow
 import net.corda.v5.application.flows.ClientRequestBody
+import net.corda.v5.application.interop.RemoteAliasLookUpService
+//import net.corda.v5.application.interop.RemoteAliasLookUpService
```
yarivus: remove comment - it's duplicate

yarivus (changes requested):

yarivus (approved):
LGTM - minor style comment left

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
SaurabhR3: Fixed in 2402153

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpServiceImpl.kt
SaurabhR3: Fixed in 7c369e5

yarivus (approved):
Even Better - LGTM

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
```diff
@@ -30,6 +34,14 @@ class FacadeInvocationFlow : ClientStartableFlow {
     override fun call(requestBody: ClientRequestBody): String {
         log.info("FacadeInvocationFlow.call() starting")
 
+        val alice = "CN=Alice Gold Alias, O=Alice Corp, L=LDN, C=GB";
```
szymonsztuka: This value should not be hard-coded in the flow.

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
```diff
@@ -30,6 +34,14 @@ class FacadeInvocationFlow : ClientStartableFlow {
...
+        val cpiName = "Gold.cpi"
```
szymonsztuka: This value should not be hard-coded in the flow.

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
```diff
@@ -30,6 +34,14 @@ class FacadeInvocationFlow : ClientStartableFlow {
...
+        val aliasMemberInfo = remoteAliasLookUpService.lookup(alice, cpiName)
+        log.info("AliasMemberInfo for Alice Gold Alias : $aliasMemberInfo")
+
+        val aliasMembers = remoteAliasLookUpService.lookup("org.corda.interop/platform/tokens/v1.0")
+        log.info("AliasMemberInfo list for facadeId : $aliasMembers")
```
szymonsztuka: The result of the lookup has no impact on the flow/ is not use later on. In this case why to add this code?

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
```diff
@@ -0,0 +1,117 @@
...
+
+@Component(service = [AliasIdentityDataService::class])
+class HardcodedAliasIdentityDataServiceImpl @Activate constructor() : AliasIdentityDataService {
+    override fun getAliasIdentityData(): MutableList<InteropGroupInfo> {
+        val aliceGroups = listOf(
+            GroupData(
+                "3dfc0aae-be7c-44c2-aa4f-4d0d7145cf08", //This is hardcoded groupId for existing texts.
+                "Interop-Gold-Silver-Bronze-Group", listOf(
+                    NetworkFacadeData(
+                        "Gold", "Gold.cpi", listOf(
```
jmacmahonr3: It wouldn't be gold.cpi as that's the file name. It's more likely to be "Gold Trading.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpServiceImpl.kt
```diff
@@ -0,0 +1,30 @@
...
+import net.corda.sandbox.type.UsedByFlow
+import net.corda.v5.application.interop.RemoteAliasLookUpService
+import net.corda.v5.interop.AliasMemberInfo
+import net.corda.v5.serialization.SingletonSerializeAsToken
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
+
+@Component(service = [RemoteAliasLookUpService::class, UsedByFlow::class], scope = PROTOTYPE)
+class RemoteAliasLookUpServiceImpl @Activate constructor() :
```
jmacmahonr3: I'd suggest renaming to just RemoteAliasLookUp

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
jmacmahonr3: As mentioned earlier cpi name could be multiple words in a String like "Big Bank Gold Trading"

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpServiceImpl.kt
jmacmahonr3: Also can we write a component test for it without using the facade invocation flow ?

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
SaurabhR3: There are following 2 reasons
1. Client flow can inject the newly created injectable service . 
2. Logging the result confirms that client flow can make a use of its api.

jmacmahonr3 (changes requested):

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
szymonsztuka: You are using this code as a pseudo unit test, nothing else. Please remove this code form the flow or make it contribute to the business logic.

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
SaurabhR3: Fixed in d05b965

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
SaurabhR3: Fixed in d05b965

yarivus (approved):
LGTM

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
SaurabhR3: It's not the unit test, this is to validate that can service be injected in client flow? and can client flow use the API of injected service? This can not be tested in Unit test. It needs a client flow.
If we remove this code how can we prove that the `RemoteAliasLookUpService` is an `Injectable Service` for client flow?

szymonsztuka: We know we can inject `RemoteAliasLookUpService` because it's an injectable service as any other services, or even this code line `@CordaInject lateinit var remoteAliasLookUpService: RemoteAliasLookUpService` proves we can inject. However the actuall usage of the service here doesn't add anything to the business logic of the flow (the result of a call is not use to invoke a facade).

SaurabhR3: Fixed in 215a0d5

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
SaurabhR3: Fixed in 215a0d5

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpServiceImpl.kt
SaurabhR3: Comment 1. Fixed in 215a0d5
Comment 2. Working on it.

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
SaurabhR3: Fixed in 215a0d5

yarivus (approved):
LGTM

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
+
+@Component(service = [RemoteAliasLookUp::class, UsedByFlow::class], scope = PROTOTYPE)
+class RemoteAliasLookUpImpl @Activate constructor() :
+    RemoteAliasLookUp, UsedByFlow, SingletonSerializeAsToken {
+
+    override fun lookup(x500Name : String, cpiName: String): AliasMemberInfo? {
+        return getAliasMemberData().firstOrNull { memberInfo ->
+            memberInfo.identifier.equals(
```
simon-johnson-r3: Is this a Java-ism? In Kotlin you should use the structural equality operator == otherwise future maintainers will think you are doing something special here - unless you are doing something special, but then I'm wondering what.

simon-johnson-r3 (changes requested):

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
szymonsztuka: What if we have more entries matching the filter, then why we are taking the first one and not the second one? 
Also `getAliasMemberData` doesn't guarantee any stable ordering, at least nothing is described about this, so it could happen the same invocation could return a different entry each time.

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
```diff
@@ -39,7 +39,6 @@ class FacadeInvocationFlow : ClientStartableFlow {
         val payload = getArgument(args, "payload")
 
         log.info("Calling facade method '$methodName@$facadeId' with payload '$payload' to $alias")
-
```
szymonsztuka, line 42: Undo line removal.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
```diff
@@ -0,0 +1,30 @@
...
+                "$x500Name@$cpiName"
+            )
+        }
+    }
+
+    override fun lookup(facadeId: String?): List<AliasMemberInfo> {
```
szymonsztuka, line 22: Just a suggestion as this is more about the design, if a client code really need to get all aliases supporting the given facade , then it can use the latter function `getAliasMemberData()` and filter manually, exactly what is in the body of this method. We don't need to complicate API.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
```diff
@@ -0,0 +1,117 @@
...
+import net.corda.interop.service.AliasIdentityDataService
+import net.corda.v5.interop.AliasMemberInfo
+import net.corda.v5.interop.HoldingIdAliasGroupInfo
+import net.corda.v5.interop.InteropGroupInfo
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import java.util.UUID
+
+@Component(service = [AliasIdentityDataService::class])
+class HardcodedAliasIdentityDataServiceImpl @Activate constructor() : AliasIdentityDataService {
```
szymonsztuka: Why this service is separated from  `RemoteAliasLookUpImpl` ? Seems like the only responsibility for this service is to store hardcoded data, then it can be just moved to a private method of 'RemoteAliasLookupImpl'.
Other note, the hardcoded data is never used, why bother to have this data then?

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
```diff
@@ -0,0 +1,117 @@
...
+    override fun getGroupName(): String {
+        return groupName
+    }
+
+    override fun getMembers(): MutableList<AliasMemberInfo> {
+        return networks.map { DummyAliasMemberInfo(member, it.network, groupName, it.cpiName, it.facadeIds) }.toMutableList()
+    }
+}
+
+data class DummyAliasMemberInfo(
```
szymonsztuka: If this is a temporary class, prefer using anonymous instances of an interface. Alos does it mean we do have only an interface `AliasMemberInfo` and we don't have any proper (aka not Dummy" in name ) call implementing those interface.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
```diff
@@ -0,0 +1,117 @@
...
+                        )
+                    )
+                )
+            )
+        )
+        return HoldingIdAliasGroupInfoImpl("Bob", aliceGroups).groups
+    }
+}
+
+data class GroupData(
```
szymonsztuka: This (and the next class) seem like should be in API, so how Cordapps can use it? I'm not sure is there different API for Cordapp and these classes are used only for InteropService?

szymonsztuka (changes requested):

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
szymonsztuka: I've amending my previous comment, after thinking, i know that I ask to remove a sample usage from a flow, and now I'm complaining about no using the code in any flow. What about creating a new flow in the existing Cordapp to use this new API with hard-coded data, the result of the flow could return data from AliasMemberInfo.  How this approach is better than adding to the existing flow - it's clearly separates the usage, then new flow aim is just to use service with hardcoded data to output it, while the same new service has now real use (for now) in the existing flow:
```
class AliasLookupTestFlow : ClientStartableFlow {

    @CordaInject
    lateinit var aliasLookup:  ...type


    @Suspendable
    override fun call(requestBody: ClientRequestBody): String {
        
        val args = requestBody.getRequestBodyAsMap(jsonMarshallingService, String::class.java, String::class.java)

        val interopGroupId = getArgument(args, "interopGroupId")
        val facadeId = getArgument(args, "facadeId")
        val alias = MemberX500Name.parse(getArgument(args, "alias"))
        //..whaterver is needed 

        val result = aliasLookup.lookup() 

        return result.toString() //If toString() doesn't return anything nice consider creating manually string with some fileds from result
    }
}
```

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
SaurabhR3: @szymonsztuka We are comparing value against the `identifier` and this `identifier` will be unique. This unique constraint we need to add at the time of alias registration. We are going to develop the alias registration in upcoming milestone.  So here we can say that only one Alias member info will be available and there won't be second.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
SaurabhR3: We discussed this and agree on this in our deep dive session hence I have added this API.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
SaurabhR3: Actually I would avoid to create the new cordapp for such reason. Ultimately what we need is to call API, after that whether you print it as a log or return it to caller does not make much difference. According to me both are not adding any value.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
SaurabhR3: I am not a fan of anonymous instances, How about changing the name from `Dummy` to `Fake` or `AliasInfo`?
Bear in mind that these classes will be removed soon.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
SaurabhR3: All the classes in `HardcodedAliasIdentityServiceImpl` are temporary and just to create the fake data. They are not going to be used by any other service.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
SaurabhR3: I have removed Dummy from classes

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
szymonsztuka: I don't want to create a cordapp, just add a new flow (_What about creating a new flow in the **existing** Cordapp ..._)

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
szymonsztuka: The whole service  `HardcodedAliasIdentityServiceImpl` looks like temporary service, so maybe for simplicity just move the method of the hardcoded data to `RemoteAliasLookUpImpl`. We do have some temporary services in our code, however I think most likely if created if you have a proper interface, but tactically if don't have the desierd implementation yet, or the temporary service requires to be a service to achieve it's role (like AliasProcessor is quasi-temporary, but it listens to a Kafka queue, so it needs to exists, and maybe it couldn't be refactored to a method or it would complicate other code).

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
```diff
@@ -0,0 +1,117 @@
...
+
+import net.corda.interop.service.AliasIdentityDataService
+import net.corda.v5.interop.AliasMemberInfo
+import net.corda.v5.interop.HoldingIdAliasGroupInfo
+import net.corda.v5.interop.InteropGroupInfo
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import java.util.UUID
+
+@Component(service = [AliasIdentityDataService::class])
```
szymonsztuka: `HardcodedAliasIdentityDataServiceImpl` is not used in the code as a Corda service, we can remove this annotation.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
```diff
@@ -0,0 +1,28 @@
...
+            it.identifier == "$x500Name@$cpiName"
+        }
+    }
+
+    override fun lookup(facadeId: String?): List<AliasMemberInfo> {
+        return getAliasMemberData().filter { it.facadeIds.contains(facadeId) }
+    }
+
+    private fun getAliasMemberData(): List<AliasMemberInfo> {
+        return HardcodedAliasIdentityDataServiceImpl().getAliasIdentityData().flatMap { it.members }.toList()
```
szymonsztuka: What about just hard-coding the result here (the list),  it would reduce number of classes and fake services (`HardcodedAliasIdentityDataServiceImpl`).

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
```diff
@@ -0,0 +1,117 @@
...
+    val cpiName: String,
+    val facadeIds: List<String>
+)
+
+data class HoldingIdAliasGroupInfoImpl(
+    private val member: String,
+    val groupData: List<GroupData>
+) : HoldingIdAliasGroupInfo {
+    override fun getShortHash(): String {
+        return "ANYSHORTHASH"
```
szymonsztuka: This is not a hardcoded data class, we should not have  `return "ANYSHORTHASH"`, we do need a field or a code to compute a hash from other existing field.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
```diff
@@ -0,0 +1,117 @@
...
+    override fun getGroupId(): UUID {
+        return UUID.fromString(groupId)
+    }
+
+    override fun getGroupName(): String {
+        return groupName
+    }
+
+    override fun getMembers(): MutableList<AliasMemberInfo> {
+        return networks.map { AliasMember(member, it.network, groupName, it.cpiName, it.facadeIds) }.toMutableList()
```
szymonsztuka: Maybe better returning immutable list, or if these are unique it could be a set (however probably in Corda we generally do use sets?), so it can be a list.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
SaurabhR3: I did it because of following reason.

1. I don't want to create the hard coded data in RemoteAliasLookUpImpl as it is not its responsibility. Its responsibility is just to filter the data as per API.
2. When we will fetch the real data from cache, we will need one service, for that reason I have created AliasIdentityDataService
3. From the class name `HardcodedAliasIdentityDataServiceImpl` you guess the purpose.

Even if we try to hard code the result there, we need to either create the anonymous instances or data classes in RemoteAliasLookUpImpl, I personally feel it does not look good, just for reducing the some classes.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
```diff
@@ -0,0 +1,28 @@
...
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
+
+@Component(service = [RemoteAliasLookUp::class, UsedByFlow::class], scope = PROTOTYPE)
+class RemoteAliasLookUpImpl @Activate constructor() :
+    RemoteAliasLookUp, UsedByFlow, SingletonSerializeAsToken {
+
+    override fun lookup(x500Name : String, cpiName: String): AliasMemberInfo? {
+        return getAliasMemberData().firstOrNull {
```
szymonsztuka, line 17: Would a map be better data structure (not from performance point of view, as we want have many elements), but from a conceptual point of view (we search for a result which we know is unique ( this is a unique identifier`"$x500Name@$cpiName"` across the whole view of my aliases)?

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
SaurabhR3: This is hardcoded data, why do we need to compute it from other field? and from which field?

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
SaurabhR3: May be, for that we need to process the cache entry to create the map. or we can cache the data in the form of map. But that will be as part of other ticket.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
szymonsztuka: Ok.
1 . I don't think this is true, we don't know yet, but ok.
2. `AliasIdentityDataService` is not used by `RemoteAliasLookUpImpl`.
3.  `HardcodedAliasIdentityDataServiceImpl` type has already been hardcoded/leaked into `RemoteAliasLookUpImpl.getAliasMemberData` method, so for me a simpler solution would be just a function here.
`RemoteAliasLookUpImpl` doesn't use the service as an interface one, we can't inject implementation but without a need to modify the code `RemoteAliasLookUpImpl`.

yarivus (approved):
LGTM

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
SaurabhR3: 1. We know that creating or fetching cache data is not responsibility of RemoteAliasLookUp, data will be provided by another service from cache.
2. 3. I agree `AliasIdentityDataService` is not used in `RemoteAliasLookUpImpl`

I am still not convinced that we should move function and data classes in `RemoteAliasLookUpImpl` and remove the `HardcodedAliasIdentityDataServiceImpl`
We can just get rid of `AliasIdentityDataService` as we don't know how the API will look like for this.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
jmacmahonr3: Likely best to ask how MemberLookup works and model after them. Consistency is good

yarivus (approved):
LGTM

yarivus (approved):
LGTM

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
szymonsztuka: As spoken offline, yes it would be beneficial to add new new API call to this flow even if this doesn't have impact on the flow, however it shows to the reader that there will be API needed in each such flows - so in summary I'm withdrawing my objections/previous comments in this thread.

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
szymonsztuka: Fair, agree

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
SaurabhR3: Fixed in 8464599

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpServiceImpl.kt
SaurabhR3: Added FetchFacadeFromAliasIdentityFlow which can be used for testing now and when we will have the actual implementation of  RemoteAliasLookUp

testing/cpbs/test-invocation-cordapp/src/main/kotlin/com/r3/corda/testing/interop/FacadeInvocationFlow.kt
SaurabhR3: Fixed in 8464599

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
SaurabhR3: Added FetchFacadeFromAliasIdentityFlow

szymonsztuka (approved):

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/HardcodedAliasIdentityServiceImpl.kt
SaurabhR3: Removed

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
SaurabhR3: Fixed in 8464599

components/interop/interop-service/src/main/kotlin/net/corda/interop/service/impl/RemoteAliasLookUpImpl.kt
SaurabhR3: ok

yarivus (approved):
LGTM

szymonsztuka (approved):

yarivus (commented):
thx @SaurabhR3

PR #3870, opened 2023-05-13
CORE-11945: Remove sub-batching from the state and event pattern

JamesHR3:
The state and event pattern divides up a batch of records into sub-batches by splitting the batch from Kafka when it sees a second instance of a key in the batch. Each sub-batch is processed and committed separately. This has a large negative performance impact if keys in the batch repeat frequently and close together, as it means very small numbers of events are processed in one go before the commit time to Kafka is incurred.

For most usages of the state and event pattern this doesn't matter. However, the P2P markers topic typically has two or three events on the same key close to each other. Under high loads and with session messages occurring, the consumer group lag on this topic rapidly grows. The impact of this is that the P2P layer could spuriously replay messages as it takes a long time for it to process the corresponding marker.

Here the sub-batching logic is removed entirely, and the pattern updated to use the previous output state as the next input if multiple records on the same key appear in the same batch. This eliminates the consumer lag growing as described above.

One side impact of this is that the state and event listener is no longer guaranteed to see every state update on a key. This is largely a non-issue as most usages of this are waiting for a `null` state to remove a timer, which they are guaranteed to see if it has happened and are guaranteed to see non-`null` if the state is not nulled out.


rick-r3:
Is there any danger of breaking the flow model with this?  e.g. flow has 2 pending events, picks them both up and processes them both with no checkpoint in-between (that is what this might imply, yes?).

It sort of sounds like maybe it _should_ be a problem, but I actually think it isn't since the end result is the same.

rick-r3 (approved):

PR #3871, opened 2023-05-14
CORE-13165: Add a cache for static and evolvable class tags

JamesHR3:
Add a class tag cache for both static and evolvable class tags.

Kryo serialization uses static class tags to indicate which bundle a class came from, while AMQP uses evolvable ones for the same purpose. Calculating these involves iterating across all bundles in the sandbox group until the correct one is located. As these serialization operations occur frequently, this gets expensive. These tags should not change for the lifetime of the sandbox, however, so holding a map of them in memory should be safe while providing a reasonable performance boost.

In performance testing fiber serialization times have been observed to decrease from 4ms to 0.5ms with this change. AMQP serialization remains unmeasured but is likely to see a similar improvement.


libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -90,9 +91,16 @@ internal class SandboxGroupImpl(
...
+    override fun getEvolvableTag(klass: Class<*>) = evolvableTagCache.computeIfAbsent(klass) {
+        getClassTag(klass, isStaticTag = false)
+    }
 
-    override fun getEvolvableTag(klass: Class<*>) = getClassTag(klass, isStaticTag = false)
+    private val staticTagCache = ConcurrentHashMap<Class<*>, String>()
+    private val evolvableTagCache = ConcurrentHashMap<Class<*>, String>()
 
     @Suppress("ComplexMethod", "NestedBlockDepth")
     override fun getClass(className: String, serialisedClassTag: String): Class<*> {
```
rick-r3, line 118: The above looks fine to me.  With one eye on prematurely optimising, I would have thought that it is fairly evident that this method (going in the reverse direction) could also benefit from caching.  Yes, we have Fiber caching now so this isn't getting hammered by Kryo but presumably it is used extremely frequently by AMQP deserialisation ? 

`ConcurrentHashMap<Pair<String, String>, Class<*>>` and a `computeIfAbsent()` around the whole method ?

Maybe longer term we can reduce the number of calls to this method, but that is harder and wrapping in `computeIfAbsent()` is easy.

rick-r3 (commented):
Looks fine, but I think while we are at it we should cache the reverse.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
JamesHR3: I actually tried this and it seemed non-trivial to get it right - I started getting a lot of AMQP deserialization errors. However that may have been a different problem. Worth another go just to be sure.

That said, while this definitely makes a difference, it also may not be the most important thing.

rick-r3: It causing errors seems incredibly weird.

sprintengine: I found the same issue, I'd be keen to get this merged and open a separate PR to investigate improving `getClass`.

I created https://r3-cev.atlassian.net/browse/CORE-14188 to investigate the reverse

jujoramos (approved):

conal-smith (approved):

rick-r3 (approved):

PR #3875, opened 2023-05-15
NOTICK: Stablize Corda Performance

Omar-awad:


components/flow/flow-mapper-service/build.gradle
```diff
@@ -5,13 +5,15 @@ plugins {
 }
 
 dependencies {
+    implementation project(path: ':libs:metrics')
```
chrisr3, line 8: You don't need this dependency here because you are adding it below.

components/flow/flow-service/src/main/kotlin/net/corda/flow/metrics/impl/FlowMetricsImpl.kt
```diff
@@ -0,0 +1,112 @@
...
+) : FlowMetrics {
+
+    private val currentState: FlowMetricState
+    private val eventReceivedTimestampMillis: Long
+    private var fiberStartTime = clock.nowInMillis()
+    private var fiberExecutionTime: Long = 0
+
+    private companion object {
+        val objectMapper = ObjectMapper()
+    }
```
chrisr3, line 24: This is a strange place to put an `ObjectMapper`.... :thinking:

PR #3874, opened 2023-05-15
CORE-12581: run signing service direct from flow operations processor…

conal-smith:
This is a back-merge of https://github.com/corda/corda-runtime-os/pull/3801 from 5.1.

Please see Dickon's description there.

PR #3876, opened 2023-05-15
CORE-13560 - initial crypto metrics per operation type

conal-smith:


dickon (dismissed):
LGTM

rick-r3 (commented):
Makes sense.

kyriathar:
Is this conflicting with [#3874](https://github.com/corda/corda-runtime-os/pull/3874)? Should that be merged first?

ronanbrowne (approved):
LGTM - build

components/crypto/crypto-persistence/src/main/kotlin/net/corda/crypto/persistence/GetEntityManagerFactory.kt
```diff
@@ -15,32 +16,37 @@ fun getEntityManagerFactory(
...
+                        ?: throw IllegalStateException(
+                            "persistenceUnitName ${CordaDb.Crypto.persistenceUnitName} is not registered."
+                        )
                 )
-        )
-    }
-    return entityManagerFactory
+            }
+            entityManagerFactory
+        }!!
```
dickon, line 51: We should not use !! in production code

components/crypto/crypto-persistence/src/main/kotlin/net/corda/crypto/persistence/GetEntityManagerFactory.kt
```diff
@@ -15,32 +16,37 @@ fun getEntityManagerFactory(
...
-                    tenantId
-                )
-            )?.cryptoDmlConnectionId
-                ?: throw IllegalStateException(
-                    "virtual node for $tenantId is not registered."
-                ),
-            entitiesSet = jpaEntitiesRegistry.get(CordaDb.Crypto.persistenceUnitName)
-                ?: throw IllegalStateException(
-                    "persistenceUnitName ${CordaDb.Crypto.persistenceUnitName} is not registered."
+    return CordaMetrics.Metric.EntityManagerFactoryCreationTimer.builder()
```
dickon, line 19: better to use an expression body here

components/crypto/crypto-persistence/src/main/kotlin/net/corda/crypto/persistence/GetEntityManagerFactory.kt
dickon: Also, there's no test coverage on this metric, or the others.

components/crypto/crypto-persistence/src/main/kotlin/net/corda/crypto/persistence/GetEntityManagerFactory.kt
dickon: Please instrument both sides of the if(onCluster) separately so we can see what's going on. In particular, the true branch shouldn't be taking much time more than once.

dickon: Minor point, and probably a bit late now, but I'd rather see metrics like this imlemented as function annotations, so we aren't adding unnecessarily deep block nesting all over the place. For one, the change of block structure makes patches like this harder to review than it needs to be.

dickon (approved):
Okay as far as it goes; there's certainly more we can do, so approved for now, and these are ways to take this further:

We are missing a metric on `SoftCryptoServiceProviderImpl.getInstance`. It's probably nothing but I want to be sure the caching with the `lazy` works.

We should measure the SigningSerivceImpl.getOwnedKeyRecord, separately for each branch fo the `if`.

It's probably trivial, but worth measuring `SoftCryptoServicePRoviderImpl.decodePublicKey` and `CryptoSignatureSpec.toSignatureSpec` 

`SigningRepositoryFacotryImpl.getInstance`  and `CryptoServiceFacotryImpl.getInstanc` makes an EMF so is worth measuring.

`SoftCryptoServiceProviderImpl.getInstance` is expensive so let's measure that.


A few general comments:

- We generally avoid !! in production code
- There's no tests, i.e. nothing in the CI that proves these metrics do what we think they do.

components/crypto/crypto-persistence/src/main/kotlin/net/corda/crypto/persistence/GetEntityManagerFactory.kt
sprintengine: Thanks, I'll make some further updates in a later PR to address some of these comments and add more metrics. Just this one is next in line to get merged asap

PR #3883, opened 2023-05-16
CORE-13353: Add metrics for flow processing

JamesHR3:
Adds a large number of metrics around the flow engine and flow mapper.

- Flow fiber serialization time
- Flow start lag (time between submitting an event through the REST API and the flow engine processing it)
- Flow execution time (time between processing the start event and the flow completing)
- Flow event lag time (time between publishing a flow event and processing it)
- Flow event pipeline execution time (time for the flow pipeline to process a single event)
- Flow event fiber execution time (time to process a single step of the user code)
- Flow pipeline execution time (time a single flow spends being executed, vs total time to run which includes pauses while the system is doing other things)
- Flow fiber execution time (time a single flow spent in user code)
- Flow suspension time (total time a flow spent suspended)
- Flow event suspension wait time (time between a flow suspending and waking up again)
- Flow scheduled wakeup count (number of times a flow has been scheduled for a wakeup due to not seeing events for some time)
- Flow mapper event processing time (time taken for the flow mapper to process an event)
- Flow mapper deduplication count (number of start events/session init events dropped owing to the mapper state already existing)
- Flow mapper creation count (number of new flows started)
- Flow mapper cleanup count (number of mapper states being cleaned up)
- Flow mapper event lag (time between a mapper event being published and processed)
- Flow mapper expired session event count (number of session events dropped owing to them expiring)


components/flow/flow-mapper-service/build.gradle
```diff
@@ -5,13 +5,15 @@ plugins {
 }
 
 dependencies {
+    implementation project(path: ':libs:metrics')
```
jujoramos, line 8: Nitpick: isn't this redundant with `implementation project(":libs:metrics")`, defined in line 16?.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/factory/impl/FlowEventPipelineFactoryImpl.kt
```diff
@@ -86,16 +96,27 @@ class FlowEventPipelineFactoryImpl(
...
+        mdcProperties: Map<String, String>,
+        eventRecordTimestamp: Long
+    ): FlowEventPipeline {
+        val flowCheckpoint = flowCheckpointFactory.create(event.flowId, checkpoint, config)
+
+        val metrics = try {
+            flowMetricsFactory.create(eventRecordTimestamp, flowCheckpoint)
+        } catch (e: Exception) {
+            throw e
+        }
```
jujoramos: Nitpick: no need to catch it if we're gonna throw it unmodified.

components/flow/flow-service/src/main/kotlin/net/corda/flow/metrics/FlowMetricsRecord.kt
```diff
@@ -0,0 +1,13 @@
+package net.corda.flow.metrics
+
+interface FlowMetricsRecord {
```
jujoramos: Nitpick: I'd rename it to `FlowMetricsRecorder`.

components/flow/flow-service/src/main/kotlin/net/corda/flow/metrics/FlowMetricsRecord.kt
```diff
@@ -0,0 +1,13 @@
...
+    fun recordFlowEventLag(lagMilli: Long, flowEventType: String)
+    fun recordFlowStartLag(lagMilli: Long)
+    fun recordFlowSuspensionCompletion(operationName: String, executionTimeMilli: Long)
+    fun recordFiberExecution(executionTimeMillis: Long)
+    fun recordPipelineExecution(executionTimeMillis: Long, flowEventType: String)
+    fun recordTotalPipelineExecutionTime(executionTimeMillis: Long)
+    fun recordTotalFiberExecutionTime(executionTimeMillis: Long)
+    fun recordTotalSuspensionTime(executionTimeMillis: Long)
+    fun recordFlowCompletion(executionTimeMillis: Long, completionStatus:String)
```
jujoramos: We are constantly calculating the difference between two instances of `Instant`, represented as `epochMillis`, to finally record the lag/duration as an instance of `Duration` in the metrics (after transforming again `epochMillis` into `Duration`). Wouldn't it be better for all methods within this interface to receive an instance of `Duration` instead (caller methods would just execute `Duration.between(startTime, endTime))`?

jujoramos (commented):
Even though these new metrics will be really useful to us, won't they also negatively impact the performance as they involve multiple calculations in every step of the process?.
Another thing, probably not as urgent as getting this merged: it would be good to have testing and KDocs added.

jujoramos (approved):

PR #3889, opened 2023-05-16
CORE-10051 Remove assign hardware HSM path in crypto worker

kyriathar:
- removes assign hardware HSM path in crypto worker which is not used


dickon (approved):
LGTM.

PR #3890, opened 2023-05-16
CORE-13560

conal-smith:
This is a back-merge of https://github.com/corda/corda-runtime-os/pull/3801 from 5.1.

Please see Dickon's description there.


components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
```diff
@@ -69,24 +76,34 @@ class CryptoFlowOpsBusProcessor(
...
                     )
                 } else {
                     val response = executor.executeWithRetry {
-                        handleRequest(request.request, request.context)
+                        CordaMetrics.Metric.CryptoFlowOpsProcessorExecutionTime.builder()
+                            .withTag(CordaMetrics.Tag.OperationName, requestPayload::class.java.simpleName)
+                            .build()
+                            .recordCallable {
+                                handleRequest(requestPayload, request.context)
+                            }
```
jujoramos: Shouldn't this wrap the entire `try{} catch {}` block?. Aside from that, do we still need the `executeWithRetry` here? (maybe a question for @dickon ).
Having the metric recorded within `executeWithRetry` might end up recording multiple times for the same (initially failed) operation.

jujoramos (dismissed):

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
sprintengine: could be, I didn't test it after resolving this merge conflict, probably should be outside wrapping the entire try catch

jujoramos (approved):
LGTM

PR #3892, opened 2023-05-16
CORE-7287 duplicate topic keys bug

ciaran238:
## Context
When starting the combined worker it is quite common to get errors like
`2022-10-26 16:05:53.234 UTC [156] ERROR:  duplicate key value violates unique constraint "topic_pk"`
`2022-10-26 16:05:53.234 UTC [156] DETAIL:  Key (topic)=(gateway.tls.truststores) already exists.`
This is an SQL exception for duplicate key values caused by a race condition between when topics are added to the DB and a find to see if that key already exists on the DB. This race condition exists because during startup there are several workers that are executing the find for a topic at the same time, not finding the topic i.e. returning null and then trying to create that topic at the same time.
## Solution
During initialization of the DB, all topics will be pulled from the schema and populated into the DB. This means that all topics will be present before the workers exist, and so they will have no need to create any missing topics. As a result, the functionality which allowed workers to auto-create topics is now being removed, as it's no longer needed

[CORE-7287]: https://r3-cev.atlassian.net/browse/CORE-7287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/PostgresDbSetup.kt
```diff
@@ -58,11 +79,51 @@ class PostgresDbSetup(
             populateConfigDb()
             createUserConfig("admin", "admin")
             createDbUsersAndGrants()
+            createTopics()
         } else {
             log.info("Table config.config exists in $dbSuperUserUrl, skipping DB initialisation.")
         }
     }
```
simon-johnson-r3: I assume this is supposed to replace auto creation? Has that then been turned off?

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/PostgresDbSetup.kt
```diff
@@ -58,11 +79,51 @@ class PostgresDbSetup(
             populateConfigDb()
             createUserConfig("admin", "admin")
             createDbUsersAndGrants()
+            createTopics()
```
simon-johnson-r3: Maybe make this more explicit: `createTopicsOnDbMessageBus` ?

simon-johnson-r3: This will cause problems if we configure the combined worker to use Kafka, which is a common use case. If will need to be called conditionally based on something like (needs checking):
```
params.defaultParams.messaging[BUS_TYPE] == BusType.DB.name
```
You can see an example of this parameter being used here: https://github.com/corda/corda-runtime-os/blob/dcaf4a3901e04a7066a17b80e4b27289147375ff/applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/WorkerHelpers.kt#L100

libs/messaging/db-message-bus-datamodel/src/main/kotlin/net/corda/messagebus/db/datamodel/TopicEntry.kt
```diff
@@ -23,3 +23,21 @@ class TopicEntry(
         return "TopicEntry(topic='$topic', numPartitions=$numPartitions)"
     }
 }
+/**
+ * This entity represents the topic metadata.  This entity will be
+ * used to insert topics when initialising the db.
+ */
+@Entity(name = "topic")
+@Table(name = "messagebus.topic")
```
simon-johnson-r3: The rule is that entity definitions are always schema agnostic. Almost all schemas are configurable, however it is true that the message bus schema is (currently) hard coded as a parameter for the combined worker (rather than fed in on launch) here: https://github.com/corda/corda-runtime-os/blob/ec0ff3502c9ad26b27cd57d776dee33c44e72bce/applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt#L106

Here's an example of that being read: https://github.com/corda/corda-runtime-os/blob/411b9a6dc8bec4b5f7532615c855850cfa0877dd/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/configuration/DbBusConfigMergerImpl.kt#L23

The code should read this parameter and accommodate the schema at runtime in `createTopics` rather than have it hard coded in the annotation.

libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/persistence/DBAccess.kt
```diff
@@ -31,14 +31,15 @@ import javax.persistence.Tuple
...
-    private val defaultNumPartitions = 3
     private val autoCreate = true
 
     companion object {
         private val log: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
         internal val ATOMIC_TRANSACTION = TransactionRecordEntry("Atomic Transaction", TransactionState.COMMITTED)
+
+        // At the moment it's not easy to create partitions, so default value increased to 3 until tooling is available
+        // (There are multiple consumers using the same group for some topics and some stay idle if there is only 1 partition)
+        const val defaultNumPartitions = 3
```
simon-johnson-r3: This is only in here for auto topic creation. If we don't need that any more, we might consider moving it. It's the kind of thing which might be configurable externally. For now it could probably go alongside the postgres setup stuff.

simon-johnson-r3:
One thing I'm not entirely clear on. I assume this is a solution to that bug only because it avoids calling the function relating to auto topic creation in the db? Do we know why this is failing in the first place? Changing the combined worker setup so it writes topics on first startup is not such a bad idea, I just wonder if we know why auto topic creation isn't already working?

`a race condition between when topics are added to the DB and a find to see if that key already exists on the DB.` is this just something that happens on startup for some reason?

simon-johnson-r3 (changes requested):

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/PostgresDbSetup.kt
thiagoviana: If that's the case, I believe this is also something we would need to change:

**?: if (autoCreate) {**


```/**
     * If auto topic creation is enabled then will create the topic
     */
    fun getTopicPartitionMapFor(topic: String): Set<CordaTopicPartition> {
        val topicEntry = executeWithErrorHandling(
            "get topic partition map",
            allowDuplicate = true,
        ) { entityManager ->
            entityManager.find(TopicEntry::class.java, topic)
                ?: if (autoCreate) {
                    val topicEntry = TopicEntry(topic, defaultNumPartitions)
                    entityManager.persist(topicEntry)
                    topicEntry
                } else {
                    throw CordaMessageAPIFatalException("Cannot find topic $topic")
                }
        }
        val topicPartitions = mutableSetOf<CordaTopicPartition>()
        repeat(topicEntry.numPartitions) { partition ->
            topicPartitions.add(CordaTopicPartition(topic, partition))
        }
        return topicPartitions
    }```

File: applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt [hunk 5]
```diff
@@ -137,14 +139,17 @@ class CombinedWorker @Activate constructor(
         // In the future, perhaps we can simply rely on the schema for crypto defaults, and not supply a
         // default passphrase and salt but instead require them to be specified.
 
+        val createTopics = params.defaultParams.messaging[BUS_TYPE] == BusType.DB.name
```
pmck91: perhaps a comment here, describing what this flag does would aid readability

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/PostgresDbSetup.kt
pmck91: Looking at the code, it looks like that comment provided context for the function call, which has changed. So the comment should be removed. The code also default autocorrect to true

https://github.com/corda/corda-runtime-os/blob/37615e6b9612a0649065954cda2b40449a177ead/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/persistence/DBAccess.kt#L132

the check could probably just be removed entirely. In this PR or a new ticket

wzur-r3 (dismissed):
LGTM for Gradle changes

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
thiagoviana: It seems that this:

```params.defaultParams.messaging[BUS_TYPE] == BusType.DB.name```

It's comparing:

``` "DATABASE" == "DB" ```

File: libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/PostgresDbSetup.kt (hunk 1 above)
Comment (ciaran238) [reply libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/PostgresDbSetup.kt]: I have removed the auto creation as it wont be used anymore

File: libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/PostgresDbSetup.kt (hunk 2 above)
Comment (ciaran238) [reply libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/PostgresDbSetup.kt]: changed to suggested name

Comment (ciaran238) [reply libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/PostgresDbSetup.kt]: A condition to check the bus type has been created

File: libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/persistence/DBAccess.kt (hunk 4 above)
Comment (ciaran238) [reply libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/persistence/DBAccess.kt]: I moved this as suggested

File: applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt (hunk 5 above)
Comment (ciaran238) [reply applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt]: I have added this comment

File: libs/messaging/db-message-bus-datamodel/src/main/kotlin/net/corda/messagebus/db/datamodel/TopicEntry.kt (hunk 3 above)
Comment (ciaran238) [reply libs/messaging/db-message-bus-datamodel/src/main/kotlin/net/corda/messagebus/db/datamodel/TopicEntry.kt]: I have removed these changes I made to TopicEntry, and instead added a new connection in `PostgresDbSetup.kt` for the messagebus and made changes to `SqlFormatters.kt` so that columns can be annotated with `@Id`

File: testing/db-message-bus-testkit/src/main/kotlin/net/corda/db/messagebus/testkit/DBSetup.kt [hunk 6]
```diff
@@ -48,6 +50,10 @@ object DBSetup: BeforeAllCallback {
                 dataSource.connection.use { connection -> lbm.createUpdateSql(connection, cl, it) }
             }
             dataSource.connection.use { connection  ->  lbm.updateDb(connection, cl) }
+            if(!topicsCreated) {
```
ben-millar: Do we need to check this here if it's already done in `PostgresDbSetup`? Or vice versa; can we just do it in one of the two spots?

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
```diff
@@ -137,14 +139,20 @@ class CombinedWorker @Activate constructor(
...
+
         PostgresDbSetup(
             dbUrl,
             superUser,
             superUserPassword,
             dbAdmin,
             dbAdminPassword,
             dbName,
-            config.factory
+            dbBusType,
```
ben-millar: Discussed offline, just check if we need to pass this bool in.

ben-millar (commented):
Mostly LGTM, just a couple of small comments to double check

testing/db-message-bus-testkit/src/main/kotlin/net/corda/db/messagebus/testkit/DBSetup.kt
ciaran238: This is for setting up topics for the in-memory db message bus (used in integration tests) which doesnt use postgres

applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/WorkerHelpers.kt
```diff
@@ -26,7 +26,7 @@ data class PathAndConfig(val path: String, val config: Map<String, String>)
 
 enum class BusType {
     KAFKA,
-    DB
+    DATABASE
```
simon-johnson-r3, line 29: Because the name value of this enum value is used in the code, where it is compared to bootstrap parameters passed to the worker, it forms part of the worker public API. You shouldn't then change it because we do not permit breaking API changes between versions of Corda.

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/PostgresDbSetup.kt
```diff
@@ -58,6 +59,9 @@ class PostgresDbSetup(
             populateConfigDb()
             createUserConfig("admin", "admin")
             createDbUsersAndGrants()
+            if (isDbBusType) {
+                DbMessageBusSetup().createTopicsOnDbMessageBus(messageBusConnection())
```
simon-johnson-r3: Code like this, where you instantiate something, call a single method on it, and never use the instance again can point to an anti-pattern. Specifically that is more likely to be true where you must call that single method and there is no other method to call, which is the case here. In other words, if you must always do this and only this, there is no real point in instantiating anything at all.

Here I would suggest moving `createTopicsOnDbMessageBus` into the companion object of `DbMessageBusSetup` and leaving the call point as
```
DbMessageBusSetup.createTopicsOnDbMessageBus(messageBusConnection())
```
Then it remains scoped, but no temporary object instance is required.

testing/db-message-bus-testkit/src/main/kotlin/net/corda/db/messagebus/testkit/DBSetup.kt
```diff
@@ -11,6 +12,7 @@ import org.osgi.framework.FrameworkUtil
 import java.io.StringWriter
 
 object DBSetup: BeforeAllCallback {
+    var topicsCreated = false
```
simon-johnson-r3: Should this be private, and do we need it at all?

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/DbMessageBusSetup.kt
```diff
@@ -0,0 +1,76 @@
...
+import java.sql.Connection
+
+class DbMessageBusSetup {
+    companion object{
+        // At the moment it's not easy to create partitions, so default value increased to 3 until tooling is available
+        // (There are multiple consumers using the same group for some topics and some stay idle if there is only 1 partition)
+        const val defaultNumPartitions = 3
+        private val log = LoggerFactory.getLogger(DbMessageBusSetup::class.java)
+    }
+    /** creates the topics for the INMEMORY or DATABASE db message bus
```
simon-johnson-r3: Generally speaking you would stick to the code style of the surrounding code. Here that would dictate to put `/**` on it's own line.

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/DbMessageBusSetup.kt
```diff
@@ -0,0 +1,76 @@
...
+            val data: String = it.openStream()
+                .bufferedReader(Charset.defaultCharset()).use { it.readText() }
+            val parsedData: TopicDefinitions = mapper.readValue(data)
+            parsedData
+        }
+        val topicConfigs = topicDefinitions.flatMap { it: TopicDefinitions ->
+            it.topics.values
+        }
+        log.debug { "Mapping topicDefinitions to topicConfigs, topicDefinitions = $topicDefinitions \ntopicConfigs = $topicDefinitions" }
+        messageBusConnection.use { connection ->
```
simon-johnson-r3: The use of `use` here means this function closes the connection. That is not documented in the api, but you should also consider, given you pass the connection in to this function and its created externally, whether that is really what you want.

I would suggest the caller, who is in control of creating the connection would be better placed to say whether they wanted it closed or not.

testing/db-message-bus-testkit/src/main/kotlin/net/corda/db/messagebus/testkit/DBSetup.kt
```diff
@@ -48,6 +50,10 @@ object DBSetup: BeforeAllCallback {
                 dataSource.connection.use { connection -> lbm.createUpdateSql(connection, cl, it) }
             }
             dataSource.connection.use { connection  ->  lbm.updateDb(connection, cl) }
+            if(!topicsCreated) {
+                DbMessageBusSetup().createTopicsOnDbMessageBus(dataSource.connection)
```
simon-johnson-r3: `dataSource.connection` was closed already on line 52. Actually I see it was also closed on line 50 too, an existing problem! I guess this works because `close` isn't doing much, or perhaps release back to a pool such that the connection is still valid - although this will mess up pool management.

I wouldn't make the problem worse. Could you try and refactor so `close` is only called once? (note that this overlaps with the comment I made on `DbMessageBusSetup` above about closing connections.

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/DbMessageBusSetup.kt
```diff
@@ -0,0 +1,76 @@
...
+            topicConfigs.map {
+                connection.createStatement().execute(
+                    TopicEntry(it.name, defaultNumPartitions).toInsertStatement()
+                )
+            }
+            if (!connection.autoCommit) {
```
simon-johnson-r3: This strikes me as a bit non-deterministic. Here you are saying the function takes control of commits after all inserts are executed, but you also allow auto commits, where you can leave the db in a bad state if half of them are committed only.

I would suggest making the contract of this method (i.e. documenting in the header) that commits are controlled in all cases, and then guarding against that at runtime by putting a
```
require(!messageBusConnection.autoCommit()) { ... }
```
at the top of the function.

simon-johnson-r3 (changes requested):

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/DbMessageBusSetup.kt
ciaran238: 👍

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/DbMessageBusSetup.kt
ciaran238: 👍

testing/db-message-bus-testkit/src/main/kotlin/net/corda/db/messagebus/testkit/DBSetup.kt
ciaran238: 👍

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/DbMessageBusSetup.kt
ciaran238: 👍

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/PostgresDbSetup.kt
ciaran238: 👍

testing/db-message-bus-testkit/src/main/kotlin/net/corda/db/messagebus/testkit/DBSetup.kt
ciaran238: Yes, it should be private. It is needed because the `beforeAll`  may be executed more than once but the topics only need to be created once.

simon-johnson-r3 (dismissed):

applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/WorkerHelpers.kt
pmck91: Is a better solution to change params.defaultParams.messaging[BUS_TYPE] to be db?

thiagoviana (dismissed):

applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/WorkerHelpers.kt
ciaran238: I dont think so, I think it should have been `DATABASE` before. 
params.defaultParams.messaging[BUS_TYPE]  is the key for what is passed in as the message bus type, it is documented that it should be KAFKA or DATABASE ([Combined Worker README](https://github.com/corda/corda-runtime-os/tree/release/os/5.1/applications/workers/release/combined-worker#from-the-command-line))

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/DbMessageBusSetup.kt
```diff
@@ -0,0 +1,77 @@
...
+         * creates the topics for the INMEMORY or DATABASE db message bus.
+         * Commits are controlled by this method, so it is required that connections have autoCommit disabled
+         * @param messageBusConnection
+         * */
+        fun createTopicsOnDbMessageBus(messageBusConnection: Connection) {
+            require(!messageBusConnection.autoCommit) {
+                "DB Message Bus connection used for creating topics must not have autoCommit enabled"
+            }
+            val bundle = FrameworkUtil.getBundle(net.corda.schema.Schemas::class.java)
+            log.debug { "Got bundle $bundle for class (net.corda.schema.Schemas::class.java)" }
```
mbrkic-r3: I guess we don't need "::class.java" in log message

ronanbrowne (dismissed):
LGTM (build)

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/DbMessageBusSetup.kt
ciaran238: 👍

thiagoviana (approved):

ronanbrowne (approved):

PR #3900, opened 2023-05-17
CORE-13750: return 404 rather than null for missing HSM association

dickon:
This aligns better with other REST endpoints, and may be less surprising.

Before we would return 404 for a missing tenant or category, and a JSON body containing null if the tenant and category were correct. Now we return 404 in all cases, with different message bodies identifying the problem.

### DIFF
--- a/components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/HsmRestResourceImpl.kt
+++ b/components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/HsmRestResourceImpl.kt
@@ -58,7 +58,9 @@ class HsmRestResourceImpl @Activate constructor(
             "Find HSM",
             untranslatedExceptions = setOf(ResourceNotFoundException::class.java)
         ) {
-            hsmRegistrationClient.findHSM(tenantId, category.toCategory())?.expose()
+            hsmRegistrationClient.findHSM(tenantId, category.toCategory())?.expose() ?: throw ResourceNotFoundException(
+                "No association found for tenant ${tenantId} category ${category}"
+            )
         }
     }
 
--- a/components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/HsmRestResourceImpl.kt
+++ b/components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/HsmRestResourceImpl.kt
@@ -51,7 +51,7 @@ class HsmRestResourceImpl @Activate constructor(
         }
     }
 
-    override fun assignedHsm(tenantId: String, category: String): HsmAssociationInfo? {
+    override fun assignedHsm(tenantId: String, category: String): HsmAssociationInfo {
         verifyTenantId(tenantId)
         return tryWithExceptionHandling(
             logger,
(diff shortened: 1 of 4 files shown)


components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/HsmRestResourceImpl.kt
```diff
@@ -58,7 +58,9 @@ class HsmRestResourceImpl @Activate constructor(
             "Find HSM",
             untranslatedExceptions = setOf(ResourceNotFoundException::class.java)
         ) {
-            hsmRegistrationClient.findHSM(tenantId, category.toCategory())?.expose()
+            hsmRegistrationClient.findHSM(tenantId, category.toCategory())?.expose() ?: throw ResourceNotFoundException(
```
yift-r3, line 61: Should we change the return type from `HsmAssociationInfo?` to `HsmAssociationInfo`?

components/membership/membership-rest-impl/src/test/kotlin/net/corda/membership/impl/rest/v1/HsmRestResourceImplTest.kt
```diff
@@ -46,24 +46,37 @@ class HsmRestResourceImplTest {
     @Nested
     inner class ApiTests {
         @Test
-        fun `assignedHsm returns null if no HSM found`() {
+        fun `assignedHsm returns 404 if no HSM found`() {
             whenever(hsmRegistrationClient.findHSM(tenantId, TLS)).doReturn(null)
 
-            val hsm = ops.assignedHsm(tenantId, "tls")
+            val e = assertThrows<ResourceNotFoundException> { ops.assignedHsm(tenantId, "tls") }
+            assertThat(e.message).contains("No association found for tenant ${tenantId} category tls")
```
yift-r3: `assertThat(e).hasMessageContaining("No association found for tenant ${tenantId} category tls")`

dickon: done.

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/HsmRestResourceImpl.kt
dickon: good idea, done.

yift-r3: Can you change `HsmRestResource` as well? (also, run the `:processors:rest-processor:integrationTest` I think that the `swaggerBaseline.json` will need to change)

yift-r3 (dismissed):
LGTM

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/HsmRestResource.kt
```diff
@@ -35,7 +35,7 @@ interface HsmRestResource : RestResource {
      * @param category The category of the HSM; can be the value 'ACCOUNTS', 'CI', 'LEDGER', 'NOTARY', 'SESSION_INIT',
      * 'TLS', or 'JWT_KEY'.
      *
-     * @return Information on the assigned HSM, or null if no HSM is assigned.
+     * @return Information on the assigned HSM, or 404 if no HSM is assigned.
```
vkolomeyko:
```suggestion
     * @return Information on the assigned HSM, or throws `ResourceNotFoundException` if no HSM is assigned.
```

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/HsmRestResourceImpl.kt
dickon: done

vkolomeyko (approved):
LGTM

**Note:** Strictly speaking this is a breaking REST API change. But given that we do not currently have a way to support multiple versions and given that it is just a minor change to the error scenario, I am happy to allow this change "as is".
Going forward, we intend to be more rigorous to these sort of change to enable proper versioning support.

yift-r3 (approved):

PR #3901, opened 2023-05-17
CORE-13890: RPC sender - pick partition at random to distribute the load evenly

vkolomeyko:


vkolomeyko:
@JamesHR3 - I wonder if there is a way to use binaries produced from this PR to do a round of performance testing to see if it has any positive impact?

If it does and we will be willing to merge this change to `5.0`, I will raise a Jira and do rest of the formalities.

LWogan:
This change makes sense to me, but as an aside do we have any evidence of multiple RPC subscriptions using the same topic? I saw the thread on slack, but i dont think that shows two active consumers of one topic

vkolomeyko:
> This change makes sense to me, but as an aside do we have any evidence of multiple RPC subscriptions using the same topic? I saw the thread on slack, but i dont think that shows two active consumers of one topic

If we are talking about Slack discussion: https://r3holdco.slack.com/archives/C03AYSL75HQ/p1684324279011299 - this change is somewhat orthogonal. I just spotted this opportunity to distribute loa between partitions in the hope to make a performance impact.

LWogan (dismissed):
LGTM

rick-r3 (approved):

PR #3902, opened 2023-05-17
CORE-13756: Ensure AMQP deserialises a naked Char object correctly.

chrisr3:
Using AMQP to serialise and deserialise naked Java primitives is _hideously_ inefficient. Anyone using this functionality will therefore be suitably rewarded by its _miserable_ performance :no_mouth:.

The "trick" here is to register an AMQP serializer _specifically_ for the case where the declared type is `java.lang.Character` but the actual type is `java.lang.Integer`, rather than using `AMQPPrimitiveSerializer`.


mnesbit (approved):
OK

fenryka:
Question: Given we've frozen 5.0 should this be in 5.1?

chrisr3:
> Question: Given we've frozen 5.0 should this be in 5.1?

@sachingokhale  and @marcellofarina75 have agreed to lift the code-freeze for **non-breaking** changes.

fenryka (approved):
LGTM

PR #3903, opened 2023-05-17
CORE-13806: Handle loading Java primitives from sandboxes.

chrisr3:
Allow serialization "class tags" to use Java primitive types. Also use the standard OSGi constant `system.bundle` for the symbolic name of the system bundle rather than `org.apache.felix.framework`.


mnesbit (approved):
OK

PR #3905, opened 2023-05-17
CORE-10051 Remove `HSMRegistrationBusService`

kyriathar:
- removes `HSMRegistrationBusService` and inlines `HSMRegistrationBusProcessor` creation in `CryptoProcessorImpl`
- removes crypto Kafka RPC clients dependencies from Crypto worker since Crypto Kafka relays are removed
- closes previous bus processor subscriptions on subscriptions re-creations
- removes duplicate following of dependent components of `CryptoProcessorImpl`
- adds lifecycle unit test for `CryptoProcessorImpl` asserting its lifecycle in terms of bus processor subscriptions
- aligns retrying config type with config schema


dickon:
Looks good. I did wonder whether we need to close the subscriptions. I don't think we ever have, even when we had standalone bus services, but there could be a leak there if the components goes down then up often?

processors/crypto-processor/build.gradle
```diff
@@ -63,15 +63,13 @@ dependencies {
...
-    runtimeOnly project(":components:crypto:crypto-client-impl")
-    runtimeOnly project(":components:crypto:crypto-client-hsm-impl")
     runtimeOnly project(":components:crypto:crypto-persistence-impl")
     runtimeOnly project(":components:db:db-connection-manager-impl")
     runtimeOnly project(':components:crypto:crypto-service-impl')
     runtimeOnly project(':components:crypto:crypto-softhsm-impl')
     runtimeOnly project(":libs:db:db-orm-impl")
     runtimeOnly project(':libs:flows:external-event-responses-impl')
-    runtimeOnly project(":libs:lifecycle:lifecycle-impl")
+//    runtimeOnly project(":libs:lifecycle:lifecycle-impl")
```
ronanbrowne: Should this be removed ?

ronanbrowne (changes requested):
Minor comment on build.gradle

processors/crypto-processor/build.gradle
kyriathar: Removed

ronanbrowne (dismissed):
LGTM - (build changes only)

libs/crypto/crypto-config-impl/src/main/kotlin/net/corda/crypto/config/impl/RetryingConfig.kt
```diff
@@ -2,7 +2,7 @@ package net.corda.crypto.config.impl
 
 import net.corda.libs.configuration.SmartConfig
 
-class CryptoBusProcessorConfig(internal val config: SmartConfig) : SmartConfig by config {
+class RetryingConfig(internal val config: SmartConfig) : SmartConfig by config {
```
dickon, line 5: I think long term we can get rid of these config classes and use SmartConfig directly, but there's a case for leaving them in place, and for now this change is a good step since you are focussed on other things.

processors/crypto-processor/build.gradle
```diff
@@ -20,8 +20,6 @@ dependencies {
     implementation "net.corda:corda-topic-schema"
 
     implementation project(':components:configuration:configuration-read-service')
-    implementation project(':components:crypto:crypto-client')
-    implementation project(":components:crypto:crypto-client-hsm")
```
dickon, line 24: Nice, I love seeing dependencies disappear

processors/crypto-processor/src/test/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImplTest.kt
```diff
@@ -0,0 +1,164 @@
...
+            addDependency<SoftCryptoServiceProvider>()
+            addDependency<CryptoServiceFactory>()
+            addDependency<HSMService>()
+            addDependency<HSMStore>()
+            addDependency<JpaEntitiesRegistry>()
+            addDependency<DbConnectionManager>()
+            addDependency<VirtualNodeInfoReadService>()
+            addDependency<SubscriptionFactory>()
+            val cryptoProcessor = CryptoProcessorImpl(
+                coordinatorFactory,
```
dickon, line 124: minor style point, not a blocker, but I'd use all keyword arguments on any argument list this long, since if someone needs to change the argument list of CryptoProcessorImpl later they have to do a lot of counting with positional arguments, especially if they don't add at the end.

Normally I switch over to keyword arguments when the functional call extends over multiple lines.

processors/crypto-processor/src/test/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImplTest.kt
```diff
@@ -0,0 +1,164 @@
...
+                        )
+                    ),
+                    any(),
+                    any()
+                )
+            )
+                .thenReturn(hsmRegSubscription)
+        }
+
+        LifecycleTest {
```
dickon, line 112: I didn't know we could test lifecycle without bringing up OSGi, which I think this is doing. Nice.

(If this is bringing up OSGi I'd probably want it in the integrationTest directory).

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -117,16 +111,17 @@ class CryptoProcessorImpl @Activate constructor(
     private val dependentComponents = DependentComponents.of(
         ::configurationReadService,
         ::hsmStore,
-        ::cryptoOpsClient,
```
dickon, line 120: I think it was intentional to include cryptoOpsClient here, so that CryptoProcessorImpl doesn't go up until everything it needs is ready. But I don't think that's essential, so maybe dropping what we don't need is okay. 

In this case, `CryptoOpsClient` is not OSGi and not Lifecycle, so it is just pointless, so deleting it is good.

(Changed my mind twice on this :))

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -117,16 +111,17 @@ class CryptoProcessorImpl @Activate constructor(
...
         ::hsmService,
-        ::hsmRegistration,
         ::dbConnectionManager,
         ::virtualNodeInfoReadService,
     )
 
-    private val lifecycleCoordinator =
+    // We are making the below coordinator visible to be able to test the processor as if it were a `Lifecycle`
+    // using `LifecycleTest` API
+    @get:VisibleForTesting
```
dickon, line 124: Seems a bit of a hack, but I don't really mind, can't see many problems coming from this.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -89,8 +86,6 @@ class CryptoProcessorImpl @Activate constructor(
     private val subscriptionFactory: SubscriptionFactory,
     @Reference(service = ExternalEventResponseFactory::class)
     private val externalEventResponseFactory: ExternalEventResponseFactory,
-    @Reference(service = JpaEntitiesRegistry::class)
```
dickon, line 92: So we imported the same thing twice with different names? Weird, seems a mistake.

Please keep the names aligned, so keep `jpaEntitiesRegistry` and delete `entitiesRegistry` since the class is called `JpaEntitiesRegistry`.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -272,11 +272,28 @@ class CryptoProcessorImpl @Activate constructor(
...
+            messagingConfig = messagingConfig
+        )
+
         // and start the subscriptions
-        logger.trace("Starting processing on ${flowGroupName} ${Schemas.Crypto.FLOW_OPS_MESSAGE_TOPIC}")
-        flowOpsSubscription.start()
-        logger.trace("Starting processing on ${rpcGroupName} ${Schemas.Crypto.RPC_OPS_MESSAGE_TOPIC}")
-        rpcOpsSubscription.start()
+        logger.trace("Starting processing on $flowGroupName ${Schemas.Crypto.FLOW_OPS_MESSAGE_TOPIC}")
+        flowOpsSubscription!!.start()
```
dickon: I know it's probably impossible for any of these !! to fail but I still believe we should not use !! in production codepaths, instead just log an ERROR and do nothing.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -245,22 +241,26 @@ class CryptoProcessorImpl @Activate constructor(
...
+        val flowOpsProcessor = CryptoFlowOpsBusProcessor(signingService, externalEventResponseFactory, retryingConfig)
+        val rpcOpsProcessor = CryptoOpsBusProcessor(signingService, retryingConfig)
+        val hsmRegistrationProcessor = HSMRegistrationBusProcessor(hsmService, retryingConfig)
 
-        // now make both the subscriptions
+        // now make the subscriptions
         val messagingConfig = event.config.getConfig(MESSAGING_CONFIG)
         val flowGroupName = "crypto.ops.flow"
-        val flowOpsSubscription = subscriptionFactory.createDurableSubscription(
+        flowOpsSubscription?.close()
```
dickon: Is this `close` enough to kill the thread and release everything?

processors/crypto-processor/src/test/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImplTest.kt
kyriathar: Doesn't brings up OSGi, just unit tests lifecycle stuff.

libs/crypto/crypto-config-impl/src/main/kotlin/net/corda/crypto/config/impl/RetryingConfig.kt
kyriathar: Yes agreed, for now I just aligned it to be closer to schema to be more intuitive.

processors/crypto-processor/build.gradle
kyriathar: Yes we no longer need the crypto Kafka clients within the crypto worker since Kafka relays we used to have within the crypto worker are now removed.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
kyriathar: I don't think that's needed anymore, the `CryptoOpsClient` was included within the crypto worker because of the Kafka relays.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
kyriathar: That was needed to write the lifecycle unit test for the crypto processor, because the crypto processor is not a `Lifecycle`.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
kyriathar: I just thought to leave to not increase PR's diff. Valid point, will address.

processors/crypto-processor/src/test/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImplTest.kt
kyriathar: This is testing code so shouldn't matter much, named parameters seem more overhead to me in this case.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
kyriathar: Not really sure. I suppose it also kills the background thread for this subscription.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
kyriathar: Yes I think it's impossible, in other case we just use some `it` reference to start them. Applied suggestion.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
fowlerrr: Yeah, it does exactly that.  Stops the background thread so no more callback from the old subscription (after whatever is pending, of course)

fowlerrr: Having said that, subscriptions are _usually_ a good candidate for managed resources as you'll want to close them automatically on stop/close _and_ here.

Usually if your mechanism is to always `close -> create new` in the same place and you aren't doing anything else with the subscription then you should consider a managed resource.

If you're doing something else with the subscriptions then it gets trickier.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
fowlerrr: I think the correct (but out of scope) action would be to expose any created coordinators in the `LifecycleTest` suite.  I might look into that at some point.  It would be better than exposing internals.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
kyriathar: Changed to use managed resources.

fowlerrr (dismissed):

fowlerrr (approved):

JamesHR3 (approved):

driessamyn (approved):
LGTM

PR #3906, opened 2023-05-17
CORE-13755 Add persistence metrics for membership

charlieR3:
Added metrics to capture both the overall membership persistence handler execution time and the membership persistence transaction times. This is because the recent performance issues we saw in the e2e test environment was caused by the time spent getting the database connection which only became clear after comparing the transaction times to the handler execution times. 

The Corda metrics library for timers give a breakdown of the 50th, 95th, and 99th percentile times, as well as count, sum and max metrics. 

Added dashboard "Membership" to "eks-e2e-dev". This will likely need to be moved to a base image.

On Grafana we can see an overview of all persistence operations for transaction time and handler times
Screenshot 2023-05-19 at 12 10 53

Here is a closer look at the handler execution times
Screenshot 2023-05-19 at 12 11 43

A more useful view is to click in to a specific persistence operation and percentile. This is showing the 50th percentile times for `QueryRegistrationRequest`
Screenshot 2023-05-19 at 12 12 45

And here is the transaction times breakdown
Screenshot 2023-05-19 at 12 13 23

And again, the 50th percentile view for QueryRegistrationRequest
Screenshot 2023-05-19 at 12 13 51

### DIFF
--- a/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/HandlerFactories.kt
+++ b/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/HandlerFactories.kt
@@ -1,5 +1,6 @@
 package net.corda.membership.impl.persistence.service.handler
 
+import io.micrometer.core.instrument.Timer
 import net.corda.crypto.cipher.suite.KeyEncodingService
 import net.corda.avro.serialization.CordaAvroSerializationFactory
 import net.corda.data.membership.db.request.MembershipPersistenceRequest
@@ -35,9 +36,12 @@ import net.corda.libs.platform.PlatformInfoProvider
 import net.corda.membership.lib.MemberInfoFactory
 import net.corda.membership.lib.exceptions.MembershipPersistenceException
 import net.corda.membership.mtls.allowed.list.service.AllowedCertificatesReaderWriterService
+import net.corda.metrics.CordaMetrics
+import net.corda.metrics.CordaMetrics.Metric
 import net.corda.orm.JpaEntitiesRegistry
 import net.corda.utilities.time.Clock
 import net.corda.virtualnode.read.VirtualNodeInfoReadService
+import java.util.concurrent.ConcurrentHashMap
 
 @Suppress("LongParameterList")
 internal class HandlerFactories(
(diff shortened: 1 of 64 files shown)


components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/HandlerFactories.kt
```diff
@@ -93,6 +98,27 @@ internal class HandlerFactories(
         UpdateStaticNetworkInfo::class.java to { UpdateStaticNetworkInfoHandler(persistenceHandlerServices) },
     )
 
+    private val handlerTimers = ConcurrentHashMap<String, Timer>()
+    private val transactionTimers = ConcurrentHashMap<String, Timer>()
```
dimosr: Are you doing this as an optimisation or is there another reason?

I think I read at some point micrometer already caches registered metrics, so you get the same instance.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistenceHandler.kt
```diff
@@ -42,6 +48,10 @@ internal abstract class BasePersistenceHandler<REQUEST, RESPONSE>(
     val platformInfoProvider get() = persistenceHandlerServices.platformInfoProvider
     val allowedCertificatesReaderWriterService get() = persistenceHandlerServices.allowedCertificatesReaderWriterService
 
+    private val transactionTimer by lazy {
```
dimosr: minor, but where is the need for lazy coming from?

dimosr (commented):
LGTM - just a few minor comments/questions

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/HandlerFactories.kt
charlieR3: I didn't know that. I was just trying to avoid additional work during every persistence operation to build the timer. Looking at the kdocs it seems like you are correct so I can remove this.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistenceHandler.kt
charlieR3: `get()` would be better actually and more in line with the rest of the class. I'm just trying to avoid the `Accessing non-final property operation in constructor` warning for the `operation` property.

dimosr (approved):

PR #3908, opened 2023-05-18
CORE-13455 Set limits on deployments from p2p scripts to match CI deployments

charlieR3:
This adds the same CPU and memory limits as done for the CI deployments so that pods aren't running without an uper limit and so that we can more easily reproduce issues often seen in the e2e tests.

### DIFF
--- a/applications/tools/p2p-test/app-simulator/scripts/deploy.sh
+++ b/applications/tools/p2p-test/app-simulator/scripts/deploy.sh
@@ -16,6 +16,7 @@ deploy() {
      oci://corda-os-docker.software.r3.com/helm-charts/corda-dev \
      --set image.registry="corda-os-docker.software.r3.com" \
      --set kafka.replicaCount=$KAFKA_REPLICAS,kafka.zookeeper.replicaCount=$KAFKA_ZOOKEEPER_REPLICAS \
+     -f prereqs-eks.yaml \
      --render-subchart-notes \
      --timeout 10m \
      --wait
@@ -24,6 +25,7 @@ deploy() {
    helm upgrade --install corda -n $namespace oci://corda-os-docker.software.r3.com/helm-charts/release/os/5.0/corda \
      --set "imagePullSecrets={docker-registry-cred}" --set image.tag=$DOCKER_IMAGE_VERSION \
      --set image.registry="corda-os-docker.software.r3.com" --values $REPO_TOP_LEVEL_DIR/values.yaml \
+     -f corda-eks.yaml \
      --values $REPO_TOP_LEVEL_DIR/debug.yaml --wait --version $CORDA_CHART_VERSION
 }
 
--- a/applications/tools/p2p-test/app-simulator/README.md
+++ b/applications/tools/p2p-test/app-simulator/README.md
@@ -298,3 +298,7 @@ Then using [helm-docs](https://github.com/norwoodj/helm-docs), generate the READ
 ```shell
 helm-docs
 ```
+
+# Deploying Corda
+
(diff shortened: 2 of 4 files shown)


yift-r3 (dismissed):

applications/tools/p2p-test/app-simulator/scripts/deploy.sh
```diff
@@ -24,6 +25,7 @@ deploy() {
    helm upgrade --install corda -n $namespace oci://corda-os-docker.software.r3.com/helm-charts/release/os/5.0/corda \
      --set "imagePullSecrets={docker-registry-cred}" --set image.tag=$DOCKER_IMAGE_VERSION \
      --set image.registry="corda-os-docker.software.r3.com" --values $REPO_TOP_LEVEL_DIR/values.yaml \
+     -f corda-eks.yaml \
```
williamvigorr3, line 28: Should we put this argument after `--values $REPO_TOP_LEVEL_DIR/debug.yaml` (to guard against the unlikely event that someone overrides the defaults in debug.yaml and that takes precedent over the new arguments).

applications/tools/p2p-test/app-simulator/scripts/deploy.sh
```diff
@@ -16,6 +16,7 @@ deploy() {
      oci://corda-os-docker.software.r3.com/helm-charts/corda-dev \
      --set image.registry="corda-os-docker.software.r3.com" \
      --set kafka.replicaCount=$KAFKA_REPLICAS,kafka.zookeeper.replicaCount=$KAFKA_ZOOKEEPER_REPLICAS \
+     -f prereqs-eks.yaml \
```
williamvigorr3, line 19: So before we were using the default from the top level values (`./corda-runtime-os/values.yaml`) which leaves these unset. We could add the defaults to `./corda-runtime-os-fix-script/values.yaml` here (and not override them in the test pipeline to make everything easier to deploy and prevent this problem from happening to other teams, who probably don't use this script but do use the top level values) but maybe we don't want to do that because we expect users to use the top level values.

The other advantage of setting them at the top level is, that if the values in the test pipeline get updated this situation could happen again.

williamvigorr3:
Missed one: Should we also update the README: `corda-runtime-os/applications/tools/p2p-test/app-simulator/scripts/README.md` and make it clear that we are using the same values as the CI pipeline (so hopefully we remember if we decide to use the script for performance testing)?

applications/tools/p2p-test/app-simulator/scripts/deploy.sh
charlieR3: I'm not sure about setting it globally. I think these limits should be set intentionally to suit the environment and scenario they are deployed in. Especially now when there's a focus on performance - I don't want to introduce a limit globally that might impact someone without them realising.

applications/tools/p2p-test/app-simulator/scripts/deploy.sh
charlieR3: We could. I was thinking that we might want the debug yaml to take precedence over other configs so that it can overwrite defaults (although those overrides would typically be debug related)

charlieR3:
> Missed one: Should we also update the README: corda-runtime-os/applications/tools/p2p-test/app-simulator/scripts/README.md and make it clear that we are using the same values as the CI pipeline (so hopefully we remember if we decide to use the script for performance testing)?

I can add a note. Although, I don't want to add specific numbers there since it might go out of sync if we change one and not the other. I think someone should look at the deploy command and yaml configurations to see the limits set and not the README. Could be useful to mention why they are what they are though.

applications/tools/p2p-test/app-simulator/scripts/deploy.sh
williamvigorr3: I think ideally we would define a few possible values which we have tested the platform with i.e. minimum supported values (for Corda to run reliably), DB has enough resources so doesn't limit Corda performance etc. Anyway I don't want to block this.

williamvigorr3 (approved):
LGTM

PR #3909, opened 2023-05-18
CORE-13740: Change the virtual node database connection pool to have 0 idle connections

yift-r3:
Make the virtual node DML database connection pool have more than one connection and set the minimal number of connections to 0.

The API pull request is in https://github.com/corda/corda-api/pull/1128

This is the first pull request of two; the second will use this function in the members and certificate persistence service. - see https://github.com/corda/corda-runtime-os/pull/3916


components/db/db-connection-manager/src/main/kotlin/net/corda/db/connection/manager/DbConfig.kt
```diff
@@ -34,6 +34,14 @@ fun DataSourceFactory.createFromConfig(config: SmartConfig): CloseableDataSource
     val driver = configWithFallback.getString(DatabaseConfig.JDBC_DRIVER)
     val jdbcUrl = configWithFallback.getString(DatabaseConfig.JDBC_URL)
     val maxPoolSize = configWithFallback.getInt(DatabaseConfig.DB_POOL_MAX_SIZE)
+    val minPoolSize = if(
+        configWithFallback.hasPath(DatabaseConfig.DB_POOL_MIN_SIZE)
+        && !configWithFallback.getIsNull(DatabaseConfig.DB_POOL_MIN_SIZE)
```
fowlerrr: `hasPath()` already checks for `null` so you don't need the second bit, I think.

yift-r3: Yes, that is correct.

libs/db/db-core/src/main/kotlin/net/corda/db/core/HikariDataSourceFactory.kt
```diff
@@ -66,8 +66,14 @@ class HikariDataSourceFactory(
         conf.isAutoCommit = isAutoCommit
         conf.isReadOnly = isReadOnly
         conf.maximumPoolSize = maximumPoolSize
-        conf.minimumIdle = minimumPoolSize
-        conf.idleTimeout = idleTimeout.toMillis()
+        if (minimumPoolSize != null) {
+            conf.minimumIdle = minimumPoolSize
+        }
```
kyriathar: Shouldn't we say that if minimumPoolSize == null then `conf.minimumIdle = maximumPoolSize`?

yift-r3: If it's null, we can leave it as is and let Hikari manage the default.

kyriathar (dismissed):

kyriathar (approved):

dimosr (approved):

PR #3911, opened 2023-05-18
ES-403/increase job frequency

seanbrereton:
Set job to run every 3 hours instead of nightly, also add a gradle tag to allow for filtering in gradle enterpirse

Tested here(failing test expected): https://ci02.dev.r3.com/job/Corda5/job/corda-runtime-os-run-unstable/job/seanb%2FES-403%2Fincrease-job-frequency/


yarivus (approved):
LGTM

PR #3913, opened 2023-05-18
CORE-13688 Refactoring redundant code in ParameterTypeLabel

julia-filipczak:
Refactoring & adding missing annotations.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FacadeServiceImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FacadeServiceImpl.kt
@@ -1,5 +1,6 @@
 package net.corda.flow.application.services.impl
 
+import co.paralleluniverse.fibers.Suspendable
 import net.corda.flow.application.services.impl.interop.dispatch.buildDispatcher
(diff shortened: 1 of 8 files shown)


jmacmahonr3 (commented):
Hi @julia-filipczak 
Do we need to add Suspendable to things like InterOpAction and other classes in the binding package ? 

Also is there a corresponding pr for API ? Using the MemberLookUp service as a guide we can see that Services in API have suspendable on the interface

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FacadeServiceImpl.kt
```diff
@@ -1,5 +1,6 @@
 package net.corda.flow.application.services.impl
 
+import co.paralleluniverse.fibers.Suspendable
```
chrisr3: This is the wrong `@Suspendable` annotation. Please use `net.corda.v5.base.annotations.Suspendable` instead, assuming you need to annotate it at all...?

szymonsztuka (approved):

yarivus (approved):
LGTM

szymonsztuka (approved):

PR #3914, opened 2023-05-19
CORE-5616 - Add metrics for gateway processor

dimosr:
## Changes
Adding an initial high-level set of metrics for the gateway processor.

## Testing
I enabled metrics temporarily in the CI and checked metrics were emitted successfully in CI. Some example below:
* inbound/outbound request latency (p99)
Screenshot 2023-05-19 at 13 24 26
Screenshot 2023-05-19 at 13 24 38

* inbound/outbound TLS connections (increase rate)

Screenshot 2023-05-19 at 13 24 48
Screenshot 2023-05-19 at 13 24 56

The only exception is revocation checks that are not easily exercisable with our current test suite, since most tests have revocation checks disabled. But, I think it's worth getting that in with everything else now, instead of waiting for manually testing it.

### DIFF
--- a/components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/BaseHttpChannelHandler.kt
+++ b/components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/BaseHttpChannelHandler.kt
@@ -1,5 +1,6 @@
 package net.corda.p2p.gateway.messaging.http
 
+import io.micrometer.core.instrument.Counter
 import io.netty.buffer.ByteBuf
 import io.netty.channel.ChannelHandlerContext
 import io.netty.channel.SimpleChannelInboundHandler
@@ -7,12 +8,16 @@ import io.netty.handler.codec.http.HttpObject
 import io.netty.handler.ssl.SslHandshakeCompletionEvent
 import io.netty.handler.ssl.SslHandshakeTimeoutException
 import io.netty.handler.timeout.IdleStateEvent
+import net.corda.metrics.CordaMetrics
 import org.slf4j.Logger
+import java.net.InetSocketAddress
+import java.net.SocketAddress
 import java.nio.channels.ClosedChannelException
 import javax.net.ssl.SSLException
 
 abstract class BaseHttpChannelHandler(private val eventListener: HttpConnectionListener,
-                                      private val logger: Logger): SimpleChannelInboundHandler<HttpObject>() {
+                                      private val logger: Logger,
+                                      private val handlerType: HandlerType): SimpleChannelInboundHandler<HttpObject>() {
 
     private companion object {
         val messagesNotToLog = listOf(
@@ -65,8 +70,10 @@ abstract class BaseHttpChannelHandler(private val eventListener: HttpConnectionL
                 if (evt.isSuccess) {
                     val ch = ctx.channel()
                     logger.info("Handshake with ${ctx.channel().remoteAddress()} successful")
+                    getConnectionMetric(ctx.channel().remoteAddress(), "success").increment()
                     eventListener.onOpen(HttpConnectionEvent(ch))
                 } else {
+                    getConnectionMetric(ctx.channel().remoteAddress(), "failure").increment()
                     val cause = evt.cause()
                     when {
                         cause is ClosedChannelException -> logger.warn("SSL handshake closed early")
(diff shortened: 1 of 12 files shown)


components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/BaseHttpChannelHandler.kt
```diff
@@ -65,8 +70,10 @@ abstract class BaseHttpChannelHandler(private val eventListener: HttpConnectionL
                 if (evt.isSuccess) {
                     val ch = ctx.channel()
                     logger.info("Handshake with ${ctx.channel().remoteAddress()} successful")
+                    getConnectionMetric(ctx.channel().remoteAddress(), "success").increment()
```
yift-r3, line 73: Should we put the `success` and `failure` somewhere in `CordaMetrics` (or, maybe, change the `ConnectionResult` to `ConnectionSuccess` and then record a boolean)?

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/BaseHttpChannelHandler.kt
```diff
@@ -117,4 +124,24 @@ abstract class BaseHttpChannelHandler(private val eventListener: HttpConnectionL
         ctx.close()
     }
 
+    private fun getConnectionMetric(remoteAddress: SocketAddress, connectionResult: String): Counter {
+        val metricsBuilder = when(handlerType) {
```
yift-r3, line 128: Maybe, instead of `HandlerType`, we can pass ` Metric<Counter>`? (and then we won't need the enum)?

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/internal/OutboundMessageHandler.kt
```diff
@@ -170,9 +172,12 @@ internal class OutboundMessageHandler(
             trustStore,
             clientCertificateKeyStore,
         )
+        val startTime = System.currentTimeMillis()
         val responseFuture = sendMessage(destinationInfo, gatewayMessage)
             .orTimeout(connectionConfig().responseTimeout.toMillis(), TimeUnit.MILLISECONDS)
         return responseFuture.whenCompleteAsync({ response, error ->
+            val requestLatency = System.currentTimeMillis() - startTime
+            getRequestTimer(peerMessage, response).record(requestLatency, TimeUnit.MILLISECONDS)
```
yift-r3, line 180: Do we want to record the existence of an error?

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/BaseHttpChannelHandler.kt
dimosr: Based on what I've seen so far, in `CordaMetrics` we add types for metrics and tags, but not values.

In terms of a boolean value, I think leaving it as a string provides some flexibility that might be useful in the long run. For example, if we want to record extra results in the future, such as timeouts, separately. I was also thinking that some of these tags might be re-usable by different parts of the system, so boolean might constrain other uses a bit.

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/BaseHttpChannelHandler.kt
dimosr: Thought about it a bit, but I am a bit on the fence.. Feels like to me that adding the enum and incorporating the logic in the base class is a bit easier to understand (and maybe makes it harder to forget thing or set up the timer inconsistently for each case) when compared to splitting it into the sub-classes and passing in the metric itself.

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/internal/OutboundMessageHandler.kt
dimosr: I think it's implicitly recorded. Whether the response type tag is populated or not indicates if the client received a response from the server or there was an error before that. 

We can add an extra tag if you think it would be beneficial, but not completely clear to me what that would look like.

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/BaseHttpChannelHandler.kt
yift-r3: Maybe make the `Metric<Counter>` an enum member and remove the switch statement?

dimosr: Not sure if that would make things better or worse.. at the moment, `HandlerType` is a quite simple type and extensions of the base class don't even have to know about metrics - it's all an internal detail of the base class. But, if we make this a member variable, then that's not the case anymore. 

Also, if we want to add more metrics in the future, adding more of them as member variables into the `HandlerType` seems like the wrong way to go.

yift-r3: I don't like the switch case statement every time we create a connection... Alternatively, we can put it as a lazy value.

dimosr: what is your concern - performance or something else?

Because performance-wise, I don't think it will matter, so I would be keen to not prematurely optimise unless we see this is a real issue. The time for this switch is going to be minimal compared to the overall time for connection establishment. And connections are established once and kept open.

yift-r3: It's not really performance as such (creating a connection is not really part of the critical path), but more about code tidiness...

dimosr: but in terms of tidiness, the current structure seems easier to understand for me. 

Everything is contained within a single method and when statements are exhaustive, so it seems simpler compared to splitting some parts of it in a different part of the codebase.

yift-r3: It looks strange to me... but it's just my subjective opinion, nothing that should block the pull request.

yift-r3 (approved):

PR #1923, opened 2022-08-18
CORE-6148: Close the entity manager instead of reusing it

yift-r3:
Closing the node entity manager factory after using it and recreating it again every time instead of using a cache to support a large number of concurrent nodes.

PR #3916, opened 2023-05-19
CORE-13740: Use predefine pool

yift-r3:
This will revert the changes from https://github.com/corda/corda-runtime-os/pull/1923. The changes in https://github.com/corda/corda-runtime-os/pull/3909 should reduce the minimum number of connections to zero, solving the potential issue of too many idle connections.


vkolomeyko (dismissed):

dimosr (commented):
LGTM - but I think to be on the safe side, it would be better to keep the flaky e2e tests disabled still. We can then monitor the jobs for a few days to confirm if we get 100% success rate across a lot of runs or we need to make more changes first.

dimosr (approved):

PR #3917, opened 2023-05-19
CORE-13870: Unique Kafka ClientId

jujoramos:

  instead of 'AtomicLong'.


LWogan:
It might be simpler to just add the worker instance.id to the existing client id to make it unique per JVM? The instance.id is supposed to be unique per worker and is passed to it on startup. Not sure if thats still true

jujoramos:
> It might be simpler to just add the worker instance.id to the existing client id to make it unique per JVM?

I considered that option at the beginning, the problem is that our helm chart currently doesn't set the `instanceId` when starting the worker pods. If we want to do that, we should move from a `Deployment` to `StatefulSet`, or use some other mechanism when deploying the chart so it's intelligent enough as to guarantee the uniqueness of the `client.id`.

jujoramos:
Should we also update the deployment files in https://github.com/corda/corda-soak-test-deployment/tree/release/5.0 to match these changes (the idea being that developers can deploy the same thing as the pipeline does to troubleshoot issue)?. And what about `release/5.1` (both for this repository and `corda-soak-test-deployment`), should it be updated as well?.

PR #3918, opened 2023-05-19
CORE-11990 - Modify signature handling

nikinagy:
Review how we handle signed data for member info and registration request. We should store and retrieve the data in its original serialised form instead of serialising and deserialising back and forth.

Added factory method to create PersistentMemberInfo and modified method which is creating MemberInfo from PersistentMemberInfo.

Removed QueryMemberSignatureHandler and retrieving signature information along with member information from calling QueryMemberInfo() during synchronisation and data distribution. Added status filter when querying for member information so that we can easily retrieve non-pending information for data packages.


components/membership/membership-group-read-impl/src/test/kotlin/net/corda/membership/impl/read/subscription/MemberListProcessorTest.kt
```diff
@@ -113,13 +134,21 @@ class MemberListProcessorTest {
                     topicData[aliceIdentity.shortHash.value + holdingIdentity.shortHash.value] = PersistentMemberInfo(
                         aliceIdentity.toAvro(),
                         member.memberProvidedContext.toAvro(),
-                        member.mgmProvidedContext.toAvro()
+                        member.mgmProvidedContext.toAvro(),
+                        SignedContexts(
+                            member.memberProvidedContext.toAvro().toByteBuffer(),
+                            member.mgmProvidedContext.toAvro().toByteBuffer(),
+                        ),
```
charlieR3, line 141: Is this a valid scenario? I feel like the member info factory or someplace in the code should validate against this. Either use the deprecated fields or use the new fields. Not both.

libs/membership/membership-impl/src/main/kotlin/net/corda/membership/lib/impl/MemberInfoFactoryImpl.kt
```diff
@@ -1,44 +1,128 @@
...
+        }
     }
+
+    override fun createPersistentMemberInfo(
+        viewOwningMember: HoldingIdentity,
+        memberInfo: MemberInfo,
+    ): PersistentMemberInfo = PersistentMemberInfo.newBuilder()
+        .setViewOwningMember(viewOwningMember)
+        .setMemberContext(memberInfo.memberProvidedContext.toAvro())
+        .setMgmContext(memberInfo.mgmProvidedContext.toAvro())
```
charlieR3, line 101: Should we be using these fields if they are deprecated?

components/membership/synchronisation-impl/src/test/kotlin/net/corda/membership/impl/synchronisation/MgmSynchronisationServiceImplTest.kt
```diff
@@ -347,12 +360,16 @@ class MgmSynchronisationServiceImplTest {
...
     )
 
-    private fun createSignatures(members: List<MemberInfo>) = members.associate {
+    private fun createSignedInfos(members: List<MemberInfo>) = members.map {
         val name = it.name.toString()
-        it.holdingIdentity to (CryptoSignatureWithKey(
-            ByteBuffer.wrap("pk-$name".toByteArray()),
-            ByteBuffer.wrap("sig-$name".toByteArray())
-        ) to CryptoSignatureSpec("dummy", null, null))
+        SignedMemberInfo(
```
charlieR3, line 365: `SignedMemberInfo` wasn't added in this PR but I'm not sure it correctly represents signed member infos. 

It contains a `MemberInfo` and a member context signature but there's no way to see the data that was signed for the member context and there's no signature for the mgm context

components/membership/membership-p2p/src/main/kotlin/net/corda/membership/p2p/helpers/MembershipPackageFactory.kt
```diff
@@ -58,31 +54,25 @@ class MembershipPackageFactory(
...
-                    signatureAndSpec.first to signatureAndSpec.second
-                } ?: throw CordaRuntimeException("Could not find member signature and signature spec for ${it.name}")
-
             SignedMemberInfo.newBuilder()
                 .setMemberContext(
                     SignedData(
-                        ByteBuffer.wrap(serializer.serialize(it.memberProvidedContext.toAvro())),
-                        memberSignature,
-                        memberSignatureSpec
+                        ByteBuffer.wrap(serializer.serialize(it.memberInfo.memberProvidedContext.toAvro())),
```
charlieR3, line 68: Ideally `net.corda.membership.lib.SignedMemberInfo` would expose the signed data so we don't need to serialise here.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/HandlerFactories.kt
```diff
@@ -70,7 +70,7 @@ internal class HandlerFactories(
         PersistGroupParameters::class.java to { PersistGroupParametersHandler(persistenceHandlerServices) },
         PersistGroupParametersInitialSnapshot::class.java to { PersistGroupParametersInitialSnapshotHandler(persistenceHandlerServices) },
         AddNotaryToGroupParameters::class.java to { AddNotaryToGroupParametersHandler(persistenceHandlerServices) },
-        QueryMemberSignature::class.java to { QueryMemberSignatureHandler(persistenceHandlerServices) },
+        QueryMemberSignature::class.java to { QueryMemberInfoHandler(persistenceHandlerServices) },
```
charlieR3, line 73: Will this cause a runtime exception since QueryMemberInfoHandler has `override fun invoke(context: MembershipRequestContext, request: QueryMemberInfo)`?

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoActionHandler.kt
```diff
@@ -76,11 +75,24 @@ class DistributeMemberInfoActionHandler(
...
+                listOf(MEMBER_STATUS_ACTIVE, MEMBER_STATUS_SUSPENDED)
+            )
+        val updatedSignedMemberInfo = when (updatedMemberQuery) {
+            is MembershipQueryResult.Success -> updatedMemberQuery.payload
+            is MembershipQueryResult.Failure -> return recordToRequeueDistribution(key, request) {
+                logger.warn("Failed to query for updated member's info: ${updatedMemberQuery.errorMsg}." +
+                        "Distributing the member info will be reattempted.")
             }
+        }.firstOrNull() ?: return recordToRequeueDistribution(key, request) {
+            logger.info("The MemberInfo retrieved from the message bus for ${updatedMember.x500Name} is not present yet. " +
```
charlieR3, line 91:
```suggestion
            logger.info("The MemberInfo retrieved from the database for ${updatedMember.x500Name} is not present yet. " +
```

libs/membership/membership-impl/src/main/kotlin/net/corda/membership/lib/impl/MemberInfoFactoryImpl.kt
charlieR3: I guess it's necessary just to maintain the backwards compatibility. That's annoying.

charlieR3 (commented):
I think our structures for member info are confusing now. I've commented more about that on the API PR

components/membership/membership-service-impl/src/main/kotlin/net/corda/membership/service/impl/actions/DistributeMemberInfoActionHandler.kt
```diff
@@ -76,11 +75,24 @@ class DistributeMemberInfoActionHandler(
         val updatedMember = request.updatedMember
 
         val groupReader = groupReaderProvider.getGroupReader(approvedBy.toCorda())
-        val updatedMemberInfo = groupReader.lookup(filter = ACTIVE_OR_SUSPENDED)
-            .firstOrNull { it.name.toString() == updatedMember.x500Name } ?: return recordToRequeueDistribution(key, request) {
-                logger.info("The MemberInfo retrieved from the message bus for ${updatedMember.x500Name} is not present yet. " +
-                    "Republishing the distribute command to be processed later when the MemberInfo is available.")
+        val updatedMemberQuery = membershipQueryClient
```
williamvigorr3, line 78: Is there a reason this doesn't query the message bus anymore. I am not sure the member info can be out of date from the database here (as we don't need to wait for the message bus to see the published record).

components/membership/membership-group-read-impl/src/test/kotlin/net/corda/membership/impl/read/subscription/MemberListProcessorTest.kt
nikinagy: As per this page: https://medium.com/expedia-group-tech/handling-incompatible-schema-changes-with-avro-2bc147e26770#:~:text=Apache%20Avro%20has%20the%20notion,can%20also%20have%20incompatible%20changes.
We could make the deprecated fields as defaults (although since we are trying to achieve backwards it's not needed). I think the main issue is that even if we add defaults we need to add it upon creation (so say explicitly that field is null), defaults are only handled well when you use the avro builders to create objects (like in MemberInfoFactoryImpl).
But yeah, we can still do something like this, if that's fine:
`PersistentMemberInfo(
                        aliceIdentity.toAvro(),
                        null,
                        null,
                        SignedContexts(
                            member.memberProvidedContext.toAvro().toByteBuffer(),
                            member.mgmProvidedContext.toAvro().toByteBuffer(),
                        ),`

PR #3919, opened 2023-05-19
CORE-13837: Add metrics for file descriptors and JVM heap pressure.

chrisr3:
Add extra metrics concerning "JVM memory pressure" and number of open file descriptors.


yarivus (approved):
LGTM

PR #3920, opened 2023-05-19
CORE-13974: Optimise CordaClassResolver for Kryo 5.5.

chrisr3:
Update `CordaClassResolver` to write class names as ASCII unless they actually _need_ UTF-8.

This optimisation was introduced in Kryo 5.x, although our `CordaClassResolver` was still based on the Kryo 4.x version of `DefaultClassResolver`.


driessamyn (approved):
LGTM

PR #3924, opened 2023-05-22
CORE-13952: Remove tracking of RPC subscription

fowlerrr:
We shouldn't need to track the subscription as part of the lifecycle.  This can lead to spurious ERROR messages when we don't correctly manage the tracking of the subscription and our handling of the subscription itself.  (e.g. we bring the subscription down but we haven't stopped tracking it)

Removing the subscription tracking clears that up and ensures we're consistent with the rest of Corda as well.

### DIFF
--- a/components/membership/certificates-service-impl/src/main/kotlin/net/corda/membership/certificate/service/impl/CertificatesServiceImpl.kt
+++ b/components/membership/certificates-service-impl/src/main/kotlin/net/corda/membership/certificate/service/impl/CertificatesServiceImpl.kt
@@ -66,7 +66,6 @@ class CertificatesServiceImpl internal constructor(
     }
 
     private var registrationHandle: AutoCloseable? = null
-    private var subscriptionRegistrationHandle: AutoCloseable? = null
     private var configHandle: Resource? = null
     private var rpcSubscription: Resource? = null
     private val coordinator = coordinatorFactory.createCoordinator<CertificatesService>(::handleEvent)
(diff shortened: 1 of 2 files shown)


fowlerrr:
I should note that the aforementioned ERROR was causing the DB resiliency test to fail as it scrapes the logs for errors and fails when it sees them.  So, even though this ERROR isn't really a problem in running the code it's breaking a test (and won't look great in production logs)

components/membership/certificates-service-impl/src/main/kotlin/net/corda/membership/certificate/service/impl/CertificatesServiceImpl.kt
```diff
@@ -90,51 +90,36 @@ class CertificatesServiceImpl internal constructor(
...
+                        coordinator,
+                        setOf(ConfigKeys.BOOT_CONFIG, ConfigKeys.MESSAGING_CONFIG)
+                    )
                 } else {
-                    logger.warn("Unexpected event $event.")
+                    configHandle = null
+                    rpcSubscription?.close()
+                    rpcSubscription = null
                 }
+                coordinator.updateStatus(event.status)
```
yift-r3: Shouldn't we set the state to UP only after we got the messaging configuration and started the subscription?

LWogan: Yes

fowlerrr: Yes, it's done both ways in the code but this makes sense.  Moving it

yift-r3 (approved):
LGTM

PR #3930, opened 2023-05-22
CORE-13258 Avoid reaching out to database when there are no staterefs to resolve.

vlajos:
CORE-13258 Avoid reaching out to database when when there are no staterefs to resolve.

Performance tests: https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4505141266/Test+7+-+Reduce+time+spent+resolving+querying+stateRefs+within+ledger+flows+for+empty+inputs

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerStateQueryServiceImpl.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerStateQueryServiceImpl.kt
@@ -24,7 +24,7 @@ import org.osgi.service.component.annotations.Reference
 import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
 
 @Component(
-    service = [ UtxoLedgerStateQueryService::class, UsedByFlow::class ],
+    service = [UtxoLedgerStateQueryService::class, UsedByFlow::class],
     scope = PROTOTYPE
 )
 class UtxoLedgerStateQueryServiceImpl @Activate constructor(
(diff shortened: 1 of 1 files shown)


lankydan (dismissed):

blsemo (approved):
LGTM

mnesbit (approved):
Makes sense

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerStateQueryServiceImpl.kt
```diff
@@ -37,26 +37,30 @@ class UtxoLedgerStateQueryServiceImpl @Activate constructor(
...
     @Suspendable
     override fun resolveStateRefs(stateRefs: Iterable<StateRef>): List<StateAndRef<*>> {
         return recordSuspendable({ ledgerPersistenceFlowTimer(ResolveStateRefs) }) @Suspendable {
-            wrapWithPersistenceException {
-                externalEventExecutor.execute(
-                    ResolveStateRefsExternalEventFactory::class.java,
-                    ResolveStateRefsParameters(stateRefs)
-                )
-            }.map { it.toStateAndRef<ContractState>(serializationService) }
+            if (stateRefs.count() == 0) {
```
mnesbit, line 54: My only concern here is that Iterable is not necessarily readable twice, so really API should be List<StateRef>, but it is almost certainly just a List in practice

mnesbit: Having said that do make damned sure that it does work when staterefs are non-empty

yarivus (approved):
LGTM

fenryka (approved):
LGTM

PR #3932, opened 2023-05-22
CORE-13870: Unique Kafka ClientId

jujoramos:

  instead of 'AtomicLong'.

### DIFF
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/config/ResolvedSubscriptionConfig.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/config/ResolvedSubscriptionConfig.kt
@@ -20,7 +20,7 @@ data class ResolvedSubscriptionConfig(
     val subscriptionType: SubscriptionType,
     val topic: String,
     val group: String,
-    val idCounter: Long,
+    val uniqueId: String,
     val instanceId: Int,
     val pollTimeout: Duration,
     val threadStopTimeout: Duration,
@@ -38,20 +38,20 @@ data class ResolvedSubscriptionConfig(
          * @param subscriptionType Type of subscription.
          * @param subscriptionConfig User configurable values for a subscription.
          * @param messagingConfig Messaging smart config.
-         * @param counter Client counter.
+         * @param uniqueId Unique id, used to uniquely identify the subscription.
          * @return concrete class containing all config values used by a subscription.
          */
         fun merge(
             subscriptionType: SubscriptionType,
             subscriptionConfig: SubscriptionConfig,
             messagingConfig: SmartConfig,
-            idCounter: Long
+            uniqueId: String
         ): ResolvedSubscriptionConfig {
             return ResolvedSubscriptionConfig(
                 subscriptionType,
                 subscriptionConfig.eventTopic,
                 subscriptionConfig.groupName,
-                idCounter,
+                uniqueId,
                 messagingConfig.getInt(INSTANCE_ID),
                 Duration.ofMillis(messagingConfig.getLong(POLL_TIMEOUT)),
                 Duration.ofMillis(messagingConfig.getLong(THREAD_STOP_TIMEOUT)),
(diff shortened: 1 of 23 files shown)


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/config/ResolvedSubscriptionConfig.kt
```diff
@@ -64,7 +64,6 @@ data class ResolvedSubscriptionConfig(
         }
     }
 
-    val clientId = "$subscriptionType-$group-$topic-$idCounter"
-    val loggerName = clientId
+    val clientId = "$subscriptionType-$group-$topic-$uniqueId"
     val lifecycleCoordinatorName = LifecycleCoordinatorName("$topic-$subscriptionType-$group", clientId)
```
fowlerrr: I wonder if there needs to be this overlap?  Or, if we do want it, should we at least make it the same ordering?  `$subscriptionType-$group-$topic` makes sense to me as I think it's more likely to go from broad to narrow

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/listener/LoggingConsumerRebalanceListener.kt
```diff
@@ -5,30 +5,26 @@ import net.corda.messagebus.api.consumer.CordaConsumerRebalanceListener
...
+    open val log: Logger = LoggerFactory.getLogger("${this.javaClass.name}-${clientId}")
 
     /**
      * When a [partitions] are revoked write to the log.
      */
     override fun onPartitionsRevoked(partitions: Collection<CordaTopicPartition>) {
-        val partitionIds = partitions.map{ it.partition }.joinToString(",")
-        log.info("Consumer ($clientId) group name $groupName for topic $topic partition revoked: $partitionIds.")
+        val partitionIds = partitions.map { it.partition }.joinToString(",")
+        log.info("Partitions revoked: $partitionIds.")
```
fowlerrr, line 20: Do we have MDC here giving some detail about which listener is having partitions revoked?

fowlerrr: `clientId` will be in the logger name, sure.  Just checking if it's clear enough in case we need to know.

fowlerrr (dismissed):
Pretty much happy with it.  Just a couple of suggestions.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/listener/LoggingConsumerRebalanceListener.kt
jujoramos: We don't have any `MDC` set with these properties, no. I removed `clientId`, `groupName` and `topic` from the log message because they are already included within the `loggerName` (as you correctly stated) and the duplication ends up causing confusion. Do you see any problems with that (keep in mind that `groupName` and `topic` are already part of the `clientId`, so we are not loosing any information here)?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/config/ResolvedSubscriptionConfig.kt
jujoramos: Thought about this but forgot to make the change, thanks.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/listener/LoggingConsumerRebalanceListener.kt
fowlerrr: It's all still parseable/searchable, right?  Probably okay.

fowlerrr (approved):
LGTM

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/listener/LoggingConsumerRebalanceListener.kt
jujoramos: Yep, all still parseable and searchable 👍

rick-r3 (approved):

PR #3935, opened 2023-05-22
CORE-14578 Add transaction ID to MDC properties

nkovacsx:
### Overview

The requirement is to add the `corda.transactionId` property to log messages to produce something like this:

```
2023-05-22 15:29:12.205 [Thread-77] INFO  net.corda.flow.external.events.impl.ExternalEventManagerImpl {corda.client.id=58674ba1-b178-4bb9-bb95-e47db7a10eac, corda.external.event.id=6a0837c2-3712-40
7a-9c44-9391366ad2f4, corda.transactionId=SHA-256D:73767A812847482B0F6016C883913036E01F6C07DC08FDF12022C4E7A4E532E1, flow.id=f905a38b-d6a2-4b16-9646-9f9ff4553976, vnode.id=DF2A5B3D8589} - Processing
 response for external event with id '6a0837c2-3712-407a-9c44-9391366ad2f4'
```

We added a logic where adding a property to the flow context that is prefixed with `corda.logged` will result in that property passed to the MDC and thus logged. 

In `UtxoFinalityFlowV1` we'll add the `corda.logged.transactionId` to the flow context properties. This way the transaction id will be logged throughout the processors.


lankydan (dismissed):
I'm happy with this.

@simon-johnson-r3 can you have a look at this to check you're happy with the flow context properties usage.

@LWogan can you also look since you did the original MDC implementation iirc.

lankydan:
@nkovacsx I forgot one thing, you should do the same in the receive finality flow as well.

LWogan:
this looks fine to me from MDC usage perspective

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
```diff
@@ -48,7 +50,22 @@ class FlowEventProcessorImpl(
     ): StateAndEventProcessor.Response<Checkpoint> {
         val flowEvent = event.value
         val mdcProperties = flowMDCService.getMDCLogging(state, flowEvent, event.key)
-        return withMDC(mdcProperties) {
+
+        val loggedContextProperties = try {
```
simon-johnson-r3: `getMDCLogging` takes `state` and returns `mdcProperties`, then here you get more properties from `state`
would it be better to roll this into `getMDCLogging`?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
```diff
@@ -48,7 +50,22 @@ class FlowEventProcessorImpl(
...
+            state?.flowState
+                ?.let(::FlowStateManager)
```
simon-johnson-r3: In two minds about this. It feels quite heavyweight (and out of its original scope) to create a `FlowStateManager` to extract context properties, and it copies the entire flow state. Although I do see why that is convenient also, and replicating the code to read them directly would create some duplication.

You could instead consider creating just a FlowStackImpl from the stack directly (this would make it mutable but you'd not be mutating it).

@LWogan what are your thoughts on this?

libs/utilities/src/main/kotlin/net/corda/utilities/LogUtils.kt
```diff
@@ -78,6 +81,12 @@ fun <R> withMDC(mdcProperties: Map<String, String>, block: () -> R) : R {
     }
 }
 
+fun Map<String, String>.selectLoggedProperties() = filter {
```
simon-johnson-r3: I'm not so keen on using extension functions like this because it's not useful generically to call this on a `Map<String, String>`. An extension of `ExternalEventContext` so you'd call e.g. `request.flowExternalEventContext.contextPropertiesForLogging` might be more appropriate as that is generic to `ExternalEventContext`?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
nkovacsx: Good point, I'll move this logic to `getMDCFromCheckpoint` as we can only extract this if we have a checkpoint.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
nkovacsx: I'm happy to modify this if needed, let's wait for @LWogan

libs/utilities/src/main/kotlin/net/corda/utilities/LogUtils.kt
nkovacsx: Problem is we have very few dependencies in the `:libs:utilities` module and not sure if we want to introduce more.

simon-johnson-r3: I'd argue that's because this is in the wrong place. It's not function of logging so much as of external event handling. I'd also suggest putting the context prefixes in here is the wrong place, not least because I expect `corda` is defined elsewhere already and you've copied it here.

nkovacsx: Where would you suggest placing this then?

simon-johnson-r3: It's difficult because you pull the flow context properties from both external events and flow state, even thought they have nothing in common. So there are no obvious common libs. Possibly flow-utils, given they are flow context properties, which would just introduce a dependency to flow-utils on the crypto service, which is a little unusual although it already has a flow lib dependency in external-event-response. And arguably it's starting to deal in flow things at the point it's looking at flow context properties anyway.

What is certain is `corda.` should be taken from `FlowContextProperties.CORDA_RESERVED_PREFIX` which is part of the `net.corda.v5.application.flows` api package. Otherwise we have to maintain it in two places, and there's no obvious link between them.

I don't have a good answer right now, perhaps just fix the re-definition of `corda.`. Otherwise my suggestion would perhaps be not to try and make this a common method given it's only short anyway.

simon-johnson-r3: Possibly we can just make this explicit, e.g.:
```
translateFlowContextToMDC(flowContextProperties: Map<String, String>) : Map<String, String>
```

nkovacsx: applied your suggestions

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
LWogan: I'd agree with Simon that its a bit heavyweight,  but i dont feel strongly either way.

If you created a FlowStackImpl you'd need to duplicate the code in the FlowContext method flattenPlatformProperties to get the properties out.

You'd ideally want to build FlowStackBasedContext object to reduce that duplication using the state as input to build the FlowStack first. If you go that route, you probably wouldn't be able to reuse the builder in FlowStateManager as that class needs direct access to the stack and the flowContext.

nkovacsx: Will leave it as it is for now, raised https://r3-cev.atlassian.net/browse/CORE-14185

simon-johnson-r3 (dismissed):

LWogan (dismissed):
LGTM

lankydan (dismissed):

simon-johnson-r3 (dismissed):

lankydan (dismissed):

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowMDCServiceImpl.kt
```diff
@@ -78,9 +80,25 @@ class FlowMDCServiceImpl : FlowMDCService {
             MDC_CLIENT_ID to startContext.requestId,
             MDC_FLOW_ID to flowId
         )
+
+        // Extract properties starting with `corda.logged`
+        val loggedContextProperties = try {
+            translateFlowContextToMDC(
+                state.flowState
+                // TODO CORE-14185 Instantiating the FlowStateManager is heavyweight, find a better way to do this
+                .let(::FlowStateManager)
```
mnesbit: This looks pretty horrid and setting up a try block isn't free either even if no exception is thrown. Can we check there isn't a simpler approach to extracting this stuff?

nkovacsx: @LWogan and @simon-johnson-r3 mentioned something about the FlowStackImpl but not sure whether that would be any better tbh.

nkovacsx: I managed to re-write it like this: (no try/catch and no FlowStateManager instantiation)
```
val platformProperties = state.flowState?.flowStackItems?.let {
            FlowStackImpl(state.flowState.flowStackItems).flowStackItems.flatMap { flowStackItem ->
                flowStackItem.contextPlatformProperties.items.map { keyValuePair ->
                    keyValuePair.key to keyValuePair.value
                }
            }.toMap()
        } ?: emptyMap()
```

Would this be better @mnesbit @simon-johnson-r3 ?

mnesbit (dismissed):
OK

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowMDCServiceImpl.kt
```diff
@@ -78,9 +81,19 @@ class FlowMDCServiceImpl : FlowMDCService {
             MDC_CLIENT_ID to startContext.requestId,
             MDC_FLOW_ID to flowId
         )
+
+        val platformProperties = state.flowState?.flowStackItems?.let {
+            FlowStackBasedContext(
+                FlowStackImpl(state.flowState?.flowStackItems!!)
```
simon-johnson-r3: You're only in here if `state.flowState?.flowStackItems` is non null, so the !! should be redundant. The reason it's not is because you are not using the object reference passed through (i.e. the `it` although I'd give it a name like `flowStackItems`) and the compiler recognises the value could have changed between line 87 and 85.

nkovacsx: Thanks! I know it cannot be smart casted but I figured it's not worth creating a separate `val` just for this, having `!!` should be safe anyway. If you feel like it's worth moving it to a `val` and letting the compiler smart cast this I'm happy to do it.

simon-johnson-r3: I'm saying you already have a val for it. It's the `it` given to you by the `let` scope function.

```
val platformProperties = state.flowState?.flowStackItems?.let {
            FlowStackBasedContext(
                FlowStackImpl(it)
```
should do it, but
```
val platformProperties = state.flowState?.flowStackItems?.let { flowStackItems ->
            FlowStackBasedContext(
                FlowStackImpl(flowStackItems)
```
would be more readable.

nkovacsx: not sure how I missed this 🤦 Thanks, it sure is Friday!

mnesbit (approved):
OK

simon-johnson-r3 (approved):

PR #3938, opened 2023-05-22
CORE-13908: signing cache fixes

dickon:
Extend signing key metadata cache implementation in SigningServiceImpl to be used properly when signing from flows. Improve and extend test coverage while reducing duplication.

### DIFF
--- a/components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
+++ b/components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
@@ -121,14 +121,16 @@ class SigningServiceImpl(
         tenantId: String,
         keyIds: List<ShortHash>,
     ): Collection<SigningKeyInfo> {
+        // Since we are looking up by short IDs there's nothing we can do to handle clashes
+        // on short IDs in this case, at this layer. We must therefore rely on the database
+        // uniqueness constraints to stop clashes from being created.
+
         val cachedKeys =
             cache.getAllPresent(keyIds.mapTo(mutableSetOf()) {
                 CacheKey(tenantId, it)
             }).mapTo(mutableSetOf()) { it.value }
-        // TODO handle short ID clashes
         if (cachedKeys.size == keyIds.size) return cachedKeys
         val notFound: List<ShortHash> = keyIds - cachedKeys.map { it.id }.toSet()
-
         val fetchedKeys = signingRepositoryFactory.getInstance(tenantId).use {
             it.lookupByPublicKeyShortHashes(notFound.toMutableSet())
         }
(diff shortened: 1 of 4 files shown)


components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
```diff
@@ -126,6 +125,7 @@ class SigningServiceImpl(
             cache.getAllPresent(keyIds.mapTo(mutableSetOf()) {
                 CacheKey(tenantId, it)
             }).mapTo(mutableSetOf()) { it.value }
+        // TODO handle short ID clashes
```
rick-r3: Is this referring to ambiguity in the API as to what happens if there are duplicates in the `keyIds` list.  Right now the returned collection contains each key only once and in no particular order which could be confusing to a caller perhaps, however it does work, even if there are duplicates.

rick-r3: Actually, scratch that as you may go to the database if `notFound` is empty.  Maybe you could just put at line 131:

`if(notFound.isEmpty()) return cachedKeys`

?

dickon: Yes, and I think I agree there's nothing we can do to check for clashes in this function. I'll note that in a comment, which won't need retesting.

I believe `notFound` cannot be empty, since in that case cachedKeys.size == keyIds.size, which we check at line 129: 

`        if (cachedKeys.size == keyIds.size) return cachedKeys`

rick-r3: I don't think this is true.  If the input `List` is of length 2, with the same hash in there twice, `cachedKeys.size` will be 1 and `keyIds.size` will be 2.  So you fall through.  Then the implementation of `minus` results in `notFound` being empty because both will be filtered out by the implementation:

https://github.com/JetBrains/kotlin/blob/924c28507067cbfbf78a6509ea89eabe496e34ca/libraries/stdlib/common/src/generated/_Collections.kt#L3157

I think it is worth a test, and the extra line of code.

dickon: Mmm, yes, though I don't think it's likely since why would caller waste bits in the Avro payload on asking for the same public key hash twice? 

However, since the list comes from Avro and I don't think Avro supports sets and the test is very cheap we may as well.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
```diff
@@ -122,13 +121,16 @@ class SigningServiceImpl(
...
+        // on short IDs in this case, at this layer. We must therefore rely on the database
+        // uniqueness constraints to stop clashes from being created.
+
         val cachedKeys =
             cache.getAllPresent(keyIds.mapTo(mutableSetOf()) {
                 CacheKey(tenantId, it)
             }).mapTo(mutableSetOf()) { it.value }
         if (cachedKeys.size == keyIds.size) return cachedKeys
         val notFound: List<ShortHash> = keyIds - cachedKeys.map { it.id }.toSet()
-
```
rick-r3, line 131: As per slack comment, you need `if (notFound.isEmpty()) return cachedKeys` here too.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
dickon: Implemented.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
dickon: done

rick-r3 (approved):

PR #3939, opened 2023-05-22
CORE-13987 Ledger metrics

lankydan:
Add metrics for:

- `flow.crypto.time`
- `ledger.flow.verification.time`
- `ledger.verification.contract.total.time`
- `ledger..verification.contract.time`
- `ledger.verification.contract.count`
- `ledger.flow.persistence.time`
- `ledger.backchain.resolution.chain.length`
- `serialization.amqp.serialization.time`
- `serialization.amqp.deserialization.time`

This involved wiring in `CurrentSandboxGroupContext` to a lot of services and making it available in all sandboxes.

The `SeriazliationService` implementations in `:flow-service` and `SerializationServiceProxy` in other sandboxes provide sreialization metrics without adding directly to the serialization library.

`recordSuspendable` added to allow recording within fibers, which is needed do to `Timer`s being closable and failing serialization.

`recordOptionally` added to allow recording of `Timer` metrics that are only recorded if they surpass a certain duration. This is used by the serialization metrics.


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/processor/PersistenceRequestProcessor.kt
```diff
@@ -62,6 +64,9 @@ class PersistenceRequestProcessor(
                             .toSet()
 
                         val sandbox = entitySandboxService.get(holdingIdentity, cpkFileHashes)
+
+                        currentSandboxGroupContext.set(sandbox)
```
JamesHR3, line 68: Out of interest, why did you need to set the `currentSandboxGroupContext` here?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerGroupParametersPersistenceServiceImpl.kt
```diff
@@ -32,27 +38,39 @@ class UtxoLedgerGroupParametersPersistenceServiceImpl @Activate constructor(
...
-                PersistSignedGroupParametersIfDoNotExistExternalEventFactory::class.java,
-                with(signedGroupParameters) {
-                    PersistSignedGroupParametersIfDoNotExistParameters(
-                        bytes,
-                        signature,
-                        signatureSpec
-                    )
-                }
-            )
+        recordSuspendable({ ledgerPersistenceFlowTimer("findUnconsumedStatesByType") }) @Suspendable {
```
JamesHR3: Does this timer have the right string associated with it?

Probably worth putting the timer strings as constants somewhere.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerPersistenceServiceImpl.kt
```diff
@@ -123,4 +137,13 @@ class UtxoLedgerPersistenceServiceImpl @Activate constructor(
         }
 
     private fun serialize(payload: Any) = ByteBuffer.wrap(serializationService.serialize(payload).bytes)
+
+    // TODO align with ledger persistence operation names
+    private fun ledgerPersistenceFlowTimer(operationName: String): Timer {
```
JamesHR3: I wouldn't be averse to making the operation name an enum, although that might be a touch overkill.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerGroupParametersPersistenceServiceImpl.kt
```diff
@@ -32,27 +38,39 @@ class UtxoLedgerGroupParametersPersistenceServiceImpl @Activate constructor(
 
     @Suspendable
     override fun find(hash: SecureHash): SignedGroupParameters? {
-        return wrapWithPersistenceException {
-            externalEventExecutor.execute(
-                FindSignedGroupParametersExternalEventFactory::class.java,
-                FindSignedGroupParametersParameters(hash.toString())
-            )
-        }.singleOrNull()
+        return recordSuspendable({ ledgerPersistenceFlowTimer("findUnconsumedStatesByType") }) @Suspendable {
```
JamesHR3: Think this is also the wrong name.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerGroupParametersPersistenceServiceImpl.kt
```diff
@@ -32,27 +38,39 @@ class UtxoLedgerGroupParametersPersistenceServiceImpl @Activate constructor(
...
+                            signature,
+                            signatureSpec
+                        )
+                    }
+                )
+            }
         }
     }
+
+    private fun ledgerPersistenceFlowTimer(operationName: String): Timer {
```
JamesHR3: Is this the correct metric that you're recording under?

components/ledger/ledger-verification/src/main/kotlin/net/corda/ledger/verification/processor/impl/VerificationRequestProcessor.kt
```diff
@@ -58,15 +59,18 @@ class VerificationRequestProcessor(
...
+                        currentSandboxGroupContext.set(sandbox)
+
                         requestHandler.handleRequest(sandbox, request)
                     } catch (e: Exception) {
                         errorResponse(request.flowExternalEventContext, e)
+                    } finally {
+                        currentSandboxGroupContext.remove()
                     }.also {
-                        CordaMetrics.Metric.LedgerTransactionVerificationTime
+                        CordaMetrics.Metric.Ledger.TransactionVerificationTime
```
JamesHR3, line 71: Did you take out the UTXO ledger type because there are no others that are relevant?

libs/ledger/ledger-utxo-transaction-verifier/src/main/kotlin/net/corda/ledger/utxo/transaction/verifier/UtxoLedgerTransactionContractVerifier.kt
```diff
@@ -13,30 +15,54 @@ import net.corda.v5.ledger.utxo.transaction.UtxoLedgerTransaction
...
-                    exceptionMessage = ex.message ?: "The thrown exception did not provide a failure message."
-                )
-            )
-        }
-    }
 
-    if (failureReasons.isNotEmpty()) {
-        throw ContractVerificationException(transaction.id, failureReasons)
-    }
+    CordaMetrics.Metric.Ledger.ContractVerificationTime
```
JamesHR3, line 22: This function is starting to buckle a little under the weight of metrics. Worth splitting out the contract verify step into its own function to keep the nesting a little more sensible?

libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
```diff
@@ -461,6 +519,13 @@ object CordaMetrics {
          */
         LedgerType("ledger.type"),
 
+        /**
+         * The contract class name of a contract being executed.
+         */
+        LedgerContractName("ledger.contract.name"),
+
+        SerializedClass("serialized.class"),
```
JamesHR3, line 530: Missing a kdoc here.

We should however think about whether this is going to cause a cardinality problem. It's certainly possible in more complicated applications, and it would be harder to take out once present.

JamesHR3 (commented):
One thing that occurs to me reading this is that at the ledger layer we may be more interested in transactions than flows, so I wonder if there's a useful metric about transaction persistence (or finalization?) rate per virtual node?

I'm also wondering if it's worth timing individual ledger operations from within a flow when we get some timings around that from how long the external event processors take (or even from how long we suspend with certain external event actions). Overall timings of how long a flow took inside particular ledger operations might still be interesting though. Do you feel you get better data for individual events by doing this?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/processor/PersistenceRequestProcessor.kt
lankydan: Because it wasn't previously being used in this processor and the `SerializationService` now uses it to record metrics.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerGroupParametersPersistenceServiceImpl.kt
lankydan: Ah thanks, they should just be the method names atm.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerPersistenceServiceImpl.kt
lankydan: We could do that in a later change imo.

components/ledger/ledger-verification/src/main/kotlin/net/corda/ledger/verification/processor/impl/VerificationRequestProcessor.kt
lankydan: Yes.

libs/ledger/ledger-utxo-transaction-verifier/src/main/kotlin/net/corda/ledger/utxo/transaction/verifier/UtxoLedgerTransactionContractVerifier.kt
lankydan: It doesn't really bother me because even with the metrics making the method bigger, it is still easy to follow.

libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
lankydan: Discussed with @JamesHR3, we will resolve this by only recording metrics for classes that take longer than 1ms to serialize or deserialize.

JamesHR3 (approved):

driessamyn (approved):
LGTM

PR #3943, opened 2023-05-23
CORE-13870: Unique Kafka ClientId (#3932)

jujoramos:

  instead of 'AtomicLong'.

(cherry picked from commit 07ef2715ed8bd42905056d8b8311d153fb8665bf)


jujoramos:
Closing as cherry-picking seems to be forbidden (forward merge should be used instead).

PR #3944, opened 2023-05-23
CORE-14040: Add metrics for events processed per flow and suspension count

JamesHR3:
Adds two new metrics to the flow pipeline:

- Total count of events processed for a single flow
- Total suspension count in a flow.

In the steady state these should be constant provided the platform is running flows, so a change in the counts here could indicate some issue with the platform (e.g. higher than normal session replay counts).

Some sample graphs using these metrics below:

event-suspension-count

This shows a sample flow using sessions. The first graph shows the total event count for the flow and its responder. The second shows the percentage of events that led to a fiber suspension - i.e. the percentage that resulted in user code being executed.


JamesHR3:
Draft until better names chosen.

emilybowe (dismissed):
LGTM

rick-r3 (approved):

PR #3865, opened 2023-05-12
CORE-13164 Interop / Facade Demo: Add Facades

ac101m:
This PR adds flows which uses Facade calls (client and server-side dispatchers) to existing interop token flows.
FacadeService contains 3 hard-coded Tokens facade definitions:  
"org.corda.interop/platform/tokens/v1.0", "org.corda.interop/platform/tokens/v2.0", and "org.corda.interop/platform/tokens/v3.0".
Added several testing flows, the most important 2 are `SimpleSwapFlow` and `SwapFlow`.
Added deployment script to deploy 2 networks with notaries (needed for any UTXO like Cordapps).
Minor fixes to Facade Proxy - added few `@Serializable` and fix/workaround for `Method` object equality issue in FacadeInterfaceBinding.kt.

PR #3945, opened 2023-05-23
CORE-12555 Interop / Fixes to allow suspending Weft proxy

szymonsztuka:
Update to Quasar version (0.9.1 SNAPSHOT) with improved Java Dynamic Proxy serialization. Avoid serialization issues with `java.lang.reflect.Constructor<*>` as a constructor parameter by replacing it with `Class<*>`. We anyway require a single constructor for the output parameters class, so when it is needed we can derive the given constructor from the `Class<*>` object.
Add several `@Suspendable` (maybe a bit prematurely).

This is extracted code from a bigger PR https://github.com/corda/corda-runtime-os/pull/3865


ronanbrowne (approved):
LGTM ( build changes )

build.gradle
```diff
@@ -131,7 +131,7 @@ allprojects {
         mavenLocal()
 
         maven {
-            url = "$artifactoryContextUrl/corda-dependencies"
+            url = "$artifactoryContextUrl/corda-dependencies-dev"
             content {
                 // For Quasar release.
```
chrisr3: Should now say
```groovy
// For Quasar SNAPSHOT.
```
because `corda-dependencies-dev` is a SNAPSHOT repository.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/binding/FacadeInterfaceBinding.kt
```diff
@@ -108,8 +108,16 @@ sealed class FacadeOutParameterBindings {
...
+    data class DataClassOutParameterBindings(private val clazz: Class<*>, val bindings: List<DataClassPropertyBinding>)
         : FacadeOutParameterBindings() {
+        init {
+            require(clazz.constructors.size == 1) {
+                "Facade ParameterBinding supports classes with a single constructor only."
+            }
+        }
+
+        val constructor: Constructor<*>
+            get() = clazz.constructors.single()
```
chrisr3: Can you be more precise that just using `single()`? AMQP has a similar problem because it needs to identify a class's "primary constructor" but cannot assume that the class was written in Kotlin. It solves this by requiring the constructor be annotated as `@ConstructorForDeserialization` in the case that there's a choice.

The other option is requiring a "no-args" constructor, which would allow you to use `clazz.getConstructor()`.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/binding/internal/BindingContext.kt
```diff
@@ -445,6 +448,7 @@ private class DataClassOutParametersBindingContext(
 
 }
 
+@Suspendable
 private inline fun <reified T : Annotation> AnnotatedElement.readAnnotation(): T? =
     if (isAnnotationPresent(T::class.java)) getAnnotation(T::class.java) else null
```
chrisr3: Does an `inline` function _need_ to be `@Suspendable`? I don't think it should exist on the call-stack at runtime. Does Quasar even have anything to instrument?

chrisr3 (commented):
Just a few thoughts.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/binding/FacadeInterfaceBinding.kt
szymonsztuka: Thank for pointing out this. So far we tested with Kotlin data classes only (and so I don't expect impose only "no-args" constructor) , I think we create ticket to make sure to retest with with Java.  If we require a single, e.g. default constructor for class written in Java + setters/getters, then I think this code would work as well, or there is something more complex issue?

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/binding/internal/BindingContext.kt
szymonsztuka: Indeed, fixed (removed) in 2eb0fc8a1c36684f82905380d091a53c3a56ba9c

build.gradle
szymonsztuka: Fixed in 2eb0fc8a1c36684f82905380d091a53c3a56ba9c

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/binding/FacadeInterfaceBinding.kt
chrisr3: The "issue" is that Kotlin defines a "primary constructor" which can be identified from the `KClass` object, whereas Java _doesn't_. Setting properties via the constructor also means that the class doesn't need any "setter methods" and so allows instances to be "immutable", which is no bad thing either.

szymonsztuka: I've created new ticket for the issue with Java and added TODO in the code linking to this ticket - see commit 8bf710e533a3ce3d900b5d3c2c5921976053813e

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/binding/creation/FacadeInterfaceBindings.kt
```diff
@@ -21,6 +22,7 @@ object FacadeInterfaceBindings {
      * @return A [FacadeInterfaceBinding] which maps methods in the bound interface to methods in the facade.
      */
     @JvmStatic
+    @Suspendable
```
chrisr3: This class belongs to package `net.corda.flow.application.services.impl.interop.binding.creation`, i.e. a sub-package of `net.corda.flow.application.services.impl`. So how is this `static` function expected to be a "public entry-point"? Who is your "public"? CorDapp developers are only allowed access to `net.corda.v5.*` API packages, so they're _definitely_ excluded.

In fact, I doubt this function will be visible outside of the `corda-flow-service` bundle, just like all its other `*.impl.*` classes.

Any why `@JvmStatic` anyway? This implies you expect this function to be invoked from Java, which is a bit odd for a private function inside a Kotlin bundle :thinking:.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/proxies/FacadeClientProxy.kt
```diff
@@ -31,6 +32,7 @@ object FacadeProxies {
      */
     @Suppress("UNCHECKED_CAST")
     @JvmStatic
```
chrisr3: You cannot expose a `static` function via an OSGi service, so whose Java code are you expecting to invoke this? CorDapp developers are only allows to use `net.corda.v5.*` API packages.

chrisr3 (commented):
I don't understand why you've declared some private implementation functions to be `static`.

And I assure you that these _are_ private implementation functions because not only will OSGi not export their packages from the `corda-flow-services` bundle, but our own sandbox code will prevent CorDapps from touching them with a 100km pole.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/binding/creation/FacadeInterfaceBindings.kt
szymonsztuka: For the last question, in the original code (when this was a standalone project) `FacadeInterfaceBindings` was one of the entry point for client code, yes now when moving into Corda/Flow worker this is no longer the case (a bit off-topic,  we may need to move it to Corda API as lower level API for Cordapp developers).
I will remove `@JvmStatic`.
If this is not an API method, then it doesn't need to be `@Suspendable` as well, does it?

chrisr3: > If this is not an API method, then it doesn't need to be @Suspendable as well, does it?

Probably not. I certainly don't see how it could exist on a call-stack that includes CorDapp flow code.

yarivus (commented):
checkpointing review

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/binding/creation/FacadeInterfaceBindings.kt
szymonsztuka: I've left `@Static` annotation for now but added TODOs in two places, it was originally intended to be public entry point for developers, but as you pointed out is no the case now. We will revisit later if toe make it proper public or drop @Static, see commit https://github.com/corda/corda-runtime-os/pull/3945/commits/af43adb99655e934a434741c69e4743b5fcd82d8,
I've removed `@Suspendable`

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/binding/creation/FacadeInterfaceBindings.kt
```diff
@@ -21,8 +20,10 @@ object FacadeInterfaceBindings {
      * @param boundInterface The [Class] to bind the facade to.
      * @return A [FacadeInterfaceBinding] which maps methods in the bound interface to methods in the facade.
      */
+    // TODO originally the method (and the class) were envisioned as public API and it could be used in Java code
+    //  (hence @JvmStatic annotation), however effectively it's not Corda APi (it's hidden from a Cordapp code),
```
yarivus:
```suggestion
    //  (hence @JvmStatic annotation), however effectively it's not Corda API (it's hidden from a Cordapp code),
```

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/proxies/FacadeClientProxy.kt
```diff
@@ -31,8 +31,10 @@ object FacadeProxies {
      * @param jsonMarshaller A [JsonMarshaller] to use when reading/writing JSON blobs
      */
     @Suppress("UNCHECKED_CAST")
+    // TODO originally the method was envisioned as public API and could be used in Java code (hence @JvmStatic annotation),
+    //  however effectively it's not Corda APi (it's hidden from a Cordapp code),
```
yarivus:
```suggestion
    //  however effectively it's not Corda API (it's hidden from a Cordapp code),
```

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/binding/creation/FacadeInterfaceBindings.kt
szymonsztuka: I've decided to drop this whole lengthy TODO and remove @Static annotation for now, as it not used as Chris pointed out. I've created a tracking ticket that if we decide to make these methods public for Cordapp developers (and they may be used from within Java Cordapps) then we need to add back those annotation

yarivus (approved):
LGTM

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/interop/binding/creation/FacadeInterfaceBindings.kt
szymonsztuka: I've removed all `@Suspendable` annotation added originally in this PR, after testing there is no impact of them.

yarivus (approved):
LGTM

PR #3947, opened 2023-05-23
CORE-14072: Logging - Change date in filename to sortable Y-M-D

josephzunigadaly:
Change date format from `MM-dd-yyyy` to `yyyy-MM-dd` so filenames sort correctly.


owenstanford (approved):

mnesbit (approved):
OK, though I will point out this isn't the config used when in containers

mnesbit (approved):
OK, though I will point out this isn't the config used when in containers

mnesbit (approved):
OK, though I will point out this isn't the config used when in containers

PR #3948, opened 2023-05-23
CORE-13872: RPCSender - Manual Partition Assignment

jujoramos:
Instead of subscribing to the responder topic (requires as many
partitions as subscribers), manually assign a random partition in the
RPC message pattern to reduce the total number of required partitions
to bootstrap a Corda cluster.


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaRPCSenderImpl.kt
```diff
@@ -95,19 +90,32 @@ internal class CordaRPCSenderImpl<REQUEST : Any, RESPONSE : Any>(
...
                 ).use {
-                    it.subscribe(
-                        listOf(getRPCResponseTopic(config.topic)),
-                        partitionListener
+                    // TODO: reload assignment based on configurable interval to handle operators dynamically adding partitions.
+                    val availablePartitions = it.getPartitions(responseTopic)
+                    // Randomly assign 1/3rd of the partitions to this consumer
+                    responsePartitions = availablePartitions
+                        .shuffled()
+                        .take(ceil(availablePartitions.size / 3.0).toInt())
```
rick-r3: This 1/3rd of partitions idea is interesting.  Is this preferable to 1 partition ?

In this model, the consumer will be receiving 1/3rd of the topic traffic, so there is a scalability limit.  At some point the combined traffic could overwhelm the ability of the consumer to receiver the data and filter it looking for appropriate responses.  And by consumer I might mean the infrastructure the consumer is running on.  e.g. network and CPU limits.

If it was one partition, then that scalability limit is removed.  However, an individual consumer is then limited to the throughput of a single partition - which hopefully isn't a problem because we already are proposing to have more senders than partitions, but the scalability of the cluster can be increased by adding workers.  I think this is preferable, as it maintains our horizontal scaling promise, but I'm interested in the thoughts of others.

Was part of the thinking to reduce the risk of one partition being randomly chosen for 100% of the traffic from multiple workers, which is a form of the Birthday Paradox and actually ends up pretty likely ?  e.g. with 10 partitions and 10 senders, we'll presumably end up with some partitions with multiple senders and some with zero.  But I think that is a better problem to have than a horizontal scalability limit.  And indeed this is expected when we have more senders than partitions, and the chances that some have zero is then also diminished.

@JamesHR3 @LWogan

rick-r3 (commented):
I'm aware this is a WIP :-) 

Largely looks like what I think James described.  I commented on the partition selection strategy as I think that is interesting.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaRPCSenderImpl.kt
jujoramos: I was out today but yes, I already have some local changes with the consumer randomly assigning itself one single partition. This PR was a quick first attempt to run all the tests suites 👍.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaRPCSenderImpl.kt
```diff
@@ -85,29 +79,41 @@ internal class CordaRPCSenderImpl<REQUEST : Any, RESPONSE : Any>(
...
                     String::class.java,
                     RPCResponse::class.java
                 ).use {
-                    it.subscribe(
-                        listOf(getRPCResponseTopic(config.topic)),
-                        partitionListener
-                    )
-                    // Note that Lifecycle UP and DOWN are handled by the RPCConsumerRebalanceListener based on whether
-                    // partitions are available to this sender or not.
+                    // TODO: we might want to reload the assignment based on a configurable interval to
```
rick-r3: File a JIRA to follow up on this?  Not necessarily for 5.0.   I imagine we might well have other problems elsewhere in the codebase if the partitions are dynamically adjusted besides this, and even then I think this carries on working, just not using all partitions.

rick-r3 (commented):
LGTM

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaRPCSenderImpl.kt
jujoramos: It won't stop working, it just won't use all the available partitions as you stated, yeah.
Instead of creating a ticket (it's probably going to be an epic so all the message patterns are updated), though, I'll remove the `TODO`: if the operator increases the amount of partitions after initialisation the sender should continue to work just fine as it already has the partition assigned. If the operator decreases the amount of partitions, I'd expect the next `poll` to throw a fatal exception (we can't read from a non existing partition), in which case the sender should recreate itself using the current available partitions, so it should continue to work just fine anyways 👍 .

fowlerrr (approved):
LGTM.  I do wonder if we'll start favouring `assignment` over subscriptions when we can.

rick-r3 (approved):

PR #3953, opened 2023-05-24
CORE-14127: Set explicit file permissions for CPK disk cache.

chrisr3:
Set file permissions for the CPKs that we cache on disk to be `rw-------`.
Also set the permissions for this cache directory to be `rwx------`.
Set workspace directory permissions to `rwxr-xr-x`.
Set temporary directory permissions to `rwx------`.

It is also probably a Very Good Idea :tm: to close a `ByteChannel` once we've finished writing to it.
And convert some Kotlin classes which only contain `companion object` into a Kotlin `object`.


mnesbit:
Worth just sanity checking on windows? Looks sensible though.

mnesbit (approved):
I chatted to @driessamyn and we agree that one day we may want to share cached binaries, but we don't use that right now and I think the security issues here are significant and need fixing.

PR #3954, opened 2023-05-24
CORE-13908 Put back signing keys cache as per CORE-9550

kyriathar:


components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
```diff
@@ -359,21 +382,28 @@ class SigningServiceImpl(
...
+
+    @VisibleForTesting
+    internal fun lookupByFullKeyId(tenantId: String, requestedFullKeyId: SecureHash): SigningKeyInfo? {
+        val keyId = ShortHash.of(requestedFullKeyId)
+        return cache.get(CacheKey(tenantId, keyId)) {
+            val repository = signingRepositoryFactory.getInstance(tenantId)
+            repository.findKeyByFullId(requestedFullKeyId)
+        }?.let {
+            // This is to make sure cached key by short id (db one looks with full id so should be OK) is the actual
+            // requested key and not a different one that clashed on key id (short key id).
```
rick-r3, line 404: But if the `fullId` from the cache does not match, you don't fall back to the database?  i.e. the caller above is going to throw an exception.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
```diff
@@ -336,16 +357,18 @@ class SigningServiceImpl(
...
         if (publicKey is CompositeKey) {
             val leafKeysIdsChunks = publicKey.leafKeys.map {
                 it.fullIdHash(schemeMetadata, digestService) to it
             }.chunked(KEY_LOOKUP_INPUT_ITEMS_LIMIT)
             for (chunk in leafKeysIdsChunks) {
-                val found = signingRepositoryFactory.getInstance(tenantId).lookupByPublicKeyHashes(
-                    chunk.map { it.first }.toMutableSet()
-                )
+                val found = signingRepositoryFactory.getInstance(tenantId)
+                        .lookupByPublicKeyHashes(
```
rick-r3: Does this method `lookupByPublicKeyHashes` also cache ?  My impression is not given the methods above wrap it with caching.   Shouldn't this also go via a cache ?

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/SigningServiceCacheTests.kt
```diff
@@ -0,0 +1,270 @@
...
+                mock(),
+                signingRepositoryFactory,
+                cipherSchemeMetadata,
+                digestService,
+                cache
+            )
+    }
+
+    @Test
+    fun `lookupByKeyIds returns requested keys from cache if all requested keys are in cache`() {
```
dickon, line 61: Way too much repetition here; I'd sorted all this out so the test was maintainable and you've made it much longer again

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
kyriathar: Good catch. Changed it to use `SigningService.lookupSigningKeysByPublicKeyHashes` instead which looks up in the cache first.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
kyriathar: The cache is by short key id only. Once in this block it means that if found on cache we make sure it actually matches the found `fullId`, otherwise it is a clash on short key id so different key so return not found(null). if it was found in the DB because it looks by full key id means no way clashing on short key id, so that's there to take care of found on cache branch.

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/SigningServiceCacheTests.kt
kyriathar: All these tests exercise the different paths, going through cache and if not cached going through DB

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
rick-r3: But my point is you are not going to the DB when the cache entry for the short ID is found, but then you have the clash.  Shouldn't you get -> compare -> db lookup ?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/SigningService.kt
```diff
@@ -65,18 +65,18 @@ interface SigningService {
...
-        keyIds: List<ShortHash>,
+        keyIds: Set<ShortHash>,
     ): Collection<SigningKeyInfo>
 
     /**
      * Looks for keys by full key ids.
      *
      * @param tenantId The tenant's id which the keys belong to.
-     * @param fullKeyIds Key ids to look keys for.
+     * @param requestedFullKeyIds Key ids to look keys for.
```
dickon, line 75: The word "requested" is unnecessary; it is part of the parameters of the request through to the signing service.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
```diff
@@ -87,15 +88,15 @@ class SigningServiceImpl(
...
+    data class OwnedKeyRecord(val publicKey: PublicKey, val data: SigningKeyInfo)
 
     override fun getSupportedSchemes(tenantId: String, category: String): List<String> {
         logger.debug { "getSupportedSchemes(tenant=$tenantId, category=$category)" }
         val ref = cryptoServiceFactory.findInstance(tenantId = tenantId, category = category)
         return ref.instance.supportedSchemes.map { it.key.codeName }
     }
 
-
+    // This is admin operation - not to worry about performance for now
```
dickon, line 99: Are you sure?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
kyriathar: Oh okay, I think the answer to this is that we can't have clashes for the same tenant. Means if we have found a <tenantId, shortKeyId> pair it has to be either that or the key belongs to someone else.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
kyriathar: `querySigningKeys` is getting called only from `CryptoOpsBusProcessor` so in any case it doesn't seem to be involved in flow-crypto interactions.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
```diff
@@ -118,49 +119,68 @@ class SigningServiceImpl(
...
-            it.lookupByPublicKeyShortHashes(notFound.toMutableSet())
-        }
-        fetchedKeys.forEach { cache.put(CacheKey(tenantId, it.id), it) }
+        return if (cachedKeys.size == keyIds.size) {
+            cachedKeys
+        } else {
+            val notFound = keyIds - cachedKeys.mapTo(mutableSetOf()) { it.id }
+            val fetchedKeys =
+                signingRepositoryFactory.getInstance(tenantId).use {
+                    it.lookupByPublicKeyShortHashes(notFound)
```
dickon, line 137: You've reformatted the code, making it longer and more nested and used slgithly different types. Doesn't really matter, just means I had to spend a few minutes  figuring out whether there's any semantic change here.  

Both versions are using sets for the computation.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
```diff
@@ -118,49 +119,68 @@ class SigningServiceImpl(
...
-                .map { foundKey ->
-                    foundKey.also {
-                        cache.put(CacheKey(tenantId, it.id), it)
-                    }
+        return if (cachedKeysByFullId.size == requestedFullKeyIds.size) {
+            cachedKeysByFullId
+        } else {
+            val notFound =
+                requestedFullKeyIds - cachedKeysByFullId.mapTo(mutableSetOf()) { it.fullId }
+            // We look for keys in DB by their full key ids so not risking a clash here
```
dickon, line 174: again, I don't believe there's any useful functional change here, your version is just different and longer.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
dickon: There's no SigningServiceImpl unit test coverage on this branch with composite keys yet, which is what I wanted @kyriathar to actually address. I'm not sure if we use this branch or not.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
rick-r3: Ah, okay, this defends against someone asking you to sign with a key that hash collides with one of yours, and you end up confusing it for they key you do have with the same hash, rather than saying you don't have that public (and private) key.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
kyriathar: Added test asserting cache and database hits for composite key lookup (hits for composite key leaves)

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/SigningServiceGeneralTests.kt
```diff
@@ -451,44 +450,4 @@ class SigningServiceGeneralTests {
         assertThat(fullIdsCap.allValues.single()).isEqualTo(setOf(hashB))
     }
 
-    @Test
```
dickon: Why delete this test?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
dickon: I'm not sure yet that flows are the only frequent interaction in the test cases being studied.

dickon (changes requested):
Sorry @kyriathar, we need the production change to be as small as possible at this stage in 5.0, and that means we really should not go from List to Set. Only performance fixes and blocking bugs can go in.

I will push on with https://github.com/corda/corda-runtime-os/pull/3938  and try to implement all Rick's feedback.

Kyriakos and I worked carefully through this and 3938 which are 2 different versions of the same effort and I think we've agreed there's no important behavioural differences.

Maybe this could go in 5.1. However, given that Avro has no set type, and this code is only used in production based on Avro messages and has to respond in Avro, I we'll probably decide that the List type is the best choice.

dickon:
My attempt to progerss this before I realised it was just too nuch unnecessary change  for 5.0 is at https://github.com/corda/corda-runtime-os/tree/dickon/CORE-13908-with-sets

PR #3955, opened 2023-05-25
CORE-13740: Add serial and allow conflicts in suspension and activation

yift-r3:
See the e2e pull request in https://github.com/corda/corda-e2e-tests/pull/109

### DIFF
--- a/testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MGMUtils.kt
+++ b/testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MGMUtils.kt
@@ -309,7 +309,8 @@ private fun ClusterInfo.createMgmRegistrationContext(
  */
 fun ClusterInfo.suspendMember(
     mgmHoldingId: String,
-    x500Name: String
+    x500Name: String,
+    serialNumber: Int? = null,
 ) = cluster {
     assertWithRetry {
         timeout(15.seconds)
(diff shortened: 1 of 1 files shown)


testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MGMUtils.kt
```diff
@@ -309,18 +309,19 @@ private fun ClusterInfo.createMgmRegistrationContext(
...
+    serialNumber: Int? = null,
 ) = cluster {
     assertWithRetry {
         timeout(15.seconds)
         interval(1.seconds)
         command {
             post(
                 "/api/v1/mgm/$mgmHoldingId/suspend",
-                "{ \"x500Name\": \"$x500Name\" }"
+                "{ \"x500Name\": \"$x500Name\", \"serialNumber\": $serialNumber }"
```
dimosr, line 321: Does the REST endpoint work fine with if `serialNumber` is null? If so, should we omit specifying it at all in cases it's null?

 I remember there was an issue with blank strings.

dimosr (commented):
LGTM - just a minor comment

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MGMUtils.kt
williamvigorr3: Hmm I don't think `"{ \"x500Name\": \"$x500Name\", \"serialNumber\": null }"` is valid json, so we should probably fix this.

williamvigorr3 (commented):
Everything apart from the open comment looks fine.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MGMUtils.kt
yift-r3: The rest endpoint work with null as if the field is not there.

(The issue with blank string is that it behaves as null)

yift-r3: (Am I missing something? It looks like perfectly valid JSON to me)

williamvigorr3: Yeah I think you are correct (having read the json spec).

dimosr (approved):

PR #3956, opened 2023-05-25
CORE-14142: Call flow started when initiated by session event

JamesHR3:
We've missed a call to `flowStarted` for initiated flows, which means that timing metrics are broken for these. Adding this in here should fix this up.


JamesHR3:
Draft while confirming the fix.

rick-r3 (dismissed):

mnesbit (approved):
OK

mnesbit (approved):
OK

PR #3959, opened 2023-05-25
CORE-13837: Add disk-usage metrics for CPK caches.

chrisr3:
Create a `DiskSpace` metric, and use it to monitor the available disk space in the CPK and chunk cache directories.

Note that `DiskSpace` is determined via `File::getUsableSpace` and `File::getTotalSpace`, which operate on the disk partition that `File` object refers to. This means that:
- We need an actual `File` instance to create one of these metrics.
- A server may be created to have _many_ disk partitions, and there is no reason to assume that two different paths share the same partition.
- Conversely, a single disk partition may contain several directories that we may wish to monitor, and so it wouldn't make sense to have metrics for each of them. Most likely, we'd only need to monitor their common parent directory.

We should therefore _carefully_ consider which directories we need to monitor. We certainly don't need to monitor _every_ directory.

### DIFF
--- a/components/virtual-node/cpk-read-service-impl/src/main/kotlin/net/corda/cpk/read/impl/CpkReadServiceImpl.kt
+++ b/components/virtual-node/cpk-read-service-impl/src/main/kotlin/net/corda/cpk/read/impl/CpkReadServiceImpl.kt
@@ -21,6 +21,8 @@ import net.corda.lifecycle.StopEvent
 import net.corda.lifecycle.createCoordinator
 import net.corda.messaging.api.subscription.config.SubscriptionConfig
 import net.corda.messaging.api.subscription.factory.SubscriptionFactory
+import net.corda.metrics.CordaMetrics
+import net.corda.metrics.CordaMetrics.Tag.ContentsType
 import net.corda.schema.Schemas
 import net.corda.schema.configuration.ConfigKeys.BOOT_CONFIG
 import net.corda.schema.configuration.ConfigKeys.MESSAGING_CONFIG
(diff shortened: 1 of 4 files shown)


JamesHR3:
The code looks sane, but I'm wondering what this metric looks like in terms of how it would display on a dashboard?

chrisr3:
> The code looks sane, but I'm wondering what this metric looks like in terms of how it would display on a dashboard?

This uses the built-in `DiskSpaceMetrics` which consists of two `Guage`s, and also adds a `ContentsType` tag to describe what the disk space is being used for:
```java
Iterable<Tag> tagsWithPath = Tags.concat(tags, "path", absolutePath);
Gauge.builder("disk.free", path, File::getUsableSpace)
    .tags(tagsWithPath)
    .description("Usable space for path")
    .baseUnit(BaseUnits.BYTES)
    .strongReference(true)
    .register(registry);
Gauge.builder("disk.total", path, File::getTotalSpace)
    .tags(tagsWithPath)
    .description("Total space for path")
    .baseUnit(BaseUnits.BYTES)
    .strongReference(true)
    .register(registry);
```
I don't understand your concerns about "how it would display". This is just the raw metric data - display it however you like :shrug:.

components/virtual-node/cpk-read-service-impl/src/main/kotlin/net/corda/cpk/read/impl/CpkReadServiceImpl.kt
```diff
@@ -147,12 +149,20 @@ class CpkReadServiceImpl (
             return
         }
 
+        // Monitor the amount of available disk space in these two directories.
+        CordaMetrics.Metric.DiskSpace(cpkCacheDir).builder()
+            .withTag(ContentsType, "CPKs")
```
mbrkic-r3, line 154: Maybe more descriptive names "CPK cache partition" and "CPK parts partition"?

libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
```diff
@@ -456,6 +460,13 @@ object CordaMetrics {
              */
             object DeserializationTime : Metric<Timer>("serialization.amqp.deserialization.time", CordaMetrics::timer)
         }
+
+        class DiskSpace(path: Path) : Metric<NoopMeter>("", meter = { _, tags ->
+            if (path.fileSystem == FileSystems.getDefault()) {
```
mbrkic-r3, line 465: Why do we check this?

chrisr3: Because this is the _only_ filesystem that supports `Path.toFile()`. Unit tests that use JimFS would fail without this check.

components/virtual-node/cpk-read-service-impl/src/main/kotlin/net/corda/cpk/read/impl/CpkReadServiceImpl.kt
chrisr3: I think that's overkill. There's no reason to assume that either of these directories has its own dedicated partition.

JamesHR3 (approved):

wzur-r3 (approved):
LGTM for Gradle files changes

driessamyn (approved):
LGTM

PR #3960, opened 2023-05-25
CORE-14065: adjust poll frequency for idle topic consumers

Tom-Fitzpatrick:
Write up for this investigation: https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4496851084/Experiment+with+increased+consumer+poll+timeout+to+reduce+poll+frequency+of+idle+topics


libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
```diff
@@ -27,6 +27,9 @@ consumer = ${common} {
     session.timeout.ms = 6000
     # Frequency of heartbeats between the consumer and the broker. Should be no higher than 1/3 of the session timeout.
     heartbeat.interval.ms = 2000
+
+    fetch.max.wait.ms = 1000
+    fetch.min.bytes = 16000
```
rick-r3: I think we will want to leave this min.bytes as 1 so consumers return immediately if there is data, but appreciate you might be experimenting to see what effects are.

Tom-Fitzpatrick: Ah thanks, I was curious to see the effects, but it seems to have broken the e2e tests anyway, so will focus on the max.wait.ms now.

rick-r3: The other thing to be aware of is that `fetch.max.wait.ms = 1000` puts you very close to `session.timeout.ms = 6000` and `heartbeat.interval.ms = 2000` - or at least closer, which might make things less stable in your tests.   Increasing those other settings temporarily (maybe add another '0' on the end of each) might help with stability issues if you see them.  

Note @fowlerrr is looking at increasing these other settings anyway.... https://github.com/corda/corda-runtime-os/pull/3964

libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
```diff
@@ -27,6 +27,8 @@ consumer = ${common} {
     session.timeout.ms = 6000
     # Frequency of heartbeats between the consumer and the broker. Should be no higher than 1/3 of the session timeout.
     heartbeat.interval.ms = 2000
+
```
vkolomeyko: Probably needs a comment explaining what this parameter controls.

Tom-Fitzpatrick: Thanks, I added a comment

rick-r3 (approved):

PR #3961, opened 2023-05-25
CORE-13987: Add AMQP serialization metrics to all sandboxes, and simplify CurrentSandboxGroupContext.

chrisr3:
`CurrentSandboxGroupContext` is a singleton platform component, and so does not need to declare any of the sandbox marker interfaces.

`FlowSandboxServiceImpl` needs this component to advertise `SingletonSerializeAsToken` so that it can be registered with the Checkpoint serializer as a "non-injectable service".

Wrap `SerializationService` inside `SerializationMetricsWrapper` to ensure that serialization metrics are available in all cases.

### DIFF
--- a/components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/SerializationMetricsWrapper.kt
+++ b/components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/SerializationMetricsWrapper.kt
@@ -1,0 +1,44 @@
+package net.corda.sandbox.serialization.amqp
+
+import io.micrometer.core.instrument.Timer
+import net.corda.metrics.CordaMetrics
+import net.corda.metrics.recordOptionally
+import net.corda.sandboxgroupcontext.SandboxGroupContext
+import net.corda.v5.application.serialization.SerializationService
+import net.corda.v5.serialization.SerializedBytes
+
+class SerializationMetricsWrapper(
+    private val serializationService: SerializationService,
+    sandboxGroupContext: SandboxGroupContext
+) : SerializationService {
+    private val holdingId = sandboxGroupContext.virtualNodeContext.holdingIdentity.shortHash.toString()
+
+    override fun <T : Any> serialize(obj: T): SerializedBytes<T> {
+        return CordaMetrics.Metric.Serialization.SerializationTime
+            .builder()
+            .forVirtualNode(holdingId)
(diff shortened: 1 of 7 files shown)


lankydan (dismissed):

components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/SerializationMetricsWrapper.kt
```diff
@@ -0,0 +1,44 @@
+package net.corda.sandbox.serialization.amqp
+
+import io.micrometer.core.instrument.Timer
+import net.corda.metrics.CordaMetrics
+import net.corda.metrics.recordOptionally
+import net.corda.sandboxgroupcontext.SandboxGroupContext
+import net.corda.v5.application.serialization.SerializationService
+import net.corda.v5.serialization.SerializedBytes
+
+class SerializationMetricsWrapper(
```
lankydan, line 10: Does this change any behaviour? The code I wrote should have already been including the serialization metrics in the db and verification sandboxes. The flow sandbox serialization service had its metrics added to its implementation of `SerializationService`. Are we going to make the same change there?

chrisr3: By adding the metrics in `SerializationServiceProxy`, we left the instance inside the `SandboxGroupContext` unaffected. So a `PERSISTENCE` sandbox would have no AMQP metrics if it obtained its `SerializationService` by invoking:
```kotlin
fun SandboxGroupContext.getSerializationService(): SerializationService =
    getObjectByKey(RequireSandboxAMQP.AMQP_SERIALIZATION_SERVICE)
        ?: throw CordaRuntimeException(
            "Entity serialization service not found within the sandbox for identity: " +
                    "${virtualNodeContext.holdingIdentity}"
        )
```
This `SandboxSerializationProxy` is only for `PERSISTENCE` and `VERIFICATION` sandbox services that ask for a `SerializationService` by `@Reference`.

I don't think it's a good idea for AMQP serialization metrics to be added by multiple different "Sources of Truth".

lankydan: I see, that is good to know for future reference.

lankydan (dismissed):

lankydan (approved):

driessamyn (approved):
LGTM

PR #3963, opened 2023-05-26
CORE-13858 - Record inbound message rate in the loopback case

dimosr:
Record inbound message rate in the loopback case


dimosr:
Closing this since it was merged in 5.0 (and will be merged forward in time).

PR #3964, opened 2023-05-26
CORE-13866: Increase Kakfa timeouts

fowlerrr:
The assumption is that the current timeouts are too slow and are leading to increased rebalancing.  This has been through limited testing using the performance tests.

More testing should help stabilise these numbers and it's likely that as other areas improve we'll be able to refine them further.

### DIFF
--- a/libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
+++ b/libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
@@ -21,12 +21,14 @@ consumer = ${common} {
     # batch may contain fewer than 500 records.
     max.poll.records = 500
     # Time to allow between polls on a consumer before a rebalance occurs that removes this consumer's partitions.
-    max.poll.interval.ms = 100000
+    max.poll.interval.ms = 300000
     # Timeout of heartbeats between the consumer and the broker. If no heartbeat is received in this timeframe, a
     # rebalance will occur.
-    session.timeout.ms = 6000
+    session.timeout.ms = 60000
     # Frequency of heartbeats between the consumer and the broker. Should be no higher than 1/3 of the session timeout.
-    heartbeat.interval.ms = 2000
+    heartbeat.interval.ms = 20000
+    # How long to wait to allow consumers to join the group before performing a rebalance
+    group.initial.rebalance.delay.ms = 60000
 }
 
 # Defaults for all producers.
--- a/libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
+++ b/libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
@@ -27,8 +27,6 @@ consumer = ${common} {
     session.timeout.ms = 60000
     # Frequency of heartbeats between the consumer and the broker. Should be no higher than 1/3 of the session timeout.
     heartbeat.interval.ms = 20000
-    # How long to wait to allow consumers to join the group before performing a rebalance
-    group.initial.rebalance.delay.ms = 60000
 }
 
 # Defaults for all producers.
--- a/libs/messaging/kafka-message-bus-impl/src/test/kotlin/net/corda/messagebus/kafka/config/MessageBusConfigResolverTest.kt
+++ b/libs/messaging/kafka-message-bus-impl/src/test/kotlin/net/corda/messagebus/kafka/config/MessageBusConfigResolverTest.kt
@@ -152,7 +152,7 @@ class MessageBusConfigResolverTest {
                 CLIENT_ID_PROP to "consumer-$CLIENT_ID",
                 ISOLATION_LEVEL_PROP to "read_committed",
                 BOOTSTRAP_SERVERS_PROP to "localhost:9092",
-                SESSION_TIMEOUT_PROP to "6000",
+                SESSION_TIMEOUT_PROP to "60000",
                 AUTO_OFFSET_RESET_PROP to "earliest"
             )
             val properties = Properties()


driessamyn (dismissed):
LGTM but would like some input but would like somebody else to sanity check.
@JamesHR3 ? @Omar-awad ? @rick-r3 ?

libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
```diff
@@ -21,12 +21,14 @@ consumer = ${common} {
...
     # Time to allow between polls on a consumer before a rebalance occurs that removes this consumer's partitions.
-    max.poll.interval.ms = 100000
+    max.poll.interval.ms = 300000
     # Timeout of heartbeats between the consumer and the broker. If no heartbeat is received in this timeframe, a
     # rebalance will occur.
-    session.timeout.ms = 6000
+    session.timeout.ms = 60000
     # Frequency of heartbeats between the consumer and the broker. Should be no higher than 1/3 of the session timeout.
-    heartbeat.interval.ms = 2000
+    heartbeat.interval.ms = 20000
```
jujoramos, line 29: From the definitive Kafka 2nd Edition:
> session.timeout.ms and heartbeat.interval.ms
The amount of time a consumer can be out of contact with the brokers while still considered alive defaults to 10 seconds. If more than session.timeout.ms passes without the consumer sending a heartbeat to the group coordinator, it is considered dead and the group coordinator will trigger a rebalance of the consumer group to allocate partitions from the dead consumer to the other consumers in the group. This property is closely related to heartbeat.interval.ms, which controls how frequently the Kafka consumer will send a heartbeat to the group coordinator, whereas ses⁠sion.timeout.ms controls how long a consumer can go without sending a heartbeat. Therefore, those two properties are typically modified together—heartbeat.​interval.ms must be lower than session.timeout.ms and is usually set to one-third of the timeout value. So if session.timeout.ms is 3 seconds, heartbeat.​inter⁠val.ms should be 1 second. Setting session.timeout.ms lower than the default will allow consumer groups to detect and recover from failure sooner but may also cause unwanted rebalances. Setting session.timeout.ms higher will reduce the chance of accidental rebalance but also means it will take longer to detect a real failure.

Any reason for not using the 1/3 rule? (just out of curiosity).

libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
```diff
@@ -21,12 +21,14 @@ consumer = ${common} {
     # batch may contain fewer than 500 records.
     max.poll.records = 500
     # Time to allow between polls on a consumer before a rebalance occurs that removes this consumer's partitions.
-    max.poll.interval.ms = 100000
+    max.poll.interval.ms = 300000
```
jujoramos, line 24: Isn't 5 minutes too long for the system to detect that a consumer is unresponsive? 😓

libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
```diff
@@ -21,12 +21,14 @@ consumer = ${common} {
...
     # Timeout of heartbeats between the consumer and the broker. If no heartbeat is received in this timeframe, a
     # rebalance will occur.
-    session.timeout.ms = 6000
+    session.timeout.ms = 60000
     # Frequency of heartbeats between the consumer and the broker. Should be no higher than 1/3 of the session timeout.
-    heartbeat.interval.ms = 2000
+    heartbeat.interval.ms = 20000
+    # How long to wait to allow consumers to join the group before performing a rebalance
+    group.initial.rebalance.delay.ms = 60000
```
jujoramos: This change not only implies that the initial rebalance is delayed during the initial `Corda` deployment (good ✅), but also that Kafka will wait 1 minute to execute a rebalance after a consumer joins an already formed group, meaning that we are consciously introducing at least an extra 1 minute latency during failover (bad 🔴 ?). Reducing the `max.poll.interval.ms` to 30 seconds (instead of 5 minutes) might help to mitigate this as the rebalance waiting time for joining members is capped by `max.poll.interval.ms` (if I'm reading [this](https://kafka.apache.org/documentation/#upgrade_1100_notable) correctly, that is).
I'm pretty sure the `resiliency-tests` can not cope with this at the moment, they declare a failure whenever any flow or member registration takes longer than 1 minute.

jujoramos: Another thing I've just noticed while searching through the docs, the `group.initial.rebalance.delay.ms` seems to be a broker property, not something you can configure on the client side.

libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
fowlerrr: Aren't I following the 1/3 rule?  I meant to.  As is `60000` * `1/3` == `20000`?

libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
fowlerrr: I didn't realise that second bit.   And 5 minutes for the poll was certainly overeager.  I do expect we'll have to tweak these once we have more data and that was really just a shot in the dark.

I'll pull the poll back to 30 seconds as you suggest for now.

libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
jujoramos: Yes you are, I'm certainly not good at maths on Fridays 🤣 .

libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
driessamyn: `group.initial.rebalance.delay.ms` is indeed a broker setting (and we should set it), but it does _not_ impact rebalancing when somebody joins the group, only when a new group is created:
https://docs.confluent.io/platform/current/installation/configuration/broker-configs.html#group-initial-rebalance-delay-ms
> The amount of time the group coordinator will wait for more consumers to join a new group before performing the first rebalance. A longer delay means potentially fewer rebalances, but increases the time until processing begins.

So I think it _is_ safe for us to set to a fairly long time (but yes, it's in the wrong place here).

jujoramos: You're right, I was reading [old docs](https://kafka.apache.org/documentation/#upgrade_1100_notable) from when the configuration property was added to Kafka:
> A new config, group.initial.rebalance.delay.ms, was introduced. This config specifies the time, in milliseconds, that the GroupCoordinator will delay the initial consumer rebalance. The rebalance will be further delayed by the value of group.initial.rebalance.delay.ms as new members join the group, up to a maximum of max.poll.interval.ms. The default value for this is 3 seconds. During development and testing it might be desirable to set this to 0 in order to not delay test execution time.

rick-r3:
Can we try and get metrics showing us the rebalances and when they happen relative (time-wise) to a "run" under load, before and after?  Put them on the JIRA.

I'd like to know the distribution (is it all at the beginning?), and how much improvement these changes bring.

fowlerrr:
> Can we try and get metrics showing us the rebalances and when they happen relative (time-wise) to a "run" under load, before and after? Put them on the JIRA.
> 
> I'd like to know the distribution (is it all at the beginning?), and how much improvement these changes bring.

I have some runs with these changes and with a baseline which is a `HEAD`.

Baseline:
image

With changes:
image


It's only a single run of each but it appears better.

fowlerrr:
I can pop the above in the JIRA but it's not really a lot of information.  I'll see if I can get more than a headline number from that dashboard

jujoramos:
@fowlerrr: `corda-rest-worker` rebalances decreased quite a lot after commit a189d56f9f5a14b30e6bd29a5192b26cad15c49f got merged. Can you make sure to rebase your changes against the latest `release/os/5.0` and run the tests again? (I'm making this comment based on the images you recently shared).

fowlerrr:
> I can pop the above in the JIRA but it's not really a lot of information. I'll see if I can get more than a headline number from that dashboard

Put some results in the JIRA.  Running a few other tests as well, but they'll be ancillary to this change

libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
jujoramos: Looks like 5 minutes is the default used by Kafka, so ignore my comment 👎

rick-r3 (approved):

PR #3966, opened 2023-05-26
CORE-13920 message pattern lib metrics

conal-smith:
I've now updated [flow dashboard](https://github.com/corda/engineering-helm/pull/50/files#diff-f4bd8bfd6097f7b6628b3835244f804d8173df626065098302e11e3ff6ebc576) following renaming of some "messaging" metrics to "consumer" (as they cover all consumers).

See grafana dashboard of flow worker with these updates [here](https://grafana.non-functional-tests.corda.cloud/d/Conal-FlowWorker/conal-flow-worker?orgId=1&from=1687544802980&to=1687545779676).

See the message pattern dashboard [here](https://grafana.non-functional-tests.corda.cloud/d/Conal-Messaging/conal-messaging-patterns?orgId=1&from=1687544779198&to=1687545767527)

See engineering-helm changes [here](https://github.com/corda/engineering-helm/pull/50).

### DIFF
--- a/libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
+++ b/libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
@@ -49,7 +49,7 @@ class CordaKafkaConsumerImpl<K : Any, V : Any>(
     private val consumer: Consumer<Any, Any>,
     private var defaultListener: CordaConsumerRebalanceListener? = null,
     private val chunkDeserializerService: ConsumerChunkDeserializerService<K, V>,
-    private val consumerMetricsBinder : MeterBinder,
+    private val consumerMetricsBinder: MeterBinder,
 ) : CordaConsumer<K, V> {
     private var currentAssignment = mutableSetOf<Int>()
     private val bufferedRecords = mutableMapOf<Int, List<ConsumerRecord<Any, Any>>>()
@@ -74,47 +74,68 @@ class CordaKafkaConsumerImpl<K : Any, V : Any>(
 
     @Suppress("TooGenericExceptionCaught")
     override fun poll(timeout: Duration): List<CordaConsumerRecord<K, V>> {
-        val polledRecords = try {
-            consumer.poll(timeout)
-        } catch (ex: Exception) {
-            when (ex) {
-                is AuthorizationException,
-                is AuthenticationException,
-                is IllegalArgumentException,
-                is IllegalStateException,
-                is ArithmeticException,
-                is FencedInstanceIdException,
-                is InconsistentGroupProtocolException,
-                is InvalidOffsetException,
-                -> {
-                    logErrorAndThrowFatalException("Error attempting to poll.", ex)
-                }
+        return recordConsumerPollTime {
+            val polledRecords = try {
+                consumer.poll(timeout)
+            } catch (ex: Exception) {
+                when (ex) {
+                    is AuthorizationException,
+                    is AuthenticationException,
+                    is IllegalArgumentException,
+                    is IllegalStateException,
+                    is ArithmeticException,
+                    is FencedInstanceIdException,
+                    is InconsistentGroupProtocolException,
+                    is InvalidOffsetException,
+                    -> {
+                        logErrorAndThrowFatalException("Error attempting to poll.", ex)
+                    }
 
-                is WakeupException,
-                is InterruptException,
-                is KafkaException,
-                -> {
-                    logWarningAndThrowIntermittentException("Error attempting to poll.", ex)
-                }
+                    is WakeupException,
+                    is InterruptException,
+                    is KafkaException,
+                    -> {
+                        logWarningAndThrowIntermittentException("Error attempting to poll.", ex)
+                    }
 
-                else -> logErrorAndThrowFatalException("Unexpected error attempting to poll.", ex)
-            }
-        }
-        clearBuffersForUnassignedPartitions()
-
-        val recordsToReturn = mutableListOf<CordaConsumerRecord<K, V>>()
-        polledRecords.groupBy { it.partition() }.forEach { (partition, records) ->
-            val bufferedRecords = bufferedRecords[partition] ?: emptyList()
-            if (bufferedRecords.isNotEmpty()) {
-                log.trace {
-                    "Taking  ${bufferedRecords.size} buffered records from partition $partition and adding them to the polled records" +
-                            " of size ${records.size}"
+                    else -> logErrorAndThrowFatalException("Unexpected error attempting to poll.", ex)
                 }
             }
-            recordsToReturn.addAll(parseRecords(partition, bufferedRecords.plus(records)))
-        }
 
-        return recordsToReturn.sortedBy { it.timestamp }
+            clearBuffersForUnassignedPartitions()
+
+            val recordsToReturn = mutableListOf<CordaConsumerRecord<K, V>>()
+            polledRecords.groupBy { it.partition() }.forEach { (partition, records) ->
+                recordPolledRecordsPerPartition(partition, records)
+                val bufferedRecords = bufferedRecords[partition] ?: emptyList()
+                if (bufferedRecords.isNotEmpty()) {
+                    log.trace {
+                        "Taking  ${bufferedRecords.size} buffered records from partition $partition and adding them to the polled records" +
+                                " of size ${records.size}"
+                    }
+                }
+                recordsToReturn.addAll(parseRecords(partition, bufferedRecords.plus(records)))
+            }
+
+            recordsToReturn.sortedBy { it.timestamp }
+        }
+    }
+
+    private fun <T : Any> recordConsumerPollTime(poll: () -> T): T {
+        return CordaMetrics.Metric.Messaging.ConsumerPollTime.builder()
+            .withTag(CordaMetrics.Tag.MessagePatternClientId, config.clientId)
+            .build()
+            .recordCallable {
+                poll.invoke()
+            }!!
+    }
+
+    private fun recordPolledRecordsPerPartition(partition: Int, records: List<ConsumerRecord<*, *>>) {
+        CordaMetrics.Metric.Messaging.ConsumerBatchSize.builder()
+            .withTag(CordaMetrics.Tag.MessagePatternClientId, config.clientId)
+            .withTag(CordaMetrics.Tag.Partition, "$partition")
+            .build()
+            .record(records.size.toDouble())
     }
 
     /**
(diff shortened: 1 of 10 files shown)


libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
```diff
@@ -114,7 +121,22 @@ class CordaKafkaConsumerImpl<K : Any, V : Any>(
             recordsToReturn.addAll(parseRecords(partition, bufferedRecords.plus(records)))
         }
 
-        return recordsToReturn.sortedBy { it.timestamp }
+        return recordsToReturn.sortedBy { it.timestamp }.also {
+            CordaMetrics.Metric.Messaging.ConsumerPollTime.builder()
```
conalsmith-r3, line 125: This is consumer poll time for all consumers, previously we had it only for S&E consumers. I haven't removed `MessagingPollTime`, but could do so if you're happy with that @JamesHR3?

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
```diff
@@ -114,7 +121,22 @@ class CordaKafkaConsumerImpl<K : Any, V : Any>(
...
+                .withTag(CordaMetrics.Tag.MessagePatternClientId, config.clientId)
+                .build()
+                .record(Duration.ofMillis(System.nanoTime() - startTime))
+        }
+    }
+
+    private fun recordPolledRecordsPerPartition(partition: Int, records: List<ConsumerRecord<*, *>>) {
+        recordsConsumedMetricCache.computeIfAbsent(partition) {
+            CordaMetrics.Metric.Messaging.ConsumerRecordsConsumedCount.builder()
```
sprintengine: Removed `messaging.batch.size` in place of this.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/CordaKafkaProducerImpl.kt
```diff
@@ -140,12 +140,28 @@ class CordaKafkaProducerImpl(
...
+            .mapValues { (_, records) -> records.size }
+            .forEach { (topic, count) ->
+                CordaMetrics.Metric.Messaging.ProducerChunksGenerated.builder()
+                    .withTag(CordaMetrics.Tag.MessagePatternClientId, config.clientId)
+                    .withTag(CordaMetrics.Tag.Topic, topic)
+                    .build()
+                    .record(count.toDouble())
+            }
+
+    }
```
sprintengine, line 202: Tested this locally with `SendReceiveAllMessagingFlow` and saw that it works, the current perf test suites don't create msgs large enough to send chunked.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/CordaKafkaProducerImpl.kt
```diff
@@ -140,12 +140,28 @@ class CordaKafkaProducerImpl(
...
         }
     }
 
+    private fun recordChunksCountPerTopic(cordaProducerRecords: List<CordaProducerRecord<*, *>>) {
+        cordaProducerRecords.groupBy { it.topic }
+            .mapValues { (_, records) -> records.size }
+            .forEach { (topic, count) ->
+                CordaMetrics.Metric.Messaging.ProducerChunksGenerated.builder()
+                    .withTag(CordaMetrics.Tag.MessagePatternClientId, config.clientId)
+                    .withTag(CordaMetrics.Tag.Topic, topic)
```
conalsmith-r3, line 158: arguably we may not need topic here, if we have the clientId, and we could build the metric at class level rather than rebuilding each time chunking is used.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/StateAndEventConsumerImpl.kt
```diff
@@ -59,13 +61,31 @@ internal class StateAndEventConsumerImpl<K : Any, S : Any, E : Any>(
...
+
+    private fun recordCurrentStatesMetrics() {
+        currentStates.keys.forEach { partition ->
+            currentStatesMetricCache.computeIfAbsent(partition) {
+                CordaMetrics.Metric.Messaging.StateAndEventConsumerInMemoryStoreCount.builder()
+                    .withTag(CordaMetrics.Tag.MessagePatternClientId, config.clientId)
+                    .withTag(CordaMetrics.Tag.Partition, "$partition")
+                    .build()
+            }
+                .record((currentStates[partition]?.keys?.size ?: 0).toDouble())
```
sprintengine: `currentStates` is a map of partitions to a map of states, where the key is the state key and the value is a pair of the timestamp and the state.

For the given partition we record the number of state keys.

Though this seems to have some unexpected data:
- states increasing by exactly 6 each time for each partition
- all partitions have exactly the same number of states

image

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
sprintengine: This replaces `MessagingPollTime`

libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
```diff
@@ -614,6 +594,51 @@ object CordaMetrics {
...
+                computation
+            )
+
+            /**
+             * Measure for the number of in-memory states held in consumers with partitions.
+             */
+            class PartitionedConsumerInMemoryStore(computation: Supplier<Number>) : ComputedValue<Nothing>(
+                "consumer.partitioned.inmemory.store",
+                computation
+            )
```
sprintengine, line 646: Micrometer metrics don't like when you use one metric name built in two consumers where one sets a tag ("partition") and the other doesn't, so I have two separate ones to avoid confusion

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/StateAndEventConsumerImpl.kt
```diff
@@ -60,18 +61,20 @@ internal class StateAndEventConsumerImpl<K : Any, S : Any, E : Any>(
...
-    private val statePollTimer = CordaMetrics.Metric.MessagePollTime.builder()
-        .withTag(CordaMetrics.Tag.MessagePatternType, MetricsConstants.STATE_AND_EVENT_PATTERN_TYPE)
-        .withTag(CordaMetrics.Tag.MessagePatternClientId, config.clientId)
-        .withTag(CordaMetrics.Tag.OperationName, MetricsConstants.STATE_POLL_OPERATION)
-        .build()
-
-    private val eventPollTimer = CordaMetrics.Metric.MessagePollTime.builder()
-        .withTag(CordaMetrics.Tag.MessagePatternType, MetricsConstants.STATE_AND_EVENT_PATTERN_TYPE)
-        .withTag(CordaMetrics.Tag.MessagePatternClientId, config.clientId)
-        .withTag(CordaMetrics.Tag.OperationName, MetricsConstants.EVENT_POLL_OPERATION)
```
conalsmith-r3, line 72: I think both event and state consumers might have clientId "eventConsumer" so if we change the correct one to be "stateConsumer" we can discern between state polls and event polls

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/consumer/StateAndEventConsumerImpl.kt
```diff
@@ -60,18 +61,20 @@ internal class StateAndEventConsumerImpl<K : Any, S : Any, E : Any>(
     private val partitionsToSync = ConcurrentHashMap.newKeySet<CordaTopicPartition>()
     private val inSyncPartitions = ConcurrentHashMap.newKeySet<CordaTopicPartition>()
 
-    private val statePollTimer = CordaMetrics.Metric.MessagePollTime.builder()
-        .withTag(CordaMetrics.Tag.MessagePatternType, MetricsConstants.STATE_AND_EVENT_PATTERN_TYPE)
-        .withTag(CordaMetrics.Tag.MessagePatternClientId, config.clientId)
-        .withTag(CordaMetrics.Tag.OperationName, MetricsConstants.STATE_POLL_OPERATION)
-        .build()
```
sprintengine, line 67: We can discern poll operation type "event" or "state" from the client ID - State consumer has a client ID of 
```
"${config.clientId}-stateConsumer"
```
and eventConsumer:
```
"${config.clientId}-eventConsumer"
```

jujoramos (commented):
It would be good to add tests verifying that the newly added metrics work. Otherwise, LGTM.

libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
```diff
@@ -614,6 +594,51 @@ object CordaMetrics {
...
+
+            /**
+             * Counter for the number of records a consumer has consumed.
+             */
+            object ConsumerRecordsConsumedCount : Metric<DistributionSummary>("consumer.records.consumed", Metrics::summary)
```
conalsmith-r3, line 634: Might be better to remove "Count" from this name

rick-r3 (dismissed):

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
```diff
@@ -49,11 +51,13 @@ class CordaKafkaConsumerImpl<K : Any, V : Any>(
     private val consumer: Consumer<Any, Any>,
     private var defaultListener: CordaConsumerRebalanceListener? = null,
     private val chunkDeserializerService: ConsumerChunkDeserializerService<K, V>,
-    private val consumerMetricsBinder : MeterBinder,
+    private val consumerMetricsBinder: MeterBinder,
 ) : CordaConsumer<K, V> {
     private var currentAssignment = mutableSetOf<Int>()
     private val bufferedRecords = mutableMapOf<Int, List<ConsumerRecord<Any, Any>>>()
 
+    private val recordsConsumedMetricCache = ConcurrentHashMap<Int, DistributionSummary>()
```
charlieR3: Do you need to cache these? Micrometer caches based on name a tags already.

charlieR3 (dismissed):
LGTM - one minor comment

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
sprintengine: You're right thanks, I've removed this hashmap

charlieR3 (approved):
LGTM

rick-r3 (approved):

PR #3968, opened 2023-05-26
CORE-14127: Detect whether filesystem supports POSIX.

chrisr3:
Windows _still_ doesn't support POSIX file attributes _after over 40 years!_ Detect whether the filesystem has a `PosixFileAttributeView` before passing any POSIX file attributes.


driessamyn (approved):
LGTM

PR #3969, opened 2023-05-26
CORE-13926 Cache find my signing keys

lankydan:
Cache the results of `findMySigningKeys`.

Add dependency inversion for evicting entries in a cache when a sandbox is evicted. This should make it easier for people to add sandboxed caches correctly.

This requires caches to implement `SandboxedCache`, and make there service available under that interface for OSGi to find it. `SandboxedCacheEvicter` will then handle evicting from all `SandboxedCache`s whenever a sandbox is evicted.

The writeup of the performance impact can be read here - https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4496883838/Test+3+-+Reduce+time+spent+in+SigningService.findMySigningKeys


components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxedCacheEvicterImpl.kt
```diff
@@ -0,0 +1,60 @@
...
+        if (!sandboxGroupContextComponent.removeEvictionListener(sandboxGroupType, ::onEviction)) {
+            log.error("FAILED TO REMOVE EVICTION LISTENER")
+        }
+    }
+
+    private fun onEviction(vnc: VirtualNodeContext) {
+        log.debug("Sandbox {} has been evicted", vnc)
+        for (ref in bundleContext.getServiceReferences(SandboxedCache::class.java, NON_PROTOTYPE_SERVICES)) {
+            bundleContext.getService(ref)?.also { cache ->
+                log.error(
```
rick-r3: Just info ?

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/MySigningKeysCacheImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    // TODO Access configuration to setup the cache
+    private companion object {
+        private const val MY_SIGNING_KEYS_CACHE_MAX_SIZE_PROPERTY_NAME = "net.corda.flow.application.crypto.cache.maximumSize"
+        private const val MY_SIGNING_KEYS_EXPIRE_AFTER_WRITE_SECONDS_PROPERTY_NAME =
+            "net.corda.flow.application.crypto.cache.expireAfterWriteSeconds"
+    }
+
+    private val maximumSize = java.lang.Long.getLong(MY_SIGNING_KEYS_CACHE_MAX_SIZE_PROPERTY_NAME, 10000)
+    private val expireAfterWriteSeconds = java.lang.Long.getLong(MY_SIGNING_KEYS_EXPIRE_AFTER_WRITE_SECONDS_PROPERTY_NAME, 600)
```
rick-r3: I'm not sure I get the rationale with these caches (maybe the Fibre one too) of having them expire.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/MySigningKeysCacheImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    private val cache: Cache<CacheKey<PublicKey>, CacheValue> = CacheFactoryImpl().build(
+        "My-Signing-Key-Cache",
+        Caffeine.newBuilder()
+            .expireAfterAccess(Duration.ofSeconds(expireAfterWriteSeconds))
+            .maximumSize(maximumSize)
+    )
+
+    @Suspendable
+    override fun get(keys: Set<PublicKey>): Map<PublicKey, PublicKey?> {
```
rick-r3, line 67: It does feel like there's an abstract base class implementation of `SandboxedCache` to be had here, that wraps the keys with the current holding identity for you.  But not essential now.

rick-r3 (commented):
Looks logically correct to me.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/MySigningKeysCacheImpl.kt
lankydan: Possibly, I just had it based on the flow one, happy to change it if someone has any suggestions.

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxedCacheEvicterImpl.kt
```diff
@@ -0,0 +1,60 @@
...
+                    "Evicting cached items from ${cache::class.java} with holding identity: ${vnc.holdingIdentity} and sandbox type: " +
+                            vnc.sandboxGroupType
+                )
+                cache.remove(vnc.holdingIdentity, vnc.sandboxGroupType)
+            }
+        }
```
chrisr3: _Derp_, I wasn't expecting to see any "OSGi Bongo" going on. Are you _sure_ this is doing what you expect? You will only "see" services that are visible from this service's own `BundleContext`, and I'm _hoping_ that no-one would create a `SandboxedCache` with `PROTOTYPE` scope anyway, so I'm not sure that you _need_ the `NON_PROTOTYPE_SERVICES` filter :thinking:.

If nothing else, invoking `getService()` without also invoking `ungetService()` later is [going to mess up the framework's reference counting](https://docs.osgi.org/javadoc/r4v43/core/org/osgi/framework/BundleContext.html#getService(org.osgi.framework.ServiceReference)):
> A bundle's use of a service is tracked by the bundle's use count of that service. Each time a service's service object is returned by [getService(ServiceReference)](https://docs.osgi.org/javadoc/r4v43/core/org/osgi/framework/BundleContext.html#getService(org.osgi.framework.ServiceReference)) the context bundle's use count for that service is incremented by one. Each time the service is released by [ungetService(ServiceReference)](https://docs.osgi.org/javadoc/r4v43/core/org/osgi/framework/BundleContext.html#ungetService(org.osgi.framework.ServiceReference)) the context bundle's use count for that service is decremented by one.

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxedCacheEvicterImpl.kt
```diff
@@ -0,0 +1,60 @@
...
+@Component(service = [SandboxedCacheEvicter::class])
+class SandboxedCacheEvicterImpl @Activate constructor(
+    private val bundleContext: BundleContext,
+    @Reference(service = SandboxGroupContextComponent::class)
+    private val sandboxGroupContextComponent: SandboxGroupContextComponent
+): SandboxedCacheEvicter {
+
+    private companion object {
+        private const val NON_PROTOTYPE_SERVICES = "(!(${Constants.SERVICE_SCOPE}=${Constants.SCOPE_PROTOTYPE}))"
+        val log: Logger = LoggerFactory.getLogger(SandboxedCacheEvicterImpl::class.java)
```
chrisr3: I think
```kotlin
private val log
```
is better because then Kotlin won't bother generating a `getLog()` method. To be fair, this is a "singleton" service and so you don't need to create `log` inside the `companion object` either.

Kotlin won't moan about the field having a platform type if it's `private`.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/sandbox/impl/FlowSandboxServiceImpl.kt
```diff
@@ -54,22 +52,7 @@ class FlowSandboxServiceImpl @Activate constructor(
...
-    fun shutdown() {
-        if (!sandboxGroupContextComponent.removeEvictionListener(SandboxGroupType.FLOW, ::onEviction)) {
-            logger.warn("FAILED TO REMOVE EVICTION LISTENER")
-        }
-    }
-
-    private fun onEviction(vnc: VirtualNodeContext) {
-        logger.debug("Sandbox {} has been evicted", vnc)
-        flowFiberCache.remove(vnc.holdingIdentity.toAvro())
+        sandboxedCacheEvicter.setSandboxGroupType(SandboxGroupType.FLOW)
```
chrisr3: But `SandboxCacheEvictor` is a singleton service, and so this effectively forbids `PERSISTENCE` and `VERIFICATION` sandboxes from using it too. That's not very friendly, considering that `EvictionListener` supports all sandbox types.

chrisr3 (commented):
You've added `SandboxedCacheEvictor` to `sandbox-group-context`, but only `FLOW` sandboxes will be able to use it. Was that your intent?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/sandbox/impl/FlowSandboxServiceImpl.kt
lankydan: I intend for it to be used by all the sandboxes. But, I think you're right in that this isn't correct.

My thinking was that I needed a way to configure the sandbox type passed into `sandboxGroupContextComponent.addEvictionListener(sandboxGroupType, ::onEviction)`.

I could have it so the `SandboxedCache`s themselves specify the sandbox group types (if they are used by multiple) that they are interested in and then the evicter finds these caches and adds the eviction listeners. But that would mean we register many listeners.

I could also move the code below:

```kotlin
override fun setSandboxGroupType(sandboxGroupType: SandboxGroupType) {
        require(this.sandboxGroupType == null) { "The sandbox group type cannot be set multiple times" }
        this.sandboxGroupType = sandboxGroupType
        if (!sandboxGroupContextComponent.addEvictionListener(sandboxGroupType, ::onEviction)) {
            log.error("FAILED TO ADD EVICTION LISTENER")
        }
    }

    @Suppress("unused")
    @Deactivate
    fun shutdown() {
        val sandboxGroupType = requireNotNull(this.sandboxGroupType) { "The sandbox group type has not been set when evicting" }
        if (!sandboxGroupContextComponent.removeEvictionListener(sandboxGroupType, ::onEviction)) {
            log.error("FAILED TO REMOVE EVICTION LISTENER")
        }
    }
```

Into an abstract base class for `SandboxedCache`s. Then remove the `SandboxedCacheEvicter` all together and have each cache register its own eviction listener.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/MySigningKeysCacheImpl.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import java.security.PublicKey
+
+@Component(service = [MySigningKeysCache::class])
+class MySigningKeysCacheImpl @Activate constructor(
+    @Reference(service = CurrentSandboxGroupContext::class)
+    private val currentSandboxGroupContext: CurrentSandboxGroupContext,
+    @Reference(service = SandboxGroupContextComponent::class)
+    private val sandboxGroupContextComponent: SandboxGroupContextComponent
```
chrisr3: Doesn't this just need to be `CacheEviction`? We'll obviously need to export the `CacheEviction` interface too:
```kotlin
@Component(service = [SandboxGroupContextComponent::class, CacheEviction::class])
class SandboxGroupContextComponentImpl @Activate constructor(
```
Although I don't know how well this would interact with `FakeSandboxGroupContextComponent`.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/MySigningKeysCacheImpl.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    private data class CacheValue(val publicKey: PublicKey?)
+
+    // TODO Access configuration to setup the cache
+    private companion object {
+        private val log: Logger = LoggerFactory.getLogger(MySigningKeysCacheImpl::class.java)
+        private const val MY_SIGNING_KEYS_CACHE_MAX_SIZE_PROPERTY_NAME = "net.corda.flow.application.crypto.cache.maximumSize"
+    }
+
+    private val maximumSize = java.lang.Long.getLong(MY_SIGNING_KEYS_CACHE_MAX_SIZE_PROPERTY_NAME, 10000)
```
chrisr3, line 38: This looks like an undocumented configuration parameter. Is it possible for non-`Lifecycle` components to use `SmartConfig`? How about `BundleContext.getProperty()` instead of a Java system property?

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/SigningServiceImpl.kt
```diff
@@ -35,11 +36,13 @@ class SigningServiceImpl @Activate constructor(
...
-    private val keyEncodingService: KeyEncodingService
+    private val keyEncodingService: KeyEncodingService,
+    @Reference(service = MySigningKeysCache::class)
+    private val mySigningKeysCache: MySigningKeysCache
 ) : SigningService, UsedByFlow, SingletonSerializeAsToken {
 
     private companion object {
-        private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
+        private val log: Logger = LoggerFactory.getLogger(SigningServiceImpl::class.java)
     }
```
chrisr3, line 46: Is this a singleton service? You don't need to create a `companion object` for a service which is already a singleton.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
```diff
@@ -3,22 +3,32 @@ package net.corda.flow.fiber.cache.impl
...
+import org.osgi.service.component.annotations.Reference
 import org.slf4j.LoggerFactory
+import java.time.Duration
 
 @Suppress("unused")
 @Component(service = [FlowFiberCache::class])
-class FlowFiberCacheImpl @Activate constructor() : FlowFiberCache {
+class FlowFiberCacheImpl @Activate constructor(
+    @Reference(service = SandboxGroupContextComponent::class)
+    private val sandboxGroupContextComponent: SandboxGroupContextComponent
```
chrisr3: This only need to be `CacheEviction` - like with `MySigningKeysCacheImpl`.

chrisr3 (commented):
I don't think we need to inject the full `SandboxGroupContextComponent` into either `FlowFiberCacheImpl` or `MySigningKeysCacheImpl` when just `CacheEviction` will do instead.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/MySigningKeysCacheImpl.kt
```diff
@@ -0,0 +1,89 @@
...
+        if (keys.isNotEmpty()) {
+            val virtualNodeContext = currentSandboxGroupContext.get().virtualNodeContext
+            cache.putAll(keys.map { (key, value) -> CacheKey(virtualNodeContext, key) to CacheValue(value) }.toMap())
+        }
+    }
+
+    override fun remove(holdingIdentity: HoldingIdentity, sandboxGroupType: SandboxGroupType) {
+        cache.invalidateAll(cache.asMap().keys.filter { it.holdingIdentity == holdingIdentity })
+        cache.cleanUp()
+    }
```
chrisr3: Curious - you have not used the `sandboxGroupType` parameter.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
```diff
@@ -51,9 +83,10 @@ class FlowFiberCacheImpl @Activate constructor() : FlowFiberCache {
...
 
-    override fun remove(holdingIdentity: HoldingIdentity) {
+    override fun remove(holdingIdentity: HoldingIdentity, sandboxGroupType: SandboxGroupType) {
         logger.debug { "Flow fiber cache removing holdingIdentity $holdingIdentity" }
-        val keysToInvalidate = cache.asMap().keys.filter { holdingIdentity == it.identity }
+        val holdingIdentityToRemove = holdingIdentity.toAvro()
+        val keysToInvalidate = cache.asMap().keys.filter { holdingIdentityToRemove == it.identity }
         cache.invalidateAll(keysToInvalidate)
         cache.cleanUp()
     }
```
chrisr3: You haven't used the `sandboxGroupType` parameter here either. Are you sure you need it at all?

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/MySigningKeysCacheImpl.kt
lankydan: I've left it the same as the property retrieval in `FlowFiberCacheImpl`

rick-r3 (approved):

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/MySigningKeysCacheImpl.kt
```diff
@@ -0,0 +1,89 @@
...
+
+@Component(service = [MySigningKeysCache::class])
+class MySigningKeysCacheImpl @Activate constructor(
+    @Reference(service = CurrentSandboxGroupContext::class)
+    private val currentSandboxGroupContext: CurrentSandboxGroupContext,
+    @Reference(service = CacheEviction::class)
+    private val cacheEviction: CacheEviction
+) : MySigningKeysCache, SandboxedCache {
+
+    private data class CacheValue(val publicKey: PublicKey?)
```
kyriathar, line 30: Could we remove `CacheValue`? I think it hides what it holds.

libs/virtual-node/sandbox-group-context/src/main/kotlin/net/corda/sandboxgroupcontext/SandboxedCache.kt
```diff
@@ -0,0 +1,44 @@
...
+ */
+interface SandboxedCache {
+
+    /**
+     * A cache key holding sandbox information and the _real_ key to the cache.
+     *
+     * @property holdingIdentity The [HoldingIdentity] of the sandbox that added the key-value pair.
+     * @property key The real key of the cache.
+     */
+    data class CacheKey<T>(val holdingIdentity: HoldingIdentity, val key: T) {
```
kyriathar, line 26: I don't think we expose this anywhere in the API of the cache, should it be internalized to the implementation

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/SigningServiceImpl.kt
```diff
@@ -61,13 +64,20 @@ class SigningServiceImpl @Activate constructor(
 
     @Suspendable
     override fun findMySigningKeys(keys: Set<PublicKey>): Map<PublicKey, PublicKey?> {
-        return recordSuspendable({ cryptoFlowTimer("findMySigningKeys") }) @Suspendable {
-            val compositeKeys: Set<CompositeKey> = keys.filterIsInstanceTo(linkedSetOf(), CompositeKey::class.java)
-            val plainKeys = keys - compositeKeys
+        val operation = operation@ @Suspendable {
```
kyriathar, line 67: Could we revert back to what it was before, I think this val adds some overhead which is a bit unclear too of what it does.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
```diff
@@ -3,22 +3,32 @@ package net.corda.flow.fiber.cache.impl
...
 import org.slf4j.LoggerFactory
+import java.time.Duration
 
 @Suppress("unused")
 @Component(service = [FlowFiberCache::class])
-class FlowFiberCacheImpl @Activate constructor() : FlowFiberCache {
+class FlowFiberCacheImpl @Activate constructor(
+    @Reference(service = CacheEviction::class)
+    private val cacheEviction: CacheEviction
+) : FlowFiberCache, SandboxedCache {
```
kyriathar, line 28: `FlowFiberCache` is already a `SandboxedCache` I think

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/MySigningKeysCache.kt
```diff
@@ -0,0 +1,30 @@
...
+     *      * </ul>
+     */
+    fun get(keys: Set<PublicKey>): Map<PublicKey, PublicKey?>
+
+    /**
+     * Put signing keys into the cache.
+     *
+     * @param keys The signing keys to cache.
+     */
+    fun putAll(keys: Map<out PublicKey, PublicKey?>)
```
kyriathar, line 29: I think this cache could just be holding plain keys (i.e. not composite). 

The flattening of composite keys should be done by the caller and then this just gets which of the plain keys found to be _mine_.

kyriathar (commented):
I feel the cache abstraction could be omitted and instead we use the cache directly wherever needed (`SigningServiceImpl`?). Do we know it's going to be needed elsewhere so then maybe "service" cache makes sense to be re-used then.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/MySigningKeysCache.kt
kyriathar: Discussed offline with @lankydan. These APIs are actually OK because we need to also cache the negatives, i.e. the not owned keys so we save going to Crypto worker for those. So please disregard the above comment.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/MySigningKeysCacheImpl.kt
kyriathar: Maybe the cache could become HoldingId -> Map<PublicKey, PublicKey?> then we no longer need that `CacheValue` (discussed with @lankydan its purpose is to pass in to cache not null references) and also saves a bit of holding id overhead in cache keys too

PR #3971, opened 2023-05-26
CORE-14178 - Adjust gateway metrics to populate same set of tags for a specific metric name

dimosr:
Adjust gateway metrics to populate same set of tags for a specific metric name.

I also adjusted the code around metrics for inbound/outbound connections to make it less error-prone to not specifying the same set of tags always and make it a bit easier to read.

### DIFF
--- a/components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/internal/OutboundMessageHandler.kt
+++ b/components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/internal/OutboundMessageHandler.kt
@@ -232,9 +232,8 @@ internal class OutboundMessageHandler(
     ): Timer {
         val builder = CordaMetrics.Metric.OutboundGatewayRequestLatency.builder()
         builder.withTag(CordaMetrics.Tag.DestinationEndpoint, peerMessage.header.address)
-        if (response != null) {
-            builder.withTag(CordaMetrics.Tag.HttpResponseType, response.statusCode.code().toString())
-        }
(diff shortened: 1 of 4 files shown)


components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/internal/OutboundMessageHandler.kt
```diff
@@ -232,9 +232,8 @@ internal class OutboundMessageHandler(
     ): Timer {
         val builder = CordaMetrics.Metric.OutboundGatewayRequestLatency.builder()
         builder.withTag(CordaMetrics.Tag.DestinationEndpoint, peerMessage.header.address)
-        if (response != null) {
-            builder.withTag(CordaMetrics.Tag.HttpResponseType, response.statusCode.code().toString())
-        }
+        val responseType = response?.statusCode?.code()?.toString() ?: "none"
```
yift-r3, line 235: Maybe `NOT_APPLICABLE_TAG_VALUE` instead of `"none"`?

yift-r3 (approved):

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/internal/OutboundMessageHandler.kt
dimosr: The reason I went for `none` in this case was to indicate a bit more strongly that there was no response from the server (e.g. timeout or client error).

driessamyn (approved):
LGTM

PR #3979, opened 2023-05-30
CORE-14204 Use cashed EntityManagerFactory for virtual node reconciliation

mbrkic-r3:
- Using cached `EntityManagerFactory` in `VirtualNodeReconciliationContext`

Note: This cannot be merged until proper cache is used: https://github.com/corda/corda-runtime-os/pull/3979#discussion_r1209974585

### DIFF
--- a/components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DbConnectionOpsCachedImpl.kt
+++ b/components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DbConnectionOpsCachedImpl.kt
@@ -7,7 +7,7 @@ import net.corda.db.schema.CordaDb
 import net.corda.libs.configuration.SmartConfig
 import net.corda.orm.JpaEntitiesRegistry
 import net.corda.orm.JpaEntitiesSet
-import java.util.*
+import java.util.UUID
 import java.util.concurrent.ConcurrentHashMap
 import javax.persistence.EntityManager
 import javax.persistence.EntityManagerFactory
@@ -55,6 +55,15 @@ class DbConnectionOpsCachedImpl(
     ): EntityManagerFactory {
         return cache.computeIfAbsent(Pair(name,privilege)) {
             delegate.getOrCreateEntityManagerFactory(name, privilege, entitiesSet)
+        }
+    }
+
+    override fun getOrCreateEntityManagerFactory(
+        connectionId: UUID,
+        entitiesSet: JpaEntitiesSet
+    ): EntityManagerFactory {
+        return cache.computeIfAbsent(Pair(connectionId.toString(), DbPrivilege.DML)) {
+            delegate.getOrCreateEntityManagerFactory(connectionId, entitiesSet)
         }
     }
 }
--- a/components/crypto/crypto-softhsm-impl/src/test/kotlin/net/corda/crypto/softhsm/impl/TestSigningRepositoryImpl.kt
+++ b/components/crypto/crypto-softhsm-impl/src/test/kotlin/net/corda/crypto/softhsm/impl/TestSigningRepositoryImpl.kt
@@ -1,6 +1,5 @@
 package net.corda.crypto.softhsm.impl
 
-import javax.persistence.EntityManagerFactory
 import net.corda.crypto.core.CryptoTenants
 import net.corda.crypto.persistence.getEntityManagerFactory
 import net.corda.db.connection.manager.DbConnectionManager
@@ -17,6 +16,7 @@ import org.mockito.kotlin.mock
 import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
 import org.mockito.kotlin.verifyNoMoreInteractions
+import javax.persistence.EntityManagerFactory
 
 class TestSigningRepositoryImpl {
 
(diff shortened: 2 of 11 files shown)


components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DbConnectionOpsCachedImpl.kt
```diff
@@ -57,4 +57,13 @@ class DbConnectionOpsCachedImpl(
             delegate.getOrCreateEntityManagerFactory(name, privilege, entitiesSet)
         }
     }
+
+    override fun getOrCreateEntityManagerFactory(
+        connectionId: UUID,
+        entitiesSet: JpaEntitiesSet
+    ): EntityManagerFactory {
+        return cache.computeIfAbsent(Pair(connectionId.toString(), DbPrivilege.DML)) {
```
driessamyn, line 65: This cannot be merged as is.
This change means vnode connection pools  will always be cached, which means postgres will run out of connections.
There are tickets on the DB worker team to address this.

driessamyn (changes requested):

components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DbConnectionOpsCachedImpl.kt
mbrkic-r3: right, proper cache should be used

> // TODO - replace with caffeine cache

mbrkic-r3:
Connections are created each time also here:

@conalsmith-r3 @fowlerrr 
https://github.com/corda/corda-runtime-os/blob/release/os/5.0/components/crypto/crypto-persistence/src/main/kotlin/net/corda/crypto/persistence/GetEntityManagerFactory.kt#L35

@yift-r3 @charlieR3 
https://github.com/corda/corda-runtime-os/blob/release/os/5.0/components/membership/certificates-service-impl/src/main/kotlin/net/corda/membership/certificate/service/impl/DbClientImpl.kt#L100
https://github.com/corda/corda-runtime-os/blob/release/os/5.0/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistenceHandler.kt#L79

dimosr:
> @yift-r3 @charlieR3
https://github.com/corda/corda-runtime-os/blob/release/os/5.0/components/membership/certificates-service-impl/src/main/kotlin/net/corda/membership/certificate/service/impl/DbClientImpl.kt#L100
https://github.com/corda/corda-runtime-os/blob/release/os/5.0/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistenceHandler.kt#L79

The membership components were changed in the 5.1 branch to re-use a cached EMF (https://github.com/corda/corda-runtime-os/pull/3916). We can consider porting that to 5.0.

But, this was preceded by some changes that allow the DB connection pool to expand/shrink on demand depending on activity and I think that should also come with it, if that's the case (https://github.com/corda/corda-runtime-os/pull/3909). Yiftach's been doing some testing with onboarding a large number of vnodes in a network to ensure there are no issues there, which I think is ongoing but we haven't seen issues with up to 75 vnodes.

components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DbConnectionOpsCachedImpl.kt
driessamyn: But first the challenge around managing connections in the pool need to be solved.

driessamyn:
> > @yift-r3 @charlieR3
> > https://github.com/corda/corda-runtime-os/blob/release/os/5.0/components/membership/certificates-service-impl/src/main/kotlin/net/corda/membership/certificate/service/impl/DbClientImpl.kt#L100
> > https://github.com/corda/corda-runtime-os/blob/release/os/5.0/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistenceHandler.kt#L79
> 
> The membership components were changed in the 5.1 branch to re-use a cached EMF (#3916). We can consider porting that to 5.0.
> 
> But, this was preceded by some changes that allow the DB connection pool to expand/shrink on demand depending on activity and I think that should also come with it, if that's the case (#3909). Yiftach's been doing some testing with onboarding a large number of vnodes in a network to ensure there are no issues there, which I think is ongoing but we haven't seen issues with up to 75 vnodes.

This is where it gets messy and why I had preferred this was all managed in one place by one team.
It _should_ be managed in the DbConnectionManager, but not like this PR.
My understanding was that your changes, @dimosr , were tactical, so they will need to be pulled out once it's resolved centrally. But I worry we'll have overlapping caches and end up with a mess.

Tejas and @simon-johnson-r3 have this on the backlog.

components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DbConnectionOpsCachedImpl.kt
mbrkic-r3: Sure, at this point this PR is mainly used to see how does caching EMF improve results of https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4491870437/Test+7+-+Simple+flow+scenario+-+Varying+number+of+virtual+nodes (running initial test against this branch at the moment).

dimosr:
> This is where it gets messy and why I had preferred this was all managed in one place by one team.
It should be managed in the DbConnectionManager, but not like this PR.
My understanding was that your changes, @dimosr , were tactical, so they will need to be pulled out once it's resolved centrally. But I worry we'll have overlapping caches and end up with a mess.

> Tejas and @simon-johnson-r3 have this on the backlog.

No, that wasn't the tactical fix we were originally planning. Initially, we were considering to add an extra layer of cache on the membership side, but after our discussions we abandoned that. The changes that we merged were considered by us to be the proper long-term solution, i.e. adjusting the membership components so that they use the `getOrCreateEntityManagerFactory` method which provides cached EMFs and also adjusting the DB code so that vnode connection pools start from 0 (instead of directly consuming the full size), expand depending on demand up to the max size and return idle connections back.

mbrkic-r3:
This PR is not needed as another one  (that disables reconciliation) will be opened for performance testing purposes, and there is already another ticket to address reconciliation performance.

PR #3980, opened 2023-05-30
CORE-13934 bypass backchain resolution if there are no dependencies

nkovacsx:
### Overview

The backchain resolution flow was initiated even if the given transaction had no dependencies. 
We now added a check and a trace log to not call it in certain situations. 
This will also improve performance in issuance situations where other parties are involved.

Perf testing: https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4496752831/Test+5+Bypassing+back-chain+resolution+if+transaction+has+no+dependencies

Please note that this document contains important information about performance impact of exceeding sandbox cache size too.

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
@@ -8,6 +8,7 @@ import net.corda.ledger.utxo.flow.impl.flows.finality.getVisibleStateIndexes
 import net.corda.ledger.utxo.flow.impl.flows.finality.v1.FinalityNotarizationFailureType.Companion.toFinalityNotarizationFailureType
 import net.corda.ledger.utxo.flow.impl.transaction.UtxoSignedTransactionInternal
 import net.corda.sandbox.CordaSystemFlow
+import net.corda.utilities.trace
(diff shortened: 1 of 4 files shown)


lankydan (dismissed):
Change is fine, needs perf testing.

lankydan:
Probably add a unit test for it as well.

nkovacsx:
@lankydan added unit test

lankydan (approved):

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -69,7 +70,14 @@ class UtxoReceiveFinalityFlowV1(
         if (log.isDebugEnabled) {
```
thiagoviana, line 70: Outside your work, but I wonder if this test could not being made internally to avoid repeated checks everywhere? 

The same for isTraceEnabled.

lankydan: The logs are like this because of https://r3-cev.atlassian.net/browse/CORE-8856

relyafi (approved):
LGTM, nice use of the test framework to verify we don't trigger the subflow.

Think it might be worth adding to the PR description that it will improve performance in issuance situations where other parties are involved.

rick-r3 (approved):

PR #3981, opened 2023-05-30
CORE-13146: Eliminate Hikari Pool warning

vkolomeyko:
Before this change there were multiple instances of warning like this in the Combined Worker log:
```
2023-05-30 11:02:16.214 [SecManagerForkJoinPool-worker-2] WARN  com.zaxxer.hikari.HikariConfig {} - HikariPool-56 - idleTimeout has been set but has no effect because the pool is operating as a fixed size pool.
```
During production run, same short of warning is likely to be occurring in DB Worker log.


driessamyn (approved):
LGTM

PR #3989, opened 2023-05-31
CORE-14242: Upgrade Kotlin Metadata 0.6.0 -> 0.6.1.

chrisr3:
Upgrade to [kotlinx-metadata-jvm 0.6.1](https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/ChangeLog.md), which is able to parse Kotlin 2.0 metdata.

> This release uses Kotlin 1.8.20 with metadata version 1.8, and as a special case, is able to read metadata of version 2.0. This is done as an incentive to test K2 compiler and 2.0 language version. No other changes were made and no migration is needed.


driessamyn (approved):
LGTM

PR #3990, opened 2023-05-31
CORE-14246: Ensure CPK chunks are created and assembled atomically.

chrisr3:
This issue concerns `CpkReadServiceImpl` and consequently all services which use it, such as the sandboxes and flow processing etc. `CpkReadServiceImpl` receives CPK chunks for assembly directly from Kafka.

Prevent CPKs and CPK chunks from being read until they have been written _entirely_ to disk. We achieve this by writing the contents to an invisible temporary file in the target directory, and only moving this _atomically_ to its expected name once writing is complete. Also make the final file "read only" to protect against accidental subsequent writes.

I have confirmed that `ATOMIC_MOVE` works on Windows.

The `assembleFileFromChunks()` function is used by `ChunkDbWriterFactoryImpl` to verify that a set of chunks fetched from the database defines a valid CPI. It apparently achieves concurrency by constantly overwriting the disk with the same information, which sounds I/O intensive to me. However, I don't think that counts as a "bug". (And if it does then it's likely too difficult to fix for Corda 5.0.)


driessamyn (approved):
LGTM

PR #3991, opened 2023-05-31
CORE-14203: Fix Session Event Wrapping

jujoramos:
Ensure that all messages sent to 'p2p.out' topic are wrapped as
instaces of 'AppMessage' to prevent deserialization issues.

### DIFF
--- a/libs/flows/flow-mapper-impl/src/test/kotlin/net/corda/flow/mapper/impl/executor/SessionInitExecutorTest.kt
+++ b/libs/flows/flow-mapper-impl/src/test/kotlin/net/corda/flow/mapper/impl/executor/SessionInitExecutorTest.kt
@@ -103,12 +103,12 @@ class SessionInitExecutorTest {
 
     @Test
     fun `Subsequent OUTBOUND SessionInit messages get passed through if no ACK received from first message`(){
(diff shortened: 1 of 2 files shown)


thiagoviana (dismissed):

JamesHR3 (dismissed):
@LWogan @emilybowe - think this is the fix we were talking about yesterday?

LGTM

emilybowe (dismissed):
LGTM

libs/flows/flow-mapper-impl/src/test/kotlin/net/corda/flow/mapper/impl/executor/SessionInitExecutorTest.kt
```diff
@@ -142,4 +142,4 @@ class SessionInitExecutorTest {
 
         assertThat(resultOutbound.outputEvents).isEmpty()
```
LWogan, line 147: i think we should assert that the class type of the output event is AppMessage

jujoramos: Done 👍

LWogan (approved):
LGTM

rick-r3 (approved):

PR #3992, opened 2023-05-31
CORE-13353: redact salt in logs

dickon:
https://r3-cev.atlassian.net/browse/CORE-14015 tracks the merge work
Following procedure at https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4396089494/Keeping+release+branches+in+Sync

### DIFF
--- a/applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/WorkerHelpers.kt
+++ b/applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/WorkerHelpers.kt
@@ -34,7 +34,7 @@ class WorkerHelpers {
     companion object {
         private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
         private const val BOOT_CONFIG_PATH = "net/corda/applications/workers/workercommon/boot/corda.boot.json"
-        private val SENSITIVE_ARGS = setOf("-ddatabase.pass", "-spassphrase", "-msasl.jaas.config")
+        private val SENSITIVE_ARGS = setOf("-ddatabase.pass", "-spassphrase", "-msasl.jaas.config", "-ssalt")
 
         /**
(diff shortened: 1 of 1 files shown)


applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/WorkerHelpers.kt
```diff
@@ -34,7 +34,7 @@ class WorkerHelpers {
     companion object {
         private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
         private const val BOOT_CONFIG_PATH = "net/corda/applications/workers/workercommon/boot/corda.boot.json"
-        private val SENSITIVE_ARGS = setOf("-ddatabase.pass", "-spassphrase", "-msasl.jaas.config")
+        private val SENSITIVE_ARGS = setOf("-ddatabase.pass", "-spassphrase", "-msasl.jaas.config", "-ssalt")
```
driessamyn, line 37: We don't have a unit test covering this (which would be a bit tricky). Can you run a manual test and confirm with a copy of the logs? (or take from the jenkins E2E test logs if possible)

dickon: I did think about that question. If this one turns out to be low priority, i.e. the consensus decision is that a salt is not confidential, and we should redact it anyway, then maybe I'm happy to let QA rerun their test, which is probably automated, rather than also implement our own.

driessamyn: Y. It's probably not worth an automated test, but you should probably run a manual one anyway. I suspect this is in the logs on Jenkins.

dickon:
Closed since we have decided the salt can be logged.

PR #3993, opened 2023-05-31
CORE-14290 Publish percentile histogram for timer metrics to allow calculation of percentiles on aggregation of metrics

charlieR3:
Publish percentile histogram for timer metrics to allow calculation of percentiles on aggregation of metrics. 
Client side publication of metrics does not allow for aggregation of metrics on the server side. Publishing the percentile histogram allows us to group by operation type and namespace and get the percentile values.

It should be possible to remove the usage of `publishPercentiles` for timer metrics, but to allow others time to switch to the histogram for rendering percentile graphs, this will be removed in a follow-on PR.


conal-smith (approved):

Omar-awad:
Looks good to me. We need this to have better insights on the latency of our functions.

JamesHR3 (approved):

driessamyn (approved):

PR #3994, opened 2023-05-31
CORE-13862 Add metrics in DB worker

kyriathar:
- adds metrics for: 
  - DB entity persistence request processing time
  - DB entity persistence request Kafka lag (i.e. how long the stayed on Kafka)
  - Reconciliation run time
  - Number of Reconciled records

### DIFF
--- a/components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityMessageProcessor.kt
+++ b/components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityMessageProcessor.kt
@@ -70,6 +70,13 @@ class EntityMessageProcessor(
                 // We received a [null] external event therefore we do not know the flow id to respond to.
                 return@mapNotNull null
             } else {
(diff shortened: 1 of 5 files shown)


components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityMessageProcessor.kt
```diff
@@ -67,6 +70,13 @@ class EntityMessageProcessor(
                 // We received a [null] external event therefore we do not know the flow id to respond to.
                 return@mapNotNull null
             } else {
+                CordaMetrics.Metric.Db.EntityPersistenceRequestLag.builder()
+                    .withTag(CordaMetrics.Tag.OperationName, request.request::class.java.name)
+                    .build()
+                    .record(
+                        Duration.ofMillis(Instant.now().toEpochMilli() - event.timestamp)
```
driessamyn, line 77: I think we have to be very careful with this metric. 
This relies on clock sync between workers, and measuring in ms seems like we could be reporting completely inaccurate numbers if there is even the slightest clock drift.

driessamyn (approved):
LGTM

PR #3998, opened 2023-06-01
CORE-13987: Make Timer.recordOptionally() more accurate.

chrisr3:
`System.nanoTime()` is the preferred way of timing durations. We also don't need to compute the duration more than once.


driessamyn (approved):
LGTM

PR #4002, opened 2023-06-01
CORE-14319 add compression to kafka producers

LWogan:
This was seen to double the throughput of messages through the message patterns

### DIFF
--- a/libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
+++ b/libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
@@ -40,6 +40,8 @@ producer = ${common} {
     # Ensures that messages are not lost due to broker failure at inopportune moments. This forces acknowledgements from
     # all in-sync replicas before continuing, which ensures the message has reached at least one broker.
     acks = all
+    # Ensure producers use compression when sending data
+    compression.type = zstd
 }
 
 # Roles that particular consumers or producers could be taking. By tying consumers and producers to the roles they are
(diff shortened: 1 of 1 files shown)


jujoramos (dismissed):

libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
```diff
@@ -17,9 +17,9 @@ consumer = ${common} {
     isolation.level = read_committed
     # Force consumers to commit their offsets manually. Again required for transactionality.
     enable.auto.commit = false
-    # Retrieve 500 records maximum per poll. Note that poll will return immediately if any records are available, so a
-    # batch may contain fewer than 500 records.
-    max.poll.records = 500
+    # Retrieve 2000 records maximum per poll. Note that poll will return immediately if any records are available, so a
+    # batch may contain fewer than 2000 records.
+    max.poll.records = 2000
```
rick-r3: I would be cautious about this setting, as it increases the maximum batch size we might pick up.   e.g. we could pick up 2000 database requests in one go, which take e.g. 5 ms each, which then means the batch takes 10 seconds to complete - making our latency of flows worse.

LWogan: Fair, ill revert, I was testing with dumb processors. The batches are very small right now though. Perhaps we need to configure per topic.

LWogan: While were on the subject, we currently use default setting for fetch of 1mb per partition assigned. We are currently recommending  5x partition per consumer in non functional repo helm chart comments. So this would be 5mb per poll max. But if we decide to go 1 partition per consumer for operational reasons. we need to bump the fetch max size up for consumers that process larger messages as they wont come close to the 500. I had set it to 5mb per partition in this pr which showed better throughput (not just because of compression), but im unclear what the real guidance is on partitions per consumer.

rick-r3 (approved):

PR #4003, opened 2023-06-01
CORE-14320: Create a ComputedValue gauge metric.

chrisr3:
Create a `Gauge` metric that wraps `Supplier<Number>` instead of `AtomicInteger`. The advantage of `Supplier<Number>` is that the gauge will automatically compute a new value each time without constantly needing for one to be "set" manually.


driessamyn (approved):
LGTM

PR #4008, opened 2023-06-02
CORE-13942 Add single-cluster mode to app simulator scripts

YashNabar:
Updates the app simulator deployment scripts to enable single-cluster mode. Also includes a fix in the Receiver to filter out messages not meant for the simulator (e.g. messages meant for MGM in single cluster mode should be ignored by the simulator). To run in single-cluster mode, set environment variable `CLUSTER_MODE=SINGLE_CLUSTER`.


applications/tools/p2p-test/app-simulator/src/main/kotlin/net/corda/p2p/app/simulator/Receiver.kt
```diff
@@ -70,14 +70,27 @@ class Receiver(private val subscriptionFactory: SubscriptionFactory,
...
-                val authenticatedMessage = it.value!!.message as AuthenticatedMessage
-                //Only JSON deserialize messages from another app-simulator (not sent by the MGM for example).
-                if (authenticatedMessage.header.subsystem == APP_SIMULATOR_SUBSYSTEM) {
-                    val payload = objectMapper.readValue<MessagePayload>(authenticatedMessage.payload.array())
-                    val messageReceivedEvent = MessageReceivedEvent(payload.sender,
-                        authenticatedMessage.header.messageId, payload.sendTimestamp, now, Duration.between(payload.sendTimestamp, now))
-                    Record(destinationTopic, messageReceivedEvent.messageId, objectMapper.writeValueAsString(messageReceivedEvent))
-                } else {
+                try {
+                    val authenticatedMessage = it.value!!.message as AuthenticatedMessage
```
williamvigorr3: Should we do:
```
val authenticatedMessage = it.value!!.message as? AuthenticatedMessage

//Only JSON deserialize messages from another app-simulator (not sent by the MGM for example).
if (authenticatedMessage?.header.subsystem == APP_SIMULATOR_SUBSYSTEM) {
```
Instead of the try catch block.

applications/tools/p2p-test/app-simulator/scripts/onBoardCluster.sh
```diff
@@ -320,26 +320,38 @@ on_board_node() {
...
    echo Node $2 Ledger Key Id: $NODE_LEDGER_KEY_ID
 
-   # Generate Key Pair for TLS for MGM
-   NODE_TLS_KEY_ID=$(assign_hsm_and_generate_tls_key_pair $1)
-   echo NODE $2 TLS Key Id: $NODE_TLS_KEY_ID
+   # Generate Key Pair for TLS for Member
+   if [ $CLUSTER_MODE == "SINGLE_CLUSTER" ]
+   then
+     NODE_TLS_KEY_ID=$6
+     echo NODE $2 TLS Key Id: $NODE_TLS_KEY_ID
```
williamvigorr3: I think `$6` is the `$MGM_HOLDING_ID_SHORT_HASH`. So I think you are echoing the wrong thing here which is a bit confusing? I don't think `NODE_TLS_KEY_ID` is used for `SINGLE_CLUSTER`? So this block could be empty.

applications/tools/p2p-test/app-simulator/scripts/onBoardCluster.sh
```diff
@@ -352,6 +364,6 @@ java -jar $CA_JAR --home=./ca create-ca
 
 on_board_mgm
 
-on_board_node $A_RPC $A_X500_NAME $A_GATEWAY_ADDRESS a_tls $A_GATEWAY_ENDPOINT
+on_board_node $A_RPC $A_X500_NAME $A_GATEWAY_ADDRESS a_tls $A_GATEWAY_ENDPOINT $MGM_HOLDING_ID_SHORT_HASH
```
williamvigorr3: I think we can remove the last argument.

applications/tools/p2p-test/app-simulator/scripts/onBoardCluster.sh
```diff
@@ -53,7 +53,7 @@ build_cpi() {
    keytool -genkeypair -alias "signing key 1" -keystore signingkeys.pfx -storepass "keystore password" -dname "cn=CPI Plugin Example - Signing Key 1, o=R3, L=London, c=GB" -keyalg RSA -storetype pkcs12 -validity 4000
    keytool -importcert -keystore signingkeys.pfx -storepass "keystore password" -noprompt -alias gradle-plugin-default-key -file "$SCRIPT_DIR/gradle-plugin-default-key.pem"
 
-   $CORDA_CLI_DIR/build/generatedScripts/corda-cli.sh package create-cpi --cpb test-cordapp-$CORDA_VERSION-SNAPSHOT-package.cpb --group-policy $1 --cpi-name "test cordapp" --cpi-version "1.0.0.0-SNAPSHOT" --file test-cordapp-5.0.0.0-SNAPSHOT-package.cpi --keystore signingkeys.pfx --storepass "keystore password" --key "signing key 1" 
+   $CORDA_CLI_DIR/build/generatedScripts/corda-cli.sh package create-cpi --cpb test-cordapp-$CORDA_VERSION-SNAPSHOT-package.cpb --group-policy $1 --cpi-name "test cordapp $2" --cpi-version "1.0.0.0-SNAPSHOT" --file test-cordapp-5.0.0.0-SNAPSHOT-package.cpi --keystore signingkeys.pfx --storepass "keystore password" --key "signing key 1"
```
williamvigorr3, line 56: I assume it is Okay to put a comma in the CPI name (if we add the X500 name)?

williamvigorr3 (commented):
Left a few minor comments. Looks good.

applications/tools/p2p-test/app-simulator/src/main/kotlin/net/corda/p2p/app/simulator/Receiver.kt
YashNabar: It still throws a `ClassCastException` which needs to be caught

applications/tools/p2p-test/app-simulator/scripts/onBoardCluster.sh
YashNabar: Yes, it doesn't complain

applications/tools/p2p-test/app-simulator/scripts/onBoardCluster.sh
YashNabar: Ah yes, I'll correct that, thanks!

applications/tools/p2p-test/app-simulator/src/main/kotlin/net/corda/p2p/app/simulator/Receiver.kt
williamvigorr3: Can we put the try catch block around line 74 only?

yift-r3: Maybe `it.value?.message as? AuthenticatedMessage`

dimosr: You can also add an extra if statement (or `when` clause) if you want to avoid all the `?`, e.g. something like
```
if (it.value!!.message is AuthenticatedMessage) {
    // existing logic
} else {
   null
}
```

I find it a bit harder to follow the logic when there are many chained `?`, but might be just me. And there's also a trade-off of a slightly more nested logic.

YashNabar: You're right, didn't really need that try-catch, I've corrected it in [697e990](https://github.com/corda/corda-runtime-os/pull/4008/commits/697e99096f83b5a3a2af0bc6033c203b33da21f3). Thanks both!

williamvigorr3 (approved):
LGTM As long as @dimosr is happy with the chained `?`. I don't think it matters too much in this case as the nullable code was generated by AVRO and so cannot be null in practise but it maybe is a bit confusing to someone reading the code, who isn't familiar with that.

dimosr (approved):
No big deal for the chained `?`. It's personal preference, so if people disagree I won't insist.

applications/tools/p2p-test/app-simulator/src/main/kotlin/net/corda/p2p/app/simulator/Receiver.kt
YashNabar: Missed your comment before @dimosr, sure I can change that

PR #4009, opened 2023-06-02
CORE-13987: Make more Timer metrics use System.nanoTime().

chrisr3:
We should be using `System.nanoTime()` to compute durations of elapsed time. The `Timer.recordCallable` API already uses `System.nanoTime()` internally, and this PR fixes low-hanging fruit where we were using the "wall time" clock.


driessamyn (approved):
LGTM

PR #4014, opened 2023-06-04
CORE-14320: Reimplement MemberListCache metrics with ComputedValue.

chrisr3:
Replace `SettableGauge` with `ComputedValue` so that we don't constantly need to update each gauge's value "by hand".

The ultimate goal here is to remove `SettableGauge` entirely.

---
A `ComputedValue` gauge is actually trickier than it first appears because the `MeterRegistry` hold a strong reference to the `Supplier<Number>` lambda, and hence to any objects that the lambda uses. It is therefore prudent to remove such gauges when no longer needed to allow their objects to be garbage-collected.

Alternatively, we can create a `ComputedValue` gauge around a single object and provide a `ToDoubleFunction` lambda to compute a property for it. The `MeterRegistry` will then only hold a weak reference to the object, and the gauge will return NaN once the object is garbage-collected. The gauge still needs to be removed from the `MeterRegistry` though.


charlieR3 (dismissed):
LGTM

charlieR3 (approved):
LGTM

driessamyn (approved):
LGTM

PR #4019, opened 2023-06-05
CORE-14350: Standardise publisher clientIds for metrics traceability

fowlerrr:
By having a standard we should be able to better track Kafka metrics in the cluster

### DIFF
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/config/ResolvedPublisherConfig.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/config/ResolvedPublisherConfig.kt
@@ -1,6 +1,7 @@
 package net.corda.messaging.config
 
 import java.time.Duration
+import java.util.UUID
 import net.corda.libs.configuration.SmartConfig
 import net.corda.messaging.api.publisher.config.PublisherConfig
 import net.corda.schema.configuration.BootConfig.INSTANCE_ID
@@ -25,14 +26,17 @@ internal data class ResolvedPublisherConfig(
          * @return Concrete class containing all config values used by a publisher.
          */
         fun merge(publisherConfig: PublisherConfig, messagingConfig: SmartConfig): ResolvedPublisherConfig {
+            val transactionalFlag = if (publisherConfig.transactional) "TRANSACTIONAL" else "ASYNCHRONOUS"
+            val clientId = "$transactionalFlag--${publisherConfig.clientId}--${publisherConfig.topic}--${UUID.randomUUID()}"
             return ResolvedPublisherConfig(
-                publisherConfig.clientId,
+                clientId,
                 messagingConfig.getInt(INSTANCE_ID),
                 publisherConfig.transactional,
                 Duration.ofMillis(messagingConfig.getLong(CLOSE_TIMEOUT)),
                 messagingConfig
             )
         }
+
     }
 
     val loggerName = "PUBLISHER-$clientId"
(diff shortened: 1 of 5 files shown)


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/config/ResolvedPublisherConfig.kt
```diff
@@ -25,14 +26,17 @@ internal data class ResolvedPublisherConfig(
          * @return Concrete class containing all config values used by a publisher.
          */
         fun merge(publisherConfig: PublisherConfig, messagingConfig: SmartConfig): ResolvedPublisherConfig {
+            val transactionalFlag = if (publisherConfig.transactional) "TRANSACTIONAL" else "ASYNCHRONOUS"
+            val clientId = "$transactionalFlag--${publisherConfig.clientId}--${publisherConfig.topic}--${UUID.randomUUID()}"
```
Omar-awad: Need to check the clientId against a regex of acceptable Id formats to avoid Ids like "This is the Avro schema publisher".

vlajos: I have added a space removal from the clientIds, and removed it from one place manually.
Also checked the the other ones. Mostly tests, and they are not super consistent, but I have not found anything suspicious.
(`-`, `.`, `_` were the extras on the top of the normal characters.)

Omar-awad:
Please can you add the topic names for the functions that triggers main processes like:
- Wake up calls
- Cleanups
- State consumption

LWogan:
> Please can you add the topic names for the functions that triggers main processes like:
> 
> * Wake up calls
> * Cleanups
> * State consumption

Kafka producers can send to any topic. In Corda it is common in most of our subscriptions that the producer will send to many topics in a single transaction. It doesn't make sense to tie a producer to a topic for its client id. The consumer group name might be a better way to tie a producer and consumer together for the patterns which use both as part of the consumer&produce loop?

Omar-awad:
Knowing the benefits of having one producer for all is network efficient the downside is it proves difficult to optimize each poroducer to a particular use case or behaviour which are generally tied to a selected group of topics. The intention here in any case is not to make a decision of which best method suites us rather than to make an effort to have meaningful names that we can use to gather telemetry from. Relying on client Id's is not optimal but works in the short term knowing that we have 1 topic per producer in practice on majority of the cases in Corda.

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/publisher/config/PublisherConfig.kt
```diff
@@ -4,8 +4,10 @@ package net.corda.messaging.api.publisher.config
  * Class to store the required params to create a Publisher.
  * @property clientId Is an identifier of the source of a record.
  * @property transactional True to publish records as a transaction, false to send asynchronously.
+ * @property topic If there will only be one topic that this publisher sends to then it can put in the clientId for metrics
  */
 data class PublisherConfig (
     val clientId: String,
     val transactional: Boolean = true,
+    val topic: String = "general",
```
driessamyn, line 12: This looks a bit weird.
Shouldn't this be a list of topics, and what is the magic `general`value?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/config/ResolvedPublisherConfig.kt
```diff
@@ -25,14 +26,22 @@ internal data class ResolvedPublisherConfig(
          * @return Concrete class containing all config values used by a publisher.
          */
         fun merge(publisherConfig: PublisherConfig, messagingConfig: SmartConfig): ResolvedPublisherConfig {
+            val transactionalFlag = if (publisherConfig.transactional) "TRANSACTIONAL" else "ASYNCHRONOUS"
```
driessamyn, line 29: `Asynchronous` isn't the opposite of `Transactional`

I'm not  convinced that the clientID is the right place to record this.
It may make more sense to just trace/debug level log the publisherConfig when the producer is created.

driessamyn:
@thiagoviana , I discussed this with @Omar-awad , I think it's worth somebody from your team to have a look. If we have inconsistency in the way clients are named, my view is that we should fix that at source.

fowlerrr:
We've gone in a different direction and this work will no longer be needed.

PR #4022, opened 2023-06-06
CORE-14343: Add tracing sample rate configuration

josephzunigadaly:
* Add new command line option `--trace-samples-per-second`
* Defaults to 1 sample per second if not set
* Allows specifying unsigned numeric values as samples per second
* Allows "unlimited" value to always sample

### DIFF
--- a/libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
+++ b/libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
@@ -22,13 +22,16 @@ import zipkin2.reporter.AsyncReporter
 import zipkin2.reporter.Reporter
 import zipkin2.reporter.brave.ZipkinSpanHandler
 import zipkin2.reporter.urlconnection.URLConnectionSender
-import java.util.OptionalInt
 import java.util.Stack
 import java.util.concurrent.ExecutorService
 import javax.servlet.Filter
 
+sealed interface SampleRate
+object Unlimited : SampleRate
+data class PerSecond(val samplesPerSecond: Int) : SampleRate
+
 @Suppress("TooManyFunctions")
-class BraveTracingService(serviceName: String, zipkinHost: String, samplesPerSecond: OptionalInt) : TracingService {
+class BraveTracingService(serviceName: String, zipkinHost: String, samplesPerSecond: SampleRate) : TracingService {
 
     private val resourcesToClose = Stack<AutoCloseable>()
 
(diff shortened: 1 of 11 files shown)


libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
```diff
@@ -24,8 +25,13 @@ import zipkin2.reporter.urlconnection.URLConnectionSender
 import java.util.Stack
 import java.util.concurrent.ExecutorService
 import javax.servlet.Filter
+
+sealed interface SampleRate
+object Unlimited : SampleRate
+data class PerSecond(val samplesPerSecond: Int) : SampleRate
```
vkolomeyko:
```suggestion
internal sealed interface SampleRate
private object Unlimited : SampleRate
private data class PerSecond(val samplesPerSecond: Int) : SampleRate
```

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
```diff
@@ -24,8 +25,13 @@ import zipkin2.reporter.urlconnection.URLConnectionSender
...
+
 @Suppress("TooManyFunctions")
-class BraveTracingService(serviceName: String, zipkinHost: String) : TracingService {
+class BraveTracingService(serviceName: String, zipkinHost: String, samplesPerSecond: SampleRate) : TracingService {
```
vkolomeyko:
```suggestion
internal class BraveTracingService(serviceName: String, zipkinHost: String, samplesPerSecond: SampleRate) : TracingService {
```

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
```diff
@@ -35,12 +41,17 @@ class BraveTracingService(serviceName: String, zipkinHost: String) : TracingServ
             .addScopeDecorator(MDCScopeDecorator.get())
             .build()
 
+        val sampler = when (samplesPerSecond) {
+            is PerSecond -> RateLimitingSampler.create(samplesPerSecond.samplesPerSecond)
+            is Unlimited -> Sampler.ALWAYS_SAMPLE
+        }
+
```
vkolomeyko, line 57: I would add an `INFO` log line informing how frequently we will be collecting tracing samples.

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
josephzunigadaly: Made them all internal for access from `TraceThis.kt`

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
josephzunigadaly: Done

vkolomeyko (approved):
LGTM

fenryka (approved):
LGTM

driessamyn:
Should this not be a dynamic config that can be set through the config API ?

(not suggesting we undo this change, but wondering it that needs to be a follow-up JIRA?)

vkolomeyko:
> Should this not be a dynamic config that can be set through the config API ?
> 
> (not suggesting we undo this change, but wondering it that needs to be a follow-up JIRA?)

IMO that would be un-necessary and overcomplicated as every worker would need to listen to config update and change tracing sampling in response.
If C5 cluster operator wants to change sampling rate, then can always modify value in Helm chart and update their deployment.

PR #4024, opened 2023-06-06
CORE-7032 Use batch send and receive in receive and finality flow

nkovacsx:
### Overview

Use batch sending and receiving in finality and receive finality flows. We used to send data one-by-one to counterparties but we can replace these with batch sending.

Performance testing: https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4504518727/Test+6+Use+batch+send+receive

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
@@ -92,16 +92,11 @@ class UtxoFinalityFlowV1(
     @Suppress("MaxLineLength")
     @Suspendable
     private fun receiveSignaturesAndAddToTransaction(): Pair<UtxoSignedTransactionInternal, Map<FlowSession, List<DigitalSignatureAndMetadata>>> {
-        val signaturesPayloads = sessions.associateWith { session ->
-            try {
-                log.debug { "Requesting signatures from ${session.counterparty} for transaction $transactionId" }
-                @Suppress("unchecked_cast")
-                session.receive(Payload::class.java) as Payload<List<DigitalSignatureAndMetadata>>
-            } catch (e: CordaRuntimeException) {
-                log.warn("Failed to receive signatures from ${session.counterparty} for transaction $transactionId")
-                persistInvalidTransaction(initialTransaction)
-                throw e
-            }
+        val signaturesPayloads = flowMessaging.receiveAllMap(
+            sessions.associateWith { Payload::class.java }
+        ).mapValues {
+            @Suppress("unchecked_cast")
+            it.value as Payload<List<DigitalSignatureAndMetadata>>
         }
 
         var transaction = initialTransaction
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
@@ -77,8 +77,8 @@ class UtxoFinalityFlowV1(
 
     @Suspendable
     private fun sendTransactionAndBackchainToCounterparties() {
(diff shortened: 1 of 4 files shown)


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -77,8 +77,8 @@ class UtxoFinalityFlowV1(
 
     @Suspendable
     private fun sendTransactionAndBackchainToCounterparties() {
+        flowMessaging.sendAll(initialTransaction, sessions.toSet())
```
mnesbit, line 80: Do we correctly handle empty session set? Perhaps just add an isEmpty check around this

rick-r3 (approved):

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -92,16 +92,17 @@ class UtxoFinalityFlowV1(
...
+            ).mapValues {
                 @Suppress("unchecked_cast")
-                session.receive(Payload::class.java) as Payload<List<DigitalSignatureAndMetadata>>
-            } catch (e: CordaRuntimeException) {
-                log.warn("Failed to receive signatures from ${session.counterparty} for transaction $transactionId")
-                persistInvalidTransaction(initialTransaction)
-                throw e
+                it.value as Payload<List<DigitalSignatureAndMetadata>>
             }
+        } catch (e: CordaRuntimeException) {
```
mnesbit, line 102: There are possibly some non-CordaRuntimeExceptions possible here e.g. the cast could fail. So not sure if catch is too narrow, though the re-throw really needs to be a CordaRuntimeException. Probably not to be for tonight, but needs a JIRA to revisit

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
nkovacsx: `sendAll` will not do anything if the sessions set is empty:
```
if (sessions.isEmpty()) {
            return
        }
```
Same with `toSet`, it's not sensitive to empty collections.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
nkovacsx: The logic hasn't changed from the previous impl, this is how it was before. Probably need to reconsider.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
mnesbit: OK

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
mnesbit: revisit later

mnesbit (approved):
OK

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
nkovacsx: Raised https://r3-cev.atlassian.net/browse/CORE-14402

PR #4026, opened 2023-06-06
CORE-14383 - Doubled the maximum memory that a Kafka pod can use

filipesoliveira:


applications/tools/p2p-test/app-simulator/scripts/prereqs-eks.yaml
```diff
@@ -4,7 +4,7 @@ kafka:
       memory: 390Mi
       cpu: 1000m
     limits:
-      memory: 800Mi
```
filipesoliveira, line 7: Do we want to also increase the max. memory for the app-simulator?

.ci/e2eTests/prereqs-eks.yaml
```diff
@@ -11,7 +11,7 @@ kafka:
       memory: 390Mi
```
vkolomeyko:
```suggestion
      memory: 780Mi
```

applications/tools/p2p-test/app-simulator/scripts/prereqs-eks.yaml
```diff
@@ -4,7 +4,7 @@ kafka:
       memory: 390Mi
```
vkolomeyko:
```suggestion
      memory: 780Mi
```

applications/tools/p2p-test/app-simulator/scripts/prereqs-eks.yaml
dimosr: yes I think it's worth keeping it in line with CI (at least the defaults) as there were cases in the past where we were trying to reproduce issues manifesting in CI and we couldn't because issues were happening only in under-resourced environments

dimosr (dismissed):
LGTM

vkolomeyko (approved):
LGTM

dimosr (approved):

dimosr:
wondering - should we merge this change in 5.0 instead and bring it in 5.1 as part of a forward merge?

Suspecting the issues we were seeing will be present in 5.0 too.

vkolomeyko:
> wondering - should we merge this change in 5.0 instead and bring it in 5.1 as part of a forward merge?
> 
> Suspecting the issues we were seeing will be present in 5.0 too.

I guess we can merge it in `5.0` but it will involve more scrutiny as for:
- Is it fixing a critical bug?
- Is it improving performance?

And only limited set of people can approve anything into `5.0` at the moment.

I would merge it to `5.1` for now. If this will be a re-occurring pattern in `5.0` we would be able to backport this change there as well.

dimosr:
that sounds reasonable 👍

wzur-r3 (approved):
LGTM

PR #4034, opened 2023-06-07
ES-582 Prefer worker Kafka values

davidcurrie:
The check should prefer worker-specific Kafka values over the default values, even if the worker uses `value` and the defaults use `valueFrom` to replicate the behavior in the Helm chart.


jujoramos (approved):
LGTM

PR #4038, opened 2023-06-07
CORE-14387 Added zipkin and io.micrometer to the quasar exclude list.

owenstanford:
Excludes io.zipkin and io.micrometer to the quasar exclude list to ensure they are not instrumented by quasar


vkolomeyko (approved):

driessamyn (approved):
LGTM

PR #4035, opened 2023-06-07
CORE-13796: Merge 5.0 into 5.1

josephzunigadaly:
Reminder: No squashing when integrating this PR

PR #4039, opened 2023-06-07
CORE-13862: Fix timer resolution issues found during forward merge

josephzunigadaly:
This PR is a response to this feedback:
https://github.com/corda/corda-runtime-os/pull/4035#pullrequestreview-1468000050

New PR because we need to apply the fixes to the 5.0 branch, the PR linked above targets 5.1.

### DIFF
--- a/components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
+++ b/components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
@@ -74,19 +74,19 @@ internal class ReconcilerEventHandler<K : Any, V : Any>(
     private fun reconcileAndScheduleNext(coordinator: LifecycleCoordinator) {
         logger.info("Initiating reconciliation")
         var reconciliationOutcome = "FAILED"
-        val startTime = System.currentTimeMillis()
+        val startTime = System.nanoTime()
         var reconciliationRunTime = startTime
         var reconciledCount = 0
         try {
             reconciledCount = reconcile()
             reconciliationOutcome = "SUCCEEDED"
-            reconciliationRunTime = System.currentTimeMillis() - startTime
+            reconciliationRunTime = System.nanoTime() - startTime
             logger.info("Reconciliation completed in $reconciliationRunTime ms")
             scheduleNextReconciliation(coordinator)
         } catch (e: Throwable) {
             // An error here could be a transient or not exception. We should transition to `DOWN` and wait
             // on subsequent `RegistrationStatusChangeEvent` to see if it is going to be a `DOWN` or an `ERROR`.
-            reconciliationRunTime = System.currentTimeMillis() - startTime
+            reconciliationRunTime = System.nanoTime() - startTime
             logger.warn("Reconciliation failed. Terminating reconciliations", e)
             coordinator.updateStatus(LifecycleStatus.DOWN)
         } finally {
@@ -94,7 +94,7 @@ internal class ReconcilerEventHandler<K : Any, V : Any>(
                 .withTag(CordaMetrics.Tag.OperationName, name)
                 .withTag(CordaMetrics.Tag.OperationStatus, reconciliationOutcome)
                 .build()
-                .record(Duration.ofMillis(reconciliationRunTime))
+                .record(Duration.ofNanos(reconciliationRunTime))
 
             CordaMetrics.Metric.Db.ReconciliationRecordsCount.builder()
                 .withTag(CordaMetrics.Tag.OperationName, name)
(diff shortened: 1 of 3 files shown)


components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
```diff
@@ -74,27 +74,27 @@ internal class ReconcilerEventHandler<K : Any, V : Any>(
     private fun reconcileAndScheduleNext(coordinator: LifecycleCoordinator) {
         logger.info("Initiating reconciliation")
         var reconciliationOutcome = "FAILED"
-        val startTime = System.currentTimeMillis()
+        val startTime = System.nanoTime()
         var reconciliationRunTime = startTime
```
chrisr3: Perhaps `reconciliationEndTime` rather than `reconciliationRunTime`? Then you could calculate `reconciliationRunTime` once inside the `finally` block.

Is this legal?
```kotlin
val reconciliationEndTime: Long
var reconciledCount = 0
try {
    try {
        reconciledCount = reconcile()
    } finally {
        reconciliationEndTime = System.nanoTime()
    }
    reconciliationOutcome = "SUCCEEDED"
    ...
```
I think recent versions of Kotlin allow you to declare a `var` without an initialiser _provided_ the code _guarantees_ it to be initialised before it is used.

josephzunigadaly: It's legal except in this case the finally counts as possibly not initialised.

josephzunigadaly: this is done:

>Perhaps reconciliationEndTime rather than reconciliationRunTime? Then you could calculate reconciliationRunTime once inside the finally block.

components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
```diff
@@ -74,27 +74,27 @@ internal class ReconcilerEventHandler<K : Any, V : Any>(
...
+        val startTime = System.nanoTime()
+        var reconciliationEndTime = startTime
         var reconciledCount = 0
         try {
             reconciledCount = reconcile()
             reconciliationOutcome = "SUCCEEDED"
-            reconciliationRunTime = System.currentTimeMillis() - startTime
-            logger.info("Reconciliation completed in $reconciliationRunTime ms")
+            reconciliationEndTime = System.nanoTime()
+            logger.info("Reconciliation completed in ${reconciliationEndTime - startTime} ns")
```
chrisr3: This isn't a very useful message - humans don't think of time in nanoseconds :wink:. Perhaps log both `reconciliationOutcome` and `reconciliationRunTime` (in ms) in the `finally` block?

To be fair, I'm not convinced `System.nanoTime()` is _accurate_ to the nanosecond either!

josephzunigadaly: I'll do that

josephzunigadaly: Done

components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
```diff
@@ -74,27 +74,28 @@ internal class ReconcilerEventHandler<K : Any, V : Any>(
...
         } catch (e: Throwable) {
             // An error here could be a transient or not exception. We should transition to `DOWN` and wait
             // on subsequent `RegistrationStatusChangeEvent` to see if it is going to be a `DOWN` or an `ERROR`.
-            reconciliationRunTime = System.currentTimeMillis() - startTime
+            reconciliationEndTime = System.nanoTime()
             logger.warn("Reconciliation failed. Terminating reconciliations", e)
             coordinator.updateStatus(LifecycleStatus.DOWN)
         } finally {
+            val reconciliationTime = Duration.ofNanos(reconciliationEndTime - startTime)
+            logger.info("Reconciliation completed in ${reconciliationTime.toMillis()} ms")
```
kyriathar: I think that was meant to mark (successfully) completed. Now it will log in any case. Perhaps it would make sense now, to add in the logged string the `reconciliationOutcome` as well?

kyriathar: Although we do (warn) log in case of a failure above but still maybe more clear.

josephzunigadaly: Like this?

```kotlin
logger.info("Reconciliation $reconciliationOutcome in ${reconciliationTime.toMillis()} ms")
```

josephzunigadaly: `reconciliationOutcome` expands to SUCCEEDED/FAILED

josephzunigadaly: Done

kyriathar (approved):

chrisr3 (approved):
LGTM

driessamyn (approved):
LGTM

PR #4040, opened 2023-06-07
CORE-14425: Create JUnit extensions for testing with CordaMetrics.

chrisr3:
- Create JUnit extensions to integrate managing `CordaMetrics`  with the JUnit lifecycle.
- Use JUnit `@ResourceLock` to ensure each test has exclusive access to the global `MeterRegistry`. (`CordaMetrics` _own_ tests cannot use this extension themselves without creating a circular dependency.)
- Also tweak `SandboxGroupContextCacheTest` in attempt to resolve flaky behaviour.

---
`@ResourceLock` is only effective when used with JUnit's built-in parallel execution mode. Gradle's `--parallel` option appears to work by dividing the work-load up among separate "worker" JVMs.


driessamyn (approved):
LGTM

PR #4042, opened 2023-06-08
CORE-14321: Reduce logging at debug level in state and event pattern

JamesHR3:
At debug level the logs become swamped by a few state and event pattern logs in a tight loop. This PR removes one of these and reduces the levels for another to try and improve debuggability.


driessamyn (approved):
LGTM

PR #4047, opened 2023-06-08
CORE-14335: make SigningKeyInfo.keyMaterial non-nullable

dickon:
SiginingKeyInfo.keyMaterial used to be nullable, in order to support HSMs where we do not have the key material (i.e. we do not have access to an encrypted private key). Tlet's remove support for missing key material in signing key records since it can never happen in production right now with our current feature set. This is a step towards removing the HSM layers in order to squash together code in order to improve performance,

### DIFF
--- a/components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
+++ b/components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
@@ -14,7 +14,6 @@ import net.corda.crypto.core.publicKeyIdFromBytes
 import net.corda.crypto.persistence.SigningKeyInfo
 import net.corda.crypto.persistence.SigningKeyOrderBy
 import net.corda.crypto.persistence.SigningKeyStatus
-import net.corda.crypto.persistence.SigningPublicKeySaveContext
 import net.corda.crypto.persistence.SigningWrappedKeySaveContext
 import net.corda.crypto.persistence.WrappingKeyInfo
 import net.corda.crypto.persistence.db.model.WrappingKeyEntity
@@ -54,7 +53,8 @@ class SigningRepositoryTest : CryptoRepositoryTest() {
     private val defaultTenantId = "Memento Mori"
     private val defaultMasterKeyName = "Domination's the name of the game"
 
-    private fun createNewPubKeyInfo(): SigningKeyInfo {
+    private fun createPrivateKeyInfo(): SigningKeyInfo {
+        val privKey = SecureHashUtils.randomBytes()
         val unique = UUID.randomUUID().toString()
         val key = SecureHashUtils.randomBytes()
         val keyId = publicKeyIdFromBytes(key)
@@ -65,25 +65,15 @@ class SigningRepositoryTest : CryptoRepositoryTest() {
             tenantId = "t-$unique".take(12),
             category = "c-$unique",
             alias = "a-$unique",
-            hsmAlias = "ha-$unique",
+            hsmAlias = null,
             publicKey = key,
-            keyMaterial = null,
             schemeCodeName = "FOO",
-            masterKeyAlias = null,
             externalId = "e-$unique",
-            encodingVersion = null,
             timestamp = Instant.now().toSafeWindowsPrecision(),
             hsmId = "hi-$unique".take(36),
-            status = SigningKeyStatus.NORMAL
-        )
-    }
-
-    private fun createPrivateKeyInfo(): SigningKeyInfo {
-        val privKey = SecureHashUtils.randomBytes()
-        return createNewPubKeyInfo().copy(
+            status = SigningKeyStatus.NORMAL,
             keyMaterial = privKey,
             encodingVersion = 1,
-            hsmAlias = null,
             masterKeyAlias = defaultMasterKeyName
         )
     }
@@ -99,7 +89,7 @@ class SigningRepositoryTest : CryptoRepositoryTest() {
         val pubKey = mock<PublicKey> {
             on { encoded } doReturn(info.publicKey)
         }
-        return GeneratedWrappedKey(pubKey, info.keyMaterial!!, info.encodingVersion!!)
+        return GeneratedWrappedKey(pubKey, info.keyMaterial, info.encodingVersion!!)
     }
 
     private fun createKeyScheme(info: SigningKeyInfo) =
@@ -164,24 +154,8 @@ class SigningRepositoryTest : CryptoRepositoryTest() {
             repo.savePrivateKey(ctx)
         }
 
-        val publicKeys = (0..3).map { _ ->
-            val info = createNewPubKeyInfo()
-            val ctx = createSigningPublicKeySaveContext(info)
-            repo.savePublicKey(ctx)
-        }
-        return privateKeys + publicKeys
-    }
 
-    fun createSigningPublicKeySaveContext(info: SigningKeyInfo) : SigningPublicKeySaveContext {
-        val pubKey = createGeneratedPublicKey(info)
-        return SigningPublicKeySaveContext(
-            key = pubKey,
-            alias = info.alias,
-            category = info.category,
-            keyScheme = createKeyScheme(info),
-            externalId = info.externalId,
-            hsmId = info.hsmId,
-        )
+        return privateKeys
     }
 
     fun createSigningWrappedKeySaveContext(info: SigningKeyInfo) : SigningWrappedKeySaveContext {
@@ -197,49 +171,6 @@ class SigningRepositoryTest : CryptoRepositoryTest() {
         )
     }
 
-    @ParameterizedTest
-    @MethodSource("emfs")
-    fun `savePublicKey and find`(emf: EntityManagerFactory) {
-        val info = createNewPubKeyInfo()
-        val pubKey = createGeneratedPublicKey(info)
-        val ctx = createSigningPublicKeySaveContext(info)
-
-        val repo = SigningRepositoryImpl(
-            emf,
-            info.tenantId,
-            cipherSchemeMetadata,
-            digestService,
-            createLayeredPropertyMapFactory(),
-        )
-
-        repo.savePublicKey(ctx)
-
-        val found = repo.findKey(pubKey.publicKey)
-
-        assertThat(found)
-            .usingRecursiveComparison().ignoringFields("timestamp")
-            .isEqualTo(info)
-    }
-
-    @ParameterizedTest
-    @MethodSource("emfs")
-    fun `savePublicKey twice fails`(emf: EntityManagerFactory) {
-        val info = createNewPubKeyInfo()
-        val ctx = createSigningPublicKeySaveContext(info)
-
-        val repo = SigningRepositoryImpl(
-            emf,
-            info.tenantId,
-            cipherSchemeMetadata,
-            digestService,
-            createLayeredPropertyMapFactory(),
-        )
-
-        repo.savePublicKey(ctx)
-        assertThrows<PersistenceException> {
-            repo.savePublicKey(ctx)
-        }
-    }
 
     @ParameterizedTest
     @MethodSource("emfs")
--- a/components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
+++ b/components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
@@ -53,7 +53,7 @@ class SigningRepositoryTest : CryptoRepositoryTest() {
     private val defaultTenantId = "Memento Mori"
     private val defaultMasterKeyName = "Domination's the name of the game"
 
-    private fun createPrivateKeyInfo(): SigningKeyInfo {
+    private fun createSigningKeyInfo(): SigningKeyInfo {
         val privKey = SecureHashUtils.randomBytes()
         val unique = UUID.randomUUID().toString()
         val key = SecureHashUtils.randomBytes()
@@ -149,7 +149,7 @@ class SigningRepositoryTest : CryptoRepositoryTest() {
 
         saveWrappingKey(emf, defaultMasterKeyName)
         val privateKeys = (0..2).map { _ ->
-            val info = createPrivateKeyInfo()
+            val info = createSigningKeyInfo()
             val ctx = createSigningWrappedKeySaveContext(info)
             repo.savePrivateKey(ctx)
         }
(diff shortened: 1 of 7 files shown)


libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/SigningAliasSpec.kt
```diff
@@ -4,6 +4,8 @@ import net.corda.crypto.cipher.suite.schemes.KeyScheme
 import net.corda.v5.crypto.SignatureSpec
 import java.security.PublicKey
 
+// TODO remove
+
```
kyriathar: Let's please remove this TODO

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/SigningAliasSpec.kt
```diff
@@ -19,7 +21,7 @@ class SigningAliasSpec(
     val hsmAlias: String,
     override val publicKey: PublicKey,
     override val keyScheme: KeyScheme,
-    override val signatureSpec: SignatureSpec
+    override val signatureSpec: SignatureSpec,
```
kyriathar: Can we please revert this unneeded change

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/SharedSecretAliasSpec.kt
```diff
@@ -3,6 +3,7 @@ package net.corda.crypto.cipher.suite
 import net.corda.crypto.cipher.suite.schemes.KeyScheme
 import java.security.PublicKey
 
+// TODO remove
```
kyriathar: Let's please remove this TODO

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
```diff
@@ -268,12 +268,11 @@ fun SigningKeyEntity.joinSigningKeyInfo(em: EntityManager): SigningKeyInfo {
...
-    val wrappingKey = if (signingKeyMaterialEntity != null) {
-        em.createQuery(
-            "FROM WrappingKeyEntity WHERE id=:wrappingKeyId", WrappingKeyEntity::class.java,
-        ).setParameter("wrappingKeyId", signingKeyMaterialEntity.wrappingKeyId).resultList.singleOrNull()
-    } else null
+        .resultList.singleOrNull() ?: throw InvalidObjectException("private key material for $id not found")
+    val wrappingKey = em.createQuery(
+        "FROM WrappingKeyEntity WHERE id=:wrappingKeyId", WrappingKeyEntity::class.java
+    ).setParameter("wrappingKeyId", signingKeyMaterialEntity.wrappingKeyId).resultList.singleOrNull()
+
```
kyriathar: Not really sure why we would had to change this bit of code? Nothing changed at entities level I believe.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
```diff
@@ -268,12 +268,11 @@ fun SigningKeyEntity.joinSigningKeyInfo(em: EntityManager): SigningKeyInfo {
         "FROM ${SigningKeyMaterialEntity::class.java.simpleName} WHERE signingKeyId=:signingKeyId",
         SigningKeyMaterialEntity::class.java
     ).setParameter("signingKeyId", id)
-        .resultList.singleOrNull()
-    val wrappingKey = if (signingKeyMaterialEntity != null) {
-        em.createQuery(
-            "FROM WrappingKeyEntity WHERE id=:wrappingKeyId", WrappingKeyEntity::class.java,
-        ).setParameter("wrappingKeyId", signingKeyMaterialEntity.wrappingKeyId).resultList.singleOrNull()
-    } else null
+        .resultList.singleOrNull() ?: throw InvalidObjectException("private key material for $id not found")
```
kyriathar: Not really sure if we should be using this exception type?

components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
```diff
@@ -54,7 +53,8 @@ class SigningRepositoryTest : CryptoRepositoryTest() {
     private val defaultTenantId = "Memento Mori"
     private val defaultMasterKeyName = "Domination's the name of the game"
 
-    private fun createNewPubKeyInfo(): SigningKeyInfo {
+    private fun createPrivateKeyInfo(): SigningKeyInfo {
```
kyriathar: Maybe call this `createSigningKeyInfo`

kyriathar (commented):
A few comments.

I believe we should also be removing `GeneratedPublicKey` and `SigningPublicKeySaveContext` which are types to hold signing keys whose private parts were meant to be stored into hardware HSMs?

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
Sakpal: @kyriathar Can you please provide a little context on why shouldn't we be using this Exception?

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/SigningAliasSpec.kt
dickon: This is a coding standard question. I've been putting them in since @fowlerrr regarded having them as part of the r3 coding standard.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
Sakpal: My understanding is previously we were doing a NULL check  `(signingKeyMaterialEntity != null)`  and if `signingKeyMaterialEntity` was NULL, we were assigning `wrappingKey` as NULL and continue our execution.

With newer changes, we are halting/breaking the execution of this function call.

@dickon a couple of questions:
- Are we okay with the execution flow of breaking this function call if  `signingKeyMaterialEntity` is NULL? Are you planning to implement a strict flow break if these details are missing as per requirement?
-  Have we handled this exception at the caller side of this method? I am guessing this Is something which will come up in following PR's?

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/SharedSecretAliasSpec.kt
dickon: Later in the patch series I remove this class. I think leaving TODOs in code is reasonable, and an important part of taking small steps. Alternatively we could use JIRAs but for tiny things like this it seems over the top.

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/SigningAliasSpec.kt
dickon: see above

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
dickon: The goal of this change is to make `SigningKeyInfo.keyMaterial` non-null. In situations where this routines fail to find key material in the database, we used to return a null in `SigningKeyInfo.keyMaterial`; we can't do that any more, so we throw an exception in such cases. Now we are removing external HSM support it should never happen, so an exception feels right.

components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
dickon: It was called `createPrivateKeyInfo` before, I haven't introduced the name. But I do think I prefer `createSigningKeyInfo`. So maybe it is worth the time to go back and change the name. Waiting to see what the decision is on the other conversations on this PR.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
dickon: https://docs.oracle.com/javase/8/docs/api/java/io/InvalidObjectException.html says:

says "Indicates that one or more deserialized objects failed validation tests. The argument should provide the reason for the failure." 

... which seems to be the closest of the java standard  exceptions to what we are hitting here, which is that we've got some data from the database and it doesn't validate against our object model. 

I understand the Corda 5 architecutral standard for exceptions is to use the most appropriate standard exceptions in cases like this.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
kyriathar: So that suggests erroneous behavior before. 

It's a different thing finding/ not finding the `SigningKeyMaterialEntity` and a different thing the `SigningKeyMaterialEntity.keyMaterial` to be nullable/ not nullable I believe. Or maybe we mean nullability in the DB by finding/ not finding the record.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
kyriathar: A `requireNotNull` should be ok I think. `InvalidObjectException` seems to be related to deserialization whereas here we need something to indicate not found (in the DB) I believe.

chrisr3: `InvalidObjectException` relates to Java's _own_ serialization though, doesn't it? It doesn't appear to be relevant here. We're not even talking about AMQP!

fowlerrr: `IllegalArgumentException` seems correct to me.  That or a custom one.  But, what we're saying at this point, is that the `id` wasn't found.  That's typically an illegal argument in Java, no?

dickon: @fowlerrr maybe, though It's not so much that the argument is invalid, more what's in the database is unexpected. I can live with that though so I'll make the change.

components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
```diff
@@ -164,24 +154,8 @@ class SigningRepositoryTest : CryptoRepositoryTest() {
...
 
-        val publicKeys = (0..3).map { _ ->
-            val info = createNewPubKeyInfo()
-            val ctx = createSigningPublicKeySaveContext(info)
-            repo.savePublicKey(ctx)
-        }
-        return privateKeys + publicKeys
-    }
 
-    fun createSigningPublicKeySaveContext(info: SigningKeyInfo) : SigningPublicKeySaveContext {
```
Sakpal, line 175: Nice work here @dickon for getting rid of unused code which had tests as well, reducing overall effort to manage it further 🙌

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
Sakpal: Nice work here @dickon  to remove the unpredictability from our old code around availability of Keys through HSM 🙌

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/SigningAliasSpec.kt
chrisr3: Gee, I "love" it when coding standards are agreed in secret by a small group of people behind everyone else's backs :no_mouth:.

This extra comma is unnecessary.

components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
dickon: done

components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
dickon: Thanks

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
dickon: I went with `requiteNotNull`, hence throwing `IllegalArgumentException`.

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/SharedSecretAliasSpec.kt
dickon: Clarified.

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/SigningAliasSpec.kt
dickon: Clarified.

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/SigningAliasSpec.kt
dickon: Reverted.

kyriathar (approved):

Sakpal (commented):
LGTM 👍

PR #4048, opened 2023-06-08
POC - se tracing

driessamyn:


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
```diff
@@ -185,7 +192,60 @@ internal class StateAndEventSubscriptionImpl<K : Any, S : Any, E : Any>(
...
+
+        fun recordCommitTx() {
+            commitDurationMs = timeToMillis()
+        }
+
+        fun recordTotal() {
+            totalDuractionMs = timeToMillis()
+        }
+
+        private fun timeToMillis(): Long {
```
owenstanford: Maybe double to avoid rounding sub millisecond metrics

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
```diff
@@ -262,8 +355,12 @@ internal class StateAndEventSubscriptionImpl<K : Any, S : Any, E : Any>(
                 })
                 deadLetterRecords.clear()
             }
+            log.info("DLQ: ${(System.nanoTime()-now).div(1000000.toDouble())}")
```
driessamyn, line 358: FYI @Omar-awad 
I updated this to include the DLQ block you were concerned was skewing the time measurements.
From my test just now, it looks like that generally contributes to <1ms (as expected).

driessamyn:
FYI @Omar-awad , @JamesHR3 
This may be useful when validating the metrics. I don't intend to merge this, but it was useful for analysing the timings from the trace.

I used the below Jupyter notebook to scrape the results from the log into a spreadsheet.

```python
import glob
import re

pat = re.compile(r'\[(.*)\/(.*)\] Batch Size: (\d*), UpdateStates: (.*)ms, Process: (.*)ms, Begin: (.*)ms, Send: (.*)ms, DLQ: (.*)ms, Send Offsets: (.*)ms, Commit: (.*)ms, Total:(.*)ms, Avg: (.*)ms, Max:(.*)ms')

for file in glob.glob(log_dir):
    logs = [ { 
        "state": l.group(1),
        "event": l.group(2),
        "batch": l.group(3),
        "update-states": float(l.group(4)),
        "process": float(l.group(5)),
        "begin": float(l.group(6)),
        "send": float(l.group(7)),
        "dlq": float(l.group(8)),
        "send-offset": float(l.group(9)),
        "commit": float(l.group(10)),
        "total": float(l.group(11)),
        "avg": float(l.group(12)),
        "max": float(l.group(13)),
        } for l in [pat.search(line) for line in open(file)] if l is not None ]
    
```

```python
import pandas as pd

df = pd.DataFrame(columns=["state","event","batch","process","begin","send","send-offset","commit","total","avg","max"])

entry = pd.DataFrame.from_dict(logs)

df = pd.concat([df, entry], ignore_index=True)

df
```

```python
with pd.ExcelWriter('se-trace_single_trace.xlsx') as writer: 
    df.to_excel(writer)
```

PR #4050, opened 2023-06-09
ES-89: removed obsolete parameter of shared pipeline

wzur-r3:
* removed obsolete `nexusAppId` parameter
* we no longer run checks with Nexus and the functionality is gone from shared pipelines

This is just a cosmetic change and it does not modify any behaviour.


yarivus (approved):
LGTM

ronanbrowne (approved):

PR #4052, opened 2023-06-09
CORE-11037 Read extra properties from DbConfig

kyriathar:
* Reads the following datasource properties from configuration on datasources creation:

  *  idle timeout
  *  max lifetime
  *  keepalive time
  *  validation timeout

* Adds the above datasource properties to VNode datasource configuration with hardcoded values

* Makes datasource parameters not optional now that they are always passed in, in prod path(s)

### DIFF
--- a/components/db/db-connection-manager/src/main/kotlin/net/corda/db/connection/manager/DbConfig.kt
+++ b/components/db/db-connection-manager/src/main/kotlin/net/corda/db/connection/manager/DbConfig.kt
@@ -86,6 +86,10 @@ fun createDbConfig(
     jdbcUrl: String? = null,
     maxPoolSize: Int? = null,
     minPoolSize: Int? = null,
+    idleTimeout: Int,
+    maxLifetime: Int,
+    keepaliveTime: Int,
+    validationTimeout: Int,
     key: String = "database-password"
 ): SmartConfig {
     var config =
@@ -99,6 +103,11 @@ fun createDbConfig(
         config = config.withValue(DatabaseConfig.DB_POOL_MAX_SIZE, ConfigValueFactory.fromAnyRef(maxPoolSize))
     if(null != minPoolSize)
         config = config.withValue(DatabaseConfig.DB_POOL_MIN_SIZE, ConfigValueFactory.fromAnyRef(minPoolSize))
+
+    config = config.withValue(DatabaseConfig.DB_POOL_IDLE_TIMEOUT, ConfigValueFactory.fromAnyRef(idleTimeout))
+    config = config.withValue(DatabaseConfig.DB_POOL_MAX_LIFETIME, ConfigValueFactory.fromAnyRef(maxLifetime))
+    config = config.withValue(DatabaseConfig.DB_POOL_KEEP_ALIVE_TIME, ConfigValueFactory.fromAnyRef(keepaliveTime))
+    config = config.withValue(DatabaseConfig.DB_POOL_VALIDATION_TIMEOUT, ConfigValueFactory.fromAnyRef(validationTimeout))
     return config
 
 }
--- a/components/db/db-connection-manager/src/main/kotlin/net/corda/db/connection/manager/DbConfig.kt
+++ b/components/db/db-connection-manager/src/main/kotlin/net/corda/db/connection/manager/DbConfig.kt
@@ -11,6 +11,7 @@ import net.corda.libs.configuration.validation.getConfigurationDefaults
 import net.corda.schema.configuration.ConfigKeys
 import net.corda.schema.configuration.DatabaseConfig
 import org.slf4j.LoggerFactory
+import java.time.Duration
 import javax.sql.DataSource
 
 internal object DbConfig {
@@ -39,6 +40,11 @@ fun DataSourceFactory.createFromConfig(config: SmartConfig): CloseableDataSource
     } else {
         null
     }
+    val idleTimeout = configWithFallback.getInt(DatabaseConfig.DB_POOL_IDLE_TIMEOUT).toLong().run(Duration::ofSeconds)
+    val maxLifetime = configWithFallback.getInt(DatabaseConfig.DB_POOL_MAX_LIFETIME).toLong().run(Duration::ofSeconds)
+    val keepaliveTime = configWithFallback.getInt(DatabaseConfig.DB_POOL_KEEP_ALIVE_TIME).toLong().run(Duration::ofSeconds)
+    val validationTimeout =
+        configWithFallback.getInt(DatabaseConfig.DB_POOL_VALIDATION_TIMEOUT).toLong().run(Duration::ofSeconds)
 
     val username = if (configWithFallback.hasPath(DatabaseConfig.DB_USER)) configWithFallback.getString(DatabaseConfig.DB_USER) else
         throw DBConfigurationException(
@@ -59,6 +65,10 @@ fun DataSourceFactory.createFromConfig(config: SmartConfig): CloseableDataSource
         password = password,
         maximumPoolSize = maxPoolSize,
         minimumPoolSize = minPoolSize,
+        idleTimeout = idleTimeout,
+        maxLifetime = maxLifetime,
+        keepaliveTime = keepaliveTime,
+        validationTimeout = validationTimeout
     )
 }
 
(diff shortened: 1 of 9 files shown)


components/db/db-connection-manager/src/main/kotlin/net/corda/db/connection/manager/DbConfig.kt
```diff
@@ -31,6 +32,8 @@ fun DataSourceFactory.createFromConfig(config: SmartConfig): CloseableDataSource
     DbConfig.log.debug("Given configuration: ${config.toSafeConfig().root().render(ConfigRenderOptions.concise())}")
     DbConfig.log.debug("Fallback configuration: ${dbFallbackConfig.root().render(ConfigRenderOptions.concise())}")
 
+    // TODO all the following configuration reading should probably be changed to mandatory (i.e. throw if not found),
```
simon-johnson-r3: if this is true right now, we should just do it and remove the TODO

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDbFactoryImpl.kt
```diff
@@ -169,7 +169,9 @@ internal class VirtualNodeDbFactoryImpl(
 
             // TODO support for CharArray passwords in SmartConfig
             val config = createDbConfig(
```
simon-johnson-r3, line 175: Here you will need to create a config with all the new values. For now you'll need to hard code the values, like we do at line 30. At some point we'll need to move all those hard coded values elsewhere. The actual values should be preserved from what we have now. We might need to double check them with @driessamyn too (the actual values I mean).

simon-johnson-r3 (changes requested):

components/db/db-connection-manager/src/main/kotlin/net/corda/db/connection/manager/DbConfig.kt
kyriathar: Done

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDbFactoryImpl.kt
kyriathar: Done I think. I can see we treat as optional the below parameters within `fun createDbConfig`: 
* `jdbcUrl`
* `maxPoolSize`
* `minPoolSize`
* `key`

However, we are always passing in arguments for those at call site. Should we make them non optional? (Some of them are nullable while we always passing non values too, so nullable could be removed as well for those I think)

simon-johnson-r3: Yes I think so. `minPoolSize` can be nullable all the way to the HikariCP API, but I'm not sure that's a good idea as it leaks details of that through the abstraction. I assume by the time we call `createDbConfig` we already have a final version of `maxPoolSize` from whatever config and as such we can always set `minPoolSize` ourselves if the config value is `null`.

kyriathar: Removed unneeded optional and nullable which are not needed as per the call site.

components/db/db-connection-manager/src/main/kotlin/net/corda/db/connection/manager/DbConfig.kt
```diff
@@ -34,12 +35,22 @@ fun DataSourceFactory.createFromConfig(config: SmartConfig): CloseableDataSource
...
     }
 
+    val idleTimeout =
+        configWithFallback.getInt(DatabaseConfig.DB_POOL_IDLE_TIMEOUT).toLong().run(Duration::ofSeconds)
+    val maxLifetime =
+        configWithFallback.getInt(DatabaseConfig.DB_POOL_MAX_LIFETIME).toLong().run(Duration::ofSeconds)
+    val keepaliveTime =
+        configWithFallback.getInt(DatabaseConfig.DB_POOL_KEEP_ALIVE_TIME).toLong().run(Duration::ofSeconds)
+    val validationTimeout =
+        configWithFallback.getInt(DatabaseConfig.DB_POOL_VALIDATION_TIMEOUT).toLong().run(Duration::ofSeconds)
```
driessamyn: Can we add the unit of time to the var name? (E.g. `_MS`, `_S`) so it's easier to figure out what is expected?

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDbFactoryImpl.kt
driessamyn: I'm not sure what the actual values should be, but I think they need to come from config too.

I disagree with the min=0, max=10 value, because I can't see how we'd ever need 10 vnode connections in a pool, unless we have a bug.

components/db/db-connection-manager/src/main/kotlin/net/corda/db/connection/manager/DbConfig.kt
```diff
@@ -34,12 +36,22 @@ fun DataSourceFactory.createFromConfig(config: SmartConfig): CloseableDataSource
     val driver = configWithFallback.getString(DatabaseConfig.JDBC_DRIVER)
     val jdbcUrl = configWithFallback.getString(DatabaseConfig.JDBC_URL)
     val maxPoolSize = configWithFallback.getInt(DatabaseConfig.DB_POOL_MAX_SIZE)
+    // The below `hashPath` takes care the case where DB_POOL_MIN_SIZE is null (DB_POOL_MIN_SIZE will always be present)
```
simon-johnson-r3: typo: hasPath

simon-johnson-r3 (changes requested):

simon-johnson-r3 (dismissed):

fowlerrr (dismissed):

kyriathar:
> I'm not sure what the actual values should be, but I think they need to come from config too.

> I disagree with the min=0, max=10 value, because I can't see how we'd ever need 10 vnode connections in a pool, unless we have a bug.

From offline conversation with @yift-r3 it seems that indeed max pool size of 1 should be fine for the most part. A need for a second connection could be when an API call (admin call I believe) comes in to DB worker for the same VNode while the DB connection for that VNode is already in use (when handling an MGM persistence request for example).  

We can discuss over this and most likely bring max pool size for VNode datasources down to 1 (or 2) when populating the VNode configuration values in follow-up PR I think.

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDbFactoryImpl.kt
simon-johnson-r3: They should come from the Config yes. That is out of scope for this story however, it will be in a later one. Note that some of these values were hard coded from way back already, we haven't changed anything in that regard.

I also think it's unlikely we need 10 vnode connections but I think we need the metrics story really.

components/db/db-connection-manager/src/main/kotlin/net/corda/db/connection/manager/DbConfig.kt
kyriathar: Raised https://github.com/corda/corda-api/pull/1150 to address this comment.

kyriathar: Done

driessamyn (approved):
LGTM

PR #4056, opened 2023-06-12
CORE-11411 - Refresh certificates using openssl and longer expiration window

dimosr:
Refresh certificates used for integration tests that were due to expire next month using openssl, so that they have a longer expiration window (50 years). I also updated the README file with some extra options I had to use to create the certificates in the required way (e.g. specifying key usage, a valid country, missing mandatory x500 attributes such as organisation etc.).

As part of this change, I didn't refresh:
* the C4 certificates as they expire in ~4 years and there's no step-by-step documentation on how they had been generated so it will require some reverse engineering and familiarisation with C4 that's probably not worth the effort (also given the fact that support of C4 certificates from C5 was part of the original design, but it's not going to be a feature from what we've seen so it's very likely the tests will get disabled at some point).
* the Bob certificate, which is a revoked certificate from Tinycert. The tests exercising revocation have been disabled and they will be enabled back again after resolving CORE-11411. However, I split the truststore out into a separate file, since the other certificates will now have a truststore coming from openssl.


yift-r3 (approved):

driessamyn (approved):
LGTM

PR #4059, opened 2023-06-12
CORE-14166 Prevent infinite retries for p2p messages

charlieR3:
Infinite retries were observed after putting the link manager under a high load using the app simulator despite no message loss being observed. After investigation, it was found that in certain scenarios, a message can be retrying and also queued to retry and so when the retry ended, the message was taken from the queue and replayed again but never removed. 

When a high load is pushed through the link manager with a max concurrent replay size of 100 (i.e. the default size), the following was observed. `LinkManagerProcessedMarker` is received by the link manager `DeliveryTracker` and the concurrent replay max is not yet reached so the message is registered with the `ReplayScheduler` to start replaying. The message is sent and the `DeliveryTracker` then receives a `LinkManangerSentMarker` with a non-null state. The `DeliveryTracker` again registers this for a replay with the `ReplayScheduler` but this time other messages have been sent and the max concurrent replays limit has been reached so the message is added to the queue to replay later (even though it is already replaying). The `DeliveryTracker` then receives the `LinkManagerReceivedMarker` with a non-null state and removes the message from the replay but not the queue. The message is added from the queue to the active replays and this continues to replay infinitely since as far as the `DeliveryTracker` is concerned, that message has finished replaying so it will never try to remove it.

This PR ensures that we don't queue an already replaying message, and when we remove a replaying message ID, we unqueue it if it was also added to the queue somehow as an extra safety precaution again infinite replays.


williamvigorr3:
> Infinite retries were observed after putting the link manager under a high load using the app simulator despite no message loss being observed. After investigation, it was found that in certain scenarios, a message can be retrying and also queued to retry and so when the retry ended, the message was taken from the queue and replayed again but never removed.
> 
> When a high load is pushed through the link manager with a max concurrent replay size of 100 (i.e. the default size), the following was observed. `LinkManagerProcessedMarker` is received by the link manager `DeliveryTracker` and the concurrent replay max is not yet reached so the message is registered with the `ReplayScheduler` to start replaying. The message is sent and the `DeliveryTracker` then receives a `LinkManangerSentMarker` with a non-null state. The `DeliveryTracker` again registers this for a replay with the `ReplayScheduler` but this time other messages have been sent and the max concurrent replays limit has been reached so the message is added to the queue to replay later (even though it is already replaying). The `DeliveryTracker` then receives the `LinkManagerReceivedMarker` with a non-null state and removes the message from the replay but not the queue. The message is added from the queue to the active replays and this continues to replay infinitely since as far as the `DeliveryTracker` is concerned, that message has finished replaying so it will never try to remove it.
> 
> This PR ensures that when we remove a replaying message ID, we unqueue it is it was also added to the queue. An alternative would be to avoid ever adding it to the queue if currently replaying. This is open for discussion and I can change this. The reason I left it at just removing from the queue without avoiding adding was in case there are other cases which might case a replaying message to be queued also.

I think we should consider do both i.e. be defensive and keep the current code but also not add it to the queue if currently replaying. We originally wrote the `DeliveryTracker` before the `LinkManagerProcessedMarker` was added (before this we added a `LinkManagerSentMarker` ) I think it was assumed that `addForReplay` would only be called once per `message id`.  The `InMemorySessionReplayer` should also be safe if we change the `ReplaySchedular` as the authentication protocol is idempotent (so generates the same response messages if any session messages get replayed).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/delivery/ReplayScheduler.kt
```diff
@@ -57,7 +57,7 @@ internal class ReplayScheduler<K: SessionManager.BaseCounterparties, M>(
         ConcurrentHashMap<K, MutableSet<MessageId>>()
     private val replayInfoPerMessageId = ConcurrentHashMap<MessageId, ReplayInfo>()
     data class QueuedMessage<M>(val originalAttemptTimestamp: Long, val uniqueId: MessageId, val message: M)
-    private val queuedMessagesPerCounterparties = ConcurrentHashMap<K, MessageQueue<M>>()
+    internal val queuedMessagesPerCounterparties = ConcurrentHashMap<K, MessageQueue<M>>()
```
williamvigorr3, line 61: Should we use `@VisibleForTesting` here?

williamvigorr3 (commented):
LGTM - I left a minor comment.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/delivery/ReplayScheduler.kt
charlieR3: Added

williamvigorr3 (approved):
LGTM

PR #4064, opened 2023-06-12
CORE-13785: Fix tooling

yift-r3:
This pull request fixes a few issues with the on board plugin and the deploy script:
1. Add tolerations to the Kafka yaml files.
2. Make the script location explicit.
3. Make the allow certificate plugin insecure.

### DIFF
--- a/applications/tools/p2p-test/app-simulator/scripts/deploy.sh
+++ b/applications/tools/p2p-test/app-simulator/scripts/deploy.sh
@@ -10,30 +10,29 @@ deploy() {
    echo Creating $namespace
    kubectl delete ns $namespace || echo ''
    kubectl create ns $namespace
+   prereqs_args="--install prereqs -n $namespace  \
+                oci://corda-os-docker.software.r3.com/helm-charts/corda-dev  \
+                --set image.registry=\"corda-os-docker.software.r3.com\"  \
+                --set kafka.replicaCount=$KAFKA_REPLICAS,kafka.zookeeper.replicaCount=$KAFKA_ZOOKEEPER_REPLICAS  \
+                -f \"$SCRIPT_DIR/prereqs-eks.yaml\"  \
+                --render-subchart-notes  \
+                --timeout 10m  \
+                --wait"
+   corda_args="--install corda -n $namespace oci://corda-os-docker.software.r3.com/helm-charts/release/os/5.0/corda \
+              --set imagePullSecrets={docker-registry-cred} --set image.tag=$DOCKER_IMAGE_VERSION \
+              --set image.registry=corda-os-docker.software.r3.com --values $REPO_TOP_LEVEL_DIR/values.yaml \
+              -f \"$SCRIPT_DIR/corda-eks.yaml\" \
+              --values $REPO_TOP_LEVEL_DIR/debug.yaml --wait --version $CORDA_CHART_VERSION"
    if kubectl get ns metrics-server > /dev/null 2>/dev/null ; then
-       prereqs_eks="$SCRIPT_DIR/prereqs-eks.metrics.yaml"
-       corda_eks="$SCRIPT_DIR/corda-eks.metrics.yaml"
-   else
-       prereqs_eks="$SCRIPT_DIR/prereqs-eks.yaml"
-       corda_eks="$SCRIPT_DIR/corda-eks.yaml"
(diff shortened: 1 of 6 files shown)


applications/tools/p2p-test/app-simulator/scripts/prereqs-eks.metrics.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+postgresql:
+  primary:
+    nodeSelector:
+      node: "prereqs"
+    tolerations:
+      - effect: "NoSchedule"
+        key: "node"
+        value: "prereqs"
+    resources:
```
williamvigorr3: Should we include the other file as well (same as `corda-eks.metrics.yaml`), instead of defining these resources twice. Otherwise someone might change them in one file and forget to update the other.

tools/plugins/mgm/src/main/kotlin/net/corda/cli/plugins/mgm/AllowClientCertificate.kt
```diff
@@ -49,6 +49,7 @@ class AllowClientCertificate : Runnable {
             targetUrl = baseUrlFromClusterName(cordaClusterName, restWorkerDeploymentName)
             password = restPasswordFromClusterName(cordaClusterName)
             username = "admin"
+            insecure = true
```
williamvigorr3, line 52: Is this tool is only used internally (I think that is the case)?

williamvigorr3 (commented):
I think long term maybe we should switch to the other deployment script.

tools/plugins/mgm/src/main/kotlin/net/corda/cli/plugins/mgm/AllowClientCertificate.kt
yift-r3: Yes, internal only.

applications/tools/p2p-test/app-simulator/scripts/prereqs-eks.metrics.yaml
yift-r3: Should be better now

applications/tools/p2p-test/app-simulator/scripts/deploy.sh
```diff
@@ -10,23 +10,29 @@ deploy() {
...
+                -f \"$SCRIPT_DIR/prereqs-eks.yaml\"  \
+                --render-subchart-notes  \
+                --timeout 10m  \
+                --wait"
+   corda_args="--install corda -n $namespace oci://corda-os-docker.software.r3.com/helm-charts/release/os/5.0/corda \
+              --set imagePullSecrets={docker-registry-cred} --set image.tag=$DOCKER_IMAGE_VERSION \
+              --set image.registry=corda-os-docker.software.r3.com --values $REPO_TOP_LEVEL_DIR/values.yaml \
+              -f \"$SCRIPT_DIR/corda-eks.yaml\" \
+              --values $REPO_TOP_LEVEL_DIR/debug.yaml --wait --version $CORDA_CHART_VERSION"
+   if kubectl get ns metrics-server > /dev/null 2>/dev/null ; then
```
YashNabar, line 26: Is this a sufficient check, or should we do something like the other [repo](https://github.com/corda/corda-non-functional-test/blob/3b79e13a6748dbcf5814e20f7f06b632d3fd14df/scripts/eks.sh#L43)?
```
# Used to determine whether to override pod's toleration/affinity when deploying
PERFORMANCE_ENABLED_CLUSTER=$(if [[ ${KUBERNETES_CONTEXT} == *"eks-e2e"* ]] || [[ ${KUBERNETES_CONTEXT} == *"corda-non-functional-test"* ]]; then echo "false"; else echo "true"; fi)
```

williamvigorr3: I guess this way has the advantage that if someone spins up a new cluster without metrics it will work straight away. Very nit picky: Maybe we shouldn't hardcode `metrics-server` set its as a variable inside settings.sh.

williamvigorr3 (approved):
LGTM (when @YashNabar is happy).

applications/tools/p2p-test/app-simulator/scripts/deploy.sh
YashNabar: We can resolve this if we're happy to proceed with the namespace check

YashNabar (approved):

PR #4065, opened 2023-06-12
CORE-8808 Fix error code in Certificates API

YashNabar:
Endpoints on Certificates API now throw 404 Resource Not found in case of non-existent holding identity ID, instead of 500 Internal Server Error.

Valid hex identifier, non-existent vnode:

Screenshot 2023-06-12 at 12 29 44

Invalid hex identifier:

Screenshot 2023-06-12 at 13 12 40


yift-r3 (approved):

dimosr (approved):

ronanbrowne (approved):
LGTM

PR #4067, opened 2023-06-12
CORE-14549: Add support for --trace-samples-per-second to Helm chart

josephzunigadaly:


charts/corda/values.schema.json
```diff
@@ -239,11 +239,25 @@
                     "examples": [
                         "http://tempo:9411"
                     ]
+                },
+                "samplesPerSecond": {
+                    "type": "string",
+                    "default": "",
+                    "title": "Number of request traces to sample per second, defaults to 1 sample per second. Set to 'unlimited' to record all traces",
```
vkolomeyko:
```suggestion
                    "title": "Number of request traces to sample per second, defaults to 1 sample per second. Set to 'unlimited' to record all traces, but in this case amount of tracing data produced can be quite vast.",
```

charts/corda/values.yaml
```diff
@@ -49,6 +49,8 @@ metrics:
 tracing:
   # -- URL for endpoint to send Zipkin-format distributed traces to e.g. http://tempo:9411
   endpoint: ""
+  # --  Number of request traces to sample per second, defaults to 1 sample per second. Set to 'unlimited' to record all traces
```
vkolomeyko:
```suggestion
  # --  Number of request traces to sample per second, defaults to 1 sample per second. Set to 'unlimited' to record all traces, but in this case amount of tracing data produced can be quite vast.
```

vkolomeyko (dismissed):
LGTM

charts/corda/values.schema.json
```diff
@@ -239,11 +239,25 @@
                     "examples": [
                         "http://tempo:9411"
                     ]
+                },
+                "samplesPerSecond": {
+                    "type": "string",
+                    "default": "",
```
davidcurrie:
```suggestion
                    "default": "1",
```

vkolomeyko (approved):
LGTM

mnesbit (approved):
OK

PR #4068, opened 2023-06-12
CORE-14360: Add support for batch publish tracing

josephzunigadaly:
Before:
image

After:
image


.run/Combined Worker Local.run.xml
```diff
@@ -2,11 +2,11 @@
...
+    <option name="PROGRAM_PARAMETERS" value="--instance-id=0 --send-trace-to=http://localhost:9411 -mbootstrap.servers=localhost:9092 -mbus.busType=KAFKA -spassphrase=password -ssalt=salt -spassphrase=password -ssalt=salt -ddatabase.user=user -ddatabase.pass=password -ddatabase.jdbc.url=jdbc:postgresql://localhost:5432/cordacluster -ddatabase.jdbc.directory=$ProjectFileDir$/applications/workers/release/combined-worker/drivers -rtls.crt.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/server.crt -rtls.key.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/server.key -rtls.ca.crt.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/ca-chain-bundle.crt" />
```
josephzunigadaly: Needs revert

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/ledger/UtxoLedgerTests.kt
```diff
@@ -145,7 +145,7 @@ class UtxoLedgerTests {
         val utxoFlowResult = awaitRpcFlowFinished(aliceHoldingId, utxoFlowRequestId)
         assertThat(utxoFlowResult.flowStatus).isEqualTo(RPC_FLOW_STATUS_SUCCESS)
         assertThat(utxoFlowResult.flowError).isNull()
-
+/*
```
josephzunigadaly: Needs revert

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
```diff
@@ -70,7 +72,7 @@ internal class BraveTracingService(serviceName: String, zipkinHost: String, samp
 
         // The console reporter is useful when debugging test runs on the combined worker.
         // uncomment it to enable it.
-        val reporters = mutableListOf<Reporter<Span>>(/*Reporter.CONSOLE*/)
+        val reporters = mutableListOf<Reporter<Span>>(Reporter.CONSOLE)
```
josephzunigadaly: Need to revert

.run/Combined Worker Local.run.xml
josephzunigadaly: Done

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/ledger/UtxoLedgerTests.kt
josephzunigadaly: Done

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
josephzunigadaly: Done

libs/tracing/src/main/kotlin/net/corda/tracing/TraceContext.kt
```diff
@@ -6,4 +6,10 @@ interface TraceContext {
     fun traceRequestId(requestId: String)
 
     fun traceVirtualNodeId(vNodeId: String)
+
+    fun markInScope():AutoCloseable
```
vkolomeyko: Nit: 
```suggestion
    fun markInScope(): AutoCloseable
```

libs/tracing/src/main/kotlin/net/corda/tracing/TraceContext.kt
```diff
@@ -6,4 +6,10 @@ interface TraceContext {
...
+
+    fun errorAndFinish(e:Exception)
```
vkolomeyko: Nit:
```suggestion
    fun errorAndFinish(e: Exception)
```

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
```diff
@@ -70,10 +72,11 @@ class FlowEventProcessorImpl(
 
         val pipeline = try {
             log.trace { "Flow [${event.key}] Received event: ${flowEvent.payload::class.java} / ${flowEvent.payload}" }
-            flowEventPipelineFactory.create(state, flowEvent, config, mdcProperties, event.timestamp)
-        } catch (t: Throwable) {
```
josephzunigadaly, line 74: Do we know why it was `Throwable` to start with? Is catching `Throwable` important so that `flowEventExceptionProcessor` sees every possible exit via exceptions?

If we are only changing to `Exception` so `traceContext.error(ex)` works, we could wrap the `Throwable` by creating a `CordaRuntimeException` and passing the `Thowable` as the `cause` field.

owenstanford: yes lets do that for now to avoid any issues

libs/flows/flow-api/build.gradle
```diff
@@ -18,6 +18,7 @@ dependencies {
     implementation project(":libs:messaging:messaging")
     implementation project(":libs:metrics")
     implementation project(":libs:serialization:serialization-checkpoint-api")
+    implementation project(":libs:tracing")
```
josephzunigadaly: This should be an `api` dependency (instead of `implementation`) because `TraceContext` is visible from this libraries public API.

Using `api` tells gradle that all projects depending on flow-api will also need to depend on tracing at compile time. This `implementation` option means the dependency is not available at compile time for those projects and they will need to duplicate the dependency in their projects.

It does raise the question of whether we should pass the context as a parameter or read it from a thread local like in other parts of the system.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/CordaKafkaProducerImpl.kt
```diff
@@ -104,9 +125,24 @@ class CordaKafkaProducerImpl(
...
+    ) {
+        val traceContext = traceSend(record.headers, "send $clientId")
+        traceContext.markInScope().use {
             try {
-                producer.send(record.toKafkaRecord(topicPrefix , partition), callback?.toKafkaCallback())
+                producer.send(
+                    record.toKafkaRecord(topicPrefix, partition),
+                    toTraceKafkaCallback({ exception -> callback?.onCompletion(exception) }, traceContext)
+                )
             } catch (ex: CordaRuntimeException) {
```
josephzunigadaly, line 144: Does `CordaRuntimeException` cover everything we need? Should we widen the scope of this catch block?

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTraceContext.kt
```diff
@@ -19,6 +21,22 @@ internal class BraveTraceContext(
...
+         return tracer.withSpanInScope(span)
+    }
+
+    override fun errorAndFinish(e: Exception) {
+        span.error(e).finish()
+    }
+
+    override fun error(e: Exception) {
+        span.error(e).finish()
+    }
```
josephzunigadaly: Looks like a duplicate. I guess `error` was not supposed to call `finish`?

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveRecordTracing.kt
```diff
@@ -10,26 +10,29 @@ class BraveRecordTracing(tracing: Tracing) {
...
     private val tracingContextExtractor = tracing.propagation().extractor(recordHeaderGetter)
 
     fun nextSpan(record: Record<*, *>): Span {
-        return nextSpan(record.topic, record.headers)
+        return nextSpan(record.headers).apply { tag("topic.name", record.topic) }
     }
 
     fun nextSpan(record: EventLogRecord<*, *>): Span {
-        return nextSpan(record.topic, record.headers)
+        return nextSpan(record.headers).apply { tag("topic.name", record.topic) }
```
josephzunigadaly: I also added a topic tag in `net.corda.messagebus.kafka.producer.CordaKafkaProducerImpl#toTraceKafkaCallback`
Should we remove one?

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveBatchPublishTracing.kt
```diff
@@ -0,0 +1,51 @@
...
+        // the incoming batch can contain messages with different trace contexts.
+        batchSpans.addAll(
+            recordHeaders.mapNotNull { headers ->
+                val extracted = tracingContextExtractor.extract(headers)
+                if (extracted.context() == null) {
+                    null
+                } else {
+                    extracted
+                }
+            }.groupBy { ctx ->
```
josephzunigadaly: We could simplify this `mapNotNull` lambda by splitting the work into a transformation then a filter.

```suggestion
            recordHeaders.map(tracingContextExtractor::extract)
            .filter { it.context() != null }
            .groupBy { ctx ->
```

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
josephzunigadaly: Done

libs/flows/flow-api/build.gradle
owenstanford: yes change to API.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/CordaKafkaProducerImpl.kt
owenstanford: I don't think anything has changed for the send, any exception in the callback won't be caught here

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTraceContext.kt
owenstanford: yes!

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveRecordTracing.kt
owenstanford: yes, the one in the producer callback should cover everything

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveBatchPublishTracing.kt
owenstanford: yes If it's cleaner

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/CordaKafkaProducerImpl.kt
josephzunigadaly: I'm thinking more about `producer.send` throwing an exception. It has a few it could throw around failure to submit the messages to Kafka (authentication, timeout, and more). Do we want to capture these in `errorAndFinish` for the span?

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveRecordTracing.kt
josephzunigadaly: Done

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/CordaKafkaProducerImpl.kt
owenstanford: yes that makes sense

josephzunigadaly: Added a catch Exception block to record all exceptions

vkolomeyko (approved):
LGTM

driessamyn (approved):
LGTM

PR #4069, opened 2023-06-12
CORE-14127: Fix CPK file attributes for CpkLoaderV2.

chrisr3:
Update `CpkLoaderV2` to write CPK files atomically, with safe file attributes.

It also now appears that `cpkPartsDir` actually contains expanded CPI and CPK files rather than chunks from Kafka, so update its `DiskSpace` metric description accordingly.

---
As an aside, we appear to have _two_ caches of CPK files - one "working" cache for combining Kafka chunks, and another "temporary" cache for extracting CPI and CPK contents. I suspect that some of this disk I/O is unnecessary, but I cannot change that now.

I also notice that this _now removed_ code created a CPK file _outside_ the cache directory `cacheDir` as a sibling of its parent:
```diff
- val finalCpkFile = cacheDir.parent.resolve(hash.toHexString()).toFile()
+ val cpkFile = cacheDir.resolve(hash.toHexString())
```
That cannot possibly have been the original intention.


driessamyn (approved):
LGTM

PR #4071, opened 2023-06-13
CORE-13660: Improve synchronisation stability 

yift-r3:
This pull request is making some changes to the synchronization to make it more stable:
1. Instead of persisting the group parameters in the member worker and waiting for a reply, it will send an asynchronous request to the DB worker to do that. For that, the group parameter persistent code was changed to publish the latest group parameter into the bus as part of the handler in the persistence service.
2. If the synchronization handling fails, we will try to ask for another synchronization from the MGM.

### DIFF
--- a/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistGroupParametersHandler.kt
+++ b/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistGroupParametersHandler.kt
@@ -6,6 +6,7 @@ import net.corda.data.membership.db.request.MembershipRequestContext
 import net.corda.data.membership.db.request.command.PersistGroupParameters
 import net.corda.data.membership.db.response.command.PersistGroupParametersResponse
 import net.corda.membership.datamodel.GroupParametersEntity
+import net.corda.membership.impl.persistence.service.RecoverableException
 import net.corda.membership.lib.GroupParametersNotaryUpdater.Companion.EPOCH_KEY
 import net.corda.membership.lib.exceptions.MembershipPersistenceException
 import net.corda.membership.lib.toMap
@@ -72,6 +73,16 @@ internal class PersistGroupParametersHandler(
                 .resultList
                 .first()
                 .toAvro()
+        }
+
+        try {
+            val parameters = groupParametersFactory.create(persistedGroupParameters)
+            groupParametersWriterService.put(
+                context.holdingIdentity.toCorda(),
+                parameters,
+            )
+        } catch (e: Exception) {
+            throw RecoverableException("Could not write the group parameters to the message bus.", e)
         }
 
         return PersistGroupParametersResponse(persistedGroupParameters)
--- a/components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
+++ b/components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
@@ -32,7 +32,6 @@ import net.corda.lifecycle.RegistrationStatusChangeEvent
 import net.corda.lifecycle.StartEvent
 import net.corda.lifecycle.StopEvent
 import net.corda.lifecycle.TimerEvent
-import net.corda.membership.groupparams.writer.service.GroupParametersWriterService
 import net.corda.membership.lib.GroupParametersFactory
 import net.corda.membership.lib.MemberInfoExtension.Companion.MEMBER_STATUS_SUSPENDED
 import net.corda.membership.lib.MemberInfoExtension.Companion.id
@@ -92,7 +91,6 @@ class MemberSynchronisationServiceImpl internal constructor(
     private val clock: Clock,
     private val membershipPersistenceClient: MembershipPersistenceClient,
     private val groupParametersFactory: GroupParametersFactory,
-    private val groupParametersWriterService: GroupParametersWriterService,
 ) : MemberSynchronisationService {
     @Suppress("LongParameterList")
     @Activate
@@ -121,8 +119,6 @@ class MemberSynchronisationServiceImpl internal constructor(
         membershipPersistenceClient: MembershipPersistenceClient,
         @Reference(service = GroupParametersFactory::class)
         groupParametersFactory: GroupParametersFactory,
-        @Reference(service = GroupParametersWriterService::class)
-        groupParametersWriterService: GroupParametersWriterService,
     ) : this(
         publisherFactory,
         configurationReadService,
@@ -149,14 +145,13 @@ class MemberSynchronisationServiceImpl internal constructor(
         UTCClock(),
         membershipPersistenceClient,
         groupParametersFactory,
-        groupParametersWriterService,
     )
 
     /**
      * Private interface used for implementation swapping in response to lifecycle events.
      */
     private interface InnerSynchronisationService : AutoCloseable {
-        fun processMembershipUpdates(updates: ProcessMembershipUpdates)
+        fun processMembershipUpdates(updates: ProcessMembershipUpdates): List<Record<*, *>>
 
         fun cancelCurrentRequestAndScheduleNewOne(
             memberIdentity: HoldingIdentity,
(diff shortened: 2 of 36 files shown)


components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistGroupParametersHandler.kt
```diff
@@ -74,6 +75,16 @@ internal class PersistGroupParametersHandler(
                 .toAvro()
         }
 
+        try {
```
dimosr, line 78: Not a bad idea to move this in here, so that we write to DB & Kafka at the same point.

But, I think if we do it only here it will be a bit inconsistent. For example, other components such as `ApproveRegistrationHandler` and `MGMRegistrationService` invoke persistence handlers and then the group parameters writer service separately. Should we do something similar for the persistence operations `addNotaryToGroupParameters` & `persistGroupParametersInitialSnapshot` for consistency? Happy for it to be a follow-up JIRA, if needed, to keep this PR relatively small.

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
```diff
@@ -358,27 +354,28 @@ class MemberSynchronisationServiceImpl internal constructor(
...
                 )
-                publisher.publish(allRecords).first().get(PUBLICATION_TIMEOUT_SECONDS, TimeUnit.SECONDS)
+                allRecords + persistRecords
             } catch (e: Exception) {
                 logger.warn("Failed to process membership updates received by ${viewOwningMember.x500Name}.", e)
-                // TODO - CORE-5813 - trigger sync protocol.
                 logger.warn(
                     "Cannot recover from failure to process membership updates. ${viewOwningMember.x500Name}" +
                             " cannot initiate sync protocol with MGM as this is not implemented."
                 )
```
dimosr: this log should be removed, since we now have sync implemented.

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
```diff
@@ -428,6 +425,30 @@ class MemberSynchronisationServiceImpl internal constructor(
...
+        return try {
+            listOf(
+                createSynchronisationRequestMessage(
+                    groupReader = membershipGroupReaderProvider.getGroupReader(viewOwningMember),
+                    memberIdentity = viewOwningMember,
+                    mgm = mgm,
+                ),
+            )
+        } catch (_: Exception) {
+            // Could not create a synchronisation, schedual to create one soon
```
dimosr:
```suggestion
            // Could not create a synchronisation, schedule to create one soon
```

dimosr: Also maybe worth a warning log in this scenario? e.g.
```
logger.warn("Failed to trigger an immediate sync, will schedule on to be triggered later.")
```

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
```diff
@@ -428,6 +425,31 @@ class MemberSynchronisationServiceImpl internal constructor(
...
+                    memberIdentity = viewOwningMember,
+                    mgm = mgm,
+                ),
+            )
+        } catch (e: Exception) {
+            logger.warn("Could not create a synchronisation request. Will retry soon.", e)
+            // Could not create a synchronisation, schedual to create one soon
+            coordinator.setTimer(
+                key = "SendSyncRequest-${viewOwningMember.fullHash}",
+                delay = (random.nextDouble() * TimeUnit.MINUTES.toMillis(RESEND_NOW_MAX_IN_MINUTES)).toLong()
```
dimosr: Do we need the jitter here? I think the idea for the periodic syncs was to avoid all the workers triggering it at the same time for now, but if the distribution is processed by a single worker we might not need it in this case.

components/membership/synchronisation-impl/src/test/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImplTest.kt
```diff
@@ -367,20 +363,11 @@ class MemberSynchronisationServiceImplTest {
...
-        whenever(mockPublisher.publish(capturedPublishedList.capture())).thenReturn(
-            listOf(
-                CompletableFuture.completedFuture(
-                    Unit
-                )
-            )
-        )
 
-        synchronisationService.processMembershipUpdates(updates)
+        val publishedMemberList = synchronisationService.processMembershipUpdates(updates)
```
dimosr:
```suggestion
        val producedRecords = synchronisationService.processMembershipUpdates(updates)
```

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistGroupParametersHandler.kt
yift-r3: CORE-14593 created

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
yift-r3: Probably not

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
yift-r3: I removed that comment (it wasn't helpful) and changed the log.

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
```diff
@@ -358,26 +354,26 @@ class MemberSynchronisationServiceImpl internal constructor(
...
                 )
-                publisher.publish(allRecords).first().get(PUBLICATION_TIMEOUT_SECONDS, TimeUnit.SECONDS)
+                allRecords + persistRecords
             } catch (e: Exception) {
-                logger.warn("Failed to process membership updates received by ${viewOwningMember.x500Name}.", e)
-                // TODO - CORE-5813 - trigger sync protocol.
                 logger.warn(
-                    "Cannot recover from failure to process membership updates. ${viewOwningMember.x500Name}" +
-                            " cannot initiate sync protocol with MGM as this is not implemented."
+                    "Failed to process membership updates received by ${viewOwningMember.x500Name}. Will retry again soon.",
```
dimosr:
```suggestion
                    "Failed to process membership updates received by ${viewOwningMember.x500Name}. Will trigger a fresh sync with MGM.",
```

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
```diff
@@ -428,6 +424,30 @@ class MemberSynchronisationServiceImpl internal constructor(
...
+        return try {
+            listOf(
+                createSynchronisationRequestMessage(
+                    groupReader = membershipGroupReaderProvider.getGroupReader(viewOwningMember),
+                    memberIdentity = viewOwningMember,
+                    mgm = mgm,
+                ),
+            )
+        } catch (e: Exception) {
+            logger.warn("Failed to trigger an immediate sync, will schedule on to be triggered later.", e)
```
dimosr:
```suggestion
            logger.warn("Failed to trigger an immediate sync, will schedule one to be triggered later.", e)
```

dimosr (commented):
just a couple of final minor suggestions

dimosr (approved):

PR #4072, opened 2023-06-13
CORE-12172 Fix error handling for static network notary updates

YashNabar:
Fixes an issue where a static network registration request with an invalid notary context fails with the wrong error reason. This was because a `MembershipPersistenceException` was thrown in case of invalid registration context, which would then be caught in `MemberOpsAsyncProcessor`, and replayed. Now an `InvalidGroupParametersUpdateException` is thrown, which leads to the request being converted to "INVALID" with the right error message.

**Static Network**
1. Missing notary protocol versions
Screenshot 2023-06-12 at 17 03 07

2. Missing notary protocol name
Screenshot 2023-06-12 at 17 12 26

**Dynamic Network**
1. Missing notary protocol versions
```
membership.db.rpc.ops] WARN  net.corda.membership.impl.persistence.service.MembershipPersistenceRPCProcessor {} - Exception thrown while processing membership persistence request: Cannot add notary to notary service 'O=NotaryServiceA, L=London, C=GB' - protocol versions are missing.
```
2. Missing notary protocol name
```
membership.db.rpc.ops] WARN  net.corda.membership.impl.persistence.service.MembershipPersistenceRPCProcessor {} - Exception thrown while processing membership persistence request: Cannot add notary to group parameters - notary protocol must be specified to create new notary service 'O=NotaryServiceB, L=London, C=GB'.
```


dimosr (approved):
LGTM

PR #4073, opened 2023-06-13
CORE-11657 SubFlow metric fix

lankydan:
The `isSubFlow` flag was incorrectly set as `true` for only the top-level flow.

When a SubFlow was completed correctly it was recorded as FAILED instead of COMPLETED.


emilybowe (approved):
LGTM

rick-r3 (approved):

PR #4075, opened 2023-06-13
CORE-14435: always pass SigningWrappedSpec into CryptoService.sign

dickon:
Currently `CryptoService.sign` took a `SigningSpec `as an argument, which is an interface which had 2 implementations; the method is polymorphic. One variant, `SigningAliasSpec`, was intended for use when signing when we only know the alias of a key, but not the private key itself. This is not supported, and we do not want to carry the complexity. 

Therefore, delete the interface and `SigningAliasSpec `which is an implementation class we never use, since it was part of the multiple HSM support. This change will make it easier to improve the caching behaviour of the crypto processor, since the current complexity we are addressing here means there can be multiple crypto services, which brings complexity to the fast path.

I considered making `SigningWrappedSpec `be called `SigningSpec`. I felt that would potentially be more confusing at this stage, but that's a marginal call. Long run, a lot of the `Spec` classes in the crypto processor can be replaced by a small number of arguments to various functions.

This is a step toward CORE-10051.


Sakpal (commented):
LGTM 👍

libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
```diff
@@ -721,7 +721,7 @@ object CordaMetrics {
         /**
          * Identifies the signature signing scheme name to create signatures during crypto signing operations.
          */
-        SignatureSpec("signature.spec"),
+        SigningWrappedSpec("signature.spec"),
```
kyriathar: I don't think we want this changed. It is still used as a tag key for `spec.signatureSpec.signatureName`.

dickon: yeah, good spot. Reverted. Testing now.

Sakpal (approved):
LGTM

PR #4077, opened 2023-06-13
CORE-14540: Explicitly exclude Snappy library dependency

vkolomeyko:
We should not need this dependency as long as we are not using `snappy` compression type in Kafka topics configuration.


ronanbrowne (approved):
LGTM

PR #4078, opened 2023-06-14
CORE-14583: change logging level to info for updateVirtualNodeState endpoint

Tom-Fitzpatrick:
The updateVirtualNodeState endpoint has been failing occasionally in various e2e tests. This PR changes the logging level in this endpoint to be info, so that we can hopefully get a better idea of where it is going wrong.


vkolomeyko (approved):
LGTM

PR #4079, opened 2023-06-14
CORE-14538: Remove full stack trace logging

vkolomeyko:
Exception like `ResourceNotFoundException` can be quite "normal" and there is no point logging them at INFO level. Should anyone will be willing to know more, it is possible to see the full stacktrace at DEBUG level if it is enabled.


dimosr (approved):
LGTM

josephzunigadaly (approved):

PR #4081, opened 2023-06-14
CORE-14519 Re-routing data messages from FlowMapper to P2P.out in GlobalPostProcessing

ben-millar:


JamesHR3 (commented):
Gist of this looks about right. I'd verify it through trace data etc, and it would need more work to productionize if we go down that route.

PR #4083, opened 2023-06-14
CORE-14523 Skipping P2P layer

mbrkic-r3:
Sending messages from the Flow Mapper to the flow event topic rather than sending to p2p.


pmck91:
Whats left to do at this point: 

- all OUTBOUND messages need to be a FlowMapperEvent type as opposed to AppMessage

JamesHR3 (commented):
Looks about right. Worth confirming with the trace data that this does as expected, and again some discussion would be required to productionize something like this.

PR #4084, opened 2023-06-14
CORE-11037 Move `DBConfig.kt` contents out of api module to use sites

kyriathar:
* Removes `DBConfig.kt` file and move its contents (i.e. fun `DataSourceFactory.createFromConfig` and `fun createDbConfig`) to use sites, because its contents i.e. these functions were not very much related conceptually. This also removes overhead from DB manager API module, as none of these functions were used from any of the module's types.


components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DataSourceFactoryHelper.kt
```diff
@@ -0,0 +1,80 @@
...
+
+    /** Default values from configuration schema */
+    private val dbFallbackConfig = getConfigurationDefaults(ConfigKeys.DB_CONFIG, DB_SCHEMA_VER)
+
+    /**
+     * Creates a [DataSource] using the [config].
+     *
+     * @throws DBConfigurationException If required configuration attributes are missing.
+     */
+    fun DataSourceFactory.createFromConfig(config: SmartConfig): CloseableDataSource {
```
simon-johnson-r3: I find it a bit unusual that this is in the internal object. Is there a reason for scoping them like this?

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDbFactoryImpl.kt
```diff
@@ -210,3 +214,36 @@ internal class VirtualNodeDbFactoryImpl(
      */
     private fun String.toSmartConfig() = smartConfigFactory.create(ConfigFactory.parseString(this))
 }
+
+@Suppress("LongParameterList")
+@VisibleForTesting
+internal fun createDbConfig(
```
simon-johnson-r3: I'd be tempted to make this a private function of `VirtualNodeDbFactoryImpl`
I assume it's internal for testing? Maybe see if you can derive the successful operation of this method from some other use of the class instance's public api? Chances are if you wrote this from scratch, you would have made it private and not bothered to unit test it, and we only got to this point because of emergent design.

simon-johnson-r3 (changes requested):

components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DataSourceFactoryHelper.kt
kyriathar: Good point, will move it out of the object.

kyriathar: Done

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDbFactoryImpl.kt
kyriathar: Made `createDbConfig` private, removed its previous dedicated test, and expanded different existing test to assert more expected VNode datasource configurations.

components/virtual-node/virtual-node-write-service-impl/src/test/kotlin/net/corda/virtualnode/write/db/impl/VirtualNodeDbFactoryImplTest.kt
```diff
@@ -62,11 +64,18 @@ class VirtualNodeDbFactoryImplTest {
             request,
         )
 
-        val vaultDmlConfig = dbs[VirtualNodeDbType.VAULT]?.dbConnections?.get(DbPrivilege.DML)?.config
-        verify(vaultDmlConfig!!).withValue(DatabaseConfig.DB_POOL_MAX_SIZE, ConfigValueFactory.fromAnyRef(10))
+        val vaultDmlConfig = dbs[VirtualNodeDbType.VAULT]?.dbConnections?.get(DbPrivilege.DML)?.config!!
+        verify(vaultDmlConfig, never()).withValue(eq(DatabaseConfig.JDBC_DRIVER), any())
+        verify(vaultDmlConfig).withValue(DatabaseConfig.JDBC_URL, ConfigValueFactory.fromAnyRef(JDBC_URL))
+        // TODO A VNode DML max pool size datasource needs to be changed to 1. This will change in follow-up PR.
```
simon-johnson-r3: I don't think we should have TODO comments in the os code. When you change this the test would fail anyway, so it can't be forgotten.

simon-johnson-r3 (changes requested):

components/virtual-node/virtual-node-write-service-impl/src/test/kotlin/net/corda/virtualnode/write/db/impl/VirtualNodeDbFactoryImplTest.kt
kyriathar: Good point, removed

simon-johnson-r3 (approved):

PR #4087, opened 2023-06-15
CORE-13893 Cache the result of state ref resolution

nkovacsx:
More info and performance testing: https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4519723031/Test+8+-+Cache+the+result+of+StateRef+resolve


blsemo (dismissed):
LGTM

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/StateRefCache.kt
```diff
@@ -0,0 +1,10 @@
+package net.corda.ledger.utxo.flow.impl.cache
+
+import net.corda.ledger.utxo.data.state.LazyStateAndRefImpl
+import net.corda.v5.ledger.utxo.StateRef
+
+interface StateRefCache {
+    fun get(stateRefs: Set<StateRef>): Map<StateRef, LazyStateAndRefImpl<*>?>
+
+    fun putAll(stateRefs: Map<StateRef, LazyStateAndRefImpl<*>?>)
```
lankydan: Why does this need to be `LazyStateAndRefImpl` instead of `StateAndRef`?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerStateQueryServiceImpl.kt
```diff
@@ -54,12 +58,23 @@ class UtxoLedgerStateQueryServiceImpl @Activate constructor(
...
+                val cachedStateRefs = stateRefCache.get(stateRefs.toSet())
+                val nonCachedStateRefs = stateRefs - cachedStateRefs.keys
+
+                if (nonCachedStateRefs.isNotEmpty()) {
+                    val resolvedStateRefs = wrapWithPersistenceException {
+                        externalEventExecutor.execute(
+                            ResolveStateRefsExternalEventFactory::class.java,
+                            ResolveStateRefsParameters(
+                                stateRefs
+                            )
```
lankydan: Doesn't need `stateRefs` to be on a new line.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerStateQueryServiceImpl.kt
```diff
@@ -54,12 +58,23 @@ class UtxoLedgerStateQueryServiceImpl @Activate constructor(
...
-                wrapWithPersistenceException {
-                    externalEventExecutor.execute(
-                        ResolveStateRefsExternalEventFactory::class.java,
-                        ResolveStateRefsParameters(stateRefs)
-                    )
-                }.map { it.toStateAndRef<ContractState>(serializationService) }
+                val cachedStateRefs = stateRefCache.get(stateRefs.toSet())
+                val nonCachedStateRefs = stateRefs - cachedStateRefs.keys
+
+                if (nonCachedStateRefs.isNotEmpty()) {
```
lankydan, line 64: I don't think this logic is correct.

We've pulled the `cachedStateRefs` from the cache, but then we haven't included them in the returned output when we go to the database.

So in the if statement it should return something like `cachedStateRefs + resolvedStateRefs`.

Do you think it would also be worth quickly ordering the output `StateAndRef`s to match the order of the `StateRef`s? I'm not super fussed about this but might be worth doing?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/impl/StateRefCacheImpl.kt
```diff
@@ -0,0 +1,89 @@
...
+                SandboxedCache.CacheKey(virtualNodeContext, it)
+            }).map {
+                    (key, value) -> key.key to value.lazyStateRef
+            }.toMap()
+        } else {
+            emptyMap()
+        }
+    }
+
+    override fun putAll(stateRefs: Map<StateRef, LazyStateAndRefImpl<*>?>) {
```
lankydan: We could have this take in a `List<StateAndRef>` instead and pull the `key`s from the `StateAndRef.getRef`.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/impl/StateRefCacheImpl.kt
```diff
@@ -0,0 +1,89 @@
...
+    @Reference(service = CacheEviction::class)
+    private val cacheEviction: CacheEviction
+) : SandboxedCache, StateRefCache {
+
+    private data class CacheValue(val lazyStateRef: LazyStateAndRefImpl<*>?)
+
+    // TODO Access configuration to setup the cache
+    private companion object {
+        private val log: Logger = LoggerFactory.getLogger(StateRefCacheImpl::class.java)
+        private const val STATE_REF_CACHE_SIZE_PROPERTY_NAME = "net.corda.ledger.utxo.flow.cache.maximumSize"
```
lankydan: Would use a more specific property name, probably include state and ref or state in there somewhere.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerStateQueryServiceImpl.kt
```diff
@@ -54,12 +58,23 @@ class UtxoLedgerStateQueryServiceImpl @Activate constructor(
...
+                            ResolveStateRefsExternalEventFactory::class.java,
+                            ResolveStateRefsParameters(
+                                stateRefs
+                            )
+                        )
+                    }.map { it.toStateAndRef<ContractState>(serializationService) }
+                    stateRefCache.putAll(resolvedStateRefs.associateBy { it.ref })
+                    resolvedStateRefs
+                } else {
+                    cachedStateRefs.values.mapNotNull { it?.deserializedStateAndRef }
```
lankydan: We shouldn't be accessing `deserializedStateAndRef` directly (its only really there for the serializer. If we switch to `StateAndRef`s being in the cache then we shouldn't need to worry about this. If the `LazyStateAndRefImpl` actually held in the cache has/is accessed then that will maintain its lazy value being initialised or not.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/impl/StateRefCacheImpl.kt
```diff
@@ -0,0 +1,89 @@
...
+@Suppress("unused")
+@Component(service = [StateRefCache::class])
+class StateRefCacheImpl @Activate constructor(
+    @Reference(service = CurrentSandboxGroupContext::class)
+    private val currentSandboxGroupContext: CurrentSandboxGroupContext,
+    @Reference(service = CacheEviction::class)
+    private val cacheEviction: CacheEviction
+) : SandboxedCache, StateRefCache {
+
+    private data class CacheValue(val lazyStateRef: LazyStateAndRefImpl<*>?)
```
lankydan: You shouldn't need `CacheValue` afaik. I used it in `MySigningKeysCache` because caffiene doesn't allow nulls.

We don't want to cache a `StateRef` to `null` because that `StateAndRef` _could_ exist in the future and the cache would prevent it from being accessed from the database.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/StateRefCache.kt
```diff
@@ -0,0 +1,10 @@
+package net.corda.ledger.utxo.flow.impl.cache
+
+import net.corda.ledger.utxo.data.state.LazyStateAndRefImpl
+import net.corda.v5.ledger.utxo.StateRef
+
+interface StateRefCache {
```
vlajos: Is not this a State**And**RefCache?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/StateRefCache.kt
nkovacsx: Can't remember exactly at this point, I think it was @vlajos 's suggestion. Probably because we can save extra deserializations sometimes?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerStateQueryServiceImpl.kt
nkovacsx: re. logic, good point
`to match the order of the StateRefs` -- I'm not sure if I understand

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/impl/StateRefCacheImpl.kt
nkovacsx: I don't have a strong opinion about this, can change it if you feel like.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/impl/StateRefCacheImpl.kt
nkovacsx: good point

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/StateRefCache.kt
nkovacsx: Good point

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/StateRefCache.kt
vlajos: AFAIK we wanted to delay deserialization as long as possible.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerStateQueryServiceImpl.kt
lankydan: Users have requested some state and refs by providing some state refs, so I thought providing them in the same order the state refs were passed in. That said, we take an `Iterable` as input so i don't think that has an ordering.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/impl/StateRefCacheImpl.kt
lankydan: I'd have a list passed in.

Also just noticed, should we call this `StateAndRefCache` instead?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/StateRefCache.kt
nkovacsx: I ended up using keeping `LazyStateAndRefImpl` under the hood but the API is now using `StateAndRef<*>`.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerStateQueryServiceImpl.kt
nkovacsx: Yeah I double checked and couldn't find any ordering.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/impl/StateRefCacheImpl.kt
nkovacsx: Yeah applied both of your suggestions.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/impl/StateAndRefCacheImpl.kt
```diff
@@ -0,0 +1,85 @@
...
+    @Reference(service = CurrentSandboxGroupContext::class)
+    private val currentSandboxGroupContext: CurrentSandboxGroupContext,
+    @Reference(service = CacheEviction::class)
+    private val cacheEviction: CacheEviction
+) : SandboxedCache, StateAndRefCache {
+
+    // TODO Access configuration to setup the cache
+    private companion object {
+        private val log: Logger = LoggerFactory.getLogger(StateAndRefCacheImpl::class.java)
+        private const val STATE_REF_CACHE_SIZE_PROPERTY_NAME = "net.corda.ledger.utxo.flow.state.cache.maximumSize"
```
vlajos: The const's name says `STATE_REF`, the key itself says `state`.
I would change both to `StateAndRef`

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/impl/StateAndRefCacheImpl.kt
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    private val maximumSize = java.lang.Long.getLong(STATE_REF_CACHE_SIZE_PROPERTY_NAME, 10000)
+
+    private val cache: Cache<SandboxedCache.CacheKey<StateRef>, StateAndRef<*>> = CacheFactoryImpl().build(
+        "State-Ref-Cache",
```
vlajos: I would add `and` to this one too.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/StateAndRefCache.kt
```diff
@@ -0,0 +1,22 @@
...
+
+    /**
+     * Returns a map of [StateRef] and [StateAndRef] associations from the cache for the given [stateRefs].
+     * Please note that this function WON'T return `null`s. This means, if a there's no [StateAndRef]
+     * object for a given [StateRef], it will not be present in the returned map.
+     */
+    fun get(stateRefs: Set<StateRef>): Map<StateRef, StateAndRef<*>>
+
+    /**
+     * Inserts the provided [stateRefs] list to the cache. It will be keyed on the [StateRef].
```
vlajos:
```suggestion
     * Inserts the provided [stateAndRefs] list to the cache. It will be keyed on the [StateRef].
```

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/StateAndRefCache.kt
```diff
@@ -0,0 +1,22 @@
...
+     */
+    fun putAll(stateRefs: List<StateAndRef<*>>)
```
vlajos:
```suggestion
    fun putAll(stateAndRefs: List<StateAndRef<*>>)
```

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/cache/impl/StateAndRefCacheImpl.kt
nkovacsx: good point thanks!

vlajos (commented):
I think `findSignedLedgerTransactionWithStatus` gets some stateAndRefs from the database too since https://github.com/corda/corda-runtime-os/commit/baf444063d18e528f608fb6a284d1aeb71a9e920
It would be good to add those to the cache as well.

vlajos (approved):

PR #4088, opened 2023-06-15
CORE-14564 - Review backward compatibility for RegistrationStatus

nikinagy:
Adding new values for enums in avro is not considered as a backward compatible change.
We’ve added a new value STARTED_PROCESSING_BY_MGM to flag in-progress registrations. This change unfortunately broke the registration when doing cross-version testing. This PR contains the fix for this issue.

### DIFF
--- a/components/membership/membership-p2p-impl/src/main/kotlin/net/corda/membership/impl/p2p/handler/SetOwnRegistrationStatusHandler.kt
+++ b/components/membership/membership-p2p-impl/src/main/kotlin/net/corda/membership/impl/p2p/handler/SetOwnRegistrationStatusHandler.kt
@@ -11,7 +11,6 @@ import net.corda.membership.lib.SetOwnRegistrationStatusV2
 import net.corda.schema.Schemas.Membership.REGISTRATION_COMMAND_TOPIC
 import net.corda.schema.registry.AvroSchemaRegistry
 import net.corda.schema.registry.deserialize
-import net.corda.v5.base.exceptions.CordaRuntimeException
 import net.corda.virtualnode.toCorda
 import java.nio.ByteBuffer
 
@@ -22,10 +21,10 @@ internal class SetOwnRegistrationStatusHandler(
         header: AuthenticatedMessageHeader,
         payload: ByteBuffer
     ): Record<String, RegistrationCommand> {
-        val response = try {
-            avroSchemaRegistry.deserialize(payload)
-        } catch (e: CordaRuntimeException) {
+        val response = if (avroSchemaRegistry.getClassType(payload) == SetOwnRegistrationStatus::class.java) {
             avroSchemaRegistry.deserialize<SetOwnRegistrationStatus>(payload).convertToNewVersion()
+        } else {
+            avroSchemaRegistry.deserialize(payload)
         }
         return Record(
             REGISTRATION_COMMAND_TOPIC,
@@ -41,11 +40,16 @@ internal class SetOwnRegistrationStatusHandler(
 
     private fun SetOwnRegistrationStatus.convertToNewVersion() : SetOwnRegistrationStatusV2 {
         val status = when(newStatus) {
+            RegistrationStatus.NEW -> RegistrationStatusV2.NEW
+            RegistrationStatus.SENT_TO_MGM -> RegistrationStatusV2.SENT_TO_MGM
             RegistrationStatus.RECEIVED_BY_MGM -> RegistrationStatusV2.RECEIVED_BY_MGM
+            RegistrationStatus.PENDING_MEMBER_VERIFICATION -> RegistrationStatusV2.PENDING_MEMBER_VERIFICATION
             RegistrationStatus.PENDING_MANUAL_APPROVAL -> RegistrationStatusV2.PENDING_MANUAL_APPROVAL
             RegistrationStatus.PENDING_AUTO_APPROVAL -> RegistrationStatusV2.PENDING_AUTO_APPROVAL
             RegistrationStatus.APPROVED -> RegistrationStatusV2.APPROVED
             RegistrationStatus.DECLINED -> RegistrationStatusV2.DECLINED
+            RegistrationStatus.INVALID -> RegistrationStatusV2.INVALID
+            RegistrationStatus.FAILED -> RegistrationStatusV2.FAILED
             else -> throw IllegalArgumentException("Unknown status '${newStatus.name}' received.")
         }
         return SetOwnRegistrationStatusV2(registrationId, status)
--- a/components/membership/membership-p2p-impl/src/main/kotlin/net/corda/membership/impl/p2p/handler/SetOwnRegistrationStatusHandler.kt
+++ b/components/membership/membership-p2p-impl/src/main/kotlin/net/corda/membership/impl/p2p/handler/SetOwnRegistrationStatusHandler.kt
@@ -2,12 +2,16 @@ package net.corda.membership.impl.p2p.handler
 
 import net.corda.data.membership.command.registration.RegistrationCommand
 import net.corda.data.membership.command.registration.member.PersistMemberRegistrationState
-import net.corda.data.membership.p2p.v2.SetOwnRegistrationStatus
+import net.corda.data.membership.common.RegistrationStatus
+import net.corda.data.membership.p2p.SetOwnRegistrationStatus
 import net.corda.messaging.api.records.Record
 import net.corda.data.p2p.app.AuthenticatedMessageHeader
+import net.corda.membership.lib.RegistrationStatusV2
+import net.corda.membership.lib.SetOwnRegistrationStatusV2
 import net.corda.schema.Schemas.Membership.REGISTRATION_COMMAND_TOPIC
 import net.corda.schema.registry.AvroSchemaRegistry
 import net.corda.schema.registry.deserialize
+import net.corda.v5.base.exceptions.CordaRuntimeException
 import net.corda.virtualnode.toCorda
 import java.nio.ByteBuffer
 
@@ -18,7 +22,11 @@ internal class SetOwnRegistrationStatusHandler(
         header: AuthenticatedMessageHeader,
         payload: ByteBuffer
     ): Record<String, RegistrationCommand> {
-        val response = avroSchemaRegistry.deserialize<SetOwnRegistrationStatus>(payload)
+        val response = try {
+            avroSchemaRegistry.deserialize(payload)
+        } catch (e: CordaRuntimeException) {
+            avroSchemaRegistry.deserialize<SetOwnRegistrationStatus>(payload).convertToNewVersion()
+        }
         return Record(
             REGISTRATION_COMMAND_TOPIC,
             "${response.registrationId}-${response.newStatus}-${header.destination.toCorda().shortHash}",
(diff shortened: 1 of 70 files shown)


components/membership/membership-p2p-impl/src/main/kotlin/net/corda/membership/impl/p2p/handler/SetOwnRegistrationStatusHandler.kt
```diff
@@ -18,7 +22,11 @@ internal class SetOwnRegistrationStatusHandler(
         header: AuthenticatedMessageHeader,
         payload: ByteBuffer
     ): Record<String, RegistrationCommand> {
-        val response = avroSchemaRegistry.deserialize<SetOwnRegistrationStatus>(payload)
+        val response = try {
+            avroSchemaRegistry.deserialize(payload)
+        } catch (e: CordaRuntimeException) {
+            avroSchemaRegistry.deserialize<SetOwnRegistrationStatus>(payload).convertToNewVersion()
+        }
```
dimosr: Can we switch that to use conditional logic, instead of relying on exception handling? i.e. use `getClassType` you could find the type first and then operate conditionally based on the type to deserialize to the right one.

components/membership/membership-p2p-impl/src/main/kotlin/net/corda/membership/impl/p2p/handler/SetOwnRegistrationStatusHandler.kt
```diff
@@ -30,4 +38,16 @@ internal class SetOwnRegistrationStatusHandler(
...
     }
+
+    private fun SetOwnRegistrationStatus.convertToNewVersion() : SetOwnRegistrationStatusV2 {
+        val status = when(newStatus) {
+            RegistrationStatus.RECEIVED_BY_MGM -> RegistrationStatusV2.RECEIVED_BY_MGM
+            RegistrationStatus.PENDING_MANUAL_APPROVAL -> RegistrationStatusV2.PENDING_MANUAL_APPROVAL
+            RegistrationStatus.PENDING_AUTO_APPROVAL -> RegistrationStatusV2.PENDING_AUTO_APPROVAL
+            RegistrationStatus.APPROVED -> RegistrationStatusV2.APPROVED
+            RegistrationStatus.DECLINED -> RegistrationStatusV2.DECLINED
+            else -> throw IllegalArgumentException("Unknown status '${newStatus.name}' received.")
```
dimosr, line 53: I know that some values are not sent across, but should we map all the values here for completeness?

If you do that, `when` will become exhaustive and you might be able to get rid of the else clause too.

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/VersionedMessageBuilder.kt
```diff
@@ -0,0 +1,52 @@
...
+
+import net.corda.data.membership.common.RegistrationStatus
+import net.corda.data.membership.p2p.SetOwnRegistrationStatus
+import org.apache.avro.specific.SpecificRecordBase
+
+private val regexForV50000 = "500[0-9][0-9]$".toRegex()
+private val regexForV50100 = "501[0-9][0-9]$".toRegex()
+
+fun retrieveRegistrationStatusMessage(platformVersion: Int, registrationId: String, status: String): SpecificRecordBase {
+    val platformVersionString = platformVersion.toString()
```
dimosr: why don't we perform gating based on the numerical value here, instead of using regex? e.g. something like
``` 
if (platformVersion < 50100) {
    // use v1
} else {
    // use v2
}
```
I think it'll be a bit easier to understand and maintain.

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/VersionedMessageBuilder.kt
```diff
@@ -0,0 +1,52 @@
...
+    return when {
+        regexForV50000.matches(platformVersionString) -> SetOwnRegistrationStatus(registrationId, retrieveV1Status(status))
+        regexForV50100.matches(platformVersionString) -> SetOwnRegistrationStatusV2(registrationId, retrieveV2Status(status))
+        else -> throw IllegalArgumentException("Unknown platform version: $platformVersion")
+    }
+}
+
+private fun retrieveV1Status(status: String): RegistrationStatus {
```
dimosr: While looking at this, I was thinking of the following alternative:
* doing automatic conversion of the string value to the corresponding enum, instead of doing it manually for every value. I guess in that case we wouldn't be able to have logic around statuses that are sent vs ones that remain in the cluster, but do we really need this at this level (or can we convert everything)?
* If the first suggestion is followed, I think the exception would have to go. But, I think we could replace it with an optional return type for the cases where the string can't be mapped to that version of registration status and we'd need to adjust the call site to handle that properly (i.e. if it's not present not send the status at all, but continue normally with the rest).

To be honest, I thought about both approaches and each one has its pros and cons, so I don't think I have a strong opinion. But, keen to hear what you think.

libs/membership/membership-common/src/test/kotlin/net/corda/membership/lib/VersionedMessageBuilderTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import java.util.UUID
+
+class VersionedMessageBuilderTest {
+    private companion object {
+        val registrationId = UUID.randomUUID().toString()
+    }
+
+    @Test
+    fun `SetOwnRegistrationStatus version 1 messages are built as expected`() {
+        with(retrieveRegistrationStatusMessage(50001, registrationId, RegistrationStatus.APPROVED.name)) {
```
dimosr: minor, but for completeness we could test all the values in each of those tests (using a loop over a list of values to avoid having a ton of test cases).

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/VersionedMessageBuilder.kt
nikinagy: In terms of maintainability I think your suggestion is better overall, because we won't need to think about what is being reasonably sent.
We can easily use `RegistrationStatus.valueOf(status)`, which by default throws and IllegalArgumentException if the string is not known: `java.lang.IllegalArgumentException: No enum constant net.corda.data.membership.common.v2.RegistrationStatus.dummyStatus`
I think this should be enough. There is a very little chance that someone (a developer from our team) will randomly put there a string.

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/VersionedMessageBuilder.kt
```diff
@@ -0,0 +1,14 @@
+package net.corda.membership.lib
+
+import net.corda.data.membership.common.RegistrationStatus
+import net.corda.data.membership.p2p.SetOwnRegistrationStatus
+
+fun retrieveRegistrationStatusMessage(platformVersion: Int, registrationId: String, status: String) =
```
dimosr: I think it's worth capturing `IllegalArgumentException`s here, return null or something else that represents invalid conversion and handle it upstream by not sending the status update to the member. 

This would help be a bit more defensive in scenarios, such as when trying to send newer values to older members that they don't know about. In practice, not much of a problem for now, since the only newly added value (STARTED_PROCESSING_BY_MGM) is not exchanged. But, it would be good to protect against this issue for the future, in case we add new values that are exchanged.

dimosr (commented):
just a minor thing - otherwise LGTM

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/VersionedMessageBuilder.kt
nikinagy: done in [68fc4ee](https://github.com/corda/corda-runtime-os/pull/4088/commits/68fc4eea82d58ffac87838e5d12dd0b638a461d8)

ronanbrowne (dismissed):
LGTM (build - presuming `5.1.0.747-MGM-alpha-1687185551196` is the correct version your pinning to)

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
```diff
@@ -190,17 +193,20 @@ internal class StartRegistrationHandler(
...
-            outputRecords.add(persistMemberStatusMessage)
+            if (statusUpdateMessage != null) {
+                p2pRecordsFactory.createAuthenticatedMessageRecord(
+                    source = mgmHoldingId.toAvro(),
+                    destination = pendingMemberHoldingId.toAvro(),
+                    content = statusUpdateMessage,
+                    minutesToWait = 5,
+                    filter = PENDING
+                )
+            } else { null }?.let { outputRecords.add(it) }
```
dimosr:
```suggestion
            if (statusUpdateMessage != null) {
                val record = p2pRecordsFactory.createAuthenticatedMessageRecord(
                    source = mgmHoldingId.toAvro(),
                    destination = pendingMemberHoldingId.toAvro(),
                    content = statusUpdateMessage,
                    minutesToWait = 5,
                    filter = PENDING
                )
                outputRecords.add(record)
            }
```

components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/ApproveRegistrationHandlerTest.kt
```diff
@@ -262,6 +262,17 @@ class ApproveRegistrationHandlerTest {
         assertThat(holdingIdentityCaptor.firstValue).isEqualTo(mgm.holdingIdentity)
     }
 
+    @Test
+    fun `invoke does not send registration status update message when status cannot be retrieved`() {
```
dimosr, line 268: I think these unit tests are misleading and are not testing what their name implies, because what you want to exercise is what happens if `retrieveRegistrationStatusMessage` returns null, not `createAuthenticatedMessageRecord` returns null.

I think the problem you have here is `retrieveRegistrationStatusMessage` is not easily mockable/fakable. It would be easier if it was on an object you could inject, for example.

If you don't want to change that now to get that in, we should at least remove these tests.

dimosr:
Can you also remind me why we are pinning to alpha versions here, instead of beta?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
nikinagy: addressed in [1e618c0](https://github.com/corda/corda-runtime-os/pull/4088/commits/1e618c0f0df8097ba22affd974cc4ebda81ee906)

components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/ApproveRegistrationHandlerTest.kt
nikinagy: addressed in [1e618c0](https://github.com/corda/corda-runtime-os/pull/4088/commits/1e618c0f0df8097ba22affd974cc4ebda81ee906) by using static mock

dimosr (dismissed):

dimosr (approved):

PR #4090, opened 2023-06-15
CORE-14583: change logging level to info for updateVirtualNodeState endpoint

Tom-Fitzpatrick:
The updateVirtualNodeState endpoint has been failing occasionally in various e2e tests. This PR changes the logging level in this endpoint to be info, so that we can hopefully get a better idea of where it is going wrong.

Recent example of a flaky build: https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda5-e2e-tests/detail/release%2F5.0/705/tests/


vkolomeyko (approved):
LGTM

driessamyn:
> so that we can hopefully get a better idea of where it is going wrong.

Do we see this flake in the 5.1 branch too? If so, I'd rather this temporary logging level change was only merged there.

Tom-Fitzpatrick:
The flake is appearing in 5.1 also, and I have this logging there already. Sure, I can wait until we see the failure on that branch. Thanks.

PR #4091, opened 2023-06-15
CORE-14536: Make static registration with notery more idempotent

yift-r3:
This change allows the static registration to retry if the persistence had failed with a notary member.

1. It will allow a member with the same name and same notary service to try and register again.
2. It will use the group parameters from the bus (if available) instead of the database - the bus is published before the database (the static network info) is persisted.
3. It will not try to update the static member info if there are no changes to the group parameters.

### DIFF
--- a/components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/staticnetwork/StaticMemberRegistrationService.kt
+++ b/components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/staticnetwork/StaticMemberRegistrationService.kt
@@ -102,7 +102,7 @@ class StaticMemberRegistrationService(
     internal val keyEncodingService: KeyEncodingService,
     private val cryptoOpsClient: CryptoOpsClient,
     val configurationReadService: ConfigurationReadService,
-    private val coordinatorFactory: LifecycleCoordinatorFactory,
+    coordinatorFactory: LifecycleCoordinatorFactory,
     private val hsmRegistrationClient: HSMRegistrationClient,
     private val memberInfoFactory: MemberInfoFactory,
     private val persistenceClient: MembershipPersistenceClient,
@@ -265,7 +265,7 @@ class StaticMemberRegistrationService(
             )
             (records + createHostedIdentity(member, groupPolicy)).publish()
 
-            persistGroupParameters(memberInfo, staticMemberList)
+            persistGroupParameters(memberInfo, staticMemberList, membershipGroupReader)
 
             persistRegistrationRequest(registrationId, memberInfo)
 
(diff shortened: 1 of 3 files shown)


components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/staticnetwork/StaticMemberRegistrationService.kt
```diff
@@ -293,9 +293,10 @@ class StaticMemberRegistrationService(
 
     private fun persistGroupParameters(
         memberInfo: MemberInfo,
-        staticMemberList: List<StaticMember>
+        staticMemberList: List<StaticMember>,
+        membershipGroupReader: MembershipGroupReader,
     ) {
-        val staticNetworkInfo = getCurrentStaticNetworkConfigWithRetry(memberInfo)
+        val staticNetworkInfo = getCurrentStaticNetworkConfigWithRetry(membershipGroupReader, memberInfo)
```
dimosr, line 299: > It will use the group parameters from the bus (if available) instead of the database - the bus is published before the database (the static network info) is persisted.

Looking at the code, it seems like static network information is persisted first (as part of `getCurrentStaticNetworkConfigWithRetry`, lline 299 in `StaticMemberRegistrationService`) before the publication of the group parameters to the message bus (which is now done as part of `persistenceClient.persistGroupParameters`, line 306 in `StaticMemberRegistrationService`), no?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/staticnetwork/StaticMemberRegistrationService.kt
```diff
@@ -591,14 +600,17 @@ class StaticMemberRegistrationService(
...
                     .reduceOrNull { acc, it -> acc.apply { retainAll(it) } }
                     ?: emptySet()
+                val latestGroupParameters = membershipGroupReader.groupParameters?.toAvro() ?: currentStaticNetworkInfo.groupParameters
+                latestGroupParameters.addNotary(
+                    memberInfo,
+                    currentProtocolVersions,
+                    keyEncodingService,
+                    clock
+                )
+            }?.let { groupParameters ->
```
dimosr, line 610: > It will not try to update the static member info if there are no changes to the group parameters.

where is that done exactly? Having a very quick look at `latestGroupParameters.addNotary` it seems like it always returns a value, so the next commands would always execute.

dimosr (commented):
Code looks good to me. Just a few questions on your comments in the description (left them inline in the relevant places to make it a bit easier).

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/staticnetwork/StaticMemberRegistrationService.kt
yift-r3: We don't. I will revert this in the next commit.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/staticnetwork/StaticMemberRegistrationService.kt
yift-r3: The `addNotary` can call either `updateExistingNotaryService` or `addNewNotaryService`. The `updateExistingNotaryService` will return nulls if the current group parameters already have those keys.

dimosr (approved):
LGTM

PR #4093, opened 2023-06-15
CORE-12457 Consolidate SignedGroupParameters fields

jennyang-r3:
Overview
Consolidated variable names to align with SignedGroupParameters avro schema

Test note
unit & integration tests modified accordingly

### DIFF
--- a/libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/InternalGroupParameters.kt
+++ b/libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/InternalGroupParameters.kt
@@ -15,11 +15,11 @@ interface InternalGroupParameters: GroupParameters {
      * The serialized form of the group parameters is always the source of truth over deserialized params to support
      * signing.
      */
-    val bytes: ByteArray
+    val groupParameters: ByteArray
 
     /**
      * Returns the [SecureHash] of the group parameters. The group parameters hash is a hash over the group parameters
-     * serialised byte array available as [bytes].
+     * serialised byte array available as [groupParameters].
      */
     val hash: SecureHash
 }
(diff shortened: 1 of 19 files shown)


libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/InternalGroupParameters.kt
```diff
@@ -15,11 +15,11 @@ interface InternalGroupParameters: GroupParameters {
      * The serialized form of the group parameters is always the source of truth over deserialized params to support
      * signing.
      */
-    val bytes: ByteArray
+    val groupParameters: ByteArray
```
nkovacsx, line 18: Looking at `groupParameters` someone might confuse it with the actual group parameters object and not a serialised byte array. Calling this `serializedGroupParameters` would make more sense to me. That's just my opinion though, let's see what @lankydan thinks.

lankydan: I don't mind too much, I'll let @vlajos have the final say since he raised the ticket.

lankydan (approved):
Wait for @vlajos's reply to my comment, otherwise looks good.

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/InternalGroupParameters.kt
vlajos: That would diverge it again from the Avro class.
Or the Avro class should need to be adjusted as well to keep them consistent.
I think that would be a bit too much for the current ticket's scope.
Also that API is the @corda/corda-platform-network-team's domain.

charlieR3 (approved):
LGTM

PR #4094, opened 2023-06-15
CORE-14522 Record durable message pattern output message metrics.

owenstanford:


libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
```diff
@@ -19,6 +18,7 @@ import java.nio.file.Path
 import java.util.function.Supplier
 import java.util.function.ToDoubleFunction
 import java.util.function.ToLongFunction
+import io.micrometer.core.instrument.Tag as micrometerTag
```
chrisr3, line 21: Why has this `import` moved? (IntelliJ _has_ been known to be dumb...)

PR #4096, opened 2023-06-15
CORE-14335: pass signing repository into crypto service

dickon:
We are working towards improving caching of signing keys, which is currently split between `components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt` and `net/corda/crypto/service/impl/SigningServiceImpl.kt`. A practical production-grade implementation of signing key caching is best done by combing those 2 layers, which we were planning anyway (CORE-10051). To make such a combination requires the signing repository to be available for the crypto service. So, we do that here, as an incremental step.

### DIFF
--- a/components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt
+++ b/components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt
@@ -22,6 +22,7 @@ import net.corda.crypto.hes.core.impl.deriveDHSharedSecret
 import net.corda.crypto.impl.SignatureInstances
 import net.corda.crypto.impl.getSigningData
 import net.corda.crypto.persistence.WrappingKeyInfo
+import net.corda.crypto.softhsm.SigningRepositoryFactory
 import net.corda.crypto.softhsm.WrappingRepositoryFactory
 import net.corda.crypto.softhsm.deriveSupportedSchemes
 import net.corda.metrics.CordaMetrics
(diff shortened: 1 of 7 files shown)


components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt
```diff
@@ -45,6 +46,8 @@ const val PRIVATE_KEY_ENCODING_VERSION: Int = 1
  *
  * @param wrappingRepositoryFactory which provides a factory for [WrappingRepository], which provides save and
  *        find operations for wrapping keys on a specific tenant.
+ * @param signingRepositoryFactory which provides a factory for [SigningRepository], which provides save and
+ *        find operations for signing keys fon a specific tenant.
```
anien, line 50: Shouldn't we call it `signingKeyRepositoryFactory`? I think it just aligns better of what it does.

dickon: We follow the convention that the parameter name should match the type name, only with a lower case initial letter, and I think it is good to keep that, so I'd want to change the class and file names as well.

The class is called `SigningRepository` and maybe it should be called `SigningKeysRepository` (and similary then we'd want to do `WrappingKeysRepository`. Not something I want to change in this branch, since we have a big enough change already.

At some point we may want to do a big code reorg since a lot of the `corda-runtime-os.git/components` crypto code could  move to `corda-runtime-os/libs` and reviewing all the class names might be worthwhile.

anien (approved):
LGTM.

dickon:
This was merged as part of https://github.com/corda/corda-runtime-os/pull/4151 since I struggled at the time to get a green build so merged a later stage in the development instead (and got the commit message slightly incomplete, and that can be addressed when it goes to a release branch).

PR #4097, opened 2023-06-15
CORE-14335: further structural changes to support cache improvements

dickon:
Pass signing repository into crypto service.

Make HSMServiceImpl a simple class, delete dead code

doOSGI hsm service impl

Only test with wrapping key present, since we've dropped support for keyMaterial being null (i.e. public key only mode)

Remove secondary constructor on SigningServiceImpl

(cherry-picked from  425c7926c89e87e9b5508564437fc023c2487588 which is in 5.1 so that the 5.0 and 5.1 crypto workers will be in sync once all the merging is done.

We want the changes here to simplify the upcoming caching change)

Co-authored-by: Kyriakos Tharrouniatis <kyriakos.tharrouniatis@r3.com>

(This PR includes a change in https://github.com/corda/corda-runtime-os/pull/4096)


components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/SigningServiceGeneralTests.kt
```diff
@@ -52,14 +53,15 @@ import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
 import java.security.PublicKey
 import java.time.Instant
-import java.util.UUID
+import java.util.*
```
anien: This wouldn't pass detekt

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt
```diff
@@ -45,6 +46,8 @@ const val PRIVATE_KEY_ENCODING_VERSION: Int = 1
  *
  * @param wrappingRepositoryFactory which provides a factory for [WrappingRepository], which provides save and
  *        find operations for wrapping keys on a specific tenant.
+ * @param signingRepositoryFactory which provides a factory for [SigningRepository], which provides save and
+ *        find operations for signing keys fon a specific tenant.
```
anien: typo: fon -> for

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -166,66 +182,125 @@ class CryptoProcessorImpl @Activate constructor(
...
+            }
+        )
+    }
+
+    private fun startHSMService() {
+        hsmService = HSMServiceImpl(hsmStore, cryptoService)
+        CryptoConsts.Categories.all.forEach { category ->
+            CryptoTenants.allClusterTenants.forEach { tenantId ->
+                hsmService.assignSoftHSM(tenantId, category)
+                logger.trace("Assigned SOFT HSM for $tenantId:$category")
```
anien, line 285: Is there a reason why we don't track failed associations anymore?

processors/crypto-processor/src/test/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImplTest.kt
```diff
@@ -26,19 +33,27 @@ import net.corda.processors.crypto.CryptoProcessor
...
 import org.mockito.kotlin.whenever
+import java.util.*
 
 class CryptoProcessorImplTest {
 
     companion object {
+        // Tempting to use SmartConfig here even from unit tests, but probably better to mock it out
+        // since SmartConfig is Corda corda. Or could we use TypeSafeConfig directly heere?
+        // TODO remove hard coded strings e.g. "expireAfterAccessMins" and instead use string constants
+        //   from CryptoConfigUtils exclusively
```
anien, line 56: As I mentioned privately, I am more keen on having tickets for TODO comments to they can be tracked better and prioritise in the work. But as it's still in a feature branch, I am okay to leave it for now.

processors/crypto-processor/src/test/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImplTest.kt
```diff
@@ -61,8 +76,18 @@ class CryptoProcessorImplTest {
...
                 whenever(it.getConfig("waitBetweenMills")).thenReturn(waitBetweenMillsConfig)
             }
+            val wrappingKeysList = listOf(mock<SmartConfig>().also {
+                whenever(it.getString(ALIAS)).thenReturn("alpha")
+                whenever(it.getString(SALT)).thenReturn("speak")
+                whenever(it.getString(PASSPHRASE)).thenReturn("friend")
+            })
+            val hsmConfig = mock<SmartConfig>().also {
+                whenever(it.getConfigList(eq(WRAPPING_KEYS))).thenReturn(wrappingKeysList)
+                whenever(it.getString(DEFAULT_WRAPPING_KEY)).thenReturn("alpha")
```
anien: Do you want to have the same string for `ALIAS` and `DEFAULT_WRAPPING_KEY`?

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/SigningServiceGeneralTests.kt
dickon: Looks like we don't run detekt on this file. I didn't know that!

The defualt WildcardImport exclude java.util.*, see https://detekt.dev/docs/rules/style/#wildcardimport

dickon: However the change is pointless and unintended, so I've reverted it.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt
dickon: fixed

processors/crypto-processor/src/test/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImplTest.kt
dickon: Yes, it has to be. The default wrapping key needs to be one of the keys in wrappingKeysList, and there's only one so it has to match.

dickon: I added a comment to explain.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
dickon: We now only do the soft HSM association, not hardware HSM assocations. That's due to a change in 5.1 I include (but it was missing from the PR description,  and I've put that right). If the soft HSM association fails, we'll throw an exception here and fail to configure and declare outselves as up, so it would be tracked and handled as any other OSGi lifecycle component startup error.

anien: Thanks for the explanation!

anien (dismissed):
LGTM.

anien (approved):
LGTM.

PR #4107, opened 2023-06-19
CORE-11763 - Remove explicit references to version number for fake-ca tooling

dimosr:
Remove explicit references to version number for fake-ca tooling.


charlieR3 (approved):
LGTM

PR #4108, opened 2023-06-19
CORE-14709: Adding `apiVersion` dimension to REST API

vkolomeyko:
Initial implementation of the [REST API versioning design](https://github.com/corda/platform-eng-design/blob/master/core/corda-5/corda-5.0/rpc/rest-api-versioning/index.md).


gradle.properties
```diff
@@ -114,7 +114,7 @@ slingVersion=3.3.6
 
 # REST dependency versions
 javalinVersion = 4.6.7
-swaggerVersion = 2.2.9
+swaggerVersion = 2.2.12
```
vkolomeyko, line 120: Strictly speaking, this change is unrelated, but it was hooking-up and old version of Jackson library which in turn was hooking old version of SnakesYaml library which we (correctly) do not have in our dependency graph and this prevented `OpenApiCompatibilityTest` from running locally.

libs/rest/rest-client/src/integrationTest/kotlin/net/corda/rest/client/RestIntegrationTestBase.kt
```diff
@@ -10,6 +11,7 @@ abstract class RestIntegrationTestBase {
         fun isServerInitialized() = ::server.isInitialized
         val userAlice = User(FakeSecurityManager.USERNAME, FakeSecurityManager.PASSWORD, setOf())
         val securityManager = FakeSecurityManager()
-        val context = RestContext("1", "api", "RestContext test title ", "RestContext test description")
+        val context = RestContext("api", "RestContext test title ", "RestContext test description")
+        val apiVersion = RestApiVersion.C5_1
```
josephzunigadaly, line 15: I would like to suggest turning this into a list of currently supported versions and then adding a version parameter to the tests using this as the data source. So JUnit calls the test methods multiple times, once for each version.

libs/rest/rest/src/main/kotlin/net/corda/rest/annotations/RestApiVersion.kt
```diff
@@ -0,0 +1,15 @@
...
+ * It must be updated with any new Corda version for new path to become available.
+ */
+enum class RestApiVersion(val versionPath: String, val parentVersion: RestApiVersion?) {
+
+    C5_0("v1", null),
+    C5_1("v5_1", C5_0),
+    C5_2("v5_2", C5_1),
+
+    // Whenever new version is added, please re-visit `HttpResource.kt` and `RestEndpoint.kt` to change default
+    // version aliases.
```
josephzunigadaly, line 14: Can we implement toString to return `versionPath`?

I believe that would allow us to use:
```kotlin
"${restServerSettings.context.basePath}/$apiVersion/"
```
instead of:
```kotlin
"${restServerSettings.context.basePath}/${apiVersion.versionPath}/"
```

libs/rest/rest/src/main/kotlin/net/corda/rest/annotations/RestEndpoint.kt
```diff
@@ -4,6 +4,9 @@ import net.corda.rest.ResponseCode
 import net.corda.rest.exception.ResourceNotFoundException
 import net.corda.rest.response.ResponseEntity
 
+import net.corda.rest.annotations.RestApiVersion.C5_0 as MIN_SUPPORTED
+import net.corda.rest.annotations.RestApiVersion.C5_1 as CURRENT
```
josephzunigadaly, line 8: This is here and in `libs/rest/rest/src/main/kotlin/net/corda/rest/annotations/HttpRestResource.kt`

Should we add some constants for min supported and current so we can share values across these files?

josephzunigadaly (approved):
Looks good. I left some comments but I'd happy for this PR to go in and the comments considered for future work instead of this PR.

libs/rest/rest-client/src/integrationTest/kotlin/net/corda/rest/client/RestIntegrationTestBase.kt
vkolomeyko: Yes, I was thinking of doing it in the PR to follow. We can even employ `@ParameterizedTest` feature for that.
This will be especially important when our REST API starts to diverge from one version to another.

For now, for most of the tests while porting them, I used `v1` in the URL path and for some (like this one) I have actually made a switch to version 5.1, i.e. path `v5_1`.

libs/rest/rest/src/main/kotlin/net/corda/rest/annotations/RestApiVersion.kt
vkolomeyko: As discussed, I am a bit cautious about using `toString()` method to drive application logic. Hence, I though it would be better to use `apiVersion.versionPath` explicitly. Especially given that there are not too many places where it is being used.

libs/rest/rest/src/main/kotlin/net/corda/rest/annotations/RestEndpoint.kt
vkolomeyko: Yes, I did try it:
image
looks like it is not possible to create a `const` of type `RestApiVersion` :-(

Making it just `val` is of course possible, but then we cannot use as a default value for annotations as it has to be a compile-time constant.

Import alias is a best option I could get so far.

dimosr (approved):
approving for @corda/corda5-team-leads - owned files

PR #4111, opened 2023-06-19
CORE-14749 Skip event bus for events for the same topic/key

mbrkic-r3:
Allow Subscription Processors to feed subset of events back into itself without going to bus first


JamesHR3 (commented):
This looks to me like it should do the right sort of thing - essentially adding the new events into the current processing batch.

PR #4112, opened 2023-06-20
CORE-13565 Add extra datasource properties to db bootstrap config

kyriathar:
* adds extra datasource properties to Cluster DB config values and passes them in to DB aware workers start up arguments (for Crypto & DB workers)
* adds extra datasource properties to DB config values and passes them in to `create-db-config` CLI command (for `CRYPTO` & `RBAC` schemas)

### DIFF
--- a/charts/corda-lib/templates/_bootstrap.tpl
+++ b/charts/corda-lib/templates/_bootstrap.tpl
@@ -516,7 +516,9 @@ a second init container to execute the output SQL to the relevant database
              {{- " '--name'" -}}, 'corda-{{ .longName | default .name }}', 
              {{- " '--jdbc-url'" -}}, 'jdbc:{{ include "corda.clusterDbType" . }}://{{ required "A db host is required" .Values.db.cluster.host }}:{{ include "corda.clusterDbPort" . }}/{{ include "corda.clusterDbName" . }}{{- if .schema }}?currentSchema={{.schema }}{{- end -}}', 
              {{- " '--jdbc-pool-max-size'" -}}, {{ (index .Values.bootstrap.db (.dbName | default .name)).dbConnectionPool.maxSize | quote }},
-             {{- " '--jdbc-pool-min-size'" -}}, {{ (index .Values.bootstrap.db (.dbName | default .name)).dbConnectionPool.minSize | quote }},
+             {{- with (index .Values.bootstrap.db (.dbName | default .name)).dbConnectionPool.minSize -}}
+                {{- " '--jdbc-pool-min-size'" -}}, {{ . | quote }},
+             {{- end -}}
              {{- " '--idle-timeout'" -}}, {{ (index .Values.bootstrap.db (.dbName | default .name)).dbConnectionPool.idleTimeoutSeconds | quote }},
              {{- " '--max-lifetime'" -}}, {{ (index .Values.bootstrap.db (.dbName | default .name)).dbConnectionPool.maxLifetimeSeconds | quote }},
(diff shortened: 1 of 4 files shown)


davidcurrie (dismissed):
LGTM

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -516,6 +516,11 @@ a second init container to execute the output SQL to the relevant database
              {{- " '--name'" -}}, 'corda-{{ .longName | default .name }}', 
              {{- " '--jdbc-url'" -}}, 'jdbc:{{ include "corda.clusterDbType" . }}://{{ required "A db host is required" .Values.db.cluster.host }}:{{ include "corda.clusterDbPort" . }}/{{ include "corda.clusterDbName" . }}{{- if .schema }}?currentSchema={{.schema }}{{- end -}}', 
              {{- " '--jdbc-pool-max-size'" -}}, {{ (index .Values.bootstrap.db (.dbName | default .name)).dbConnectionPool.maxSize | quote }}, {{- " " -}}
+             {{- " '--jdbc-pool-min-size'" -}}, {{ (index .Values.bootstrap.db (.dbName | default .name)).dbConnectionPool.minSize | quote }}, {{- " " -}}
+             {{- " '--idle-timeout'" -}}, {{ (index .Values.bootstrap.db (.dbName | default .name)).dbConnectionPool.idleTimeoutSeconds | quote }}, {{- " " -}}
+             {{- " '--max-lifetime'" -}}, {{ (index .Values.bootstrap.db (.dbName | default .name)).dbConnectionPool.maxLifetimeSeconds | quote }}, {{- " " -}}
+             {{- " '--keepalive-time'" -}}, {{ (index .Values.bootstrap.db (.dbName | default .name)).dbConnectionPool.keepaliveTimeSeconds | quote }}, {{- " " -}}
```
davidcurrie: I don't think you need to explicitly add the space on the end of each argument given that you're including a space on the start of the next one.

davidcurrie (dismissed):
👍

davidcurrie (dismissed):
🚀

charts/corda/values.schema.json
```diff
@@ -752,6 +756,43 @@
...
+                                                {
+                                                    "type": "integer",
+                                                    "minimum": 0
+                                                },
+                                                {
+                                                    "type": "null"
+                                                }
+                                            ],
+                                            "default": null,
+                                            "title": "minimum JDBC connection pool size for crypto DB"
```
simon-johnson-r3, line 771: I think we should describe what `null` means

simon-johnson-r3: I think we should describe what `null` means
(and all occurrences below)

charts/corda/values.yaml
```diff
@@ -246,6 +246,16 @@ bootstrap:
       dbConnectionPool:
         # -- maximum JDBC connection pool size for crypto DB
         maxSize: 5
+        # -- minimum JDBC connection pool size for crypto DB
```
simon-johnson-r3: probably add a note about null here too, it's the first place most end users will encounter it

simon-johnson-r3 (changes requested):

charts/corda/values.schema.json
kyriathar: Good point. Addressed.

charts/corda/values.yaml
kyriathar: Done

simon-johnson-r3 (approved):

davidcurrie (approved):
LGTM

PR #4117, opened 2023-06-20
CORE-14145: Use group reader before going to db

yift-r3:
Adds some more logs to the membership persistence for asynchronous communication.
Ensure that we check the membership group reader (accessible from memory) before going to the persistence layer (for static registration).


components/membership/membership-persistence-client-impl/src/main/kotlin/net/corda/membership/impl/persistence/client/MembershipPersistenceOperationImpl.kt
```diff
@@ -116,5 +116,8 @@ internal class MembershipPersistenceOperationImpl<T>(
             key = request.context.requestId,
             value = MembershipPersistenceAsyncRequest(request),
         ),
-    )
+    ).also {
+        val requestId = request.context.requestId
```
charlieR3: Nit pick but is it necessary to assign to a `val`? 

```
logger.info("Sending async membership persistence RPC request ID: ${request.context.requestId}.")
```

charlieR3 (dismissed):
One minor comment

charlieR3 (approved):
LGTM

PR #4118, opened 2023-06-20
CORE-14188: Add a tag to class cache to SandboxGroupImpl.getClass(String, String)

josephzunigadaly:
PR to run tests and see what effect this change has.

### DIFF
--- a/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
+++ b/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
@@ -105,6 +105,7 @@ internal class SandboxGroupImpl(
 
     private val staticTagCache = ConcurrentHashMap<Class<*>, String>()
     private val evolvableTagCache = ConcurrentHashMap<Class<*>, String>()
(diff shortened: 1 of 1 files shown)


libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -114,48 +115,50 @@ internal class SandboxGroupImpl(
...
-                    throw SandboxException(
-                        "Class $className was not from a bundle, and could not be found in the system classloader."
-                    )
+        return reverseTagCache.computeIfAbsent((className to serialisedClassTag)){
+            val classTag = classTagFactory.deserialise(serialisedClassTag)
+
+            when (classTag.classType) {
+                ClassType.NonBundleClass -> {
+                    try {
+                        primitiveTypes[className] ?: bundleUtils.loadClassFromSystemBundle(className)
```
chrisr3, line 126: FYI this will conflict when merged into 5.1 because `primitiveTypes` has been moved into
```kotlin
bundleUtils.loadClassFromSystemBundle(className)
```
Hence the correct merge for 5.1 will be:
```kotlin
try {
    bundleUtils.loadClassFromSystemBundle(className)
} catch (e: ClassNotFoundException) {
```

PR #4119, opened 2023-06-20
CORE-12500 Combine TestGroupPolicyProviders & write integration test …

jennyang-r3:
# Overview
This PR does two different things:
- combine the three similar `TestGroupPolicyProvider`
- write a persistence integration test for new methods `persistSignedGroupParametersIfDoNotExist` and `findSignedGroupParameters` which we previously didn't have a test for in CORE-7269 and verified it's working


components/membership/group-policy-test-common/build.gradle
```diff
@@ -0,0 +1,16 @@
+plugins {
+    id 'corda.common-publishing'
```
vlajos: If I remember right we do not want to publish test modules. So this publishing plugin is not needed.

components/membership/group-policy-test-common/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+    id 'corda.common-library'
+}
+
+description 'Group policy provider component'
```
vlajos:
```suggestion
description 'Group policy provider fake component for testing'
```
or something similar?

vlajos: Probably the whole module could be moved from `components/membership/` to `testing/`...
The testing modules lives there.

components/membership/registration-impl/src/integrationTest/kotlin/net/corda/membership/impl/registration/MemberRegistrationIntegrationTest.kt
```diff
@@ -294,10 +294,10 @@ class MemberRegistrationIntegrationTest {
 
     @Test
     fun `dynamic member registration service publishes unauthenticated message to be sent to the MGM`() {
-        groupPolicyProvider.putGroupPolicy(TestGroupPolicy())
-
+        val groupPolicy = TestGroupPolicy()
         val member = HoldingIdentity(memberName, groupId)
         val context = buildTestContext(member)
+        groupPolicyProvider.putGroupPolicy(member, groupPolicy)
```
nikinagy: You can just do this, since groupPolicy is a one-time used val:
```suggestion
        groupPolicyProvider.putGroupPolicy(member, TestGroupPolicy())
```

ronanbrowne (dismissed):
LGTM (only gradle change reviewed)

nikinagy (dismissed):
LGTM (platform-network team owned code)

lankydan (approved):

nikinagy (approved):
LGTM (network)

ronanbrowne (approved):

PR #4120, opened 2023-06-20
CORE-14552: Throw 409 when upgrading vNode with cpi that is already installed

Tom-Fitzpatrick:
This adds a validation to the virtual node upgrade process to check that the Checksum is different to the currently installed one. It throws a 409 if the user is trying to upgrade to the same CPI. 

A corresponding test is being added in corda-e2e-tests:
https://github.com/corda/corda-e2e-tests/pull/130


josephzunigadaly:
It's better to return success like it was the first time. The desired state has been achieved. Returning an error gives users something to investigate when all is well.

vkolomeyko:
> It's better to return success like it was the first time. The desired state has been achieved. Returning an error gives users something to investigate when all is well.

I guess it is a debate to have. I sort of agree with what QA said on the ticket, if we are trying to upgrade to CPI version we are already on, it is a NOP and arguably the caller might have had a different CPI in mind, hence the error code returned.
But I do not have a strong opinion here as long as we are consistent across most of our REST API methods.

vkolomeyko (approved):
LGTM

PR #4126, opened 2023-06-21
CORE-14719 - New version of config schema caused failure on platform upgrade

nikinagy:
Config schema changes were not picked up upon platform upgrade. We will use a default value until a forced upgrade happens via the config reconciler which results in values being received to configure the ExpirationProcessor.

Testing:
1. Run a deployment in 5.0.
2. Do platform upgrade to 5.1.
3. Check the logs: member worker runs okay and processor could start:
log1
4. Check the current config in DB (it should have old values at this point):
lookup1
5. Update the config with empty value to force the reconciler to run:
update
6. Check the current config in DB (it should have the new values at this point as defined in the api project):
lookup2
7. Check the logs to see that ExpirationProcessor receive the config and restarted:
log2

### DIFF
--- a/components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/mgm/ExpirationProcessorTest.kt
+++ b/components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/mgm/ExpirationProcessorTest.kt
@@ -268,15 +268,15 @@ class ExpirationProcessorTest {
                 it.assertThat(capturedTimeframes.firstValue).isLessThanOrEqualTo(TimeUnit.MINUTES.toMillis(180))
                 it.assertThat(capturedTimeframes.secondValue).isLessThanOrEqualTo(TimeUnit.MINUTES.toMillis(dummyTime))
 
-                val declineEvent = capturedEvents.firstValue.invoke("")
-                it.assertThat(declineEvent).isInstanceOf(ExpirationProcessorImpl.DeclineExpiredRegistrationRequests::class.java)
-                val firstEvent = declineEvent as ExpirationProcessorImpl.DeclineExpiredRegistrationRequests
-                it.assertThat(firstEvent.expirationDate).isEqualTo(TimeUnit.MINUTES.toMillis(300))
+                val firstEvent = capturedEvents.firstValue.invoke("")
+                it.assertThat(firstEvent).isInstanceOf(ExpirationProcessorImpl.DeclineExpiredRegistrationRequests::class.java)
+                val declineEvent = firstEvent as ExpirationProcessorImpl.DeclineExpiredRegistrationRequests
+                it.assertThat(declineEvent.expirationDate).isEqualTo(TimeUnit.MINUTES.toMillis(300))
 
-                val declineEventAfterUpdate = capturedEvents.secondValue.invoke("")
-                it.assertThat(declineEventAfterUpdate).isInstanceOf(ExpirationProcessorImpl.DeclineExpiredRegistrationRequests::class.java)
-                val secondEvent = declineEventAfterUpdate as ExpirationProcessorImpl.DeclineExpiredRegistrationRequests
-                it.assertThat(secondEvent.expirationDate).isEqualTo(TimeUnit.MINUTES.toMillis(dummyTime))
+                val secondEvent = capturedEvents.secondValue.invoke("")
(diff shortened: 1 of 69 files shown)


components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/mgm/ExpirationProcessorTest.kt
```diff
@@ -244,6 +246,40 @@ class ExpirationProcessorTest {
...
+            // new config received
+            handler.firstValue.processEvent(configChangedEvent, coordinator)
+
+            verify(coordinator, times(2)).setTimer(any(), capturedTimeframes.capture(), capturedEvents.capture())
+
+            SoftAssertions.assertSoftly {
+                it.assertThat(capturedTimeframes.firstValue).isLessThanOrEqualTo(TimeUnit.MINUTES.toMillis(180))
+                it.assertThat(capturedTimeframes.secondValue).isLessThanOrEqualTo(TimeUnit.MINUTES.toMillis(dummyTime))
+
+                val lambda = capturedEvents.firstValue.invoke("")
```
charlieR3: Maybe some name like this would be better? If I'm reading it right this value isn't a lambda
```suggestion
                val declineEvent = capturedEvents.firstValue.invoke("")
```

components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/mgm/ExpirationProcessorTest.kt
```diff
@@ -244,6 +246,40 @@ class ExpirationProcessorTest {
...
+                it.assertThat(lambda).isInstanceOf(ExpirationProcessorImpl.DeclineExpiredRegistrationRequests::class.java)
+                val firstEvent = lambda as ExpirationProcessorImpl.DeclineExpiredRegistrationRequests
+                it.assertThat(firstEvent.expirationDate).isEqualTo(TimeUnit.MINUTES.toMillis(300))
+
+                val lambdaAfterUpdate = capturedEvents.secondValue.invoke("")
```
charlieR3: Same here. Naming it lambda is a bit confusing (if I'm reading this right)

charlieR3 (dismissed):
LGTM. Two nit picky comments in tests but looks good anyway.

components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/mgm/ExpirationProcessorTest.kt
nikinagy: I am not sure that is the best name here as the passed lambda is being captured here and event is gotten from it after calling invoke on it. But I am open to any other better naming suggestions.

charlieR3: Seems like `lambdaAfterUpdate` is the result of invoking the lambda rather than the lambda itself so it was a little confusing to read. It's very minor so I won't hold it up for this. Something like `declineExpiredEventAfterUpdate` better reflects what it relates to (although it's a bit long)

nikinagy: ah ok, never mind it's very Friday, you are right

components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/mgm/ExpirationProcessorTest.kt
nikinagy: done in [5759425](https://github.com/corda/corda-runtime-os/pull/4126/commits/5759425a69bee48824ca7724fff6528f426f050f)

components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/mgm/ExpirationProcessorTest.kt
nikinagy: done in [5759425](https://github.com/corda/corda-runtime-os/pull/4126/commits/5759425a69bee48824ca7724fff6528f426f050f)

charlieR3 (dismissed):

dimosr (dismissed):

charlieR3 (approved):
LGTM

dimosr (approved):

PR #4127, opened 2023-06-22
CORE-14145: Use group reader before going to db (#4117)

yift-r3:
* CORE-14145: Use group reader before going to db

* Apply review comments


dimosr (approved):
already reviewed and approved on 5.1

driessamyn (approved):
LGTM

PR #4128, opened 2023-06-22
CORE-14658: check content was found when parsing PEM format keys, to turn null pointer exception into a useful exception

dickon:
Introduce a test which is simply paring a PEM file without a valid key within it. That currently throws null pointer exception, since we deference the content without checking anything was foud. Instead, throw IllegalArgumentException which seems to be consensus exception type for bad input.

### DIFF
--- a/libs/crypto/cipher-suite-impl/src/test/kotlin/net/corda/cipher/suite/impl/CipherSchemeMetadataTests.kt
+++ b/libs/crypto/cipher-suite-impl/src/test/kotlin/net/corda/cipher/suite/impl/CipherSchemeMetadataTests.kt
@@ -552,21 +552,9 @@ class CipherSchemeMetadataTests {
     
     @Test
     fun `Should fail gracefully with a correupted PEM file`() {
-        // note deliberate typo "BENIGN" rather than "BEGIN"
-        val bad = """-----BENIGN PUBLIC KEY-----
-                MIIBojANBgkqhkiG9w0BAQEFAAOCAY8AMIIBigKCAYEAtkTc9q3S7hVqHzlMY7gl
-        KaJyfWhajLBvWO56VGci4udvnZuAZnLJDPjDgESPFQpm7NMXceZ6czXyf2t5FQ98
-        chtJbCX9/vHUK2lKueL+mK8seOuseSwCnDkU0qxRYFX1Im0p3oKefzrkDgCiQKYt
-        XU7tKUekaJF5J/nrVZYNVPpvaM9uLjOK2Ajwt/qlS3lIoeD1prDNUAe1mKxk+Pol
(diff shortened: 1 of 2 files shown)


libs/crypto/cipher-suite-impl/src/test/kotlin/net/corda/cipher/suite/impl/CipherSchemeMetadataTests.kt
```diff
@@ -549,6 +549,14 @@ class CipherSchemeMetadataTests {
         )
     }
 
+    
+    @Test
+    fun `Should fail gracefully with a correupted PEM file`() {
```
Sakpal: Minor: `correupted` spelling 😬

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/CipherSchemeMetadataProvider.kt
```diff
@@ -218,7 +218,7 @@ class CipherSchemeMetadataProvider : KeyEncodingService {
     private fun parsePemContent(pem: String): ByteArray =
         StringReader(pem).use { strReader ->
             return PemReader(strReader).use { pemReader ->
-                pemReader.readPemObject().content
+                pemReader.readPemObject()?.content?: throw IllegalArgumentException("Key not found in PEM file")
```
Sakpal: Curious: Do we explicitly avoid using the keyword `it`?

Sakpal: Minor: `Key not found in PEM file` this message is very explicit in saying missing key in the file. If I pass a JSON file, will the error message be still correct?

Alternative, shall we drop a generic exception message in this case?

dickon: @Sakpal this one is nested, so the `it` would be ambigious. Kotlin probably does have a rule to work it out but it's more clear. (Not that I wrote the original)

dickon: And I've generalised the message to say format rather than file now.

libs/crypto/cipher-suite-impl/src/test/kotlin/net/corda/cipher/suite/impl/CipherSchemeMetadataTests.kt
dickon: fixed

dickon:
@Peter Mitcham has confirmed on slack that QA would like my fix that turns the null pointer exception in to IllegalArgumentException("Key not found in PEM format") .

driessamyn (approved):
LGTM

PR #4134, opened 2023-06-22
CORE-14305 Interop - Invoke Facade Proxy from within a responder flow

szymonsztuka:
Invoke Facade Proxy from within a responder flow.
A subflow can be annotated with @InitiatingFlow, then a responder flow can starting a new session, and the infrastructure code will find a protocol name (which is anyway not used, but the presence is required).

Tested by running `com.r3.corda.demo.interop.tokens.workflows.interop.SwapFlow` flow. 
The run exposes the other issue, which is addressed by the ticket CORE-14306  ( Unable to serialize Java Lambda expression, unless explicitly declared e.g., Runnable r = (Runnable & Serializable) () -> System.out.println("Hello world!");)


jmacmahonr3 (approved):
LGTM

PR #4142, opened 2023-06-23
CORE-14869 & CORE-14872 Fix retry of initial registration

charlieR3:
Retrying an initial registration after it had been declined was not working since the MGM created a pending member info when it processed the initial registration the first time and the logic in place assumed that during the second attempt, the name was already taken by a different member. 
This change should allow a new registration to go through if the previous version of the member info was never approved.

An additional problem was found after resolving that which was that the member signature verification failed after member list distribution. This was because the persisted member info has non updateable fields around the member signatures and so the old member signatures were stored after persisting pending member info again for the same serial number. These are being switched to updateable now in line with what has been done in 5.1 as part of the re-registration work.

### DIFF
--- a/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistMemberInfoHandler.kt
+++ b/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistMemberInfoHandler.kt
@@ -61,10 +61,9 @@ internal class PersistMemberInfoHandler(
                         ),
                         LockModeType.PESSIMISTIC_WRITE
                     )
-                    val updatingPendingInfo = oldMemberInfo?.status == MEMBER_STATUS_PENDING
-                            && newMemberInfo.status == MEMBER_STATUS_PENDING
-                    val updatingSerialNumber = oldMemberInfo?.serialNumber != newMemberInfo.serial
-                    if (!updatingSerialNumber && !updatingPendingInfo) {
+                    val newPendingVersion = newMemberInfo.status == MEMBER_STATUS_PENDING
+                            && oldMemberInfo?.status == MEMBER_STATUS_PENDING
+                    if (!newPendingVersion && oldMemberInfo?.serialNumber == newMemberInfo.serial) {
                         val currentMemberContext = deserialize(oldMemberInfo.memberContext)
                         val currentMgmContext = deserialize(oldMemberInfo.mgmContext)
                         if (currentMemberContext.items != it.persistentMemberInfo.memberContext.items) {
@@ -77,17 +76,7 @@ internal class PersistMemberInfoHandler(
                         }
                         return@forEach
                     }
-                    else if (!updatingSerialNumber && updatingPendingInfo) {
-                        /**
-                         * If persisting a pending member info and the existing member info is pending with the same
-                         * serial number, delete the existing one and add the new pending info instead. This is because
-                         * the member info entity contains non-updatable fields such as the signature info, and we
-                         * cannot store new signature information as a result with removing the existing member info.
-                         * This can occur for example if a registration is declined by the mgm and submitted again by
-                         * the client.
-                         */
-                        em.remove(oldMemberInfo)
-                    }
+
                     val entity = MemberInfoEntity(
                         newMemberInfo.groupId,
                         newMemberInfo.name.toString(),
--- a/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistMemberInfoHandler.kt
+++ b/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistMemberInfoHandler.kt
@@ -45,45 +45,61 @@ internal class PersistMemberInfoHandler(
 
     override fun invoke(context: MembershipRequestContext, request: PersistMemberInfo) {
         if (request.members.isNotEmpty()) {
-            logger.info("Persisting member information.")
             transaction(context.holdingIdentity.toCorda().shortHash) { em ->
                 request.members.forEach {
-                    val memberInfo = memberInfoFactory.create(it.persistentMemberInfo)
-                    val currentMemberInfo = em.find(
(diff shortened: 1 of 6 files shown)


components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
```diff
@@ -137,14 +137,23 @@ internal class StartRegistrationHandler(
...
+             */
+            validateRegistrationRequest(
+                existingMemberInfos.isEmpty()
+                        || registrationRequest.serial != 0L
+                        || activeOrSuspendedInfo == null
+            ) { "Member already exists with the same X500 name." }
+
+            // Serial number on the request should be smaller than the current version of the requestor's MemberInfo
             validateRegistrationRequest(
                 activeOrSuspendedInfo == null || activeOrSuspendedInfo.serial <= registrationRequest.serial!!
```
dimosr: These boolean expressions are now becoming a bit hard to follow and error-prone. I was thinking it could help if we split logic explicitly between registration and re-registration, e.g. something like:
```
            validateRegistrationRequest(registrationRequest.serial != null) { "Serial cannot be null on the registration request." }
            validateRegistrationRequest(registrationRequest.serial!! < 0) { "Serial cannot be negative on the registration request." }
            if (registrationRequest.serial!! > 0) { //re-registration
                validateRegistrationRequest(activeOrSuspendedInfo != null) {
                    "Member has not registered previously."
                }
                validateRegistrationRequest(activeOrSuspendedInfo!!.serial <= registrationRequest.serial!!) {
                    "Registration request was submitted for an older version of member info. " +
                            "Please submit a new request."
                }
            } else if (registrationRequest.serial!! == 0L) { // initial registration
                validateRegistrationRequest(activeOrSuspendedInfo == null) { "Member already exists with the same X500 name." }
            }
```
Note: I assumed the serial is guaranteed to always be populated on the member side (by the platform, if not by the user), but not 100% sure if that's the case.

charlieR3: That's a good point. I'll look at that now.

charlieR3: I think serial always being set is a fair assumption. The member registration always sets a value.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistMemberInfoHandler.kt
```diff
@@ -45,45 +45,61 @@ internal class PersistMemberInfoHandler(
...
+                    else if (!updatingSerialNumber && updatingPendingInfo) {
+                        /**
+                         * If persisting a pending member info and the existing member info is pending with the same
+                         * serial number, delete the existing one and add the new pending info instead. This is because
+                         * the member info entity contains non-updatable fields such as the signature info, and we
+                         * cannot store new signature information as a result with removing the existing member info.
+                         * This can occur for example if a registration is declined by the mgm and submitted again by
+                         * the client.
+                         */
+                        em.remove(oldMemberInfo)
```
dimosr: I see serial number is not part of the primary for this entity here, so can there be similar issues with failed updates when that differs? 

I suppose in 5.0 it might not be an issue due to re-registration not being supported yet, but in general?

charlieR3: I'm not sure I fully follow. If the serial number differs it should not come in to this code block due to the `!updatingSerialNumber`. Does that answer the question or can you clarify it for me?

dimosr: Sure. The question was if the serial version number differs, does it actually need to go through similar logic, i.e. to remove the existing entry before adding a new one? Because otherwise, it would still attempt to update the non-updatable columns (and fail), right?

charlieR3: Interesting. Makes sense, but I'm not sure why we haven't seen the same exception during reregistration or suspension/activation in that case. Let me dig in to that a bit and see what's going on.

yift-r3: Shouldn't we keep the serial number as a primary key (so we can see the history of the member's info)? If not in this table, maybe in another one?

charlieR3: I was mistakenly comparing the 5.0 branch to the 5.1 branch to see why it worked OK for re-registration forgetting that that functionality is still on a feature branch. Looking at the feature branch I can see `updateable = false` was removed on the feature branch and that's why it works OK for re-registration. And it works OK for suspending and activating members because only the MGM context changes in that case and not the member context. 

The best solution in that case might be to allow those fields to be updateable rather than removing the previous row. 

In regards to changing the primary key, I'm not sure it's such a good idea to change that as a bug fix to 5.0. If we want historical member infos that might be considered new functionality and be better added and tested on 5.1.

dimosr: That makes sense. Making the field updatable seems to me like a smaller change and consistent with the 5.1 approach going forward, so preferable in this context. 

And I would agree that if we want to preserve historic member infos for audit purposes, that sounds like new functionality. Best to avoid introducing extra risk as part of this.

dimosr (approved):
LGTM

nikinagy (approved):
LGTM

ronanbrowne (approved):
LGTM to merge, based on being reviewed already / approved by product.

Note I've not actually _reviewed_ this kotlin code.

PR #4144, opened 2023-06-23
CORE-14675: Fixed vNode getting successfully upgraded with full hash

BenYip123:
Fixed the virtual node ID being incorrectly parsed causing an upgrade on the virtual node to work with the full hash when an error should be returned instead

Related to: [CORE-14675: Add test to check for error code 400 when using full hash of vNode id when upgrading vNode #133](https://github.com/corda/corda-e2e-tests/pull/133)


filipesoliveira (approved):
LGTM

vkolomeyko (approved):
LGTM

Tom-Fitzpatrick (approved):

PR #4147, opened 2023-06-23
ES-764: fix helm publishing job 

ronanbrowne:
Due to recent EKS infra changes related to expect node labels, the following modifications are requires

- `kubernetes-build-agent `updated to `gradle-build`
- yamlMergeStrategy added - this ensures correct yaml is generated for pod definitions 

Pipelines are blocked without this change, no functional application code changes.

kubernetes-build-agent label is no longer mapped to an agent as a result, when this job is kicked off, it can not acquire an Agent and just hangs.


conal-smith (approved):
LGTM

ronanbrowne:
@wzur-r3 FYI - another follow on here form previous changes

yarivus (approved):
LGTM

PR #4150, opened 2023-06-26
CORE-13126 remove initiator from list of members to finalize

vlajos:
The initiator should not call the finalization against itself, since it is unnecessary and makes the test slower.


blsemo (approved):
LGTM

PR #4152, opened 2023-06-26
CORE-14798 - Review context validation for re-registration

nikinagy:
Reregistration only works for the keys beginning with ext so members who registered on Corda 5.0 on a cluster which was then updated to Corda 5.1 cannot reregister since corda.softwareVersion changes.

**Testing**
1) Registered MGM and member on 5.0 cluster.
2) Did a platform upgrade using the image from this PR.
3) Re-registered member with additional field `ext.sample`:
Screenshot 2023-06-26 at 12 49 35
Lookup shows that member is on 50100 version after upgrade and has the new field in their member provided context:
Screenshot 2023-06-26 at 12 49 52
4) Re-registered member by using new protocol version for endpoints, the request got invalidated due to failing on the validation logic:
Screenshot 2023-06-26 at 12 51 00
5) Re-registered member by using an additional non-existing corda field `corda.sample`, the request got invalidated due to failing on the schema validation logic:
Screenshot 2023-06-26 at 12 51 45

### DIFF
--- a/components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
+++ b/components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
@@ -16,14 +16,18 @@ import net.corda.membership.impl.registration.dynamic.handler.MissingRegistratio
 import net.corda.membership.impl.registration.dynamic.handler.RegistrationHandler
 import net.corda.membership.impl.registration.dynamic.handler.RegistrationHandlerResult
 import net.corda.membership.impl.registration.dynamic.verifiers.RegistrationContextCustomFieldsVerifier
-import net.corda.membership.lib.MemberInfoExtension
 import net.corda.membership.lib.MemberInfoExtension.Companion.CREATION_TIME
 import net.corda.membership.lib.MemberInfoExtension.Companion.CUSTOM_KEY_PREFIX
+import net.corda.membership.lib.MemberInfoExtension.Companion.ENDPOINTS
+import net.corda.membership.lib.MemberInfoExtension.Companion.LEDGER_KEYS
 import net.corda.membership.lib.MemberInfoExtension.Companion.MEMBER_STATUS_ACTIVE
 import net.corda.membership.lib.MemberInfoExtension.Companion.MEMBER_STATUS_PENDING
 import net.corda.membership.lib.MemberInfoExtension.Companion.MEMBER_STATUS_SUSPENDED
 import net.corda.membership.lib.MemberInfoExtension.Companion.MODIFIED_TIME
+import net.corda.membership.lib.MemberInfoExtension.Companion.NOTARY_KEYS
+import net.corda.membership.lib.MemberInfoExtension.Companion.ROLES_PREFIX
 import net.corda.membership.lib.MemberInfoExtension.Companion.SERIAL
+import net.corda.membership.lib.MemberInfoExtension.Companion.SESSION_KEYS
 import net.corda.membership.lib.MemberInfoExtension.Companion.STATUS
 import net.corda.membership.lib.MemberInfoExtension.Companion.endpoints
 import net.corda.membership.lib.MemberInfoExtension.Companion.groupId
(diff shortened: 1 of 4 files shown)


components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
```diff
@@ -159,10 +162,17 @@ internal class StartRegistrationHandler(
...
+                                it.key.startsWith(SESSION_KEYS) ||
+                                it.key.startsWith(LEDGER_KEYS) ||
+                                it.key.startsWith(ROLES_PREFIX) ||
+                                it.key.startsWith("corda.notary")
+                    }
                 validateRegistrationRequest(
                     diff.isEmpty()
-                ) { "Only custom fields with the '$CUSTOM_KEY_PREFIX' prefix may be updated during re-registration." }
+                ) { "Only custom fields with the '$CUSTOM_KEY_PREFIX' prefix and cpi/platform information " +
+                        "may be updated during re-registration." }
```
dimosr: Given we have an explicit allow list now, should we adjust the message to indicate what exactly was invalid? e.g. something like
```
Fields ${diff.map { it.key }} cannot be added, removed or updated during re-registration.
```
Similarly in `DynamicMemberRegistrationService`.

components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationServiceTest.kt
```diff
@@ -1002,12 +1013,13 @@ class DynamicMemberRegistrationServiceTest {
         }
 
         @Test
-        fun `registration fails when non-custom properties are updated`() {
+        fun `registration fails when non-custom, non-platform or non-cpi related properties are added`() {
```
dimosr, line 1016: is this testing addition as the title implies or update of a field? Seems like endpoint fields are present in `previousRegistrationContext` too, so it's testing updates.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
```diff
@@ -159,10 +162,17 @@ internal class StartRegistrationHandler(
...
                 val pendingContext = pendingMemberInfo.memberProvidedContext.toMap()
                 val diff = ((pendingContext.entries - previousContext.entries) + (previousContext.entries - pendingContext.entries))
-                    .filterNot { it.key.startsWith(CUSTOM_KEY_PREFIX) || it.key == REGISTRATION_ID }
+                    .filter {
+                        it.key.startsWith(ENDPOINTS) ||
+                                it.key.startsWith(SESSION_KEYS) ||
+                                it.key.startsWith(LEDGER_KEYS) ||
+                                it.key.startsWith(ROLES_PREFIX) ||
+                                it.key.startsWith("corda.notary")
+                    }
```
charlieR3, line 170: Does this allow these keys samples `foo.bar`, `corda.blah`?

nikinagy: I've only tried with 'corda.sample' (testing evidence step 5), it seemed like the JSON scheme validation handles it as we guessed the other day.

charlieR3: 🤦‍♂️I think that's the second time in a week I've made that mistake.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
nikinagy: addressed in [dcd46f5](https://github.com/corda/corda-runtime-os/pull/4152/commits/dcd46f5f5553fea24cef5415fc2f39c45beaefc4)

components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationServiceTest.kt
nikinagy: Added new role information instead in [dcd46f5](https://github.com/corda/corda-runtime-os/pull/4152/commits/dcd46f5f5553fea24cef5415fc2f39c45beaefc4)

dimosr (approved):
LGTM

PR #4160, opened 2023-06-26
CORE-14355 Interop/ Facade method returns unwrapped result

szymonsztuka:
Simplify the object returns by a method call on a proxy by removing wrapping generic type `InteropAction<T>` and returned the matching type as specified in Facade JSON file.

The typical usage before the change was:
For Façade JSON specification the returned type:
```…
"out": {
  "reservation-ref": "uuid"
}
```
An interface needed to have InteropAction type for each returned type wrapping the actual type:

`fun reserveTokens(denomination: String, amount: BigDecimal, ttlMs: Long): InteropAction<UUID>`

In the Cordapp, to retrieve the actual result, a method/getter “result” would need to be called on the object implementing InteropAction, e.g:

```
val responseObject = client.reserveTokensV1("USD", BigDecimal(100))        
val response: UUID  = responseObject.result
```

After the change, the interface methods returns directly the type from Façade specification:
`fun reserveTokens(denomination: String, amount: BigDecimal, ttlMs: Long): UUID`

In the Cordapp there is no longer an intermidiare “resoponse object” holding the actual response value:

`val response : UUID = client.reserveTokensV1("USD", BigDecimal(100))    `    

Mehtods which doesn’t return any value can be specied as returning Unit or nothing, e.g.:
Before:
   
` fun releaseReservedTokens(reservationRef: UUID): InteropAction<Unit>`

now:

` fun releaseReservedTokens(reservationRef: UUID)`

Cord API changes in  https://github.com/corda/corda-api/pull/1168

The removal of `ClientAction<T>` resolved issue CORE-14306 "Allow Kryo Serialisation of ClientAction in a responder flow".


yarivus (approved):
LGTM

julia-filipczak (approved):

PR #4165, opened 2023-06-27
CORE-15062: Remove special SerializationService handling for FLOW sandboxes.

chrisr3:
Convert `SandboxSerializationInternal` service into a platform service, since it obtains its sandbox `SerializationService` using platform service `CurrentSandboxGroupContext`, and has no sandbox services as dependencies.

This means that we can create `FLOW` sandboxes' `SerializationService` in the same way as `PERSISTENCE` and `VERIFICATION` sandboxes.

It also means that flows can now only `@CordaInject` a `SerializationService` instance and not a `SerializationServiceInternal`, but this is something that can only be relevant to flows which are also `@CordaSystemFlow` anyway.

---
Since only `FLOW` sandboxes currently support `@CordaInject`, we can disable this support for `PERSISTENCE` and `VERIFICATION` sandboxes.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/serialization/SerializationServiceInternalImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/serialization/SerializationServiceInternalImpl.kt
@@ -7,6 +7,7 @@ import net.corda.sandboxgroupcontext.getObjectByKey
 import net.corda.utilities.reflection.castIfPossible
 import net.corda.v5.application.serialization.SerializationService
 import net.corda.v5.serialization.SerializedBytes
+import net.corda.v5.serialization.SingletonSerializeAsToken
 import org.osgi.service.component.annotations.Activate
 import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
@@ -14,17 +15,14 @@ import org.slf4j.LoggerFactory
 import java.io.NotSerializableException
 
 /**
- * This is a platform service that uses the current sandbox's [SerializationService].
+ * This is a platform singleton service that uses the current sandbox's [SerializationService].
  */
-@Component(service = [SerializationServiceInternal::class])
+@Component(service = [SerializationServiceInternal::class, SingletonSerializeAsToken::class])
 class SerializationServiceInternalImpl @Activate constructor(
     @Reference(service = CurrentSandboxGroupContext::class)
     private val currentSandboxGroupContext: CurrentSandboxGroupContext
-) : SerializationServiceInternal {
-
-    private companion object {
-        private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
-    }
+) : SerializationServiceInternal, SingletonSerializeAsToken {
+    private val log = LoggerFactory.getLogger(this::class.java)
 
     private val serializationService
         get(): SerializationService {
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/serialization/SerializationServiceInternalImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/serialization/SerializationServiceInternalImpl.kt
@@ -1,31 +1,26 @@
 package net.corda.flow.application.serialization
 
 import net.corda.flow.pipeline.exceptions.FlowFatalException
-import net.corda.sandbox.type.SandboxConstants.CORDA_SYSTEM_SERVICE
-import net.corda.sandbox.type.UsedByFlow
 import net.corda.sandboxgroupcontext.CurrentSandboxGroupContext
 import net.corda.sandboxgroupcontext.RequireSandboxAMQP.AMQP_SERIALIZATION_SERVICE
 import net.corda.sandboxgroupcontext.getObjectByKey
 import net.corda.utilities.reflection.castIfPossible
 import net.corda.v5.application.serialization.SerializationService
 import net.corda.v5.serialization.SerializedBytes
-import net.corda.v5.serialization.SingletonSerializeAsToken
 import org.osgi.service.component.annotations.Activate
 import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
-import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
 import org.slf4j.LoggerFactory
 import java.io.NotSerializableException
 
-@Component(
-    service = [SerializationService::class, SerializationServiceInternal::class, UsedByFlow::class],
-    property = [CORDA_SYSTEM_SERVICE],
-    scope = PROTOTYPE
-)
-class SerializationServiceImpl @Activate constructor(
+/**
+ * This is a platform service that uses the current sandbox's [SerializationService].
+ */
+@Component(service = [SerializationServiceInternal::class])
+class SerializationServiceInternalImpl @Activate constructor(
     @Reference(service = CurrentSandboxGroupContext::class)
     private val currentSandboxGroupContext: CurrentSandboxGroupContext
-) : SerializationServiceInternal, UsedByFlow, SingletonSerializeAsToken {
+) : SerializationServiceInternal {
 
     private companion object {
         private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
--- a/components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/SerializationServiceProxy.kt
+++ b/components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/SerializationServiceProxy.kt
@@ -1,26 +1,33 @@
 package net.corda.sandbox.serialization.amqp
 
+import net.corda.sandbox.type.SandboxConstants.CORDA_SYSTEM_SERVICE
+import net.corda.sandbox.type.UsedByFlow
 import net.corda.sandbox.type.UsedByPersistence
 import net.corda.sandbox.type.UsedByVerification
 import net.corda.v5.application.serialization.SerializationService
 import net.corda.v5.serialization.SerializedBytes
+import net.corda.v5.serialization.SingletonSerializeAsToken
 import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
 
 /**
- * A [SerializationService] component for persistence and verification sandboxes.
- * Flow sandboxes already have a special "fiber-aware" component of their own.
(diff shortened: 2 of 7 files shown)


components/flow/flow-service/src/main/kotlin/net/corda/flow/application/serialization/SerializationServiceInternalImpl.kt
```diff
@@ -1,31 +1,26 @@
...
-class SerializationServiceImpl @Activate constructor(
+/**
+ * This is a platform service that uses the current sandbox's [SerializationService].
+ */
+@Component(service = [SerializationServiceInternal::class])
+class SerializationServiceInternalImpl @Activate constructor(
     @Reference(service = CurrentSandboxGroupContext::class)
     private val currentSandboxGroupContext: CurrentSandboxGroupContext
-) : SerializationServiceInternal, UsedByFlow, SingletonSerializeAsToken {
+) : SerializationServiceInternal {
```
lankydan: How does this work with `SingletonSerializeAsToken` being removed?

E.g. `FlowSessionImpl` which gets kryo serialized references `SerializationServiceInternal`.

lankydan (commented):
Generally looks fine to be, left one question though.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/serialization/SerializationServiceInternalImpl.kt
chrisr3: I think the answer to your question is _"Ze oops!"_ :blush:

libs/virtual-node/sandbox-group-context/src/main/kotlin/net/corda/sandboxgroupcontext/SandboxGroupType.kt
```diff
@@ -7,10 +7,14 @@ import net.corda.sandbox.type.UsedByVerification
...
  * Enumeration of various sandbox group types.
  */
-enum class SandboxGroupType(private val typeName: String, val serviceMarkerType: Class<*>) {
-    FLOW("flow", UsedByFlow::class.java),
-    VERIFICATION("verification", UsedByVerification::class.java),
-    PERSISTENCE("persistence", UsedByPersistence::class.java);
+enum class SandboxGroupType(
+    private val typeName: String,
+    val serviceMarkerType: Class<*>,
+    val hasInjection: Boolean
```
kyriathar, line 22: Maybe worth adding a few words to document what's this flag for.

components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/SerializationServiceProxy.kt
```diff
@@ -1,26 +1,33 @@
...
         SerializationServiceProxy::class,
+        UsedByFlow::class,
         UsedByPersistence::class,
         UsedByVerification::class
     ],
+    property = [ CORDA_SYSTEM_SERVICE ],
     scope = PROTOTYPE
 )
-class SerializationServiceProxy : SerializationService, UsedByPersistence, UsedByVerification {
+class SerializationServiceProxy
```
kyriathar, line 29: Maybe not for this PR because it will need aligning at use sites but should we consider renaming `SerializationServiceProxy` to `SerializationServiceImpl`? And perhaps `SerializationServiceInternal` could be renamed to `PlatformSerializationService` instead (This naming we used in `DigestService` and `PlatformDigestService`).

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
```diff
@@ -797,6 +803,14 @@ private fun Bundle.loadMetadataService(serviceClassName: String, isMetadataServi
     return null
 }
 
+@OptIn(ExperimentalContracts::class)
+private inline fun <T> T.takeIfElse(otherwise: T, predicate: (T) -> Boolean): T {
+    contract {
+        callsInPlace(predicate, InvocationKind.EXACTLY_ONCE)
+    }
+    return if (predicate(this)) this else otherwise
+}
```
kyriathar, line 812: I can't see what this function adds. Couldn't we use one of the build-in ones to remove this function overhead?

chrisr3: Kotlin's closest equivalent to this function is `takeIf {}`, except that returns `null` if `predicate` returns `false` and I need it to return `emptyList()` instead.

I wrote this function precisely _because_ Kotlin does not contain a suitable built-in function.

components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/SerializationServiceProxy.kt
kyriathar: I had a quick look on how we bootstrapping the sandbox serialization service so maybe the above does not apply after all.

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
kyriathar: Sure if you think it's better. I guess with `takeIf` it would be `.takeIf { sandboxGroupType.hasInjection } :? emptyList()`? I just don't see we will be re-using it. But fine to keep it if you think it looks better.

components/virtual-node/sandbox-amqp/src/main/kotlin/net/corda/sandbox/serialization/amqp/SerializationServiceProxy.kt
chrisr3: We already have `net.corda.internal.serialization.SerializationServiceImpl`, which is a _true_ implementation of `SerializationService`, although not an OSGi component.

OTOH `SerializationServiceProxy` is an OSGi component that proxies for an actual instance of `SerializationServiceImpl`. I therefore believe that its name is correct.

Both `SerializationServiceInternal` and its implementation are internal to the `flow-service`, so when I say "platform" I only mean that it isn't a `PROTOTYPE` sandbox service.

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
kyriathar: Also I think it's not very intuitive as the else (`otherwise`) comes before the if (`predicate`)

chrisr3: The `takeIfElse() {}` function is part of a chain:
```kotlin
.onEach { serviceType ->
    serviceIndex.computeIfAbsent(serviceType) { HashSet() }.add(serviceRef)
}.takeIfElse(emptyList()) {
    serviceGroupType.hasInjection
}.mapNotNullTo(ArrayList()) { serviceType ->
```
All my efforts to slot `takeIf {} ?: emptyList()` into that chain were unsuccessful, hence my `inline` function.

Switching the order of the `takeIfElse` parameters would result in:
```kotlin
.onEach { serviceType ->
    serviceIndex.computeIfAbsent(serviceType) { HashSet() }.add(serviceRef)
}.takeIfElse({
    serviceGroupType.hasInjection
}, emptyList()).mapNotNullTo(ArrayList()) { serviceType ->
```
which looks worse to me.

lankydan (approved):

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
kyriathar: One last try one this one.

So I think it could be written as `takeIf {` and then below `}?.mapNotNullTo` and `}?.takeIf` which is the same pattern you have previously used further below after `.takeIf` on line 285 (`}?.also`) I believe?

chrisr3: I _need_ the `takeIf {}` to return `emptyList()` instead of `null` because the chain _must_ reach this blocK:
```kotlin
}.takeIf { serviceTypes ->
    // Every service object must implement the service marker type
    // and at least one other type too, unless declared with the
    // corda.marker.only property.
    serviceTypes.isNotEmpty() || markerOnlyFilter.match(serviceRef)
```
And _must_ then be able to enter this block if `markerOnlyFilter.match(serviceRef)` is `true`:
```kotlin
}?.also { injectableTypes ->
    logger.debug("Identified injectable service {}, holding ID={}", serviceRef, vnc.holdingIdentity)
    injectableTypes += serviceMarkerType

    // We filtered on services having a component name, so we
    // should be guaranteed to find its component description.
    serviceComponentRuntime.getComponentDescriptionDTO(serviceRef)?.also { description ->
        injectables[serviceRef] = ServiceDefinition(injectableTypes, description)
    }
}
```
However, if `.takeIf { sandboxGroupType.hasInjection }` returns `null` then the chain ends, and that would be a bug.

I suspect you have simply made the code _compile_, and haven't tested whether it actually _works_. Try running:
```bash
$ cd components
$ ../gradlew clean testOSGi
```

kyriathar (approved):

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
kyriathar: Yes indeed I only made it to compile and haven't tested it. Thank you

ben-millar (approved):

PR #4166, opened 2023-06-27
CORE-12308 Introduce MGM historic session keys

YashNabar:
Introduces a new key for the MemberContext in the MGM’s MemberInfo, containing historic session keys. The MGM will advertise any historic session-initiation keys (currently an empty list, until session key rotation support is added) that might have signed group parameters in the past, to enable ledger code to verify transactions in the backchain.


libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/MemberInfoExtension.kt
```diff
@@ -205,19 +209,23 @@ class MemberInfoExtension {
...
-                }
-            }
+            get() = getOrCalculateKeyHashes(SESSION_KEYS) { sessionInitiationKeys }
+
+        /**
+         * The now historic member session initiation keys.
+         */
+        @JvmStatic
+        val MemberInfo.historicSessionInitiationKeys: Collection<PublicKey>
+            get() = memberProvidedContext.parseList(HISTORIC_SESSION_KEYS)
```
dimosr, line 219: What does `parseList` return if the key is not specified at all? Is it an empty list?

Reason I am asking is that I don't see this field being populated yet, so trying to understand whether it will have a value of an empty list, when not present, or we need to populate it in a specific way.

charlieR3: Yes, it's an empty list. A list is created and then if no matches are found for the input prefix then that list is returned, otherwise it parses matches and populates that list.

dimosr: Thanks, should be fine then.

dimosr (approved):

PR #4168, opened 2023-06-27
CORE-14355 Interop / Rename Facade service method

szymonsztuka:
FacadeService is returning not the Facade but the proxy binding it to TokensFacade. Rename the method to "getProxy".


jmacmahonr3 (approved):
LGTM

PR #4174, opened 2023-06-28
CORE-15087 update platform version

vlajos:
Update Corda Platform version to 50000 according to spec.
Also update related tests, test CorDapps platform version too.

### DIFF
--- a/components/configuration/configuration-rest-resource-service-impl/src/test/kotlin/net/corda/configuration/rest/impl/ConfigRestResourceImplTests.kt
+++ b/components/configuration/configuration-rest-resource-service-impl/src/test/kotlin/net/corda/configuration/rest/impl/ConfigRestResourceImplTests.kt
@@ -58,7 +58,7 @@ class ConfigRestResourceImplTests {
         }
     }
 
-    private val req = UpdateConfigParameters("section", 999, TestJsonObject("a=b"), ConfigSchemaVersion(888, 0))
+    private val req = UpdateConfigParameters("section", 50000, TestJsonObject("a=b"), ConfigSchemaVersion(888, 0))
     private val successFuture = CompletableFuture.supplyAsync {
         ConfigurationManagementResponse(
             true, null, req.section, req.config.escapedJson, ConfigurationSchemaVersion(
--- a/libs/configuration/configuration-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/ConfigEntityManagerIntegrationTest.kt
+++ b/libs/configuration/configuration-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/ConfigEntityManagerIntegrationTest.kt
@@ -68,7 +68,7 @@ class ConfigEntityManagerIntegrationTest {
     @Test
     fun `can persist and read back config entities`() {
         val config = ConfigEntity(
-            "${random.nextInt()}", "a=b", 999, 0,
+            "${random.nextInt()}", "a=b", 50000, 0,
             // truncating to millis as on windows builds the micros are lost after fetching the data from Postgres
             Instant.now().truncatedTo(ChronoUnit.MILLIS), "actor"
         )
@@ -86,7 +86,7 @@ class ConfigEntityManagerIntegrationTest {
     @Test
     fun `can persist and read back config audit entities`() {
         val config = ConfigEntity(
-            "${random.nextInt()}", "a=b", 999, 0,
+            "${random.nextInt()}", "a=b", 50000, 0,
             // truncating to millis as on windows builds the micros are lost after fetching the data from Postgres
             Instant.now().truncatedTo(ChronoUnit.MILLIS), "joel"
         )
--- a/libs/packaging/packaging/test/contract-cpk/build.gradle
+++ b/libs/packaging/packaging/test/contract-cpk/build.gradle
@@ -17,8 +17,8 @@ tasks.named('jar', Jar) {
 }
 
 cordapp {
-    targetPlatformVersion 1000
-    minimumPlatformVersion 999 as Integer
+    targetPlatformVersion 50000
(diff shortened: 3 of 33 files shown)


testing/cpbs/crypto-custom-digest-one-consumer/build.gradle
```diff
@@ -8,7 +8,7 @@ description 'Corda Crypto Custom Digest Consumer One'
 group 'com.example.crypto.consumer'
 
 cordapp {
-    targetPlatformVersion = 999
+    targetPlatformVersion = 50000
```
simon-johnson-r3, line 11: I see many test cordapps use `platformVersion.toInteger()` to avoid the maintenance problem with using a literal value, would that be more appropriate here (and all the other cases in the testing/cpbs directory, and other test applications), or is this really bound to 5.0 specifically?

libs/configuration/configuration-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/ConfigEntityManagerIntegrationTest.kt
```diff
@@ -68,7 +68,7 @@ class ConfigEntityManagerIntegrationTest {
     @Test
     fun `can persist and read back config entities`() {
         val config = ConfigEntity(
-            "${random.nextInt()}", "a=b", 999, 0,
+            "${random.nextInt()}", "a=b", 50000, 0,
```
simon-johnson-r3, line 71: Is there a more generic way to achieve this such that it doesn't have to be hard coded every version? I appreciate this is out of scope of setting the value correctly however, and might prolong that task unnecessarily, but maybe we need a JIRA ticket to clean it up in later version?

libs/packaging/packaging/test/contract-cpk/build.gradle
```diff
@@ -17,8 +17,8 @@ tasks.named('jar', Jar) {
 }
 
 cordapp {
-    targetPlatformVersion 1000
-    minimumPlatformVersion 999 as Integer
+    targetPlatformVersion 50000
+    minimumPlatformVersion 50000 as Integer
```
simon-johnson-r3, line 21: See comment about use of `platformVersion.toInteger()` which might apply to all `build.gradle` files, but also here I believe `50000` already is an Integer.

libs/platform-info/src/test/kotlin/net/corda/libs/platform/impl/PlatformInfoProviderImplTest.kt
```diff
@@ -12,7 +12,7 @@ import org.osgi.framework.Version
 class PlatformInfoProviderImplTest {
 
     companion object {
-        const val PLATFORM_VERSION = "999"
```
simon-johnson-r3, line 15: if this is a test info provider, it strikes me it could be one place you don't want to hard code the real value?

components/configuration/configuration-rest-resource-service-impl/src/test/kotlin/net/corda/configuration/rest/impl/ConfigRestResourceImplTests.kt
```diff
@@ -58,7 +58,7 @@ class ConfigRestResourceImplTests {
         }
     }
 
-    private val req = UpdateConfigParameters("section", 999, TestJsonObject("a=b"), ConfigSchemaVersion(888, 0))
+    private val req = UpdateConfigParameters("section", 50000, TestJsonObject("a=b"), ConfigSchemaVersion(888, 0))
```
simon-johnson-r3, line 61: Similar to the other comment, is there a way to make this not sensitive to the real platform version (or JIRA ticket so we can avoid changes in the future)?

testing/cpbs/crypto-custom-digest-one-consumer/build.gradle
vlajos: I assumed that there might have been a point why the original version used the hard coding. Although this was just an assumption.
But happy to reference it via `platformVersion.toInteger()`

libs/configuration/configuration-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/ConfigEntityManagerIntegrationTest.kt
vlajos: I think we just need something roughly realistic in this case. So I would not upgrade it for later versions.

libs/packaging/packaging/test/contract-cpk/build.gradle
vlajos: Correct. Probably for 999 as well.
Is it OK to change all of these gradle versions to `platformVersion.toInteger()?

libs/platform-info/src/test/kotlin/net/corda/libs/platform/impl/PlatformInfoProviderImplTest.kt
vlajos: I thought we set it once to 50000 and then keep it the same 50000 for later versions.
But I do not mind changing it.

simon-johnson-r3:
I think a risk averse strategy here would be to submit as it and raise a JIRA on the other stuff. There is no customer impact to deferring the things I mention in my comments.

The only concern I would have here is we accidentally backed ourselves into a corner which made 5.1 difficult - I would like a sanity check from other reviewers on that too - but otherwise I think we can live with this.

driessamyn (approved):
LGTM.

@simon-johnson-r3 , @fowlerrr , as the authors of the design doc, would you also be able to approve if this lines up with the intention?

fowlerrr (approved):

simon-johnson-r3 (approved):

PR #4178, opened 2023-06-28
CORE-14775 Rolling back before last changes to combined worker

ben-millar:


settings.gradle
```diff
@@ -515,4 +515,4 @@ gradleEnterprise {
             }
         }
     }
-}
+} // Dummy change
```
chrisr3, line 518: But won't Jenkins still just merge this one-line change into `release/os/5.0` HEAD before building?

PR #4179, opened 2023-06-28
CORE-9901 - Perform signature verification on the member side with the right key

r3-arish:
Currently, we verify signatures with the key sent by the other side (as part of CryptoSignatureWithKey). Instead of that:

when verifying the member’s signature for member-provided context, we should verify with the key inside the member’s MemberInfo.

when verifying the MGM signature for mgm-provided context, we should be verifying with the key in the MGM’s MemberInfo.

when verifying the group parameters signature, we should be verifying with the key in the MGM’s MemberInfo. 

This is in MemberSynchronisationServiceImpl.


components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
```diff
@@ -258,14 +259,16 @@ class MemberSynchronisationServiceImpl internal constructor(
...
-        ) = with(membershipPackage.groupParameters) {
-            verifier.verify(
-                mgm.sessionInitiationKeys,
-                mgmSignature,
-                mgmSignatureSpec,
-                groupParameters.array()
-            )
-            groupParametersFactory.create(this)
+        ): InternalGroupParameters {
+            val mgmPublicKey = mgm.ledgerKeys.firstOrNull()
```
nikinagy: I think this is wrong. The mgm's key that was used for signing should be retrieved from the `membershipPackage.groupParameters.mgmSignature.publicKey`. See schemas in corda-api :
https://github.com/corda/corda-api/blob/release/os/5.1/data/avro-schema/src/main/resources/avro/net/corda/data/membership/p2p/MembershipPackage.avsc 
https://github.com/corda/corda-api/blob/release/os/5.1/data/avro-schema/src/main/resources/avro/net/corda/data/membership/SignedGroupParameters.avsc
https://github.com/corda/corda-api/blob/release/os/5.1/data/avro-schema/src/main/resources/avro/net/corda/data/crypto/wire/CryptoSignatureWithKey.avsc

nikinagy: Also we don't have ledger keys from MGM, only session and ECDH:
https://github.com/corda/corda-runtime-os/wiki/MGM-Onboarding

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
```diff
@@ -281,25 +284,47 @@ class MemberSynchronisationServiceImpl internal constructor(
...
             val viewOwningMember = updates.synchronisationMetaData.member.toCorda()
             val mgm = updates.synchronisationMetaData.mgm.toCorda()
             logger.debug { "Member $viewOwningMember received membership updates from $mgm." }
 
             return try {
                 cancelCurrentRequestAndScheduleNewOne(viewOwningMember, mgm)
+
                 val updateMembersInfo = updates.membershipPackage.memberships.memberships.map { update ->
+
+                    val memberKey = getMemberKeyFromMemberInfo(update.memberContext)
```
dimosr: I don't think this is correct either. We should be retrieving the session initiation key from the member info. Given, we have multiple ones, we can filter out the one that matches the one in the signature packet. But we shouldn't just verify blindly with that without verifying it's one of the session initiation keys advertised in the member info. 

(at this point it's not deserialised yet, so the logic might need some shuffling around to deserialise first).

r3-arish:
The changes have been around the processMembershipUpdates function in the Member Synchronisation Implementation class. The logic has been shuffled, as the session initialisation keys that are fetched from the member member info after the deserialisation has been performed. The old verify function has been removed from the Verifier class. Moreover, I have made changes in the corresponding tests for the implementation class. 

I have attached the test results below.

Screenshot 2023-07-03 at 23 01 10
Screenshot 2023-07-03 at 23 19 01
Screenshot 2023-07-04 at 00 53 44

components/membership/membership-p2p/src/test/kotlin/net/corda/membership/p2p/helpers/VerifierTest.kt
```diff
@@ -46,8 +46,11 @@ class VerifierTest {
     @Test
     fun `verify without key call the service with the correct arguments`() {
```
dimosr: I assume the test name is not fully accurate now, since the method without a key parameter has been removed. We should probably rename `verify without key..` to something like `verify with multiple keys ..` in all these tests?

components/membership/membership-p2p/src/test/kotlin/net/corda/membership/p2p/helpers/VerifierTest.kt
```diff
@@ -46,8 +46,11 @@ class VerifierTest {
...
         val data = byteArrayOf(44, 1)
+        val publicKeys = (1..3).map {
+            mock<PublicKey>()
+        }
```
YashNabar: Can we do this once on the class and then reuse that in the test cases?

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
```diff
@@ -284,30 +282,48 @@ class MemberSynchronisationServiceImpl internal constructor(
         override fun processMembershipUpdates(updates: ProcessMembershipUpdates): List<Record<*, *>> {
             val viewOwningMember = updates.synchronisationMetaData.member.toCorda()
             val mgm = updates.synchronisationMetaData.mgm.toCorda()
+            val groupReader = membershipGroupReaderProvider.getGroupReader(viewOwningMember)
+            val mgmInfo = groupReader.lookup().firstOrNull { it.isMgm } ?: throw CordaRuntimeException(
```
YashNabar, line 286: Can we reuse this `mgmInfo` below where we parse the group parameters?

components/membership/membership-p2p/src/test/kotlin/net/corda/membership/p2p/helpers/VerifierTest.kt
r3-arish: This is a good point! I will do that

dimosr (commented):
LGTM - just a minor thing

components/membership/membership-p2p/src/test/kotlin/net/corda/membership/p2p/helpers/VerifierTest.kt
r3-arish: Good catch, I have addressed it in the latest commit!

dimosr (approved):

YashNabar (approved):
LGTM

PR #4181, opened 2023-06-28
CORE-14775 Increasing smoke test timeout [do not merge]

ben-millar:
It was found that, after removing the scheduled wakeup events, external events which have hit transient errors are taking some time (~15 minutes) to retry. I've increased the smoke test timeout from 6 minutes to 30 minutes to try to determine if the flaky flow tests are timing out due to external event transient failures.

### DIFF
--- a/.ci/e2eTests/JenkinsfileCombinedWorker
+++ b/.ci/e2eTests/JenkinsfileCombinedWorker
@@ -31,6 +31,10 @@ pipeline {
         }
     }
 
+triggers {
+        cron('H H/2 * * *')
+    }
+
     environment {
         ARTIFACTORY_CREDENTIALS = credentials('artifactory-credentials')
         BUILD_CACHE_CREDENTIALS = credentials('gradle-ent-cache-credentials')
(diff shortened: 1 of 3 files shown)


.ci/e2eTests/JenkinsfileCombinedWorker
```diff
@@ -109,7 +113,7 @@ pipeline {
         }
         stage('smoketests') {
                 options {
-                    timeout(time: 30, unit: 'MINUTES')
+                    timeout(time: 120, unit: 'MINUTES')
```
chrisr3, line 112: _Yikes!_ :scream:

Imagine waiting for two hours just for your build to fail at the end...

Can't you try poking something to wake it up more frequently instead?

ben-millar: I'll debug the external events properly tomorrow, if needs be, but for now I've set up a cron job to run these tests every 2hrs overnight to see if the tests pass given some more time 😁 I don't think they'll take 2 hours, but I'm certain they won't take more than that.

chrisr3: Yes, but by increasing the test timeout you will probably end up running fewer tests overnight too.

ben-millar: If that turns out to be an issue, I'll re-do this during the winter solstice. ❄️ 

In all seriousness though, the only tests currently timing out are the flaky smoke tests for which the individual timeouts have been increased from `6` to `30` minutes. Just a quick PoC to see if those tests now no longer fail if they're given time for external events to retry which can work away in the background while I do other things.

PR #4184, opened 2023-06-29
CORE-15068 Reduce the external event retry period

lankydan:
The external event retry period relies on other events to come in after the retry period has elapsed.

Due to the increase duration between flow wakeup events, it caused flows that have to retry external events to wait a significantly long time to retry.

When an external event must be retried, it now will schedule a wakeup that matches the `sendTimestamp` set when sending the previous attempt for the external event. Therefore, decoupling the retry period from the general wakeup period of flows.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/ExternalEventResponseHandler.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/ExternalEventResponseHandler.kt
@@ -17,7 +17,7 @@ import java.time.Duration
 import java.time.Instant
 
 @Component(service = [FlowEventHandler::class])
-class ExternalEventResponseHandler @Activate constructor(
+class ExternalEventResponseHandler(
     private val clock: Clock,
     @Reference(service = ExternalEventManager::class)
     private val externalEventManager: ExternalEventManager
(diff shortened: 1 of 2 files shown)


lankydan:
This should go into `release/os/5.0` because it can cause flows to hang for 30 mins due to the wakeup period being increased to 30 mins (or it might be 15 mins, either way its not ideal).

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/ExternalEventResponseHandler.kt
```diff
@@ -54,11 +68,18 @@ class ExternalEventResponseHandler @Activate constructor(
...
+        val updatedExternalEventState = externalEventManager.processResponse(
             externalEventState,
             externalEventResponse
         )
 
+        checkpoint.externalEventState = updatedExternalEventState
+
+        if (updatedExternalEventState.status.type == ExternalEventStateType.RETRY) {
+            val sleepDuration = Duration.between(clock.instant(), updatedExternalEventState.sendTimestamp).toMillis().toInt()
+            checkpoint.setFlowSleepDuration(if (sleepDuration > 0) sleepDuration else 0)
```
ben-millar: Nit: you can avoid the condition logic here if you chain `.coerceAtLeast(0)` into the sleep duration def

ben-millar (dismissed):
LGTM pending ✅, well spotted!

ben-millar (approved):

driessamyn (approved):
LGTM

PR #4185, opened 2023-06-29
CORE-14538: Handle "%" symbol in start flow requests

Tom-Fitzpatrick:
This PR removes the `formParams` property in `ClientRequestContext`, which was throwing an error handling certain requests. Instead, we now use a function which calls the `formParamMap` method from javalin, only when we need to access form parameters.


vkolomeyko (approved):
LGTM

PR #4190, opened 2023-06-29
CORE-14592: Fix incorrect error code 500 response for negative scenario when using external messaging service

BenYip123:
- Add a BadRequestException for type ConfigurationValidationException, so that error code 400 is received instead of 500 for negative scenarios when using the external messaging service, since it is not a server error.
- E2E test for this scenario: [CORE 14592: Change invalid channel config test to check for code 400 #137](https://github.com/corda/corda-e2e-tests/pull/137)

### DIFF
--- a/components/virtual-node/cpi-upload-rest-service/build.gradle
+++ b/components/virtual-node/cpi-upload-rest-service/build.gradle
@@ -18,6 +18,7 @@ dependencies {
     implementation project(':libs:lifecycle:lifecycle')
     implementation project(':libs:messaging:messaging')
     implementation project(':libs:virtual-node:cpi-upload-endpoints')
+    implementation project(':libs:configuration:configuration-validation')
     api project(':libs:virtual-node:cpi-upload-manager')
     implementation project(':libs:virtual-node:virtual-node-endpoints')
     implementation project(':libs:packaging:packaging-core')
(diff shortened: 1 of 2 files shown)


vkolomeyko:
Looks fine to me.
But I would suggest adding an E2E test for at least one negative scenario to prove that we are indeed handling `ConfigurationValidationException` correctly.

filipesoliveira (approved):
LGTM. Good work. Can you leave a link pointing to the E2E test changes in this PR?

BenYip123:
> Looks fine to me. But I would suggest adding an E2E test for at least one negative scenario to prove that we are indeed handling `ConfigurationValidationException` correctly.

I believe https://github.com/corda/corda-e2e-tests/blob/37666d863883aad8b0925ba47faa248aeef58a1e/e2etest/src/e2eTest/kotlin/net/corda/e2etest/virtualnode/ExternalMessagingRouteConfigTest.kt#L214 already tests for this. I would just need to change the condition to check for the right error code

BenYip123:
Relevant E2E test changes for negative testing this scenario are located in this PR: [CORE 14592: Change invalid channel config test to check for code 400 #137](https://github.com/corda/corda-e2e-tests/pull/137)

simon-johnson-r3 (approved):

PR #4199, opened 2023-06-30
CORE-14749 Skip Message Bus for Same Topic/Key

Cap-j:
_Context_
One of the ideas proposed was to minimise Kafka interactions (“hops”) by avoiding them when they were not necessary, such as when messaging occurs between two parties on one cluster. This POC reduces the number of hops and the latency by about 10%, as can be seen in the test results linked in the relevant ticket.

_Solution_
This change re-routes messages that are sent between two members on one cluster by checking if the new topic is the same as the current event topic. If the topic is the same, the message is forced through locally, instead of being sent out to Kafka and back. By avoiding Kafka when it is not strictly necessary to get the message across, we are decreasing the performance drop that it causes.

_Issue_
This change sacrifices some checkpoints, as those are saved when messages are passed through Kafka. We should work on an approach that allows us to implement this change while retaining the safety of checkpoints in this use case.


tmp/ca/ca/.ca/keystore.jks
yarivus, line 1: R U sure the tmp/ca/ca/.ca/ folder should not be gitignored ?

Cap-j: Unintentional commit - deleted now, thank you!

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
```diff
@@ -230,11 +231,18 @@ internal class StateAndEventSubscriptionImpl<K : Any, S : Any, E : Any>(
...
+                while (eventsToProcess.isNotEmpty()) {
+                    val event = eventsToProcess.removeFirst()
                     stateAndEventConsumer.resetPollInterval()
-                    processEvent(event, outputRecords, updatedStates)
+                    processEvent(event, outputRecords, newEventsToProcess, updatedStates)
+                    eventsToProcess.addAll(newEventsToProcess.map {
+                        val ret = toCordaConsumerRecord(event, it)
+                        ret
+                    })
+                    newEventsToProcess.clear()
```
yarivus: I'd check if it's more performant to declare the `newEventsToProcess` in the loop iteration rather than clearing it every iteration

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
```diff
@@ -265,18 +273,33 @@ internal class StateAndEventSubscriptionImpl<K : Any, S : Any, E : Any>(
         return false
     }
 
+    private fun toCordaConsumerRecord(sourceRecord: CordaConsumerRecord<K, E>, newEvent : Record<K, E>) =
```
yarivus: this private func has no benefit, better calling the `CordaConsumerRecord` c'tor where needed.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
```diff
@@ -265,18 +273,33 @@ internal class StateAndEventSubscriptionImpl<K : Any, S : Any, E : Any>(
...
+            sourceRecord.timestamp,
+            sourceRecord.headers
+        )
+
+    @Suppress("UNCHECKED_CAST")
     private fun processEvent(
         event: CordaConsumerRecord<K, E>,
         outputRecords: MutableList<Record<*, *>>,
-        updatedStates: MutableMap<Int, MutableMap<K, S?>>
+        recordsToProcess: MutableList<Record<K, E>>,
```
yarivus: I'd put the `recordsToProcess` as the return type, this will make the code simpler here and in the call-site.

yarivus (commented):
some comments

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
Cap-j: Refactoring code to change the return type as per the last comment resolved this issue too.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
Cap-j: Addressed, thank you!

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
Cap-j: Refactored to change return type

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
```diff
@@ -230,11 +230,17 @@ internal class StateAndEventSubscriptionImpl<K : Any, S : Any, E : Any>(
...
         try {
             processorMeter.recordCallable {
-                for (event in events) {
+                while (eventsToProcess.isNotEmpty()) {
+                    val event = eventsToProcess.removeFirst()
                     stateAndEventConsumer.resetPollInterval()
-                    processEvent(event, outputRecords, updatedStates)
+                    val newEventsToProcess = processEvent(event, outputRecords, updatedStates)
+                    eventsToProcess.addAll(newEventsToProcess.map {
+                        val ret = CordaConsumerRecord(it.topic, event.partition, event.offset, it.key, it.value, event.timestamp, event.headers)
```
pmck91: we can remove the val ret part here, and ret from the next line. It was added to aid in debug logging

ben-millar:
This new logic probably needs tests in `StateAndEventSubscriptionImplTest.kt` and `StateAndEventSubscriptionIntegrationTest.kt`

yarivus (approved):
LGTM

pmck91 (approved):

driessamyn (changes requested):
This is a message pattern change, so should be considered medium to high risk.

Given the linked JIRA mentions a POC, do we have evidence that this is a good change and makes a measurable difference?
Can the attached JIRA be updated with relevant information?

FYI @JamesHR3

Cap-j:
@driessamyn For POC test results, see [this Confluence page](https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4529389617/Test+18+-+Allow+Subscription+Processors+to+feed+subset+of+events+back+into+itself+without+going+to+bus+first+MSK) [edited to correct]

thiagoviana:
> @driessamyn For POC test results, see [this Confluence page](https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4522147874/Test+16+-+Skip+P2P+layer+when+parties+are+on+the+same+cluster)

@driessamyn [This](https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4529389617/Test+18+-+Allow+Subscription+Processors+to+feed+subset+of+events+back+into+itself+without+going+to+bus+first+MSK) is what you're looking for.

@Cap-j the previously shared confluent page is not related to this PR.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
```diff
@@ -230,11 +230,16 @@ internal class StateAndEventSubscriptionImpl<K : Any, S : Any, E : Any>(
...
+        val eventsToProcess = ArrayDeque(events)
         try {
             processorMeter.recordCallable {
-                for (event in events) {
+                while (eventsToProcess.isNotEmpty()) {
+                    val event = eventsToProcess.removeFirst()
                     stateAndEventConsumer.resetPollInterval()
-                    processEvent(event, outputRecords, updatedStates)
+                    val newEventsToProcess = processEvent(event, outputRecords, updatedStates)
+                    eventsToProcess.addAll(newEventsToProcess.map {CordaConsumerRecord(
```
owenstanford, line 240: ~~Does this work with the commit offsets?~~
~~The commit offsets works out the max offset for the batch of input events, if the initial event was part of a batch and the synthetic event was processed by itself, would it not be possible for it to commit an earlier offset (albeit only relevant in a rebalance or failure scenario).~~
~~For Example:~~
~~batch 1:~~
~~- event 1 offset 0 - outputs wakeup event (that's fast tracked and added to the next batch)~~
~~- event 2 offset 1~~
~~- event 3 offset 2~~
~~batch 2: (nothing received from kafka)~~
~~- synthetic wakeup event offset 0 (copied from the source event)~~
Ignore, I've looked at the code again, I see you are processing the wakeup event as part of the same initial batch

PR #4201, opened 2023-06-30
CORE-15070: use service to check if cluster is local

emilybowe:
This PR adds a RecordFactory and slightly redesigns Init, Error and Event Executors. These executors now check if Outbound events are local via the record factory and selects which topic to publish to based on this for performance. The RecordFactory uses a lookup service from membership to do this.


thiagoviana:
As agreed, we're pointing this merge to a feature branch, not 5.0 or 5.1.

thiagoviana:
As agreed, we're moving this to a feature branch, not 5.0 or 5.1.

components/membership/locally-hosted-identities-service-impl/src/main/kotlin/net/corda/membership/locally/hosted/identities/impl/LocallyHostedIdentitiesServiceImpl.kt
```diff
@@ -80,6 +80,7 @@ class LocallyHostedIdentitiesServiceImpl(
     private fun handleEvent(event: LifecycleEvent) {
         when (event) {
             is StartEvent -> {
+                logger.info("StartEvent")
```
LWogan: dont forget to remove these logs

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/RecordFactoryImpl.kt
```diff
@@ -0,0 +1,165 @@
...
+    override fun forwardError(
+        sessionEvent: SessionEvent,
+        exceptionEnvelope: ExceptionEnvelope,
+        instant: Instant,
+        flowConfig: SmartConfig,
+        messageDirection: MessageDirection
+    ): Record<*, *> {
+        val outputTopic = getSessionEventOutputTopic(sessionEvent, messageDirection)
+        val errEvent = SessionEvent(
+            MessageDirection.INBOUND,
```
LWogan: You can move the construction of this error SessionEvent into the if (isLocalCluster(sessionEvent)) statement

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/FlowMapperHelper.kt
```diff
@@ -26,32 +25,27 @@ fun generateFlowId(): String {
...
-private fun getSourceAndDestinationIdentity(sessionEvent: SessionEvent): Pair<HoldingIdentity, HoldingIdentity> {
+fun getSourceAndDestinationIdentity(sessionEvent: SessionEvent): Pair<HoldingIdentity, HoldingIdentity> {
     return if (sessionEvent.sessionId.contains(INITIATED_SESSION_ID_SUFFIX)) {
         Pair(sessionEvent.initiatedIdentity, sessionEvent.initiatingIdentity)
     } else {
         Pair(sessionEvent.initiatingIdentity, sessionEvent.initiatedIdentity)
     }
 }
 
+fun getDestinationIdentity(sessionEvent: SessionEvent): HoldingIdentity {
```
LWogan, line 46: needs a kdoc

components/flow/flow-mapper/src/main/kotlin/net/corda/flow/mapper/factory/RecordFactory.kt
```diff
@@ -0,0 +1,42 @@
...
+import net.corda.messaging.api.records.Record
+import java.time.Instant
+
+interface RecordFactory {
+    /**
+     * Create [Record] after checking if the cluster is local.
+     * @return A record for p2p.out or local
+     */
+
+    fun forwardEvent(
```
LWogan, line 23: kdocs

LWogan (commented):
Looks good. did a quick pass. Ill need to take a closer look again later.

dimosr:
I haven't looked at the code in detail, but my understanding based on discussions so far is we're trying to skip delegating to the p2p layer completely when the identity is locally hosted? If so, there's quite a bit of extra business logic that is currently performed at the p2p layer that you might need to incorporate at the flow layer if you want to go down that path. Thinking about scenarios where that identity is suspended for example. Most of these scenarios are captured by our e2e tests, but it's not a gate for PRs here. So, I'd suggest triggering a manual run of the e2e tests (ideally the regular and also the [multi-cluster](https://ci02.dev.r3.com/job/Corda5/job/corda-e2e-tests-multi-cluster-tests/) one) with this change before merging it when it's ready to ensure there will be no regressions.

LWogan:
> I haven't looked at the code in detail, but my understanding based on discussions so far is we're trying to skip delegating to the p2p layer completely when the identity is locally hosted? If so, there's quite a bit of extra business logic that is currently performed at the p2p layer that you might need to incorporate at the flow layer if you want to go down that path. Thinking about scenarios where that identity is suspended for example. Most of these scenarios are captured by our e2e tests, but it's not a gate for PRs here. So, I'd suggest triggering a manual run of the e2e tests (ideally the regular and also the [multi-cluster](https://ci02.dev.r3.com/job/Corda5/job/corda-e2e-tests-multi-cluster-tests/) one) with this change before merging it when it's ready to ensure there will be no regressions.

We currently do the memberlookup in the flow engine before sending any messages. If the member doesn't exist, we will error the session going forward. (in C5.0 wed retry a few times but retry is being removed). Does this cover the p2p business logic you are referring to?

```
val counterparty: MemberX500Name = MemberX500Name.parse(sessionState.counterpartyIdentity.x500Name!!)
        val groupReader = membershipGroupReaderProvider.getGroupReader(context.checkpoint.holdingIdentity)
        val counterpartyExists: Boolean = null != groupReader.lookup(counterparty)
        ```

dimosr:
ok, if you also do a member lookup before (and not just relying on the locally hosted identity data), it should protect against it as by default only active members are returned. And it will be a bit more resilient against other edge cases as well, such as a locally hosted identity that hasn't registered yet trying to send a message.

PR #4206, opened 2023-07-03
CORE-15087 update platform version 5.1

vlajos:
CORE-15087 update platform version to 50100 for 5.1 according spec. Centralize it. Make tests less-dependent of the main number.

### DIFF
--- a/libs/platform-info/src/main/kotlin/net/corda/libs/platform/impl/PlatformInfoProviderImpl.kt
+++ b/libs/platform-info/src/main/kotlin/net/corda/libs/platform/impl/PlatformInfoProviderImpl.kt
@@ -12,9 +12,9 @@ class PlatformInfoProviderImpl @Activate constructor(
 ) : PlatformInfoProvider {
 
     internal companion object {
-        const val STUB_PLATFORM_VERSION = 50000
+        const val STUB_PLATFORM_VERSION = 50100
 
-        private const val DEFAULT_PLATFORM_VERSION = 50000
+        private const val DEFAULT_PLATFORM_VERSION = 50100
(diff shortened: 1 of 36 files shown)


libs/platform-info/src/main/kotlin/net/corda/libs/platform/impl/PlatformInfoProviderImpl.kt
```diff
@@ -12,9 +12,9 @@ class PlatformInfoProviderImpl @Activate constructor(
 ) : PlatformInfoProvider {
 
     internal companion object {
-        const val STUB_PLATFORM_VERSION = 50000
+        const val STUB_PLATFORM_VERSION = 50100
 
-        private const val DEFAULT_PLATFORM_VERSION = 50000
+        private const val DEFAULT_PLATFORM_VERSION = 50100
```
vlajos, line 17: Are not these supposed to be derived somehow from the root gradle's platform version?

simon-johnson-r3: The long term plan for these is not yet determined. I think this is fine for now, the inconvenience means we don't forget about it. Although I appreciate that's easier to say if you're not the one having to do something about it...

vlajos: Thanks for explaining.

simon-johnson-r3 (approved):
Ideally we'd have the stable flag in place here too. But for now I think this is the best we can achieve.

thiagoviana (approved):
LGTM - Flow Worker Related

ronanbrowne (approved):
LGTM

PR #4210, opened 2023-07-04
CORE-8958 distribute group parameters in backchain resolution

vlajos:
* Distribute signed group parameters via back-chain resolution when the receiver side does not have it already.
* Create v2 Backchain resolution flows to support the new protocol.
* Initial signed group parameters must also be saved on the responder side. Since that happens before the back-chain resolution. Also, this can happen without a network hop. However, to guarantee that we saved the one used in the received transaction, we need to move the check, which compares the current group parameters hash with the one in the metadata of the transaction before the back-chain resolution, and also we need to guarantee that the same one gets saved.

Reviewing is easier commit by commit since the v2 copy generated lots of noise.
There are some version number handling tweaks in this follow-up PR: https://github.com/corda/corda-runtime-os/pull/4213


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
```diff
@@ -54,7 +55,8 @@ class TransactionBackchainResolutionFlowVersionedFlowFactory(
 
     override fun create(version: Int, session: FlowSession): SubFlow<Unit> {
         return when {
-            version >= 1 -> TransactionBackchainResolutionFlowV1(initialTransactionIds, session)
+            version in 1..50099 -> TransactionBackchainResolutionFlowV1(initialTransactionIds, session)
+            version >= 50100 -> TransactionBackchainResolutionFlowV2(initialTransactionIds, session)
```
lankydan, line 58: I assume `50100` is the next correct version?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
```diff
@@ -54,7 +55,8 @@ class TransactionBackchainResolutionFlowVersionedFlowFactory(
 
     override fun create(version: Int, session: FlowSession): SubFlow<Unit> {
         return when {
-            version >= 1 -> TransactionBackchainResolutionFlowV1(initialTransactionIds, session)
+            version in 1..50099 -> TransactionBackchainResolutionFlowV1(initialTransactionIds, session)
```
lankydan: We could leave the original `version >= 1` (or set it to `version >= 50000`) if we invert the order of the when block so that the V2 version is first.

I think that makes more sense generally as we'd expect the newest version to be the code path that is hit the most.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v2/TransactionBackchainReceiverFlowV2.kt
```diff
@@ -0,0 +1,168 @@
...
+    lateinit var utxoLedgerPersistenceService: UtxoLedgerPersistenceService
+
+    @CordaInject
+    lateinit var utxoLedgerGroupParametersPersistenceService: UtxoLedgerGroupParametersPersistenceService
+
+    @CordaInject
+    lateinit var utxoLedgerMetricRecorder: UtxoLedgerMetricRecorder
+
+    @Suspendable
+    override fun call(): TopologicalSort {
```
lankydan, line 49: We need to make sure @nkovacsx is aware of these changes since he's changing the backchain code but in the older version atm. He'll need to change both if yours is merged first, or you have to bring over his changes etc.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v2/TransactionBackchainReceiverFlowV2.kt
```diff
@@ -0,0 +1,168 @@
...
+                                "for transaction ($retrievedTransactionId)"
+                    }
+                    val retrievedSignedGroupParameters = session.sendAndReceive(
+                        SignedGroupParameters::class.java,
+                        TransactionBackchainRequestV2.GetSignedGroupParameters(groupParametersHash)
+                    )
+                    utxoLedgerGroupParametersPersistenceService.persistIfDoesNotExist(retrievedSignedGroupParameters)
+                }
+
+                val (status, _) = utxoLedgerPersistenceService.persistIfDoesNotExist(retrievedTransaction, UNVERIFIED)
```
lankydan: We need to do a lot of work to optimise this code. Being performance is being looked at a lot atm we need to ensure that we're not making things worse.

We should at a minimum combine the persisting of the group parameters and the received transaction.

We might want to try and take things further tbh.

e.g. can we proactively send the group parameters along with the requested transaction. On the side sending the transactions we can keep a check of which parameters we've already sent to the peer, so that if a transaction has the same parameters we can choose to not send it the second times.

Maybe we should discuss this because it might be reasonably involved to achieve some of these optimisations, but I think they're important to make.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -83,6 +96,24 @@ class UtxoReceiveFinalityFlowV1(
...
     }
 
+    @Suspendable
+    private fun verifyLatestGroupParametersAreUsed(initialTransaction: UtxoSignedTransactionInternal): SignedGroupParameters {
+        val currentGroupParameters = currentGroupParametersService.get()
+        val txGroupParametersHash =
+            (initialTransaction.metadata as TransactionMetadataInternal).getMembershipGroupParametersHash()
+        if (
+            txGroupParametersHash != currentGroupParameters.hash.toString()
+        ) {
```
lankydan: Why do we have the `()` on different lines?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
vlajos: Based on https://github.com/corda/platform-eng-design/blob/ca17d9154d89d4484a747544b366edcb86b2f19d/core/corda-5/corda-5.0/the-host/platform-version-number.md that means 5.1.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
vlajos: I have swapped the conditions.
I think it is a bit easier to read/it is more explicit to have the range bounded from both ends.
But I can be convinced that the other way is better.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
vlajos: The two variables before the `if` were inlined initially, and the `if`'s condition was long and wrapped. IntelliJ's formatting did not remove the extra line when I introduced the variables.
I have fixed this manually.

lankydan (approved):

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
```diff
@@ -3,6 +3,7 @@ package net.corda.ledger.utxo.flow.impl.flows.backchain
 import net.corda.flow.application.services.VersioningService
 import net.corda.flow.application.versioning.VersionedReceiveFlowFactory
 import net.corda.ledger.utxo.flow.impl.flows.backchain.v1.TransactionBackchainResolutionFlowV1
+import net.corda.ledger.utxo.flow.impl.flows.backchain.v2.TransactionBackchainResolutionFlowV2
```
relyafi, line 6: This PR hasn't introduced this pattern, but does it make sense to have the version as part of the package path? We already identify this as a new version in the class name.

The reason I ask is that presumably there is no guarantee that all classes in a package are going to evolve version at the same time? As different classes upgrade, there is not going to be a cohesive "this is the version 2 of backchain resolution" package, there may be some flows relating to backchain resolution that are at v2, others at v3 etc.

relyafi (commented):
I've not done a full review yet, as I thought it worth clarifying if we need to introduce full-fledged V2 flows at this point?

The new V2 flows share a lot of logic with the V1 flows. The strongest example of this is `TransactionBackchainResolutionFlowV2` which only differs in terms of the name of the subflow it is invoking.

Have we considered if there are any cases where it would be safe to simply update an existing flow to switch on the platform version? Or is that inherently difficult due to the fact flows may be checkpointed and later resume on a new platform version?

At any rate, I'd strongly recommend seeing if we can reduce the duplication that we have here, even if it's extracting logic out into common functions. Multiple copies of code like this would be a maintenance nightmare if we have to fix bugs etc. in future - each new version of the flow will potentially need changing and testing.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
vlajos: > This PR hasn't introduced this pattern, but does it make sense to have the version as part of the package path? We already identify this as a new version in the class name.
> 

I like that the files related to a specific version are clearly encapsulated in a package.
It is easy to see which files are supposed to work together.

> The reason I ask is that presumably there is no guarantee that all classes in a package are going to evolve version at the same time? As different classes upgrade, there is not going to be a cohesive "this is the version 2 of backchain resolution" package, 

There is no guarantee that all classes change.
On the other hand, we need to upgrade only when the flow protocol changes, which means at least the sender and the receiver change.  So generally, at least two classes will change.
In this particular case, three of the four files changed.

> there may be some flows relating to backchain resolution that are at v2, others at v3 etc.

The flow version negotiating logic guarantees that all participants will use the latest version, which is available for all participants.
No versioned flow is supposed to be started without this negotiation wrapper flow.
But this is already integrated into all relevant places.

vlajos:
> I've not done a full review yet, as I thought it worth clarifying if we need to introduce full-fledged V2 flows at this point?

A new request is being introduced to the Backchain resolution flow.
If the receiver starts to use it without a compatible sender, then the sender will not recognize it, and it will throw, and the flow will abort.

> The new V2 flows share a lot of logic with the V1 flows. The strongest example of this is `TransactionBackchainResolutionFlowV2` which only differs in terms of the name of the subflow it is invoking.
> 
> Have we considered if there are any cases where it would be safe to simply update an existing flow to switch on the platform version? Or is that inherently difficult due to the fact flows may be checkpointed and later resume on a new platform version?

The new request makes the flow upgrade inevitable.
AFAIK the CPK distribution will be something similar, so ideally, we should add these new changes to the same new version.

> 
> At any rate, I'd strongly recommend seeing if we can reduce the duplication that we have here, even if it's extracting logic out into common functions. Multiple copies of code like this would be a maintenance nightmare if we have to fix bugs etc. in future - each new version of the flow will potentially need changing and testing.

Yeah. I agree. It would be hard to maintain.
Ideally, we should not have too many protocol-breaking changes.
For this particular version, I think we will have at least one more. (CPKs.)

Sharing code between the versions can be tricky.
It will make cross-version situations harder to understand. And they are usually already complicated.
Anyway If that is preferred by the team, I am happy to implement some changes.

relyafi:
Thanks @vlajos for the comments. I think assuming that protocol changes will be rare is optimistic, given we're already making one before we've even got GA out of the door. That said, it may be the case that we will be making more changes early on as we refine our initial solution, and things will eventually settle down.

I think your point regarding the flows being grouped together under the same version is true if we can guarantee that we have a separate package path for each specific functional use-case. For example, the current package path implies that all "backchain" related functionality will up-version together. Whilst that is true for now, because we only have "backchain resolution", and a single way of doing it, that may not hold in future.

blsemo:
>Have we considered if there are any cases where it would be safe to simply update an existing flow to switch on the platform version? Or is that inherently difficult due to the fact flows may be checkpointed and later resume on a new platform version?

That won't work as we need to consider the platform version of the node running the flow and of the peers - this is exactly what the flow versioning mechanism does.

blsemo:
I don't think we can have a general pattern how versioned flows should always be done because it depends on the very specific circumstances.
In this particulcar case, it might be enough to have one set of classes that has the code for v1 and v2, and has a constructor parameter telling it whether to invoked the v2 behaviour or not. The version switching flow then could just create the subflow with either flag depending on the agreed version.

vlajos:
Alternative PR: https://github.com/corda/corda-runtime-os/pull/4265 with reusing the same class for both versions.

vlajos:
Closing in favor of https://github.com/corda/corda-runtime-os/pull/4265.

PR #4213, opened 2023-07-04
CORE-8958 centralize platform versions

vlajos:
Introduce `net.corda.libs.platform.PlatformVersion` and use it.

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
@@ -3,6 +3,7 @@ package net.corda.ledger.utxo.flow.impl.flows.backchain
 import net.corda.flow.application.services.VersioningService
 import net.corda.flow.application.versioning.VersionedReceiveFlowFactory
 import net.corda.ledger.utxo.flow.impl.flows.backchain.v1.TransactionBackchainResolutionFlowV1
+import net.corda.libs.platform.PlatformVersion.CORDA_5_1
 import net.corda.sandbox.CordaSystemFlow
 import net.corda.v5.application.flows.CordaInject
 import net.corda.v5.application.flows.SubFlow
@@ -54,12 +55,12 @@ class TransactionBackchainResolutionFlowVersionedFlowFactory(
 
     override fun create(version: Int, session: FlowSession): SubFlow<Unit> {
         return when {
-            version >= 50100 -> TransactionBackchainResolutionFlowV1(
+            version >= CORDA_5_1.platformVersion -> TransactionBackchainResolutionFlowV1(
                 initialTransactionIds,
                 session,
                 TransactionBackChainResolutionVersion.V2
             )
-            version in 1..50099 -> TransactionBackchainResolutionFlowV1(
+            version in 1 until CORDA_5_1.platformVersion -> TransactionBackchainResolutionFlowV1(
                 initialTransactionIds,
                 session,
                 TransactionBackChainResolutionVersion.V1
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
@@ -55,12 +55,12 @@ class TransactionBackchainResolutionFlowVersionedFlowFactory(
 
     override fun create(version: Int, session: FlowSession): SubFlow<Unit> {
         return when {
-            version >= CORDA_5_1.platformVersion -> TransactionBackchainResolutionFlowV1(
+            version >= CORDA_5_1.value -> TransactionBackchainResolutionFlowV1(
                 initialTransactionIds,
                 session,
                 TransactionBackChainResolutionVersion.V2
             )
-            version in 1 until CORDA_5_1.platformVersion -> TransactionBackchainResolutionFlowV1(
+            version in 1 until CORDA_5_1.value -> TransactionBackchainResolutionFlowV1(
                 initialTransactionIds,
                 session,
                 TransactionBackChainResolutionVersion.V1
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainSenderFlow.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainSenderFlow.kt
@@ -3,6 +3,7 @@ package net.corda.ledger.utxo.flow.impl.flows.backchain
 import net.corda.flow.application.services.VersioningService
 import net.corda.flow.application.versioning.VersionedSendFlowFactory
 import net.corda.ledger.utxo.flow.impl.flows.backchain.v1.TransactionBackchainSenderFlowV1
+import net.corda.libs.platform.PlatformVersion.CORDA_5_1
 import net.corda.sandbox.CordaSystemFlow
 import net.corda.v5.application.flows.CordaInject
 import net.corda.v5.application.flows.SubFlow
@@ -53,12 +54,12 @@ class TransactionBackchainSenderFlowVersionedFlowFactory(
 
     override fun create(version: Int, sessions: List<FlowSession>): SubFlow<Unit> {
         return when {
-            version >= 50100 -> TransactionBackchainSenderFlowV1(
+            version >= CORDA_5_1.platformVersion -> TransactionBackchainSenderFlowV1(
                 headTransactionIds,
                 sessions.single(),
                 TransactionBackChainResolutionVersion.V2
             )
-            version in 1..50099 -> TransactionBackchainSenderFlowV1(
+            version in 1 until CORDA_5_1.platformVersion -> TransactionBackchainSenderFlowV1(
                 headTransactionIds,
                 sessions.single(),
                 TransactionBackChainResolutionVersion.V1
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainSenderFlow.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainSenderFlow.kt
@@ -54,12 +54,12 @@ class TransactionBackchainSenderFlowVersionedFlowFactory(
 
     override fun create(version: Int, sessions: List<FlowSession>): SubFlow<Unit> {
         return when {
-            version >= CORDA_5_1.platformVersion -> TransactionBackchainSenderFlowV1(
+            version >= CORDA_5_1.value -> TransactionBackchainSenderFlowV1(
                 headTransactionIds,
                 sessions.single(),
                 TransactionBackChainResolutionVersion.V2
             )
-            version in 1 until CORDA_5_1.platformVersion -> TransactionBackchainSenderFlowV1(
+            version in 1 until CORDA_5_1.value -> TransactionBackchainSenderFlowV1(
                 headTransactionIds,
                 sessions.single(),
                 TransactionBackChainResolutionVersion.V1
--- a/components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlowVersionedFlowFactoryTest.kt
+++ b/components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlowVersionedFlowFactoryTest.kt
@@ -23,7 +23,7 @@ class TransactionBackchainResolutionFlowVersionedFlowFactoryTest {
     }
 
     @Test
-    fun `with last 5_0 platform version creates TransactionBackchainResolutionFlowV1`() {
+    fun `with last potential 5_0 platform version creates TransactionBackchainResolutionFlowV1`() {
         val flow = factory.create(CORDA_5_1.value - 1, mock())
         assertThat(flow).isExactlyInstanceOf(TransactionBackchainResolutionFlowV1::class.java)
         assertSame(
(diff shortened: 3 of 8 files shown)


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
```diff
@@ -4,6 +4,7 @@ import net.corda.flow.application.services.VersioningService
 import net.corda.flow.application.versioning.VersionedReceiveFlowFactory
 import net.corda.ledger.utxo.flow.impl.flows.backchain.v1.TransactionBackchainResolutionFlowV1
 import net.corda.ledger.utxo.flow.impl.flows.backchain.v2.TransactionBackchainResolutionFlowV2
+import net.corda.libs.platform.PlatformVersion.CORDA_5_1
```
vkolomeyko, line 6:
```suggestion
import net.corda.libs.platform.PlatformVersion.CORDA_5_1 as CURRENT_VERSION
```
Then it should be possible to use `CURRENT_VERSION` further in the file. This will help to change fewer lines in the next release when we will be introducing `CORDA_5_2`.

vlajos: @vkolomeyko I do not fully get this one. In this case, we want to introduce this new feature in a new flow from precisely 5.1.
Later versions (5.2) may or may not want to upgrade the flow, but the cutoff between v1 and v2 will be 5.1.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
```diff
@@ -55,8 +56,12 @@ class TransactionBackchainResolutionFlowVersionedFlowFactory(
 
     override fun create(version: Int, session: FlowSession): SubFlow<Unit> {
         return when {
-            version in 1..50099 -> TransactionBackchainResolutionFlowV1(initialTransactionIds, session)
-            version >= 50100 -> TransactionBackchainResolutionFlowV2(initialTransactionIds, session)
+            version in 1 until CORDA_5_1.platformVersion -> TransactionBackchainResolutionFlowV1(
```
blsemo: Not sure that is the most efficient way of writing this - doesn't this create a huge array of numbers and then check if your version is in that array? Or is there a kotlin optimisation under the hood? Wouldn't this be more efficient:
```suggestion
            version < CORDA_5_1.platformVersion -> TransactionBackchainResolutionFlowV1(
```

libs/platform-info/src/main/kotlin/net/corda/libs/platform/PlatformVersion.kt
```diff
@@ -0,0 +1,6 @@
+package net.corda.libs.platform
+
+enum class PlatformVersion(val platformVersion: Int) {
+    CORDA_5_0(50000),
+    CORDA_5_1(50100),
```
blsemo, line 5: Why is 5.1 platform version 50100? What would be platfome version 51000 in this schema?

vlajos: I have followed the related spec. (https://github.com/corda/platform-eng-design/blob/ca17d9154d89d4484a747544b366edcb86b2f19d/core/corda-5/corda-5.0/the-host/platform-version-number.md)
51000 would be 5.10

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
vlajos: `until` uses `IntRange`, which is smart enough to store only the ends:
image

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
vkolomeyko: Sorry, I misread what `TransactionBackchainResolutionFlow` is meant to do.
Specifying concrete versions makes sense to me here.

libs/platform-info/src/main/kotlin/net/corda/libs/platform/PlatformVersion.kt
```diff
@@ -0,0 +1,6 @@
+package net.corda.libs.platform
+
+enum class PlatformVersion(val platformVersion: Int) {
```
lankydan: Personally I'd prefer `value` instead of `platformVersion` but not something I feel strong about.

lankydan (dismissed):

libs/platform-info/src/main/kotlin/net/corda/libs/platform/PlatformVersion.kt
vlajos: I have updated it.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
```diff
@@ -54,17 +55,17 @@ class TransactionBackchainResolutionFlowVersionedFlowFactory(
...
     override fun create(version: Int, session: FlowSession): SubFlow<Unit> {
         return when {
-            version >= 50100 -> TransactionBackchainResolutionFlowV1(
+            version >= CORDA_5_1.value -> TransactionBackchainResolutionFlowV1(
                 initialTransactionIds,
                 session,
                 TransactionBackChainResolutionVersion.V2
             )
-            version in 1..50099 -> TransactionBackchainResolutionFlowV1(
+            version in 1 until CORDA_5_1.value -> TransactionBackchainResolutionFlowV1(
```
relyafi, line 63:
```suggestion
            version in CORDA_5_0.value until CORDA_5_1.value -> TransactionBackchainResolutionFlowV1(
```

Would this make more sense logically?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainSenderFlow.kt
```diff
@@ -53,17 +54,17 @@ class TransactionBackchainSenderFlowVersionedFlowFactory(
...
     override fun create(version: Int, sessions: List<FlowSession>): SubFlow<Unit> {
         return when {
-            version >= 50100 -> TransactionBackchainSenderFlowV1(
+            version >= CORDA_5_1.value -> TransactionBackchainSenderFlowV1(
                 headTransactionIds,
                 sessions.single(),
                 TransactionBackChainResolutionVersion.V2
             )
-            version in 1..50099 -> TransactionBackchainSenderFlowV1(
+            version in 1 until CORDA_5_1.value -> TransactionBackchainSenderFlowV1(
```
relyafi, line 62:
```suggestion
            version in CORDA_5_0.value until CORDA_5_1.value -> TransactionBackchainSenderFlowV1(
```

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlowVersionedFlowFactoryTest.kt
```diff
@@ -22,8 +23,8 @@ class TransactionBackchainResolutionFlowVersionedFlowFactoryTest {
     }
 
     @Test
-    fun `with platform version 50099 creates TransactionBackchainResolutionFlowV1`() {
-        val flow = factory.create(50099, mock())
+    fun `with last 5_0 platform version creates TransactionBackchainResolutionFlowV1`() {
+        val flow = factory.create(CORDA_5_1.value - 1, mock())
```
relyafi, line 27:
```suggestion
        val flow = factory.create(CORDA_5_0.value, mock())
```

This one is a bit more questionable than the earlier comments, because this would always test 5.0, not 5.0.1 for example. So this suggestion may not be the best approach.

That said, I think the current implementation is not testing a "real" Corda version anyway; isn't it getting the _literal_ minus one value, i.e. 50099, whereas you may want the latest valid value, i.e. we should be getting the ordinal - 1 entry. Not sure how much this matters though.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainSenderFlowVersionedFlowFactoryTest.kt
```diff
@@ -22,8 +23,8 @@ class TransactionBackchainSenderFlowVersionedFlowFactoryTest {
     }
 
     @Test
-    fun `with platform version 50099 creates TransactionBackchainSenderFlowV1`() {
-        val flow = factory.create(50099, listOf(mock()))
+    fun `with last 5_0 platform version creates TransactionBackchainSenderFlowV1`() {
+        val flow = factory.create(CORDA_5_1.value - 1, listOf(mock()))
```
relyafi, line 27: Same as comment in previous test.

libs/platform-info/src/main/kotlin/net/corda/libs/platform/impl/PlatformInfoProviderImpl.kt
```diff
@@ -12,9 +13,9 @@ class PlatformInfoProviderImpl @Activate constructor(
 ) : PlatformInfoProvider {
 
     internal companion object {
-        const val STUB_PLATFORM_VERSION = 50100
+        val STUB_PLATFORM_VERSION = CORDA_5_1.value
```
relyafi, line 16: Out of curiosity, what is the difference between the "stub" and default versions?

relyafi (commented):
Some minor comments. Overall I think this is a good change, it makes the versioning more obvious when reading the code and provides a source of truth for the platform versions.

I do wonder if it would be more appropriate to have the definition in `corda-api`, but I'm guessing it isn't used there currently.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
vlajos: If we want to cover the `1 until CORDA_5_0.value` range somehow, I think the V1 flow is friendlier than the exception.
On the other hand, it should not make any difference for real/production systems.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlowVersionedFlowFactoryTest.kt
vlajos: We know this flow upgrade should happen from 5.1, so anything before that should use V1.
Regardless of whether it exists or not (yet).
Also, the current test set checks the ends of the intervals, so it includes all possible values, not just the real ones.
This makes it a bit more future-proof and less likely to need to be adjusted when new versions get released.

vlajos: Would renaming the test to
`with last potential 5_0 ...`
make this intent cleaner?

libs/platform-info/src/main/kotlin/net/corda/libs/platform/impl/PlatformInfoProviderImpl.kt
vlajos: AFAIK the value is supposed to come from the related root gradle property. And it has yet to come from there.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
relyafi: Yeah, I couldn't see a scenario where it would make sense that we would cover 1 up to 5.0 - effectively the enum definitions are replacing the integer values. So aren't we interested in the first version (ordinal 0 in the enum)?

It's not a big deal, it just looks a bit more readable and consistent to me.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlowVersionedFlowFactoryTest.kt
relyafi: > Would renaming the test to `with last potential 5_0 ...` make this intent cleaner?

Yes, that sounds like a good compromise.

vlajos:
@relyafi 
> I do wonder if it would be more appropriate to have the definition in corda-api, but I'm guessing it isn't used there currently.

We can move it there later if it proves to be needed.
I think it is more prudent to introduce it in runtime-os first.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/TransactionBackchainResolutionFlow.kt
vlajos: The usage of enums is not enforced. The platform version will come from "somewhere else".
In the current form, the PR prepares the code for as many integers as possible and tries to handle every case in the best possible way. But I guess, this is a bit subjective.

relyafi: I guess it could be enforced by changing version from an Int to the enum type, but unfortunately that would be a breaking change. Or a non-breaking way to do it would be to throw an exception if the supplied version is not in our enumeration.

If we can't enforce the version list, then it compromises the ability for the version to be a "source of truth" - all code must assume it is dealing with any arbitrary version. The enumeration is simply a convenience helper to identify "well known" Corda versions. If we're happy with that, then I guess that's ok.

vlajos: I have aimed for this second target. (convenience helper)
It is not perfect. Just a step forward.

relyafi (approved):

PR #4215, opened 2023-07-04
CORE-14658: stop using lambda in cipher suite metrics due to Quasar problems, use our own PEM parsing logic which cannot block

dickon:
Quasar seems to have trouble with lambdas in the way we use metrics, and that caused misbehaviour. To be safe, avoid lambdas and remove the use of theoretically blocking BufferReader when parsing PEM content. This involves doing the top level parse of PEM content ourselves, which also allows detailed error reporting than the Bouncy Castle code. The detailed and security sensitive ASN.1 handling is still done by Bouncy Castle. With these changes, we no longer see the fiber corruption (e.g null pointers out of nowhere, lines of code being skipped) we were seeing, but since the issue is very sensitive to timing it's hard to be sure.

Add more test cases for PEM parsing now we own the logic.

This has been successfully tested on OpenShift using the same QA test case as in the bug report and a custom flow worker image with this change, although the failure is somewhat sporadic.

### DIFF
--- a/libs/crypto/cipher-suite-impl/src/test/kotlin/net/corda/cipher/suite/impl/CipherSchemeMetadataTests.kt
+++ b/libs/crypto/cipher-suite-impl/src/test/kotlin/net/corda/cipher/suite/impl/CipherSchemeMetadataTests.kt
@@ -66,6 +66,8 @@ import java.util.UUID
 import kotlin.test.assertEquals
 import kotlin.test.assertNotEquals
 import kotlin.test.assertNotNull
+import net.corda.crypto.cipher.suite.publicKeyId
+import org.bouncycastle.util.encoders.DecoderException
 
 class CipherSchemeMetadataTests {
     companion object {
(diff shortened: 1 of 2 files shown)


libs/crypto/cipher-suite-impl/src/test/kotlin/net/corda/cipher/suite/impl/CipherSchemeMetadataTests.kt
```diff
@@ -66,6 +66,8 @@ import java.util.UUID
 import kotlin.test.assertEquals
 import kotlin.test.assertNotEquals
 import kotlin.test.assertNotNull
+import net.corda.crypto.cipher.suite.publicKeyId
+import org.bouncycastle.util.encoders.DecoderException
```
chrisr3: These are in the wrong place.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/CipherSchemeMetadataProvider.kt
```diff
@@ -145,23 +151,33 @@ class CipherSchemeMetadataProvider : KeyEncodingService {
         algorithmMap[normaliseAlgorithmIdentifier(algorithm)]
             ?: throw IllegalArgumentException("Unrecognised algorithm: ${algorithm.algorithm.id}, with parameters=${algorithm.parameters}")
 
-    private fun <T : Any> recordPublicKeyOperation(operationName: String, op: () -> T): T {
-        return CordaMetrics.Metric.Crypto.CipherSchemeTimer.builder()
-            .withTag(CordaMetrics.Tag.OperationName, operationName)
-            .build()
-            .recordCallable {
-                op.invoke()
-            }!!
```
chrisr3, line 154: I _strongly_ suspect that the _real_ problem was that `AbstractTimer.recordCallable` is not instrumented because we _explicitly_ exclude `io.micrometer.**` classes from Quasar:
```java
package io.micrometer.core.instrument;

public abstract class AbstractTimer {

    @Override
    public <T> T recordCallable(Callable<T> f) throws Exception {
        final long s = clock.monotonicTime();
        try {
            return f.call();
        }
        finally {
            final long e = clock.monotonicTime();
            record(e - s, TimeUnit.NANOSECONDS);
        }
    }

}
```
We also exclude `net.corda.metrics` from being instrumented. However, if we changed that then we could try writing a
```kotlin
@Throws(Exception::class)
fun <T> Timer.record(callable: Callable<T>): T
```
function of our own that we can instrument.

chrisr3 (commented):
Corda 4 would have exploded years ago, if Quasar had as much trouble with lambdas as you suggest. I suspect the real explanation is much simpler.

libs/crypto/cipher-suite-impl/src/test/kotlin/net/corda/cipher/suite/impl/CipherSchemeMetadataTests.kt
dickon: fixed

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/CipherSchemeMetadataProvider.kt
dickon: Maybe, but I've got reasonable evidence that what I did solves the problem, and can now no longer reproduce the problem easily to confirm your fix, so pragmatically I still recommend taking this change.

blsemo (approved):
That solution looks fine to me.
Can we merge it and argue about instrumenting micrometer later?

driessamyn (approved):
I agree with @blsemo , this change seems reasonable, so let's get it merged if it resolves the issue. In the background we can do more investigation on the Quasar issue (or non-issue).

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/CipherSchemeMetadataProvider.kt
chrisr3: If you've stopped using the uninstrumented `AbstractTimer::recordCallable` then that _would_ reasonably resolve your immediate problem. However, I suspect we've used `AbstractTimer.recordCallable` in a lot of other places too, at least _some_ of which would also need instrumenting.

I've never heard of anyone having difficulty with Quasar and `BufferReader` before either :no_mouth:.

PR #4216, opened 2023-07-04
CORE-15172 - Adjust retry logic

nikinagy:
Adjust the retry logic on member side to be compatible with re-registration. At the moment if a member does not hear back from the MGM saying the registration request has been received for a given period of time, they will kick off the registration again. This will clash with the queueing logic at the MGM side.

### DIFF
--- a/components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/QueueRegistrationHandler.kt
+++ b/components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/QueueRegistrationHandler.kt
@@ -1,26 +1,56 @@
 package net.corda.membership.impl.registration.dynamic.handler.mgm
 
+import net.corda.avro.serialization.CordaAvroDeserializer
+import net.corda.avro.serialization.CordaAvroSerializationFactory
+import net.corda.data.KeyValuePairList
 import net.corda.data.membership.command.registration.RegistrationCommand
 import net.corda.data.membership.command.registration.mgm.CheckForPendingRegistration
 import net.corda.data.membership.command.registration.mgm.QueueRegistration
 import net.corda.data.membership.common.v2.RegistrationStatus
 import net.corda.data.membership.state.RegistrationState
+import net.corda.data.p2p.app.MembershipStatusFilter
 import net.corda.membership.impl.registration.dynamic.handler.RegistrationHandler
 import net.corda.membership.impl.registration.dynamic.handler.RegistrationHandlerResult
+import net.corda.membership.lib.MemberInfoExtension.Companion.PLATFORM_VERSION
+import net.corda.membership.lib.VersionedMessageBuilder
 import net.corda.membership.lib.registration.RegistrationRequest
+import net.corda.membership.p2p.helpers.P2pRecordsFactory
 import net.corda.membership.persistence.client.MembershipPersistenceClient
 import net.corda.messaging.api.records.Record
 import net.corda.schema.Schemas.Membership.REGISTRATION_COMMAND_TOPIC
+import net.corda.utilities.time.Clock
+import net.corda.v5.base.exceptions.CordaRuntimeException
 import net.corda.virtualnode.toCorda
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
 internal class QueueRegistrationHandler(
+    private val clock: Clock,
     private val membershipPersistenceClient: MembershipPersistenceClient,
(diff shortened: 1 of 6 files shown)


components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/QueueRegistrationHandler.kt
```diff
@@ -60,7 +89,27 @@ internal class QueueRegistrationHandler(
 
     private fun queueRequest(
         key: String, command: QueueRegistration, registrationId: String
-    ): List<Record<String, RegistrationCommand>> {
+    ): List<Record<*, *>> {
+        val outputRecords = mutableListOf<Record<*, *>>()
```
yift-r3: Instead of a mutable list, can you keep the record as a nullable variable?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/QueueRegistrationHandler.kt
```diff
@@ -60,7 +89,27 @@ internal class QueueRegistrationHandler(
...
+
+        val context = deserialize(command.memberRegistrationRequest.memberContext.data.array())
+        val platformVersion = context.items.first { it.key == PLATFORM_VERSION }.value.toInt()
+        val statusUpdateMessage = VersionedMessageBuilder.retrieveRegistrationStatusMessage(
+            platformVersion,
+            command.memberRegistrationRequest.registrationId,
+            RegistrationStatus.RECEIVED_BY_MGM.name,
+        )
+        if (statusUpdateMessage != null) {
```
yift-r3: Probably worth adding a comment on why we are doing that.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/QueueRegistrationHandler.kt
nikinagy: We would need 2 vals then as there will be 2 records (hopefully), so I am not sure how the nullability would solve this? Can you elaborate on this?

yift-r3: A nullable val for the `statusUpdateMessage` record only, and then you can return a list of the two records if it's null or just one record if it's not null.

Alternatively, you can have an empty list or list of one item for the status update and add it to the current list.

nikinagy: addressed in [39e73df](https://github.com/corda/corda-runtime-os/pull/4216/commits/39e73df572e50c2cedc8ed43571ede1fd41af6e7)

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/QueueRegistrationHandler.kt
nikinagy: addressed in [39e73df](https://github.com/corda/corda-runtime-os/pull/4216/commits/39e73df572e50c2cedc8ed43571ede1fd41af6e7)

yift-r3 (approved):

dimosr (approved):
LGTM - did you test this end-to-end in any way? One thing that could be done as a sanity check is setup manual approval and send a few registration requests for the same node (without approving any) to confirm that they will all get updated to `RECEIVED_BY_MGM` regardless of any queuing.

nikinagy:
> LGTM - did you test this end-to-end in any way? One thing that could be done as a sanity check is setup manual approval and send a few registration requests for the same node (without approving any) to confirm that they will all get updated to `RECEIVED_BY_MGM` regardless of any queuing.

No, I haven't done any manual tests.

nikinagy:
Testing done:
1) Configured manual approval.
2) Submitted 3 requests: first request got to `PENDING_MANUAL_APPROVAL`, others were in `RECEIVED_BY_MGM` and got queued as expected.
Screenshot 2023-07-05 at 09 24 20
3) MGM approved first request, remaining requests got processed and moved to declined due to serial clash:
Screenshot 2023-07-05 at 09 26 49

PR #4217, opened 2023-07-04
CORE-14309 Interop - fix java.lang.reflect.method serialization issue

szymonsztuka:
New Interopability Facade code contains a data class with `java.lang.reflect.Method` field: 
```
data class FacadeMethodBinding(
    val facadeMethod: FacadeMethod,
    val interfaceMethod: Method,
     ....
```
This field was failing when being serialized by Kryo. 
This is a potential fix to add a custom serializer - in this PR.
The other option would be to replace `val interfaceMethod: Method` with fields (or data class with 3 fields) allowing to construct an object of  `Method` type - so exactly as the new serializer does, this would not require new code in Kryo and would shift the extra code to the Facade.


szymonsztuka:
@chrisr3 Hi Chris is acceptable to add a serializer for the `java.lang.reflect.Method` type to Kryo, or it would be better to restructure the "offending" field into custom class which can be processed by Kryo out-of-box -see the PR description.

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/serializers/MethodSerializer.kt
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Suppress("SpreadOperator", "TooGenericExceptionThrown")
+    override fun read(kryo: Kryo, input: Input, type: Class<out Method>): Method {
+        val declaringClassName = input.readString()
+        val methodName = input.readString()
+        val parameterTypes = kryo.readObject(input, arrayOf<Class<*>>()::class.java)
+
+        return try {
+            val declaringClass = Class.forName(declaringClassName)
```
chrisr3, line 29: This code will be running inside an OSGi framework, but you are effectively trying to load "declareingClassName" (whatever that may be, and whatever bundle it may live in) from `MethodSerializer.class.getClassloader()`.

I really don't see how this could ever work.

Perhaps you could declare your `Method` field to be `transient`, and then recreate it should it ever be `null`?

szymonsztuka: It's working (at least in the tested cases) as the error mentioned here https://r3-cev.atlassian.net/browse/CORE-14309 is fixed by this PR.

_Perhaps you could declare your Method field to be transient, and then recreate it should it ever be null?_. -yes so more or less, the second option from my question and then no changes to Kyro serialization.

chrisr3: > It's working (at least in the tested cases) as the error mentioned here https://r3-cev.atlassian.net/browse/CORE-14309 is fixed by this PR.

Uh huh. By all mean, please explain how `Class.forName()` is able to locate the correct bundle / class loader to load your class by name... :thinking:

szymonsztuka: Thank @chrisr3 for insights, I don't know why it works or why it shouldn't. 
As this is a feature branch, we would like to merge this code to unblock some other work, however we have a ticket https://r3-cev.atlassian.net/browse/CORE-15338 to properly implement it - remove the need for custom Kryo Serializer as per your comment (any my original description).

PR #4218, opened 2023-07-05
CORE-13570 Default datasource properties for vnode connections

kyriathar:
* introduces vnode datasource config for DDL and DML connections in DB table: `db_connections` which only contains "pool" config.
* adds dedicated vnode datasource configs for DDL and DML connections in helm charts.


charts/corda/values.schema.json
```diff
@@ -913,6 +913,134 @@
                                 }
                             }]
                         },
+                        "virtualNodesDdlPoolConfig": {
+                            "type": "object",
+                            "default": {},
+                            "title": "TODO configuration",
```
davidcurrie: Resolve `TODO` here and elsewhere

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -145,8 +145,10 @@ spec:
...
+        {{- include "corda.generateAndExecuteSql" ( dict "name" "vnodes-dml-pool-config" "longName" "virtual-nodes-dml-pool-config" "dbName" "virtualNodesDmlPoolConfig" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "DB_CLUSTER" "sequenceNumber" 9) | nindent 8 }}
+        {{- include "corda.generateAndExecuteSql" ( dict "name" "crypto" "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "CRYPTO_DB_USER"  "schema" "CRYPTO" "sequenceNumber" 11) | nindent 8 }}
+        {{- include "corda.generateAndExecuteSql" ( dict "name" "rest"  "Values" .Values "Chart" .Chart "Release" .Release "environmentVariablePrefix" "REST_API_ADMIN"  "schema" "RBAC"  "searchPath" "RBAC" "subCommand" "create-user-config" "namePostfix" "admin" "sqlFile" "rbac-config.sql" "sequenceNumber" 13) | nindent 8 }}
         - name: 11-create-db-users-and-grant
```
davidcurrie: the sequence numbers should follow on...
```suggestion
        - name: 15-create-db-users-and-grant
```

charts/corda/values.yaml
```diff
@@ -323,6 +323,36 @@ bootstrap:
             name: ""
             # -- the password secret key
             key: ""
+    virtualNodesDdlPoolConfig:
```
davidcurrie: On naming: everything in here is "config". I'd also be tempted to structure it to be more in line with the current `crypto` and `rbac` stanzas. So, without getting too carried away with layers of nesting, maybe something like:
```
virtualNode:
  ddlConnectionPool:
     ...
  dmlConnectionPool:
     ...
  ```
  
  Also, a whole load more `TODO` in here that need resolving.

**Review by davidcurrie — CHANGES_REQUESTED** [review]
Reviewed for chart changes...

File: charts/corda/values.yaml (hunk 3 above)
Comment (kyriathar) [reply charts/corda/values.yaml]: Good catch this one. I pushed a renaming change, but kept structure same, to (**I think**) be a bit more consistent with how other configurations (e.g. `rbac` and `crypto`) map to a dedicated DB row each (i.e. a row per db connection in `db_connection` table).

File: charts/corda/values.yaml [hunk 4]
```diff
@@ -323,6 +323,38 @@ bootstrap:
             name: ""
             # -- the password secret key
             key: ""
+    # Virtual Nodes DDL connections configuration for Corda managed Virtual Nodes DBs
```
simon-johnson-r3: I'm not sure the term "DBs" here makes sense. At this level we only support using the cluster db as a vnode db, and that would seem to be aligned with the fact this is under a `db:` parent.

Perhaps consider:
Virtual Nodes DDL connection configuration for each Corda managed Virtual Node schema in the db

And the same for DML

charts/corda/values.schema.json
```diff
@@ -913,6 +913,134 @@
                                 }
                             }]
                         },
+                        "virtualNodesDdl": {
+                            "type": "object",
+                            "default": {},
+                            "title": "Virtual Nodes DDL connections configuration for Corda managed Virtual Nodes DBs",
```
simon-johnson-r3: See comment on values.yaml for proposed slight rewording

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDbFactoryImpl.kt
```diff
@@ -237,12 +222,31 @@ private fun createDbConfig(
     if (jdbcDriver != null)
         config = config.withValue(DatabaseConfig.JDBC_DRIVER, ConfigValueFactory.fromAnyRef(jdbcDriver))
     config = config.withValue(DatabaseConfig.JDBC_URL, ConfigValueFactory.fromAnyRef(jdbcUrl))
-    config = config.withValue(DatabaseConfig.DB_POOL_MAX_SIZE, ConfigValueFactory.fromAnyRef(maxPoolSize))
-    if (minPoolSize != null)
+
```
simon-johnson-r3: This function should be surplus to requirements after this change, so it probably needs a little attention.

Before this change:
- All db connections are represented by `DataSource`s
- All `DataSource`s are created via configs
- Configs come from db_connections rows, or in the case of the cluster/config, from startup params (because it has to)
- Vnodes are a special case. We don't have anywhere to grab a config from, so we create a config at runtime from hard coded values, using this function `createDbConfig`.

After this change:
- Vnodes get config from rows in the db_connections table, alleviating the need to create a config at runtime.

As discussed there is an issue of the JDBL url and authentication. So I would expect a helper method to inject those into the config we already have. That roughly tallies with the parameters I see passed here. So I think they key is to change this function to reflect that only.

charts/corda/values.yaml
dickon: I'm not a huge fan of stating DDL and DML here; not everyone is going to know those acronyms, and what's in this stanza is not the DDL or DML itself, but rather connection details for setting up the definitions in the database, or doing CRUD oeprations.

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeWriterFactory.kt
```diff
@@ -97,7 +101,27 @@ internal class VirtualNodeWriterFactory(
             publisher
         )
 
-        val virtualNodeDbFactory = VirtualNodeDbFactoryImpl(dbConnectionManager, virtualNodeDbAdmin, schemaMigrator)
+        // TODO The following configs and `virtualNodeDbFactory` do not depend on dynamic config so they could be extracted out to not be
```
simon-johnson-r3: This may be true, but it wasn't affected by this PR. I would delete this comment and raise a ticket on that. But I'm not even entirely sure we need to raise that, it's an optimisation we might not benefit from.

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeWriterFactory.kt
```diff
@@ -97,7 +101,27 @@ internal class VirtualNodeWriterFactory(
...
+        //  re-created on dynamic config update.
+        val virtualNodesDdlPoolConfig =
+            requireNotNull(dbConnectionManager.getDataSourceConfig(VIRTUAL_NODES_DDL_POOL_CONFIG, DbPrivilege.DDL)) {
```
simon-johnson-r3: `require...` functions throw `IllegalArgumentException` which would be a bit odd in the middle of a function where no arguments are being interrogated. I think you might want `check...` equivalent which will throw `IllegalStateException` and is designed for more runtime state like things.

simon-johnson-r3 (changes requested):

charts/corda-lib/templates/_bootstrap.tpl
kyriathar: Done

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeWriterFactory.kt
dickon: Nice tip, @simon-johnson-r3 , I've probably been overusin `require`.

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeWriterFactory.kt
kyriathar: Sure, it shouldn't make much difference I agree, only re-creating it while we couldn't. Removing comment as it's minor.

charts/corda/values.schema.json
kyriathar: I believe I resolved them

components/db/db-connection-manager/src/main/kotlin/net/corda/db/connection/manager/DbConnectionsRepository.kt
```diff
@@ -75,4 +76,6 @@ interface DbConnectionsRepository {
      * @return The cluster DB [DataSource]
      */
     fun getClusterDataSource(): CloseableDataSource
+
+    fun getDataSourceConfig(name: String, privilege: DbPrivilege): Config?
```
dickon: Needs kdoc. But really I don't know what you're trying to achieve here, and I'm not 100% sure about building out ahead of implementing the methods but having both getClusterDataSource and getDataSourceConfig confuses me a bit. I can live with this for now though.

dickon (commented):
It seems a shame to have to list out these config parameters in so many places, there's some inconclusive chat on slack about this.

components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/LateInitDbConnectionOps.kt
```diff
@@ -49,6 +50,9 @@ class LateInitDbConnectionOps: DbConnectionOps {
     override fun getDataSource(config: SmartConfig): CloseableDataSource =
         delegate.getDataSource(config)
 
+    override fun getDataSourceConfig(name: String, privilege: DbPrivilege): Config? =
```
driessamyn: I think this needs to be `SmartConfig` (same everywhere else) as it can contain "secrets" (e.g. password).

If we want to delay turning into a config object (I can see we turn it into `SmartConfig` later on), then we should just treat this as a String I think.

kyriathar: I believe at this point we are only reading the vnodes datasource pool configuration from the DB so no secrets at this point (?) and then we are turning it into a complete vnode connection configuration which contains a secret which is the password of the DB user.

components/db/db-connection-manager/src/main/kotlin/net/corda/db/connection/manager/DbConnectionsRepository.kt
kyriathar: It's a good point actually. Here we are trying to handle this "special" case of vnodes datasource configuration, out of which we only need the pool configuration. 

Whereas in all other cases configurations in `db_connection` table maps to a `Datasource` (per record/ row).

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeWriterFactory.kt
kyriathar: Done

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeWriterFactory.kt
kyriathar: Removed

charts/corda/values.yaml
kyriathar: Good point.

Updated as suggested.

charts/corda/values.schema.json
kyriathar: Done

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDbFactoryImpl.kt
kyriathar: Good catch! Yes this function is currently only being used to create Vnode datasource config.

Updated its name to `createVirtualNodeDbConfig`.

components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/LateInitDbConnectionOps.kt
driessamyn: We're explicitly stating in the documentation that any config object can contain a secret. I.e. we don't define what is secret and what isn't, that is down to the operator.
Whether it's likely one of those things will be a secret is a different question. But what do we gain here by turning it into `Config`?

kyriathar: I believe only making the API a bit more explicit/ clear but not much more than that.

simon-johnson-r3: I think @driessamyn has a point, and I think this relates to my comment below too. And also the comment @dickon makes actually.

The intent of this change is to make vnode connection configs more like other connection configs. 

I think `VirtualNodeDbFactoryImpl` instead of having a `createVirtualNodeDbConfig` function, should perhaps have access to `dbConnectionManager`, from which it extracts a data source directly, just like everything else. The only difference here is you need to inject (replace) the jdbc url and authentication.

You have 2 options for that
1) Pass those into a new api `getDataSource(jdbcOverrides...)`
~~2) Get the smart config and write over them in `VirtualNodeDbFactoryImpl`~~

simon-johnson-r3: Actually that wouldn't work for (2) because this is a get method, you'd potentially be sharing a datasource and overwriting values in it which wouldn't work. Might have to do (1).

simon-johnson-r3: Don't think I was thinking option (2) through actually...

kyriathar: I see what you mean, good point.

I think the issue with that is that we need to get a a datasource config to pass it into `DbConnectionImpl` (it needs a `SmartConfig`  at this point).

kyriathar:
For this work, we are going to go with the approach in [#1186](https://github.com/corda/corda-api/pull/1186) and [#4349](https://github.com/corda/corda-runtime-os/pull/4349).

Adding the vnode pool configs in DB table `db_connections` (which is what the current PR does) looked worse since it was only pool config and not a connection.

PR #4223, opened 2023-07-05
CORE-13164 Interop/ Facade demo flows enabled by Facade serialization and useage fixes

szymonsztuka:
Show full usage of interop API:
- removal of hardcoded Interop Identity name in Swap flow - it's now hardcoded inside lookup API and it matches the deployment script ("C=GB, L=London, O=Bob2 Alias")
- selecting InteropGroup ID in flow REST request (for now it needs to match hardcoded "3dfc0aae-be7c-44c2-aa4f-4d0d7145cf08"
- sending back inner-ledger response via session was moved outside the subflow - to highlight the subflow deals only with calling a proxy
Minor service code improvements - checking fro null etc.


testing/cpbs/interop-demo/deploy.sh
```diff
@@ -5,7 +5,7 @@ context=docker-desktop
 namespace=corda
 groupPolicyFile=GroupPolicy1
 cpi=demo1
-identity1=("C=GB, L=London, O=Alice" "C=GB, L=London, O=Bob" "C=GB, L=London, O=Notary")
```
jmacmahonr3, line 8: What's the reason for this change? Seems unnecessary ?

jmacmahonr3 (approved):
LGTM but will ask some questions in stand up

testing/cpbs/interop-demo/deploy.sh
szymonsztuka: In a loop below the script takes the last entry as ALICE_HASH, so this needs to be reversed.

PR #4224, opened 2023-07-05
CORE-15052: Don't follow subscription status in MembershipPersistenceServiceImpl

yift-r3:
* Add timeouts to the static member onboarding utility to prevent the test from running forever.
* Stop following the statuses of the subscriptions, as this is no longer needed and likely cause a timing issue.

I ran it a few times, and the issue seemed to have gone.

@fowlerrr - FYI

### DIFF
--- a/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/MembershipPersistenceServiceImpl.kt
+++ b/components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/MembershipPersistenceServiceImpl.kt
@@ -93,7 +93,6 @@ class MembershipPersistenceServiceImpl @Activate constructor(
     private var retryManager: MembershipPersistenceAsyncRetryManager? = null
 
     private var dependencyServiceHandle: RegistrationHandle? = null
-    private var subHandle: RegistrationHandle? = null
     private var configHandle: AutoCloseable? = null
 
     override val isRunning: Boolean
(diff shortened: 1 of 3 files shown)


components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/MembershipPersistenceServiceImpl.kt
```diff
@@ -231,14 +219,13 @@ class MembershipPersistenceServiceImpl @Activate constructor(
...
-        asyncSubscription = secondSubscription
-        subHandle = coordinator.followStatusChangesByName(
-            setOf(
-                firstSubscription.subscriptionName,
-                secondSubscription.subscriptionName,
-            )
+
+        coordinator.updateStatus(
+            LifecycleStatus.UP,
+            "Received config and started RPC topic subscription."
```
dimosr: Not that relevant to the change here but given we're at it, should this be updated now given we start 2 subscriptions, one on `MEMBERSHIP_DB_RPC_TOPIC` and `MEMBERSHIP_DB_ASYNC_TOPIC`?

dimosr (commented):
Code LGTM - just a minor comment. 

Also for my understanding could you elaborate on:
> Stop following the statuses of the subscriptions, as this is no longer needed and likely cause a timing issue

Was there a specific reason we needed this before, which doesn't exist anymore?

yift-r3:
> Code LGTM - just a minor comment.
> 
> Also for my understanding could you elaborate on:
> 
> > Stop following the statuses of the subscriptions, as this is no longer needed and likely cause a timing issue
> 
> Was there a specific reason we needed this before, which doesn't exist anymore?

I'm not sure it was ever needed. I'm following the lead from https://github.com/corda/corda-runtime-os/pull/3924

dimosr (approved):

PR #4229, opened 2023-07-06
CORE-14335: cache signing keys based on public key, irrespecitve of tenant ID

dickon:
Keep a single cache mapping signing key hashes to metadata about that key, shared across the entire crypto processor instance, plus a cache of short hashes to full hashes. Previously, the crypto processor stored a cache per tenant ID (i.e. per vnode), and would not store information about signing keys owned by other tenants. Therefore, we saw various flows repeatedly check ownership of a signing key and the crypto processor going to the database each time. (While this was being implemented, we also arranged for the ledger to cache key ownership. That reduces but does not eliminate the benefit of this change).

To facilitate this change, we combine the signing service with the crypto service, and remove more of the support for non-soft HSMs. That removes around 2300 lines of code; those layers and conditonals will all have tiny performance impacts. We had cleanup tickets in the backlog about all these changes, which this PR implements in order to provide a viable implementation of the new caching strategy. In particular:

- CORE-15297: limit use of OSGi to the top level crypto processor instance
- CORE-15276: remove the unused crypto throttling layer, which was never triggered. The architectural choice of Corda 5 is to retry from the top, not at each layer, and crypto is never at the top. Mutliple levels of retires are problematic.
- CORE-15277: move the handling of exceptions out of a proxy layer into particular places where exceptions are expected. The main motivation was to reduce the number of places we have to touch on each interface change.

This required some movement of pieces to keep the package import graph from becoming cyclic with the above changes. In particular SigningKeyInfo and SigningKeySaveContext more into the persistence package.
remove many test cases, which are testing layers that no longer exist, and move other test cases to reflect the newly simplified structure.

### DIFF
--- a/components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/CryptoOperationsTests.kt
+++ b/components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/CryptoOperationsTests.kt
@@ -87,10 +87,10 @@ class CryptoOperationsTests {
         @JvmStatic
         @BeforeAll
         fun setup() {
-            factory = TestServicesFactory()
+            tenantId = UUID.randomUUID().toString()
+            factory = TestServicesFactory(tenantId)
             schemeMetadata = factory.schemeMetadata
             verifier = factory.verifier
-            tenantId = UUID.randomUUID().toString()
             category = CryptoConsts.Categories.LEDGER
             CryptoConsts.Categories.all.forEach {
                 factory.tenantInfoService.populate(tenantId, it, factory.cryptoService)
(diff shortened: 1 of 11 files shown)


Sakpal:
@dickon if I recollect correctly from our previous conversations, the Caching change has performance benefits right?

Did you get a chance to run performance tests on the changes? If yes, can you please post the link to results of the tests?

dickon:
> @dickon if I recollect correctly from our previous conversations, the Caching change has performance benefits right?
> 
> Did you get a chance to run performance tests on the changes? If yes, can you please post the link to results of the tests?

I referred to this in the description here:

> That reduces but does not eliminate the benefit of this change, but means we'd need new flows to see a performance benefit that shows up against experimental noise.

so no, I've not run the settlement app since :
(a) this isn't going to 5.0 where that's needed and 
(b) we won't see any observable benefit plus
(c) there's only one cluster which is trusted to produce menaingful results, and  I understand is exclusively for 5.0 performance work. 

We need to address making sure that we manage performance as 5.1 (or whatever the next release is called) is developed, and maybe that'll bring in PR gates, but that's a project-wide effort which hasn't delivered (or been started?) yet

dickon:
Closing due to a design problem with ShortHash collisions, which would be a regression.

PR #4230, opened 2023-07-06
CORE-8695 Add batching logic to backchain resolution

nkovacsx:
### Overview
This PR includes:
- Batching logic for the backchain resolution. This is either coming from the config `corda.ledger.utxo.backchain.batchsize` or if this is not available for some reason it will default to `BACKCHAIN_DEFAULT_BATCH_SIZE`.
- In order to retrieve the ledger config from a flow we had to make certain modifications so that the flow context can include other configs not just the flow config.

### Testing
A custom set of flow were used to generate an `n` sized backchain and multiple scenarios were tested for performance. Please see more details here: https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4536270888/Test+9+-+Add+batching+to+back-chain+resolution.


components/flow/flow-service/src/main/kotlin/net/corda/flow/application/config/impl/FlowConfigServiceImpl.kt
```diff
@@ -0,0 +1,33 @@
...
+
+@Component(
+    service = [ FlowConfigService::class, UsedByFlow::class ],
+    property = [SandboxConstants.CORDA_SYSTEM_SERVICE],
+    scope = ServiceScope.PROTOTYPE
+)
+class FlowConfigServiceImpl @Activate constructor(
+    @Reference(service = FlowFiberService::class)
+    private val flowFiberService: FlowFiberService
+) : FlowConfigService, UsedByFlow, UsedByPersistence, UsedByVerification, SingletonSerializeAsToken {
```
lankydan: Doesn't need `UsedByPersistence` and `UsedByVerification`

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/config/impl/FlowConfigServiceImpl.kt
```diff
@@ -0,0 +1,33 @@
...
+    override fun getFlowConfigs(): Map<String, SmartConfig> {
+        return flowFiberService.getExecutingFiber().getExecutionContext().configs
+    }
+
+    override fun getLedgerConfig(): SmartConfig? {
+        return getFlowConfigs()[UTXO_LEDGER_CONFIG]
+    }
```
lankydan: I'd rather not have a `getLedgerConfig` method in the flow module. Can we pass in the `key` into this method and have `getConfig(key)` instead?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -42,9 +48,18 @@ class TransactionBackchainReceiverFlowV1(
...
 
+        val ledgerConfig = flowConfigService.getLedgerConfig()
+
+        val batchSize = if (ledgerConfig != null && ledgerConfig.hasPath(BACKCHAIN_BATCH_CONFIG_PATH)) {
+            ledgerConfig.getInt(BACKCHAIN_BATCH_CONFIG_PATH)
+        } else {
+            log.warn("Config path for backchain batch size does not exist, " +
+                    "defaulting batch size to $BACKCHAIN_DEFAULT_BATCH_SIZE")
+            BACKCHAIN_DEFAULT_BATCH_SIZE
+        }
```
lankydan: We shouldn't need this as the defaults in the config should handle it afaik.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/config/impl/FlowConfigServiceImpl.kt
blsemo: I would even word it stronger - because you are not getting ledger config - you are getting UTXO ledger config. The flow layer shouldn't care whether you're using a ledger,  and if yes,  what ledger that is.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainSenderFlowV1.kt
```diff
@@ -37,9 +41,7 @@ class TransactionBackchainSenderFlowV1(private val headTransactionIds: Set<Secur
                         utxoLedgerPersistenceService.find(id)
                             ?: throw CordaRuntimeException("Requested transaction does not exist locally")
                     }
-                    // sending in batches of 1
-                    // TODO Switch to [FlowMessaging.sendAll]
-                    transactions.map { session.send(listOf(it)) }
+                    flowMessaging.sendAll(transactions, setOf(session))
```
lankydan: Why do we need to use `sendAll` here? It should be `session.send(transactions)`

components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowExecutorImpl.kt
```diff
@@ -19,6 +19,7 @@ import net.corda.messaging.api.subscription.factory.SubscriptionFactory
 import net.corda.schema.Schemas.Flow.FLOW_EVENT_TOPIC
 import net.corda.schema.configuration.ConfigKeys.FLOW_CONFIG
 import net.corda.schema.configuration.ConfigKeys.MESSAGING_CONFIG
+import net.corda.schema.configuration.ConfigKeys.UTXO_LEDGER_CONFIG
```
blsemo: Is that import used in this file?

libs/flows/flow-api/src/main/kotlin/net/corda/flow/application/services/FlowConfigService.kt
```diff
@@ -0,0 +1,18 @@
...
+
+    /**
+     * Retrieves all the available configs from the flow context.
+     */
+    fun getFlowConfigs(): Map<String, SmartConfig>
+
+    /**
+     * Retrieves the ledger config from the flow context if available.
+     */
+    fun getLedgerConfig(): SmartConfig?
```
blsemo: As mentioned above, I would prefer
```
fun getConfig( configKey: String ) : SmartConfig?
```

blsemo (commented):
Looks ok to me - just the flow config service interface change.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -42,9 +48,16 @@ class TransactionBackchainReceiverFlowV1(
 
         val sortedTransactionIds = TopologicalSort()
 
+        val ledgerConfig = flowConfigService.getConfig(UTXO_LEDGER_CONFIG)
+
+        requireNotNull(ledgerConfig) {
+            "Could not find UTXO ledger config in the flow context with key: $UTXO_LEDGER_CONFIG"
+        }
```
lankydan: I think this require should be in `getConfig`. For now, lets just make the assumption that all config either exists with a default value or should throw an error upon not being found.

Otherwise things look fine to me.

ronanbrowne (dismissed):
LGTM ( for build files )  although you have a conflict at moment it looks like

libs/flows/flow-api/src/main/kotlin/net/corda/flow/application/services/FlowConfigService.kt
```diff
@@ -0,0 +1,14 @@
+package net.corda.flow.application.services
+
+import net.corda.libs.configuration.SmartConfig
+import net.corda.v5.base.annotations.DoNotImplement
+
+@DoNotImplement
+interface FlowConfigService {
```
lankydan, line 5: It's an internal interface so it doesn't need this annotation.

lankydan (dismissed):

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/factory/impl/FlowEventPipelineFactoryImpl.kt
```diff
@@ -96,20 +98,25 @@ class FlowEventPipelineFactoryImpl(
...
+        configs: Map<String, SmartConfig>,
         mdcProperties: Map<String, String>,
         traceContext:TraceContext,
         eventRecordTimestamp: Long
     ): FlowEventPipeline {
-        val flowCheckpoint = flowCheckpointFactory.create(event.flowId, checkpoint, config)
+        val flowCheckpoint = flowCheckpointFactory.create(
+            event.flowId,
+            checkpoint,
+            configs[FLOW_CONFIG]!!
```
LWogan: Never a good idea to have !! in the code.
 use configs.getConfig(FLOW_CONFIG)

components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowExecutorImpl.kt
```diff
@@ -67,17 +67,26 @@ class FlowExecutorImpl constructor(
     override fun onConfigChange(config: Map<String, SmartConfig>) {
         try {
             val messagingConfig = toMessagingConfig(config)
-            val flowConfig = config.getConfig(FLOW_CONFIG)
-                .withValue(PROCESSOR_TIMEOUT, ConfigValueFactory.fromAnyRef(messagingConfig.getLong(PROCESSOR_TIMEOUT)))
-                .withValue(MAX_ALLOWED_MSG_SIZE, ConfigValueFactory.fromAnyRef(messagingConfig.getLong(MAX_ALLOWED_MSG_SIZE)))
+            config.getConfig(FLOW_CONFIG) // We need to make sure the flow config is present
```
LWogan: this line isnt doing anything

components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowExecutorImpl.kt
```diff
@@ -67,17 +67,26 @@ class FlowExecutorImpl constructor(
...
+            val updatedConfigs = config.mapValues {
```
LWogan: Do you think this might be cleaner? I'm not sure. ill leave it up to you. 


```
            val updatedConfigs = config.toMutableMap()
            updatedConfigs.compute(FLOW_CONFIG) { _, old ->
                old?.withValue(
                    PROCESSOR_TIMEOUT, ConfigValueFactory.fromAnyRef(messagingConfig.getLong(PROCESSOR_TIMEOUT))
                )?.withValue(
                    MAX_ALLOWED_MSG_SIZE, ConfigValueFactory.fromAnyRef(messagingConfig.getLong(MAX_ALLOWED_MSG_SIZE))
                )
            }
`

**Review by LWogan — COMMENTED** [review]
A couple comments

File: components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowExecutorImpl.kt (hunk 10 above)
Comment (nkovacsx) [reply components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowExecutorImpl.kt]: It will throw an exception if the `FLOW_CONFIG` doesn't exist. This is how it used to be before the change.

File: components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/factory/impl/FlowEventPipelineFactoryImpl.kt (hunk 9 above)
Comment (nkovacsx) [reply components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/factory/impl/FlowEventPipelineFactoryImpl.kt]: Good point, thanks

File: components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowExecutorImpl.kt (hunk 11 above)
Comment (nkovacsx) [reply components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowExecutorImpl.kt]: I'm not sure either. It seems more complex to me but that might be just preference. We also need to make something mutable in order to use this solution which I'm not a huge fan of.

File: components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowExecutorImpl.kt (hunk 10 above)
Comment (LWogan) [reply components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowExecutorImpl.kt]: Thats not quite true. It was called before to get the config, update it and set it to a field. 

The name of the function is getConfig. I dont think it makes the code very readable to call a get to validate something exists and discard the result. How about this?

```
val flowConfig = config.getConfig(FLOW_CONFIG)
                .withValue(PROCESSOR_TIMEOUT, ConfigValueFactory.fromAnyRef(messagingConfig.getLong(PROCESSOR_TIMEOUT)))
                .withValue(MAX_ALLOWED_MSG_SIZE, ConfigValueFactory.fromAnyRef(messagingConfig.getLong(MAX_ALLOWED_MSG_SIZE)))

val updatedConfigs = config.mapValues {
                if (it.key == FLOW_CONFIG) {
                    flowConfig
                } else {
                    it.value
                }
            }
```

Also I think it would also be cleaner to do this manipulation of the configs in a separate method in this class

nkovacsx: Fair point, thanks for the suggestions!

LWogan (dismissed):
LGTM from flow perspective

lankydan (dismissed):

LWogan (commented):
Please do not merge this pr until after the flow session & wakeup removal pr is merged this morning  which is due to be released in a beta today

LWogan (approved):
LGTM

lankydan (approved):

PR #4231, opened 2023-07-06
CORE-15303: Remove Kotlin reflection from UtxoFinalityFlowV1.

chrisr3:
Do not use Kotlin reflection because it is unreliable inside an OSGi framework. The `primaryConstructor` function also assumes that all pluggable notary classes will be written in Kotlin, which is not required.

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
@@ -19,6 +19,7 @@ import net.corda.v5.application.messaging.FlowSession
 import net.corda.v5.base.annotations.Suspendable
 import net.corda.v5.base.annotations.VisibleForTesting
 import net.corda.v5.base.exceptions.CordaRuntimeException
+import net.corda.v5.base.types.MemberX500Name
 import net.corda.v5.ledger.notary.plugin.api.PluggableNotaryClientFlow
 import net.corda.v5.ledger.notary.plugin.core.NotaryExceptionFatal
 import net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction
@@ -26,7 +27,6 @@ import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import java.security.AccessController
 import java.security.PrivilegedExceptionAction
-import kotlin.reflect.full.primaryConstructor
 
 @CordaSystemFlow
 class UtxoFinalityFlowV1(
@@ -52,7 +52,7 @@ class UtxoFinalityFlowV1(
     @Suspendable
     override fun call(): UtxoSignedTransaction {
         addTransactionIdToFlowContext(flowEngine, transactionId)
-        log.trace("Starting finality flow for transaction: $transactionId")
+        log.trace("Starting finality flow for transaction: {}", transactionId)
         verifyExistingSignatures(initialTransaction)
         verifyTransaction(initialTransaction)
 
@@ -67,7 +67,7 @@ class UtxoFinalityFlowV1(
         val (notarizedTransaction, notarySignatures) = notarize(transaction)
         persistNotarizedTransaction(notarizedTransaction)
         sendNotarySignaturesToCounterparties(notarySignatures)
-        log.trace { "Finalisation of transaction $transactionId has been finished." }
+        log.trace("Finalisation of transaction {} has been finished.", transactionId)
         return notarizedTransaction
     }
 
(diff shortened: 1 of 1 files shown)


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -52,7 +52,7 @@ class UtxoFinalityFlowV1(
     @Suspendable
     override fun call(): UtxoSignedTransaction {
         addTransactionIdToFlowContext(flowEngine, transactionId)
-        log.trace("Starting finality flow for transaction: $transactionId")
+        log.trace { "Starting finality flow for transaction: $transactionId" }
```
blsemo: Are you sure that this lambda cannot blow up in an unexpected way in Quasar? We've had issues with that in the past.

blsemo (commented):
Fine with the actual change - can we double check the trace lambda?

relyafi (commented):
Change makes sense, but I wonder if there's a way of enforcing a code analysis check to prevent use of reflection generally within OSGi code if it is potentially problematic? That would be supplementary to this change, but might give us additional protection against similar issues being introduced in future as the code base evolves.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
chrisr3: There's already a `log.trace {}` lambda at the bottom of this exact same function, and I know of no rational explanation why one should be OK but not the other.

Failing that, we _could_ convert both `log.trace {}` to use SLF4J placeholders - _anything_ but this **preposterous** string concatenation!

chrisr3: Fine - I _hate_ `log.trace {}` anyway.

blsemo (approved):
LGTM

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -52,7 +52,7 @@ class UtxoFinalityFlowV1(
     @Suspendable
     override fun call(): UtxoSignedTransaction {
         addTransactionIdToFlowContext(flowEngine, transactionId)
-        log.trace("Starting finality flow for transaction: $transactionId")
+        log.trace("Starting finality flow for transaction: {}", transactionId)
```
mnesbit, line 55: Trace should not be done using direct call. It should always use lambda version to prevent the very expensive logging call

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -67,7 +67,7 @@ class UtxoFinalityFlowV1(
         val (notarizedTransaction, notarySignatures) = notarize(transaction)
         persistNotarizedTransaction(notarizedTransaction)
         sendNotarySignaturesToCounterparties(notarySignatures)
-        log.trace { "Finalisation of transaction $transactionId has been finished." }
+        log.trace("Finalisation of transaction {} has been finished.", transactionId)
```
mnesbit, line 70: as above

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -273,7 +273,7 @@ class UtxoFinalityFlowV1(
         transaction: UtxoSignedTransactionInternal
     ): PluggableNotaryClientFlow {
         return AccessController.doPrivileged(PrivilegedExceptionAction {
-            pluggableNotaryClientFlow.kotlin.primaryConstructor!!.call(
+            pluggableNotaryClientFlow.getConstructor(UtxoSignedTransaction::class.java, MemberX500Name::class.java).newInstance(
```
mnesbit, line 276: Make sense

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
chrisr3: See @blsemo's comment, where the lamba version was _explicitly_ removed from here because it could "blow up in an unexpected way in Quasar".

In which case, the placeholder version is still _waay_ better than mandatory string concatenation.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
chrisr3: If `log.trace {}` could explode with Quasar at the start of this function then I don't see why using it at the end of this same function should be any less dangerous.

driessamyn (approved):
LGTM

PR #4234, opened 2023-07-06
CORE-14573: Wrong status code reported for incorrect tenantId

anien:
When requesting a CSR for a tenant, we were not checking if a virtual node is registered for the tenant.
ValidateTenantId check was updated and it now returns 400 Bad Request if virtual node is not registered for the tenant.
As we do this check early on, tests had to be updated too to make sure they test the original functionality.

### DIFF
--- a/components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/CertificatesRestResourceImpl.kt
+++ b/components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/CertificatesRestResourceImpl.kt
@@ -8,7 +8,6 @@ import net.corda.crypto.cipher.suite.schemes.GOST3410_GOST3411_TEMPLATE
 import net.corda.crypto.client.CryptoOpsClient
 import net.corda.crypto.core.CryptoConsts
 import net.corda.crypto.core.CryptoTenants.P2P
-import net.corda.crypto.core.CryptoTenants.REST
 import net.corda.crypto.core.CryptoTenants.allClusterTenants
 import net.corda.crypto.core.DefaultSignatureOIDMap
 import net.corda.crypto.core.ShortHash
@@ -26,7 +25,6 @@ import net.corda.membership.rest.v1.CertificatesRestResource
 import net.corda.membership.rest.v1.CertificatesRestResource.Companion.SIGNATURE_SPEC
 import net.corda.rest.HttpFileUpload
 import net.corda.rest.PluggableRestResource
-import net.corda.rest.exception.BadRequestException
 import net.corda.rest.exception.InvalidInputDataException
 import net.corda.rest.exception.ResourceNotFoundException
 import net.corda.rest.messagebus.MessageBusUtils.tryWithExceptionHandling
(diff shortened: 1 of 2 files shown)


components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/CertificatesRestResourceImpl.kt
```diff
@@ -444,9 +446,21 @@ class CertificatesRestResourceImpl @Activate constructor(
...
+        } catch(e: BadRequestException) {
+            false
+        }
+        catch (e: ResourceNotFoundException) {
+            false
+        }
+
+        val isValidClusterTenant = tenantId in allClusterTenants
+
+        if ((!isValidShortHash && !isValidClusterTenant) || ( isValidShortHash && !isValidClusterTenant && !isVirtualNodeRegisteredForTenant)) {
```
dickon: I think this could be clearer as something like:

```
if (tenantId in allClusterTenants) return

if (!isValidShortHash) throw InvalidInputDataException("Provided tenantId $tenantId is not a cluster tenant and is not the right size of a holding ID")

val isVirtualNodeRegisteredForTenant = try {
            virtualNodeInfoReadService.getByHoldingIdentityShortHashOrThrow(tenantId)
            true
        } catch(e: BadRequestException) {
            false
        }
        catch (e: ResourceNotFoundException) {
            false
        }
if (!isVirtualNodeRegisteredForTenant) { 
 throw InvalidInputDataException(
                "Provided tenantId ($tenantId) was not valid.; it is not a cluster tenant and not a valid virutal node holding ID" 
 )
}
```

anien: Thanks! It's indeed much more readable.

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/CertificatesRestResourceImpl.kt
```diff
@@ -444,12 +446,25 @@ class CertificatesRestResourceImpl @Activate constructor(
...
+        } catch(e: BadRequestException) {
+            false
+        }
+        catch (e: ResourceNotFoundException) {
+            false
+        }
+
+        if (tenantId in allClusterTenants) return
+        if (!isValidShortHash) throw InvalidInputDataException("Provided tenantId $tenantId is not a cluster tenant " +
+                "and is not the right size of a holding ID.")
```
dimosr: you can move these checks further up, e.g. first check if cluster tenant, then check if valid short hash and lastly check if there's a vnode with that ID. 


This way, you'll save a query to the vnode service (for cases where it's a cluster tenant).

dimosr: Also the second message (`"Provided tenantId $tenantId is not a cluster tenant and is not the right size of a holding ID."`) does not cover all cases as the `parse` method can fail for cases where the string provided is not a valid hex string, even if it's the right size.

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/CertificatesRestResourceImpl.kt
```diff
@@ -444,12 +446,25 @@ class CertificatesRestResourceImpl @Activate constructor(
         } catch (e: ShortHashException) {
             false
         }
-        val isValidClusterTenant = (tenantId == P2P || tenantId == REST)
 
-        if (!isValidShortHash && !isValidClusterTenant) {
+        // Check if a virtual node is registered for given tenantId
+        val isVirtualNodeRegisteredForTenant = try {
```
dimosr: I think this is more `doesVirtualNodeExist` rather than `isVirtualNodeRegisteredForTenant`.

Also, have you considered throwing a 404 in that case when vnode doesn't exist, instead of a 400? I think I've seen that applied in several other APIs.

dimosr (commented):
I believe 5.0 is under freeze at the moment, so this will have to be switched to 5.1. But, since I took a look, leaving some feedback below.

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/CertificatesRestResourceImpl.kt
anien: > you can move these checks further up, e.g. first check if cluster tenant, then check if valid short hash and lastly check if there's a vnode with that ID.
> 
> This way, you'll save a query to the vnode service (for cases where it's a cluster tenant).

I actually had it like that in the first place, but though this is easier to read. But yes, I did realise we might do some unnecessary queries. I will move the checks.

anien:
Closing this PR as a [new one](https://github.com/corda/corda-runtime-os/pull/4296) was opened for 5.1 branch.

PR #4242, opened 2023-07-07
CORE-15278, CORE-10051:CORE-15278, CORE-10051: remove unused types and code that handle situations where the private key was stored in external HSM (multi-HSM concept), since we are not supporting multiple HSMs

dickon:
SiginingKeyInfo.keyMaterial used to be nullable, in order to support HSMs where we do not have the key material (i.e. we do not have access to an encrypted private key). Let's remove support for missing key material in signing key records since it can never happen in production right now with our current feature set. This is a step towards removing the HSM layers in order to squash together code in order to improve performance.

Delete the interface and SigningAliasSpec which is the other implementation class we never use, for when we might have supported HSMs in the future.

Currently CryptoService.sign took a SigningSpec as an argument, which is an interface which had 2 implementations; the method is polymorphic. One variant, SigningAliasSpec, was intended for use when signing when we only know the alias of a key, but not the private key itself. This is not supported, and we do not want to carry the complexity.

Therefore, delete the interface and SigningAliasSpec which is an implementation class we never use, since it was part of the multiple HSM support. This change will make it easier to improve the caching behaviour of the crypto processor, since the current complexity we are addressing here means there can be multiple crypto services, which brings complexity to the fast path.

I considered making SigningWrappedSpec be called SigningSpec. I felt that would potentially be more confusing at this stage, but that's a marginal call. Long run, a lot of the Spec classes in the crypto processor can be replaced by a small number of arguments to various functions.

Similarly, removed GeneratedPublicKey and SigningPublicKeySaveContext, which were unused.


components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
```diff
@@ -65,25 +65,15 @@ class SigningRepositoryTest : CryptoRepositoryTest() {
             tenantId = "t-$unique".take(12),
             category = "c-$unique",
             alias = "a-$unique",
-            hsmAlias = "ha-$unique",
+            hsmAlias = null,
```
kyriathar, line 67: Wondering if this change was needed?

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
```diff
@@ -264,16 +264,15 @@ class SigningRepositoryImpl(
 
 
 fun SigningKeyEntity.joinSigningKeyInfo(em: EntityManager): SigningKeyInfo {
-    val signingKeyMaterialEntity = em.createQuery(
+    val signingKeyMaterialEntity = requireNotNull(em.createQuery(
```
kyriathar: Perhaps this could be replaced with `checkNotNull`, since we are not checking function parameter for nulled value.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
```diff
@@ -264,16 +264,15 @@ class SigningRepositoryImpl(
...
         "FROM ${SigningKeyMaterialEntity::class.java.simpleName} WHERE signingKeyId=:signingKeyId",
         SigningKeyMaterialEntity::class.java
     ).setParameter("signingKeyId", id)
-        .resultList.singleOrNull()
-    val wrappingKey = if (signingKeyMaterialEntity != null) {
-        em.createQuery(
-            "FROM WrappingKeyEntity WHERE id=:wrappingKeyId", WrappingKeyEntity::class.java,
-        ).setParameter("wrappingKeyId", signingKeyMaterialEntity.wrappingKeyId).resultList.singleOrNull()
-    } else null
+        .resultList.singleOrNull(), { "private key material for $id not found"})
```
kyriathar: This lambda argument could be moved out of parenthesis.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
```diff
@@ -282,7 +281,7 @@ fun SigningKeyEntity.joinSigningKeyInfo(em: EntityManager): SigningKeyInfo {
         alias = alias,
         hsmAlias = hsmAlias,
         publicKey = publicKey,
-        keyMaterial = signingKeyMaterialEntity?.keyMaterial,
+        keyMaterial = signingKeyMaterialEntity.keyMaterial,
         schemeCodeName = schemeCodeName,
         masterKeyAlias = wrappingKey?.alias,
```
kyriathar: I believe `wrappingKey` is no longer nullable so this null check could be removed?

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/SigningWrappedSpec.kt
```diff
@@ -12,15 +12,15 @@ import java.security.PublicKey
  * @param keyScheme The scheme for the key used for signing operation.
  * @param signatureSpec The signature spec to use for signing, such as SHA256withECDSA, etc.
  */
-class SigningWrappedSpec(
+data class SigningWrappedSpec(
```
kyriathar, line 15: Minor but I guess if this wasn't a `data` class before probably suggests we were not really using it's `equals` or `hashCode` so maybe we could leave it as not data class.

kyriathar (commented):
I think this is a good change, towards further removing obsolete supporting of multiple HSMs concept. A few minor comments below.

Should we also consider removing as part of this change the followings types `GeneratedPublicKey` and `SigningPublicKeySaveContext`. They also seem to be meant for multiple HSMs and I believe they are not currently used.

components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
dickon: Actually yes, it is, I went back and double chcked and the test will fail without that change.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
dickon: done. Although I'm not sure which was more readable.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
dickon: done

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SigningRepositoryImpl.kt
dickon: You are probably right, so I'll make the wrappingKey string compulsory (i.e. no longer not null) and add a matching `checkNotNull` when retreiving the wrapping key. Let's see if that gets through CI.

dickon: done

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/SigningWrappedSpec.kt
dickon: we could, but I think it's good to always have`equals` or `hashCode`,  so we should use `data class`

dickon:
> Should we also consider removing as part of this change the followings types GeneratedPublicKey and SigningPublicKeySaveContext. They also seem to be meant for multiple HSMs and I believe they are not currently used.

I was wondering about that, and I felt I would like to so I have done so, and rebase the rest of the this, which was quite fiddly.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/SigningServiceImpl.kt
```diff
@@ -392,12 +377,8 @@ class SigningServiceImpl(
...
         tenantId: String,
     ): KeyMaterialSpec {
-        val keyMaterial: ByteArray = record.data.keyMaterial ?: throw IllegalStateException(
-            "The key material is null for public key ${publicKey.publicKeyId()} of tenant $tenantId  "
-        )
-        val masterKeyAlias = record.data.masterKeyAlias ?: throw IllegalStateException(
-            "The master key alias for public key ${publicKey.publicKeyId()} of tenant $tenantId must be specified, but is null"
-        )
+        val keyMaterial: ByteArray = record.data.keyMaterial
+        val masterKeyAlias = record.data.wrappingKeyAlias
```
kyriathar, line 381: Should  `masterKeyAlias` be renamed to `wrappingKeyAlias` here? Feels like some renaming made it half way through.

components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
kyriathar: So I believe the reason why this is needed now is because this was related to multi-HSM concept? I.e. have the hsm alias for the key, i.e. to which HSM the key is stored?

kyriathar (approved):
LGTM. One minor comment but happy for it to be addressed later.

I believe this should be merged by saying something like: "Remove types which were meant to hold only the public parts of the signing key and reference the HSM where their pair private parts were stored"

PR #4243, opened 2023-07-07
CORE-14335: step 3 CORE-9901 - Perform signature verification on the member side with the right key (#4179)

dickon:
Pass signing repository into crypto service.

Make HSMServiceImpl a simple class, without OSGi.

Only test with wrapping key present, since we've dropped support for keyMaterial being null (i.e. public key only mode)

Remove secondary constructor on SigningServiceImpl


dickon:
closed in favour of updated version which now has a green build

PR #4248, opened 2023-07-07
CORE-14709: Adding apiVersion dimension to REST API - fix test exception when introducing new endpoint

szymonsztuka:
After changes introduced in  CORE-14709, when a baseline was not updated for a new endpoint (the new endpoint was present in the baseline but with an older "content"/"format" ) the `OpenApiCompatibilityTest` threw Runtime Exception `Index out of bounds` and it didn't produce the error report and a new baseline.
A sample error from a sibling branch: 
https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda-runtime-os/detail/PR-4239/1/tests/


vkolomeyko (dismissed):
LGTM

vkolomeyko (approved):
LGTM

PR #4250, opened 2023-07-07
CORE-15318 - Investigate flakiness on feature branch

nikinagy:
Pre-auth token failure e2e tests started to fail recently on feature branch due to:
```
[pod/corda-caroline-membership-worker-8cb5747bf-xmtnr/corda-caroline-membership-worker] 13:01:46.381 [Thread-71] WARN  net.corda.membership.impl.registration.dynamic.RegistrationProcessor {} - Declining registration because: Registration attempted to use a pre-auth token which is not currently active for this member.
[pod/corda-caroline-membership-worker-8cb5747bf-xmtnr/corda-caroline-membership-worker] 13:01:46.382 [Thread-71] ERROR net.corda.membership.impl.registration.dynamic.RegistrationProcessor {} - Unexpected error in handling registration command.
[pod/corda-caroline-membership-worker-8cb5747bf-xmtnr/corda-caroline-membership-worker] net.corda.v5.base.exceptions.CordaRuntimeException: Failed to retrieve pending member's info for member with holding ID'BAEFE1257845'.
[pod/corda-caroline-membership-worker-8cb5747bf-xmtnr/corda-caroline-membership-worker] 	at net.corda.membership.impl.registration.dynamic.handler.mgm.DeclineRegistrationHandler.invoke(DeclineRegistrationHandler.kt:62) ~[?:?]
[pod/corda-caroline-membership-worker-8cb5747bf-xmtnr/corda-caroline-membership-worker] 	at net.corda.membership.impl.registration.dynamic.handler.mgm.DeclineRegistrationHandler.invoke(DeclineRegistrationHandler.kt:28) ~[?:?]
[pod/corda-caroline-membership-worker-8cb5747bf-xmtnr/corda-caroline-membership-worker] 	at net.corda.membership.impl.registration.dynamic.handler.RegistrationHandler$invoke$1$1.invoke(RegistrationHandler.kt:17) ~[?:?]
[pod/corda-caroline-membership-worker-8cb5747bf-xmtnr/corda-caroline-membership-worker] 	at net.corda.membership.impl.registration.dynamic.handler.RegistrationHandler$invoke$1$1.invoke(RegistrationHandler.kt:16) ~[?:?]
[pod/corda-caroline-membership-worker-8cb5747bf-xmtnr/corda-caroline-membership-worker] 	at net.corda.membership.impl.registration.dynamic.handler.RegistrationHandler.recordTimerMetric$lambda$1(RegistrationHandler.kt:36) ~[?:?]
```
It seems like the published member info is not available in time in the cache to prepare the p2p message in DeclineRegistrationHandler.


components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/DeclineRegistrationHandler.kt
```diff
@@ -57,33 +58,41 @@ internal class DeclineRegistrationHandler(
         if (!memberTypeChecker.isMgm(declinedBy)) {
             logger.warn("Trying to decline registration request: '$registrationId' by ${declinedBy.x500Name} which is not an MGM")
         }
-        val memberInfo = groupReaderProvider.getGroupReader(declinedBy.toCorda())
-            .lookup(declinedMember.toCorda().x500Name, MembershipStatusFilter.PENDING)
-            ?: throw CordaRuntimeException("Failed to retrieve pending member's info " +
-                    "for member with holding ID'${declinedMember.toCorda().shortHash}'.")
         logger.info("Declining registration request: '$registrationId' for ${declinedMember.x500Name} - ${command.reason}")
+        val pendingMemberInfo = membershipQueryClient.queryMemberInfo(declinedBy.toCorda(), listOf(declinedMember.toCorda()))
```
dimosr, line 62: I remember we discussed the option of incorporating the platform version into the DeclineRegistration command to avoid introducing RPC calls here? What was the reason this wasn't preferred?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
```diff
@@ -156,6 +137,35 @@ internal class StartRegistrationHandler(
                 }
             }
 
+            val persistentMemberInfo = PersistentMemberInfo.newBuilder()
```
dimosr, line 98: If I understand correctly if we require a member info to send the status update for declined requests, that means that in some cases validation will happen before that point and then the member will not be notified of the declined registration request (e.g. because no pending member info is persisted)?

Should we maybe write this in the beginning before we perform any validations?

nikinagy: I left a comment here saying: We don't want to overwrite existing member's information or persist member with wrong name. Most of these checks are about validating the identity of a member eg x500Name. I didn't want to overwrite existing member's data with someone else's data.
Although, some these checks (eg serial is valid/invalid) could be moved after this to make sure we notify member.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/DeclineRegistrationHandler.kt
nikinagy: I decided not to go with it, because DeclineRegistration() is issued at a lot of places and at some of these places (eg ProcessMemberVerificationResponseHandler, VerifyMemberHandler, MembershipP2PMarkersProcessor), we would still had to figure out somehow the members platform version by querying.
Also, it seemed tricky to include the versions in the command since we mostly issued it when we caught an exception. So we would need to somehow put it into the exceptions to have it when we decline.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
dimosr: right, but everything else is in principle something we could persist a pending member info so that we can notify the member? If so, we could have this check for x500 names in the beginning, follow with the persistence of the member infos and then everything else.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/DeclineRegistrationHandler.kt
dimosr: I see, that would probably have zero net benefit as we'd need to add querying in other places to save this one. Ok, we can go with this approach then.

(And on a separate note, we'll probably need to revisit the PoC of executing the handlers within the DB processor at some point)

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
```diff
@@ -86,34 +86,28 @@ internal class StartRegistrationHandler(
...
+
+            val activeOrSuspendedInfo = membershipQueryClient.queryMemberInfo(
+                mgmHoldingId,
+                listOf(pendingMemberHoldingId)
+            ).getOrThrow().lastOrNull {
+                it.status == MEMBER_STATUS_ACTIVE || it.status == MEMBER_STATUS_SUSPENDED
             }
-            validateRegistrationRequest(registrationRequest.serial!! >= 0) {
-                "Serial cannot be negative on the registration request."
+            if (registrationRequest.serial != null && registrationRequest.serial == 0L) { // name check for initial registration
```
dimosr: Will there be any issues if we keep that validation after the persistence of the pending member info, so that we can notify properly for scenarios where a user tried to use an already taken name?

nikinagy: My main problem with this whole thing is who do you notify in this case? If we send out the decline message it will use the wrong name or not? And I think this will result in notifying the wrong member. But maybe I am overthinking.

dimosr: No, you're making a good point. It seems like the decline handler is sending any notifications using the pending member info records (which I think is generally a sound approach for all handlers). So, in that case, the p2p layer should pick up the pending member info for the member that tried to register and failed and use that (i.e. use proper endpoints / session keys etc.).

Thinking about the member's side, I assume that this member must reside in a separate cluster to the one that has reserved that name. Otherwise, it should all have failed much earlier on when the operator on that side tried to create a vnode with the same name. So, routing of the message in that cluster should theoretically work without any confusion. 

All worth a test though to confirm.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/StartRegistrationHandler.kt
```diff
@@ -125,37 +106,68 @@ internal class StartRegistrationHandler(
...
                 listOf(pendingMemberHoldingId)
             ).getOrThrow().lastOrNull {
                 it.status == MEMBER_STATUS_ACTIVE || it.status == MEMBER_STATUS_SUSPENDED
             }
-            if (registrationRequest.serial!! > 0) { //re-registration
+            if (registrationRequest.serial != null && registrationRequest.serial == 0L) { // name check for initial registration
+                validateRegistrationRequest(activeOrSuspendedInfo == null) {
+                    "Member already exists with the same X500 name."
+                }
+            }
```
dimosr: I think it'd be better to preserve this in the original form as an if-else that treats initial registrations / re-registrations together, unless there was a reason not to.

It makes it a bit easier to read/understand and make the conditional smaller, since null checks on `registrationRequest.serial` have already been performed at that stage.

nikinagy: addressed in [1bcff8c](https://github.com/corda/corda-runtime-os/pull/4250/commits/1bcff8c31b10b642447221d1ca9440cfd7e8fc75)

dimosr (approved):

PR #4251, opened 2023-07-07
CORE-14685: Separate UniquenessChecker and BackingStore from Lifecycle.

chrisr3:
Splice `Lifecycle` from `UniquenessChecker` and `BackingStore` so that these components become available without also needing the lifecycle layer.

The lifecycle functionality has been moved to `UniquenessCheckerLifecycle` and `BackingStoreLifecycle`, which host the `UniquenessChecker` and `BackingStore` components respectively.

The underlying objective is for `UniquenessCheckMessageProcessor` not to depend on any components from the "lifecycle" layer, in the same way that:
- `VerifiactionRequestProcessor`
- `PersistenceRequestProcessor`
- `EntityMessageProcessor`

and others already do not.

---
In simple terms, I am replacing an interface like this:
```kotlin
interface Foo : Lifecycle
```
with two interfaces like this:
```kotlin
interface Foo
interface FooLifecycle : Foo, Lifecycle
```
I then split the `FooImpl` class into two like this:
```kotlin
@Component(service = [ Foo::class ])
class FooImpl : Foo

@Component(service = [ FooLifecycle::class ])
class FooLifecycleImpl @Activate constructor(
    @Refererence
    private val foo: Foo
) : FooLifecycle, Foo by foo
```

### DIFF
--- a/components/uniqueness/backing-store-impl/src/integrationTest/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImplIntegrationTests.kt
+++ b/components/uniqueness/backing-store-impl/src/integrationTest/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImplIntegrationTests.kt
@@ -62,7 +62,7 @@ import java.time.LocalDateTime
 import java.time.LocalDate
 import java.time.ZoneOffset
 import java.time.Duration
-import java.util.*
+import java.util.UUID
 import java.util.concurrent.Callable
 import java.util.concurrent.ExecutionException
 import java.util.concurrent.Executors
@@ -84,7 +84,7 @@ import kotlin.reflect.full.createInstance
  */
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class JPABackingStoreImplIntegrationTests {
-    private lateinit var backingStoreImpl: JPABackingStoreImpl
+    private lateinit var backingStoreImpl: JPABackingStoreLifecycleImpl
     private lateinit var lifecycleCoordinator: LifecycleCoordinator
     private lateinit var lifecycleCoordinatorFactory: LifecycleCoordinatorFactory
     private lateinit var testClock: AutoTickTestClock
@@ -139,13 +139,18 @@ class JPABackingStoreImplIntegrationTests {
         testClock = AutoTickTestClock(baseTime, Duration.ofSeconds(1))
     }
 
-    private fun createBackingStoreImpl(emFactory: EntityManagerFactory): JPABackingStoreImpl {
-        return JPABackingStoreImpl(lifecycleCoordinatorFactory,
-            JpaEntitiesRegistryImpl(),
-            mock<DbConnectionManager>().apply {
-                whenever(getOrCreateEntityManagerFactory(any(), any(), any())) doReturn emFactory
-                whenever(getClusterDataSource()) doReturn dbConfig.dataSource
-            })
+    private fun createBackingStoreImpl(emFactory: EntityManagerFactory): JPABackingStoreLifecycleImpl {
+        val jpaEntitiesRegistry = JpaEntitiesRegistryImpl()
+        val dbConnectionManager = mock<DbConnectionManager>().apply {
+            whenever(getOrCreateEntityManagerFactory(any(), any(), any())) doReturn emFactory
+            whenever(getClusterDataSource()) doReturn dbConfig.dataSource
+        }
+        return JPABackingStoreLifecycleImpl(
+            lifecycleCoordinatorFactory,
+            jpaEntitiesRegistry,
+            dbConnectionManager,
+            JPABackingStoreImpl(jpaEntitiesRegistry, dbConnectionManager)
+        )
     }
 
     private fun createEntityManagerFactory(persistenceUnitName: String = "uniqueness"): EntityManagerFactory {
--- a/components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImpl.kt
+++ b/components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImpl.kt
@@ -6,17 +6,8 @@ import net.corda.db.connection.manager.DbConnectionManager
 import net.corda.db.connection.manager.VirtualNodeDbType
 import net.corda.db.core.DbPrivilege
 import net.corda.db.schema.CordaDb
-import net.corda.lifecycle.DependentComponents
-import net.corda.lifecycle.LifecycleCoordinator
-import net.corda.lifecycle.LifecycleCoordinatorFactory
-import net.corda.lifecycle.LifecycleEvent
-import net.corda.lifecycle.RegistrationStatusChangeEvent
-import net.corda.lifecycle.StartEvent
-import net.corda.lifecycle.StopEvent
-import net.corda.lifecycle.createCoordinator
 import net.corda.metrics.CordaMetrics
 import net.corda.orm.JpaEntitiesRegistry
-import net.corda.orm.JpaEntitiesSet
 import net.corda.uniqueness.backingstore.BackingStore
 import net.corda.uniqueness.datamodel.common.UniquenessConstants.HIBERNATE_JDBC_BATCH_SIZE
 import net.corda.uniqueness.datamodel.common.UniquenessConstants.RESULT_ACCEPTED_REPRESENTATION
@@ -28,7 +19,6 @@ import net.corda.uniqueness.datamodel.impl.UniquenessCheckStateDetailsImpl
 import net.corda.uniqueness.datamodel.impl.UniquenessCheckStateRefImpl
 import net.corda.uniqueness.datamodel.internal.UniquenessCheckRequestInternal
 import net.corda.uniqueness.datamodel.internal.UniquenessCheckTransactionDetailsInternal
-import net.corda.utilities.VisibleForTesting
 import net.corda.utilities.debug
 import net.corda.v5.application.uniqueness.model.UniquenessCheckError
 import net.corda.v5.application.uniqueness.model.UniquenessCheckResult
@@ -55,8 +45,6 @@ import javax.persistence.RollbackException
  */
 @Component(service = [BackingStore::class])
 open class JPABackingStoreImpl @Activate constructor(
-    @Reference(service = LifecycleCoordinatorFactory::class)
-    coordinatorFactory: LifecycleCoordinatorFactory,
     @Reference(service = JpaEntitiesRegistry::class)
     private val jpaEntitiesRegistry: JpaEntitiesRegistry,
     @Reference(service = DbConnectionManager::class)
@@ -70,17 +58,12 @@ open class JPABackingStoreImpl @Activate constructor(
         const val MAX_ATTEMPTS = 10
     }
 
-    private val lifecycleCoordinator: LifecycleCoordinator = coordinatorFactory
-        .createCoordinator<BackingStore>(::eventHandler)
-
-    private val dependentComponents = DependentComponents.of(
-        ::dbConnectionManager
-    )
-
-    private lateinit var jpaEntities: JpaEntitiesSet
-
-    override val isRunning: Boolean
-        get() = lifecycleCoordinator.isRunning
+    init {
+        jpaEntitiesRegistry.register(
+            CordaDb.Uniqueness.persistenceUnitName,
+            JPABackingStoreEntities.classes
+        )
+    }
 
     override fun session(holdingIdentity: HoldingIdentity, block: (BackingStore.Session) -> Unit) {
 
@@ -116,16 +99,6 @@ open class JPABackingStoreImpl @Activate constructor(
                 .build()
                 .record(Duration.ofNanos(System.nanoTime() - sessionStartTime))
         }
-    }
-
-    override fun start() {
-        log.info("Backing store starting")
-        lifecycleCoordinator.start()
-    }
-
-    override fun stop() {
-        log.info("Backing store stopping")
-        lifecycleCoordinator.stop()
     }
 
     protected open inner class SessionImpl(
@@ -430,37 +403,6 @@ open class JPABackingStoreImpl @Activate constructor(
                     .withTag(CordaMetrics.Tag.SourceVirtualNode, holdingIdentity.shortHash.toString())
                     .build()
                     .record(Duration.ofNanos(System.nanoTime() - commitStartTime))
-            }
-        }
-    }
-
-    @VisibleForTesting
-    fun eventHandler(event: LifecycleEvent, coordinator: LifecycleCoordinator) {
-        log.info("Backing store received event $event")
-        when (event) {
-            is StartEvent -> {
-                dependentComponents.registerAndStartAll(coordinator)
-            }
-            is StopEvent -> {
-                dependentComponents.stopAll()
-            }
-            is RegistrationStatusChangeEvent -> {
-                jpaEntitiesRegistry.register(
-                    CordaDb.Uniqueness.persistenceUnitName,
-                    JPABackingStoreEntities.classes
-                )
-
-                jpaEntities = jpaEntitiesRegistry.get(CordaDb.Uniqueness.persistenceUnitName)
-                    ?: throw IllegalStateException(
-                        "persistenceUnitName " +
-                                "${CordaDb.Uniqueness.persistenceUnitName} is not registered."
-                    )
-
-                log.info("Backing store is ${event.status}")
-                coordinator.updateStatus(event.status)
-            }
-            else -> {
-                log.warn("Unexpected event ${event}, ignoring")
             }
         }
     }
--- a/components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreLifecycleImpl.kt
+++ b/components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreLifecycleImpl.kt
@@ -1,0 +1,90 @@
+package net.corda.uniqueness.backingstore.impl
+
+import net.corda.db.connection.manager.DbConnectionManager
+import net.corda.db.schema.CordaDb
+import net.corda.lifecycle.DependentComponents
+import net.corda.lifecycle.LifecycleCoordinator
+import net.corda.lifecycle.LifecycleCoordinatorFactory
+import net.corda.lifecycle.LifecycleEvent
+import net.corda.lifecycle.RegistrationStatusChangeEvent
+import net.corda.lifecycle.StartEvent
+import net.corda.lifecycle.StopEvent
+import net.corda.lifecycle.createCoordinator
+import net.corda.orm.JpaEntitiesRegistry
+import net.corda.orm.JpaEntitiesSet
+import net.corda.uniqueness.backingstore.BackingStore
+import net.corda.uniqueness.backingstore.BackingStoreLifecycle
+import net.corda.utilities.VisibleForTesting
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.Reference
+import org.slf4j.LoggerFactory
+
+@Suppress("unused")
+/**
+ * JPA backing store implementation, which uses a JPA compliant database to persist data.
+ */
+@Component(service = [BackingStoreLifecycle::class])
+open class JPABackingStoreLifecycleImpl @Activate constructor(
+    @Reference(service = LifecycleCoordinatorFactory::class)
+    coordinatorFactory: LifecycleCoordinatorFactory,
+    @Reference(service = JpaEntitiesRegistry::class)
+    private val jpaEntitiesRegistry: JpaEntitiesRegistry,
+    @Reference(service = DbConnectionManager::class)
+    private val dbConnectionManager: DbConnectionManager,
+    @Reference(service = BackingStore::class)
+    private val backingStore: BackingStore
+) : BackingStoreLifecycle, BackingStore by backingStore {
+
+    private companion object {
+        private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
+    }
+
+    private val lifecycleCoordinator: LifecycleCoordinator = coordinatorFactory
+        .createCoordinator<BackingStoreLifecycle>(::eventHandler)
+
+    private val dependentComponents = DependentComponents.of(
+        ::dbConnectionManager
+    )
+
+    private lateinit var jpaEntities: JpaEntitiesSet
+
+    override val isRunning: Boolean
+        get() = lifecycleCoordinator.isRunning
+
+    override fun start() {
+        log.info("Backing store starting")
+        lifecycleCoordinator.start()
+    }
+
+    override fun stop() {
+        log.info("Backing store stopping")
+        lifecycleCoordinator.stop()
+    }
+
+    @VisibleForTesting
+    fun eventHandler(event: LifecycleEvent, coordinator: LifecycleCoordinator) {
+        log.info("Backing store received event $event")
+        when (event) {
+            is StartEvent -> {
+                dependentComponents.registerAndStartAll(coordinator)
+            }
+            is StopEvent -> {
+                dependentComponents.stopAll()
+            }
+            is RegistrationStatusChangeEvent -> {
+                jpaEntities = jpaEntitiesRegistry.get(CordaDb.Uniqueness.persistenceUnitName)
+                    ?: throw IllegalStateException(
+                        "persistenceUnitName " +
+                                "${CordaDb.Uniqueness.persistenceUnitName} is not registered."
+                    )
+
+                log.info("Backing store is ${event.status}")
+                coordinator.updateStatus(event.status)
+            }
+            else -> {
+                log.warn("Unexpected event ${event}, ignoring")
+            }
+        }
+    }
+}
--- a/components/uniqueness/uniqueness-checker-client-service-impl/build.gradle
+++ b/components/uniqueness/uniqueness-checker-client-service-impl/build.gradle
@@ -10,7 +10,7 @@ dependencies {
     compileOnly 'org.osgi:osgi.core'
     compileOnly 'org.osgi:org.osgi.service.component.annotations'
 
-    implementation platform('net.corda:corda-api:$cordaApiVersion')
+    implementation platform("net.corda:corda-api:$cordaApiVersion")
 
     implementation 'net.corda:corda-application'
     implementation 'net.corda:corda-base'
--- a/components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/BatchedUniquenessCheckerImpl.kt
+++ b/components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/BatchedUniquenessCheckerImpl.kt
@@ -1,29 +1,11 @@
 package net.corda.uniqueness.checker.impl
 
-import net.corda.configuration.read.ConfigChangedEvent
-import net.corda.configuration.read.ConfigurationReadService
 import net.corda.data.ExceptionEnvelope
 import net.corda.data.uniqueness.UniquenessCheckRequestAvro
 import net.corda.data.uniqueness.UniquenessCheckResponseAvro
 import net.corda.data.uniqueness.UniquenessCheckResultMalformedRequestAvro
 import net.corda.data.uniqueness.UniquenessCheckResultUnhandledExceptionAvro
-import net.corda.flow.external.events.responses.factory.ExternalEventResponseFactory
-import net.corda.libs.configuration.SmartConfig
-import net.corda.libs.configuration.helper.getConfig
-import net.corda.lifecycle.DependentComponents
-import net.corda.lifecycle.LifecycleCoordinator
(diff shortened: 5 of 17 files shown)


relyafi (commented):
See https://r3holdco.slack.com/archives/C03A3TQMBHR/p1688740735041139 for discussion on this topic. A proper description explaining the rationale for this change should be added to this PR.

I haven't spent the time to properly review yet (which I think does need review from ledger / notaries before merge), but I am a bit concerned that it appears we are changing the design pattern to facilitate a test framework change. I think this needs an architectural review from @driessamyn, to answer two questions:

- Are we happy with the principle of splitting out the lifecycle into its own class?
- If so, shouldn't this pattern be adopted across the board? This change would make the uniqueness checker / backing store outliers compared to the rest of the codebase, which has their components implement `Lifecycle` directly. This may hamper understanding / maintenance of the code going forward.

chrisr3:
> I am a bit concerned that it appears we are changing the design pattern to facilitate a test framework change.

The "uniqueness checker" components had strongly coupled the "lifecycle" layer to the "messaging" layer, **unlike** the ledger, entity and verification components, and this PR simply disentangles them. I therefore consider the "uniqueness checker" components to be the outliers here, since I believe the original plan _was_ to keep the layers separate.

I will also point out that separating the "lifecycle" code has removed a lot of unused mocks from the unit tests; a clear indication that this change is a Good Thing :tm:.

yarivus (commented):
Looking at the code reductions due to the change seems like we now have much less dependencies across components ... so IMHO the change reflects @chrisr3 's PR description and overall a good change

chrisr3:
Note that we already use this technique to separate the lifecycle from the sandboxing components:
```kotlin
interface SandboxGroupContextService: AutoCloseable {
    // etc
}

interface SandboxGroupContextComponent : SandboxGroupContextService, CacheControl, Lifecycle
```

components/uniqueness/uniqueness-checker-client-service-impl/build.gradle
```diff
@@ -10,7 +10,7 @@ dependencies {
     compileOnly 'org.osgi:osgi.core'
     compileOnly 'org.osgi:org.osgi.service.component.annotations'
 
-    implementation platform('net.corda:corda-api:$cordaApiVersion')
+    implementation platform("net.corda:corda-api:$cordaApiVersion")
```
driessamyn, line 13: Not really related to the PR, but how did the before state even work. I thought single quotes means literal string in groovy so it wouldn't resolve `$cordaApiVersion` ?

driessamyn (approved):
LGTM

driessamyn:
> I haven't spent the time to properly review yet (which I think does need review from ledger / notaries before merge), but I am a bit concerned that it appears we are changing the design pattern to facilitate a test framework change. I think this needs an architectural review from @driessamyn, to answer two questions:
> 
>     * Are we happy with the principle of splitting out the lifecycle into its own class?
> 
>     * If so, shouldn't this pattern be adopted across the board? This change would make the uniqueness checker / backing store outliers compared to the rest of the codebase, which has their components implement `Lifecycle` directly. This may hamper understanding / maintenance of the code going forward.

The usage pattern of the lifecycle library is inconsistent across the board. I think the pattern here is used elsewhere, just not consistently.
This is something we need to resolve in the longer term, and we need an owner for this library first.

I've approved the PR as I think it's the pragmatic thing to do. @relyafi , are we missing a code owners mapping if the notary team hasn't been listed as a mandatory approver?

components/uniqueness/uniqueness-checker-client-service-impl/build.gradle
chrisr3: You are correct, and I _do not know_ how it worked before. It did seem prudent to **fix** it, though :wink:.

(Perhaps it was inheriting the correct `platform` dependency transitively from another module?)

seanbrereton (approved):
LGTM - build only

relyafi:
> > I haven't spent the time to properly review yet (which I think does need review from ledger / notaries before merge), but I am a bit concerned that it appears we are changing the design pattern to facilitate a test framework change. I think this needs an architectural review from @driessamyn, to answer two questions:
> > ```
> > * Are we happy with the principle of splitting out the lifecycle into its own class?
> > 
> > * If so, shouldn't this pattern be adopted across the board? This change would make the uniqueness checker / backing store outliers compared to the rest of the codebase, which has their components implement `Lifecycle` directly. This may hamper understanding / maintenance of the code going forward.
> > ```
> 
> The usage pattern of the lifecycle library is inconsistent across the board. I think the pattern here is used elsewhere, just not consistently. This is something we need to resolve in the longer term, and we need an owner for this library first.

Thanks @driessamyn, in that case happy to go ahead with the approach and will do a detailed review.

> I've approved the PR as I think it's the pragmatic thing to do. @relyafi , are we missing a code owners mapping if the notary team hasn't been listed as a mandatory approver?

I thought we had this, but this may have just been the limited list we had for further 5.0 changes. I'll look into fixing this properly in 5.1

components/uniqueness/backing-store-impl/src/integrationTest/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImplIntegrationTests.kt
```diff
@@ -84,7 +84,7 @@ import kotlin.reflect.full.createInstance
  */
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class JPABackingStoreImplIntegrationTests {
```
relyafi, line 86: I'm torn as to whether we need to rename all test classes, variable usages etc. to include "lifecycle" now as well, since these are no longer referencing the actual impl class.

That said, I question whether this class should be testing the lifecycle class at all - I don't think it's used much by these test cases (if at all), and the references are pretty much just doing the setup that was necessary to satisfy the lifecycle logic. It may be possible to simplify this to just test the backing store directly.

components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImpl.kt
```diff
@@ -70,17 +58,12 @@ open class JPABackingStoreImpl @Activate constructor(
...
-
-    private lateinit var jpaEntities: JpaEntitiesSet
-
-    override val isRunning: Boolean
-        get() = lifecycleCoordinator.isRunning
+    init {
+        jpaEntitiesRegistry.register(
+            CordaDb.Uniqueness.persistenceUnitName,
+            JPABackingStoreEntities.classes
+        )
```
relyafi, line 65: Is this safe? If memory serves me correctly, don't we need the database to be available prior to registering the JPA entities?

This will now cause this to be done on construction, instead of when we get a registration status change event.

components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreLifecycleImpl.kt
```diff
@@ -0,0 +1,90 @@
...
+import net.corda.uniqueness.backingstore.BackingStoreLifecycle
+import net.corda.utilities.VisibleForTesting
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.Reference
+import org.slf4j.LoggerFactory
+
+@Suppress("unused")
+/**
+ * JPA backing store implementation, which uses a JPA compliant database to persist data.
```
relyafi, line 25: This comment is copy / pasted from the existing class, and should be updated to reflect the split in functionality.

components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreLifecycleImpl.kt
```diff
@@ -0,0 +1,90 @@
...
+ */
+@Component(service = [BackingStoreLifecycle::class])
+open class JPABackingStoreLifecycleImpl @Activate constructor(
+    @Reference(service = LifecycleCoordinatorFactory::class)
+    coordinatorFactory: LifecycleCoordinatorFactory,
+    @Reference(service = JpaEntitiesRegistry::class)
+    private val jpaEntitiesRegistry: JpaEntitiesRegistry,
+    @Reference(service = DbConnectionManager::class)
+    private val dbConnectionManager: DbConnectionManager,
+    @Reference(service = BackingStore::class)
```
relyafi, line 35: Do these need to be included here, given they are already defined within the backing store?

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/BatchedUniquenessCheckerImpl.kt
```diff
@@ -63,45 +44,23 @@ import java.util.LinkedList
...
     private companion object {
-        const val GROUP_NAME = "uniqueness.checker"
-        const val UNHANDLED_EXCEPTION = "UniquenessCheckResultUnhandledException"
+        private const val UNHANDLED_EXCEPTION = "UniquenessCheckResultUnhandledException"
 
-        const val CONFIG_HANDLE = "CONFIG_HANDLE"
-        const val SUBSCRIPTION = "SUBSCRIPTION"
-
-        val log: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+        private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
```
relyafi, line 63: These look like unnecessary changes; why make these private when they are already in a private companion object?

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/BatchedUniquenessCheckerLifecycleImpl.kt
```diff
@@ -0,0 +1,129 @@
...
+import net.corda.uniqueness.checker.UniquenessChecker
+import net.corda.uniqueness.checker.UniquenessCheckerLifecycle
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.Reference
+import org.slf4j.LoggerFactory
+
+/**
+ * A batched implementation of the uniqueness checker component, which processes batches of requests
+ * together in order to provide higher performance under load.
```
relyafi, line 31: Copy / paste comment that needs updating to reflect the separation of responsibilities.

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/BatchedUniquenessCheckerLifecycleImpl.kt
```diff
@@ -0,0 +1,129 @@
...
+    @Reference(service = UniquenessChecker::class)
+    uniquenessChecker: UniquenessChecker
+) : UniquenessCheckerLifecycle, UniquenessChecker by uniquenessChecker {
+    private companion object {
+        const val GROUP_NAME = "uniqueness.checker"
+
+        const val CONFIG_HANDLE = "CONFIG_HANDLE"
+        const val SUBSCRIPTION = "SUBSCRIPTION"
+
+        private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
```
relyafi, line 55: Again, why a private within a private companion object?

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/BatchedUniquenessCheckerLifecycleImpl.kt
```diff
@@ -0,0 +1,129 @@
...
+    @Reference(service = LifecycleCoordinatorFactory::class)
+    coordinatorFactory: LifecycleCoordinatorFactory,
+    @Reference(service = ConfigurationReadService::class)
+    private val configurationReadService: ConfigurationReadService,
+    @Reference(service = SubscriptionFactory::class)
+    private val subscriptionFactory: SubscriptionFactory,
+    @Reference(service = ExternalEventResponseFactory::class)
+    private val externalEventResponseFactory: ExternalEventResponseFactory,
+    @Reference(service = BackingStoreLifecycle::class)
+    private val backingStore: BackingStoreLifecycle,
```
relyafi, line 45: Consider renaming to reflect it's the lifecycle, not the actual impl

testing/uniqueness/backing-store-fake/src/main/kotlin/net/corda/uniqueness/backingstore/impl/fake/BackingStoreImplFake.kt
```diff
@@ -31,19 +32,19 @@ import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
 @ServiceRanking(Int.MAX_VALUE)
-@Component(service = [BackingStore::class])
+@Component(service = [BackingStoreLifecycle::class, BackingStore::class])
```
relyafi, line 35: Given this class is purely for unit testing, I think this highlights that perhaps the unit tests should be testing the backing store directly, and this change should simply remove lifecycle logic from the fake? The `coordinatorFactory` is only mocked where used in the tests.

relyafi (changes requested):
Some changes suggested - unfortunately this change has already been merged prematurely (which does show the need for the notary code owners list to be put in place), so @chrisr3 you will need to figure out the best way of addressing these.

I do question whether inheritance (as opposed to composition) is the best way of handling this, especially given the delegation to the backing store within the lifecycle impl anyway. But what has been done is the more expedient change.

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/BatchedUniquenessCheckerLifecycleImpl.kt
chrisr3: Because Kotlin would generate an unnecessary `getLog()` getter function for a public property.

testing/uniqueness/backing-store-fake/src/main/kotlin/net/corda/uniqueness/backingstore/impl/fake/BackingStoreImplFake.kt
chrisr3: This is an OSGi component, and is therefore used for integration testing rather than unit testing. I have added `BackingStoreLifecycle` as a service so that `BackingStoreImplFake` can satisfy _both_ `BackingStore` _and_ `BackingStoreLifecycle` references.

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/BatchedUniquenessCheckerImpl.kt
chrisr3: This prevents Kotlin from generating an unnecessary `fun getLog(): Logger` getter method. Kotlin will only generate a field for `private val`.

components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreLifecycleImpl.kt
chrisr3: Yes, because `JPABackingStoreLifecycleImpl` still references them. However, they are both singletons and so the OSGi framework will inject the same instances into both `JPABackingStoreImpl` and JPABackingStoreLifecycleImpl`.
.

relyafi: Yes, but what I mean is, these are already defined as class members of `JPABackingStore`, so could we not use those references via the backing store instead? That said, it would mean changing the visibility from private, so perhaps that's not advisable.

components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImpl.kt
chrisr3: `JPAEntitiesRegistryImpl` is just a wrapper around a `ConcurrentHashMap`, and `JPABackingStoreEntities.classes` is defined as:
```kotlin
internal object JPABackingStoreEntities {
    val classes = setOf(
        UniquenessStateDetailEntity::class.java,
        UniquenessTransactionDetailEntity::class.java,
        UniquenessRejectedTransactionEntity::class.java
    )
}
```
There's no database usage here at all, so why _wouldn't_ it be safe? :thinking:

components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreLifecycleImpl.kt
chrisr3: I would need to change the definition of the `BackingStore` interface in order to expose the `dbConnectionManager` component. I didn't think that was advisable.

components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImpl.kt
relyafi: It could be that referencing the entities would trigger validation of the said entities against the DB schema, which would obviously fail if it was unavailable.

I'm not saying there is an issue, but I certainly think it needs due diligence to confirm. Might be as simple as killing / restarting the DB.

components/uniqueness/backing-store-impl/src/integrationTest/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImplIntegrationTests.kt
chrisr3: Actually, this test _does_ still reference the actual impl class:
```kotlin
    private fun createBackingStoreImpl(emFactory: EntityManagerFactory): JPABackingStoreLifecycleImpl {
        val jpaEntitiesRegistry = JpaEntitiesRegistryImpl()
        val dbConnectionManager = mock<DbConnectionManager>().apply {
            whenever(getOrCreateEntityManagerFactory(any(), any(), any())) doReturn emFactory
            whenever(getClusterDataSource()) doReturn dbConfig.dataSource
        }
        return JPABackingStoreLifecycleImpl(
            lifecycleCoordinatorFactory,
            jpaEntitiesRegistry,
            dbConnectionManager,
            JPABackingStoreImpl(jpaEntitiesRegistry, dbConnectionManager)
        )
    }
```

components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreLifecycleImpl.kt
relyafi: Addressed by #4257

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/BatchedUniquenessCheckerLifecycleImpl.kt
relyafi: Addressed by #4257

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/BatchedUniquenessCheckerLifecycleImpl.kt
relyafi: Addressed by #4257

components/uniqueness/backing-store-impl/src/integrationTest/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImplIntegrationTests.kt
relyafi: That's only because of the refactor that's been done where we've done a simple repoint to the lifecycle class rather than asking if we need the lifecycle at all - it was instantiating a backing store before. What I'm saying is, could this continue to instantiate a backing store, and take the lifecycle logic out of the equation?

components/uniqueness/backing-store-impl/src/main/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImpl.kt
relyafi: Actually, it looks like the register call will ultimately end up down here: https://github.com/corda/corda-runtime-os/blob/e7cbe3e1973ecab885b1158a661f5eeac94ccff5/libs/db/db-orm/src/main/kotlin/net/corda/orm/JpaEntitiesSet.kt

Looks like there is no DB schema validation logic at this point, so you're right, we should be ok. It must be the later calls (still wrapped in the event handler) that will end up doing that.

chrisr3: `JPABackingStoreLifecycleImpl` _depends_ upon `JPABackingStoreImpl`, and so OSGi _must_ instantiate the `JPABackingStoreImpl` component first. Hence the entities _must_ be available by the time `JPABackingStoreLifecycleImpl` executes:
```kotlin
    jpaEntities = jpaEntitiesRegistry.get(CordaDb.Uniqueness.persistenceUnitName)
```

components/uniqueness/backing-store-impl/src/integrationTest/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImplIntegrationTests.kt
chrisr3: My goal here was to ensure that the tests still worked correctly afterwards. I did not intend to rewrite them completely, c.f. `MessageBusIntegrationTests` which I _really_ didn't want to dig too deeply into. Having said that, I _was_ able to remove the "lifecycle" references from the `UniquenessChecker` tests.

chrisr3:
> I do question whether inheritance (as opposed to composition) is the best way of handling this, especially given the delegation to the backing store within the lifecycle impl anyway. But what has been done is the more expedient change.

I don't understand - this _does_ use composition: `JPABackingStoreLifecycleImpl` composes `JPABackingStoreImpl`, and `BatchedUniquenessCheckerLifecycleImpl` composes `BackedUniquenessCheckerImpl` :thinking:.

relyafi:
> > I do question whether inheritance (as opposed to composition) is the best way of handling this, especially given the delegation to the backing store within the lifecycle impl anyway. But what has been done is the more expedient change.
> 
> I don't understand - this _does_ use composition: `JPABackingStoreLifecycleImpl` composes `JPABackingStoreImpl`, and `BatchedUniquenessCheckerLifecycleImpl` composes `BackedUniquenessCheckerImpl` 🤔.

Yes and no - that's what I meant when I said we were delegating under the hood we can argue the implementation is using composition, but the interface is inheritance:

`interface BackingStoreLifecycle : BackingStore, Lifecycle`

That implies an "is a" rather than "has" relationship. I don't think that's quite right, but also I can understand why it's been done the way it has for expediency.

components/uniqueness/backing-store-impl/src/integrationTest/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImplIntegrationTests.kt
relyafi: I think if we start to go making changes to accommodate test frameworks in entirely unrelated bits of functionality, there is an obligation to make sure we are doing what's right for the product, not to just do what's quickest and easiest in the context of the current changes. I don't think it's right to do part of the job, and then leave someone else to fix it up properly down the line.

FWIW I suspect it isn't a rewrite required in order to change this - if that was the case it would be a strong argument against this comment as it would imply there is still a tight coupling there.

chrisr3: > I think if we start to go making changes to accommodate test frameworks in entirely unrelated bits of functionality, ...

Do you _want_ to be able to single-step debug the notary flow code from within IntelliJ? :exploding_head:.

I will repeat: Keeping the "lifecycle" and "message" layers separate is not a "New Thing", even if the notary code apparently never received the memo.

> I don't think it's right to do part of the job, and then leave someone else to fix it up properly down the line.

Whoever wrote `MessageBusIntegrationTests` apparently considered it to be "useful", and who am I to contradict them?

> leave someone else to fix it up properly

I don't see how `MessageBusIntegrationTests` is any less "proper" than it was before, seeing as its code-coverage is unchanged. You will observe that the "lifecycle" code is tested _explicitly_, as otherwise I would not have needed to change this:
```diff
                c.followStatusChangesByName(
                    setOf(
-                       LifecycleCoordinatorName.forComponent<BackingStore>(),
+                       LifecycleCoordinatorName.forComponent<BackingStoreLifecycle>(),
                        LifecycleCoordinatorName.forComponent<ConfigurationReadService>(),
-                       LifecycleCoordinatorName.forComponent<UniquenessChecker>()
+                       LifecycleCoordinatorName.forComponent<UniquenessCheckerLifecycle>()
                    )
                )
```

chrisr3:
> That implies an "is a" rather than "has" relationship. I don't think that's quite right, but also I can understand why it's been done the way it has for expediency.

Personally speaking, I consider this to be a perfectly normal and acceptable use of Java interfaces rather than "expediency", and an example of _why_  Java has extensible interfaces in the first place. As an exercise, you can browse the code-base to see how many Corda interfaces already extends `Lifecycle`:
- `AllowedCertificatesReaderWriterService`
-  ... _scroll_, _scroll_ ...
- `VirtualNodeWriteService`

components/uniqueness/backing-store-impl/src/integrationTest/kotlin/net/corda/uniqueness/backingstore/impl/JPABackingStoreImplIntegrationTests.kt
relyafi: > Do you want to be able to single-step debug the notary flow code from within IntelliJ? 🤯.

This isn't an excuse to cut corners, the same argument could be made for any software change being made.

The message bus tests are a bit more involved as they are OSGi tests, but yes, it may be the comment applies there too.

At the least, if you don't want to do the work now, I'd suggest adding comments to these test files to suggest reviewing whether the lifecycle classes need to be included. That way someone unfamiliar with the background to this code won't scratch their head wondering what value they are adding to these test cases.

chrisr3: > This isn't an excuse to cut corners, the same argument could be made for any software change being made.

Who said anything about "cutting corners"? I simply don't understand why you are repeatedly griping about "making changes to accommodate test frameworks" when that test framework is actually a Good Thing :tm:.

```kotlin
uniquenessCheckerLifecycle = BatchedUniquenessCheckerLifecycleImpl(
    lifecycleCoordinatorFactory,
    configurationReadService,
    subscriptionFactory,
    externalEventResponseFactory,
    backingStore,
    BatchedUniquenessCheckerImpl(backingStore, testClock)
).also { it.start() }
```
So `MessageBusIntegrationTests` is using both `BatchedUniquenessCheckerImpl()` and `BatchedUniquenessCheckerLifecycleImpl`, which means its value as an integration test in undiminished. I am also assuming that the "lifecycle" checking here is _intentional_, seeing as the test class name includes the words "MessageBus".

Having said that, I _can_ still strip all of the "lifecycle" code  out of `MessageBusIntegrationTests`, if the remainder would still be considered useful.

See #4257.

relyafi: Thanks for applying that. Looking at the change, I can see there is some value in retaining the lifecycle logic in `MessageBusIntegrationTests`, because it is additionally testing the external events API. As mentioned earlier, this is the more involved set of tests because it's using OSGi. So I'm happy for you to leave the lifecycle code in this case.

We seem to have completely overlooked the code this was originally made against, namely `JPABackingStoreImplIntegrationTests`. These are not OSGi based, and are using mocks. So this may be a stronger candidate for removal of the lifecycle logic.

Note this is not a demand to remove the logic, I'm simply asking for the legwork to be done to establish if keeping the lifecycle logic is justified or not. We can then make an informed decision as to whether to leave as-is (and potentially rename the test file) or make the changes to remove.

yarivus (commented):
thx @chrisr3

PR #4258, opened 2023-07-10
CORE-8958 distribute group parameters in backchain resolution inherit

vlajos:


vlajos:
This direction got discarded.

PR #4263, opened 2023-07-10
CORE-14685: Add support for UTXO flows.

chrisr3:
Add support for running UTXO flows within the driver framework. This also includes support for a non-validating notary and the verification processor.

Notary support is _not_ enabled by default, and must be explicitly requested when configuring the driver nodes:
```kotlin
@RegisterExtension
private val driver = DriverNodes(alice, bob).withNotary(sally, 1, 2).forAllTests()
```
where `sally` is the notary's `MemberX500Name`, for protocol versions 1 and 2.

This is to accommodate tests which want to define the `MemberX500Name` and protocol versions of the notary (or notaries!) themselves. And also to avoid creating notaries unnecessarily.

I have created `NotaryFlowTests` and `UtxoLedgerTests` for the driver, based on our existing smoke tests, which all execute correctly except for one UTXO tests that uses "custom queries". The driver does not yet include support for "custom queries", since Corda has not yet implemented them for HSQLDB.

Quasar's "file cache" is now enabled and created as `build/quasar-cache`. This cache can be shared by all driver tests belonging to the same Gradle module, and should be safe for concurrent access (i.e. multiple driver tests running in parallel.)

---
Corda's Crypto code is still being rewritten, after which I am hoping that `DriverCryptoServiceProviderImpl` can be migrated from `crypto-softhsm-impl` to `driver-engine`.


seanbrereton (approved):
LGTM - Build

testing/driver/driver-engine/src/main/kotlin/net/corda/testing/driver/config/SmartConfigProvider.kt
```diff
@@ -34,7 +34,7 @@ class SmartConfigProvider {
             .withValue(PROCESSING_MAX_RETRY_ATTEMPTS, ConfigValueFactory.fromAnyRef(5))
             .withValue(PROCESSING_MAX_RETRY_DELAY, ConfigValueFactory.fromAnyRef(5000L))
             .withValue(PROCESSING_MAX_FLOW_SLEEP_DURATION, ConfigValueFactory.fromAnyRef(5000L))
-            .withValue(PROCESSOR_TIMEOUT, ConfigValueFactory.fromAnyRef(60000L))
+            .withValue(PROCESSOR_TIMEOUT, ConfigValueFactory.fromAnyRef(600000L))
```
yarivus, line 37: was this intentional increasing x10 ?

yarivus (commented):
see 1 question

testing/driver/driver-engine/src/main/kotlin/net/corda/testing/driver/config/SmartConfigProvider.kt
chrisr3: Yes, because I discovered that my debugging sessions were unexpectedly timing out! :scream:

Ultimately, this is one configuration option that we are likely to want to expose to users.

PR #4265, opened 2023-07-10
CORE-8958 distribute group parameters in backchain resolution shared class

vlajos:
* Distribute signed group parameters via back-chain resolution when the receiver side does not have it already.
* Create a v2 version of Backchain resolution flows to support the new protocol.
* The same set of classes implements both versions. But they are instantiated with a different version parameter.
* Initial signed group parameters must also be saved on the responder side since that happens before the back-chain resolution. Also, this can happen without a network hop. However, to guarantee that we saved the one used in the received transaction, we need to move the check, which compares the current group parameters hash with the one in the metadata of the transaction before the back-chain resolution, and also we need to guarantee that the same one gets saved.


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/base/TransactionBackchainReceiverFlowBase.kt
```diff
@@ -53,7 +67,7 @@ class TransactionBackchainReceiverFlowV1(
             @Suppress("unchecked_cast")
             val retrievedTransactions = session.sendAndReceive(
                 List::class.java,
-                TransactionBackchainRequestV1.Get(batch)
+                TransactionBackchainRequestBase.Get(batch)
```
lankydan: This doesn't work because a 5.0 node wouldn't know this type.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/base/TransactionBackchainReceiverFlowBase.kt
```diff
@@ -1,26 +1,34 @@
...
 import net.corda.v5.application.messaging.FlowSession
 import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.base.exceptions.CordaRuntimeException
 import net.corda.v5.crypto.SecureHash
 import net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction
 import org.slf4j.LoggerFactory
 
 @CordaSystemFlow
-class TransactionBackchainReceiverFlowV1(
+class TransactionBackchainReceiverFlowBase(
```
blsemo: I don't like the "base" in the classname here - it is not a base class, after all.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/base/TransactionBackchainReceiverFlowBase.kt
lankydan: If we wanted to switch to this way of coding I think we'd at least need to switch to a V2 version first.

lankydan: Or we would need do some changes to 5.0 instead.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -83,6 +100,23 @@ class UtxoReceiveFinalityFlowV1(
...
+                "Transactions can be created only with the latest membership group parameters. " +
+                        "Current: ${currentGroupParameters.hash} Transaction's: $txGroupParametersHash"
+            log.warn(message)
+            persistInvalidTransaction(initialTransaction)
+            throw CordaRuntimeException(message)
+        }
+        signedGroupParametersVerifier.verifySignature(currentGroupParameters)
+        return currentGroupParameters
+    }
+
```
blsemo, line 119: Isn't that part of a different change/PR?

blsemo (commented):
Looks good to me, just not happy with the naming of the classes  `...Base` sounds too much like an abstract base class.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/base/TransactionBackchainReceiverFlowBase.kt
vlajos: Nice catch. Thanks.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/base/TransactionBackchainReceiverFlowBase.kt
vlajos: It was V1. And it seems that the request class should keep its original name with this approach. (@lankydan 's comment.)
I guess I can remove `Base` from the ones not sent across the wire.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
vlajos: The original version was in a different PR. But a big chunk of it needed to be moved to other places. (I wrote a bit about it in the last section of the PR's description.)

blsemo: Ok, reread it. It makes sense now.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/base/TransactionBackchainReceiverFlowBase.kt
vlajos: I replaced it with `main`.
It is not much better, but removing it would make it more confusing. (The caller/versioning flows have the same names already.)

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/base/TransactionBackchainReceiverMainFlow.kt
```diff
@@ -1,26 +1,35 @@
-package net.corda.ledger.utxo.flow.impl.flows.backchain.v1
+package net.corda.ledger.utxo.flow.impl.flows.backchain.base
```
blsemo: I am not sure that changing the package name to `base` has any benefit. Maybe we could leave it at `v1` and just put a comment somewhere explaining that v2 is just an extension of v1?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/base/TransactionBackchainReceiverMainFlow.kt
```diff
@@ -101,6 +117,46 @@ class TransactionBackchainReceiverFlowV1(
...
+            return
+        }
+        val retrievedTransactionId = retrievedTransaction.id
+        val groupParametersHash = parseSecureHash(requireNotNull(
+            (retrievedTransaction.metadata as TransactionMetadataInternal).getMembershipGroupParametersHash()
+        ) {
+            "Received transaction does not have group parameters in its metadata."
+        })
+
+        if (utxoLedgerGroupParametersPersistenceService.find(groupParametersHash) != null) {
```
blsemo: I hope this is cached somewhere? Because I can see that being called very often with the exact same hash...

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/base/TransactionBackchainReceiverFlowBase.kt
vlajos: I have reverted this rename.

blsemo (commented):
Can't we keep the name as v1 or make it v2 (and add a note that they have a switch so can v1 and v2 can unterstand each other)?
I don't like completely breaking out of the versioning naming scheme.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/base/TransactionBackchainReceiverMainFlow.kt
vlajos: Yes. `GroupParametersCache` is responsible for that.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/base/TransactionBackchainReceiverMainFlow.kt
vlajos: I have renamed that to `main` already. Would that be better?
Or would you still prefer to keep it as v1?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainSenderFlowV1.kt
```diff
@@ -12,10 +14,23 @@ import net.corda.v5.crypto.SecureHash
...
+ * V2 is essentially an extension of V1, so in order to avoid huge code duplication,
+ * we kept V1 class implementing both.
+ */
+
 @CordaSystemFlow
-class TransactionBackchainSenderFlowV1(private val headTransactionIds: Set<SecureHash>, private val session: FlowSession) : SubFlow<Unit> {
+class TransactionBackchainSenderFlowV1(
+    private val headTransactionIds: Set<SecureHash>,
+    private val session: FlowSession,
+    val version: TransactionBackChainResolutionVersion
```
lankydan, line 27: Should this version be private?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainSenderFlowV1.kt
```diff
@@ -53,10 +71,35 @@ class TransactionBackchainSenderFlowV1(private val headTransactionIds: Set<Secur
...
 
+    @Suspendable
+    private fun handleSignedGroupParametersRequest(request: TransactionBackchainRequestV1.GetSignedGroupParameters) {
+        if (version == TransactionBackChainResolutionVersion.V1) {
+            val message =
+                "Backchain resolution of $headTransactionIds - GetSignedGroupParameters is " +
+                        "not available in TransactionBackchainSenderFlowV1 V1"
+            log.warn(message)
+            throw CordaRuntimeException(message)
+        }
```
lankydan: Worth noting that I don't think its possible to hit this branch if this flow is V1 and the peer is V2. It would fail deserialization instead.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainSenderFlowV1.kt
```diff
@@ -53,10 +71,35 @@ class TransactionBackchainSenderFlowV1(private val headTransactionIds: Set<Secur
                     }
                     return
                 }
+                is TransactionBackchainRequestV1.GetSignedGroupParameters ->
```
lankydan, line 75: Not a big fan of the naming but I understand why it ends up becoming like this.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -1,26 +1,39 @@
...
 import net.corda.v5.application.messaging.FlowSession
 import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.base.exceptions.CordaRuntimeException
 import net.corda.v5.crypto.SecureHash
 import net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction
 import org.slf4j.LoggerFactory
 
+/**
+ * V2 is essentially an extension of V1, so in order to avoid huge code duplication,
+ * we kept V1 class implementing both.
```
blsemo:
```suggestion
 * The V2 protocoll is an extension of the V1 protocoll, which can be enabled via a switch (on both sides). In order to avoid huge code duplication,
 * we kept V1 class implementing both protocolls and added a switch that makes it behave according to the V2 protocoll.
```
I would suggest this for all instances.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainRequestV1.kt
```diff
@@ -3,8 +3,14 @@ package net.corda.ledger.utxo.flow.impl.flows.backchain.v1
 import net.corda.v5.base.annotations.CordaSerializable
 import net.corda.v5.crypto.SecureHash
 
+/**
+ * This class is transferred both in V1 and V2 of Transaction Backchain resolution flows.
+ * So its name/package should not be changed (without flow version upgrade).
+ */
+
```
blsemo: I think this comment is obsolete now we kept the V1 class anyway.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainResolutionFlowV1.kt
```diff
@@ -16,10 +17,16 @@ import net.corda.v5.crypto.SecureHash
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
+/**
+ * V2 is essentially an extension of V1, so in order to avoid huge code duplication,
+ * we kept V1 class implementing both.
```
blsemo:
```suggestion
 * The V2 protocoll is an extension of the V1 protocoll, which can be enabled via a switch (on both sides). In order to avoid huge code duplication,
 * we kept V1 class implementing both protocolls and added a switch that makes it behave according to the V2 protocoll.
```

lankydan (dismissed):
Generally, I'm fine with the change but honestly I'd rather take the duplication of code because its a clean cut of what V1 does and what V2 does.

Part of me feels that if we want to have the code shared better, we'd be better off creating V2 now and then making an equivalent of `TransactionBackchainRequestV1` which is an interface and named `TransactionBackchainRequest`. We can then add new types under that as needed to handle new functionality we add in the future. 

In hindsight I'd prefer that I didn't make it a sealed class, but I went on the assumption we'd just duplicate everything.

I'm ok with approving the PR to move things forward, but wanted to put out my thoughts before doing so.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainSenderFlowV1.kt
```diff
@@ -12,10 +14,23 @@ import net.corda.v5.crypto.SecureHash
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
+/**
+ * V2 is essentially an extension of V1, so in order to avoid huge code duplication,
+ * we kept V1 class implementing both.
```
blsemo:
```suggestion
 * The V2 protocoll is an extension of the V1 protocoll, which can be enabled via a switch (on both sides). In order to avoid huge code duplication,
 * we kept V1 class implementing both protocolls and added a switch that makes it behave according to the V2 protocoll.
```

blsemo (commented):
Looks good now, but I would like some comments to be updated.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainSenderFlowV1.kt
vlajos: I made it public for the tests. Probably not the greatest reason, but it looked pretty harmless.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
vlajos: Thanks. I have updated them.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainRequestV1.kt
vlajos: I have removed this one.

blsemo (approved):
LGTM

PR #4268, opened 2023-07-10
CORE-13937: two party finality optimisation

jennyang-r3:
# Overview
This PR optimises two-party finality flow, where there is the possibility to skip second session communication by knowing the number of parties involved. There wouldn't be unseen signatures if there were two parties, so there is no need to send or receive those signatures. Instead, this implementation skips this part.


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -63,7 +63,11 @@ class UtxoFinalityFlowV1(
         val (transaction, signaturesReceivedFromSessions) = receiveSignaturesAndAddToTransaction()
         verifyAllReceivedSignatures(transaction, signaturesReceivedFromSessions)
         persistTransactionWithCounterpartySignatures(transaction)
-        sendUnseenSignaturesToCounterparties(transaction, signaturesReceivedFromSessions)
+
+        if (sessions.size > 2) {
```
vlajos: Generally, `number of parties` = `sessions +1`, so this will be true only when there are at least 4 parties.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -79,7 +83,7 @@ class UtxoFinalityFlowV1(
 
     @Suspendable
     private fun sendTransactionAndBackchainToCounterparties() {
-        flowMessaging.sendAll(initialTransaction, sessions.toSet())
+        flowMessaging.sendAll(mapOf("initialTransaction" to initialTransaction, "numberOfParties" to sessions.size), sessions.toSet())
```
vlajos: Sessions do not contain the initiator.
Maybe `numberOfCounterParties`? or `numberOfSessions`?

vlajos: Also, the string constants should be moved to the base class to let us reuse them on both the sender and receiver sides.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -58,17 +59,24 @@ class UtxoReceiveFinalityFlowV1(
...
+
         transaction = receiveNotarySignaturesAndAddToTransaction(transaction)
         persistNotarizedTransaction(transaction)
         return transaction
     }
 
     @Suspendable
     private fun receiveTransactionAndBackchain(): UtxoSignedTransactionInternal {
-        val initialTransaction = session.receive(UtxoSignedTransactionInternal::class.java)
+        val payload = session.receive(Map::class.java) as Map<String, *>
```
vlajos: Is this cast needed? `as Map<String, *>` ?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
jennyang-r3: I thought it should be at least 3 parties, why there should be 4 parties? or is that typo?

vlajos: Yes. It should be at least 3 parties, but this `if (sessions.size > 2)` is not true for 3 parties.
Since 3 parties = 2 sessions + the initiator.
It should look like this:
```
if (sessions.size > 1) {
```

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -63,7 +66,11 @@ class UtxoFinalityFlowV1(
         val (transaction, signaturesReceivedFromSessions) = receiveSignaturesAndAddToTransaction()
         verifyAllReceivedSignatures(transaction, signaturesReceivedFromSessions)
         persistTransactionWithCounterpartySignatures(transaction)
-        sendUnseenSignaturesToCounterparties(transaction, signaturesReceivedFromSessions)
+
+        if (sessions.size + INITIATOR > 2) {
```
relyafi: Having an `INITIATOR` constant does help a little bit with readability, but I'd suggest adding a comment explaining why we only do this if the parties are greater than 2.

Personally, I would remove the constant and just use the value inline, or change the > to >=. I'd also explain why we are doing this check in a comment. I think that's clearer, and avoids adding a constant that might have limited use.

lankydan:
We definitely need a V2 version of the flow based on the code change.

Are we going to do this in a similar way to what @vlajos did with the backchain resolution PR or duplicate things?

vlajos:
> We definitely need a V2 version of the flow based on the code change.
> 
> Are we going to do this in a similar way to what @vlajos did with the backchain resolution PR or duplicate things?

Yes. I suggested implementing first the change without versioning and then decide about how to version it exactly.
I am not too sure about the answer to the second question, let's discuss it on slack.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/UtxoFinalityConstants.kt
```diff
@@ -0,0 +1,5 @@
+package net.corda.ledger.utxo.flow.impl.flows.finality
+
+const val INITIAL_TRANSACTION = "INITIAL_TRANSACTION"
+const val NUMBER_OF_COUNTER_PARTIES = "NUMBER_OF_PARTIES"
+const val INITIATOR = 1
```
relyafi: Personally, I don't think the `INITIATOR` constant adds much value. The value of 1 doesn't really make sense outside of where it is used in the code, we are simply using this to say that our parties count is 1 higher than the number of sessions. But this is very specific to particular checks in the code, so I don't think this is a good candidate for a constant.

I can see how `INITIAL_TRANSACTION` and `NUMBER_OF_COUNTER_PARTIES` are useful in the context of defining elements of the flow message payloads, but as it stands, there is nothing to indicate what these relate to - `INITIAL_TRANSACTION` could mean anything. It could also clash with another constants file which defines the same value.

An easy way to resolve this is to use an object to "namespace" the constants, see [here](https://github.com/corda/corda-runtime-os/blob/5ef733499c9d2757d82931f870aae7a9e844e45e/libs/uniqueness/common/src/main/kotlin/net/corda/uniqueness/datamodel/common/UniquenessConstants.kt) for an example of this. What we could do, is wrap these settings in an object, e.g. `FinalityFlowPayload`. This would then be referenced in code, e.g. `FinalityFlowPayload.INITIAL_TRANSACTION`.

Personally, I'm a big fan of using objects to namespace generally, as I think it aids code maintainability (in case there are clashes with other files).

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -79,7 +86,12 @@ class UtxoFinalityFlowV1(
 
     @Suspendable
     private fun sendTransactionAndBackchainToCounterparties() {
-        flowMessaging.sendAll(initialTransaction, sessions.toSet())
+        flowMessaging.sendAll(
+            mapOf(
```
lankydan: I think we should send a POJO instead of a map.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -73,17 +77,24 @@ class UtxoReceiveFinalityFlowV1(
             session.send(payload)
             throw CordaRuntimeException(payload.message)
         }
-        transaction = receiveSignaturesAndAddToTransaction(transaction)
-        verifyAllReceivedSignatures(transaction)
-        persistenceService.persist(transaction, TransactionStatus.UNVERIFIED)
+
+        if (numberOfParties > 2) {
```
relyafi: I don't think this approach is necessarily wrong, but there is an alternative to consider here. This flow doesn't actually need to know the number of parties, it just needs to know whether it needs to wait to receive additional signatures.

An alternative to this would be to define a boolean variable (and also in the payload), e.g. `waitForAdditionalSignatures`. This would in my opinion be a bit clearer from an API / code perspective.

It is however a bit less flexible than sending the number of parties - there may be a need in future to use this information for something else, so I think the current approach is also valid. If we do keep this approach, I would suggest adding comments to make it clear why we are doing this check.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -79,7 +86,12 @@ class UtxoFinalityFlowV1(
...
+                INITIAL_TRANSACTION to initialTransaction,
+                NUMBER_OF_COUNTER_PARTIES to sessions.size
+            ), sessions.toSet()
```
relyafi: Personally I find having dedicated payload data classes to be more readable than defining maps, but having discussed with @vlajos he has convinced me that this approach makes it easier to add new attributes without breaking compatibility. So I think this approach is fine (subject to the earlier comment on constant namespacing).

relyafi (commented):
Generally looks good. I'm keen to make sure our constants are namespaced, but the other comments are mostly other options to consider.

relyafi (commented):
Sorry, forgot to mention in my original review - it would also be good to get an indication as to whether this has led to performance improvements, by running a before and after with the performance tests, see https://github.com/corda/corda-non-functional-test/tree/release/5.1/apps/ledger-performance.

In particular, the `single-issue-multi-party` scenario would allow you to control how many participants there are in the transaction. I suspect that wouldn't invoke backchain resolution, but it would require additional signatures.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
jennyang-r3: @lankydan wdyt as opposed to pojo?

lankydan: That's true, but we can do something similar to the group parameters? In that it has its own type, which exposes methods for getting data from the object, but the underlying data is held within a map instead of POJO fields.

Something like:

```kotlin
data class Data @ConstructorForDeserialization constructor(private val map: Map<String, Any?>) {

  constructor(initialTransaction: UtxoSignedTransaction, numberOfCounterParties: Int) : this(
    mapOf(
      INITIAL_TRANSACTION to initialTransaction,
      NUMBER_OF_COUNTER_PARTIES to numberOfCounterParties
    )
  )

  val initialTransaction get() = map[INITIAL_TRANSACTION]

 ...
}
```

We only need to ensure that the constructor for deserialization is never changed. The secondary constructor is to simplify the usage and restricts the constants to be internal to the POJO. The secondary constructor can change between versions without breaking things since its not be used to deserialize the object.

Generally, I want to stop our code from being plagued with constants everywhere to access data from maps.

Just implement the PR ignoring the comment about the map, I want to discuss this with @blsemo first.

lankydan:
> Sorry, forgot to mention in my original review - it would also be good to get an indication as to whether this has led to performance improvements, by running a before and after with the performance tests, see https://github.com/corda/corda-non-functional-test/tree/release/5.1/apps/ledger-performance.
> 
> In particular, the `single-issue-multi-party` scenario would allow you to control how many participants there are in the transaction. I suspect that wouldn't invoke backchain resolution, but it would require additional signatures.

@jennyang-r3 finish the PR first before doing any perf testing as it will take a while to set up.

Me or @nkovacsx could run these tests instead to save time.

relyafi (commented):
This looks ok to me now, but will defer approval until @lankydan has had chance to discuss with @blsemo as that might change the direction of the payload logic.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
relyafi: > That's true, but we can do something similar to the group parameters? In that it has its own type, which exposes methods for getting data from the object, but the underlying data is held within a map instead of POJO fields.
> ...

This is an interesting idea, it does add a bit more complexity to the data class, but potentially is the best of both worlds. As long as any attributes added in future are optional or have defaults, this could work.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -43,6 +47,13 @@ class UtxoFinalityFlowV1(
 
     private val transactionId = initialTransaction.id
 
+    /*
+    * only if the number of sessions(counterparties) is more than one,
+    * it should wait for additional signatures.
+    * Otherwise, it can be skipped since there isn't unseen signatures
```
vlajos:
```suggestion
    * if the number of sessions(counterparties) is more than one,
    * it should wait for the signatures of the other parties.
    * Otherwise, it can be skipped since no other parties can produce more signatures.
    * (V1 flow does not support bypassing that step.)
```

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -43,6 +47,13 @@ class UtxoFinalityFlowV1(
...
+    */
+    private val waitForAdditionalSignatures = !(version == UtxoFinalityVersion.V2 && sessions.size == 1)
```
vlajos:
```suggestion
    private val waitForAdditionalSignatures = (version == UtxoFinalityVersion.V1 || sessions.size >1)
```
looks a bit simpler for me.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -73,17 +78,30 @@ class UtxoReceiveFinalityFlowV1(
...
         }
-        transaction = receiveSignaturesAndAddToTransaction(transaction)
-        verifyAllReceivedSignatures(transaction)
-        persistenceService.persist(transaction, TransactionStatus.UNVERIFIED)
+
+        if (waitForAdditionalSignatures) {
+            transaction = receiveSignaturesAndAddToTransaction(transaction)
+            verifyAllReceivedSignatures(transaction)
+            persistenceService.persist(transaction, TransactionStatus.UNVERIFIED)
+        }
```
vlajos:
```suggestion
        if (waitForAdditionalSignatures) {
            transaction = receiveSignaturesAndAddToTransaction(transaction)
        }
        verifyAllReceivedSignatures(transaction)
        if (waitForAdditionalSignatures) {
            persistenceService.persist(transaction, TransactionStatus.UNVERIFIED)
        }
```
The `verifyAllReceivedSignatures` is still needed, but the other two are not.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1Test.kt
```diff
@@ -473,18 +480,45 @@ class UtxoReceiveFinalityFlowV1Test {
...
         verify(flowEngine, never()).subFlow(TransactionBackchainResolutionFlow(signedTransaction.dependencies, session))
     }
 
-    private fun callReceiveFinalityFlow(validator: UtxoTransactionValidator = UtxoTransactionValidator { }) {
-        val flow = UtxoReceiveFinalityFlowV1(session, validator)
+    @Test
+    fun `skip receiving and persisting signatures when there are only two parties`() {
+        whenever(session.receive(Map::class.java)).thenReturn(receivedPayloadV2ForTwoParties)
+        whenever(signedTransaction.addMissingSignatures()).thenReturn(signedTransactionWithOwnKeys to listOf(signature1))
+        whenever(session.receive(List::class.java)).thenReturn(listOf(signature2))
```
vlajos: This receive mock is not needed here.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1Test.kt
```diff
@@ -473,18 +480,45 @@ class UtxoReceiveFinalityFlowV1Test {
...
-        callReceiveFinalityFlow()
+        callReceiveFinalityFlow(UtxoFinalityVersion.V1)
 
         verify(flowEngine, never()).subFlow(TransactionBackchainResolutionFlow(signedTransaction.dependencies, session))
     }
 
-    private fun callReceiveFinalityFlow(validator: UtxoTransactionValidator = UtxoTransactionValidator { }) {
-        val flow = UtxoReceiveFinalityFlowV1(session, validator)
+    @Test
+    fun `skip receiving and persisting signatures when there are only two parties`() {
```
vlajos: I would highlight somewhere that these are V2 tests. (In the name of the tests? Or in a comment before the first?)

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
lankydan: Discussed with @blsemo and decided to go with POJO containing a `Map`, with `get()` fields that access properties from the map.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -43,6 +47,13 @@ class UtxoFinalityFlowV1(
 
     private val transactionId = initialTransaction.id
 
+    /*
+    * if the number of sessions(counterparties) is more than one,
+    * it should wait for additional signatures.
+    * Otherwise, it can be skipped since there isn't unseen signatures
+    */
+    private val waitForAdditionalSignatures = (version == UtxoFinalityVersion.V1 || sessions.size > 1)
```
lankydan: Don't need the `()`

lankydan: I don't feel that `waitForAdditionalSignatures` is the right naming here. Since we're really asking, "should we send signatures to other counter parties?".

So maybe `shouldSendAdditionalSignatures` would be a better name.

That said, maybe we put this logic into `sendUnseenSignaturesToCounterparties` instead. In which case the if can just be `version == UtxoFinalityVersion.V1 || sessions.size > 1` because in the context of the method it makes sense.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -32,7 +35,8 @@ import java.security.PrivilegedExceptionAction
 class UtxoFinalityFlowV1(
     private val initialTransaction: UtxoSignedTransactionInternal,
     private val sessions: List<FlowSession>,
-    private val pluggableNotaryClientFlow: Class<PluggableNotaryClientFlow>
+    private val pluggableNotaryClientFlow: Class<PluggableNotaryClientFlow>,
+    val version: UtxoFinalityVersion
```
lankydan, line 38: Can the `version` be private?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -28,14 +31,16 @@ import org.slf4j.LoggerFactory
...
+    private val validator: UtxoTransactionValidator,
+    val version: UtxoFinalityVersion
 ) : UtxoFinalityBaseV1() {
 
     private companion object {
         private val log: Logger = LoggerFactory.getLogger(UtxoReceiveFinalityFlowV1::class.java)
     }
 
     override val log: Logger = UtxoReceiveFinalityFlowV1.log
+    private var waitForAdditionalSignatures = true
```
lankydan: I would rather we tried to stay away from storing data in property variables in flows. There's no reason we can't, but it can make it confusing as to where it comes from.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -73,17 +78,32 @@ class UtxoReceiveFinalityFlowV1(
...
-        val initialTransaction = session.receive(UtxoSignedTransactionInternal::class.java)
+        lateinit var initialTransaction: UtxoSignedTransactionInternal
+
+        if (version == UtxoFinalityVersion.V1) {
+            initialTransaction = session.receive(UtxoSignedTransactionInternal::class.java)
+        } else {
+            val payload = session.receive(Map::class.java)
+            initialTransaction = payload[INITIAL_TRANSACTION] as UtxoSignedTransactionInternal
+            waitForAdditionalSignatures = payload[WAIT_FOR_ADDITIONAL_SIGNATURES] as Boolean
+        }
```
lankydan: We shouldn't really be using `lateinit var` in many scenarios.

We can change this code to remove the need for it.

```kotlin
val (initialTransaction, waitForAdditionalSignatures) = if (version == UtxoFinalityVersionV1) {
  initialTransaction = session.receive(UtxoSignedTransactionInternal::class.java) to true
} else {
  val payload = session.receive(Map::class.java)
  val initialTransaction = payload[INITIAL_TRANSACTION] as UtxoSignedTransactionInternal
  val waitForAdditionalSignatures = waitForAdditionalSignatures = payload[WAIT_FOR_ADDITIONAL_SIGNATURES] as Boolean
  initialTransaction to waitForAdditionalSignatures
}
```

We can then return `initialTransaction` and `waitForAdditionalSignatures` from this method.

lankydan: What do others think? @vlajos

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -73,17 +78,32 @@ class UtxoReceiveFinalityFlowV1(
...
-        verifyAllReceivedSignatures(transaction)
-        persistenceService.persist(transaction, TransactionStatus.UNVERIFIED)
+
+        if (waitForAdditionalSignatures) {
+            transaction = receiveSignaturesAndAddToTransaction(transaction)
+        }
+            verifyAllReceivedSignatures(transaction)
+        if (waitForAdditionalSignatures) {
+            persistenceService.persist(transaction, TransactionStatus.UNVERIFIED)
+        }
```
lankydan: Putting this into a method would be a good idea, we'd need to return the pass the transaction into the method and return the updated one from it.

I would also change the if statements slightly since they look at bit odd (in my opinion).

```kotlin
if (waitForAdditionalSignatures) {
  transaction = receiveSignaturesAndAddToTransaction(transaction)
  verifyAllReceivedSignatures(transaction)
  persistenceService.persist(transaction, TransactionStatus.UNVERIFIED)
} else {
  verifyAllReceivedSignatures(transaction)
}
```

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1Test.kt
```diff
@@ -1029,11 +1030,105 @@ class UtxoFinalityFlowV1Test {
         verify(flowEngine, never()).subFlow(TransactionBackchainSenderFlow(TX_ID, sessionBob))
     }
 
-    private fun callFinalityFlow(signedTransaction: UtxoSignedTransactionInternal, sessions: List<FlowSession>) {
+    @Test
+    fun `not sending unseen signatures to counterparties when there are only two parties for V2`() {
```
lankydan: I'd write this like either:

`do not send unseen signatures to counterparties when there are only two parties for V2`

or

`when there are only two parties for V2 do not send unseen signatures to counterparties`.

lankydan (commented):
@vlajos can you take a look at some of my code suggestions since some of them are debatable/preference.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
jennyang-r3: This field is accessed by unit test. Is there a way to access to a private field using JUnit? If so, this can be private I guess.

lankydan: K, just leave it then.

There isn't really a good way to make it private but use it from a test.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
jennyang-r3: Shall I pass this variable as a parameter to functions instead then?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
blsemo: Yes, I agree - lateinit on a stack variable is a code smell in kotlin. Should not be necessary.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
jennyang-r3: `sendTransactionAndBackchainToCounterparties` needs this variable to send receive flow before `sendUnseenSignaturesToCounterparties` is executed. But I think I can define this logic in `call()` and pass `shouldSendAdditionalSignatures ` to those methods to handle this logic in the methods instead.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
relyafi: Can we apply the `@VisibleForTesting` annotation at a constructor argument level, to flag the intent?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
vlajos: Yes, `shouldSend` is more accurate than `waitFor`. `Transfer` could be another option, but I am fine with `shouldSend` too.

Regarding inlining the variable: It is needed in two places. I would keep it as an explicit variable.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
vlajos: Yes. `receiveTransactionAndBackchain` could return that value to the main `call()`, which could pass it forward where needed.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
vlajos: I agree too. @lankydan 's suggestion looks good to me.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/UtxoFinalityFlow.kt
```diff
@@ -40,7 +41,13 @@ class UtxoFinalityFlowVersionedFlowFactory(
...
-            version >= 1 -> UtxoFinalityFlowV1(transaction, sessions, pluggableNotaryClientFlow)
+            version >= CORDA_5_1.value -> UtxoFinalityFlowV1(transaction, sessions, pluggableNotaryClientFlow, UtxoFinalityVersion.V2)
+            version in 1 until CORDA_5_1.value -> UtxoFinalityFlowV1(
+                transaction,
+                sessions,
+                pluggableNotaryClientFlow,
+                UtxoFinalityVersion.V1
+            )
             else -> throw IllegalArgumentException()
         }
```
vlajos, line 47: This looks a bit more concise:
```
val version when {
    version >= CORDA_5_1.value -> UtxoFinalityVersion.V2
    version in 1 until CORDA_5_1.value -> UtxoFinalityVersion.V1
    else -> throw IllegalArgumentException()
}
return UtxoFinalityFlowV1(transaction, sessions, pluggableNotaryClientFlow, version)
```
With a `getFinalityFlowVersionForPlatformVersion(platformVersion)` function, its counterpart from the `ReceiveFinality` could be removed too.

(Sorry, I did not think about this when I wrote the original pattern...)

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import net.corda.v5.base.annotations.ConstructorForDeserialization
+import net.corda.v5.base.annotations.CordaSerializable
+import net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction
+
+@CordaSerializable
+data class FinalityPayload @ConstructorForDeserialization constructor(val map: Map<String, Any?>) {
+    private companion object {
+        const val INITIAL_TRANSACTION = "INITIAL_TRANSACTION"
+        const val WAIT_FOR_ADDITIONAL_SIGNATURES = "WAIT_FOR_ADDITIONAL_SIGNATURES"
```
vlajos: Ideally, having this consistent with the variables in finality/receive finality would be nice.
To be able to be consistent with both, probably a side-agnostic name would be better. 
Maybe `TRANSFER_ADDITIONAL_SIGNATURES` everywhere?

(This is relatively cosmetic/subjective. Feel free to ignore it.)

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -194,8 +192,12 @@ class UtxoFinalityFlowV1(
     @Suspendable
     private fun sendUnseenSignaturesToCounterparties(
         transaction: UtxoSignedTransactionInternal,
-        signaturesReceivedFromSessions: Map<FlowSession, List<DigitalSignatureAndMetadata>>
+        signaturesReceivedFromSessions: Map<FlowSession, List<DigitalSignatureAndMetadata>>,
+        shouldSendAdditionalSignatures: Boolean
     ) {
+        if (!shouldSendAdditionalSignatures) {
```
vlajos: I think this `if` can stay in the main `call()`

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -73,17 +76,39 @@ class UtxoReceiveFinalityFlowV1(
...
+        var verifiedTransaction = transaction
+        if (shouldWaitForAdditionalSignatures) {
+            verifiedTransaction = receiveSignaturesAndAddToTransaction(verifiedTransaction)
+            verifyAllReceivedSignatures(verifiedTransaction)
+            persistenceService.persist(verifiedTransaction, TransactionStatus.UNVERIFIED)
+        } else {
+            verifyAllReceivedSignatures(verifiedTransaction)
+        }
+
+        return verifiedTransaction
```
vlajos: It is better to avoid `verify` in this case. (It is used for quite a few things already. The base class has 5 verifySomething methods.)

`transactionWithAllSignatorySignatures` would be the correct name. But it is pretty ugly... :)

Maybe just "kinda" inlining it?
```suggestion
        return if (shouldWaitForAdditionalSignatures) {
            receiveSignaturesAndAddToTransaction(transaction).also {
              verifyAllReceivedSignatures(it)
              persistenceService.persist(it, TransactionStatus.UNVERIFIED)
            }
        } else {
            verifyAllReceivedSignatures(transaction)
            transaction
        }
```

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -73,17 +76,39 @@ class UtxoReceiveFinalityFlowV1(
...
+    }
+
+    @Suspendable
+    private fun receiveTransactionAndBackchain(): Pair<UtxoSignedTransactionInternal, Boolean> {
+        val (initialTransaction, shouldWaitForAdditionalSignatures) = if (version == UtxoFinalityVersion.V1) {
+            session.receive(UtxoSignedTransactionInternal::class.java) to true
+        } else {
+            val payload = session.receive(FinalityPayload::class.java)
+            payload.initialTransaction as UtxoSignedTransactionInternal to payload.waitForAdditionalSignatures as Boolean
```
vlajos: Could we move the casts to the getters?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
jennyang-r3: I agree with you

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
```diff
@@ -0,0 +1,22 @@
...
+        const val TRANSFER_ADDITIONAL_SIGNATURES = "TRANSFER_ADDITIONAL_SIGNATURES"
+    }
+    constructor(initialTransaction: UtxoSignedTransaction, waitForAdditionalSignatures: Boolean) : this(
+        mapOf(
+            INITIAL_TRANSACTION to initialTransaction,
+            TRANSFER_ADDITIONAL_SIGNATURES to waitForAdditionalSignatures
+        )
+    )
+    val initialTransaction get() = map[INITIAL_TRANSACTION] as UtxoSignedTransactionInternal
+    val waitForAdditionalSignatures get() = map[TRANSFER_ADDITIONAL_SIGNATURES] as Boolean
```
vlajos: The property's name is still the old one. `waitForAdditionalSignatures`

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -51,6 +54,13 @@ class UtxoFinalityFlowV1(
 
     @Suspendable
     override fun call(): UtxoSignedTransaction {
+        /*
+        * if the number of sessions(counterparties) is more than one,
+        * it should wait for additional signatures.
+        * Otherwise, it can be skipped since there isn't unseen signatures
+        */
+        val shouldSendAdditionalSignatures = version == UtxoFinalityVersion.V1 || sessions.size > 1
```
vlajos: I would also rename this one to `transfer...`

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -48,7 +51,7 @@ class UtxoReceiveFinalityFlowV1(
 
     @Suspendable
     override fun call(): UtxoSignedTransaction {
-        val initialTransaction = receiveTransactionAndBackchain()
+        val (initialTransaction, shouldWaitForAdditionalSignatures) = receiveTransactionAndBackchain()
```
vlajos: and this one too.

vlajos (approved):
LGTM

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -97,7 +121,7 @@ class UtxoReceiveFinalityFlowV1(
                 "Transaction with id ${initialTransaction.id} has no dependencies so backchain resolution will not be performed."
             }
         }
-        return initialTransaction
+        return Pair(initialTransaction, transferAdditionalSignatures)
```
lankydan, line 124: you can do `initialTransaction to transferAdditionalSignatures` instead of `Pair` here. That said, it doesn't really matter.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
```diff
@@ -0,0 +1,22 @@
+package net.corda.ledger.utxo.flow.impl.flows.finality
+
+import net.corda.ledger.utxo.flow.impl.transaction.UtxoSignedTransactionInternal
+import net.corda.v5.base.annotations.ConstructorForDeserialization
+import net.corda.v5.base.annotations.CordaSerializable
+import net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction
+
+@CordaSerializable
+data class FinalityPayload @ConstructorForDeserialization constructor(val map: Map<String, Any?>) {
```
lankydan, line 9: There is a bit more work we need to do to this class to make it backwards compatible in all scenarios (that I can think of).

We'll want the map to be `Map<String, ByteArray>` instead. This means that we deserialize each value as needed so that if a future version sends an object that an older version receives but doesn't know how to deserialize then it ok, because that object is never deserialized as its not accessed. The current code would still try to deserialize it.

To achieve this, we'll want to change the input map to hold `ByteArrays` and then pass in a `SerializationService` which serializes each value in the map.

Happy to merge the PR as it is and then we can make another PR for the same ticket that does this.

lankydan: We'd also need to change the getters to be `lazy` evaluated properties similar to `LazyStateAndRefImpl`, which means we'd also need a kryo serializer because kryo doesn't handle the lazy properties well. Note, a kryo serializer was added to `LazyStateAndRefImpl` for that exact reason and we'd probably go with the same fix.

lankydan (approved):

PR #4271, opened 2023-07-11
CORE-15358: Avoid closing the retry manager

yift-r3:
The life cycle coordinator can't always manage when one closes and immediately open a coordinator with the same name. Instead, with this change the coordinator will always be alive, and we will just stop and start it when needed.


dimosr (approved):
LGTM

PR #4272, opened 2023-07-11
CORE-15257: Handle RestApiVersions in EndpointNameConflictValidator when endpoint name conflicts occurs

Tom-Fitzpatrick:
This relaxes the requirements in EndpointNameConflictValidator to allow endpoints with the same path and endpoint type if they have different, non overlapping versions.

### DIFF
--- a/libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
+++ b/libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
@@ -2,6 +2,7 @@ package net.corda.rest.tools.annotations.validation
 
 import net.corda.rest.RestResource
 import net.corda.rest.annotations.RestApiVersion
+import net.corda.rest.annotations.retrieveApiVersionsSet
 import net.corda.rest.tools.annotations.validation.utils.EndpointType
 import net.corda.rest.tools.annotations.validation.utils.endpointPath
 import net.corda.rest.tools.annotations.validation.utils.endpointType
@@ -23,31 +24,28 @@ internal class EndpointNameConflictValidator(private val clazz: Class<out RestRe
 
     override fun validate(): RestValidationResult = validateSameTypeEndpoints(clazz.endpoints)
 
-    private data class VersionRange(val minVersion: RestApiVersion, val maxVersion: RestApiVersion) {
-        fun overlapsWith(other: VersionRange): Boolean {
-            return other.minVersion <= maxVersion && other.maxVersion >= minVersion
-        }
-    }
-
     private fun validateSameTypeEndpoints(endpoints: List<Method>): RestValidationResult {
-        val pathsAndTypesToVersions = mutableMapOf<Pair<String?, EndpointType>, MutableList<VersionRange>>()
+        val pathsToTypesAndVersions = mutableSetOf<Triple<String?, EndpointType, RestApiVersion>>()
         return endpoints.fold(RestValidationResult()) { total, method ->
             val endpointType = method.endpointType
-            total + pathsAndTypesToVersions.validateNoDuplicatePath(method, endpointType)
+            total + pathsToTypesAndVersions.validateNoDuplicatePath(method, endpointType)
         }
     }
 
-    private fun MutableMap<Pair<String?, EndpointType>, MutableList<VersionRange>>.validateNoDuplicatePath(
+    private fun MutableSet<Triple<String?, EndpointType, RestApiVersion>>.validateNoDuplicatePath(
         method: Method,
         type: EndpointType
     ): RestValidationResult {
         val path = method.endpointPath(type)?.lowercase()
-        val newVersionRange = VersionRange(method.restApiVersions.minVersion, method.restApiVersions.maxVersion)
-        return if (this[path to type]?.any { it.overlapsWith(newVersionRange) } == true) {
-            RestValidationResult(listOf(error(path, type, method)))
-        } else {
-            this.getOrPut(path to type) { mutableListOf() }.add(newVersionRange)
-            RestValidationResult()
+        val newVersions = retrieveApiVersionsSet(method.restApiVersions.minVersion, method.restApiVersions.maxVersion)
+
+        newVersions.forEach {
+            val version = Triple(path, type, it)
+            if (!this.add(version)) {
+                return RestValidationResult(listOf(error(path, type, method)))
+            }
         }
+
+        return RestValidationResult()
     }
 }
--- a/libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
+++ b/libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
@@ -18,9 +18,10 @@ internal class EndpointNameConflictValidator(private val clazz: Class<out RestRe
 
     companion object {
         fun error(path: String?, type: EndpointType, method: Method, conflictingMethod: Method): String =
-            "Duplicate endpoint path '$path' with $type HTTP method for version range (${method.restApiVersions.minVersion} -> " +
-                    "${method.restApiVersions.maxVersion}) in '${method.declaringClass.simpleName}.${method.name}'." +
-                    "Conflicting method: '${conflictingMethod.declaringClass.simpleName}.${conflictingMethod.name}'"
+            "Duplicate endpoint path '$path' with $type HTTP method in '${method.declaringClass.simpleName}.${method.name}' " +
+                    "for version range (${method.restApiVersions.minVersion} -> ${method.restApiVersions.maxVersion})." +
+                    "Conflicting method: '${conflictingMethod.declaringClass.simpleName}.${conflictingMethod.name}' " +
+                    "with versions (${conflictingMethod.restApiVersions.minVersion} -> ${conflictingMethod.restApiVersions.maxVersion})."
     }
 
     override fun validate(): RestValidationResult = validateSameTypeEndpoints(clazz.endpoints)
@@ -40,16 +41,17 @@ internal class EndpointNameConflictValidator(private val clazz: Class<out RestRe
         val path = method.endpointPath(type)?.lowercase()
         val newVersions = retrieveApiVersionsSet(method.restApiVersions.minVersion, method.restApiVersions.maxVersion)
 
+        val conflicts = mutableSetOf<String>()
         newVersions.forEach {
             val version = Triple(path, type, it)
             if (this.keys.contains(version)) {
                 val existingMethod = this.getValue(version)
-                return RestValidationResult(listOf(error(path, type, method, existingMethod)))
+                conflicts.add(error(path, type, method, existingMethod))
             } else {
                 this[version] = method
             }
         }
 
-        return RestValidationResult()
+        return RestValidationResult(conflicts.toList())
     }
 }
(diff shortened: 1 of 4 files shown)


filipesoliveira (dismissed):
LGTM

libs/rest/rest-tools/src/test/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidatorTest.kt
```diff
@@ -15,7 +16,25 @@ import kotlin.reflect.jvm.javaMethod
...
+            abstract fun test()
+
+            @HttpGET("/test", minVersion= RestApiVersion.C5_1)
+            @Suppress("unused")
+            abstract fun test2()
+        }
+
+        val result = EndpointNameConflictValidator(TestInterface::class.java).validate()
+
+        assertEquals(0, result.errors.size)
```
josephzunigadaly: I would recommend planning for the failure of asserts. This assert works and we may never see it fail, but in general I'd suggest considering what the error message will look like.

If we break the validator, this assert will print "expected: <0> but was: <1>". On failure, it would give the developer a head start if we used an assert that prints the list.

This assert:
```kotlin
assertEquals(emptyList<String>(), result.errors)
```
would output:
```text
expected: <[]> but was: <[Duplicate endpoint path '/test' for GET HTTP method in 'TestInterface.test2'.]>
```

libs/rest/rest-tools/src/test/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidatorTest.kt
```diff
@@ -15,7 +16,25 @@ import kotlin.reflect.jvm.javaMethod
...
+    }
+
+    @Test
+    fun `validate withEndpointNameConflictOnSamePathSameVersions errorListContainsError`() {
```
josephzunigadaly: This test covers overlap because the versions are not defined and that means all versions are implied.

I think we should cover versions that are defined and overlap. Including a version range completely enclosed within another. As in (1 - 4) and (2 - 3).

libs/rest/rest-tools/src/test/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidatorTest.kt
Tom-Fitzpatrick: Ah interesting, great point, I'll change it to assert in this way

libs/rest/rest-tools/src/test/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidatorTest.kt
Tom-Fitzpatrick: Fair point, would be better to get coverage over more overlapping versioning scenarios. I'll add these. Thanks :)

libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
```diff
@@ -15,28 +17,36 @@ internal class EndpointNameConflictValidator(private val clazz: Class<out RestRe
...
 
-    private fun MutableSet<Pair<String?, EndpointType>>.validateNoDuplicatePath(
+    private fun MutableMap<Pair<String?, EndpointType>, MutableList<VersionRange>>.validateNoDuplicatePath(
         method: Method,
         type: EndpointType
     ): RestValidationResult {
         val path = method.endpointPath(type)?.lowercase()
-        return if (this.contains(path to type)) {
+        val newVersionRange = VersionRange(method.restApiVersions.minVersion, method.restApiVersions.maxVersion)
+        return if (this[path to type]?.any { it.overlapsWith(newVersionRange) } == true) {
```
josephzunigadaly: It's always output one error with the details of one method but the error is caused by two methods. It might be useful to include the other method too.

libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
```diff
@@ -15,28 +17,36 @@ internal class EndpointNameConflictValidator(private val clazz: Class<out RestRe
...
     companion object {
         fun error(path: String?, type: EndpointType, method: Method): String =
-            "Duplicate endpoint path '$path' for $type HTTP method in '${method.declaringClass.simpleName}.${method.name}'."
+            "Duplicate endpoint path '$path' with $type HTTP method for version range (${method.restApiVersions.minVersion} -> " +
+                    "${method.restApiVersions.maxVersion}) in '${method.declaringClass.simpleName}.${method.name}'."
     }
 
     override fun validate(): RestValidationResult = validateSameTypeEndpoints(clazz.endpoints)
 
+    private data class VersionRange(val minVersion: RestApiVersion, val maxVersion: RestApiVersion) {
```
vkolomeyko: Note that we already have similar sort container `MinMaxRestApiVersions`.

Also when testing for overlap, IMO it would be best to use logic similar to `net.corda.rest.server.impl.apigen.processing.APIStructureRetriever#retrieveApiVersionsSet` (it will need moving to a different module).
Once `Set` is obtained it should be fine to test for overlap.

Reason being is that when we test with `<=` - it relies on the `ordinal` value of enum, i.e. on the order in which enum is defined which is not super-reliable.

libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
```diff
@@ -15,28 +17,36 @@ internal class EndpointNameConflictValidator(private val clazz: Class<out RestRe
...
+        val pathsAndTypesToVersions = mutableMapOf<Pair<String?, EndpointType>, MutableList<VersionRange>>()
         return endpoints.fold(RestValidationResult()) { total, method ->
             val endpointType = method.endpointType
-            total + pathsToTypes.validateNoDuplicatePath(method, endpointType)
+            total + pathsAndTypesToVersions.validateNoDuplicatePath(method, endpointType)
         }
     }
 
-    private fun MutableSet<Pair<String?, EndpointType>>.validateNoDuplicatePath(
+    private fun MutableMap<Pair<String?, EndpointType>, MutableList<VersionRange>>.validateNoDuplicatePath(
```
vkolomeyko: Alternative could be use of `MutableSet<Triple<String?, EndpointType, RestApiVersion>>` then `overlapsWith` will not be necessary as for every version of the method it would be possible to check if there is already an entry in a `Set`.

libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
Tom-Fitzpatrick: Thanks, that makes sense. I refactored this to use `retrieveApiVersionsSet` as you suggested, which seems a lot tidier. In order to make it available here though I moved the method to `libs/rest/rest/src/main/kotlin/net/corda/rest/annotations/RestApiVersion.kt`. Do you think this is suitable move, or would there be somewhere better for this sort of method?

libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
Tom-Fitzpatrick: Good point, I have modified it to also specify the conflicting method in the error now. Thanks

libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
Tom-Fitzpatrick: I ended up using a map so I could track the method that it's conflicting with for the error, but I went with this general approach and got rid of the `overlapsWith` method. Thanks

libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
```diff
@@ -23,31 +24,28 @@ internal class EndpointNameConflictValidator(private val clazz: Class<out RestRe
...
-            RestValidationResult(listOf(error(path, type, method)))
-        } else {
-            this.getOrPut(path to type) { mutableListOf() }.add(newVersionRange)
-            RestValidationResult()
+        val newVersions = retrieveApiVersionsSet(method.restApiVersions.minVersion, method.restApiVersions.maxVersion)
+
+        newVersions.forEach {
+            val version = Triple(path, type, it)
+            if (!this.add(version)) {
+                return RestValidationResult(listOf(error(path, type, method)))
```
vkolomeyko: It is probably best not to use `return` here but accumulate all the conflicts into a `list` and then return `RestValidationResult` once at the end.

Also, when calling `error` it would probably makes sense to pass `version` as well to have full details where the conflict has occurred.

vkolomeyko (commented):
Looks fine to me. One minor comment posted.

libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
```diff
@@ -14,30 +17,39 @@ import java.lang.reflect.Method
...
-        } else {
-            this.add(path to type)
-            RestValidationResult()
+        val newVersions = retrieveApiVersionsSet(method.restApiVersions.minVersion, method.restApiVersions.maxVersion)
+
+        newVersions.forEach {
+            val version = Triple(path, type, it)
+            if (this.keys.contains(version)) {
+                val existingMethod = this.getValue(version)
+                return RestValidationResult(listOf(error(path, type, method, existingMethod)))
```
vkolomeyko: It is probably best not to use `return` here but accumulate all the conflicts into a `list` and then return `RestValidationResult` once at the end.

Also, when calling `error` it would probably makes sense to pass `version` as well to have full details where the conflict has occurred.

libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
vkolomeyko: Yes, it is fine to have `retrieveApiVersionsSet` relocated to that module.

libs/rest/rest-tools/src/main/kotlin/net/corda/rest/tools/annotations/validation/EndpointNameConflictValidator.kt
Tom-Fitzpatrick: Ah, good catch, I fixed this, and I also added a test to cover a scenario where a new endpoint version range conflicts with more than one existing one, so we see errors reported for both conflicts. 

I added the version details for the conflicted method to the error message too, we had them in the method already, so I didn't need to pass it in separately. 

This is what the error looks like now:

`Duplicate endpoint path '/test' with GET HTTP method in 'TestInterface.test2' for version range (C5_1 -> C5_2). Conflicting method: 'TestInterface.test' with versions (C5_0 -> C5_1).`

vkolomeyko (approved):
LGTM

josephzunigadaly (approved):

PR #4274, opened 2023-07-11
CORE-15101: Move CheckClusterRolesE2eTest, UserRoleAssociationE2eTest and LimitedUserAuthorizationE2eTest

BenYip123:
- Remove CheckClusterRolesE2eTest, UserRoleAssociationE2eTest and LimitedUserAuthorizationE2eTest

Moved to `corda-e2e-tests` repo in this PR: [CORE-15101: Add CheckClusterRolesE2eTest, UserRoleAssociationE2eTest and LimitedUserAuthorizationE2eTest #148](https://github.com/corda/corda-e2e-tests/pull/148)


testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterBuilder.kt
```diff
@@ -357,11 +357,15 @@ class ClusterBuilder {
     fun getPermissionByQuery(
         limit: Int,
         permissionType: String,
+        groupVisibility: String? = null,
```
vkolomeyko: I would remove those changes, especially given that underlying E2E PR is passing without those changes.

The story with `virtualNode` argument may change in the future and I would refrain from increasing footprint of it in our testing code.

BenYip123: I was thinking it would be easier to have all the optional parameters readily available, but I will change it back as to avoid more changes in E2E repo

vkolomeyko (approved):
LGTM

PR #4277, opened 2023-07-11
CORE-15127 Support bundle - Handle error when namespace isn't set

jacob-scott3141:


support_bundle.sh
```diff
@@ -6,6 +6,11 @@ else
   namespace="$1"
 fi
 
+if [ -z "$namespace" ]; then
```
davidcurrie: in the name of consistency
```suggestion
if [ -z "${namespace}" ]; then
```

davidcurrie (approved):
LGTM

PR #4279, opened 2023-07-11
CORE-15389: Populate the X509 name field if the cert verification fails

anien:
The signature of the notary CPB now includes the certificate chain.
We were relying on the fact that if the validation of the certificate fails, if shows the index of the certificate which failed the validation. Unfortunately the documentation says, it MAY include this index, but it doesn't have to. So we can't rely on that information. 
We now instead always print the X509 name of the first certificate in the chain if the index is not set.


dickon (approved):
LGTM

Bit poor that the validator didn't return a useful index. Long term may be we'll get an improved java.security.cert validator or find another one, but going with the standard seems okay for now.

driessamyn (approved):
LGTM

PR #4283, opened 2023-07-11
CORE-15389: Add test to verify the message if root CA is incorrect in verification

anien:
New tests:
- Valid scenario when root CA is included in the signature of CPB.
- Negative scenario when different root CA is used to verify the signature of CPB which was signed with another root CA. It also checks the error message.


driessamyn (dismissed):
LGTM

driessamyn (approved):

PR #4284, opened 2023-07-12
CORE-9519 API to change group parameters

YashNabar:
Adds an endpoint to the MGM API to update group parameters. The MGM may update the current group parameters by submitting a map of the updated version (which may only include changes to the minimum platform version and custom ("ext.") properties). The platform populates notary information to construct the final group parameters, which are then persisted and distributed to active members on the network.
https://github.com/corda/corda-api/pull/1175

Testing:

1. Member's view of group parameters -
Screenshot 2023-07-12 at 09 38 59
  
  
2. MGM updates group parameters -
Screenshot 2023-07-12 at 09 40 16
  
  
3. Updated group parameters (with updated epoch and modified time fields) are available to the member via the group reader -
Screenshot 2023-07-12 at 09 40 34


filipesoliveira:
Hi @YashNabar. Can you also share which error codes are used for each endpoint? I have noticed that your screenshots of the Swagger webpage does not list them.

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/InternalGroupParametersUtils.kt
```diff
@@ -29,4 +29,6 @@ fun InternalGroupParameters.toPersistentGroupParameters(
             }
         )
     )
-}
\ No newline at end of file
+}
+
+fun InternalGroupParameters.toMap() = entries.associate { it.key to it.value }
```
filipesoliveira: Is there a reason why this function is not part of the interface `InternalGroupParameters`? Why can't the logic in the file `InternalGroupParametersUtils.kt` be in the class itself? Extending a class is great when we don't have access to the source code of a given class. But this is not the case.

charlieR3 (dismissed):
LGTM

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/InternalGroupParametersUtils.kt
YashNabar: These are simply mapping functions intended to be used as extension utilities similar to those from the Kotlin Standard Library. It's largely a question of coding style, but I believe it's fairly standard to define such utils as extensions.

YashNabar:
> Hi @YashNabar. Can you also share which error codes are used for each endpoint? I have noticed that your screenshots of the Swagger webpage does not list them.

@filipesoliveira I have added the list to the PR description.

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/MGMRestResource.kt
```diff
@@ -475,4 +475,29 @@ interface MGMRestResource : RestResource {
...
+        path = "{holdingIdentityShortHash}/group-parameters",
+        description = "This API allows you to make changes to the group parameters by submitting an updated version " +
+                "of the group parameters.",
+        responseDescription = "The newly updated group parameters"
+    )
+    fun updateGroupParameters(
+        @RestPathParameter(description = "The holding identity ID of the MGM")
+        holdingIdentityShortHash: String,
+        @ClientRequestBodyParameter(description = "Updated version of the group parameters")
+        newGroupParameters: Map<String, String>,
```
vkolomeyko: I wonder if it would be a better idea to introduce a dedicated type `GroupParameters` which wraps `Map<String, String>`.
Then it would be possible to use the same type as input value and as a return value, as well as use the same type in  `MemberLookupRestResource.viewGroupParameters`. This would show that all these incoming parameters and return results are related (and strong typed), rather than having seemingly unrelated maps in multiple places.
This will also make OpenAPI definition more concise.

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/InternalGroupParametersUtils.kt
filipesoliveira: I see. My main points are that I find it difficult to understand why code logic that belongs to a class is in a different file or that one has to import a function before such function can be used as a function of the class. Similarly, this strategy makes it harder to find all functions available for a given class.

I understand that it is a coding style. But I wonder if this helps people navigate the code and avoid code duplication. In practice, this strategy is hiding information from the developer which it does not seem beneficial in this scenario. 

Anyway, if this something that people usually do, I won't oppose to it. I am only highlighting some concerns.

filipesoliveira:
> > Hi @YashNabar. Can you also share which error codes are used for each endpoint? I have noticed that your screenshots of the Swagger webpage does not list them.
> 
> @filipesoliveira I have added the list to the PR description.

Can you see the list of errors on the Swagger page? That is important since people will use Swagger to understand the possible outputs of a given endpoint.

libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/InternalGroupParametersUtils.kt
YashNabar: I have moved the new function in [2b99002](https://github.com/corda/corda-runtime-os/pull/4284/commits/2b99002aa1f2c381c2a8062e6ad98d3ba9fddda8)

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/MGMRestResource.kt
YashNabar: Addressed in [2b99002](https://github.com/corda/corda-runtime-os/pull/4284/commits/2b99002aa1f2c381c2a8062e6ad98d3ba9fddda8)

YashNabar:
> > > Hi @YashNabar. Can you also share which error codes are used for each endpoint? I have noticed that your screenshots of the Swagger webpage does not list them.
> > 
> > 
> > @filipesoliveira I have added the list to the PR description.
> 
> Can you see the list of errors on the Swagger page? That is important since people will use Swagger to understand the possible outputs of a given endpoint.

Discussed offline. The default list is visible on Swagger (cropped from attached screenshots).

vkolomeyko (dismissed):
LGTM (for @corda/rest )

gradle.properties
```diff
@@ -45,7 +45,7 @@ bouncycastleVersion=1.73
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.0.0.xx-SNAPSHOT to pick up maven local published copy
 #cordaApiVersion=5.1.0.xxx-SNAPSHOT
-cordaApiVersion=5.1.0.6-beta+
+cordaApiVersion=5.1.0.6-alpha-1689155242210
```
filipesoliveira: Don't forget to change this back to beta. before merging the PR.

charlieR3 (dismissed):

filipesoliveira (dismissed):

vkolomeyko (dismissed):
LGTM

charlieR3 (dismissed):

filipesoliveira (dismissed):

filipesoliveira (approved):

charlieR3 (approved):

PR #4287, opened 2023-07-12
CORE-14775 update the db consumer to check that the records read are contiguous and not pending, also filter out any records that are aborted

LWogan:
If records are being committed in a transaction, its possible a subsequent transaction can populate the topic record table with data first. The poll can then skip records that have not been populated yet to the table. e.g poll position is 50, but next record in the table is offset 54, due to a transaction with records 51, 52, 53 being mid insert.

Offsets are reserved separately to the insertion logic which is why a table can have missing records temporarily while transactions are inserted.


applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
```diff
@@ -564,8 +565,7 @@ class FlowTests {
             .isEqualTo("${bobX500}=echo:m1; ${charlyX500}=echo:m2")
     }
 
-    @Test
-    @Disabled
+    @RepeatedTest(20)
```
ben-millar, line 568: Don't forget to remove before merging

ben-millar (approved):

LWogan:
closing for now.  @ben-millar will take back over this work and open a new pr

PR #4289, opened 2023-07-12
CORE-13164 Interop/Facade demo flows enabled by facade fixes

szymonsztuka:
Facade Demo Issue flows will produce a meaningful message when number of expected notaries in the network differs from 1.


jmacmahonr3 (approved):
An ambitious change. 
I'll allow it.

PR #4291, opened 2023-07-12
CORE-9483 Enable sending of primitives across sessions

ben-millar:
This PR is a follow-up to https://github.com/corda/corda-runtime-os/pull/3768, which renders the original obsolete.

## Background
Historically, the sending and receiving of primitive types across sessions caused failures at runtime and, for this reason, checks were added to prevent this from being done. The investigation as part of CORE-9483 revealed that there were multiple independent causes for these failures; the core serializer bugs preventing the AMQP serialization of raw primitives was fixed by @chrisr3 in PR https://github.com/corda/corda-runtime-os/pull/3904, and so this PR handles the final issues on the flow session side.

## Overview
We expose APIs for sending and receiving data across sessions in both the `FlowSessionImpl` and the `FlowMessagingImpl` classes which generally follow the form:
```
send(Object payload) // Java
send(payload: Any)   // Kotlin
```
```
receive(Class<R> receiveType)   // Java
receive(receiveType: Class<R>)  // Kotlin
```
When we send a payload, its type is inferred and it's serialized as such before being sent to the counterparty. When we receive an payload, we must specify the expected type so that it can be deserialized and used by the client code.

Notably, where `send` accepts _any_ object, `receive` accepts the generic type `Class<R>` which carries more information and is required for the runtime reflection which allows us to cast and use the deserialized bytes correctly.

## Problem
When we try to send a primitive value, it will be autoboxed at runtime and therefore will be serialized as a boxed class before sending. However, the corresponding receive call will still be expecting a primitive type back and will fail when trying to deserialize the received bytes as such.

## Fix
To fix this, we simply add a check to the `receive` calls. If we're trying to receive a primitive type, we should interpret the received bytes as the boxed representation of that type (as we can guarantee it's been boxed when passing through the `send(Object)` interface). It can then be implicit cast back to the expected value by the `<R>` return type of the receive functions.

This PR implements that change and removes the `requireBoxType` checks which are no longer needed. It also adds some unit testing to `SerializationHelper.kt` as I noticed there was none there when checking through the code.

### DIFF
--- a/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
+++ b/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
@@ -211,7 +211,9 @@ class FlowTests {
         assertThat(flowResult.flowError).isNull()
         assertThat(flowResult.json.command).isEqualTo("flow_session_primitives")
         assertThat(flowResult.json.result)
-            .isEqualTo("${bobX500}=Successfully received 8 items.")
+            .isEqualTo("Successfully received 8 items.\n" +
+                    "Successfully received 8 items from receiveAll.\n" +
+                    "Successfully received 8 items from receiveAllMap.\n")
     }
 
     @Test
--- a/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
+++ b/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
@@ -196,6 +196,25 @@ class FlowTests {
     }
 
     @Test
+    fun `Flow Session - Send and receive primitive values across a session`() {
+        val requestBody = RpcSmokeTestInput().apply {
+            command = "flow_session_primitives"
+            data = mapOf("sessions" to bobX500)
+        }
+
+        val requestId = startRpcFlow(bobHoldingId, requestBody)
+
+        val flowResult = awaitRestFlowResult(bobHoldingId, requestId)
+
+        assertThat(flowResult.flowStatus).isEqualTo(RPC_FLOW_STATUS_SUCCESS)
+        assertThat(flowResult.json).isNotNull
+        assertThat(flowResult.flowError).isNull()
+        assertThat(flowResult.json.command).isEqualTo("flow_session_primitives")
+        assertThat(flowResult.json.result)
+            .isEqualTo("${bobX500}=Successfully received 8 items.")
+    }
+
+    @Test
     fun `Persistence - persist a single entity`() {
         val id = UUID.randomUUID()
         val flowResult = persistDog(id)
--- a/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
+++ b/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
@@ -428,27 +428,6 @@ class FlowTests {
     }
 
     @Test
-    fun `Flow Session - Send and receive primitive values across a session`() {
-        val requestBody = RpcSmokeTestInput().apply {
-            command = "flow_session_primitives"
-            data = mapOf("sessions" to bobX500)
-        }
-
-        val requestId = startRpcFlow(bobHoldingId, requestBody)
-
-        val flowResult = awaitRestFlowResult(bobHoldingId, requestId)
-
-        assertThat(flowResult.flowStatus).isEqualTo(RPC_FLOW_STATUS_SUCCESS)
-        assertThat(flowResult.json).isNotNull
-        assertThat(flowResult.flowError).isNull()
-        assertThat(flowResult.json.command).isEqualTo("flow_session_primitives")
-        assertThat(flowResult.json.result)
-            .isEqualTo("Successfully received 8 items.\n" +
-                    "Successfully received 8 items from receiveAll.\n" +
-                    "Successfully received 8 items from receiveAllMap.\n")
-    }
-
-    @Test
     fun `Crypto - Sign and verify bytes`() {
         val requestBody = RpcSmokeTestInput().apply {
             command = "crypto_sign_and_verify"
--- a/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
+++ b/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
@@ -428,6 +428,27 @@ class FlowTests {
     }
 
     @Test
+    fun `Flow Session - Send and receive primitive values across a session`() {
+        val requestBody = RpcSmokeTestInput().apply {
+            command = "flow_session_primitives"
+            data = mapOf("sessions" to bobX500)
+        }
+
+        val requestId = startRpcFlow(bobHoldingId, requestBody)
+
+        val flowResult = awaitRestFlowResult(bobHoldingId, requestId)
+
+        assertThat(flowResult.flowStatus).isEqualTo(RPC_FLOW_STATUS_SUCCESS)
+        assertThat(flowResult.json).isNotNull
+        assertThat(flowResult.flowError).isNull()
+        assertThat(flowResult.json.command).isEqualTo("flow_session_primitives")
+        assertThat(flowResult.json.result)
+            .isEqualTo("Successfully received 8 items.\n" +
+                    "Successfully received 8 items from receiveAll.\n" +
+                    "Successfully received 8 items from receiveAllMap.\n")
+    }
+
+    @Test
     fun `Crypto - Sign and verify bytes`() {
         val requestBody = RpcSmokeTestInput().apply {
             command = "crypto_sign_and_verify"
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
@@ -100,22 +100,25 @@ class FlowSessionImpl(
     @Suspendable
     override fun <R : Any> sendAndReceive(receiveType: Class<R>, payload: Any): R {
         verifySessionStatusNotErrorOrClose(sourceSessionId, flowFiberService)
-        requireBoxedType(receiveType)
         val request = FlowIORequest.SendAndReceive(mapOf(getSessionInfo() to serialize(payload)))
         val received = fiber.suspend(request)
+
         setSessionConfirmed()
-        return deserializeReceivedPayload(received, receiveType)
+
+        return processReceivedPayload(receiveType, received)
     }
 
     @Suspendable
     override fun <R : Any> receive(receiveType: Class<R>): R {
         verifySessionStatusNotErrorOrClose(sourceSessionId, flowFiberService)
-        requireBoxedType(receiveType)
         val request = FlowIORequest.Receive(setOf(getSessionInfo()))
         val received = fiber.suspend(request)
+
         setSessionConfirmed()
-        return deserializeReceivedPayload(received, receiveType)
+
+        return processReceivedPayload(receiveType, received)
     }
+
     @Suspendable
     override fun send(payload: Any) {
         verifySessionStatusNotErrorOrClose(sourceSessionId, flowFiberService)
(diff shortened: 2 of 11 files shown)


libs/serialization/serialization-amqp/src/test/kotlin/net/corda/internal/serialization/amqp/SerializationHelperTests.kt
```diff
@@ -0,0 +1,113 @@
...
+
+import com.google.common.reflect.TypeToken
+import net.corda.v5.base.annotations.CordaSerializable
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertThrows
+import org.junit.jupiter.api.Assertions.assertDoesNotThrow
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+
+class SerializationHelperTests {
```
LWogan, line 12: nice test coverage!

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -114,12 +114,24 @@ internal fun Type.isSubClassOf(type: Type): Boolean {
...
+    when (type) {
+        is Type -> {
+            // See CORDA-2782 for explanation of the special exemption made for Comparable
+            if (!hasCordaSerializable(type.asClass()) && type.asClass() != java.lang.Comparable::class.java) {
+                throw AMQPNotSerializableException(
+                    type,
+                    "Class \"${type}\" is not annotated with @CordaSerializable.")
+            }
+        }
+        is Collection<*> -> {
```
LWogan: I think you should add a kdoc to this function to adequately explain this function. It will be confusing to anyone without context.

LWogan (commented):
This is great work Ben! The logic looks good to me. Given that its in a sensitive area, i'd like to get @mnesbit s input before approving.

We might need to check the kdocs in corda-api on these apis to ensure theres no comments saying "you cant send primitive types" as there might be. You might need to check with the docs team as well to see if theres any guidance given to cordapp developers telling them the same.

emilybowe (commented):
LGTM, pending @mnesbit review
Good explanation of this in the demo earlier, made this much easier to go through 👍

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
```diff
@@ -585,6 +585,27 @@ class FlowTests {
             .isEqualTo("${bobX500}=Completed. Sum:18")
     }
 
+    @Test
```
ben-millar: ~~NB: This test still has the potential to fail on the combined worker until this PR is merged, fixing the DBCordaConsumerImpl: https://github.com/corda/corda-runtime-os/pull/4304~~

The DBCordaConsumer change has been merged. Risk is mitigated.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
```diff
@@ -100,22 +100,25 @@ class FlowSessionImpl(
...
     override fun <R : Any> sendAndReceive(receiveType: Class<R>, payload: Any): R {
         verifySessionStatusNotErrorOrClose(sourceSessionId, flowFiberService)
-        requireBoxedType(receiveType)
         val request = FlowIORequest.SendAndReceive(mapOf(getSessionInfo() to serialize(payload)))
         val received = fiber.suspend(request)
+
         setSessionConfirmed()
-        return deserializeReceivedPayload(received, receiveType)
+
+        return processReceivedPayload(receiveType, received)
```
fenryka: minor comment, but why are we changing the established context of the parameter order? I'm not saying if one is better than the other but it feels a little weird to invert it unless we're bringing things inline with a wider set of established conventions

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
```diff
@@ -100,22 +100,25 @@ class FlowSessionImpl(
...
     override fun <R : Any> receive(receiveType: Class<R>): R {
         verifySessionStatusNotErrorOrClose(sourceSessionId, flowFiberService)
-        requireBoxedType(receiveType)
         val request = FlowIORequest.Receive(setOf(getSessionInfo()))
         val received = fiber.suspend(request)
+
         setSessionConfirmed()
-        return deserializeReceivedPayload(received, receiveType)
+
+        return processReceivedPayload(receiveType, received)
```
fenryka: same comment as above on inverting the parameter ordering

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -114,12 +114,36 @@ internal fun Type.isSubClassOf(type: Type): Boolean {
...
 }
 
-fun requireCordaSerializable(type: Type) {
-    // See CORDA-2782 for explanation of the special exemption made for Comparable
-    if (!hasCordaSerializable(type.asClass()) && type.asClass() != java.lang.Comparable::class.java) {
-        throw AMQPNotSerializableException(
-                type,
-                "Class \"$type\" is not annotated with @CordaSerializable.")
+/**
+ * Checks if the given [type] is Corda serializable (if it has or inherits the [CordaSerializable] annotation),
```
fenryka: Technically this function doesn't check this (in that it returns true or false), it enforces it

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -114,12 +114,36 @@ internal fun Type.isSubClassOf(type: Type): Boolean {
...
+    when (type) {
+        is Type -> {
+            // See CORDA-2782 for explanation of the special exemption made for Comparable
+            if (!hasCordaSerializable(type.asClass()) && type.asClass() != java.lang.Comparable::class.java) {
+                throw AMQPNotSerializableException(
+                    type,
+                    "Class \"${type}\" is not annotated with @CordaSerializable.")
+            }
+        }
+        is Collection<*> -> {
```
fenryka: should we check that the collection type is on our allowed set of collections (If that's still a thing, i.e. those with stable and ordering)

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -114,12 +114,36 @@ internal fun Type.isSubClassOf(type: Type): Boolean {
...
+ * If the passed [type] is a [Collection], this function will be recursively called for each item.
+ *
+ * @param type The type to be checked for Corda serializability.
+ * @throws AMQPNotSerializableException if [type] is not annotated with [CordaSerializable]
+ *                                      and is not an instance of [java.lang.Comparable].
+ * @throws AMQPNoTypeNotSerializableException if [type] is not a valid type, or null, and cannot be serialized.
+ *
+ * @see [CORDA-2782] for the special exemption made for `Comparable`.
+ */
+fun requireCordaSerializable(type: Any?) {
```
fenryka: Question: What do we expect people to do with this function? IIRC the serialiser enforces this, so it is already "required", this function throws earlier in the call stack potentially, but then we're adding an overhead of runtime checking each and every time we call this and thus could do the check at compile time (at least for CorDaps)... so I guess I'm just curious why add this function when this is already enforced

libs/serialization/serialization-amqp/src/test/kotlin/net/corda/internal/serialization/amqp/SerializationHelperTests.kt
```diff
@@ -0,0 +1,113 @@
...
+    @Test
+    fun `asClass should correctly convert type to class`() {
+        val type = object : TypeToken<List<String>>() {}.type
+        val clazz = type.asClass()
+
+        assertEquals(List::class.java, clazz)
+    }
+
+    @Test
+    fun `isArray should correctly determine if type is an array type`() {
```
fenryka, line 22: Should this also test primitive Arrays int[] for example?

fenryka (commented):
Just a few comments, feel free to ignore as otherwise looks fine

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
ben-millar: Whoops! I'll refactor that, good catch.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/SerializationHelper.kt
ben-millar: Actually, this function already existed. This change just extends it to handle collections on a per-element basis (because this is how AMQP will handle them, too).

Originally, I modified it as I was going to use it to enforce that types passed to `send`/`receive` were Corda serializable but, actually, this wouldn't work as we could be passing a concrete type which is annotated serializable, but which is inherited from an interface which is not.

I decided to keep the change in the PR anyway, as I think it's more correct.

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/SerializationHelper.kt
ben-millar: I looked into this. There is an allowed set of collections in the `CollectionSerializer` covering lists and sets, but this is just for the collection types not covered by other serializers (I.e., we have an `ArraySerializer`, `MapSerializer`, etc.).

I'm sure there are some collection types (¿`EnumSet`?) that we can't handle, but it feels too heavy handed to pull together all of these allowed sets from various serializers just for this check. Good question tho, it was informative to check this stuff.

libs/serialization/serialization-amqp/src/test/kotlin/net/corda/internal/serialization/amqp/SerializationHelperTests.kt
ben-millar: c-style arrays in Kotlin, TIL! Added a test case

libs/serialization/serialization-amqp/src/main/kotlin/net/corda/internal/serialization/amqp/SerializationHelper.kt
fenryka: I guess my concern would be that in adding this enforcement check, we're adding a lot of runtime overhead to something that is already enforced later on, i.e. it can't go wrong so we're going to spend a few instructions earlier on also checking that... happy path we do more work, unhappy path we catch it earlier. 

NOTE: by "a lot of" we're talking maybe 20 instructions at most so it's really not significant in any way... I'm just old enough to remember when that would've been an issue we'd have wanted to avoid, old lady shouts at cloud :D

ben-millar: Yeah, actually, I think you're right; it hasn't really been sitting well with me since yesterday 🤔 I'm gonna follow the KISS principal and roll this change back.

testing/cpbs/test-cordapp/src/main/kotlin/com/r3/corda/testing/smoketests/flow/RpcSmokeTestFlow.kt
```diff
@@ -249,6 +250,24 @@ class RpcSmokeTestFlow : ClientStartableFlow {
         return outputs.joinToString("; ")
     }
 
+    @Suspendable
```
simon-johnson-r3, line 253: The scope of smoke tests is being reduced in favour of e2e tests, and I think given this is a specific feature test might be a better fit there. I think @owenstanford is leading that, perhaps he has an opinion?

simon-johnson-r3:
One of the issues that concerned me in the JIRA ticket is that this support is not well documented, and in fact if you try and send something that is not `@CordaSerializable` the error message leads you to believe only `@CordaSerializable` types are allowed. Can we make it clear in the public api that primitive types are now supported? I don't think this is a breaking change to the contract as we are widening it.

simon-johnson-r3 (changes requested):

testing/cpbs/test-cordapp/src/main/kotlin/com/r3/corda/testing/smoketests/flow/RpcSmokeTestFlow.kt
ben-millar: I agree, I'll move this to the E2E test repo. I just wanted it to run against the PR to demonstrate that the changes are working before I port it.

ben-millar: > Can we make it clear in the public api that primitive types are now supported?

Are you looking for a change to the docs on the API side?

ben-millar: API PR to add context https://github.com/corda/corda-api/pull/1184/files

simon-johnson-r3 (approved):

LWogan (approved):
LGTM. As mentioned in my last review, please check with docs team in case there's anything regarding the old restriction that needs to be removed. Will need a release note at a minimum

PR #4296, opened 2023-07-13
CORE-14573: Wrong status code reported for incorrect tenantId

anien:
When requesting a CSR for a tenant, we were not checking if a virtual node is registered for the tenant.
ValidateTenantId check was updated and it now returns 404 RESOURCE_NOT_FOUND if virtual node is not registered for the tenant.
As we do this check early on, tests had to be updated too to make sure they test the original functionality.

### DIFF
--- a/components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/CertificatesRestResourceImpl.kt
+++ b/components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/CertificatesRestResourceImpl.kt
@@ -9,6 +9,7 @@ import net.corda.crypto.client.CryptoOpsClient
 import net.corda.crypto.core.CryptoConsts
 import net.corda.crypto.core.CryptoTenants.P2P
 import net.corda.crypto.core.CryptoTenants.REST
+import net.corda.crypto.core.CryptoTenants.allClusterTenants
 import net.corda.crypto.core.DefaultSignatureOIDMap
 import net.corda.crypto.core.ShortHash
 import net.corda.crypto.core.ShortHashException
@@ -26,6 +27,7 @@ import net.corda.membership.rest.v1.CertificatesRestResource
 import net.corda.membership.rest.v1.CertificatesRestResource.Companion.SIGNATURE_SPEC
 import net.corda.rest.HttpFileUpload
 import net.corda.rest.PluggableRestResource
+import net.corda.rest.exception.BadRequestException
 import net.corda.rest.exception.InvalidInputDataException
 import net.corda.rest.exception.ResourceNotFoundException
 import net.corda.rest.messagebus.MessageBusUtils.tryWithExceptionHandling
@@ -120,8 +122,6 @@ class CertificatesRestResourceImpl @Activate constructor(
         contextMap: Map<String, String?>?,
     ): String {
         validateTenantId(tenantId)
-        // Check if a virtual node is registered for given tenantId
-        virtualNodeInfoReadService.getByHoldingIdentityShortHashOrThrow(tenantId)
 
         val key = tryWithExceptionHandling(logger, "find key with ID $keyId for $tenantId") {
             cryptoOpsClient.lookupKeysByIds(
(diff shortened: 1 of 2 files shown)


dimosr (dismissed):
LGTM

components/membership/membership-rest-impl/src/test/kotlin/net/corda/membership/impl/rest/v1/CertificatesRestResourceImplTest.kt
```diff
@@ -597,11 +609,40 @@ class CertificatesRestResourceImplTest {
...
+
+        private fun registerVirtualNodeForTenantId(holdingIdentityShortHash: String) {
+            val nodeHoldingIdentity = mock<HoldingIdentity> {
+                on { x500Name } doReturn MemberX500Name.parse("O=Alice, L=LDN, C=GB")
+            }
+            val nodeInfo = mock<VirtualNodeInfo> {
+                on { holdingIdentity } doReturn nodeHoldingIdentity
+            }
+            whenever(
+                virtualNodeInfoReadService.getByHoldingIdentityShortHash(
```
yift-r3: Can you add this to the definition of `virtualNodeInfoReadService` and remove this function?

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/CertificatesRestResourceImpl.kt
```diff
@@ -442,20 +442,17 @@ class CertificatesRestResourceImpl @Activate constructor(
...
     private fun validateTenantId(tenantId: String) {
-        val isValidShortHash = try {
+        if (tenantId in allClusterTenants) return
+
+        try {
             ShortHash.parse(tenantId)
-            true
         } catch (e: ShortHashException) {
-            false
+            throw InvalidInputDataException("Provided tenantId $tenantId is not a cluster tenant or " +
```
kyriathar: I believe this message needs updating to only mention `tenantId` which is of type `ShortHash` is not a valid  holding id (Should be shorter and hopefully more clear).

Or if you want to keep the first part then "or" needs to be an "and" I think.

kyriathar: Actually on second look, message is correct, so please disregard the above. Perhaps making shorter to only mention not valid short id might be a bit more clear.

anien: `Provided tenantId $tenantId is not a valid holding identity ID.` would be fine with you?

components/membership/membership-rest-impl/src/test/kotlin/net/corda/membership/impl/rest/v1/CertificatesRestResourceImplTest.kt
anien: I don't think I can do that. I need to parametrized the function and it won't be possible in the definition of `virtualNodeInfoReadService`. I can potentially move some bits to the definition (`nodeHoldingIdentity` and `nodeInfo`), and then just run `virtualNodeInfoReadService.getByHoldingIdentityShortHash` inline in the tests.

yift-r3: You always call it with the same argument (`holdingIdentityShortHash`), so we can move its definition to the parent class and add the mocking to the definition of `virtualNodeInfoReadService`.  If you want it to fail in some tests, you can define a second holding identity...

anien: Okay, managed to get it working completely in the definition. Thanks for the suggestion!

yift-r3 (approved):

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/CertificatesRestResourceImpl.kt
```diff
@@ -442,20 +442,16 @@ class CertificatesRestResourceImpl @Activate constructor(
...
     private fun validateTenantId(tenantId: String) {
-        val isValidShortHash = try {
+        if (tenantId in allClusterTenants) return
+
+        try {
             ShortHash.parse(tenantId)
-            true
         } catch (e: ShortHashException) {
-            false
+            throw InvalidInputDataException("Provided tenantId $tenantId is not a valid holding identity ID.")
```
kyriathar, line 450: Here, we swallow the actual issue with the `ShortHash` parsing which should be described in `ShortHashException`. I know did the same before but thought I'd flag it anyway.

anien: Yes, I noticed that. But this is all facing REST calls, where `InvalidInputDataException` is `HttpApiException` compare to `ShortHashException` which is not. So in this case it makes sense to swallow the non http api `ShortHashException` and throw `HttpApiException`.

kyriathar: What I had in my mind was something like: 
`throw InvalidInputDataException("this exception's message",  shortHashException)`

But then I saw that `InvalidInputDataException` API won't let you nest exceptions. Maybe the message of the `ShortHashException` could be added to the `InvalidInputDataException.details` which seems to be there for that reason, to give a bit more info as to why the short hash was not valid.

PR #4297, opened 2023-07-13
CORE-13377: Move `loginName` to path param for GET User API

vkolomeyko:
I would recommend to start review with `UserEndpoint.kt` file to get an idea what sort of change we are making to REST resource.

With Smoke/E2E tests passing, I have also done a round of Dev testing locally with Combined Worker.
It is still possible to use `v1` version of the API, however request serviced will contain a `Warning` header like so:
image

Newer version of API will not contain such a warning:
image

But in both cases requests serviced successfully.


BenYip123 (approved):

Tom-Fitzpatrick (approved):
LGTM

PR #4298, opened 2023-07-13
CORE-14186 - Message patterns do not close old consumer and producers before recreating new ones

pmck91:
adding close to the consumer and producer in EventLogSubscriptionImpl.


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/EventLogSubscriptionImpl.kt
```diff
@@ -107,6 +107,9 @@ internal class EventLogSubscriptionImpl<K : Any, V : Any>(
                     ForwardingRebalanceListener(config.topic, config.clientId, partitionAssignmentListener)
                 }
                 val consumerConfig = ConsumerConfig(config.group, config.clientId, ConsumerRoles.EVENT_LOG)
+
+                consumer?.close()
+
```
chrisr3: Why do you need this? Both `CordaConsumer` and `CordaProducer` extend `AutoCloseable` and so `consumer` and `producer` are _guaranteed_ to be closed when their respective `use {}` blocks exit.

FYI `use` is Kotlin's equivalent of Java's `try-with-resources`, so it looks to me like you're now closing both of these twice.

Actually, I'm wondering if you need either of
```kotlin
var consumer: CordaConsumer<K, V>?
var producer: CordaProducer?
```
in the first place:
```kotlin
val consumerConfig = ConsumerConfig(config.group, config.clientId, ConsumerRoles.EVENT_LOG)
cordaConsumerBuilder.createConsumer(
    // etc...
).use { cordaConsumer ->
    cordaConsumer.subscribe(config.topic)
    val producerConfig = ProducerConfig(config.clientId, config.instanceId, true, ProducerRoles.EVENT_LOG, false)
    cordaProducerBuilder.createProducer(producerConfig, config.messageBusConfig) { data ->
        log.warn("Failed to serialize record from ${config.topic}")
        deadLetterRecords.add(data)
    }.use { cordaProducer ->
        threadLooper.updateLifecycleStatus(LifecycleStatus.UP)
        pollAndProcessRecords(cordaConsumer, cordaProducer)
    }
}
```

LWogan: I didn't notice we are already using `use` blocks for producers as well, which can be seen on line 134 of the new pr. This means producers should close gracefully when theres exceptions. This makes the issue seen more confusing.

This JIRA was raised after some strange behavior in some performance tests. We were seeing old producers linger around and cause exceptions when new producers were created.

```
net.corda.messaging.api.exception.CordaMessageAPIFatalException: FatalError occurred aborting transaction for CordaKafkaProducer with clientId DURABLE-net.corda.messaging.api.exception.CordaMessageAPIFatalException: FatalError occurred aborting transaction for CordaKafkaProducer with clientId DURABLE-DURABLE.Group-flow.event-d1404c80-fc94-4e1f-b2f0-6594c928dcf1
....
...
Caused by: org.apache.kafka.common.errors.ProducerFencedException: There is a newer producer with the same transactionalId which fences the current one.DURABLE.Group-flow.event-d1404c80-fc94-4e1f-b2f0-6594c928dcf1
....
...
Caused by: org.apache.kafka.common.errors.ProducerFencedException: There is a newer producer with the same transactionalId which fences the current one.
```

The new producer was being instantiated as a result of some errors from kafka after 2 of the 3 brokers crashed. 

```
16:40:53.993 [kafka-producer-network-thread | producer-DURABLE-DURABLE.Group-flow.event-d1404c80-fc94-4e1f-b2f0-6594c928dcf1] WARN  org.apache.kafka.clients.producer.internals.Sender {} - [Producer clientId=producer-DURABLE-DURABLE.Group-flow.event-d1404c80-fc94-4e1f-b2f0-6594c928dcf1, transactionalId=DURABLE-DURABLE.Group-flow.event-d1404c80-fc94-4e1f-b2f0-6594c928dcf1-1286414107] Got error produce response with correlation id 58205 on topic-partition lorcan-p2p.out-19, retrying (2147483646 attempts left). Error: NETWORK_EXCEPTION. Error Message: Disconnected from node 1
```

I wonder is it possible the close() for the producers failed due some kafka brokers being down at the time. Then when kafka came back up and the new producer was created, kafka still had reference to the old one and through the fenced exception.

In that case adding an additional close before creating new ones might be an additional safety net.

additional detail here:https://r3holdco.slack.com/archives/C039F5SAFNH/p1685092738091709

chrisr3: Kotlin's `use {}` is equivalent to:
```java
val consumer = createConsumer()
try {
    // do stuff
} finally {
    consumer.close()
}
```
Using a `try...finally` is the best "safety net" there is, so I don't see how adding an extra spurious `close()` can add anything except confusion.

The only way that an exception can leave either `consumer` or `producer` unclosed is if it occurs _outside_ their respective `use` blocks, e.g. an exception from anything here would currently leave `consumer` open:
```kotlin
val producerConfig = ProducerConfig(config.clientId, config.instanceId, true, ProducerRoles.EVENT_LOG, false)
producer = cordaProducerBuilder.createProducer(producerConfig, config.messageBusConfig) { data ->
    log.warn("Failed to serialize record from ${config.topic}")
    deadLetterRecords.add(data)
}
```
Moving it _inside_ `consumer.use {}` would fix that.

---
Edit: I thought my earlier suggestion was just a "tidy up", but actually it's also fixing a bug where `createProducer()` could throw an exception and leave the `consumer` open...

pmck91: So after a bit of replicating and testing some stuff out, I was able to replicate the fenced exception after forcing the close to fail. @chrisr3 your suggestion seems to fix the issue, I wasn't able to replicate it with your changes. But I'm not sure why? I found this article on how `use` can go wrong if yield is used, and its possible that nested `use` blocks like this may run into similar issues 

https://betterprogramming.pub/breaking-try-catch-finally-in-kotlin-b95059a5673f

chrisr3 (approved):
LGTM

LWogan (approved):
Happy to approve this as this change has positive benefits. I think the issue I saw in the performance test will be very hard to reproduce under the exact same conditions (kafka running out of resources). I don't have any doubts that the use call close. I'm still a little worried the implementation of the producer.close() can fail to execute properly but we have logging in place to capture that so it should be debugable if we see it again. Additionally the POD would restart anyway so the system would recover.

PR #4299, opened 2023-07-13
CORE-15351: log trace span to separate log file

ZawarR3:
When deploying Corda with the combined worker, if the following command line argument is provided: `  --send-trace-to=http://localhost:9411 `, the trace data (trace spans) is sent over to the ZipKin Server, as a valid host URL has been provided, however if no URL is provided the tracing data is not generated nor logged anywhere. With this change, when no ZipKin host URL is provided the trace spans will be logged locally into a file called `corda-trace-spans.log`. Developers have the flexibility to configure logging behaviour through the log4j2.xml file. They can choose to enable or disable this local logging with or without ZipKin and can also opt to direct logs to the console instead of log files if they prefer.


libs/rest/rest-server-impl/build.gradle
```diff
@@ -17,7 +17,8 @@ dependencies {
     implementation project(':libs:rest:rest-security-read')
     implementation project(':libs:rest:json-serialization')
     implementation project(':libs:metrics')
-    implementation project(':libs:tracing')
+    implementation project(':libs:' +
+            'tracing')
```
josephzunigadaly: This looks like an accidental edit. Please can we undo this change?

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
```diff
@@ -70,20 +70,26 @@ internal class BraveTracingService(serviceName: String, zipkinHost: String, samp
...
+        if (zipkinHost != null){
 
-        val zipkinUrl = "$zipkinHost/api/v2/spans"
-        val spanAsyncReporter =
-            AsyncReporter.create(URLConnectionSender.create(zipkinUrl))
-                .also(resourcesToClose::push)
-        reporters.add(spanAsyncReporter)
+            // The console reporter is useful when debugging test runs on the combined worker.
+            // uncomment it to enable it.
+            val reporters = mutableListOf<Reporter<Span>>(Reporter.CONSOLE)
```
josephzunigadaly: `Reporter.CONSOLE` should still be commented out

```suggestion
            val reporters = mutableListOf<Reporter<Span>>(/*Reporter.CONSOLE*/)
```

osgi-framework-bootstrap/src/main/resources/log4j2.xml
```diff
@@ -60,6 +84,10 @@
             <AppenderRef ref="Ledger-Detailed"/>
         </logger>
 
+        <logger name="brave.Tracer" level="info">
```
josephzunigadaly: I think the default level should be "off" for this so that it can be enabled when wanted but normally is switched off.

osgi-framework-bootstrap/src/main/resources/log4j2.xml
```diff
@@ -50,6 +50,30 @@
                 </Delete>
             </DefaultRolloverStrategy>
         </RollingFile>
+
+        <RollingFile name="Trace-Spans"
+                     fileName="logs/corda-trace-spans.log"
+                     filePattern="logs/corda-trace-spans.%d{MM-dd-yyyy}.%i.log"
```
josephzunigadaly: I think `yyyy-MM-dd` is a better date format because it is always sortable.

osgi-framework-bootstrap/src/main/resources/log4j2.xml
owenstanford: We should exclude `brave.tracer` from the other loggers so we only get the trace output in the trace log file.

ZawarR3: I agree, currently working on it.

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
```diff
@@ -70,27 +71,38 @@ internal class BraveTracingService(serviceName: String, zipkinHost: String?, sam
...
+        tracingBuilder.addSpanHandler(spanHandler)
+        tracingBuilder.build().also(resourcesToClose::push)
+    }
+
+    private class LogReporter : Reporter<Span> {
+        private val logger: Logger = Logger.getLogger(LogReporter::class.java.name)
 
-            tracingBuilder.addSpanHandler(spanHandler)
+        override fun report(span: Span) {
+                logger.info(span.toString())
```
owenstanford, line 98: We should check to see if logging is enabled before logging to save the cost of the  span.toString()
```if (!logger.isLoggable(Level.INFO)) return ;```

**Review by owenstanford — COMMENTED** [review]
We should be able to delete the `NoopTracingService` class now.

File: libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt (hunk 5 above)
Comment (ZawarR3) [reply libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt:98]: is this okay: `       
 override fun report(span: Span) {
            if (logger.isLoggable(Level.INFO)) {
                logger.info(span.toString())
            }
        }`

Comment (owenstanford) [reply libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt:98]: yes

File: osgi-framework-bootstrap/src/main/resources/log4j2.xml [hunk 6]
```diff
@@ -10,6 +10,9 @@
                      filePattern="logs/corda.%d{yyyy-MM-dd}.%i.log"
                      ignoreExceptions="false">
             <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %-5level %logger{36} %X - %msg%n"/>
+            <Filters>
+                <RegexFilter regex=".*brave\.Tracer.*" onMatch="DENY" onMismatch="NEUTRAL"/>
+            </Filters>
```
ZawarR3: I forgot to remove this filter, it's not needed. I'll remove it on the next commit.

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
```diff
@@ -22,16 +22,18 @@ import zipkin2.reporter.AsyncReporter
 import zipkin2.reporter.Reporter
 import zipkin2.reporter.brave.ZipkinSpanHandler
 import zipkin2.reporter.urlconnection.URLConnectionSender
-import java.util.Stack
+import java.util.*
```
simon-johnson-r3: I'm not sure we allow wildcard imports, although admittedly the Kotlin standard don't explicitly prevent it.

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
```diff
@@ -70,23 +72,42 @@ internal class BraveTracingService(serviceName: String, zipkinHost: String, samp
...
-        val zipkinUrl = "$zipkinHost/api/v2/spans"
-        val spanAsyncReporter =
-            AsyncReporter.create(URLConnectionSender.create(zipkinUrl))
-                .also(resourcesToClose::push)
-        reporters.add(spanAsyncReporter)
+        val reporters = mutableListOf<Reporter<Span>>()
 
-        val spanHandler = ZipkinSpanHandler.create(reporter)
+        //Establish zipkin connection iff url host is provided and create respective reporter
+        if (zipkinHost != null){
```
simon-johnson-r3: there should be a space between ) and { here, I would run this through the intellij formatter to save you some time and prevent these kinds of formatting comments in PRs generally

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
```diff
@@ -70,23 +72,42 @@ internal class BraveTracingService(serviceName: String, zipkinHost: String, samp
...
+            val zipkinUrl = "$zipkinHost/api/v2/spans"
+            val spanAsyncReporter =
+                AsyncReporter.create(URLConnectionSender.create(zipkinUrl))
+                    .also(resourcesToClose::push)
+            reporters.add(spanAsyncReporter)
+        }
+
+        // LogReporter will report trace spans to local log files
+        val logReporter = LogReporter()
```
simon-johnson-r3: Kotlin can be quite a terse language and generally developers try to embrace that. Here I would suggest
```
reporters.add(LogReporter())
```
It's not a competition to reduce the number of code lines, but removal of variables which are only used once and which call a trivial constructor illustrates to the reader this is the only point of use, and it's one thing Kotlin is quite good at promoting IMHO

josephzunigadaly (dismissed):

simon-johnson-r3 (changes requested):

simon-johnson-r3 (dismissed):

simon-johnson-r3 (approved):

PR #4301, opened 2023-07-13
CORE-15260: Handle RestApiVersions in ResourceNameConflictValidator when resource path conflicts occur

Tom-Fitzpatrick:
This relaxes the requirements in ResourceNameConflictValidator to allow resources with the same path if they have different, non overlapping versions.


vkolomeyko (approved):
LGTM

PR #4302, opened 2023-07-14
CORE-14685: Extend DSL to support changing virtual node status.

chrisr3:
Extend the driver's simulated network and DSL so that virtual nodes can be `SUSPENDED` or `PENDING` as well as `ACTIVE`.

### DIFF
--- a/testing/driver/driver-engine/src/main/kotlin/net/corda/testing/driver/sandbox/Constants.kt
+++ b/testing/driver/driver-engine/src/main/kotlin/net/corda/testing/driver/sandbox/Constants.kt
@@ -22,18 +22,3 @@ const val DEFAULT_KEY_SCHEME = ECDSA_SECP256R1_CODE_NAME
 const val WRAPPING_KEY_ALIAS = "master"
 
 const val WAIT_MILLIS = 100L
-
-inline fun <T> MutableCollection<T>.extractAllTo(
-    destination: MutableCollection<T>,
-    predicate: (T) -> Boolean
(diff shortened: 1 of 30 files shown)


testing/driver/driver-engine/src/main/kotlin/net/corda/testing/driver/sandbox/Constants.kt
```diff
@@ -22,3 +22,18 @@ const val DEFAULT_KEY_SCHEME = ECDSA_SECP256R1_CODE_NAME
...
+    val iterator = iterator()
+    while (iterator.hasNext()) {
+        val item = iterator.next()
+        if (predicate(item)) {
+            destination += item
+            iterator.remove()
+        }
+    }
+    return destination
+}
```
yarivus: This aux function doesn't seem related to the Constants.kt file ... maybe it should bne moved to another file ?

yarivus (commented):
See comment

testing/driver/driver-engine/src/main/kotlin/net/corda/testing/driver/sandbox/Constants.kt
chrisr3: Maybe I could also rename the file to something more inclusive than "Constants"? Although personally speaking, I _hate_ "Utils"-like classes because they just end up becoming "buckets of stuff".

Need to think about this... :thinking:

yarivus (commented):
thx @chrisr3

PR #4304, opened 2023-07-14
CORE-14775 Preventing DBCordaConsumerImpl from skipping over pending records

ben-millar:
## Overview
After investigation, the root cause of the test flakiness described in CORE-14775 was found to be a bug in `DBCordaConsumerImpl` which allowed for records in `TransactionState.PENDING` to be skipped over if they were followed by records with the state `TransactionState.COMMITTED`. This scenario can occur when one transaction is still in progress but a subsequent transaction has already been completed.

## Fix
To prevent this, the logic in the `DBCordaConsumerImpl` was modified such that:
1. We cannot read beyond the last `PENDING` record.
2. Offsets must be read sequentially.

## Testing
Following the changes, I ran the previously flaky `Flow Session - Initiate multiple sessions and exercise the flow messaging apis()` test locally on the combined worker and verified that it successfully passed 50 times in a row (where previously it had been failing every 2-3 runs, as seen also on Jenkins).

I also added unit tests to cover the previously failing scenario, and refactored the existing unit tests to tighten up their logic against the correct behaviour.

### DIFF
--- a/libs/messaging/db-message-bus-impl/src/test/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImplTest.kt
+++ b/libs/messaging/db-message-bus-impl/src/test/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImplTest.kt
@@ -28,7 +28,7 @@ import java.time.Instant
 internal class DBCordaConsumerImplTest {
 
     companion object {
-        private const val topic = "topic"
+        private const val defaultTopic = "topic"
         private val defaultConfig = ResolvedConsumerConfig(
             "group",
             "client",
@@ -46,9 +46,9 @@ internal class DBCordaConsumerImplTest {
         // Null inputs to deserializer are converted to "" by the @NotNull decorator
         private val nullValue = "".toByteArray()
 
-        private val partition0 = CordaTopicPartition(topic, 0)
-        private val partition1 = CordaTopicPartition(topic, 1)
-        private val partition2 = CordaTopicPartition(topic, 2)
+        private val partition0 = CordaTopicPartition(defaultTopic, 0)
+        private val partition1 = CordaTopicPartition(defaultTopic, 1)
+        private val partition2 = CordaTopicPartition(defaultTopic, 2)
     }
 
     @Mock
@@ -75,6 +75,32 @@ internal class DBCordaConsumerImplTest {
         )
     }
 
+    private fun getTopicRecords(
+        topic: String = defaultTopic,
+        partition: CordaTopicPartition = partition0,
+        startOffset: Long = 0L,
+        key: ByteArray = serializedKey,
+        value: ByteArray? = serializedValue,
+        header: String = serializedHeader,
+        txState: TransactionState = TransactionState.COMMITTED,
+        count: Int = 1): List<TopicRecordEntry>
+    {
+        val timestamp = Instant.parse("2022-01-01T00:00:00.00Z")
+
+        return (0 until count).mapIndexed { index, _ ->
+            TopicRecordEntry(
+                topic,
+                partition.partition,
+                startOffset + index,
+                key,
+                value,
+                header,
+                TransactionRecordEntry("id", txState),
+                timestamp
+            )
+        }
+    }
+
     @Test
     fun `consumer correctly assigns topic partitions`() {
         val consumer = makeConsumer()
@@ -92,7 +118,7 @@ internal class DBCordaConsumerImplTest {
         consumer.assign(topicPartitions)
 
         assertThatExceptionOfType(CordaMessageAPIFatalException::class.java).isThrownBy {
-            consumer.subscribe(topic, null)
+            consumer.subscribe(defaultTopic, null)
         }
     }
 
@@ -101,7 +127,7 @@ internal class DBCordaConsumerImplTest {
         val consumer = makeConsumer()
         val topicPartitions = setOf(partition0, partition1, partition2)
 
-        consumer.subscribe(topic, null)
+        consumer.subscribe(defaultTopic, null)
 
         assertThatExceptionOfType(CordaMessageAPIFatalException::class.java).isThrownBy {
             consumer.assign(topicPartitions)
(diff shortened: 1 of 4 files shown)


libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
```diff
@@ -60,6 +60,23 @@ class CordaKafkaConsumerImpl<K : Any, V : Any>(
 
     private companion object {
         private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
+
```
ben-millar, line 63: This is a non-functional change to separate the config (list of exceptions) from the logic to simplify the `poll()` function. We also now avoid the linear check through each exception type and replace it with a constant time set membership check.

Note that `is` performs additional sub-typing checks which our `class.java` equality check does not, but those checks are not required as we're comparing literal types.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
```diff
@@ -78,27 +95,16 @@ class CordaKafkaConsumerImpl<K : Any, V : Any>(
...
-                    is KafkaException,
-                    -> {
+                    in transientExceptions -> {
                         logWarningAndThrowIntermittentException("Error attempting to poll.", ex)
                     }
-
-                    else -> logErrorAndThrowFatalException("Unexpected error attempting to poll.", ex)
+                    else -> {
+                        logErrorAndThrowFatalException("Unexpected error attempting to poll.", ex)
+                    }
```
ben-millar: This is a non-functional change to separate the config (list of exceptions) from the logic to simplify the `poll()` function. We also now avoid the linear check through each exception type and replace it with a constant time set membership check.

Note that `is` performs additional sub-typing checks which our `class.java` equality check does not, but those checks are not required as we're comparing literal types.

LWogan: This should be done in a different pr, and applied to all the functions that have this pattern, not just the close func.

libs/messaging/db-message-bus-impl/src/test/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImplTest.kt
```diff
@@ -289,60 +212,68 @@ internal class DBCordaConsumerImplTest {
         val fromOffset = ArgumentCaptor.forClass(Long::class.java)
         val timestamp = Instant.parse("2022-01-01T00:00:00.00Z")
 
-        val transactionRecord1 = TransactionRecordEntry("id", TransactionState.COMMITTED)
-        val transactionRecord2 = TransactionRecordEntry("id2", TransactionState.PENDING)
+        val committedRecords = getTopicRecords(
+            startOffset = 0,
+            txState = TransactionState.COMMITTED,
```
LWogan, line 217: Need an additional test case I think to handle the new filtering logic to specifically exclude ABORTED transaction records.

LWogan (commented):
Mostly looks good! Couple small comments.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
ben-millar: Makes sense. I figured I'd tidy as I went rather than opening a ticket for it, but I'll revert this commit to keep it clean

libs/messaging/db-message-bus-impl/src/test/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImplTest.kt
```diff
@@ -147,137 +160,47 @@ internal class DBCordaConsumerImplTest {
...
-                transactionRecord1,
-                timestamp
-            ),
-        )
+        val pollResult = getTopicRecords(partition = partition0, startOffset = 0, count = 4)
 
+        val fromOffset = ArgumentCaptor.forClass(Long::class.java)
         whenever(dbAccess.getMaxCommittedPositions(any(), any())).thenAnswer { mapOf(partition0 to 0L) }
-        whenever(dbAccess.getLatestRecordOffsets()).thenAnswer { mapOf(partition0 to 7L) }
+        whenever(dbAccess.getLatestRecordOffsets()).thenAnswer { mapOf(partition0 to 4L) }
```
emilybowe, line 167: so before, it was `partition0 to 7L` and the assert was `assertThat(consumer.position(partition0)).isEqualTo(8)` so 0, 1, 2, 3, 4, 5, 6, 7 = 8
and now it's `partition0 to 4L` and the assert is `assertThat(consumer.position(partition0)).isEqualTo(4)` so 0, 1, 2, 3, 4, I would  have expected = 5 ? not sure why it's 4 and not 5, maybe some other logic changed I don't understand here.

ben-millar: I think you're misreading it as a sequence, so `partition0 to 7L` isn't generating elements 0 through 7, it's mapping the `0th` partition to `7L` so that when we call `dbAccess.getLatestRecordOffsets()`, it'll return an offset of 7 for the `0th` partition.

This original logic is a little odd, though, because it inserts 4 records with the offsets 0, 2, 5 and 7. Today, the logic is such that our offsets need to be read sequentially, so we're using the offsets 0, 1, 2 and 3 – records with these offsets are returned from our function call:

```
getTopicRecords(partition = partition0, startOffset = 0, count = 4)
```

So once we've polled records 0 through 3, our next expected offset is 4 (Or 4L; L just means it's a long, so we can have like 9 quintillion offsets... just in case).

Really good question tho, hopefully the answer was interesting/useful.

LWogan (approved):
LGTM

thiagoviana (approved):
LGTM.

PR #4307, opened 2023-07-14
CORE-14235 Fix `findUnconsumedVisibleStatesByType` mapping

lankydan:
Correctly map output infos to states, similar to fixes already applied to `resolveStateRefs`.


relyafi (approved):
LGTM

blsemo (approved):
LGTM

driessamyn (approved):
It'd be good to see a unit/integration test that replicates the issue and validates the fix, but I'm happy for that to be in a follow-up PR.

PR #4308, opened 2023-07-14
CORE-13943: Change deploy script to make it work

yift-r3:
Some changes to the deploy scripts:
1. Add support for metrics toleration in the simulator chart.
2. Increase limits to more reasonable ones.
3. Pass different arguments (like number of instances, number of partitions...) as environment variables.
4. Make sure the MGM deployment is scaled down.
5. Save all temporary files into a `.gitignore` directory.
6. Allow to run the script from any location.
7. Read the version from the Gradle property file.
8. Allow running the sender with a custom Yaml file.


applications/tools/p2p-test/app-simulator/charts/app-simulator/templates/app-simulator-sender-job.yaml
```diff
@@ -62,6 +62,14 @@ spec:
       {{ include "appSimulator.kafkaSaslInitContainer" . | nindent 8 }}        
       volumes:
       {{- include "appSimulator.volumes" . | nindent 8 }}
+      {{- with .Values.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      tolerations:
```
YashNabar, line 70: Should we use the value from values.yaml here? Also in the other templates

applications/tools/p2p-test/app-simulator/charts/app-simulator-db/values.yaml
```diff
@@ -20,6 +20,12 @@ postgresql:
     password:
   # primary (read/write instance) configuration.
   primary:
+    nodeSelector:
+      node: "prereqs"
+    tolerations:
+      - effect: "NoSchedule"
+        key: "node"
+        value: "prereqs"
```
dimosr: Isn't this specific to the environment the chart is deployed into? If so, we should probably not include it in the defaults here. 

Maybe we should have a separate file that contains environment-specific settings (like this one) and add it separately? I can see we have something similar for `prereqs-eks.metrics.yaml`

Similar for the app-simulator values file.

dimosr: While continuing the review, I noticed that we actually have `app-simulator-eks.metrics.yaml`. So, maybe what I described above is already in place and this is a leftover that can be removed?

applications/tools/p2p-test/app-simulator/scripts/corda-eks.yaml
```diff
@@ -1,7 +1,7 @@
...
   requests:
-    memory: "620Mi"
-    cpu: "500m"
+    memory: "8192Mi"
+    cpu: "4000m"
   limits:
-    memory: "1250Mi"
-    cpu: "1000m"
+    memory: "8192Mi"
+    cpu: "4000m"
```
dimosr: I think people are using these scripts for everyday testing as well right? In that case, it might be better to keep resources reasonably low by default and let people override them only when needed for performance testing. 

Similar for prereqs-eks.yaml.

applications/tools/p2p-test/app-simulator/scripts/deploy.sh
```diff
@@ -35,16 +39,17 @@ deploy() {
...
 }
 
+helm registry login corda-os-docker.software.r3.com -u $CORDA_ARTIFACTORY_USERNAME -p $CORDA_ARTIFACTORY_PASSWORD
+helm registry login corda-os-docker-unstable.software.r3.com -u $CORDA_ARTIFACTORY_USERNAME -p $CORDA_ARTIFACTORY_PASSWORD
+
 if [ $# -eq 0 ]
 then
-  declare -a namespaces=($A_CLUSTER_NAMESPACE $B_CLUSTER_NAMESPACE $MGM_CLUSTER_NAMESPACE)
+  WORKER_REPLICAS=1 KAFKA_REPLICATION_FACTOR=1 KAFKA_REPLICAS=1 KAFKA_ZOOKEEPER_REPLICAS=1 KAFKA_PARTITION_COUNT=10 deploy $MGM_CLUSTER_NAMESPACE &
+  declare -a namespaces=($A_CLUSTER_NAMESPACE $B_CLUSTER_NAMESPACE)
```
dimosr, line 49: * I see these are set in `settings.sh` similar to other parameters. What is the reason for also having them here?
* I see that this code deploys the MGM cluster separately, instead of just including it in the `namespaces` variable and let the following loop deploy it. What was the reason behind this change?

yift-r3: Because the MGM cluster is only used to onboard the members it can have a smaller cluster.

dimosr: I see.. worth leaving a comment to explain the rationale.

applications/tools/p2p-test/app-simulator/scripts/runScenario.sh
```diff
@@ -0,0 +1,139 @@
...
+echo "Warm up"
+run_sender
+
+totalNumberOfMessages=60000
+interBatchDelay="PT0.3S"
+batchSize=40
+stop="no"
+latency="0.22"
+until (( $(echo "$latency > 1.0" |bc -l) ));  do
+  echo 'Waiting a minutes before starting sender'
```
dimosr:
```suggestion
  echo 'Waiting a minute before starting sender'
```

applications/tools/p2p-test/app-simulator/scripts/runScenario.sh
```diff
@@ -0,0 +1,139 @@
...
+  stop="no"
+  until [[ "$stop" == "yes" ]];  do
+    echo 'Waiting...'
+    sleep 1
+    echo 'Checking how many messages had been sent...'
+    sent=$(kubectl exec -n $NAMESPACE_PREFIX-db db-postgresql-0 -- env PGPASSWORD=$dbPassword psql -U postgres -d app_simulator -c 'select COUNT(*) from sent_messages;' -t 2>/dev/null | xargs)
+    received=$(kubectl exec -n $NAMESPACE_PREFIX-db db-postgresql-0 -- env PGPASSWORD=$dbPassword psql -U postgres -d app_simulator -c 'select COUNT(*) from received_messages;' -t  2>/dev/null | xargs)
+
+    echo "Sent [$sent] messages and received [$received] messages"
+    if [[ "$sent" == "$received" ]]; then
```
dimosr: In order to be fully correct, should we also check `$sent == $totalNumberOfMessages` here? Otherwise, there's a chance (albeit low) that the two numbers might temporarily match.

applications/tools/p2p-test/app-simulator/scripts/runScenario.sh
```diff
@@ -0,0 +1,139 @@
+#!/bin/bash
```
dimosr: Should we create a folder under `scripts` with the name `horizontal-scaling-testing` (or something similar) and put the script and the scenario files there? 

It would make things a bit easier to navigate as we start adding more scripts.

applications/tools/p2p-test/app-simulator/scripts/prereqs-eks.yaml
```diff
@@ -1,18 +1,20 @@
 kafka:
   resources:
     requests:
-      memory: 780Mi
-      cpu: 1000m
+      memory: 8192Mi
```
dimosr: this seems to still be used by `deploy`. I think we need to do something similar with `corda-eks-large` / `corda-eks-small`.

dimosr (commented):
Left a few more comments and resolved the previous ones that have been resolved. But, there seem to be some unaddressed. 

For example see [here](https://github.com/corda/corda-runtime-os/pull/4308#discussion_r1263888634) and [here](https://github.com/corda/corda-runtime-os/pull/4308#discussion_r1264994823). I think they are talking about the same thing, i.e. using tolerations from a values file in the helm charts instead of hardcoding the values and extracting both tolerations and node selector values into a separate values file instead of having them in the default one, since they shouldn't apply to all environments by default.

applications/tools/p2p-test/app-simulator/scripts/runScenario.sh
yift-r3: I think the sender will commit all the messages to the sent table in the database, so the number in `$sent` should be correct when the job is done. It's different from `totalNumberOfMessages` because it also includes the previous messages.

dimosr: yes, you're right. While writing this, I was thinking that the job will terminate early and let the load generator run, but in the default mode (`ONE_OFF`) I believe it waits until all the messages have been sent. So, there shouldn't be a problem here.

applications/tools/p2p-test/app-simulator/charts/app-simulator/templates/app-simulator-sender-job.yaml
yift-r3: I don't know helm good enough to do it.

applications/tools/p2p-test/app-simulator/charts/app-simulator/values.yaml
```diff
@@ -30,7 +30,11 @@ resources:
   requests: {}
 
 # -- node labels for pod assignment, see https://kubernetes.io/docs/user-guide/node-selection/
-nodeSelector: {}
+nodeSelector: { node: "prereqs" }
```
davidcurrie: nit: although allowable, generally it's better not to mix YAML and JSON (the exception is the use of `{}` to indicate that it's expecting an object here which you can't convey in YAML)
```suggestion
nodeSelector:
  node: "prereqs"
```

applications/tools/p2p-test/app-simulator/charts/app-simulator/templates/app-simulator-sender-job.yaml
dimosr: That's a great opportunity to learn more about it then :) 

I am sure @davidcurrie would be happy to give any tips, if needed.

dimosr (approved):
LGTM

PR #4309, opened 2023-07-14
CORE-15553 Clean up stale inbound sessions

charlieR3:
Investigations showed that e2e tests which failed when testing multi cluster suspension scenarios caused downstream failures by creating an unhealthy link manager worker. On closer inspection, it was observed that a suspended member was creating sessions with another party and these inbound sessions on the non-suspended member’s side were never cleared which eventually led to an out of memory error.
This PR tracks inbound sessions and removes them if the timeout similar to outbound sessions. This also removes pending sessions from the pending sessions list when they become active.

The below shows before and after metrics from running a single test (`onboard, suspend, fail to communicate, activate member and communicate` from `MembershipSuspensionMultiCluster`) which failed after suspending a member on cluster-b which had an open session with a member on cluster-a. These metrics are from cluster-a in this scenario.

**Before Fix:**
Screenshot 2023-07-17 at 16 46 37
Screenshot 2023-07-17 at 16 46 46

**After Fix:**
Screenshot 2023-07-17 at 17 16 25
Screenshot 2023-07-17 at 17 16 16


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -855,6 +862,7 @@ internal class SessionManagerImpl(
             SessionManager.Counterparties(ourIdentityInfo.holdingIdentity, peer.holdingIdentity),
             session.getSession()
         )
+        inboundSessionEstablished(message.header.sessionId)
```
dimosr: I had a vague memory that we were intentionally avoiding considering the session active at this point, because we can't be sure the session message will be delivered on the other side. I can now see there's a comment a few lines below for this.

Were you aware of this / did you consider any issues that might arise? For example, if we remove it from pending and responder handshake gets lost, the other side will replay and we might fail process any replays.

Not sure if it's a big problem, since the session might be cleaned up on the other side and a new one established. But, I wanted to make sure you're aware about this and put some thought into it.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -1009,20 +1019,29 @@ internal class SessionManagerImpl(
...
+         * [lastReceivedTimestamp]: The last time we received a message using this Session.
+         */
+        class TrackedInboundSession(
+            @Volatile
+            var lastReceivedTimestamp: Long,
+        )
+
         fun stopTrackingAllSessions() {
-            trackedSessions.clear()
+            trackedOutboundSessions.clear()
```
dimosr, line 1042: Looking at `applyNewConfiguration` where this is called from, it seems like we are clearing up pending/active inbound sessions too. So, we probably need to clean up the tracked inbound sessions here too?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -1009,20 +1019,29 @@ internal class SessionManagerImpl(
...
         }
 
         fun stopTrackingSpecifiedSession(sessionId: String) {
```
dimosr: Maybe rename to `stopTrackingSpecifiedOutboundSession` now?

P.S.: Seems like this is called from the code path that refreshes key (to effectively do key rotation), which ideally we'd do at the receiver side for inbound sessions as a guardrail too so that we don't rely on the sender. But, not worth doing as part of this change. I'll raise a separate ticket in our backlog.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -1074,22 +1093,43 @@ internal class SessionManagerImpl(
...
                     throw IllegalStateException("A message was acknowledged before the HeartbeatManager was started.")
                 }
-                val sessionInfo = trackedSessions[sessionId] ?: return@withLifecycleLock
+                val sessionInfo = trackedOutboundSessions[sessionId] ?: return@withLifecycleLock
                 logger.trace("Message acknowledged with on a session with Id $sessionId.")
                 sessionInfo.lastAckTimestamp = timeStamp()
             }
         }
 
+        fun dataMessageReceived(sessionId: String) {
```
dimosr, line 1120: Do we also need a `sessionMessageReceived` for inbound sessions and initialise the timer for expiry there (instead of on the first data message)?

Thinking of scenarios where a peer starts setting up a session and fails mid-way before the session is properly established (so no data messages arrive for that session).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -1074,22 +1093,43 @@ internal class SessionManagerImpl(
...
+            dominoTile.withLifecycleLock {
+                if (!isRunning) {
+                    throw IllegalStateException("A session message was received before the HeartbeatManager was started.")
```
dimosr:
```suggestion
                    throw IllegalStateException("A data message was received before the HeartbeatManager was started.")
```

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
dimosr: After moving on with the review and reaching the point where I commented about tracking session messages for inbound sessions too, I realised it might be better to skip doing this and keep it as before?

That way, any session that hasn't received the first data message will remain as pending (and will be cleaned up if no further session messages arrive). Once the first data message arrives, it will be converted to active and from that point on expiry will work based on data messages.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -1100,8 +1140,26 @@ internal class SessionManagerImpl(
             }
         }
 
+        private fun inboundSessionTimeout(sessionId: String) {
+            val sessionInfo = trackedInboundSessions[sessionId] ?: return
+            val timeSinceLastReceived = timeStamp() - sessionInfo.lastReceivedTimestamp
+            if (timeSinceLastReceived >= config.get().sessionTimeout.toMillis()) {
+                logger.info(
+                    "Inbound session $sessionId timed out due to inactivity and it will be cleaned up."
```
dimosr: I think we should make this log a bit less ambiguous e.g.
```
"Inbound session $sessionId has not received any messages for the configured timeout threshold (${config.get().sessionTimeout.toMillis()} ms) so it will be cleaned up."
```

And similarly for the equivalent message for outbound sessions.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
dimosr: FYI, created this one https://r3-cev.atlassian.net/browse/CORE-15578

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
charlieR3: Ah, I missed that comment lower down. I had seen the log message printing that inbound sessions were considered established at this point so I assumed the session could be treated as established from then on. No negative effects were observed from the test runs.

I can revert this to how it was before. Would it make sense to also track pending sessions and close them similar to how active inbound sessions are tidied up? Since this error scenario showed that data messages aren't going through due to the suspension, if I remove this line we will have the same issue as before. It will just take a longer to reach the out of memory exception since the active inbound sessions are timed out.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
charlieR3: Good point, thanks

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
dimosr: yes, it makes sense to have expiry logic for pending inbound sessions too. This is what I was touching on [in this comment](https://github.com/corda/corda-runtime-os/pull/4309#discussion_r1266489082). 

It would also be consistent/symmetric with outbound sessions, where tracking starts on the first session message.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -1074,22 +1099,52 @@ internal class SessionManagerImpl(
...
+        }
+
         private fun sessionTimeout(counterparties: SessionCounterparties, sessionId: String) {
-            val sessionInfo = trackedSessions[sessionId] ?: return
+            val sessionInfo = trackedOutboundSessions[sessionId] ?: return
             val timeSinceLastAck = timeStamp() - sessionInfo.lastAckTimestamp
             if (timeSinceLastAck >= config.get().sessionTimeout.toMillis()) {
                 logger.info(
                     "Outbound session $sessionId (local=${counterparties.ourId}, remote=${counterparties.counterpartyId}) timed " +
                         "out due to inactivity and it will be cleaned up."
```
dimosr: Can we also adjust this message similar to the other one in `inboundSessionTimeout` to replace `due to inactivity` for clarity?

dimosr (commented):
LGTM - just a comment on logs

dimosr (approved):

PR #4312, opened 2023-07-16
CORE-12001 - Enable support for running commands for specific REST endpoint and decouple from k8s

r3-arish:
## Desc

This PR adds the ability to run any command from the network plugin against an explicit endpoint, as specified by the user. This feature will provide more flexibility and control to the users over where and how they want to run the commands.

## Changes

1. Extended RestCommand in network plugin commands to inherit all the arguments including the endpoint.

2. Utilized RestClientUtils for handling REST client related operations.

3. Removed explicit/implicit dependencies of the plugin commands to k8s environments. Previously, some commands in the plugin were automatically discovering the REST endpoint and the password using kubectl. This behavior has been removed to ensure consistency with other plugins.

## Summary of changes

_AllowClientCertificate.kt_ file: The RestCommand class was added as a superclass, and the createRestClient method was used to create a REST client. The run method was modified to call the allowAndListCertificates method instead of performMembersLookup.

_BaseOnboard.kt_ file: This file was modified extensively, with changes to the setupClient, configureGateway, createTlsKeyIdNeeded, and run methods.

_MemberList.kt_ file: The RestCommand class was added as a superclass, and the performMembersLookup method was modified to use the createRestClient method to create a REST client.

_NetworkPluginWrapper.kt_ file: Minor change as the NetworkPlugin class was modified to include a space after the colon in the class definition.

_OnBoardMember.kt_ file: The RestCommand class was added as a superclass, and the performMembersLookup method was removed. The createCpi method was modified to use the createRestClient method to create a REST client. The run method was modified to call the setupClient, configureGateway, createTlsKeyIdNeeded, and onboardMember methods.

_OnboardMgm.kt_ file: The RestCommand class was added as a superclass, and the saveGroupPolicy method was modified to use the createRestClient method to create a REST client. The cpiFileChecksum property was modified to use the createRestClient method to create a REST client. The run method was modified to call the setupClient, configureGateway, createTlsKeyIdNeeded, and onboardMgm methods.

## Results

Screenshot 2023-07-13 at 23 02 23
Screenshot 2023-07-16 at 21 30 53
Screenshot 2023-07-16 at 21 32 09
Screenshot 2023-07-16 at 22 12 54
Screenshot 2023-07-13 at 23 01 31

## Edit

New Results for Member List

Screenshot 2023-07-25 at 13 07 58


tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/AllowClientCertificate.kt
```diff
@@ -2,78 +2,58 @@ package net.corda.cli.plugins.network
...
             client.start().also { connection ->
                 val mgm = connection.proxy
+
                 subjects.forEach { subject ->
                     println("\t Allowing $subject")
-                    mgm.mutualTlsAllowClientCertificate(
-                        mgmShortHash,
-                        subject,
-                    )
+                    mgm.mutualTlsAllowClientCertificate(mgmShortHash, subject)
```
YashNabar: Should we use `executeWithRetry()` from `RestClientUtils` for all these calls? Also in the other classes

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -118,51 +130,53 @@ abstract class BaseOnboard : Runnable {
...
-        val id = cpi.use { jarInputStream ->
+        createRestClient(CpiUploadRestResource::class).use { client ->
+            cpi.use { jarInputStream ->
+                client.start().also { connection ->
+                    val resource = connection.proxy
+                    val upload = HttpFileUpload(content = jarInputStream, fileName = "$name.cpi")
+                    val id = resource.cpi(upload).id
+
+                    // Wait for some time before checking the status of the CPI upload.
+                    Thread.sleep(2000)  // You can adjust this delay as per your requirements.
```
YashNabar: We should use `executeWithRetry()` from `RestClientUtils` here as well if we can instead of implementing our own retry logic

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -118,51 +130,53 @@ abstract class BaseOnboard : Runnable {
...
-            val status = try {
-                response.bodyOrThrow().`object`.get("status")
-            } catch (e: OnboardException) {
-                "Not ready yet"
-            }
-            if (status == "OK") {
-                return response.body.`object`.get("cpiFileChecksum").toString()
+            createRestClient(CpiUploadRestResource::class).use { client ->
+                client.start().also { connection ->
+                    val resource = connection.proxy
```
YashNabar: We can probably do something like this everywhere to avoid the single-use vals and `also{ }` blocks. You could refer to the existing usages of `createRestClient` and `executeWithRetry` for consistency -
```suggestion
                createRestClient(Example::class).use { client ->
                val endpoint = executeWithRetry(waitDuration, "Sample operation name") {
                    client.start().proxy
                }
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -140,17 +137,11 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
-                .bodyOrThrow()
-                .`object`
-                .get("cpis") as JSONArray
-            val exists = currentCpis.filterIsInstance<JSONObject>()
-                .mapNotNull {
-                    it.get("cpiFileChecksum") as? String
-                }.any {
-                    it == cpiFileChecksum
-                }
+            val restClient = createRestClient(CpiUploadRestResource::class)
```
YashNabar: Do we need this val?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -140,17 +137,11 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
+            val currentCpis = restClient.use { client ->
+                client.start().proxy.getAllCpis().cpis
+            }
+            val exists = currentCpis.any { it.cpiFileChecksum == cpiFileChecksum }
```
YashNabar:
```suggestion
            if (currentCpis.any { it.cpiFileChecksum == cpiFileChecksum })
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMgm.kt
```diff
@@ -110,7 +114,7 @@ class OnboardMgm : Runnable, BaseOnboard() {
 
     private val cpi by lazy {
         val parametersCpiFile = cpiFile
```
YashNabar: Is this needed?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -173,64 +173,58 @@ abstract class BaseOnboard : Runnable {
...
-                        .bodyOrThrow()
-                        .`object`
-                        .get("flowP2pOperationalStatus") == "ACTIVE"
-                ) {
-                    return@waitForVirtualNode
+        createRestClient(VirtualNodeRestResource::class).use { client ->
+            executeWithRetry(WAIT_DURATION, "Virtual Node status check") {
+                val response = client.start().proxy.getVirtualNode(shortHashId)
+                if (response.flowP2pOperationalStatus == OperationalStatus.ACTIVE) {
+                    return@executeWithRetry
```
YashNabar: Will this be retried if no exception is thrown?

YashNabar (commented):
Looking good overall, adding some minor comments

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
r3-arish: Yes, it should be retried if an exception is thrown. The executeWithRetry function does have the implementation to handle any exceptions thrown during the execution of the block and it handles them accordingly.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: It has been addressed

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMgm.kt
r3-arish: It has been addressed

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
r3-arish: Screenshot 2023-07-21 at 01 25 46

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -118,52 +134,38 @@ abstract class BaseOnboard : Runnable {
...
-            .setDefaultBasicAuth("admin", restPassword)
-            .defaultBaseUrl(url)
-    }
-
     internal class OnboardException(message: String) : Exception(message)
 
-    internal fun <T> HttpResponse<T>.bodyOrThrow(): T {
-        if (!this.isSuccess) {
-            throw OnboardException("Error: ${this.body}")
+    internal class ResourceNotReadyException(message: String) : Exception(message)
```
YashNabar: Can we extract this into a separate exceptions package, as this is not specific to this command

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -263,137 +259,177 @@ abstract class BaseOnboard : Runnable {
...
+            executeWithRetry(WAIT_DURATION, "Check Registration Progress") {
+                val status = client.start().proxy.checkSpecificRegistrationProgress(holdingId, registrationId)
+
+                when (val registrationStatus = status.registrationStatus) {
+                    RegistrationStatus.APPROVED -> return@executeWithRetry
+                    RegistrationStatus.DECLINED,
+                    RegistrationStatus.INVALID,
+                    RegistrationStatus.FAILED -> throw OnboardException("Status of registration is $registrationStatus.")
+
+                    else -> println("Status of registration is $registrationStatus")
```
YashNabar: Same as my previous comment - we will need to throw an exception in the else case to trigger a retry

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMgm.kt
```diff
@@ -109,9 +113,8 @@ class OnboardMgm : Runnable, BaseOnboard() {
     }
 
     private val cpi by lazy {
-        val parametersCpiFile = cpiFile
-        if(parametersCpiFile != null) {
-            return@lazy parametersCpiFile
+        if (cpiFile != null) {
+            return@lazy cpiFile
```
YashNabar:
```suggestion
            return@lazy cpiFile!!
```
This will remove the need for the not-null assertion below

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMgm.kt
```diff
@@ -136,28 +141,25 @@ class OnboardMgm : Runnable, BaseOnboard() {
...
+                .filter { it.groupPolicy?.contains("CREATE_ID") ?: false }
+                .map { it.cpiFileChecksum }
+                .firstOrNull()
+        }
+        if (existingHash != null) {
+            return@lazy existingHash
+        }
 
-        uploadCpi(cpi.inputStream(), cpbName)
+        uploadCpi(cpi!!.inputStream(), cpbName)
```
YashNabar: Not needed after applying previous comment

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
r3-arish: Exceptions package within the network plugin?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
r3-arish: Will address it

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMgm.kt
r3-arish: Understood

dimosr:
For some reason, I tried to cancel some previous review that had stale comments and GH posted the comments 🤦 , letting you know in case you got a notification. I removed the comments to avoid confusion and I'll take a look again now and publish anything relevant again.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
dimosr: I don't think this is a very elegant way to achieve what we want here, i.e. throwing an error in a scenario that's not really exceptional just to let that exception be caught by `executeWithRetry` and get retried. Furthermore, looking at the code of `executeWithRetry` if you do this, you will end up with an error log of the form "Unable to perform operation" even though the operation was performed fine. 

Here are a few ways I can think we can achieve this in a better way (open to other suggestions):
* Adjust the `executeWithRetry` to accept a lambda that will be used to determine if the result of the operation is what was expected, in which case it can be returned, or otherwise continue retrying with an appropriate log message (different to the one shown when the operation fails).
* Create a separate utility function that does only this looping without caring about any exceptions (as these will be caught by `executeWithRetry`).

The second one might be a bit simpler, judging from the code in `executeWithRetry` at the moment, but trying to write the code for each approach will probably give a better idea of what's better and easier to maintain.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/AllowClientCertificate.kt
```diff
@@ -2,78 +2,62 @@ package net.corda.cli.plugins.network
...
+
+    private fun allowAndListCertificates() {
         if (subjects.isEmpty()) {
             println("No subjects to allow")
             return
         }
 
-        val command = Command()
+        createRestClient(MGMRestResource::class).use { client ->
+            executeWithRetry(waitDuration, "Allow and list certificates") {
```
dimosr: Ideally, `executeWithRetry` should be invoked with a single operation not multiple ones. Think for example about the scenario where the first one succeeds and the second one fails, we'd be replaying the first one for no reason.

Also, the less code you include in the executeWithRetry (ideally only the remote operation), the easier it is to reason about. 

I'd do instead:
```
val mgm = client.start().proxy

executeWithRetry(waitDuration, "Allow and list certificates") {
    println("Allowing certificates...")
    subjects.forEach { subject ->
        println("\t Allowing $subject")
        mgm.mutualTlsAllowClientCertificate(mgmShortHash, subject)
    }
    println("Success!")
}

executeWithRetry(waitDuration, "Allow and list certificates") {
    mgm.mutualTlsListClientCertificate(mgmShortHash).forEach { subject ->
        println("Certificate with subject $subject is allowed")
    }
}
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -29,14 +48,17 @@ import java.security.KeyPairGenerator
...
 
-abstract class BaseOnboard : Runnable {
+abstract class BaseOnboard : Runnable, RestCommand() {
     private companion object {
         const val P2P_TLS_KEY_ALIAS = "p2p-tls-key"
         const val P2P_TLS_CERTIFICATE_ALIAS = "p2p-tls-cert"
         const val SIGNING_KEY_ALIAS = "signing key 1"
         const val SIGNING_KEY_STORE_PASSWORD = "keystore password"
         const val GRADLE_PLUGIN_DEFAULT_KEY_ALIAS = "gradle-plugin-default-key"
+        val WAIT_DURATION: Duration = Duration.of(300, ChronoUnit.SECONDS)
```
dimosr: Seems like we have that defined in several places all with the same value. Can we extract it into a separate class or utilities file and re-use it from there everywhere?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
dimosr: if it's only used within the network plugin, a separate package in the plugin should be fine for the time being.

But, see also [my other comment](https://github.com/corda/corda-runtime-os/pull/4312#discussion_r1270645920) as this exception will go away if we go with that approach.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
dimosr: Thinking about it, I am not even sure `executeWithRetry` is necessary at the moment given the network plugin doesn't have a strong need to be tolerant against individual request failures. Given we've already wrapped a lot of them, no harm keeping them if we want.

But, I see the original comment [here](https://github.com/corda/corda-runtime-os/pull/4312#discussion_r1266822570) was around not implementing retry logic ourselves. And `executeWithRetry` is designed to handle only exceptions at the moment. So, just implementing a utility function that does that continuous checking of an invariant might be much simpler really (just a while loop with a sleep).

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -173,64 +175,58 @@ abstract class BaseOnboard : Runnable {
...
+                hsmAssociation.hsmId.ifEmpty {
+                    null
+                }
+            } ?: run {
+                val response = createRestClient(KeysRestResource::class).use { keyClient ->
+                    keyClient.start().proxy.generateKeyPair(
+                        holdingId, "$holdingId-$category", category, "CORDA.ECDSA.SECP256R1"
+                    )
+                }
+                response.id
```
dimosr: Having this a single block makes it quite hard to read, so I would split it. Also, do we need the hsm id for anything? If I understand correctly, we just need to ensure the operation to assign an HSM was successful.

So, something like the following could work and would much easier to understand in my opinion:
```
createRestClient(HsmRestResource::class).use { client ->
    executeWithRetry(WAIT_DURATION, "Assign Soft HSM") {
        client.start().proxy.assignSoftHsm(holdingId, category)
    }
}

val response = createRestClient(KeysRestResource::class).use { keyClient ->
    keyClient.start().proxy.generateKeyPair(
        holdingId, "$holdingId-$category", category, "CORDA.ECDSA.SECP256R1"
    )
}
return response.id
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -263,137 +259,177 @@ abstract class BaseOnboard : Runnable {
...
+                        usage = "p2p-tls",
+                        alias = P2P_TLS_CERTIFICATE_ALIAS,
+                        certificates = listOf(
+                            HttpFileUpload(
+                                certificate,
+                                "certificate.pem",
+                            )
+                        )
+                    )
+                }
```
dimosr: Similar here, don't wrap multiple operations in a single `executeWithRetry` and worth extracting outside anything that's not a remote request e.g.:
```
val csr = createRestClient(CertificatesRestResource::class).use { client ->
    executeWithRetry(WAIT_DURATION, "Generate CSR") {
        client.start().proxy.generateCsr(
            tenantId = "p2p",
            keyId = tlsKeyId,
            x500Name = certificateSubject,
            subjectAlternativeNames = listOf(p2pHost),
            contextMap = null
        )
    }
}

val csrCertRequest = csr.reader().use { reader ->
    PEMParser(reader).use { parser ->
        parser.readObject()
    }
} as? PKCS10CertificationRequest ?: throw OnboardException("CSR is not a valid CSR: $csr")

createRestClient(CertificatesRestResource::class).use { client ->
    ca.signCsr(csrCertRequest).toPem().byteInputStream().use { certificate ->
        executeWithRetry(WAIT_DURATION, "Generate CSR") {
            client.start().proxy.importCertificateChain(
                usage = "p2p-tls",
                alias = P2P_TLS_CERTIFICATE_ALIAS,
                certificates = listOf(
                    HttpFileUpload(
                        certificate,
                        "certificate.pem",
                    )
                )
            )
        }
    }
}
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/MemberList.kt
```diff
@@ -85,4 +60,31 @@ class MemberList : RestCommand(), Runnable {
...
+                    organization,
+                    organizationUnit,
+                    locality,
+                    state,
+                    country
+                ).members
+            }
+        }
+
+        println(result)
```
dimosr: Looking at your test evidence in the description of the PR, seems like this one prints the classes themselves which is not very easy to read (while previously I suspect we were printing the JSON response).

Can we print this in a JSON prettified way? Seems like we do something similar in `GenerateGoupPolicy.run()` using jackson, so you can use that a starting point.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
r3-arish: thanks, will address it.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
r3-arish: thanks, will address it.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/AllowClientCertificate.kt
r3-arish: thanks, will address it.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/AllowClientCertificate.kt
```diff
@@ -2,80 +2,68 @@ package net.corda.cli.plugins.network
...
-                        mgmShortHash,
-                        subject,
-                    )
-                }
+            subjects.forEach { subject ->
+                println("\t Allowing $subject")
+                mgm.mutualTlsAllowClientCertificate(mgmShortHash, subject)
+            }
+            println("Success!")
+            true // Return true to indicate the invariant is satisfied
```
dimosr: The fact that you always return `true` here is a code smell..

Seems like we didn't have any polling previously here, so why are we adding it now?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -136,12 +133,14 @@ abstract class BaseOnboard : Runnable, RestCommand() {
...
-    internal class ResourceNotReadyException(message: String) : Exception(message)
-
     protected fun uploadCpi(cpi: InputStream, name: String): String {
         return createRestClient(CpiUploadRestResource::class).use { client ->
             cpi.use { jarInputStream ->
-                executeWithRetry(WAIT_DURATION, "Check CPI status") {
+                checkInvariant(
+                    maxAttempts = MAX_ATTEMPTS,
+                    waitInterval = WAIT_INTERVAL,
+                    errorMessage = "Check CPI status: Invariant check failed after maximum attempts."
```
dimosr: I don't think you need a `checkInvariant` here either.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -157,12 +156,16 @@ abstract class BaseOnboard : Runnable, RestCommand() {
 
     private fun checkCpiStatus(id: String): String {
         return createRestClient(CpiUploadRestResource::class).use { client ->
-            executeWithRetry(WAIT_DURATION, "Check CPI status") {
+            checkInvariant(
+                maxAttempts = MAX_ATTEMPTS,
+                waitInterval = WAIT_INTERVAL,
+                errorMessage = "Check CPI status: Invariant check failed after maximum attempts."
```
dimosr: This is going to be shown to the user, so it should be something that's as explanatory as possible. 

In general, you can keep using messages that were there previously, e.g. "CPI request $id is not ready yet!".

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -176,13 +179,13 @@ abstract class BaseOnboard : Runnable, RestCommand() {
 
     private fun waitForVirtualNode(shortHashId: String) {
         createRestClient(VirtualNodeRestResource::class).use { client ->
-            executeWithRetry(WAIT_DURATION, "Virtual Node status check") {
+            checkInvariant(
+                maxAttempts = MAX_ATTEMPTS,
+                waitInterval = WAIT_INTERVAL,
+                errorMessage = "Virtual Node status check: Invariant check failed after maximum attempts."
```
dimosr: Similarly: "Virtual Node $shortHashId is not active yet!"

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -212,21 +215,23 @@ abstract class BaseOnboard : Runnable, RestCommand() {
...
-                }
-                response.id
+        createRestClient(HsmRestResource::class).use { client ->
+            checkInvariant(
+                maxAttempts = MAX_ATTEMPTS,
+                waitInterval = WAIT_INTERVAL,
+                errorMessage = "Assign Soft HSM: Invariant check failed after maximum attempts."
+            ) {
+                client.start().proxy.assignSoftHsm(holdingId, category)
+                true // Return true to indicate the invariant is satisfied
```
dimosr: Similar here, do we need `checkInvariant`?

I can see we were doing previously `.asJson().isSuccess`. What was the purpose of that, e.g. was it to recover from failures? If that's the case, it might be good to replace it with `executeWithRetry` that serves that purpose.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -263,137 +264,210 @@ abstract class BaseOnboard : Runnable {
...
+                    take = 20,
+                    orderBy = "NONE",
+                    category = "TLS",
+                    schemeCodeName = null,
+                    alias = P2P_TLS_KEY_ALIAS,
+                    masterKeyAlias = null,
+                    createdAfter = null,
+                    createdBefore = null,
+                    ids = null,
+                ).isNotEmpty()
```
dimosr: I don't think we need `checkInvariant` here. Looking at the original code, it was just checking whether a TLS exists already in the cluster (in case another member has already created one as part of onboarding) and skip the next steps if so. 

Did you test this btw? I suspect that this might actually fail, because on the first onboarding no keys will exist and this will loop continuously until it times out.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -263,137 +264,210 @@ abstract class BaseOnboard : Runnable {
...
-
-        val generateCsrResponse = Unirest.post("/certificates/p2p/$tlsKeyId/")
-            .body(
-                mapOf(
-                    "x500Name" to certificateSubject,
-                    "subjectAlternativeNames" to listOf(p2pHost)
+
+        if (hasKeys) return
+
+        val tlsKeyId = createRestClient(KeysRestResource::class).use { client ->
```
dimosr, line 294: I don't think we need `checkInvariant for all the remaining methods here as they don't actually check any invariant.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -334,8 +360,13 @@ abstract class BaseOnboard : Runnable, RestCommand() {
...
 
         createRestClient(NetworkRestResource::class).use { client ->
-            executeWithRetry(WAIT_DURATION, "Setup Network") {
+            checkInvariant(
+                maxAttempts = MAX_ATTEMPTS,
+                waitInterval = WAIT_INTERVAL,
+                errorMessage = "Setup Network: Invariant check failed after maximum attempts."
+            ) {
                 client.start().proxy.setupHostedIdentities(holdingId, request)
+                true // Return true to indicate the invariant is satisfied
```
dimosr: same - `checkInvariant` not needed

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -369,16 +400,23 @@ abstract class BaseOnboard : Runnable, RestCommand() {
 
     private fun waitForFinalStatus(registrationId: String) {
         createRestClient(MemberRegistrationRestResource::class).use { client ->
-            executeWithRetry(WAIT_DURATION, "Check Registration Progress") {
+            checkInvariant(
+                maxAttempts = MAX_ATTEMPTS,
+                waitInterval = WAIT_INTERVAL,
+                errorMessage = "Check Registration Progress: Invariant check failed after maximum attempts."
```
dimosr: Better message would be useful e.g. "Registration with ID $registrationId was not completed."

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -369,16 +400,23 @@ abstract class BaseOnboard : Runnable, RestCommand() {
...
-                    RegistrationStatus.APPROVED -> return@executeWithRetry
+                    RegistrationStatus.APPROVED -> true // Return true to indicate the invariant is satisfied
                     RegistrationStatus.DECLINED,
                     RegistrationStatus.INVALID,
                     RegistrationStatus.FAILED -> throw OnboardException("Status of registration is $registrationStatus.")
 
-                    else -> println("Status of registration is $registrationStatus")
+                    else -> {
+                        println("Status of registration is $registrationStatus")
+                        false // Return false to indicate the invariant is not yet satisfied
```
dimosr: The logic operates based on whether the result is null or not (not whether it's true or false). So, I don't think this is doing what you intended to.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/MemberList.kt
```diff
@@ -61,13 +62,15 @@ class MemberList : RestCommand(), Runnable {
...
 
-    private val waitDuration: Duration = Duration.of(300, ChronoUnit.SECONDS)
-
-    private fun performMembersLookup() {
+    private fun performMembersLookup(): String {
         requireNotNull(holdingIdentityShortHash) { "Holding identity short hash was not provided." }
 
         val result: List<RestMemberInfo> = createRestClient(MemberLookupRestResource::class).use { client ->
-            executeWithRetry(waitDuration, "Members lookup") {
+            checkInvariant(
```
dimosr: same - `checkInvariant` not needed

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/MemberList.kt
```diff
@@ -81,10 +84,33 @@ class MemberList : RestCommand(), Runnable {
...
+            }
+        }
     }
 
     override fun run() {
-        performMembersLookup()
+        val result = performMembersLookup()
+        val objectMapper = jacksonObjectMapper()
+
+        // add pretty printer and override indentation to make the nested values look better and the file more presentable
```
dimosr: Can you update the PR description with how it looks with this change?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMgm.kt
```diff
@@ -79,6 +79,7 @@ class OnboardMgm : Runnable, BaseOnboard() {
                         json.readTree(response)
                     )
                 println("Group policy file created at $groupPolicyFile")
+                true // Return true to indicate the invariant is satisfied
```
dimosr: same here - `checkInvariant` not needed

dimosr (commented):
There are a lot of places where `checkInvariant` has been introduced where it's probably not needed. Commented on most of them while reviewing the diff of the latest changes. But, please scan the full PR and make sure we only do it where necessary.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/AllowClientCertificate.kt
```diff
@@ -2,80 +2,53 @@ package net.corda.cli.plugins.network
...
+        }
+        println("Success!")
 
-                println("Success!")
-                mgm.mutualTlsListClientCertificate(mgmShortHash).forEach {
-                    println("Certificate with subject $it is allowed")
-                }
-            }
+        mgm.mutualTlsListClientCertificate(mgmShortHash).forEach { subject ->
+            println("Certificate with subject $subject is allowed")
```
dimosr: Don't we need to wrap this with `use` similar to what we've done in other places? e.g.
```
createRestClient(MGMRestResource::class).use {
            val proxy = it.start().proxy

            ....
}
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -118,52 +132,43 @@ abstract class BaseOnboard : Runnable {
...
+                try {
+                    val status = client.start().proxy.status(id)
+                    if (status.status == "OK") {
+                        status.cpiFileChecksum
+                    } else {
+                        null
+                    }
+                } catch (e: MissingRequestedResourceException) {
+                    // Handle the exception here
+                    println("An error occurred while checking the CPI status: ${e.message}")
```
dimosr: This message will be printed during the time we wait for the cpi to be uploaded even though it's an expected condition. We should probably remove it.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -118,52 +132,43 @@ abstract class BaseOnboard : Runnable {
...
+            ) {
+                try {
+                    val status = client.start().proxy.status(id)
+                    if (status.status == "OK") {
+                        status.cpiFileChecksum
+                    } else {
+                        null
+                    }
+                } catch (e: MissingRequestedResourceException) {
+                    // Handle the exception here
```
dimosr: This comment is not providing any extra insight on top of the existing code. A useful comment would be something that gives information on the why e.g.
```
This exception can be thrown while the CPI upload is being processed, so we catch it and re-try.
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -173,64 +178,70 @@ abstract class BaseOnboard : Runnable {
...
+                maxAttempts = MAX_ATTEMPTS,
+                waitInterval = WAIT_INTERVAL,
+                errorMessage = "Virtual Node $shortHashId is not active yet!"
+            ) {
+                try {
+                    val response = client.start().proxy.getVirtualNode(shortHashId)
+                    response.flowP2pOperationalStatus == OperationalStatus.ACTIVE
+                } catch (e: MissingRequestedResourceException) {
+                    // Handle the exception here
+                    println("An error occurred while waiting for the virtual node: ${e.message}")
```
dimosr: same comment about comment & log message as above.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -173,64 +178,70 @@ abstract class BaseOnboard : Runnable {
...
+                maxAttempts = MAX_ATTEMPTS,
+                waitInterval = WAIT_INTERVAL,
+                errorMessage = "Assign Soft HSM for $category: Invariant check failed after maximum attempts."
             ) {
-                return@repeat
+                try {
+                    client.start().proxy.assignSoftHsm(holdingId, category)
+                } catch (e: MissingRequestedResourceException) {
+                    // Handle the exception here
+                    println("An error occurred while assigning soft HSM: ${e.message}")
```
dimosr: same

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -173,64 +178,70 @@ abstract class BaseOnboard : Runnable {
...
 
     protected fun assignSoftHsmAndGenerateKey(category: String): String {
-        repeat(10) {
-            if (
-                Unirest.post("/hsm/soft/$holdingId/$category").asJson().isSuccess
+        createRestClient(HsmRestResource::class).use { client ->
+            checkInvariant(
+                maxAttempts = MAX_ATTEMPTS,
+                waitInterval = WAIT_INTERVAL,
+                errorMessage = "Assign Soft HSM for $category: Invariant check failed after maximum attempts."
```
dimosr:
```suggestion
                errorMessage = "Assign Soft HSM operation for $category: failed after the maximum number of attempts ($MAX_ATTEMPTS)."
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -263,137 +274,183 @@ abstract class BaseOnboard : Runnable {
...
+                        RegistrationStatus.FAILED -> throw OnboardException("Status of registration is $registrationStatus.")
+
+                        else -> {
+                            println("Status of registration is $registrationStatus")
+                            null
+                        }
+                    }
+                } catch (e: Exception) {
+                    println("Error checking registration progress: ${e.message}")
+                    null // Return false to indicate the invariant is not yet satisfied
```
dimosr: this comment is not accurate anymore (null instead of false)

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -263,137 +274,183 @@ abstract class BaseOnboard : Runnable {
...
-                    throw OnboardException("Status of registration is $registrationStatus.")
-                }
-                else -> {
-                    println("Status of $x500Name registration is $registrationStatus")
+    private fun waitForFinalStatus(registrationId: String) {
+        createRestClient(MemberRegistrationRestResource::class).use { client ->
+            checkInvariant(
+                maxAttempts = MAX_ATTEMPTS,
+                waitInterval = WAIT_INTERVAL,
+                errorMessage = "Check Registration Progress: Invariant check failed after maximum attempts."
```
dimosr:
```suggestion
                errorMessage = "Check Registration Progress failed after maximum number of attempts ($MAX_ATTEMPTS)."
```

Keep in mind invariant is an internal detail, so if you print that to the console it will just be confusing to the user that has no knowledge of the internal implementation.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/Constants.kt
```diff
@@ -1,4 +1,4 @@
 package net.corda.cli.plugins.network
 
-const val MAX_ATTEMPTS = 5
-const val WAIT_INTERVAL = 1000L
\ No newline at end of file
+const val MAX_ATTEMPTS = 8
+const val WAIT_INTERVAL = 3000L
```
dimosr: Is this a reasonable default for everything?

That means that even operations that complete quite fast (say 500ms) will wait for the next poll in 3 seconds before returning. We can have a lower, more reasonable default (say poll every 1 second) and then if there are cases where we know it takes longer and we want to poll more infrequently, we can have a custom value there.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/MemberList.kt
```diff
@@ -85,4 +60,55 @@ class MemberList : RestCommand(), Runnable {
...
+                organizationUnit,
+                locality,
+                state,
+                country
+            ).members
+        }
+
+        val resultMap = mutableMapOf<String, Any>()
+        result.forEachIndexed { index, memberInfo ->
+            resultMap["member$index"] = memberInfo.copy()
```
dimosr: I think keying by the x500 name of the member will provide a better experience.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/MemberList.kt
```diff
@@ -85,4 +60,55 @@ class MemberList : RestCommand(), Runnable {
...
+                println(it)
+            }
+        }
+    }
+
+    override fun run() {
+        val result = performMembersLookup()
+        val objectMapper = jacksonObjectMapper()
+
+        // add pretty printer and override indentation to make the nested values look better and the file more presentable
```
dimosr: Left-over from copy paste? I think it can be removed.

dimosr (dismissed):
LGTM - worth @YashNabar taking another look at it.

And given the lack of e2e tests on this area, worth doing a final manual test before merging.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMgm.kt
```diff
@@ -62,20 +63,16 @@ class OnboardMgm : Runnable, BaseOnboard() {
     }
 
     private fun saveGroupPolicy() {
-        repeat(10) {
-            try {
```
YashNabar: Was this handling a scenario where this would be called too soon? Do we want to keep that retry?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMgm.kt
```diff
@@ -136,28 +134,25 @@ class OnboardMgm : Runnable, BaseOnboard() {
...
+                .filter { it.groupPolicy?.contains("CREATE_ID") ?: false }
+                .map { it.cpiFileChecksum }
+                .firstOrNull()
+        }
+        if (existingHash != null) {
+            return@lazy existingHash
+        }
 
-        uploadCpi(cpi.inputStream(), cpbName)
+        uploadCpi(cpi!!.inputStream(), cpbName)
```
YashNabar:
```suggestion
        uploadCpi(cpi.inputStream(), cpbName)
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMgm.kt
r3-arish: This won't be needed as we have the wait interval

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -30,7 +50,7 @@ import java.security.KeyStore
 import java.security.cert.CertificateFactory
 import java.util.Date
 
-abstract class BaseOnboard : Runnable {
+abstract class BaseOnboard : Runnable, RestCommand() {
```
YashNabar, line 54: We should remove `cordaClusterName` and its usages as well

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMgm.kt
YashNabar: Discussed offline, was addressed in [65f7ae8](https://github.com/corda/corda-runtime-os/pull/4312/commits/65f7ae8524c7a9fea33aee1da29daadeffbefd5e)

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
YashNabar: On second thought, ignore my comment - I thought this was left over from the k8s commands that were removed. Resolving.

r3-arish:
Screenshot 2023-07-27 at 09 55 21
Screenshot 2023-07-27 at 09 55 33

@dimosr here are the results

YashNabar (commented):
LGTM - there are some conflicts that need to be resolved

r3-arish:
> LGTM - there are some conflicts that need to be resolved

I think it's resolved now

dimosr (approved):
LGTM - a bit strange the merge commit shows everything brought from the main branch. Typically, in merges it only shows the conflict changes (so close to zero always).

r3-arish:
> LGTM - a bit strange the merge commit shows everything brought from the main branch. Typically, in merges it only shows the conflict changes (so close to zero always).

Yeah, I never saw this happening - I followed the steps Github told me for a merge conflict

PR #4314, opened 2023-07-17
CORE-14685: Remove message bus implementation.

chrisr3:
Remove `db-message-bus-impl`, which was only providing its `CordaAvroSerializationFactory` implementation. Copy this implementation directly into the driver instead.


yarivus (commented):
thx @chrisr3

PR #4315, opened 2023-07-17
CORE-15441: Remove references to Beta signing keys in CorDapp packaging page

anien:
- Removes references to all Beta signing keys. References now only GA release.
- As GA release is signed with new signing keys, gradle certificate is no longer required in Corda when uploading CPI.

### DIFF
--- a/wiki/CorDapp-Packaging.md
+++ b/wiki/CorDapp-Packaging.md
@@ -94,7 +94,7 @@ If we are using the default signing key, we need to import it into our key store
 ### Trust the R3 signing key
 
 The notary CPB is signed with an R3 production code signing certificate issued by the DigiCert.
-To use it import this certificate.
+To use it import DigiCert certificate.
 
 1. Save the following text into a file named `digicert-ca.pem`
     ```text
@@ -173,7 +173,7 @@ The gradle plugin will build the CPB. Run this command to turn a CPB into a CPI:
 
 Corda will validate that uploaded CPIs are signed with a trusted key. To trust your signing keys, upload them with these commands.
 
-1. Import the Digicert signing key into Corda
+1. Import the DigiCert certificate into Corda
    ```shell
    curl --insecure -u admin:admin -X PUT -F alias="digicert-ca" -F certificate=@digicert-ca.pem https://localhost:8888/api/v1/certificates/cluster/code-signer
    ```
--- a/wiki/CorDapp-Packaging.md
+++ b/wiki/CorDapp-Packaging.md
@@ -91,9 +91,9 @@ If we are using the default signing key, we need to import it into our key store
     keytool -importcert -keystore signingkeys.pfx -storepass "keystore password" -noprompt -alias gradle-plugin-default-key -file gradle-plugin-default-key.pem
     ```
 
-### Trust the signing key
+### Trust the R3 signing key
 
-The notary CPB is signed with a production signing key.
+The notary CPB is signed with an R3 production code signing certificate issued by the DigiCert.
 To use it import this certificate.
 
 1. Save the following text into a file named `digicert-ca.pem`
--- a/wiki/CorDapp-Packaging.md
+++ b/wiki/CorDapp-Packaging.md
@@ -186,3 +186,8 @@ Corda will validate that uploaded CPIs are signed with a trusted key. To trust y
     curl --insecure -u admin:admin -X PUT -F alias="signingkey1-2022" -F certificate=@signingkey1.pem https://localhost:8888/api/v1/certificates/cluster/code-signer
     ```
     _Use an alias that will be unique over time. Consider how certificate expiry will require new certificates with the same x500 name as existing certificates and define a naming convention that covers that use case._
+1. Import the gradle plugin default key into Corda
+    ```shell
+    curl --insecure -u admin:admin -X PUT -F alias="gradle-plugin-default-key" -F certificate=@gradle-plugin-default-key.pem https://localhost:8888/api/v1/certificates/cluster/code-signer
+    ```
+   **Note:** Do not import the gradle plugin key into the Corda in production. This should only be used in development.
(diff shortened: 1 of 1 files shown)


wiki/CorDapp-Packaging.md
```diff
@@ -226,15 +173,7 @@ The gradle plugin will build the CPB. Run this command to turn a CPB into a CPI:
 
 Corda will validate that uploaded CPIs are signed with a trusted key. To trust your signing keys, upload them with these commands.
 
-1. Import the gradle plugin default key into Corda
-    ```shell
-    curl --insecure -u admin:admin -X PUT -F alias="gradle-plugin-default-key" -F certificate=@gradle-plugin-default-key.pem https://localhost:8888/api/v1/certificates/cluster/code-signer
```
simon-johnson-r3, line 231: Deleting this line means the reader will never import the `gradle-plugin-default-key` into Corda, but above they have already imported it into the `signingkeys.pfx` passed to the CPI. Surely it is required in either both places or neither?

anien: My understanding is that when you create a CPI, you sign it with your new code-signing certificate. This signing cert (if self signed) or it's root CA has to be imported into Corda to verify this CPI. 
You need `gradle-plugin-default-key` to be imported into `signingkeys.pfx` only when creating CPI, because it verifies the CPB which is signed by developer cert, before resigning with your new code-signing cert.

So essentially, you don't need `gradle-plugin-default-key` to be imported in Corda, as the CPI which is uploaded is not signed with developer certificate. I tested the steps and they work.

There is one last bit which I am a bit unsure of is if there are any other parts of Corda signed by developer cert which are imported in Corda and therefore need `gradle-plugin-default-key` for the verification. If we want to be on a safe side, we can keep the import of `gradle-plugin-default-key` to Corda. But this page was dealing with CPI, for which you don't need gradle cert in Corda.

josephzunigadaly: The gradle plugin will sign with this if no signing key is configured. So CorDapp developers will need this if they choose not to configure a signing key during development.

It should not be imported into production systems because anyone can sign with this key. So if we keep it, we should add a note/warning to say not to import into production.

anien: I have added back (as a last step) the import of gradle certificate into Corda, but added a note it should not be used in production.

simon-johnson-r3 (approved):

wiki/CorDapp-Packaging.md
```diff
@@ -91,64 +91,11 @@ If we are using the default signing key, we need to import it into our key store
     keytool -importcert -keystore signingkeys.pfx -storepass "keystore password" -noprompt -alias gradle-plugin-default-key -file gradle-plugin-default-key.pem
     ```
 
-### Trust the beta signing key
+### Trust the signing key
 
-The notary CPB was signed with a test signing key for the beta builds up to Beta3 (Hawk).
```
kyriathar, line 96: My understanding is the notary CPB should still be signed by some key? Shouldn't we be now pointing to that other (non beta (?)) key?

wiki/CorDapp-Packaging.md
```diff
@@ -91,64 +91,11 @@ If we are using the default signing key, we need to import it into our key store
...
+The notary CPB is signed with a production signing key.
```
kyriathar: I thought the notary CPB is provided by us isn't it? Which production key we are referring to here?

wiki/CorDapp-Packaging.md
```diff
@@ -91,64 +91,11 @@ If we are using the default signing key, we need to import it into our key store
     keytool -importcert -keystore signingkeys.pfx -storepass "keystore password" -noprompt -alias gradle-plugin-default-key -file gradle-plugin-default-key.pem
     ```
 
-### Trust the beta signing key
+### Trust the signing key
```
kyriathar: This is very vague now I believe, "_which signing key to trust?_"

kyriathar (commented):
A few comments mostly to do with who is to provide the notary CPB hence who is to sign it.

wiki/CorDapp-Packaging.md
anien: We can say _Trust the production signing key_, but it's also specified straight in the first sentence in this section.

wiki/CorDapp-Packaging.md
anien: It's our R3 production key it's signed with. There was a different key for few Beta versions, but since May, we have a production R3 signing key/certificate we use for signing notary CPB.

Would this be better? _The notary CPB is signed with an R3 production code signing certificate._

wiki/CorDapp-Packaging.md
anien: The request from QA was to remove Beta related instructions and focus on the GA. We now say that notary CPB is signed with a production signing key. And we are pointing to that in this section.

josephzunigadaly (approved):

wiki/CorDapp-Packaging.md
josephzunigadaly: The notary CPB is signed as part of the release process. The Digicert root covers trust for that signature.

wiki/CorDapp-Packaging.md
josephzunigadaly: It is our production key.

> Would this be better? The notary CPB is signed with an R3 production code signing certificate.

That does spell it out.

wiki/CorDapp-Packaging.md
josephzunigadaly: "Trust the production signing key" is good, or we could say "Trust the R3 signing key", or "Trust Digicert code signing certificates" (since trusting the root covers all Digicert issued code signing certs)

josephzunigadaly (approved):

kyriathar (approved):
LGTM

anien:
This is now merged in the wiki repo.

PR #4320, opened 2023-07-17
CORE-15489 tidy up fatal and intermittent exception handling

ciaran238:
In `CordaKafkaConsumerImpl::poll` there is a catch-all exception clause which segments exceptions into fatal and intermittent exceptions based on their type. It does this by using fall-through logic in a `when` statement.

It would be cleaner to pull these exceptions out into a set so we can separate the config from the logic.
The added benefit is that a set membership lookup is more efficient than the linear check in the `when` statement, and the `class.java` equality comparison avoids some unnecessary sub-typing checks done by `is`.

### DIFF
--- a/libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
+++ b/libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
@@ -60,6 +60,25 @@ class CordaKafkaConsumerImpl<K : Any, V : Any>(
 
     private companion object {
         private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
+
+        val fatalExceptions: Set<Class<out Throwable>> = setOf(
+            AuthorizationException::class.java,
+            AuthenticationException::class.java,
+            IllegalArgumentException::class.java,
+            IllegalStateException::class.java,
+            ArithmeticException::class.java,
+            FencedInstanceIdException::class.java,
+            InconsistentGroupProtocolException::class.java,
+            InvalidOffsetException::class.java,
+            CommitFailedException::class.java
+        )
+        val transientException: Set<Class<out Throwable>> = setOf(
+            TimeoutException::class.java,
+            WakeupException::class.java,
+            InterruptException::class.java,
+            KafkaException::class.java,
+            ConcurrentModificationException::class.java
+        )
     }
 
     override fun close() {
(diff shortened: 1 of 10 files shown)


libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/CordaKafkaProducerImpl.kt
```diff
@@ -357,37 +376,25 @@ class CordaKafkaProducerImpl(
...
-            is UnsupportedForMessageFormatException,
-            is AuthorizationException,
-            is FencedInstanceIdException -> {
+        when (ex::class.java) {
+            in fatalExceptions -> {
                 throw CordaMessageAPIFatalException("FatalError occurred $errorString", ex)
             }
 
-            is IllegalStateException -> {
+            in setOf(IllegalStateException::class.java) -> {
```
ben-millar: Probably don't need to move this one into a set as there's only one

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/CordaKafkaProducerImpl.kt
```diff
@@ -357,37 +376,25 @@ class CordaKafkaProducerImpl(
...
-            is KafkaException -> {
+           in transientExceptions -> {
                 if (abortTransaction) {
                     abortTransaction()
                 }
                 throw CordaMessageAPIIntermittentException("Error occurred $errorString", ex)
             }
-            is CordaMessageAPIFatalException,
-            is CordaMessageAPIIntermittentException -> { throw ex }
+            in setOf(CordaMessageAPIFatalException::class.java,
```
ben-millar: This could be extracted out into an `ApiExceptions` val, potentially

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
```diff
@@ -309,25 +315,16 @@ class CordaKafkaConsumerImpl<K : Any, V : Any>(
                 consumer.commitSync(offsets)
                 attemptCommit = false
             } catch (ex: Exception) {
-                when (ex) {
-                    is InterruptException,
-                    is TimeoutException,
-                    -> {
+                when (ex::class.java) {
+                    in transientExceptions -> {
```
ben-millar: Super nit-pick warning: This is the only check that goes transient and then fatal, the others are all done the other way around.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/CordaKafkaProducerImpl.kt
ciaran238: 👍

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/CordaKafkaProducerImpl.kt
ciaran238: 👍

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/consumer/CordaKafkaConsumerImpl.kt
ciaran238: 👍

ben-millar (approved):
LGTM pending 🟢 build

LWogan (approved):
LGTM

PR #4322, opened 2023-07-17
CORE-13937: add serializers for 2 party finality flow optimisation

jennyang-r3:
### Overview
With later versions in mind, we changed the payload design in the first interaction of finality flow so that more data can be added to it and only the required properties are deserialised.

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
@@ -7,6 +7,17 @@ import net.corda.v5.base.annotations.ConstructorForDeserialization
 import net.corda.v5.base.annotations.CordaSerializable
 import net.corda.v5.base.exceptions.CordaRuntimeException
 
+/**
+ * Payload for V2 for [UtxoFinalityFlowV1] and [UtxoReceiveFinalityFlowV1] on the first interaction.
+ * It is to
+ * - serialize and deserialize on the interaction
+ * - safely add more field into the payload in the future
+ * - deserialize only required field
+ *
+ * @constructor Creates a new instance of [FinalityPayload] data class
+ * @property initialTransaction [UtxoSignedTransactionInternal]
+ * @property transferAdditionalSignatures [Boolean]
+*/
 @CordaSerializable
 data class FinalityPayload @ConstructorForDeserialization constructor(
     val map: Map<String, ByteArray>,
@@ -41,7 +52,7 @@ data class FinalityPayload @ConstructorForDeserialization constructor(
             serializationService.deserialize<Boolean>(map[TRANSFER_ADDITIONAL_SIGNATURES]!!)
         } catch (e: Exception) {
             throw CordaRuntimeException(
-                "Deserialization of ${map[INITIAL_TRANSACTION]} into UtxoSignedTransactionInternal failed.", e)
+                "Deserialization of ${map[TRANSFER_ADDITIONAL_SIGNATURES]} into Boolean failed.", e)
         }
     }
 }
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
@@ -1,22 +1,42 @@
 package net.corda.ledger.utxo.flow.impl.flows.finality
 
 import net.corda.ledger.utxo.flow.impl.transaction.UtxoSignedTransactionInternal
+import net.corda.utilities.serialization.deserialize
+import net.corda.v5.application.serialization.SerializationService
 import net.corda.v5.base.annotations.ConstructorForDeserialization
 import net.corda.v5.base.annotations.CordaSerializable
-import net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction
 
 @CordaSerializable
-data class FinalityPayload @ConstructorForDeserialization constructor(val map: Map<String, Any?>) {
+data class FinalityPayload @ConstructorForDeserialization constructor(
+    val map: Map<String, ByteArray>,
+    private val serializationService: SerializationService
+) {
     private companion object {
         const val INITIAL_TRANSACTION = "INITIAL_TRANSACTION"
         const val TRANSFER_ADDITIONAL_SIGNATURES = "TRANSFER_ADDITIONAL_SIGNATURES"
     }
-    constructor(initialTransaction: UtxoSignedTransaction, transferAdditionalSignatures: Boolean) : this(
+
+    constructor(
+        initialTransaction: UtxoSignedTransactionInternal,
+        transferAdditionalSignatures: Boolean,
+        serializationService: SerializationService
+    ) : this(
         mapOf(
-            INITIAL_TRANSACTION to initialTransaction,
-            TRANSFER_ADDITIONAL_SIGNATURES to transferAdditionalSignatures
-        )
+            INITIAL_TRANSACTION to serializationService.serialize(initialTransaction).bytes,
+            TRANSFER_ADDITIONAL_SIGNATURES to serializationService.serialize(transferAdditionalSignatures).bytes
+        ), serializationService
     )
-    val initialTransaction get() = map[INITIAL_TRANSACTION] as UtxoSignedTransactionInternal
-    val transferAdditionalSignatures get() = map[TRANSFER_ADDITIONAL_SIGNATURES] as Boolean
+
+    val initialTransaction by lazy(LazyThreadSafetyMode.PUBLICATION) {
+        requireNotNull(map[INITIAL_TRANSACTION]) {
+            "initialTransaction is null"
+        }
+        serializationService.deserialize<UtxoSignedTransactionInternal>(map[INITIAL_TRANSACTION]!!)
+    }
+    val transferAdditionalSignatures by lazy(LazyThreadSafetyMode.PUBLICATION) {
+        requireNotNull(map[TRANSFER_ADDITIONAL_SIGNATURES]) {
+            "transferAdditionalSignatures is null"
+        }
+        serializationService.deserialize<Boolean>(map[TRANSFER_ADDITIONAL_SIGNATURES]!!)
+    }
 }
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
@@ -5,6 +5,7 @@ import net.corda.utilities.serialization.deserialize
 import net.corda.v5.application.serialization.SerializationService
 import net.corda.v5.base.annotations.ConstructorForDeserialization
 import net.corda.v5.base.annotations.CordaSerializable
+import net.corda.v5.base.exceptions.CordaRuntimeException
 
 @CordaSerializable
 data class FinalityPayload @ConstructorForDeserialization constructor(
@@ -28,15 +29,19 @@ data class FinalityPayload @ConstructorForDeserialization constructor(
     )
 
     val initialTransaction by lazy(LazyThreadSafetyMode.PUBLICATION) {
-        requireNotNull(map[INITIAL_TRANSACTION]) {
-            "initialTransaction is null"
+        try {
+            serializationService.deserialize<UtxoSignedTransactionInternal>(map[INITIAL_TRANSACTION]!!)
+        } catch (e: Exception) {
+            throw CordaRuntimeException(
+                "Deserialization of ${map[INITIAL_TRANSACTION]} into UtxoSignedTransactionInternal failed.", e)
         }
-        serializationService.deserialize<UtxoSignedTransactionInternal>(map[INITIAL_TRANSACTION]!!)
     }
     val transferAdditionalSignatures by lazy(LazyThreadSafetyMode.PUBLICATION) {
-        requireNotNull(map[TRANSFER_ADDITIONAL_SIGNATURES]) {
-            "transferAdditionalSignatures is null"
+        try {
+            serializationService.deserialize<Boolean>(map[TRANSFER_ADDITIONAL_SIGNATURES]!!)
+        } catch (e: Exception) {
+            throw CordaRuntimeException(
+                "Deserialization of ${map[INITIAL_TRANSACTION]} into UtxoSignedTransactionInternal failed.", e)
         }
-        serializationService.deserialize<Boolean>(map[TRANSFER_ADDITIONAL_SIGNATURES]!!)
     }
 }
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
@@ -5,19 +5,7 @@ import net.corda.utilities.serialization.deserialize
 import net.corda.v5.application.serialization.SerializationService
 import net.corda.v5.base.annotations.ConstructorForDeserialization
 import net.corda.v5.base.annotations.CordaSerializable
-import net.corda.v5.base.exceptions.CordaRuntimeException
 
-/**
- * Payload for V2 for [UtxoFinalityFlowV1] and [UtxoReceiveFinalityFlowV1] on the first interaction.
- * It is to
- * - serialize and deserialize on the interaction
- * - safely add more field into the payload in the future
- * - deserialize only required field
- *
- * @constructor Creates a new instance of [FinalityPayload] data class
- * @property initialTransaction [UtxoSignedTransactionInternal]
- * @property transferAdditionalSignatures [Boolean]
-*/
 @CordaSerializable
 data class FinalityPayload @ConstructorForDeserialization constructor(
     val map: Map<String, ByteArray>,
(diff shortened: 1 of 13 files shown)


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/serializer/FinalityPayloadSerializer.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    override val proxyType
+        get() = FinalityPayloadProxy::class.java
+
+    override val withInheritance
+        // FinalityPayload is final.
+        get() = false
+
+    override fun toProxy(obj: FinalityPayload): FinalityPayloadProxy {
+        return FinalityPayloadProxy(obj.initialTransaction, obj.transferAdditionalSignatures)
```
lankydan: Make the map in `FinalityPayload` public and then pass the map into the proxy instead.

This means that the serialized form of the `FinalityPayload` is just a map of string to bytes and nothing else.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/serializer/FinalityPayloadSerializer.kt
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    override fun fromProxy(proxy: FinalityPayloadProxy): FinalityPayload {
+        return FinalityPayload(proxy.initialTransaction, proxy.transferAdditionalSignatures, serializationService)
```
lankydan: Once we've pass the map into the proxy we'll want to call the constructor you have with the map and the serialization service here.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -35,6 +36,7 @@ class UtxoFinalityFlowV1(
     private val initialTransaction: UtxoSignedTransactionInternal,
     private val sessions: List<FlowSession>,
     private val pluggableNotaryClientFlow: Class<PluggableNotaryClientFlow>,
+    private val serializationService: SerializationService,
```
lankydan: I'd rather that the `SerializationServce` relied on `@CordaInject` to be injected into this class.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -294,9 +302,10 @@ class UtxoFinalityFlowV1(
...
     ): PluggableNotaryClientFlow {
         return AccessController.doPrivileged(PrivilegedExceptionAction {
-            pluggableNotaryClientFlow.getConstructor(UtxoSignedTransaction::class.java, MemberX500Name::class.java).newInstance(
-                transaction, virtualNodeSelectorService.selectVirtualNode(transaction.notaryName)
-            )
+            pluggableNotaryClientFlow.getConstructor(UtxoSignedTransaction::class.java, MemberX500Name::class.java)
+                .newInstance(
+                    transaction, virtualNodeSelectorService.selectVirtualNode(transaction.notaryName)
+                )
         })
```
lankydan: Unneeded formatting.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/serializer/kryo/FinalityPayloadKryoSerializer.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.Reference
+import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
+
+@Component(
+    service = [CheckpointInternalCustomSerializer::class, UsedByFlow::class, UsedByVerification::class],
+    property = [CORDA_UNINJECTABLE_SERVICE],
+    scope = PROTOTYPE
+)
+class FinalityPayloadKryoSerializer @Activate constructor(
```
lankydan, line 22: I don't think this will work correctly.

If someone sent us a V1 version of the payload and we have V2 with an extra field, it will try to pull an extra field out of the map that doesn't exist and cause an error.

To keep things simple, especially because our usage is not _that_ complicated. I suggest that we remove the `lazy` fields from `FinalityPayload` and keep them as getters. This means that we can make keep the writing into the kryo output as the `map` from the payload and then read it back in when deserializing.

If we want to keep the lazy fields then we'd need something like `LazyStateAndRefImplKryoSerializer`, but I don't think we'd need that because `FinalityPayload` is internal to us and we basically read everything out of it straight away at the moment.

We could always make it lazy in the future if needed without issue.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1Test.kt
```diff
@@ -515,7 +584,10 @@ class UtxoReceiveFinalityFlowV1Test {
         verify(persistenceService, times(2)).persist(signedTransactionWithOwnKeys, TransactionStatus.UNVERIFIED)
     }
 
-    private fun callReceiveFinalityFlow(version: UtxoFinalityVersion, validator: UtxoTransactionValidator = UtxoTransactionValidator { }) {
+    private fun callReceiveFinalityFlow(
+        version: UtxoFinalityVersion,
+        validator: UtxoTransactionValidator = UtxoTransactionValidator { }
+    ) {
```
lankydan, line 590: Revert this formatting change.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1Test.kt
```diff
@@ -505,7 +569,12 @@ class UtxoReceiveFinalityFlowV1Test {
     @Test
     fun `Finality flow V2 - receiving and persisting signatures when there are more than two parties`() {
 
-        whenever(signedTransaction.addMissingSignatures()).thenReturn(signedTransactionWithOwnKeys to listOf(signature1, signature2))
+        whenever(signedTransaction.addMissingSignatures()).thenReturn(
```
lankydan, line 572: There are a lot of unneeded formatting changes.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -98,7 +102,13 @@ class UtxoFinalityFlowV1(
                 initialTransaction, sessions.toSet()
             )
         } else {
-            flowMessaging.sendAll(FinalityPayload(initialTransaction, transferAdditionalSignatures), sessions.toSet())
+            flowMessaging.sendAll(
+                FinalityPayload(
+                    initialTransaction,
+                    transferAdditionalSignatures,
+                    serializationService
+                ), sessions.toSet()
```
lankydan, line 110: put `sessions.toSet()` on a new line

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/serializer/kryo/FinalityPayloadKryoSerializer.kt
```diff
@@ -0,0 +1,39 @@
...
+        output.writeClassAndObject(obj.payload)
+    }
+
+    override fun read(input: CheckpointInput, type: Class<out FinalityPayload>): FinalityPayload {
+        return FinalityPayload(
+            input.readClassAndObject() as UtxoSignedTransactionInternal,
+            input.readClassAndObject() as Boolean,
+            serialisationService
+        )
+    }
```
lankydan, line 38: This should be passing in a map into the `FinalityPayload` constructor.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
```diff
@@ -1,22 +1,36 @@
...
+import net.corda.utilities.serialization.deserialize
+import net.corda.v5.application.serialization.SerializationService
 import net.corda.v5.base.annotations.ConstructorForDeserialization
 import net.corda.v5.base.annotations.CordaSerializable
-import net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction
 
 @CordaSerializable
-data class FinalityPayload @ConstructorForDeserialization constructor(val map: Map<String, Any?>) {
+data class FinalityPayload @ConstructorForDeserialization constructor(
+    val map: Map<String, ByteArray>,
```
lankydan, line 11: We have both `map` and `payload` as public accessors to the map of data. Either `map` should be private or `payload` should be removed. We do not need both to be public. I think having the `map` public is probably the right solution,

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
```diff
@@ -1,22 +1,36 @@
...
+    private val serializationService: SerializationService
```
nkovacsx, line 12: Having a service property inside a data class is something we shouldn't do. Couldn't we just pass in the already serialized object? @lankydan what do you think?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
```diff
@@ -1,22 +1,36 @@
...
 import net.corda.ledger.utxo.flow.impl.transaction.UtxoSignedTransactionInternal
+import net.corda.utilities.serialization.deserialize
+import net.corda.v5.application.serialization.SerializationService
 import net.corda.v5.base.annotations.ConstructorForDeserialization
 import net.corda.v5.base.annotations.CordaSerializable
-import net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction
 
 @CordaSerializable
-data class FinalityPayload @ConstructorForDeserialization constructor(val map: Map<String, Any?>) {
+data class FinalityPayload @ConstructorForDeserialization constructor(
```
lankydan, line 10: Because we're using serializers now, we can remove `@ConstructorForDeserialization constructor` since we're explicitly defining how to serialize/deserialize this class now.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/UtxoFinalityFlow.kt
```diff
@@ -17,7 +17,7 @@ import net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction
 class UtxoFinalityFlow(
     private val transaction: UtxoSignedTransactionInternal,
     private val sessions: List<FlowSession>,
-    private val pluggableNotaryClientFlow: Class<PluggableNotaryClientFlow>
+    private val pluggableNotaryClientFlow: Class<PluggableNotaryClientFlow>,
```
lankydan, line 20: Undo this change.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/UtxoFinalityFlow.kt
```diff
@@ -34,7 +34,7 @@ class UtxoFinalityFlow(
 
 class UtxoFinalityFlowVersionedFlowFactory(
     private val transaction: UtxoSignedTransactionInternal,
-    private val pluggableNotaryClientFlow: Class<PluggableNotaryClientFlow>
+    private val pluggableNotaryClientFlow: Class<PluggableNotaryClientFlow>,
```
lankydan, line 37: Undo this change.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/serializer/FinalityPayloadSerializer.kt
```diff
@@ -0,0 +1,47 @@
...
+import net.corda.serialization.BaseProxySerializer
+import net.corda.serialization.InternalCustomSerializer
+import net.corda.v5.application.serialization.SerializationService
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.Reference
+import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
+
+@Component(
+    service = [InternalCustomSerializer::class, UsedByFlow::class, UsedByPersistence::class, UsedByVerification::class],
```
lankydan, line 17: We don't need `UsedByPErsistence` or `UsedByVerification` for this class.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/serializer/FinalityPayloadSerializer.kt
```diff
@@ -0,0 +1,47 @@
...
+    property = [CORDA_UNINJECTABLE_SERVICE],
+    scope = PROTOTYPE
+)
+class FinalityPayloadSerializer @Activate constructor(
+    @Reference(service = SerializationService::class)
+    private val serializationService: SerializationService
+): BaseProxySerializer<FinalityPayload, FinalityPayloadProxy>(), UsedByFlow, UsedByPersistence, UsedByVerification {
```
lankydan, line 24: Same as comment above.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/serializer/kryo/FinalityPayloadKryoSerializer.kt
```diff
@@ -0,0 +1,39 @@
...
+import net.corda.serialization.checkpoint.CheckpointInternalCustomSerializer
+import net.corda.serialization.checkpoint.CheckpointOutput
+import net.corda.v5.application.serialization.SerializationService
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.Reference
+import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
+
+@Component(
+    service = [CheckpointInternalCustomSerializer::class, UsedByFlow::class, UsedByVerification::class],
```
lankydan, line 18: We don't need `UsedByVerification` here.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/serializer/kryo/FinalityPayloadKryoSerializer.kt
```diff
@@ -0,0 +1,39 @@
...
+    property = [CORDA_UNINJECTABLE_SERVICE],
+    scope = PROTOTYPE
+)
+class FinalityPayloadKryoSerializer @Activate constructor(
+    @Reference(service = SerializationService::class)
+    private val serialisationService: SerializationService,
+) : CheckpointInternalCustomSerializer<FinalityPayload>, UsedByFlow, UsedByVerification {
```
lankydan, line 25: Remove `UsedByVerification`.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/serializer/FinalityPayloadSerializer.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    override val type
+        get() = FinalityPayload::class.java
```
nkovacsx, line 27: I don't think you need the explicit getters here.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/serializer/kryo/FinalityPayloadKryoSerializer.kt
lankydan: Did we test this, because I _think_ this would fail if it was run.

libs/flows/flow-api/src/main/kotlin/net/corda/flow/application/versioning/VersionedSendFlowFactory.kt
```diff
@@ -11,17 +11,19 @@ import net.corda.v5.application.messaging.FlowSession
...
  *     override fun create(version: Int, sessions: List<FlowSession>): SubFlow<UtxoSignedTransaction> {
- *         return when {
- *             version >= 10 -> UtxoFinalityFlowV2(transaction, sessions, pluggableNotaryClientFlow)
- *             version >= 1 -> UtxoFinalityFlowV1(transaction, sessions, pluggableNotaryClientFlow)
+ *         val finalityVersion = when {
+ *             version >= CORDA_5_1.value -> UtxoFinalityVersion.V2
+ *             version in 1 until CORDA_5_1.value -> UtxoFinalityVersion.V1
  *             else -> throw IllegalArgumentException()
  *         }
+ *         return UtxoFinalityFlowV1(transaction, sessions, pluggableNotaryClientFlow, serializationService, finalityVersion)
```
lankydan, line 26: Correct this code to match the real code.

nkovacsx (changes requested):
I'm not really happy with the fact that we pass a service into a POJO (data) class, could we do something with that?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
lankydan: It has to be this way, if we want the getters to work as they are. So if we don't pass in a `SerializationService` then the getters need to take in the service to deserialize the contents. Or, we we just have the getters return bytes are deserialize in the flow. That _might_ be simpler to understand but generally the amount of code will be the same.

@blsemo @relyafi thoughts?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/serializer/kryo/FinalityPayloadKryoSerializer.kt
jennyang-r3: If this is going to be a map, are we reading this as `Map<String, Any>?` But we have separate getters for each property in the map. Or am I making no sense?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
lankydan: Passing in the `SerializationService` into the getters might be the right solution but will wait for others to comment.

jennyang-r3:
Closing the PR since the scenario initially thought about wouldn't happen, which is why we created this PR.

PR #4324, opened 2023-07-18
CORE-12100 Fix inconsistent logging

nkovacsx:
### Overview

We used to log the full keys (encoded) and not the key ids. This has now been modified.


lankydan (approved):

kyriathar (approved):
LGTM

PR #4329, opened 2023-07-18
CORE-15138 Move ledger e2e tests from runtime OS to e2e repo 

PranushanR3:


ronanbrowne (dismissed):
LGMT (build change)

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/ledger/UtxoLedgerTests.kt
```diff
@@ -240,4 +232,16 @@ class UtxoLedgerTests {
     )
 
     data class CustomQueryFlowResponse(val results: List<String>)
+
+    internal object SecureHashSerializer : com.fasterxml.jackson.databind.JsonSerializer<SecureHash>() {
```
nkovacsx, line 211: Why do we need these de/serializers?

nkovacsx (changes requested):
Just one minor comment, I'm not sure why we need the new de/serializers.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/ledger/UtxoLedgerTests.kt
PranushanR3: This is because they are being used in line 39 and 40 in the ObjectMapper.

PranushanR3: This is being used on line 39 and 40 by the object builder. The SecureHash is then used on line 164 to parse the result.

blsemo: I think the answer Nandor was looking for is:
You had to move it here because it was re-used from the (now deleted) consensual tests (which was a code smell anyway).

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/Utils.kt
```diff
@@ -64,3 +65,5 @@ fun getHoldingIdShortHash(x500Name: String, groupId: String): String {
 fun JsonNode.parseContextMap(): Map<String, String> = fields().asSequence().map {
     it.key to it.value.textValue()
 }.toMap()
+
+fun parseSecureHashString(hashString: String) = parseSecureHash(hashString)
```
blsemo, line 69: Where is this used?

testing/e2e-test-utilities/build.gradle
```diff
@@ -22,4 +22,5 @@ dependencies {
     implementation project(':tools:plugins:package')
     implementation project(":testing:packaging-test-utilities")
     implementation "org.slf4j:slf4j-api:$slf4jVersion"
+    implementation project(':libs:crypto:crypto-core')
```
blsemo, line 25: Why is this required now?

blsemo (commented):
Looks mostly ok to me, but I wonder about the changes to the test utilities lib?

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/Utils.kt
PranushanR3: This is now used in the ConsensualLedgerTests which has been moved to the `corda5-e2e-tests` repo.

testing/e2e-test-utilities/build.gradle
PranushanR3: This is in order to have access to the parseSecureHash method used in the `utils.kt` file. This method is needed in the `utils.kt` file in order to have access to the method in the `ConsensualLedgerTests.kt` file now in the e2e repo.

lankydan (commented):
I feel similar to the other comments, I'm confused why removing one test has caused the addition of other code.

Also based on the ticket, shouldn't `Utxo Ledger - custom query can be executed and results are returned if no offset is provided and limit is maximize` be moved as well?

chrisr3:
I see a lot smoke tests being migrated out of `corda-runtime-os` recently. Can I request that the relevant CPBs in `:testing:cpbs` remain intact please, because I am still using them to test my Driver.

PranushanR3:
> I see a lot smoke tests being migrated out of `corda-runtime-os` recently. Can I request that the relevant CPBs in `:testing:cpbs` remain intact please, because I am still using them to test my Driver.

I have not touched any of the `testing:cpbs` and so it should all still be there.

chrisr3:
> I have not touched any of the `testing:cpbs` and so it should all still be there.

That's fine. My comment was addressed to the wider audience :wink:.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/Utils.kt
blsemo: If it's only used in the ConsensualLedgerTests in the e2e repo, it should probably live there? Having a method in a smoketest package that does not get called doesn't make sense.

PranushanR3: We need it to be in the smoketest package due to the fact we can't access the crypto library from the `corda5-e2e-tests` repo.

lankydan (approved):

nkovacsx (approved):

ronanbrowne (approved):

PR #4330, opened 2023-07-18
CORE-10114 Introduce custom metadata in static networks

YashNabar:
Allows members of a static network to specify custom ("ext.") properties in their registration context which will be included in their MemberInfo. Also allow properties (minimum platform version and custom "ext." parameters) to be specified in the group policy of static network CPIs which will be included in the group parameters of the network.
https://github.com/corda/corda-api/pull/1180

### Group Parameters
1. Custom parameters added to group policy (under `protocolParameters.staticNetwork`) [cropped image]
Screenshot 2023-07-18 at 17 42 54

2. Custom parameters included in the group parameters
Screenshot 2023-07-18 at 16 47 32

### Registration Context
1. Custom properties specified during registration are included in MemberInfo
Screenshot 2023-07-18 at 16 45 29


libs/membership/membership-impl/src/main/kotlin/net/corda/membership/lib/impl/grouppolicy/v1/MemberGroupPolicyImpl.kt
```diff
@@ -105,6 +107,16 @@ class MemberGroupPolicyImpl(rootNode: JsonNode) : MemberGroupPolicy {
...
             output.toList()
         }
+
+        override val staticNetworkGroupParameters: Map<String, String>? =
+            staticNetwork?.getOptionalStringMap(GROUP_PARAMETERS)?.apply {
+                val verifierResult = GroupParametersUpdateVerifier().verify(this)
+                if (verifierResult is GroupParametersUpdateVerifier.Result.Failure) {
+                    with(verifierResult.reason) {
+                        throw BadGroupPolicyException(this)
+                    }
```
charlieR3: Thjis seems a bit tidier:
```suggestion
                throw BadGroupPolicyException(verifierResult.reason)
```

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/staticnetwork/StaticMemberRegistrationService.kt
```diff
@@ -233,6 +237,11 @@ class StaticMemberRegistrationService(
                 ex,
             )
         }
+        val customFieldsValid = customFieldsVerifier.verify(context)
+        if (customFieldsValid is RegistrationContextCustomFieldsVerifier.Result.Failure) {
+            logger.info(customFieldsValid.reason)
```
charlieR3: Similar message for the exception would be good.
`"Registration failed. ${customFieldsValid.reason}"`
If the registration ID is available it would be useful to include it also.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/staticnetwork/StaticMemberRegistrationService.kt
```diff
@@ -489,7 +501,7 @@ class StaticMemberRegistrationService(
             PLATFORM_VERSION to platformInfoProvider.activePlatformVersion.toString(),
             MEMBER_CPI_NAME to cpi.name,
             MEMBER_CPI_VERSION to cpi.version,
-        ) + optionalContext
+        ) + optionalContext + customFields
```
charlieR3, line 505: It could be better to have custom fields first, then add the platform properties. Otherwise someone might be able to overwrite a platform property with a custom field input.

charlieR3: Or is this relying on the schema validation to ensure customFields is only `ext.` prefixed values?

YashNabar: The schema validation is done before this, so it should be safe since they can only have "ext." properties at this stage

charlieR3 (dismissed):

charlieR3 (dismissed):

gradle.properties
```diff
@@ -45,7 +45,7 @@ bouncycastleVersion=1.73
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.0.0.xx-SNAPSHOT to pick up maven local published copy
 #cordaApiVersion=5.1.0.xxx-SNAPSHOT
-cordaApiVersion=5.1.0.8-beta+
```
josephzunigadaly, line 48: This should be reverted before merging.

simon-johnson-r3 (dismissed):

charlieR3 (approved):

simon-johnson-r3 (approved):

PR #4334, opened 2023-07-19
CORE-15497: make use of notary passed from request body

jennyang-r3:
This PR changed api of `registerStaticMember` and context notary body in MGM registration for e2e testing, so that we can make use of notary that has passed to request body of demo flow, as opposed to notary not being used previously. Now the update will use either notary that was specified in call site or whatever notary that was available in network.

### DIFF
--- a/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
+++ b/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
@@ -68,6 +68,7 @@ class FlowTests {
             davidX500,
             notaryX500
         )
+        private val notaryServiceX500 = "O=MyNotaryService-${notaryHoldingId}, L=London, C=GB"
 
         val invalidConstructorFlowNames = listOf(
             "com.r3.corda.testing.smoketests.flow.errors.PrivateConstructorFlow",
@@ -126,7 +127,7 @@ class FlowTests {
 
             registerStaticMember(bobHoldingId)
             registerStaticMember(charlieHoldingId)
-            registerStaticMember(notaryHoldingId, isNotary = true)
+            registerStaticMember(notaryHoldingId, notaryServiceX500)
         }
 
         private val JsonNode?.command: String
--- a/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
+++ b/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
@@ -68,7 +68,7 @@ class FlowTests {
             davidX500,
             notaryX500
         )
-        private val notaryServiceX500 = "O=MyNotaryService-${notaryHoldingId}, L=London, C=GB"
+        private const val NOTARY_SERVICE_X500 = "O=MyNotaryService, L=London, C=GB"
 
         val invalidConstructorFlowNames = listOf(
             "com.r3.corda.testing.smoketests.flow.errors.PrivateConstructorFlow",
@@ -127,7 +127,7 @@ class FlowTests {
 
             registerStaticMember(bobHoldingId)
             registerStaticMember(charlieHoldingId)
-            registerStaticMember(notaryHoldingId, notaryServiceX500)
+            registerStaticMember(notaryHoldingId, NOTARY_SERVICE_X500)
         }
 
         private val JsonNode?.command: String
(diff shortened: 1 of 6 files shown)


testing/cpbs/ledger-utxo-demo-app/src/main/kotlin/com/r3/corda/demo/utxo/UtxoDemoFlow.kt
```diff
@@ -66,7 +65,8 @@ class UtxoDemoFlow : ClientStartableFlow {
                 request.members + listOf(myInfo.name.toString())
             )
 
-            val notary = notaryLookup.notaryServices.single()
+            val notary = notaryLookup.lookup(MemberX500Name.parse(request.notary))?: notaryLookup.notaryServices.single()
```
relyafi: I think this needs to be a bit smarter - as it stands, this does not differentiate between no notary being provided, and an invalid / unknown name being provided. I'd suggest:

- `request.notary` should become an optional parameter. If this is null, use `single()`
- If not null, then do the lookup and parse, but we should fail with an exception if this happens so the caller knows there was an error with their input.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
```diff
@@ -68,6 +68,7 @@ class FlowTests {
             davidX500,
             notaryX500
         )
+        private val notaryServiceX500 = "O=MyNotaryService-${notaryHoldingId}, L=London, C=GB"
```
relyafi: I'd probably remove the holding id from this string now, because the test is now in control of the name, and we only have one notary service in this test. So we don't need to dis-ambiguate the service name by adding the vnode holding id.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/token/selection/TokenBalanceQueryTests.kt
```diff
@@ -48,6 +48,7 @@ class TokenBalanceQueryTests {
         val objectMapper = ObjectMapper().apply {
             registerModule(KotlinModule.Builder().build())
         }
+        private val notaryServiceX500 = "O=MyNotaryService-${notaryHoldingId}, L=London, C=GB"
```
relyafi: See earlier comment

testing/cpbs/ledger-utxo-demo-app/src/main/kotlin/com/r3/corda/demo/utxo/UtxoDemoFlow.kt
```diff
@@ -66,7 +65,13 @@ class UtxoDemoFlow : ClientStartableFlow {
                 request.members + listOf(myInfo.name.toString())
             )
 
-            val notary = notaryLookup.notaryServices.single()
+            val notary = if (request.notary == null) {
+                notaryLookup.notaryServices.single()
```
chrisr3, line 69: We _never_ want to use `.single()` because this throws an exception if the list does not contain _exactly_ one element, whereas a network can happily contain several notaries. I would suggest using `.random()` instead, but I suspect the _real_ solution is just to _insist_ that the request _does_ specify a notary.

relyafi: I completely disagree with this; I'd rather we have an obvious failure indicating the flow hasn't been designed to handle multiple notaries, than it silently selecting an arbitrary notary.

We had issues in C4 with people seeing widespread use of `first()` and people considering that a precedent and using that in real CorDapps. I want to avoid any repeat of that, which is why we are consistently trying to use `single()` throughout our demo and test flows etc. if it hasn't been designed to handle multiple notaries.

relyafi: Also, yes we could mandate that a notary must be specified, but that puts unnecessary burden on test writers who predominantly will be writing tests where they have only set up a single notary. Again we're setting a precedent here for how our test flows behave.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/ledger/UtxoLedgerTests.kt
```diff
@@ -52,6 +52,8 @@ class UtxoLedgerTests {
     private val charlieHoldingId: String = getHoldingIdShortHash(charlieX500, groupId)
     private val notaryHoldingId: String = getHoldingIdShortHash(notaryX500, groupId)
 
+    private val notaryServiceX500 = "O=MyNotaryService, L=London, C=GB"
```
relyafi: This could be a `const` (same as in `FlowTest`)

relyafi (commented):
1 minor comment, otherwise LGTM.

I'd also remove reference to the E2E tests from the PR title - it's true that the Jira ticket covers this scope, but this PR is only changing the common components and smoke tests in runtime-os

blsemo (approved):
LGTM

PR #4339, opened 2023-07-20
CORE-15451 Snyk waivers for kubernetes-client

davidcurrie:
Adds waivers for two vulnerabilities introduced via `kubernetes-client` that do not impact Corda as they only apply to the use of the dependent libraries on the server side, not as a client.


ronanbrowne (approved):
LGTM

PR #4340, opened 2023-07-20
CORE-15003: Update product version for feature (#4156)

fowlerrr:
This will prevent this branch from interfereing with the libs stored in Artifactory


ronanbrowne (approved):
LGTM

PR #4341, opened 2023-07-20
CORE-14232 - Upgrade json schema validator library

dimosr:
## Changes
Requires changes from corda-api [here](https://github.com/corda/corda-api/pull/1183).

Upgrading the `json-schema-validator` library to get rid of some spurious WARNING logs around keywords that are not identified (e.g. `defs` / `then` / `else`) even though they are supported by that version of the spec. Some more discussion and details available on [this issue](https://github.com/networknt/json-schema-validator/issues/79). I didn't upgrade up to the latest version of the spec (`V202012`) on purpose, because when I attempted it there were errors on some of our tests, e.g. `ExternalChannelsConfigValidatorTest` was failing expecting exceptions in case of unknown properties that weren't caught by validation. The release notes for that version of the spec do not provide any commentary on backwards compatibility in contrast to previous versions, so it's possible there are breaking changes. Given the scope of this is just to eliminate the warning, upgrading to V201909 should be sufficient. I also didn't upgrade to the latest version of `json-schema-validator` (1.0.86), because that depends on version 2.0.7 of SLF4j, whereas Corda uses version 1.7.36 which is the only one provided at the OSGi level, thus leading to OSGi resolution errors. I upgraded to the latest version of the library that depends on version 1.7.36 of SLF4j, while also containing the fix we need. I also removed the exclusion of `com.fasterxml.jackson.core:jackson-databind` as a transitive dependency of it in the `:libs:serialisation` module, since that has now been upgraded to the latest version that doesn't have a CVE.

I also had to force use of a specific version (1.7.3) of `com.ethlo.time`, because the one used by the library by default (1.7.0) does not have an OSGi bundle. 

## Testing
I verified that the warning logs of the form `WARN  com.networknt.schema.JsonMetaSchema {} - Unknown` are not present in the combined worker logs anymore. 

I also run the [e2e-tests](https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda5-e2e-tests/detail/PR-164/5/pipeline/) with these changes to sanity check there are no regressions there.


gradle.properties
```diff
@@ -45,7 +45,7 @@ bouncycastleVersion=1.73
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.0.0.xx-SNAPSHOT to pick up maven local published copy
 #cordaApiVersion=5.1.0.xxx-SNAPSHOT
-cordaApiVersion=5.1.0.9-beta+
+cordaApiVersion=5.1.0.9-alpha-1689862358467
```
vkolomeyko: So there is a `corda-api` PR which probably needs approving first?
We also need not forget rollback back to `beta` version.

libs/configuration/configuration-validation/build.gradle
```diff
@@ -22,6 +22,9 @@ dependencies {
     implementation "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"
     implementation "com.typesafe:config:$typeSafeConfigVersion"
     implementation "com.networknt:json-schema-validator:$networkntJsonSchemaVersion"
+    constraints {
+        implementation("com.ethlo.time:itu:$comEthloTimeItuVersion")
```
vkolomeyko:
```suggestion
        implementation("com.ethlo.time:itu:$comEthloTimeItuVersion") {
            because "Version bundled with current version of 'com.networknt:json-schema-validator' does not have OSGi manifest."
        }
```

vkolomeyko (commented):
Overall looks fine. Few minor comments posted.

gradle.properties
dimosr: yes, this one: https://github.com/corda/corda-api/pull/1183

I typically hold the dependency on the `alpha` version until before merging to ensure they are compatible. And switch back to beta right after merging the `corda-api` PR.

kyriathar (dismissed):
OK (packaging)

ronanbrowne (dismissed):
LGTM - pending alpha version being changed back

yift-r3 (dismissed):

yift-r3 (approved):

vkolomeyko (approved):
LGTM

kyriathar (approved):

ronanbrowne (approved):

PR #4343, opened 2023-07-20
ES-794/multi cluster pipeline

seanbrereton:
Add a pipeline that runs tests with the multicluster tag

New Jenkins job: https://ci02.dev.r3.com/job/Corda5/job/corda-runtime-os-multi-cluster-tests/

Tested: https://ci02.dev.r3.com/job/Corda5/job/corda-runtime-os-multi-cluster-tests/job/seanb%252FES-794%252Fmulti-cluster-pipeline/6/

### DIFF
--- a/.ci/e2eTests/JenkinsfileMultiClusterTest
+++ b/.ci/e2eTests/JenkinsfileMultiClusterTest
@@ -1,9 +1,10 @@
 @Library('corda-shared-build-pipeline-steps@5.1') _
 
 endToEndPipeline(
-    assembleAndCompile: false,
-    dailyBuildCron: '',
-    multiCluster: false,
+    assembleAndCompile: true,
+    dailyBuildCron: '0 */3 * * *',
+    multiCluster: true,
     gradleTestTargetsToExecute: ['smokeTest', 'e2eTest'],
     usePackagedCordaHelmChart: false,
+    gradleAdditionalArgs : '-PrunMultiClusterTests -Dscan.tag.MultiCluster'
 )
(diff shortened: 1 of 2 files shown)


yarivus (dismissed):
LGTM

.ci/e2eTests/JenkinsfileMultiClusterTest
```diff
@@ -0,0 +1,10 @@
+@Library('corda-shared-build-pipeline-steps@5.1') _
+
+endToEndPipeline(
+    assembleAndCompile: true,
```
charlieR3: Not my area of expertise but should this match the single cluster config and be `false`?

.ci/e2eTests/JenkinsfileMultiClusterTest
```diff
@@ -0,0 +1,10 @@
...
+    dailyBuildCron: '0 */12 * * *',
+    multiCluster: true,
+    gradleTestTargetsToExecute: ['smokeTest', 'e2eTest'],
+    usePackagedCordaHelmChart: false,
+    gradleAdditionalArgs : '-PrunMultiClusterTests -Dscan.tag.MultiCluster'
```
charlieR3, line 8: It might be useful to add `-PrunMultiClusterTests` to the unstable jenkins pipeline also so the unstable tests job is running both single and multi cluster unstable tests.

seanbrereton: If a multicluster test is also set as unstable it will still be run on the unstable pipeline. So it shouldn't be needed

charlieR3: Cool, thanks. I thought the logic would ignore them if they were tagged with multi cluster and the run multi cluster flag wasn't set.

charlieR3 (dismissed):

.ci/e2eTests/JenkinsfileMultiClusterTest
seanbrereton: Changed to keep in line with single cluster

yarivus (approved):
LGTM

PR #4344, opened 2023-07-20
CORE-14685: Implement custom queries for HSQLDB.

chrisr3:
Extends Corda's custom queries to support HSQLDB as well as Postgres. This has required mapping some Java functions into SQL to implement selected query operations on JSON-formatted columns.

I have post-processed the custom query expressions as follows, e.g.:
```sql
WHERE visible_state.custom_representation->a->>b::int = :value
```
is transformed to:
```sql
WHERE CAST( JsonFieldAsText( JsonFieldAsObject( CAST(visible_state.custom_representation AS VARCHAR(32768)), 'a'), 'b') AS int) = :value
```
And
```sql
WHERE visible_state.custom_representation->a?b
```
becomes:
```sql
WHERE HasJsonKey( JsonFieldAsObject( CAST(visible_state.custom_representation AS VARCHAR(32768)), 'a'), 'b')
```


testing/db-hsqldb-json/build.gradle
```diff
@@ -0,0 +1,22 @@
+plugins {
+    id 'corda.common-library'
+    id 'corda.common-publishing'
```
ronanbrowne, line 3: Just checking, this and the other testing/* module you added, do we need 

`corda.common-publishing` applied.

what will be consuming this outside of this repo that we need them published somewhere?

chrisr3: I am writing a testing tool for CorDapp developers to test their flows with, and they will need this module in order for HSQLDB to handle Corda's custom queries. So "yes", this new module will need to be published.

And ditto for the `ledger-hsqldb`, `driver` and `driver-engine` modules.

ronanbrowne (approved):
LGTM - blt files

testing/db-hsqldb-json/build.gradle
ronanbrowne: got it, thanks

buildSrc/src/main/groovy/corda.java-only.gradle
```diff
@@ -0,0 +1,7 @@
+pluginManager.withPlugin('org.jetbrains.kotlin.jvm') {
+    tasks.named('compileKotlin') {
+        doFirst {
+            throw new InvalidUserCodeException("Module '${project.path}' should only contain Java classes")
+        }
+    }
+}
```
relyafi, line 7: It's not clear to me how this change relates to the stated intent of this PR?

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/EntityManagerFactoryFactoryImpl.kt
```diff
@@ -71,7 +71,7 @@ class EntityManagerFactoryFactoryImpl(
         classLoaders: List<ClassLoader>,
         configuration: EntityManagerConfiguration
     ): EntityManagerFactory {
-        log.info("Creating for $persistenceUnitName")
+        log.info("Creating for {}", persistenceUnitName)
```
relyafi, line 74: Unnecessary change?

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/EntityManagerFactoryFactoryImpl.kt
```diff
@@ -58,8 +58,8 @@ class EntityManagerFactoryFactoryImpl(
     ): EntityManagerFactory {
         return create(
             persistenceUnitName,
-            entities.map { it.canonicalName },
-            entities.map { it.classLoader }.distinct(),
+            entities.map(Class<*>::getCanonicalName),
+            entities.map(Class<*>::getClassLoader).distinct(),
```
relyafi, line 62: Does this have anything to do with the stated changes?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/AbstractVaultNamedQueryParserImpl.kt
```diff
@@ -4,19 +4,19 @@ import net.corda.ledger.persistence.query.parsing.converters.VaultNamedQueryConv
 import net.corda.ledger.persistence.query.parsing.expressions.VaultNamedQueryExpressionParser
 import net.corda.ledger.persistence.query.parsing.expressions.VaultNamedQueryExpressionValidator
 
-class VaultNamedQueryParserImpl(
+abstract class AbstractVaultNamedQueryParserImpl(
```
relyafi, line 7: I think this change is unnecessary - the "parser" is parsing custom query expressions written by a CorDapp, which _so happen_ to use Postgres syntax. CorDapps are not going to write different parsers for different DBs, it should be transparent.

So supporting different DBs should require no changes to the parser, it should be purely the converter classes (which convert from the custom query tokens into the DB specific SQL) that should be changed.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Keywords.kt
```diff
@@ -4,230 +4,234 @@ interface Keyword : Token
...
+        return (this === other) || (other is JsonField)
     }
 
     override fun hashCode(): Int {
-        return javaClass.hashCode()
+        return this::class.java.hashCode()
+    }
+}
+
+class HsqldbJsonField : Keyword, HasLeftParenthesis {
```
relyafi, line 105: Again this just seems wrong, no changes should be needed here - the custom query language should never reference DB specifics, it just so happens to use the same syntax as Postgres.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Keywords.kt
```diff
@@ -4,230 +4,234 @@ interface Keyword : Token
...
     override fun equals(other: Any?): Boolean {
-        if (this === other) return true
-        if (javaClass != other?.javaClass) return false
-        return true
+        return (this === other) || (other is NotEquals)
     }
 
     override fun hashCode(): Int {
-        return javaClass.hashCode()
+        return this::class.java.hashCode()
```
relyafi, line 11: Can you elaborate on why these changes are needed?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/AbstractVaultNamedQueryConverterImpl.kt
```diff
@@ -0,0 +1,84 @@
...
+import net.corda.ledger.persistence.query.parsing.SqlType
+import net.corda.ledger.persistence.query.parsing.Token
+import net.corda.ledger.persistence.query.parsing.Where
+
+abstract class AbstractVaultNamedQueryConverterImpl : VaultNamedQueryConverter {
+
+    /**
+     * Conversion rules for extra database-specific tokens.
+     */
+    protected abstract fun customConvert(token: Token): String?
```
relyafi, line 38: Doesn't make sense - there should be no database specific tokens.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/PostgresVaultNamedQueryConverter.kt
```diff
@@ -1,73 +1,7 @@
...
-                is Like -> output.append(" LIKE ")
-                is ParameterEnd -> output.append(", ")
-                else -> throw IllegalArgumentException("Invalid token in expression - $token")
-            }
-        }
-    }
-}
\ No newline at end of file
+class PostgresVaultNamedQueryConverter : AbstractVaultNamedQueryConverterImpl() {
+    override fun customConvert(token: Token): String? = null
```
relyafi, line 6: The whole point of these classes is that it's the `convert` function that should produce a different output based on the (standard) set of tokens. If most of these statements are ANSI SQL (looks like it), you might want to put most of these in an Abstract `AnsiSqVaultNamedQueryConverter` / `BaseVaultNamedQueryConverter` and then just handle specific tokens in each DB specific type.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/expressions/AbstractVaultNamedQueryExpressionParserImpl.kt
```diff
@@ -0,0 +1,168 @@
...
+import net.corda.ledger.persistence.query.parsing.Parameter
+import net.corda.ledger.persistence.query.parsing.ParameterEnd
+import net.corda.ledger.persistence.query.parsing.PathReference
+import net.corda.ledger.persistence.query.parsing.PathReferenceWithSpaces
+import net.corda.ledger.persistence.query.parsing.RightParentheses
+import net.corda.ledger.persistence.query.parsing.Select
+import net.corda.ledger.persistence.query.parsing.Token
+import net.corda.ledger.persistence.query.parsing.Where
+
+abstract class AbstractVaultNamedQueryExpressionParserImpl : VaultNamedQueryExpressionParser {
```
relyafi, line 33: Again this shouldn't be necessary, as it relates to our custom query expressions which should be DB agnostic.

relyafi (commented):
I've not reviewed the whole set of changes, because it looks to me like it could be made significantly simpler. The main point of concern is that the custom query parser logic should be DB agnostic, so it's not clear why any changes in this area are necessary; the model was designed such that only the custom query converter logic should need to be swapped out for a different DB implementation.

buildSrc/src/main/groovy/corda.java-only.gradle
chrisr3: The `db-hsqldb-json` module must become a fragment bundle of `org.hsqldb.hsqldb`, and so is written in pure Java. I have created this extra convention plugin to ensure that it _remains_ Java.

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/EntityManagerFactoryFactoryImpl.kt
chrisr3: Cleanup.

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/EntityManagerFactoryFactoryImpl.kt
chrisr3: More c;leanup.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/AbstractVaultNamedQueryParserImpl.kt
chrisr3: To understand the abstraction you must look at the extensions:
```kotlin
@Component(service = [ VaultNamedQueryParser::class ])
class PostgresVaultNamedQueryParserImpl @Activate constructor(
    @Reference(target = POSTGRES_TYPE_FILTER)
    databaseTypeProvider: DatabaseTypeProvider
): AbstractVaultNamedQueryParserImpl(
    expressionParser = PostgresVaultNamedQueryExpressionParser(),
    expressionValidator = VaultNamedQueryExpressionValidatorImpl(),
    converter = PostgresVaultNamedQueryConverter()
) {
    init {
        LoggerFactory.getLogger(this::class.java).info("Activated for {}", databaseTypeProvider.databaseType)
    }
}
```
and
```kotlin
@Component(service = [ VaultNamedQueryParser::class ])
class HsqldbVaultNamedQueryParserImpl @Activate constructor(
    @Reference(target = HSQLDB_TYPE_FILTER)
    databaseTypeProvider: DatabaseTypeProvider
): AbstractVaultNamedQueryParserImpl(
    expressionParser = HsqldbVaultNamedQueryExpressionParser(),
    expressionValidator = VaultNamedQueryExpressionValidatorImpl(),
    converter = HsqldbVaultNamedQueryConverter()
) {
    init {
        LoggerFactory.getLogger(this::class.java).info("Activated for {}", databaseTypeProvider.databaseType)
    }
}
```
In other words, it is a base class for both OSGi components, whose service type is `VaultNamedQueryParser`.

We now have two such `VaultNamedQueryParser` components, one for each DBMS type.

buildSrc/src/main/groovy/corda.java-only.gradle
relyafi: Ok. In that case I would suggest an update to `buildSrc/README.md` to document the plugin / explain the intent.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Keywords.kt
chrisr3: I transform the initial `JsonField` token into something else for HSQLDB, which also requires adding additional tokens and re-ording some of the others too. Naturally, the original `JsonField` token is removed in the process.

Custom Query -> (Tokens) -> SQL
Custom Query -> (Tokens -> HSQLDB Tokens) -> SQL

>it just so happens to use the same syntax as Postgres.

This is very unhelpful for databases whose syntax bears no resemblance whatsoever to Postgres's.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Keywords.kt
chrisr3: Because `javaClass` is [deprecated](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/java-class.html).

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/EntityManagerFactoryFactoryImpl.kt
relyafi: So no then?

If there's a good reason to be making these changes then I think there needs to be a proper explanation. Ideally we'd just have a separate ticket entirely to enforce a standard set of usages across the code base. Multiple changes like this in a PR obfuscate the intended changes of a PR.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/AbstractVaultNamedQueryConverterImpl.kt
chrisr3: It does make sense if a database has created its own tokens with their own DBMS-specific mappings. The JSON functions I wrote specifically for HSQLDB cannot possibly be expected to exist on other databases, and the `JsonField` etc tokens map to nonsense for HSQLDB.

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/EntityManagerFactoryFactoryImpl.kt
relyafi: Please can you elaborate - is there a technical reason why this is better, or is this just personal style? If it's the former, that's probably an argument for a separate PR to apply this change to logging across the board. If it's the latter, we shouldn't be using PR's as a way to change existing style to one of personal preference.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/PostgresVaultNamedQueryConverter.kt
chrisr3: > The whole point of these classes is that it's the convert function that should produce a different output based on the (standard) set of tokens. 

Then the designer must simply have failed to consider databases which are considerably _different_ to Postgres.

The initial token stream is utterly unsuitable for HSQLDB, and it took a lot of effort to transform it into something which would execute my new JSON functions in the correct order, and with appropriate `CAST` operations in the correct places.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/expressions/AbstractVaultNamedQueryExpressionParserImpl.kt
chrisr3: The custom query expressions _are_ DB agnostic. The process to parse them into tokens which can generate working native SQL is anything but.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Keywords.kt
relyafi: `Custom Query -> (Tokens -> HSQLDB Tokens) -> SQL`

I think this is fine if this is what you need to do, just the "HSQLDB Tokens" should be private to the HSQLDB converter class, and that second phase should be done there.

`This is very unhelpful for databases whose syntax bears no resemblance whatsoever to Postgres's.`

But this is inevitable - we have to have a single query language that CorDapp developers can use, and that's never going to match every database. This is why we have the parser / converter logic. Otherwise we might just as well tell CorDapp developers to write their own native SQL, which would be a terrible idea for multiple reasons.

chrisr3: The "converter" stage is "tokens in, SQL out". The rearrangement is "tokens in, other tokens out". And _only then_ can we starts "tokens in, SQL out".

yarivus (commented):
thx @chrisr3 
thx @relyafi for reviewing 👍

PR #4345, opened 2023-07-20
CORE-15008: Large network test script

yift-r3:
Introduce a script to run a large network test.


applications/tools/p2p-test/app-simulator/scripts/large-network-testing/runScenario.sh
```diff
@@ -0,0 +1,80 @@
...
+for ((memberIndex=1; memberIndex<=MEMBER_COUNT; memberIndex++)); do
+  for cluster in "${clusters[@]}"; do
+      echo "Onboard member $memberIndex onto $cluster"
+      $CORDA_CLI_DIR/build/generatedScripts/corda-cli.sh network onboard member \
+         --cpb-file $REPO_TOP_LEVEL_DIR/testing/cpbs/calculator/build/libs/*.cpb \
+         --ca "$WORKING_DIR"/ca \
+         --group-policy-file "$WORKING_DIR"/gp/groupPolicy.json \
+         --wait \
+         $cluster
+      echo "Member $memberIndex was onboarded into $cluster" >> "$reportFile"
```
dimosr: This just checks the registration succeeded, right?

I think as part of validation we should also be testing that other members in the network can see that member (to check for example that data distribution worked fine). Due to the size of the network, it might not be practical to check that every single member can see them if that makes the test extremely slow so we might have to make a compromise there (e.g. maybe check for a few random members?).

dimosr (commented):
I started reviewing the latest changes, but then as I realised that this bash script started becoming quite hard to read/understand and maintain. 

How about writing it as a junit (e2e) test that we keep disabled for now? It seems like it would be an easy thing to do adjusting one of our existing e2e tests and will probably be much easier to maintain going forward.

applications/tools/p2p-test/app-simulator/src/test/kotlin/net/corda/p2p/tests/LargeNetworkTest.kt
```diff
@@ -0,0 +1,220 @@
...
+                    "$port:8888"
+                ).start().also { process ->
+                    Runtime.getRuntime().addShutdownHook(
+                        thread(false) {
+                            process.destroy()
+                        }
+                    )
+                    val reader = process.inputStream.bufferedReader()
+                    reader.readLine()
+                    reader.readLine()
```
dimosr: what's the reason for these 2 invocations? worth a comment clarifying.

applications/tools/p2p-test/app-simulator/src/test/kotlin/net/corda/p2p/tests/LargeNetworkTest.kt
```diff
@@ -0,0 +1,220 @@
...
+            if (index % 5 == 0) {
+                println("Validate network")
+                eventually(
+                    duration = Duration.ofMinutes(10),
+                    waitBetween = Duration.ofSeconds(10),
+                    retryAllExceptions = true,
+                ) {
+                    validateNetwork()
+                }
+            }
```
dimosr: In principle, I think we'd need to check every member has been propagated properly. If I understand correctly, this checks only every 5th member. 

Also, if we check the whole network every time, it might become a bit impractical for large networks. You will have a better idea seeing latencies from your runs. If that's the case what we could do is for every new member, we could pick a few random members in the network already (say 5) and verify that they can see that new member. 

Last, it would be nice to capture somewhere some basic statistics on how long each registration took to complete and how long for the member info to propagate. If it's too much work, happy to do it separately but seems something that could be done in a relatively crude way without too much work.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MembershipUtils.kt
```diff
@@ -46,7 +46,7 @@ const val DEFAULT_SIGNATURE_SPEC = "SHA256withECDSA"
  */
 @Suppress("LongParameterList")
 fun ClusterInfo.onboardMember(
-    cpb: String,
```
dimosr, line 49: Will this be a breaking change? Do we also need a change in the e2e-tests repo for this?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -173,7 +173,7 @@ abstract class BaseOnboard : Runnable {
     protected abstract val registrationContext: Map<String, Any?>
 
     private fun waitForVirtualNode(shortHashId: String) {
-        repeat(10) {
+        repeat(240) {
```
dimosr, line 176: is this needed now?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/Helpers.kt
```diff
@@ -39,8 +39,10 @@ internal object Helpers {
                 "deployment/$restWorkerDeploymentName",
                 "$port:8888"
             )
-                .inheritIO()
                 .start().also { process ->
+                    val reader = process.inputStream.bufferedReader()
+                    reader.readLine()
+                    reader.readLine()
```
dimosr, line 45: is this needed now?

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MembershipUtils.kt
yift-r3: I don't see how it will break the e2e-tests (removing the question mark will break it)

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
yift-r3: Not for the test, but I think it's a good idea. The plugin tends to fail too many times because of this.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/Helpers.kt
yift-r3: I think so (not for the test, but it will make the plugin more reliable. At least until we will remove it all together next week)

applications/tools/p2p-test/app-simulator/src/test/kotlin/net/corda/p2p/tests/LargeNetworkTest.kt
yift-r3: Comment added

applications/tools/p2p-test/app-simulator/src/test/kotlin/net/corda/p2p/tests/LargeNetworkTest.kt
yift-r3: It takes about a second to check each member. I changed it to check 20 random members.

It checks the entire network, not just the latest member, so doing it for every fifth member should be okay.

Added output on the time it took. It's really helpful. Thanks.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
dimosr: ok, I think that's being refactored by Arish as part of the other PR atm, but I think he put a relatively high limit there too.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MembershipUtils.kt
dimosr: good point

applications/tools/p2p-test/app-simulator/src/test/kotlin/net/corda/p2p/tests/LargeNetworkTest.kt
dimosr: > It checks the entire network, not just the latest member, so doing it for every fifth member should be okay.

There could be an issue if the number of members is not an exact multiple of 5, in which case the test might report that everything worked ok while data distribution didn't work for some of the last members. But, given we are testing quite large network sizes, an inaccuracy of that granularity is probably not a big deal in this context.

dimosr (approved):
LGTM - at some point, we should probably look into adding something that integrates slightly better with a Jenkins job. But, this should be sufficient for now.

PR #4346, opened 2023-07-20
CORE-15660: Use fewer database connections, and close them afterwards.

chrisr3:
Reuse some database connections while installing the database, and close them when we've finished with them.

This should mostly affect the tests running on Jenkins.


simon-johnson-r3 (approved):

kyriathar (approved):
OK (db)

PR #4347, opened 2023-07-20
CORE-14685: Stub handler for Token Cache events.

chrisr3:
UTXO flows also generate `service.token.event` events. The driver doesn't currently have a use for these, so intercept them and dispatch them to a "logging only" handler.

Also tidy up the flow event loop slightly, using `pollFirst()` instead of `removeFirst()`, and with a single response queue.


ronanbrowne (approved):
LGTM _(build change)_

yarivus (commented):
thx @chrisr3

PR #4349, opened 2023-07-21
CORE-13570 Default datasource properties for vnode connections

kyriathar:
Reads vnode datasource pool configurations (DDL & DML) through configuration read service and offers it to `VirtualNodeDbFactoryImpl`, to be used on vnode datasource configurations creations.

api PR: [#1186](https://github.com/corda/corda-api/pull/1186)


components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDbFactoryImpl.kt
```diff
@@ -215,20 +204,22 @@ internal class VirtualNodeDbFactoryImpl(
     private fun String.toSmartConfig() = smartConfigFactory.create(ConfigFactory.parseString(this))
 }
 
+const val VNODE_POOL_MAX_SIZE = "max_size"
```
simon-johnson-r3: Should these not go in the api repo? Or do we never do that?

kyriathar: Yes good catch, most (if not all) of these config keys constants are in api repo. Will move them there for consistency.

simon-johnson-r3 (dismissed):

driessamyn (commented):
LGTM (needs API version swapped out after API merge).

simon-johnson-r3 (approved):

PR #4350, opened 2023-07-21
CORE-15702 Resolve memory leak when repeatedly trying to establish a session

charlieR3:
e2e test `MembershipSuspensionMultiCluster` `onboard, suspend, fail to communicate, activate member and communicate` is currently failing after suspending a member and attempting to send a flow between and active member and a suspended member. The test never reactivates the member because the test failed before it tried. After this, the link manager's heap space fills up and it eventually crashes due to an OOM error and restarts. 

On inspection of the heap dumps it became clear that instances of `BouncyCastleProvider` were taking the majority of the heap space. This PR remove initialisation of new instances of `BouncyCastleProvider` in favour of referencing by name during session creation so that in this particular scenario, when session establishment is repeated, the heap space isn't filled up.

Before:
Screenshot 2023-07-21 at 11 00 30
Screenshot 2023-07-21 at 11 33 09
Screenshot 2023-07-21 at 11 34 25

After:
Screenshot 2023-07-21 at 13 02 32
Screenshot 2023-07-21 at 13 03 17

### DIFF
--- a/components/link-manager/src/nonOsgiIntegrationTest/kotlin/net/corda/p2p/P2PLayerEndToEndTest.kt
+++ b/components/link-manager/src/nonOsgiIntegrationTest/kotlin/net/corda/p2p/P2PLayerEndToEndTest.kt
@@ -149,6 +149,10 @@ class P2PLayerEndToEndTest {
         }
     }
 
+    init {
+        Security.addProvider(BouncyCastleProvider())
+    }
+
     private val bootstrapConfig = SmartConfigFactory.createWithoutSecurityServices()
         .create(ConfigFactory.empty().withValue(INSTANCE_ID, ConfigValueFactory.fromAnyRef(1)))
         .withValue(MessagingConfig.MAX_ALLOWED_MSG_SIZE, ConfigValueFactory.fromAnyRef(10000000))
@@ -480,9 +484,6 @@ class P2PLayerEndToEndTest {
         checkRevocation: Boolean,
         keyTemplate: KeySchemeTemplate,
     ) : AutoCloseable {
-        init {
-            Security.addProvider(BouncyCastleProvider())
-        }
         private val p2pPort by lazy {
             ServerSocket(0).use {
                 it.localPort
--- a/components/link-manager/src/nonOsgiIntegrationTest/kotlin/net/corda/p2p/P2PLayerEndToEndTest.kt
+++ b/components/link-manager/src/nonOsgiIntegrationTest/kotlin/net/corda/p2p/P2PLayerEndToEndTest.kt
@@ -114,6 +114,7 @@ import java.security.KeyFactory
 import java.security.KeyPairGenerator
 import java.security.KeyStore
 import java.security.PublicKey
+import java.security.Security
 import java.security.Signature
 import java.security.spec.PKCS8EncodedKeySpec
 import java.time.Duration
@@ -479,6 +480,9 @@ class P2PLayerEndToEndTest {
         checkRevocation: Boolean,
         keyTemplate: KeySchemeTemplate,
     ) : AutoCloseable {
+        init {
+            Security.addProvider(BouncyCastleProvider())
+        }
         private val p2pPort by lazy {
             ServerSocket(0).use {
                 it.localPort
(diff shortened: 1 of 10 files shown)


libs/p2p-crypto/src/test/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocolTest.kt
```diff
@@ -29,6 +31,14 @@ class AuthenticationProtocolTest {
     private val partyBMaxMessageSize = 1_500_000
     private val aliceX500Name = MemberX500Name.parse("CN=alice, OU=MyUnit, O=MyOrg, L=London, S=London, C=GB")
 
+    companion object {
+        @BeforeAll
+        @JvmStatic
+        fun setup() {
+            Security.addProvider(BouncyCastleProvider())
+        }
```
chrisr3, line 39: ~Doesn't BouncyCastle automatically register itself as a Security Provider when added to the classpath? This obviously cannot happen _inside_ an OSGi framework, but `AuthenticationProtocolTest` looks like an ordinary JUnit test to me :thinking:.~

OK, apparently not.... :grin:

processors/link-manager-processor/src/main/kotlin/net/corda/processors/p2p/linkmanager/internal/LinkManagerProcessorImpl.kt
```diff
@@ -95,6 +97,8 @@ class LinkManagerProcessorImpl @Activate constructor(
             is BootConfigEvent -> {
                 configurationReadService.bootstrapConfig(event.config)
 
+                Security.addProvider(BouncyCastleProvider())
```
chrisr3, line 100: Perhaps add this provider at the top of the provider list?
```kotlin
Security.insertProviderAt(BouncyCastleProvider(), 1)
```

components/link-manager/src/nonOsgiIntegrationTest/kotlin/net/corda/p2p/P2PLayerEndToEndTest.kt
```diff
@@ -479,6 +480,9 @@ class P2PLayerEndToEndTest {
         checkRevocation: Boolean,
         keyTemplate: KeySchemeTemplate,
     ) : AutoCloseable {
+        init {
+            Security.addProvider(BouncyCastleProvider())
```
yift-r3: Is there a reason not to use the same provider for all the hosts?

processors/link-manager-processor/src/main/kotlin/net/corda/processors/p2p/linkmanager/internal/LinkManagerProcessorImpl.kt
yift-r3: We need our `DelegatedKeystoreProvider` to be the first one.

libs/p2p-crypto/src/test/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocolTest.kt
charlieR3: Yeah I tried testing this first without adding the provider in the link manager processor and it couldn't find the bouncy castle provider. Same for the unit tests.

components/link-manager/src/nonOsgiIntegrationTest/kotlin/net/corda/p2p/P2PLayerEndToEndTest.kt
charlieR3: No reason, no. I'll adjust that. Thanks.

processors/link-manager-processor/src/main/kotlin/net/corda/processors/p2p/linkmanager/internal/LinkManagerProcessorImpl.kt
charlieR3: Thanks @yift-r3. I wasn't aware of that. Do you know why that is out of curiosity?

yift-r3: Because we want to interrupt netty SSL signing to send the signing requests to the crypto worker (instead of doing it in memory by Java or BouncyCastle).

charlieR3: Thanks

yift-r3 (approved):

dimosr (approved):

PR #4351, opened 2023-07-21
CORE-15703 Added singleton object for initializing codesigner cert and test CPB in smoketests

ben-millar:
### Overview
A number of our smoke tests rely on a codesigner certificate and test CPI being present in the test cluster before running. To achieve that today, we use the `@Order` annotation to force some tests which persist these artefacts in `VirtualNodeRestTest` to run first, however this only works when running all the tests. If a developer wants to run a single test, they must remember to manually run the virtual node tests first or their tests will fail.

### Fix
I've added a call in the `@BeforeAll` block of each test class to conditionally upload the code signer cert to the cluster if it's not already present; this will be evaluated when each test class is initialized, meaning the order of tests no longer matters.

In the `VirtualNodeRestTest` class I've instead done this in a `@BeforeEach` block, and defined a `@SkipInitialization` annotation which has been placed on the `can import codesigner certificate()` function, meaning that the first time this test is run, we do not handle the certificate upload extrinsic to the test.

This separation of responsibility between classes and the preservation of the cert and CPB tests was implemented following CR comments.

### Additional Notes
I've also extracted some helper functions to a `ClusterUtils` class, and created a new `TestConstants` file so that some shared constants can be re-used between classes. These, along with the existing `Utils.kt` have been moved to a new `/utils` sub-directory.

I've also reused logic from `testing>e2e-test-utilities` as appropriate within the tests to cut down on code duplication.

### Alternatives considered
I also considered using lazy evaluation of the upload status and the [ExecutionCondition](https://www.baeldung.com/junit-5-conditional-test-execution) feature in junit to intercept test executions and upload the artefacts if not present, but I wasn't able to get the extension auto-detection working to have this apply to all tests in the suite.


lankydan:
So in the ledger extensions repo I did the `@BeforeAll` approach you said by adding:

```kotlin
private val retryTimeout = 120.seconds
private val retryInterval = 1.seconds

fun uploadTrustedCertificate() {
    cluster {
        assertWithRetry {
            // Certificate upload can be slow in the combined worker, especially after it has just started up.
            timeout(retryTimeout)
            interval(retryInterval)
            command { importCertificate(CODE_SIGNER_CERT, "code-signer", "cordadev") }
            condition { it.code == 204 }
        }
    }
}
```

But I wasn't trying to solve uploading an additional CPB like you have done in this PR.

My main issue with this change, is that `private val vNodeInitializer = VirtualNodeInitializer` by itself which is then never used looks pretty odd.

I wonder if it would make more sense to have an `abstract base` test class that has the same information. It could potentially hold the singleton part of your code and possibly even take things like the CPB location so that it could upload the ledger or flow CPBs instead of the default test CPB being uploaded.

One concern I have in general is tests overlapping more. With the current setup each test file runs with a new CPI and set of virtual nodes, preventing overlap between tests outside of the same test class. With this change, there _could_ be conflicts. I say _could_ because I doubt any of these test classes are doing things that could clash with each other.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/flow/FlowTests.kt
```diff
@@ -96,6 +97,8 @@ class FlowTests {
 
         val jacksonObjectMapper = jacksonObjectMapper()
 
+        private val vNodeInitializer = VirtualNodeInitializer
```
josephzunigadaly: We should make it clear when we are running code. A developer looking at this line would never imagine that it triggers certificate import and CPI upload. Can you pull the functionality out into a function so we can see a function call in tests that use it?

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/virtualnode/VirtualNodeInitializer.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    private val groupId = UUID.randomUUID().toString()
+
+    /**
+     * This will run once, the first time the [VirtualNodeInitializer] is referenced in a given test run.
+     * With that in mind, there's no need to do any lazy evaluation to determine if the cert and CPI exist.
+     */
+    init {
+        uploadCertificate()
+        cluster { eventuallyUploadCpi(TEST_CPB_LOCATION, cpiName) }
```
josephzunigadaly: When a test needs a different CorDapp, they will be excluded from using this. If you split this into two, the upload certificate may still be useful to tests that use different CorDapps.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/virtualnode/VirtualNodeInitializer.kt
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * Upload a codesigner certificate to the cluster in order to enable signing functionalities.
+     * The certificate upload process might be slow in the combined worker, especially after
+     * it has just started up. Therefore, we use a retry mechanism to handle potential delays.
+     */
+    private fun uploadCertificate() {
+        cluster {
```
josephzunigadaly: This only imports into cluster B. If you extract the lambda into a `ClusterBuilder` extension method, you'll have a useful method that works on any of the three clusters.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/virtualnode/VirtualNodeRestTest.kt
```diff
@@ -54,44 +47,8 @@ class VirtualNodeRestTest {
...
-            val actualChecksum = getCpiChecksum(cpiName)
-
-            assertThat(actualChecksum).isNotNull.isNotEmpty
-
-            assertThat(cpiHash).withFailMessage(ERROR_CPI_NOT_UPLOADED).isNotNull
-
-            assertThat(actualChecksum).isEqualTo(cpiHash)
-        }
+        private val vNodeInitializer = VirtualNodeInitializer
     }
```
josephzunigadaly, line 69: Please don't delete these tests. You may have copied the code for reuse but we still want these tests in the test report.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/virtualnode/VirtualNodeInitializer.kt
ben-millar: Good point. I plan to full the function `eventuallyUploadCpi()` out into a new utils file so that it can be used either from this `VirtualNodeInitializer`, or from elsewhere.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/virtualnode/VirtualNodeRestTest.kt
ben-millar: Sure! My thinking was that we're now implicitly testing this behaviour anyway, but I can work these back in.

ben-millar: I've re-introduced the tests, and added a new annotation which will skip the certificate pre-req check and upload for the upload codesigner certificate test so it can be evaluated fairly.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/virtualnode/VirtualNodeRestTest.kt
```diff
@@ -152,25 +125,10 @@ class VirtualNodeRestTest {
     fun `can create virtual node with holding id and CPI`() {
         cluster {
             val cpiFileChecksum = getCpiChecksum(cpiName)
-
-            eventuallyCreateVirtualNode(cpiFileChecksum, aliceX500)
+            getOrCreateVirtualNodeFor(aliceX500, cpiFileChecksum)
```
ben-millar: This is not entirely correct as if another test creates another virtual node, this test will essentially be skipped and the existed/cached copy returned. I'll look into this some more.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
```diff
@@ -123,14 +123,8 @@ fun ClusterInfo.getOrCreateVirtualNodeFor(
...
-        val holdingId = createVNodeRequest["requestId"].textValue()
-
-        // Wait for the vNode creation to propagate through the system before moving on
-        eventually(duration = Duration.ofSeconds(30)) {
-            assertThat(getVNode(holdingId).code).isNotEqualTo(404)
-        }
-
-        holdingId
+        val requestId = createVNodeRequest["requestId"].textValue()
+        awaitVirtualNodeOperationStatusCheck(requestId)
```
ben-millar, line 127: A more comprehensive helper function already exists which does the same thing; utilizing that here.

ben-millar:
Thanks so much for the PR comments! Very valid points. I've addressed the comments in my latest changes, but broadly:
- I've made it more explicit that a function is being called when we persist the test artefacts.
- These artefacts are now persisted per-test-class to further avoid any spillover between tests.
- The helper functions have been made more generic and extracted out into a utils class.
- The CPB and cert upload tests have been re-added, and logic has been added to ensure the cert upload helper functions are skipped for the cert test.

josephzunigadaly (dismissed):

josephzunigadaly (approved):

PR #4352, opened 2023-07-22
ES-618: Add Gradle version catalog implementation for corda-runtime-os

knguyen-r3:
**Description:**
As part improving version dependencies within Corda we have added [Gradle Version Catalog](https://docs.gradle.org/current/userguide/platforms.html) for corda-runtime-os. This will allow us to use the newly created central repository [corda-version-catalog](https://github.com/corda/corda-version-catalog) for maintaining and managing 3rd parties dependency that is use within corda-runtime-os.

**Changes:**
Added implementation of using Gradle Version Catalog for corda-runtime-os repo in build.gradle file
Replace all version reference occurrences which use gradle.properties to version catalog in build.gradle files.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,6 @@ artifactoryContextUrl=https://software.r3.com/artifactory
 kotlin.code.style=official
 kotlinVersion=1.8.21
 kotlin.stdlib.default.dependency=false
-kotlinMetadataVersion = 0.6.0
 
 org.gradle.java.installations.auto-download=false
 org.gradle.jvmargs=-Dfile.encoding=UTF-8
@@ -26,32 +25,15 @@ bndVersion=6.4.0
 cordaGradlePluginsVersion=7.0.3
 detektPluginVersion=1.22.+
 internalPublishVersion=1.+
-internalDockerVersion=1.+
 dependencyCheckVersion=0.46.+
-snakeyamlVersion=2.0
 dokkaVersion=1.8.+
-# Implementation dependency versions
-activationVersion=1.2.0
-ariesDynamicFrameworkExtensionVersion=1.3.6
-antlrVersion=2.7.7
-asmVersion=9.4
-avroVersion=1.11.1
-commonsVersion = 1.7
-caffeineVersion = 3.1.6
-commonsLangVersion = 3.12.0
-commonsTextVersion = 1.10.0
-bouncycastleVersion=1.73
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.0.0.xx-SNAPSHOT to pick up maven local published copy
 #cordaApiVersion=5.0.0.xxx-SNAPSHOT
 cordaApiVersion=5.0.0.765-beta+
 
-disruptorVersion=3.4.4
-felixConfigAdminVersion=1.9.26
 felixVersion=7.0.5
 felixScrVersion=2.2.6
-felixSecurityVersion=2.8.4
-guavaVersion=32.1.1-jre
 # Hibernate cannot be upgraded to 6.x due to missing OSGi support
 hibernateVersion=5.6.15.Final
 hikariCpVersion=5.0.1
@@ -170,3 +152,7 @@ snykVersion = 0.4
 # License
 licenseName = The Apache License, Version 2.0
 licenseUrl = http://www.apache.org/licenses/LICENSE-2.0.txt
+
+# Gradle Version Catalog
+cordaVersionCatalog=1.0.0-beta-+
+releaseVersionCatalog=release-5.0
--- a/gradle.properties
+++ b/gradle.properties
@@ -35,18 +35,11 @@ cordaApiVersion=5.0.0.765-beta+
 felixVersion=7.0.5
 felixScrVersion=2.2.6
 # Hibernate cannot be upgraded to 6.x due to missing OSGi support
-hibernateVersion=5.6.15.Final
-hikariCpVersion=5.0.1
-jacksonVersion=2.15.0
-jaxbVersion = 2.3.1
 jbossTransactionApiSpecVersion=1.1.1.Final
 jetbrainsAnnotationsVersion=24.0.1
-kafkaClientVersion=3.4.0_1
 # NOTE: Kryo cannot easily be upgraded as it needs a Quasar change.
 #  Check with one of the group leads before changing.
-kryoVersion = 5.5.0
 kryoSerializersVersion = 0.45
-kotlinCoroutinesVersion=1.6.4
 # Liquibase upgrade to 4.20 and 4.21 failed - see CORE-12612 for more details
 liquibaseVersion = 4.19.0
 # Needed by Liquibase:
--- a/gradle.properties
+++ b/gradle.properties
@@ -39,18 +39,10 @@ jbossTransactionApiSpecVersion=1.1.1.Final
 jetbrainsAnnotationsVersion=24.0.1
 # NOTE: Kryo cannot easily be upgraded as it needs a Quasar change.
 #  Check with one of the group leads before changing.
-kryoSerializersVersion = 0.45
 # Liquibase upgrade to 4.20 and 4.21 failed - see CORE-12612 for more details
-liquibaseVersion = 4.19.0
 # Needed by Liquibase:
-beanutilsVersion=1.9.4
 log4jVersion = 2.20.0
-micrometerVersion=1.11.0
-nettyVersion = 4.1.94.Final
 # com.networknt:json-schema-validator cannot be upgraded beyond 1.0.66 because it becomes dependent on com.ethlo.time which is not OSGi compatible.
-networkntJsonSchemaVersion = 1.0.66
-osgiCmVersion = 1.6.1
-osgiNamespaceServiceVersion = 1.0.0
 osgiServiceComponentVersion = 1.5.1
 osgiUtilFunctionVersion = 1.2.0
 osgiUtilPromiseVersion = 1.3.0
--- a/libs/kotlin-reflection/build.gradle
+++ b/libs/kotlin-reflection/build.gradle
@@ -21,16 +21,16 @@ configurations {
 dependencies {
     api 'org.jetbrains.kotlin:kotlin-osgi-bundle'
     compileOnly 'org.osgi:osgi.annotation'
-    compileOnly "org.jetbrains.kotlinx:kotlinx-metadata-jvm:$kotlinMetadataVersion"
-    compileOnly "org.ow2.asm:asm:$asmVersion"
+    compileOnly libs.kotlinx.metadata
+    compileOnly libs.ow2.asm
     compileOnly "co.paralleluniverse:quasar-osgi-annotations:$quasarVersion"
     implementation platform("net.corda:corda-api:$cordaApiVersion")
 
     testCompileOnly 'org.jetbrains:annotations'
     testImplementation "org.assertj:assertj-core:$assertjVersion"
     testImplementation "org.junit.jupiter:junit-jupiter-params:$junit5Version"
-    testRuntimeOnly "org.jetbrains.kotlinx:kotlinx-metadata-jvm:$kotlinMetadataVersion"
-    testRuntimeOnly "org.ow2.asm:asm:$asmVersion"
+    testRuntimeOnly libs.kotlinx.metadata
+    testRuntimeOnly libs.ow2.asm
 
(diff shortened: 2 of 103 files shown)


settings.gradle
```diff
@@ -58,6 +58,43 @@ plugins {
...
+                    password = settings.ext.find('cordaArtifactoryPassword') ?: System.getenv('CORDA_ARTIFACTORY_PASSWORD')
+                }
+            }
+        }
+    }
+    versionCatalogs {
+        libs {
+            from("net.corda:corda-version-catalog:$cordaVersionCatalog")
+        }
+    }
```
chrisr3: Are you suggesting that all of this project's external dependencies should have their versions controlled by an _external_ artifact, rather than its own `gradle.properties` file?

I'm sorry, but this sounds like an absolutely _**terrible**_ idea! Will developers be expected to wrangle over changing this catalogue like we already do for `corda-api`? (E.g. a Slack channel for people to say "I'm taking the lock!".) And now change _two_ repositories instead of one, just to (say) upgrade ASM 9.5 to 9.6?

I've checked [ES-618](https://r3-cev.atlassian.net/browse/ES-618) and it doesn't even have a description, let alone provide any reasoning for such a disruptive change.

@driessamyn, @mnesbit, @fenryka are you aware of this?

chrisr3:
If we _must_ have a Gradle catalogue then it should be published as an artifact of the `corda-runtime-os` repository. The current catalogue also contains some internal version numbers that are of no interest to third parties, e.g. dependencies which are wrapped or shaded into our own bundles.

Alternatively, we could _not_ do this.

libs/kotlin-reflection/build.gradle
```diff
@@ -48,7 +48,7 @@ def jar = tasks.named('jar', Jar) {
 Bundle-Name: Corda Kotlin Reflection
 Bundle-SymbolicName: \${project.group}.kotlin-reflection
 Sealed: true
--includeresource: @kotlinx-metadata-jvm-${kotlinMetadataVersion}.jar
+-includeresource: @kotlinx-metadata-jvm-${libs.versions.kotlinMetadataVersion.get()}.jar
```
chrisr3, line 51: `kotlinx-metadata-jvm` is an _internal_ dependency that has absolutely _no_ business being advertised in an external catalogue.

gradle.properties
```diff
@@ -170,3 +145,7 @@ snykVersion = 0.4
 # License
 licenseName = The Apache License, Version 2.0
 licenseUrl = http://www.apache.org/licenses/LICENSE-2.0.txt
+
+# Gradle Version Catalog
+cordaVersionCatalog=1.0.0-beta-+
+releaseVersionCatalog=release-5.0
```
chrisr3, line 143: _**YIKES!!!**_ :scream: 

How will you _**guarantee**_ then the next dynamic update _never_ includes a single "breaking change" in _any_ of the (presumably) _many_ projects using the catalog?

settings.gradle
knguyen-r3: Hi Chris, at the moment this is still under development. Here is the [link](https://github.com/corda/platform-eng-design/pull/550) to the design document. Regarding your question, if we are updating ASM from 9.5 to 9.6 we only need to change it once. This can be done by updating the corresponding version of ASM in the toml file in [corda-version-catalog](https://github.com/corda/corda-version-catalog) repo. Then there will be a separate pipeline that will use this Artifactory version and test it by running full build with all the Corda 5 repos. Only when this is built successfully then we can merge the update of version in [corda-version-catalog](https://github.com/corda/corda-version-catalog) repo. No changes will be need to be done in corda-runtime-os for updating version.

gradle.properties
knguyen-r3: As mentioned above all update to version will be done in [corda-version-catalog](https://github.com/corda/corda-version-catalog). Which will then release a unique Artifactory version and we can run full build for all C5 repos which will test for any issue using that newly publish version.

chrisr3: > As mentioned above all update to version will be done in [corda-version-catalog](https://github.com/corda/corda-version-catalog). Which will then release a unique Artifactory version and we can run full build for all C5 repos which will test for any issue using that newly publish version.

Thanks, but that is still a _hideous_ barrier to development, and I don't believe the full implications for Corda developers have been thought through. To my mind, the notion of performing the same "locking dance" as we currently perform for changes to `corda-api` just to update a dependency which is of _zero_ interest to anyone outside `corda-runtime-os` (e.g. `kotlinx-metadata-jvm`) is preposterous. At the moment, I am able to update dependencies quite easily, which allows me to find the time to do it.

Conversely, make it too difficult and I simply _won't_ be able to find the time. And this approach _does_ strike me as too difficult.

knguyen-r3:
> If we _must_ have a Gradle catalogue then it should be published as an artifact of the `corda-runtime-os` repository. The current catalogue also contains some internal version numbers that are of no interest to third parties, e.g. dependencies which are wrapped or shaded into our own bundles.
> 
> Alternatively, we could _not_ do this.

This is a very good point and it was our original design when we first started working on this project. Doing this mean that we will have multiple branches in [corda-version-catalog](https://github.com/corda/corda-version-catalog) and each branch will be corresponding to a release version. However after discussing with @driessamyn we decided that we will have multiples toml file for a branch and each toml file will be corresponding to a release version. This will allow us to update multiple release dependencies version with 1 PR instead of each PR for a release version.

So instead of:
- from("net.corda:corda-version-catalog:release-5.0"), from("net.corda:corda-version-catalog:release-5.1"), ... other versions

We will have:
- from("net.corda:release-5.0:$cordaVersionCatalog") - which will be use by all C5 repos with release 5.0, from("net.corda:release-5.1:$cordaVersionCatalog") - which will be use by all C5 repos with release 5.1, ... other versions

chrisr3:
> This is a very good point and it was our original design when we first started working on this project. Doing this mean that we will have multiple branches in [corda-version-catalog](https://github.com/corda/corda-version-catalog) and each branch will be corresponding to a release version. However after discussing with @driessamyn we decided that we will have multiples toml file for a branch and each toml file will be corresponding to a release version. This will allow us to update multiple release dependencies version with 1 PR instead of each PR for a release version.

I haven't the faintest idea what a TOML file is, but a quick Google search suggests that TOML is a just a file format, e.g. YAML. In which case you appear to be discussing "implementation details", whereas I am discussing the adverse impact this PR will have on Corda development :confused:.

> This will allow us to update multiple release dependencies version with 1 PR instead of each PR for a release version.

What would you be updating here? Is the Build Team now going to be responsible for upgrading Corda to the latest versions of Kotlin, SLF4J, etc?

gradle.properties
```diff
@@ -3,7 +3,6 @@ artifactoryContextUrl=https://software.r3.com/artifactory
 kotlin.code.style=official
 kotlinVersion=1.8.21
 kotlin.stdlib.default.dependency=false
-kotlinMetadataVersion = 0.6.0
```
chrisr3, line 6: This version belongs in `corda-runtime-os`.

gradle.properties
```diff
@@ -26,56 +25,24 @@ bndVersion=6.4.0
...
 # Change to 5.0.0.xx-SNAPSHOT to pick up maven local published copy
 #cordaApiVersion=5.0.0.xxx-SNAPSHOT
 cordaApiVersion=5.0.0.765-beta+
 
-disruptorVersion=3.4.4
-felixConfigAdminVersion=1.9.26
 felixVersion=7.0.5
 felixScrVersion=2.2.6
-felixSecurityVersion=2.8.4
-guavaVersion=32.1.1-jre
```
chrisr3, line 54: I have _no idea_ how you have selected which versions should be included in the catalog, and which should remain in `corda-runtime-os`.

chrisr3:
> We will have:
> 
>    from("net.corda:release-5.0:$cordaVersionCatalog") - which will be use by all C5 repos with release 5.0,
>    from("net.corda:release-5.1:$cordaVersionCatalog") - which will be use by all C5 repos with release 5.1, ... other versions

But if this is what you want, then surely you _**do**_ want to publish the catalog as an artifact of the `corda-runtime-os` build!?

settings.gradle
chrisr3: > if we are updating ASM from 9.5 to 9.6 we only need to change it once

Do you understand that there's likely to be more involved in updating a dependency then just changing one number in some catalogue? It could add a new feature that we want to use, or fix a bug that we would no longer need to work around. It could deprecate or even remove old APIs, resulting in compiler warnings and/or errors. (We consider warnings to be errors anyway.) _In which case, the change to the catalog would need to be coordinated with simultaneous changes to all affected repositories._

We already have this problem coordinating changes between `corda-api` and `corda-runtime-os`, and I find that to be onerous and unpleasant enough. But my reaction to needing to coordinate such changes across _three or more_ repositories is nothing short of complete revulsion.

I'll also point out that projects would also likely be using a library like ASM for completely different reasons (I'm assuming here that you know what ASM is, and understand what it does...) and so we might not _want_ to update it across all projects simultaneously in the first place.

gradle.properties
driessamyn: I don't actually agree with this.
Considering the number of repos (and more to come) and branch combinations, having one central place to update the version is a good thing.
I don't agree that this means things will automagically  update. For some dependencies, that may be true, and build automation will help us here, but for others, the version catalogue itself is versioned, so repos will need to opt-in.

settings.gradle
knguyen-r3: Yes I do, as I said if it's just for updating the version for security or some other reason then we only need to update the corda-version-catalog repo. But if changes need to be made in corda-runtime-os, then we need to use the artifact version published in corda-version-catalog development and it might involve 2 repos.

gradle.properties
knguyen-r3: At the moment, I'm suppose to remove third party dependencies in these repos. And for some version is quite unclear.

knguyen-r3:
> > We will have:
> > from("net.corda:release-5.0:$cordaVersionCatalog") - which will be use by all C5 repos with release 5.0,
> > from("net.corda:release-5.1:$cordaVersionCatalog") - which will be use by all C5 repos with release 5.1, ... other versions
> 
> But if this is what you want, then surely you _**do**_ want to publish the catalog as an artifact of the `corda-runtime-os` build!?

What is the point of publish for corda-runtime-os at the moment? The reason we are doing this is to have a share corda-version-catalog which will have the share versions that will be use in **all** C5 repos. In the future we might add a corda-runtime-os if it's needed however for current implementation it's not needed. We are extracting it to a **single** repo, to have new artifactory in corda-runtime-os, corda-api, ... is no different to current implementation. Where we have to manage **many** repositories gradle.properties file. If you look at other PR where I have for this version catalog change, https://github.com/corda/corda-e2e-tests/pull/157, https://github.com/corda/corda-cli-plugin-host/pull/183 . It's all reference to the same corda-version-catalog Artifactory

settings.gradle
chrisr3: > as I said if it's just for updating the version for security or some other reason then we only need to update the corda-version-catalog repo

"Security" may be a _reason_ for updating something, but there is absolutely no reason to assume that the updated version will be a "drop in" replacement.

Every time someone says "we only need to update the corda-version-catalog repo", the more convinced I become that this change is ill-conceived and must be abandoned.

driessamyn: This doesn't assume things are drop-in replacements, but when they are, it's a single change rather than a change in x * y places, where x is the number of repos (min 5, more to be added) and y is the number of supported branches.
The proposal leaves flexibility to update the version for a particular release version and in a drop-in or opt-in way, so gives us that flexibility.

chrisr3: > This doesn't assume things are drop-in replacements, but when they are, it's a single change rather than a change in x * y places, where x is the number of repos (min 5, more to be added) and y is the number of supported branches. The proposal leaves flexibility to update the version for a particular release version and in a drop-in or opt-in way, so gives us that flexibility.

Except you don't know when an upgrade is not going to be a drop-in replacement until it suddenly isn't, e.g. that JUnit 5.9.2 -> 5.9.3 upgrade, which also happened to break all `@ParameterizedTest`s that used `@MethodSource`, but _only_ for those projects with `TestOSGi` tasks.

It's like you're suggesting we stop wearing seat-belts in cars, because we aren't _always_ involved in a road traffic accidents.

driessamyn: No. Precisely not. CI is our seatbelt in this case. 
We can think it's a drop in replacement, but no harm done for thinking wrong because CI will highlight there is an issue (granted, we have to figure out where this happens, and this has been discussed).

PR #4353, opened 2023-07-22
CORE-8434: Create an OSGi requirement for sandboxes-testkit.

chrisr3:
Define a `@RequireSandboxTestkit` annotation so that clients of `:testing:sandboxes-testkit` can declare an explicit OSGi dependency. Otherwise Bnd's resolver may not decide to include the testkit in the bundles passed to the `TestOSGi` task.

The point is that the new `@RequireSandboxTestkit` annotation creates a new OSGi requirement that only `:testing:sandboxes-testkit` can satisfy. Hence annotating anything as `@RequireSandboxTestkit` _forces_ the Bnd resolver to include the testkit bundle.

This is all caused by a limitation of Bnd's `effective:=active` resolution: it can only notice that an instance of a particular service _exists_, and not whether that instance will be enabled at runtime.


dickon (approved):
LGTM, other than a dislike of Bnd, but since we've got precedent in corda-runtime-os and you've got a decent description the ticket so be it.

kyriathar (approved):
OK (db & sandboxing)

PR #4359, opened 2023-07-25
CORE-15182 - Added a check that ensures no contract state type has more than one UTXO token observer

filipesoliveira:
The verification is performed when the sandbox is created. This means the verification is only done during runtime and a flow can failed because more than one UTXO token observer was found. This is not ideal but can be improved in the future.

Refactored the code so the values of the container that maps a `token state` to `UTXO token observer` is a single instance and not a list.

### DIFF
--- a/components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
+++ b/components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
@@ -194,16 +194,16 @@ class EntitySandboxServiceImpl @Activate constructor(
     ): Map<Class<ContractState>, UtxoLedgerTokenStateObserver<ContractState>?> {
 
         return tokenStateObserverMap.entries.associate { contractStateTypeToObserver  ->
-            val numberOfObservers = entry.value.size
+            val numberOfObservers = contractStateTypeToObserver.value.size
 
             if (numberOfObservers > 1) {
-                val observerTypes = contractStateTypeToObserver.value.map { it.stateType.name }
+                val observerTypes = contractStateTypeToObserver.value.map { observer -> observer.stateType.name }
                 throw IllegalStateException(
                     "More than one observer found for the contract state. " +
-                            "Contract state: ${entry.key}, observers: $observerTypes"
+                            "Contract state: ${contractStateTypeToObserver.key}, observers: $observerTypes"
                 )
             }
-            contractStateTypeToObserver.key to entry.value.singleOrNull()
+            contractStateTypeToObserver.key to contractStateTypeToObserver.value.singleOrNull()
         }
     }
 
--- a/components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
+++ b/components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
@@ -182,17 +182,19 @@ class EntitySandboxServiceImpl @Activate constructor(
 
         validateTokenStateObservers(tokenStateObserverMap)
 
-        // Create a map that contains a single UTXO token observer for each token state
-        // This away, people know that the there will never be more than one UTXO token observer
-        val validatedTokenStateObserverMap = tokenStateObserverMap.entries.associate { it.key to it.value.singleOrNull() }
+        ctx.putObjectByKey(SANDBOX_TOKEN_STATE_OBSERVERS, requireSingleObserverToState(tokenStateObserverMap))
 
-        ctx.putObjectByKey(SANDBOX_TOKEN_STATE_OBSERVERS, validatedTokenStateObserverMap)
         logger.debug {
             "Registered token observers: ${tokenStateObserverMap.mapValues { (_, observers) ->
                 observers.map { it::class.java.name }}
             }"
         }
     }
+
+    private fun requireSingleObserverToState(
+        tokenStateObserverMap: Map<Class<ContractState>, List<UtxoLedgerTokenStateObserver<ContractState>>>
+    ) =
+        tokenStateObserverMap.entries.associate { it.key to it.value.singleOrNull() }
 
     private fun validateTokenStateObservers(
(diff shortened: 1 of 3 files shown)


owenstanford (dismissed):
LGTM

components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
```diff
@@ -179,14 +179,38 @@ class EntitySandboxServiceImpl @Activate constructor(
...
+    private fun validateTokenStateObservers(
+        tokenStateObserverMap: Map<Class<ContractState>, List<UtxoLedgerTokenStateObserver<ContractState>>>
+    ) {
+
+        tokenStateObserverMap.forEach { contractStateTypeToObservers ->
+            val numberOfObservers = contractStateTypeToObservers.value.size
+
+            if (numberOfObservers > 1) {
+                val observerTypes = contractStateTypeToObservers.value.map { it.stateType.toString() }
+                throw SandboxException(
```
lankydan: Not sure if we should throw a sandbox exception or an illegal state. Can you check how we do it for registering vault named query factories, since we should try keep them consistent. I'll look later but I'm out atm.

components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
```diff
@@ -179,14 +179,38 @@ class EntitySandboxServiceImpl @Activate constructor(
...
 
+    private fun validateTokenStateObservers(
+        tokenStateObserverMap: Map<Class<ContractState>, List<UtxoLedgerTokenStateObserver<ContractState>>>
+    ) {
+
+        tokenStateObserverMap.forEach { contractStateTypeToObservers ->
+            val numberOfObservers = contractStateTypeToObservers.value.size
+
+            if (numberOfObservers > 1) {
+                val observerTypes = contractStateTypeToObservers.value.map { it.stateType.toString() }
```
lankydan: I'd do name of the type rather than tostring, it prints out better

components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
```diff
@@ -179,14 +179,38 @@ class EntitySandboxServiceImpl @Activate constructor(
             .toSet()
             .mapNotNull { getObserverFromClassName(it, ctx) }
             .groupBy { it.stateType }
-        ctx.putObjectByKey(SANDBOX_TOKEN_STATE_OBSERVERS, tokenStateObserverMap)
+
+        validateTokenStateObservers(tokenStateObserverMap)
+
+        // Create a map that contains a single UTXO token observer for each token state
```
lankydan: If you name the validate method, something like requireSingleObserverToState, then you can remove the comment.

components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
filipesoliveira: If you are referring to the file `VaultNamedQueryRegistryImpl.kt`, it uses `IllegalArgumentException`.

filipesoliveira: I did not manage to find the file you are talking about. But illegal state makes more sense.

lankydan: 👍

components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
```diff
@@ -182,18 +182,20 @@ class EntitySandboxServiceImpl @Activate constructor(
 
         validateTokenStateObservers(tokenStateObserverMap)
 
-        // Create a map that contains a single UTXO token observer for each token state
-        // This away, people know that the there will never be more than one UTXO token observer
-        val validatedTokenStateObserverMap = tokenStateObserverMap.entries.associate { it.key to it.value.singleOrNull() }
+        ctx.putObjectByKey(SANDBOX_TOKEN_STATE_OBSERVERS, requireSingleObserverToState(tokenStateObserverMap))
```
lankydan, line 183: Oh sorry I meant renaming `validateTokenStateObservers` to what `requireSingleObserverToState`, maybe this suggests that we can combine those two methods anyway.

filipesoliveira: Yes, that makes sense. Your comment was clear enough. I just did not read it properly.

lankydan (dismissed):

components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
```diff
@@ -193,19 +193,18 @@ class EntitySandboxServiceImpl @Activate constructor(
         tokenStateObserverMap: Map<Class<ContractState>, List<UtxoLedgerTokenStateObserver<ContractState>>>
     ): Map<Class<ContractState>, UtxoLedgerTokenStateObserver<ContractState>?> {
 
-        tokenStateObserverMap.forEach { contractStateTypeToObservers ->
-            val numberOfObservers = contractStateTypeToObservers.value.size
+        return tokenStateObserverMap.entries.associate { entry  ->
```
kyriathar: Think I 'd keep previous naming i.e. `contractStateTypeToObservers`(?)

kyriathar (approved):
LGTM

owenstanford (approved):

lankydan (approved):

PR #4361, opened 2023-07-25
CORE-13375: use singular naming for Certificate and Key rest resources

Tom-Fitzpatrick:
This PR changes the following resources to have a singular naming pattern, in line with our other resources:

/certificates/*    ->    /certificate/*
/keys/*              ->    /key/*

This is handled using the new Rest versioning strategy as described here: [Versioning of REST API]( https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4472635529/Versioning+of+REST+API )

In short, CertificatesRestResource, CertificatesRestResourceImpl, KeysRestResource, KeysRestResourceImpl have had their versions limited to `maxVersion = RestApiVersion.C5_0` and have been marked as @Deprecated

And new versions for CertificateRestResource, CertificateRestResourceImpl, KeyRestResource, KeyRestResourceImpl have been created with `minVersion = RestApiVersion.C5_1`

The uses of /certificates/* and /keys/*  across the repo have been updated to point to the new locations. 

An accompanying change to add a small e2e test to verify that the deprecated resources at version RestApiVersion.C5_0 are still functioning here: https://github.com/corda/corda-e2e-tests/pull/173


vkolomeyko (dismissed):
LGTM

vkolomeyko (dismissed):
LGTM

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/deprecated/CertificatesRestResourceImpl.kt
```diff
@@ -119,6 +121,8 @@ class CertificatesRestResourceImpl @Activate constructor(
         subjectAlternativeNames: List<String>?,
         contextMap: Map<String, String?>?,
     ): String {
+        logger.warn("Deprecated, please use next version at /certificate/{tenantId}/{keyId}")
```
yift-r3, line 129: Can the functions here log the warning and call the functions in `CertificateRestResourceImpl`?

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/deprecated/KeysRestResourceImpl.kt
```diff
@@ -65,11 +67,15 @@ class KeysRestResourceImpl @Activate constructor(
     override fun listSchemes(
         tenantId: String,
         hsmCategory: String,
-    ): Collection<String> = tryWithExceptionHandling(logger, "list supported schemes for tenant $tenantId") {
-        cryptoOpsClient.getSupportedSchemes(
-            tenantId = tenantId,
-            category = hsmCategory.uppercase()
-        )
+    ): Collection<String> {
```
yift-r3, line 75: Same question?

components/membership/membership-rest-impl/src/test/kotlin/net/corda/membership/impl/rest/v1/KeyRestResourceImplTest.kt
```diff
@@ -1,3 +1,4 @@
+@file:Suppress("DEPRECATION")
```
yift-r3: Why is this deprecated?

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/CertificateRestResource.kt
```diff
@@ -0,0 +1,310 @@
...
+ */
+@HttpRestResource(
+    name = "Certificate API",
+    description = "The Certificates API consists of endpoints used to work with certificates and related operations. " +
+        "The API allows you to import a certificate chain, and generate a certificate signing request (CSR) to be" +
+        " submitted to a certificate authority (CA).",
+    path = "certificate",
+    minVersion = RestApiVersion.C5_1
+)
+interface CertificateRestResource : RestResource {
```
yift-r3, line 26: Can't this and the deprecated one inherit from a common interface?

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/CertificatesRestResource.kt
```diff
@@ -8,18 +8,22 @@ import net.corda.rest.annotations.HttpPUT
...
  * The Certificates API consists of endpoints used to work with certificates and related operations. The API allows you
  * to import a certificate chain, and generate a certificate signing request (CSR) to be submitted to a certificate
  * authority (CA).
  */
+@Deprecated("Deprecated in favour of CertificateRestResource")
 @HttpRestResource(
     name = "Certificates API",
     description = "The Certificates API consists of endpoints used to work with certificates and related operations. " +
         "The API allows you to import a certificate chain, and generate a certificate signing request (CSR) to be" +
         " submitted to a certificate authority (CA).",
```
yift-r3, line 27: Shouldn't we add some warning in the description?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -235,7 +235,7 @@ abstract class BaseOnboard : Runnable, RestCommand() {
             }
         }
 
-        val response = createRestClient(KeysRestResource::class).use { keyClient ->
+        val response = createRestClient(KeyRestResource::class).use { keyClient ->
```
yift-r3, line 238: @r3-arish , @YashNabar - FYI, you might need to resolve some conflicts.

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/deprecated/CertificatesRestResourceImpl.kt
Tom-Fitzpatrick: That's a good point, seeing as we are only changing the endpoint name with this, I think it makes sense to call the functions from `CertificateRestResourceImpl` as you suggest. 

It will still be easy to delete this file once we decide not to support C5_0 endpoints. I'll look into this now. Thanks for the review!

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/deprecated/KeysRestResourceImpl.kt
Tom-Fitzpatrick: Same, I'll make this change now.

components/membership/membership-rest-impl/src/test/kotlin/net/corda/membership/impl/rest/v1/KeyRestResourceImplTest.kt
Tom-Fitzpatrick: Ah, it shouldn't be, thanks for catching that I'll remove it now.

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/CertificateRestResource.kt
Tom-Fitzpatrick: Hmm, to have the methods defined in the parent interface? What would the advantage be? We would still need to annotate each of them separately I believe, in `CertificateRestResource `and `CertificatesRestResource`, so that they cover separate versions. 

I think this might also add complexity, if for example we add another version that removes an endpoint down the line. In that scenario, we would need to remove it from the parent interface, and add it only in the deprecated inheriting interface. 

Maybe I'm missing something here though?

yift-r3: OK

components/membership/membership-rest/src/main/kotlin/net/corda/membership/rest/v1/CertificatesRestResource.kt
Tom-Fitzpatrick: Sure, that's a good point. I'll add a warning in the description here and in the KeysRestResource and implementations. Thanks

josephzunigadaly (dismissed):

components/membership/membership-rest-impl/src/main/kotlin/net/corda/membership/impl/rest/v1/deprecated/CertificatesRestResourceImpl.kt
Tom-Fitzpatrick: @yift-r3 @vkolomeyko 
I looked into calling the functions in the new versions of these resources inside the deprecated ones as we discussed here. However it seemed to be causing issues with the endpoints that were being served for the deprecated resources at `v1/certificates` and `v1/keys`. I think it could be fixed easily enough, but it would be a slightly more involved solution than simply instantiating CertificateRestResource and using it as we'd hoped. 

For this reason, I think it's better to leave the full implementations of KeysRestResourceImpl and CertificatesRestResourceImpl as they were originally. 

I chatted to Viktor about this too, and we see a few other benefits to this also:
- It prevents changes to the new implementation affecting the old one in unexpected ways, particularly for any small changes that wouldn't merit a new version and might just be updated in the new impl, like changing the order of a returned list etc
- Keeps the two versions totally separate and decoupled 
- It does add more code to be conscious of, but the deprecated implementations shouldn't really have to be touched much from this point onwards, they've been active and covered by tests for a long time, so we should have confidence in them as they are now and be able to work in the newest versions of these endpoints.

yift-r3 (approved):
LGTM

josephzunigadaly (approved):

PR #4364, opened 2023-07-25
CORE-15817 Add `getOrCreateEntityManagerFactory` API by connection ID

kyriathar:
We used to acquire vnode EMFs by connection name and DB privilege, because that API is also caching the EMF to be re-used across different request for the same vnode (whereas the existing API to acquire a vnode EMF by connection ID is not caching the EMF). Both connection name and DB privilege, however, are esoteric and shouldn't be used for vnode EMFs creations.

- undo [previous work](https://github.com/corda/corda-runtime-os/pull/3916) which was using `getOrCreateEntityManagerFactory` by connection name and DB privilege
- introduce new `getOrCreateEntityManagerFactory` API by connection ID, which also caches the EMF
-  use the new `getOrCreateEntityManagerFactory` API in vnode EMFs creation call sites

### DIFF
--- a/components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DbConnectionOpsCachedImpl.kt
+++ b/components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DbConnectionOpsCachedImpl.kt
@@ -19,7 +19,7 @@ class DbConnectionOpsCachedImpl(
 
     // We should try merging the below two caches into one to, like so, make sure each connection gets one EMF only,
     //  otherwise (i.e. if we get duplicate EMFs for same connection) we end up leaking memory with
-    //  duplicate entity proxies loaded in the class loader as identified in https://r3-cev.atlassian.net/browse/CORE-15806.
+    //  duplicate entity proxies loaded in the class loader as identified in CORE-15806.
     private val cache = ConcurrentHashMap<Pair<String,DbPrivilege>, EntityManagerFactory>()
 
     private val cacheByConnectionId = ConcurrentHashMap<UUID, EntityManagerFactory>()
--- a/components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DbConnectionOpsCachedImpl.kt
+++ b/components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DbConnectionOpsCachedImpl.kt
@@ -17,12 +17,11 @@ class DbConnectionOpsCachedImpl(
     private val entitiesRegistry: JpaEntitiesRegistry
     ): DbConnectionOps by delegate {
 
-    // TODO - replace with caffeine cache
+    // We should try merging the below two caches into one to, like so, make sure each connection gets one EMF only,
+    //  otherwise (i.e. if we get duplicate EMFs for same connection) we end up leaking memory with
+    //  duplicate entity proxies loaded in the class loader as identified in https://r3-cev.atlassian.net/browse/CORE-15806.
     private val cache = ConcurrentHashMap<Pair<String,DbPrivilege>, EntityManagerFactory>()
 
-    // TODO Maybe we could consider replacing the above cache with the one below. All `db_connection`s have/ get an ID.
-    //  Currently the below cache is not cleared on overwriting a connection (`putConnection`).
-    //  In theory I think that should be OK for now since we don't allow "re-creating" a vnode (I believe).
     private val cacheByConnectionId = ConcurrentHashMap<UUID, EntityManagerFactory>()
 
     private fun removeFromCache(name: String, privilege: DbPrivilege) {
@@ -67,7 +66,6 @@ class DbConnectionOpsCachedImpl(
         connectionId: UUID,
         entitiesSet: JpaEntitiesSet
     ): EntityManagerFactory {
-        // TODO Should we be preventing DDL connections from being cached?
         return cacheByConnectionId.computeIfAbsent(connectionId) {
             delegate.createEntityManagerFactory(connectionId, entitiesSet)
         }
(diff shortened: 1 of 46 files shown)


components/db/db-connection-manager/src/main/kotlin/net/corda/db/connection/manager/DbConnectionOps.kt
```diff
@@ -140,4 +140,6 @@ interface DbConnectionOps {
      * @return
      */
     fun createEntityManagerFactory(connectionId: UUID, entitiesSet: JpaEntitiesSet): EntityManagerFactory
+
+    fun getOrCreateEntityManagerFactory(connectionId: UUID, entitiesSet: JpaEntitiesSet): EntityManagerFactory
```
chrisr3, line 144: _Another_ variant of `getOrCreateEntityManagerFactory`? We already have _two_:
```kotlin
fun getOrCreateEntityManagerFactory(db: CordaDb, privilege: DbPrivilege): EntityManagerFactory

fun getOrCreateEntityManagerFactory(name: String, privilege: DbPrivilege, entitiesSet: JpaEntitiesSet): EntityManagerFactory
```

It worries me that a new function with such a _very_ similar signature could be doing something radically different :worried:.

But as I understand it, _every_ new `EntityManagerFactory` instance will allocate memory that can never be reclaimed without also unloading those entities' host bundle. Which means that we _cannot_ allow the platform to create an open-ended number of `EntityManagerFactory` instances outside of a sandbox.

kyriathar: ~~Sounds to me like a public sandbox references types (entities) belonging to user bundles? So could it be we cannot unload those user bundles then till we free the respective EMFs memory?~~

Back to this PR, I believe it doesn't change the above behavior (i.e. how we are caching the EMFs).

It's that we didn't want to get vnode EMFs via connection name and DB privilege as these are quite internal/ esoteric but rather the more appropriate pointer which is the connection ID.

sprintengine: I tested using the EMF cache from the `VirtualNodeReconciliationContext`, so reconcilers no longer re-created EMFs for every reconciliation run, for every vnode. I can confirm it fixes the mem leak described in https://r3-cev.atlassian.net/browse/CORE-15806:

A comparison, about 40 minutes apart, the reconcilers in question are running every 5 seconds:
```
after 10min
Loaded classes: 26,103
Allocated All Pools: 345 MB
Used Metaspace: 190 MB
Used CodeHeap 'profiled nmethods': 74 MB
Used CodeHeap 'non-profiled nmethods': 37 MB
Used Compressed Class Space: 17 MB
Used CodeHeap 'non-nmethods': 1.7 MB

after 50min
Loaded classes: 26,103
Allocated All Pools: 346 MB
Used Metaspace: 190 MB
Used CodeHeap 'profiled nmethods': 73 MB
Used CodeHeap 'non-profiled nmethods': 37 MB
Used Compressed Class Space: 17 MB
Used CodeHeap 'non-nmethods': 1.7 MB
```

simon-johnson-r3: Discussed offline. We're going to go for a formalised design which limits to 1 EMF per row in the db connections db via an EMF store. The api which is called to get an EMF would have no effect on the number of EMFs and essentially would do the same underneath, it's simply looking up the row/EMF using different data. It's not quite a cache, although in principal it will exhibit the same behaviour we see in this PR.

simon-johnson-r3 (changes requested):

components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DbConnectionOpsCachedImpl.kt
```diff
@@ -20,6 +20,11 @@ class DbConnectionOpsCachedImpl(
     // TODO - replace with caffeine cache
     private val cache = ConcurrentHashMap<Pair<String,DbPrivilege>, EntityManagerFactory>()
 
+    // TODO Maybe we could consider replacing the above cache with the one below. All `db_connection`s have/ get an ID.
```
simon-johnson-r3: If we want to submit this PR as a first step to alleviate the user experience problem, I would at least make this comment accurate in terms of the direction we wish to go in.

components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DbConnectionOpsCachedImpl.kt
```diff
@@ -17,9 +17,13 @@ class DbConnectionOpsCachedImpl(
     private val entitiesRegistry: JpaEntitiesRegistry
     ): DbConnectionOps by delegate {
 
-    // TODO - replace with caffeine cache
+    // We should try merging the below two caches into one to, like so, make sure each connection gets one EMF only,
+    //  otherwise (i.e. if we get duplicate EMFs for same connection) we end up leaking memory with
+    //  duplicate entity proxies loaded in the class loader as identified in https://r3-cev.atlassian.net/browse/CORE-15806.
```
yift-r3: Can we avoid the link to Jira in the source code? CORE-15806 should be enough.

components/membership/certificates-service-impl/src/integrationTest/kotlin/net/corda/membership/certificate/service/impl/CertificatesServiceImplTest.kt
```diff
@@ -79,7 +86,7 @@ internal class CertificatesServiceImplTest {
             dbConnectionManagerMock,
             jpaEntitiesRegistryMock,
             mock(),
-        )
+            virtualNodeInfoReadServiceMock)
```
yift-r3: Nitpick: can you add a comma and a new line?

components/membership/certificates-service-impl/src/test/kotlin/net/corda/membership/certificate/service/impl/DbClientImplTest.kt
```diff
@@ -136,17 +157,6 @@ class DbClientImplTest {
 
     @Test
     fun `retrieveCertificates will not close the node factory`() {
```
yift-r3: The name of this test should be `retrieveCertificates will not close the cluster factory`

simon-johnson-r3 (dismissed):
Approving on the basis this works in principle like our desired solution. That should be recorded in another JIRA ticket, but will not make substantial differences and it's better this is resolved sooner rather than later.

yift-r3 (approved):
LGTM

simon-johnson-r3 (approved):

PR #4365, opened 2023-07-26
CORE-15386 Modify ledger named queries main query to simplify post processing

nkovacsx:
### Overview

The main query for ledger named queries has been modified so the post processing can be simpler.

### Test plan

Existing unit, e2e tests. Testing the query manually for multiple database scenarios.


lankydan (commented):
Change looks good to me.

Discussed with @nkovacsx to do similar changes to the `findUnconsumedStates` and `resolveStates` which will greatly simplify our code.

relyafi:
> Change looks good to me.
> 
> Discussed with @nkovacsx to do similar changes to the `findUnconsumedStates` and `resolveStates` which will greatly simplify our code.

We may well change these as part of the multi-DB work anyway, but the SQL will be pretty much identical (assuming we use HQL instead of entities to find) so probably not wasted work.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -92,65 +91,45 @@ class VaultNamedQueryExecutorImpl(
...
+
+            request.parameters.filter { it.value != null }.forEach { rec ->
+                val bytes = rec.value.array()
+                query.setParameter(rec.key, serializationService.deserialize(bytes))
             }
 
+            query.firstResult = request.offset
+            query.maxResults = request.limit
+
+            query.resultList as List<Tuple>
```
relyafi, line 126: I think the cast may be redundant, as we've already specified the query return type as a tuple.

relyafi (dismissed):
One minor comment but not necessarily worth another change for, otherwise LGTM

vlajos (dismissed):

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -92,65 +91,47 @@ class VaultNamedQueryExecutorImpl(
...
+                        tc_output_info.data as output_info_data
+                        tc_output.data AS output_data FROM
+                        $UTXO_VISIBLE_TX_TABLE AS visible_states
+                        JOIN $UTXO_TX_COMPONENT_TABLE AS tc_output_info
+                             ON visible_states.transaction_id = tc_output_info.transaction_id
+                             AND visible_states.leaf_idx = tc_output_info.leaf_idx
+                             AND tc_output_info.group_idx = ${UtxoComponentGroup.OUTPUTS_INFO.ordinal}
+                        JOIN $UTXO_TX_COMPONENT_TABLE AS tc_output
+                             ON tc_output_info.transaction_id = tc_output.transaction_id
+                             AND tc_output_info.leaf_idx = tc_output.leaf_idx
```
vlajos, line 109: I would use `visible_states.XXX` on the left side of these.
But I am not too sure if it has any real impact. (I suspect it can help influence the query optimizer to avoid two deep nested join in favour of two one deeps.)
(Same for the other queries.)

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -4,7 +4,6 @@ import net.corda.data.KeyValuePair
 import net.corda.data.KeyValuePairList
 import net.corda.data.persistence.EntityResponse
 import net.corda.data.persistence.FindWithNamedQuery
-import net.corda.ledger.persistence.common.ComponentLeafDto
```
vlajos, line 7: Does this PR remove all usages of `ComponentLeafDto`?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
```diff
@@ -62,45 +62,18 @@ class UtxoPersistenceServiceImpl(
...
-            ?: emptyMap()
-        return componentGroups[outputsIdx]?.mapNotNull {
-            val info = outputInfos[it.transactionId to it.leafIndex]
-            requireNotNull(info) {
-                "Missing output info at index [${it.leafIndex}] for UTXO transaction with ID [${it.transactionId}]"
-            }
+        return entityManagerFactory.transaction { em ->
+            repository.findUnconsumedVisibleStatesByType(em)
+        }.filter {
             val contractState = serializationService.deserialize<ContractState>(it.data)
```
lankydan, line 106: So 💩, glad we know how to remove this line of code in a change _hopefully_ coming soon.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
```diff
@@ -103,68 +104,78 @@ class UtxoRepositoryImpl @Activate constructor(
...
+                    AND tc_output_info.leaf_idx = tc_output.leaf_idx
+                    AND tc_output.group_idx = ${UtxoComponentGroup.OUTPUTS.ordinal}
                 JOIN {h-schema}utxo_transaction_status AS ts
-                    ON ts.transaction_id = tc.transaction_id
-                WHERE tc.group_idx IN (:groupIndices)
+                    ON ts.transaction_id = tc_output.transaction_id
                 AND rts.consumed IS NULL
                 AND ts.status = :verified
-                ORDER BY tc.group_idx, tc.leaf_idx""",
+                ORDER BY tc_output.created, tc_output.transaction_id, tc_output.leaf_idx, tc_output.group_idx
```
lankydan: Do we need ordering by `group_idx`? Since we've removed the separate rows for state infos and `group_idx` isn't included anywhere in the SELECT. Same for the other queries.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
```diff
@@ -103,68 +104,78 @@ class UtxoRepositoryImpl @Activate constructor(
...
             """
-                SELECT tc.transaction_id, tc.group_idx, tc.leaf_idx, tc.data
-                FROM {h-schema}utxo_transaction_component AS tc
-                JOIN {h-schema}utxo_visible_transaction_state AS rts
-                    ON rts.transaction_id = tc.transaction_id
-                    AND rts.leaf_idx = tc.leaf_idx
+                SELECT tc_output.transaction_id, 
+				tc_output.leaf_idx, 
+				tc_output_info.data as output_info_data,
+                tc_output.data AS output_data FROM
```
lankydan: Formatting is odd.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
relyafi: Did you mean that the `tc_output` joins should reference visible states instead of `tc_output_info`? We could probably test this using either the PSQL command line or a tool like DBeaver and run the queries with `EXPLAIN ANALYSE` to see if there's a difference.

Also, personally from a style perspective I prefer seeing the "target" table on the LHS rather than RHS, because for me that reflects that we are "assigning" the key fields of the target table based on the specified values. (Comment applies to all queries)

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
relyafi: Agreed - I think generally it makes sense that we should be ordering based on the "primary" table fields given the 1-1 mappings between tables.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
vlajos: @relyafi 
> Did you mean that the `tc_output` joins should reference visible states instead of `tc_output_info`?

Yes.

> We could probably test this using either the PSQL command line or a tool like DBeaver and run the queries with `EXPLAIN ANALYSE` to see if there's a difference.

This may depend on the data in the tables and on the general state of the database.
(Maybe my whole comment is premature optimization...)

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
nkovacsx: Good point, modifying

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
nkovacsx: not sure what happened here, looks good in the IDE

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
nkovacsx: Yeah, good point. Deleted the class completely.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
nkovacsx: No unfortunately, JPA returns `List getResultList();` a List with no generic type here so we need to explicitly cast it.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
relyafi: > This may depend on the data in the tables and on the general state of the database. (Maybe my whole comment is premature optimization...)

True, and agree. In the absence of data (or a need to optimize) my view would be either approach is fine right now, I'd just favor making the code as easy to understand as possible.

nkovacsx: Will leave it as is for now, we can get back to it later if needed.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
```diff
@@ -103,68 +104,78 @@ class UtxoRepositoryImpl @Activate constructor(
...
         return entityManager.createNativeQuery(
             """
-                SELECT tc.transaction_id, tc.group_idx, tc.leaf_idx, tc.data
-                FROM {h-schema}utxo_transaction_component AS tc
-                JOIN {h-schema}utxo_visible_transaction_state AS rts
-                    ON rts.transaction_id = tc.transaction_id
-                    AND rts.leaf_idx = tc.leaf_idx
+                SELECT tc_output.transaction_id, 
+			    tc_output.leaf_idx, 
+			    tc_output_info.data as output_info_data,
```
lankydan, line 113: this formatting is weird also.

lankydan (dismissed):
Some formatting things, but fine to approve.

vlajos (dismissed):

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
lankydan: Confirmed with @nkovacsx that GH is displaying this weird for some reason.

lankydan (dismissed):

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
relyafi: @lankydan @nkovacsx there's no tabs in there right? that might be an explanation.

relyafi (approved):
LGTM

PR #4366, opened 2023-07-26
CORE-15771: Update Ledger JPA entities for testing datamodel.

chrisr3:
Java 17 no longer allows non-`static` `final` fields to be modified via Java reflection. Alter the definitions of these JPA entities so that Hibernate can update them, and no longer needs to use reflection anyway.


relyafi (approved):
The changes look fine to me.

This will cause a merge conflict with my UTXO ledger POC (as that has moved all of the UTXO entities) but I think the changes are sufficiently simple to not cause a big headache, so we should be ok to merge these as-is.

PR #4368, opened 2023-07-26
CORE-15770: Update CLI plugin's SqlFormatter to understand Hibernate property access.

chrisr3:
Update `SqlFormatter` to check for JPA property annotations on both the underlying field and the getter method.

Log a warning for each `val` property that has JPA annotations. This will become an error once all JPA entities have been updated.

### DIFF
--- a/tools/plugins/initial-config/src/test/kotlin/net/corda/cli/plugin/initialconfig/TestSqlFormattersForFieldAccess.kt
+++ b/tools/plugins/initial-config/src/test/kotlin/net/corda/cli/plugin/initialconfig/TestSqlFormattersForFieldAccess.kt
@@ -1,0 +1,202 @@
+package net.corda.cli.plugin.initialconfig
+
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import java.time.Instant
+import javax.persistence.Column
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.Table
+import javax.persistence.Version
+import org.junit.jupiter.api.Disabled
+
+class TestSqlFormattersForFieldAccess {
+
+    @Entity
+    @Table(name = "testtable", schema = "testschema")
+    private class TestEntity(
+        @Suppress("Unused")
+        @Column(name = "testyMcTestFace", nullable = false)
+        var testFace: String,
+
+        @Suppress("Unused")
+        @Column(name = "testNumber")
+        var testNum: Int,
+
+        @Suppress("Unused")
+        @Column(name = "testDate")
+        var testDate: Instant? = null,
+
+        @Suppress("Unused")
+        @JoinColumn(name = "crossRef")
+        var reference: ReferencedEntity? = null
+    )
+
+    @Entity
+    @Table(name = "testtable", schema = "testschema")
+    private class VersionedTestEntity(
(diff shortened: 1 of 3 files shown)


tools/plugins/initial-config/src/test/kotlin/net/corda/cli/plugin/initialconfig/TestSqlFormattersForFieldAccess.kt
```diff
@@ -11,34 +11,34 @@ import javax.persistence.JoinColumn
 import javax.persistence.Table
 import javax.persistence.Version
 
-class TestSqlFormatters {
+class TestSqlFormattersForFieldAccess {
```
simon-johnson-r3, line 15: The `TestSqlFormattersForPropertyAccess` test is good, because it tests the target state of JPA entity declarations.

Changing this test however means you're testing an interim state. One where `val` is changed to `var` but the `@get` annotations are still not added.

Shouldn't these remain `val` whilst the rest of the codebase is using `val` to test the current state of the code? It appears that if this was merged, right now we'd have no test for what our code is actually doing?

chrisr3: > Changing this test however means you're testing an interim state. One where val is changed to var but the @get annotations are still not added.

As I have already explained, JPA annotations should not be applied to a `val` property for "field access" since this will not work on Java 17, due to the property's underlying field being _truly_ `final`.

I suppose I could modify the code to throw an exception for such `val` properties, except that is likely to prove unpopular in the immediate term.

simon-johnson-r3: That is not what I was getting at. You have a separate ticket to replace the `val` everywhere with `var`. Until that ticket is complete we have `val` all over the place. If you submit this PR without a test for `val` how to you know they work in the current `val` form?

chrisr3: `var` = (`val` + "setter method")

Nothing is removed by doing this, so what are you expecting to break?

Or more explicitly:
- `val` -> `KProperty1`
- `var` -> `KMutableProperty1`

where `interface KMutableProperty1 : KProperty1`

chrisr3: > If you submit this PR without a test for `val` how to you know they work in the current `val` form?

In that case, would you care to approve #4367 please?

simon-johnson-r3: Yes, let's do it that way round.

simon-johnson-r3 (dismissed):

tools/plugins/initial-config/src/test/kotlin/net/corda/cli/plugin/initialconfig/TestSqlFormattersForFieldAccess.kt
simon-johnson-r3: I've approved them both, let's merge the other first.

kyriathar (approved):
OK (db)

PR #4369, opened 2023-07-26
CORE-15297: refactor crypto processor to reduce layering

dickon:
Combine the signing service with the crypto service, and remove more of the support for non-soft HSMs. That removes around 1200 lines of code, and address these tickets:

- CORE-15276: remove the unused crypto throttling layer, which was never triggered. The architectural choice of Corda 5 is to retry from the top, not at each layer, and crypto is never at the top. Mutliple levels of retires are problematic.
- CORE-15277: move the handling of exceptions out of a proxy layer into particular places where exceptions are expected. The main motivation was to reduce the number of places we have to touch on each interface change.

This required some movement of pieces to keep the package import graph from becoming cyclic with the above changes. In particular SigningKeyInfo and SigningKeySaveContext more into the persistence package, and TenantInfoService moves from crypto-service to crypto-softhsm-impl. We remove many test cases, which are testing layers that no longer exist, and move other test cases to reflect the newly simplified structure. All the tests in SigningServiceGeneralTests move to SoftCryptoServiceGeneralTests.

We also remove the `hsmId` and `instance` fields from `CryptoServiceRef`, since those are associated with layers that have been collapsed.


components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/SigningServiceGeneralTests.kt
```diff
@@ -1,507 +0,0 @@
...
-import org.mockito.kotlin.mock
-import org.mockito.kotlin.times
-import org.mockito.kotlin.verify
-import java.security.PublicKey
-import java.time.Instant
-import java.util.UUID
-import java.util.concurrent.TimeUnit
-
-// TODO - move content into SigningServiceImplTest
-class SigningServiceGeneralTests {
```
Sakpal, line 56: Major: Have we moved this tests somewhere else? Or these tests are not relevant anymore with the changes made?

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/HSMRepositoryImpl.kt
```diff
@@ -81,6 +79,8 @@ class HSMRepositoryImpl(
             val association =
                 findHSMAssociationEntity(em, tenantId)
                     ?: createAndPersistAssociation(em, tenantId, CryptoConsts.SOFT_HSM_ID, masterKeyPolicy)
+                findHSMAssociationEntity(em, tenantId)
+                    ?: createAndPersistAssociation(em, tenantId, CryptoConsts.SOFT_HSM_ID, masterKeyPolicy)
```
Sakpal: Major: Duplicate/Redundant line?

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/SigningServiceGeneralTests.kt
dickon: That's a good question, which I address in the description:

> We remove many test cases, which are testing layers that no longer exist, and move other test cases to reflect the newly simplified structure. All the tests in SigningServiceGeneralTests move to SoftCryptoServiceGeneralTests.

So all the tests in this file are now in SoftCryptoServiceGeneralTests.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/HSMRepositoryImpl.kt
dickon: Actually I already fixed this, it looks like it was a merge error. So if you reload the PR in github or repull the branch it should disappear.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt
```diff
@@ -81,12 +111,23 @@ class SoftCryptoService(
 
     override val extensions = listOf(CryptoServiceExtensions.REQUIRE_WRAPPING_KEY)
 
+    private fun <R> recoverable(description: String, block: () -> R) = try {
```
Sakpal, line 126: Curious: Can you help me understand the use of this method?

dickon: Recoverable runs a block of code, and catches and wraps certain exceptions in a way that lets the caller know what is required. This was used to be done  in the decorator layer before, and it requirement is part of the way Corda delivers robustness. We could just have a try/catch constrict with the exception wrapping logic in each place, but this way is easier to write and read, provided the reader is familiar with the function.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt
```diff
@@ -56,23 +81,28 @@ const val PRIVATE_KEY_ENCODING_VERSION: Int = 1
...
-    private val schemeMetadata: CipherSchemeMetadata,
+    private val signingRepositoryFactory: SigningRepositoryFactory,
+    override val schemeMetadata: CipherSchemeMetadata,
     private val defaultUnmanagedWrappingKeyName: String,
     private val unmanagedWrappingKeys: Map<String, WrappingKey>,
     private val digestService: PlatformDigestService,
     private val wrappingKeyCache: Cache<String, WrappingKey>?,
+    private val shortHashCache: Cache<ShortHash, SecureHash>?,
     private val privateKeyCache: Cache<PublicKey, PrivateKey>?,
+    private val signingKeyInfoCache: Cache<SecureHash, SigningKeyInfo>,
```
Sakpal: Major: Is this new support for cache, related to the caching of signing keys?

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt
```diff
@@ -218,42 +350,235 @@ class SoftCryptoService(
...
+            spec.wrappingKeyAlias,
+            tenantId
+        ).unwrap(
             spec.keyMaterial
         ).also {
             privateKeyCache?.put(publicKey, it)
         }
 
+
+    override fun querySigningKeys(
```
Sakpal, line 412: Curious: Is all of this functions added below, moved from some other files?

libs/crypto/crypto-config-impl/src/main/kotlin/net/corda/crypto/config/impl/CryptoConfigUtils.kt
```diff
@@ -170,6 +170,7 @@ fun SmartConfig.toConfigurationSecrets(): ConfigurationSecrets = ConfigurationSe
 fun SmartConfig.signingService(): CryptoSigningServiceConfig =
     CryptoSigningServiceConfig(this)
 
+// TODO remove me, dead?
```
Sakpal: Major: Can we mark this in JIRA instead?

libs/crypto/crypto-config-impl/src/main/kotlin/net/corda/crypto/config/impl/CryptoHSMConfig.kt
```diff
@@ -3,6 +3,7 @@ package net.corda.crypto.config.impl
 import com.typesafe.config.ConfigList
 import net.corda.libs.configuration.SmartConfig
 
+// TODO remove me, dead?
```
Sakpal: Major: Can we mark this in JIRA instead?

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SigningKeyInfo.kt
```diff
@@ -69,4 +74,18 @@ data class SigningKeyInfo(
         result = 31 * result + status.hashCode()
         return result
     }
+
+    fun convertToCryptoSigningKey(keyEncodingService: KeyEncodingService): CryptoSigningKey = CryptoSigningKey(
```
Sakpal: Major: This looks like a Utility function for converting one data object to another. Any particular reason why we are keeping this in one of the POJO's and not a separate utils file?

Primary concerns, this is adding some business logic in this Data class(POJO) and adds reference to other classes in it, thus creating a dependency.

libs/crypto/crypto-config-impl/src/main/kotlin/net/corda/crypto/config/impl/CryptoConfigUtils.kt
dickon: done

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt
dickon: yes, they all came from  `SigningServiceImpl`

libs/crypto/crypto-config-impl/src/main/kotlin/net/corda/crypto/config/impl/CryptoHSMConfig.kt
dickon: Created https://r3-cev.atlassian.net/browse/CORE-15824, TOOD deleted

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SigningKeyInfo.kt
dickon: On the other hand having it here improves locality of reference within the code, which I think is more important than the other considerations you mention.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/TenantInfoServiceImpl.kt
```diff
@@ -1,60 +1,51 @@
...
     override fun lookup(tenantId: String, category: String): HSMAssociationInfo? {
         logger.debug { "findAssignedHSM(tenant=$tenantId, category=$category)"  }
-        return openRepository().use {
+        return hsmRepositoryFactory().use {
             it.findTenantAssociation(tenantId, category)
         }
     }
 
-    private fun ensureWrappingKey(association: HSMAssociationInfo) {
+    private fun ensureWrappingKey(association: HSMAssociationInfo, cryptoService: CryptoService) {
```
simon-johnson-r3, line 48: You appear to have added a requirement to pass a member property to a member function here, is that a mistake?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
```diff
@@ -94,6 +94,8 @@ class CryptoFlowOpsBusProcessor(
                             ExceptionEnvelope("Expired", "Expired at $expireAt")
                         )
                     } else {
+                        // TODO CORE-15264 - remove the retry here, Corda 5 architecture is that the flow should do the retries.
```
simon-johnson-r3: I would phrase this a bit less like there's something missing, because that isn't true, we just over engineered a retry. How about just `CORE-15264 will remove the retry here...`

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/TenantInfoService.kt
```diff
@@ -0,0 +1,33 @@
...
+     * @param tenantId the tenant (virtual node or special purpose cluster level 
+     * @param category indicates what the keys will be used for, and should be one of the constants 
+     *                 defined in CryptoConsts.Categories
+     * @param cryptoService a reference to the crypto service, which is used to allocate a wrapping key
+     *                      if necessary.
+     * 
+     * @return Information about crypto processing for a specific category of keys in a specific tenant.
+     *         Includes `masterKeyAlias` which identifies the per-tenant master key used to wrap keys.
+     */
+    fun populate(tenantId: String, category: String, cryptoService: CryptoService): HSMAssociationInfo
```
simon-johnson-r3, line 29: typically we inject dependent services, is there a reason you pass one here?

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/CryptoService.kt
```diff
@@ -150,6 +162,142 @@ interface CryptoService {
...
+     * @return The public part of the pair.
+     */
+    @Suppress("LongParameterList")
+    fun generateKeyPair(
+        tenantId: String,
+        category: String,
+        alias: String? = null,
+        externalId: String? = null,
+        scheme: KeyScheme,
+        context: Map<String, String> = EMPTY_CONTEXT,
```
simon-johnson-r3, line 267: `context: Map<String, String> = emptyMap()` would be fine and more normal here (and other places) I think? Kotlin already provides an easy means to do this already. It saves the reader from looking up what EMPTY_CONTEXT is.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SigningKeyInfo.kt
simon-johnson-r3: I think this should be `toCryptoSigningKey` as in `toString` etc.

If you want to break the dependency in the data class, you could make it an extension function - and some members of the Kotlin community, including the author of Effective Kotlin would suggest that in cases like this where only public properties are used. I'm not sure you'd get a great deal of benefit from that in this case though.

simon-johnson-r3 (changes requested):

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoFlowOpsBusProcessor.kt
dickon: I removed the TODO, that seems to be the way we are going.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/TenantInfoService.kt
dickon: I clarified this in the docstring; the trouble is that there's a dependency cycle here.

dickon: and I could flatten TenantInfoSerivce into CryptoServiceImpl, and have not done so yet since CryptoServiceImpl is already large.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SigningKeyInfo.kt
dickon: Name changed.

I'd like to read up on the extension method point; my initial thinking is that in this case it'd be a surprising departure from standard OOP technique. But I don't think we should hold this work up on this point; I seem to recall examples going either way on this.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt
dickon: I clarified this is in a kdoc.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/SigningKeyInfo.kt
dickon: Having just bought and studied Item 45 and 46 in Effecitve Kotlin I think on balance the authors of that book would prefer my style rather an extension method here.  I should really read the whole book.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/TenantInfoServiceImpl.kt
dickon: I discussed this directly with Simon; he hadn't noticed I'd dropped cryptoService as as member property, due to the cycle.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/CryptoService.kt
dickon: This is code pasted from SigningService and is in 5.1 (https://github.com/corda/corda-runtime-os/blob/13c6ffded20a55ba01d46a19ee37bfee680805ab/components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/SigningService.kt#L19 )  I take the point about style but due to the size and time elapsed on this PR I think I should minimise the number of changes. And this point doesn't feel it rates a JIRA ticket, nor would it be the first style issue I'd want to address.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/TenantInfoServiceImpl.kt
dickon: Furthermore, we discussed other ways of getting rid of the cycle, but what we have works and is pretty low impact on the code so I'd like to move on.

simon-johnson-r3: I'd suggest here a complete decoupling of the classes, by introducing either an interface or higher order function, e.g.
```
populate(tenantId, category) { wrappingKeyAlias, context ->
  cryptoService.createWrappingKey(false, wrappingKeyAlias, context)
}
```

simon-johnson-r3 (approved):
I'm approving this with some minor comments still outstanding. They don't affect the logic of the code. Over time it would be good to fix these.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/TenantInfoService.kt
dickon: https://r3-cev.atlassian.net/browse/CORE-15853 to revisit this, a story on the tech debt epic.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/CryptoService.kt
dickon: https://r3-cev.atlassian.net/browse/CORE-15854 to track this, story in tech debt epic.

PR #4371, opened 2023-07-26
CORE-15642 - Persist token attributes in the output table

filipesoliveira:
Before this change the columns `tokenType`, `tokenIssuerHash`, `tokenSymbol`, `tokenTag`, `tokenOwnerHash`, and `tokenAmount` were not being populated.

Output table after updating the code:
image

### DIFF
--- a/components/ledger/ledger-persistence/src/integrationTest/kotlin/net/corda/ledger/persistence/utxo/tests/UtxoPersistenceServiceImplTest.kt
+++ b/components/ledger/ledger-persistence/src/integrationTest/kotlin/net/corda/ledger/persistence/utxo/tests/UtxoPersistenceServiceImplTest.kt
@@ -82,8 +82,12 @@ import java.security.spec.ECGenParameterSpec
 import java.time.Duration
 import java.time.Instant
 import java.time.temporal.ChronoUnit
+import java.util.Random
 import java.util.concurrent.atomic.AtomicInteger
 import javax.persistence.EntityManagerFactory
+import net.corda.v5.ledger.utxo.observer.UtxoToken
+import net.corda.v5.ledger.utxo.observer.UtxoTokenFilterFields
+import net.corda.v5.ledger.utxo.observer.UtxoTokenPoolKey
 
 @ExtendWith(ServiceExtension::class, BundleContextExtension::class)
 @TestInstance(PER_CLASS)
@@ -330,7 +334,25 @@ class UtxoPersistenceServiceImplTest {
             transactionStatus,
             visibleStatesIndexes
         )
-        persistenceService.persistTransaction(transactionReader)
+
+        // Create the utxo tokens
+        val tokenType = "token type"
+        val tokenSymbol = "token symbol"
+        val tokenTag = "token tag"
+        val issuerHash = newRandomSecureHash()
+        val ownerHash = newRandomSecureHash()
+        val tokenAmount = BigDecimal(1)
+        val utxoTokenMap = createUtxoTokenMap(
+            transactionReader,
+            tokenType,
+            tokenSymbol,
+            tokenTag,
+            issuerHash,
+            ownerHash,
+            tokenAmount
+        )
+
+        persistenceService.persistTransaction(transactionReader, utxoTokenMap)
 
         val entityFactory = UtxoEntityFactory(entityManagerFactory)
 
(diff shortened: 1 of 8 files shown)


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
```diff
@@ -154,13 +159,21 @@ class UtxoPersistenceServiceImpl(
...
             repository.persistTransactionOutput(
                 em,
                 transactionIdString,
                 UtxoComponentGroup.OUTPUTS.ordinal,
                 stateIndex,
                 stateAndRef.state.contractState::class.java.canonicalName,
-                timestamp = nowUtc
+                utxoToken?.poolKey?.tokenType,
+                utxoToken?.poolKey?.issuerHash?.toString(),
+                stateAndRef.state.notaryName.toString(),
```
owenstanford: The notary name is not used, we should look at removing this field.

components/ledger/ledger-persistence/src/integrationTest/kotlin/net/corda/ledger/persistence/utxo/tests/UtxoPersistenceServiceImplTest.kt
```diff
@@ -405,7 +405,7 @@ class UtxoPersistenceServiceImplTest {
                     assertThat(dbInput.field<String>("type")).isEqualTo(transactionOutput::class.java.canonicalName)
                     assertThat(dbInput.field<String>("tokenType")).isNull()
                     assertThat(dbInput.field<String>("tokenIssuerHash")).isNull()
-                    assertThat(dbInput.field<String>("tokenNotaryX500Name")).isNull()
+                    assertThat(dbInput.field<String>("tokenNotaryX500Name")).isEqualTo(notaryX500Name.toString())
```
owenstanford: same as the comment below I think we need to remove the notary field. 
We should set the token fields passed to the test so we can assert they are set in the DB

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
filipesoliveira: I have raised a ticket so we can look into updating the database table: https://r3-cev.atlassian.net/browse/CORE-15800.

components/ledger/ledger-persistence/src/integrationTest/kotlin/net/corda/ledger/persistence/utxo/tests/UtxoPersistenceServiceImplTest.kt
```diff
@@ -330,7 +334,25 @@ class UtxoPersistenceServiceImplTest {
...
+        val tokenAmount = BigDecimal(1)
+        val utxoTokenMap = createUtxoTokenMap(
+            transactionReader,
+            tokenType,
+            tokenSymbol,
+            tokenTag,
+            issuerHash,
+            ownerHash,
+            tokenAmount
+        )
```
lankydan, line 353: better to use named parameters here which should reduce the lines of code a lot?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistTransactionRequestHandler.kt
```diff
@@ -34,18 +34,21 @@ class UtxoPersistTransactionRequestHandler @Suppress("LongParameterList") constr
...
         val isTransactionVerified = transaction.status == TransactionStatus.VERIFIED
 
+        val listOfPairsStateAndUtxoToken = transaction.getVisibleStates().values.toList().toTokens(tokenObservers)
         val outputTokenRecords = if (isTransactionVerified) {
             utxoOutputRecordFactory.getTokenCacheChangeEventRecords(
                 holdingIdentity,
-                transaction.getVisibleStates().values.toList().toTokens(tokenObservers),
+                listOfPairsStateAndUtxoToken,
                 transaction.getConsumedStates(persistenceService).toTokens(tokenObservers)
             )
```
lankydan, line 43: Maybe this should change or be renamed slightly, since we're using it generate records and data to store the database. I'd argue the records should be based on the data we're going to store rather than what we have now.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistTransactionRequestHandler.kt
```diff
@@ -34,18 +34,21 @@ class UtxoPersistTransactionRequestHandler @Suppress("LongParameterList") constr
...
         } else {
             listOf()
         }
 
+        val unconsumedTokensMap = listOfPairsStateAndUtxoToken.associate { it.first.ref to it.second }
+
         // persist the transaction
-        persistenceService.persistTransaction(transaction)
+        persistenceService.persistTransaction(transaction, unconsumedTokensMap)
```
lankydan: Should this map be called `tokens` instead? The main thing is, can we make it consistent with the parameter for `persistTransaction`? (e.g. `utxoTokenMap` or `unconsumedTokensMap`, we should stick to one or choose another name).

vlajos (dismissed):

components/ledger/ledger-persistence/src/integrationTest/kotlin/net/corda/ledger/persistence/utxo/tests/UtxoPersistenceServiceImplTest.kt
filipesoliveira: The variables are also used in the `assert`calls. This avoids duplicating the values and lets people know where the values must match.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistTransactionRequestHandler.kt
filipesoliveira: I understand your point of view. But for this ticket, let's keep the changes to the bare minimum. There is a ticket with higher priority that needs to be done and requires the database table to be populated correctly.

lankydan (approved):

vlajos (approved):

PR #4372, opened 2023-07-26
CORE-15818: upgrade runtime repo to Java 17

jennyang-r3:
## Overview
Update runtime repo to use JAVA 17 JDK.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
-import static org.gradle.api.JavaVersion.VERSION_11
+import static org.gradle.api.JavaVersion.VERSION_17
 import static org.gradle.jvm.toolchain.JavaLanguageVersion.of
-import static org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_11
+import static org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_17
 import static org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_1_8
 import org.jetbrains.dokka.gradle.DokkaTask
 
@@ -36,7 +36,8 @@ tasks.named("wrapper") {
     distributionType = Wrapper.DistributionType.BIN
 }
 
-def javaVersion = VERSION_11
+def javaVersion = VERSION_17
+def javaVersionNumber = javaVersion.majorVersion.toInteger()
 
 snyk {
     api = project.findProperty("snykApiToken") ?: System.getenv('SNYK_TOKEN')
@@ -51,13 +52,14 @@ allprojects {
     pluginManager.withPlugin('java') {
         java {
             toolchain {
-                languageVersion = of(javaVersion.majorVersion.toInteger())
+                languageVersion = of(javaVersionNumber)
             }
 
             withSourcesJar()
         }
 
         tasks.withType(JavaCompile).configureEach {
+            options.release = javaVersionNumber
             options.encoding = 'UTF-8'
             options.compilerArgs += [ '-XDenableSunApiLintControl', '-parameters' ]
         }
@@ -93,7 +95,7 @@ allprojects {
             allWarningsAsErrors = true
             languageVersion = KOTLIN_1_8
             apiVersion = KOTLIN_1_8
-            jvmTarget = JVM_11
+            jvmTarget = JVM_17
             javaParameters = true   // Useful for reflection.
             freeCompilerArgs.addAll([
                 "-Xjvm-default=all"
@@ -110,7 +112,7 @@ allprojects {
     pluginManager.withPlugin('biz.aQute.bnd.builder') {
         tasks.withType(aQute.bnd.gradle.TestOSGi).configureEach {
             javaLauncher = javaToolchains.launcherFor {
-                languageVersion = of(javaVersion.getMajorVersion().toInteger())
+                languageVersion = of(javaVersionNumber)
             }
         }
     }
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FlowEngineImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FlowEngineImpl.kt
@@ -1,3 +1,4 @@
+@file:Suppress("deprecation")
 package net.corda.flow.application.services.impl
 
 import net.corda.data.flow.output.FlowStates
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FlowEngineImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FlowEngineImpl.kt
@@ -1,4 +1,3 @@
-@file:Suppress("deprecation")
 package net.corda.flow.application.services.impl
 
 import net.corda.data.flow.output.FlowStates
@@ -22,7 +21,6 @@ import org.osgi.service.component.annotations.Activate
 import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
 import org.osgi.service.component.annotations.ServiceScope.PROTOTYPE
-import java.security.AccessController
 import java.security.PrivilegedActionException
 import java.security.PrivilegedExceptionAction
 import java.util.UUID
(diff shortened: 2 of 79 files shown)


build.gradle
```diff
@@ -51,13 +52,14 @@ allprojects {
...
             toolchain {
-                languageVersion = of(javaVersion.majorVersion.toInteger())
+                languageVersion = of(javaVersionNumber)
             }
 
             withSourcesJar()
         }
 
         tasks.withType(JavaCompile).configureEach {
+            options.release = javaVersionNumber
```
chrisr3, line 62: Using JDK17 to compile Corda is one thing, generating Java 17 bytecode is another. FYI we _could_ continue to generate Java 11 bytecode by setting this to 11.

build.gradle
```diff
@@ -93,7 +95,7 @@ allprojects {
             allWarningsAsErrors = true
             languageVersion = KOTLIN_1_8
             apiVersion = KOTLIN_1_8
-            jvmTarget = JVM_11
+            jvmTarget = JVM_17
```
chrisr3, line 98: The Kotlin 1.8 compiler can _still_ only generate Java **8** bytecode instructions. All you're really doing here is just setting the compiled classes' version number to "61.0" (i.e. "Java SE 17"), thus preventing any JVM < 17 from loading them.

Sounds like "false advertising" to me... :shrug:.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FlowEngineImpl.kt
```diff
@@ -1,3 +1,4 @@
+@file:Suppress("deprecation")
```
chrisr3: This is a rather powerful warning to ignore. Would `@Suppress("removal")` suffice instead? Either here, or in any of the other places? (I'm assuming this is for SecurityManager-related stuff.)

chrisr3 (commented):
Once upon a time, I knew someone who wrote C code, but with "//" comments instead of "/* */". Because his code could therefore only be compiled by a C++ compiler, he _claimed_ that meant his code was actually C++!

Personally speaking, I was _never_ convinced. But that was almost 30 years ago. Strange that I should remember him now... :wink:.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/services/impl/FlowEngineImpl.kt
jennyang-r3: I tried using `@Surpress("removal")` here and/or other places, but it wouldn't compile it. Is it okay to leave it as it is or is there any other way around?

jennyang-r3: Did this:

```kotlin
@Suspendable
    override fun <R> subFlow(subFlow: SubFlow<R>): R {

        try {
            @Suppress("deprecation", "removal")
            java.security.AccessController.doPrivileged(PrivilegedExceptionAction {
                getFiberExecutionContext().sandboxGroupContext.dependencyInjector.injectServices(subFlow)
            })
        } catch (e: PrivilegedActionException) {
            throw e.exception
        }
```

I've removed the import and use the fully qualified name in the method, so we do not need to suppress at the file level. Both "deprecation" and "removal" are need to get intellij to not show any read lines. Just using "deprecation" is enough to get the code to compile though, but intellij shows it as an error in the editor.

build.gradle
jennyang-r3: The task says to update it to 17 - https://r3-cev.atlassian.net/browse/CORE-15477.

> Repos to be updated to JAVA 17 JDK, compiling to JAVA 17 bytecode:
> 
> [GitHub - corda/corda-runtime-os](https://github.com/corda/corda-runtime-os/)

@driessamyn should we not be doing this?

driessamyn: I discussed this with Rick and we were trying to find pros/cons and couldn't think of much of an up or downside, but we thought that we may as well change it as we're changing the JDK anyway.
For corda-api, we decided it may be better to keep it to J11 bytecode.

components/ledger/ledger-persistence/src/test/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImplTest.kt
```diff
@@ -294,10 +297,37 @@ class UtxoPersistenceServiceImplTest {
         }
     }
 
-    private inline fun <reified T : ContractState> createStateAndRef(returnState: T): StateAndRef<T> {
-        val txState = mock<TransactionState<T>> {
-            on { contractState } doReturn returnState
+    public class MockTransactionState<T : ContractState>(
```
driessamyn: It might be worth adding a comment to explain why this isn't using a mock.

BTW, I think the answer on this SO question explains why we hit this: 
https://stackoverflow.com/questions/75528685/getcanonicalname-returns-null-when-running-in-a-different-java-version

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/UtxoLedgerServiceImpl.kt
```diff
@@ -1,3 +1,4 @@
+@file:Suppress("deprecation")
```
vlajos: What deprecation do these suppress?

josephzunigadaly (dismissed):
REST owned files look ok

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/UtxoLedgerServiceImpl.kt
jennyang-r3: deprecation of `java.security.AccessController`. Shall I change it to specific lines of code?

dimosr (dismissed):
Approving for @corda/corda-platform-network-team "owned" files

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/UtxoLedgerServiceImpl.kt
jennyang-r3: something like this:
> @Suspendable
>    override fun <R> subFlow(subFlow: SubFlow<R>): R {
>
>        try {
>            @Suppress("deprecation", "removal")
>            java.security.AccessController.doPrivileged(PrivilegedExceptionAction {
>                getFiberExecutionContext().sandboxGroupContext.dependencyInjector.injectServices(subFlow)
>            })
>        } catch (e: PrivilegedActionException) {
>            throw e.exception
>        }

jujoramos: > deprecation of java.security.AccessController. Shall I change it to specific lines of code?

Id' say so, yes, otherwise we might end up introducing other deprecations without noticing it (same applies for other files in the `PR` where we are suppressing all deprecations within the entire file).

driessamyn: I think it may be better to add to the specific lines of code as otherwise we may miss other deprecation warnings.

simon-johnson-r3 (commented):
Assuming the deprecation suppression is made more specific as per other comments, this looks fine to me.

jennyang-r3:
What has changed since the approvals is only deprecation annotations from file level -> specific line

jujoramos (approved):
LGTM

vlajos (approved):
ledger: OK

ronanbrowne (approved):
LGTM - blt

relyafi (approved):
Uniqueness changes look ok

yarivus (approved):
LGTM

josephzunigadaly (approved):
REST owned files look ok

driessamyn (approved):
LGTM

dimosr (approved):

PR #4373, opened 2023-07-26
CORE-15770: Update remaining JPA entities to use var properties with property access

chrisr3:
Java 17 forbids Java reflection from updating non-static `final` fields, which means we can no longer declare JPA-annotated properties as `val`. Change them to `var` so that the underlying field cannot be `final`.

~Also move annotations to the "getter" methods so that Hibernate no longer needs to use reflection to update them either.~
Hibernate's "property access" is not very compatible with Kotlin's "null safety", since Hibernate seems to want to set properties to `null` occasionally, but Kotlin "setter" methods prevent it unconditionally. So I'll revert back to using "field access" here for now.

However, I'll leave the _capability_ for our entities to use Hibernate "property access" in place, since theoretically it's still a valid option.

### DIFF
--- a/components/crypto/crypto-persistence-model/src/main/kotlin/net/corda/crypto/persistence/db/model/HSMAssociationEntity.kt
+++ b/components/crypto/crypto-persistence-model/src/main/kotlin/net/corda/crypto/persistence/db/model/HSMAssociationEntity.kt
@@ -19,7 +19,7 @@ class HSMAssociationEntity(
      */
     @Id
     @Column(name = "id", nullable = false, updatable = false, length = 36)
-    val id: String,
+    var id: String,
 
     /**
      * Tenant which the configuration belongs to.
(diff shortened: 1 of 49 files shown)


vkolomeyko (approved):
LGTM for @corda/rest

simon-johnson-r3:
Where you've set `nullable = false` have you 1) checked that matches the schema in liquibase 2) checked that we have nullability checks turned on in hibernate - otherwise this has no effect as it's only used for schema generation via hibernate according to the docs.

chrisr3:
> Where you've set `nullable = false` have you 1) checked that matches the schema in liquibase 2) checked that we have nullability checks turned on in hibernate - otherwise this has no effect as it's only used for schema generation via hibernate according to the docs.

I checked that it was consistent with the schema, yes. But if it only serves documentation purposes or consistency with other annotations then that's good enough for me.

kyriathar (approved):
OK (packaging & crypto)

lankydan (approved):

thiagoviana (approved):
LGTM - for @corda/flow-worker

nkovacsx (approved):

yift-r3 (approved):
LGTM

PR #4375, opened 2023-07-27
CORE-15403: fix Snyk file formatting so waivers are correctly applied

ronanbrowne:
Fix indentation on `.Snyk` file, which is sensitive to correct yaml indentation, waivers are not applied unless formatting is properly followed.


yarivus (commented):
thx @ronanbrowne

yarivus (approved):
LGTM

PR #4376, opened 2023-07-27
CORE-4815 Added check if all sessions in subflow are already closed or errored before skipping suspension

ciaran238:
Check if all the sessions in a subflow are already closed or errored and then skipping suspending. This saves the flow the time of suspending only to do nothing and scheduling a wakeup.


ben-millar (approved):
LGTM

PR #4377, opened 2023-07-27
CORE-15815: Reduce the number of open connections

yift-r3:
Add a cache of key stores in the `TrustStoresMap` so that two identical sets of certificates will have the same key store, and the same key in the connection manager, which will reduce the number of connections in the connection manager.

### DIFF
--- a/components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/TrustStoresMap.kt
+++ b/components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/TrustStoresMap.kt
@@ -1,5 +1,8 @@
 package net.corda.p2p.gateway.messaging.http
 
+import com.github.benmanes.caffeine.cache.Cache
+import com.github.benmanes.caffeine.cache.Caffeine
+import net.corda.cache.caffeine.CacheFactoryImpl
 import net.corda.crypto.utils.convertToKeyStore
 import net.corda.data.p2p.GatewayTruststore
 import net.corda.libs.configuration.SmartConfig
@@ -17,6 +20,7 @@ import net.corda.v5.base.types.MemberX500Name
 import org.slf4j.LoggerFactory
 import java.security.KeyStore
 import java.security.cert.CertificateFactory
+import java.time.Duration
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.ConcurrentHashMap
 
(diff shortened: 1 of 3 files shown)


components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/TrustStoresMap.kt
```diff
@@ -44,6 +50,18 @@ internal class TrustStoresMap(
...
     private val entriesPerKey = ConcurrentHashMap<String, TruststoreKey>()
     private val trustRootsPerHoldingIdentity = ConcurrentHashMap<TruststoreKey, TrustedCertificates>()
+    private val keyStorePool: Cache<Collection<String>, KeyStore> = CacheFactoryImpl().build(
+        "P2P-Key-Store-Pool",
+        Caffeine.newBuilder()
+            .maximumSize(KEY_STORE_POOL_SIZE)
+            .expireAfterAccess(KEEP_KEY_STORE)
+    )
+
+    private fun TrustedCertificates.cachedKeyStore(): KeyStore? {
```
dimosr: can't this return `KeyStore` instead?

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/TrustStoresMap.kt
```diff
@@ -55,11 +73,11 @@ internal class TrustStoresMap(
 
     fun getTrustStore(sourceX500Name: MemberX500Name, destinationGroupId: String) =
         trustRootsPerHoldingIdentity[TruststoreKey(sourceX500Name, destinationGroupId)]
-            ?.trustStore
+            ?.cachedKeyStore()
             ?: throw IllegalArgumentException("Unknown trust store for source X500 name ($sourceX500Name) " +
                     "and group ID ($destinationGroupId)")
 
-    fun getTrustStores() = trustRootsPerHoldingIdentity.values.mapNotNull { it.trustStore }
+    fun getTrustStores() = trustRootsPerHoldingIdentity.values.mapNotNull { it.cachedKeyStore() }
```
dimosr: if you convert `cachedKeyStore` to return a non-nullable type, this could also be simplified to:
```suggestion
    fun getTrustStores() = trustRootsPerHoldingIdentity.values.map { it.cachedKeyStore() }
```

components/gateway/src/test/kotlin/net/corda/p2p/gateway/messaging/http/TrustStoresMapTest.kt
```diff
@@ -224,4 +225,42 @@ class TrustStoresMapTest {
...
+        )
+
+        testObject.getTrustStore(MemberX500Name.parse(source1X500Name), groupId)
+        testObject.getTrustStore(MemberX500Name.parse(source2X500Name), groupId)
+
+        verify(keyStore, times(1)).setCertificateEntry(any(), any())
+    }
+
+    @Test
+    fun `trust store creat the key sore more than once when the certificates are different`() {
```
dimosr:
```suggestion
    fun `trust store creat the key store more than once when the certificates are different`() {
```

dimosr (approved):

dimosr:
@yift-r3 thinking about this a bit more, a better way of resolving this might actually be to adjust `DestinationInfo`, so that it takes the actual certificates (e.g. as PEM strings). In this way, keying for caching will have proper equality checks and will behave as expected. And then these are converted to keystores only at the point of use, e.g. when establishing the connections. 

It also seems the same issue can happen with `clientCertificatesKeyStore` in the case of mutual TLS. So, we need to change this too to fix the issue properly.

PR #4379, opened 2023-07-28
CORE-15815: Reduce the number of open connections in mutual TLS

yift-r3:
This pull request changes the destination info key to a valid hash map key so the number of connections will stay minimal.

### DIFF
--- a/components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/DynamicX509ExtendedTrustManager.kt
+++ b/components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/DynamicX509ExtendedTrustManager.kt
@@ -131,6 +131,6 @@ internal class DynamicX509ExtendedTrustManager(
     private fun getAllX509TrustManagers(): List<X509ExtendedTrustManager> =
         trustStoresMap
             .getTrustStores()
-            .mapNotNull { it.trustStore }
+            .map { it.trustStore }
             .flatMap { it.x509ExtendedTrustManager() }
 }
--- a/components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/DynamicX509ExtendedTrustManager.kt
+++ b/components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/DynamicX509ExtendedTrustManager.kt
@@ -129,5 +129,8 @@ internal class DynamicX509ExtendedTrustManager(
     }
 
     private fun getAllX509TrustManagers(): List<X509ExtendedTrustManager> =
-        trustStoresMap.getTrustStores().flatMap { it.x509ExtendedTrustManager() }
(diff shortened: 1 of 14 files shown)


components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/http/DynamicX509ExtendedTrustManager.kt
```diff
@@ -129,5 +129,8 @@ internal class DynamicX509ExtendedTrustManager(
     }
 
     private fun getAllX509TrustManagers(): List<X509ExtendedTrustManager> =
-        trustStoresMap.getTrustStores().flatMap { it.x509ExtendedTrustManager() }
+        trustStoresMap
+            .getTrustStores()
+            .mapNotNull { it.trustStore }
```
dimosr:
```suggestion
            .map { it.trustStore }
```

components/gateway/src/test/kotlin/net/corda/p2p/gateway/messaging/DynamicKeyStoreTest.kt
```diff
@@ -437,5 +438,69 @@ class DynamicKeyStoreTest {
...
+            )
+            val storeTwo = dynamicKeyStore.ClientKeyStore(
+                listOf(certificateOne, certificateTwo, certificateThree), "id2"
+            )
+
+            assertThat(storeOne).isNotEqualTo(storeTwo)
+        }
+
+        @Test
+        fun `equals return true when ids and certifiactes are the same`() {
```
dimosr:
```suggestion
        fun `equals return true when ids and certificates are the same`() {
```

components/gateway/src/test/kotlin/net/corda/p2p/gateway/messaging/DynamicKeyStoreTest.kt
```diff
@@ -437,5 +438,69 @@ class DynamicKeyStoreTest {
...
+            val certificateOne = mock<Certificate>()
+            val certificateTwo = mock<Certificate>()
+            val certificateThree = mock<Certificate>()
```
dimosr, line 495: The equality of the dynamic key store depends on the equality of the underlying certificates, so these unit tests do not exercise the scenario where there are 2 different `Certificate` instances with the same content. Looking at the docs, they seem to imply that it will return true. But, did you sanity check that any other way (e.g. by creating some real certificates and comparing them)?

(and same applies for hashcode)

yift-r3: Yes, I did. I tested it with a few certificates. The equals and hashCode are consistent.

dimosr (approved):

dickon (approved):
LGTM for crypto changes

PR #4380, opened 2023-07-28
CORE-15747: Re-register member when upgrading vNode

BenYip123:
- Add OSGi services MembershipGroupReaderProvider,MemberResourceClient, MembershipQueryClient, so member information is accessible in VirtualNodeUpgradeOperationHandler.kt
- Trigger registration after vNode upgrade
- Update corda.serial if it exists
- Update VirtualNodeUpgradeOperationHandlerTest to reflect changes

Topic permission added for DB worker in `corda-api`: [CORE-15747: Add db to topic permission #1189](https://github.com/corda/corda-api/pull/1189)
Enable test in `corda-e2e-tests`: [CORE-15747: Enable virtual node upgrade updates CPI info in the MGM test #177](https://github.com/corda/corda-e2e-tests/pull/177)


gradle.properties
```diff
@@ -45,7 +45,7 @@ bouncycastleVersion=1.73
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.0.0.xx-SNAPSHOT to pick up maven local published copy
 #cordaApiVersion=5.1.0.xxx-SNAPSHOT
-cordaApiVersion=5.1.0.10-beta+
+cordaApiVersion=5.1.0.10-alpha-1690550393144
```
BenYip123: Reminder to revert this

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -139,6 +165,27 @@ internal class VirtualNodeUpgradeOperationHandler(
...
+
+            var hasSubmitted = false
+            while (!hasSubmitted) {
+                val registrationResult =
+                    memberResourceClient
+                        .startRegistration(upgradedVNodeInfo.holdingIdentity.shortHash, registrationContext)
+                if (registrationResult.registrationStatus == SubmittedRegistrationStatus.SUBMITTED) {
+                    hasSubmitted = true
+                }
+            }
```
josephzunigadaly: What are the conditions that lead to a retry?

BenYip123: When the registrationStatus is `SubmittedRegistrationStatus.NOT_SUBMITTED` it will keep retrying

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -139,6 +147,27 @@ internal class VirtualNodeUpgradeOperationHandler(
...
         )
+
+        // Re-register the member once the virtual node has been upgraded, so that the member CPI version is up-to-date
+        if(membershipQueryClient.isRunning) {
+            val x500Name = membershipGroupReaderProvider.getGroupReader(upgradedVNodeInfo.holdingIdentity).owningMember
+            val registrationRequest = membershipQueryClient
+                .queryRegistrationRequests(upgradedVNodeInfo.holdingIdentity, x500Name, limit = 1)
+                .getOrThrow()
+                .first()
+            val registrationContext = registrationRequest.memberProvidedContext.toMap()
```
nikinagy: You will need to make sure this does not include the optional serial field from the previous context by using a filter. I would also advise to update it instead to your current serial by performing a member lookup.

Your re-registration attempt will be declined if you use and outdated serial: https://github.com/corda/corda-runtime-os/wiki/Member-Re%E2%80%90registration#optional-serial-number

nikinagy (commented):
One thing that I mentioned on the network slack channel as well: We have queueing in place in case there are more than 1 request being submitted, see details: https://github.com/corda/corda-runtime-os/wiki/Member-Re%E2%80%90registration#request-queue

If a user decided to submit a request manually at the same time when vnode upgrade triggers an automated re-registration the chances are being high your re-registration attempt will be declined.

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -110,6 +121,40 @@ internal class VirtualNodeUpgradeOperationHandler(
 
         publishVirtualNodeInfo(upgradedVNodeInfo)
 
+        // Re-register the member once the virtual node has been upgraded, so that the member CPI version is up-to-date
+        if (membershipQueryClient.isRunning &&
```
josephzunigadaly: When does `membershipQueryClient.isRunning` return false?

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -110,6 +121,40 @@ internal class VirtualNodeUpgradeOperationHandler(
...
+            membershipQueryClient.queryRegistrationRequests(upgradedVNodeInfo.holdingIdentity, limit = 1)
+                .getOrThrow().isNotEmpty()
```
josephzunigadaly: We should remove the `getOrThrow()` call. We can replace it with an `is MembershipQueryResult.Success` condition. It would need storing the `queryRegistrationRequests` in a variable and splitting this into a nested if statement.

We are happy to skip this when `isRunning` is false. It it not consistent to throw an exception at this point.

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -110,6 +121,40 @@ internal class VirtualNodeUpgradeOperationHandler(
...
+        ) {
+            val x500Name = membershipGroupReaderProvider.getGroupReader(upgradedVNodeInfo.holdingIdentity).owningMember
+            val registrationRequest = membershipQueryClient
+                .queryRegistrationRequests(upgradedVNodeInfo.holdingIdentity, x500Name, limit = 1)
+                .getOrThrow()
```
josephzunigadaly: Another call to `getOrThrow()`

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -110,6 +121,40 @@ internal class VirtualNodeUpgradeOperationHandler(
...
+                .first()
+
+            val registrationContext = registrationRequest.memberProvidedContext.toMap().toMutableMap()
+            if (registrationContext.containsKey("corda.serial")) {
+                registrationContext["corda.serial"] =
+                    membershipGroupReaderProvider.getGroupReader(upgradedVNodeInfo.holdingIdentity)
```
josephzunigadaly: `getGroupReader` is called with the same parameter on line 129. Can we store the result in a variable and reuse it?

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -110,6 +121,40 @@ internal class VirtualNodeUpgradeOperationHandler(
...
+                        .lookup(x500Name, MembershipStatusFilter.ACTIVE_OR_SUSPENDED)?.serial.toString()
```
josephzunigadaly: If `lookup` returns `null` the `toString()` call will return a string containing "null". So the value will be set to "null". Is that what we want here?

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -110,6 +121,40 @@ internal class VirtualNodeUpgradeOperationHandler(
...
+            }
+
+            var hasSubmitted = false
+            val end = LocalDateTime.now().plusSeconds(60)
+            while (!hasSubmitted) {
```
josephzunigadaly: I think we should remove this loop and make one attempt at submitting the request.

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
BenYip123: I think it is running when the workers start up, but tests were failing in `VirtualNodeUpgradeOperationHandlerTest` without that line

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -110,6 +121,40 @@ internal class VirtualNodeUpgradeOperationHandler(
...
+                if (LocalDateTime.now().isAfter(end)) {
+                    throw TimeoutException("Timed out re-registering member after vNode upgrade")
```
josephzunigadaly: We started this code block happy to skip if something was not running. Now we are throwing an exception. I think we should remove this exception.

What code paths/situations lead to this exception?

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -139,6 +184,9 @@ internal class VirtualNodeUpgradeOperationHandler(
                     "name: ${upgradedVNodeInfo.cpiIdentifier.name}, version: ${upgradedVNodeInfo.cpiIdentifier.version} " +
                     "(request $requestId)"
         )
+
+
+
```
josephzunigadaly: Did we move the new code higher up above the DB schema migrations?

Is it possible to pull the new code out of this method and run after the successful completion of this method? That would preserve the current behaviour of VNode upgrade and add the new behaviour on to the end.

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
josephzunigadaly: We can update the mock in the test to return an empty list and then remove the call to `isRunning`.

```kotlin
    private val membershipQueryClient = mock<MembershipQueryClient>().apply {
        whenever(queryRegistrationRequests(any(), anyOrNull(), any(), anyOrNull())).thenReturn(MembershipQueryResult.Success(emptyList()))
    }
```

We could also add `MembershipQueryClient` to the list of services we depend on in `net.corda.virtualnode.write.db.impl.VirtualNodeWriteEventHandler.onStartEvent`. That will tell the lifecycle code to wait for the client to become ready before starting our `VirtualNodeWriteService`.

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
BenYip123: The check on line 127 should prevent the lookup from being null and I believe the default serial is 0 on a member lookup

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
BenYip123: I was thinking if the registration was not submitted it'll keep trying until it times out due to the request queue described in the doc: https://github.com/corda/corda-runtime-os/wiki/Member-Re%E2%80%90registration#request-queue

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
josephzunigadaly: The check on line 127 is looking up registration requests. This is looking for something different. The filter alone makes it possible for `lookup` to return null.

The doc comment for lookup says:
> [filter] should be only used by the P2P and membership layers. Everywhere else we must use the default value.

Should we be using `ACTIVE_OR_SUSPENDED`?

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
BenYip123: I could try that.
For lines 111-121, it would need to run those lines again or put into a method just to get the `upgradedVNodeInfo`, unless there is another way to get the HoldingIdentity

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
BenYip123: Oh I see, you are correct. I will use the default value here.

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -190,6 +189,54 @@ internal class VirtualNodeUpgradeOperationHandler(
...
     }
 
+    // Re-register the member if the member already exists
+    // after the virtual node has been upgraded, so that the member CPI version is up-to-date
+    private fun reRegisterMember(upgradedVNodeInfo: VirtualNodeInfo) {
+        val holdingIdentity = upgradedVNodeInfo.holdingIdentity
+        if (
+            (membershipQueryClient.queryRegistrationRequests(holdingIdentity) as MembershipQueryResult.Success)
+                .payload.isNotEmpty()
+        ) {
```
josephzunigadaly: The `as` keyword will throw an exception if not of type `Success`.

You can avoid this if you take `membershipQueryClient.queryRegistrationRequests(holdingIdentity)` out of the condition, up before the if statement, store the result in a variable. Then inside the condition you can say `variable is Success && variable.payload.isNotEmpty()`. 

The language will exit the condition early and treat the condition as false if variable is not of type `Success`. No exceptions will be thrown.

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -190,6 +189,55 @@ internal class VirtualNodeUpgradeOperationHandler(
...
+            val x500Name = membershipGroupReader.owningMember
+            val registrationRequest = membershipQueryClient
+                .queryRegistrationRequests(holdingIdentity, x500Name, limit = 1)
+            val registrationContext = (registrationRequest as MembershipQueryResult.Success)
+                .payload
+                .first()
+                .memberProvidedContext
+                .toMap().toMutableMap()
+
+            if (registrationContext.containsKey("corda.serial")) {
```
nikinagy: I think you could use the const value instead of "corda.serial" from MemberInfoExtension.SERIAL

nikinagy (commented):
I am fine with this if we want to keep things at minimum and we don't mind that this does not guarantee a successful automated re-registration (since we don't have any retries).
Only added a minor comment.

seanbrereton (approved):
LGTM - BLT

nikinagy (approved):
LGTM

josephzunigadaly (approved):

PR #4384, opened 2023-07-30
CORE-15833: Fix parsing of (:parameter) and (**::int) for custom queries.

chrisr3:
Prevent parentheses from being included in either `Parameter` or `JsonCast` tokens, because `LeftParentheses` and `RightParentheses` are tokens in their own right and need to be parsed as such.


lankydan (approved):
I'll have to take your word on the regex unless I test it (hopefully my tests would catch any unwanted behaviour).

PR #4385, opened 2023-07-31
Ledger Hibernate POC

relyafi:
DO NOT MERGE - PR only to enable easy review of changes.


components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
```diff
@@ -133,7 +134,7 @@ class EntitySandboxServiceImpl @Activate constructor(
         // entity manager factory per sandbox.
 
         // TODO - add general vault entities
-        val entityClasses = EntityExtractor.getEntityClassNames(cpks).map {
+        val entityClasses = UtxoLedgerEntities.classes + EntityExtractor.getEntityClassNames(cpks).map {
```
chrisr3: This will leak non-heap memory when the `PERSISTENCE` sandbox is unloaded. Every time we create a new `PERSISTENCE` sandbox, Hibernate will create a new dynamic proxy class for each member of `UtxoLedgerEntities`, with these proxies belonging to the `UtxoLedgerEntities` classloader - i.e. bundle. The only way to release that memory is to unload the entire `UtxoLedgerEntities` bundle from the OSGi framework.

You are going to need to `getOrCreate()` a different EMF for the same data source as this EMF, and that _only_ contains platform entities. These platform EMFs will all need to be cached so that other sandboxes can share them, so it's possible that such an EMF will already have been created.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
```diff
@@ -421,41 +527,65 @@ class UtxoRepositoryImpl @Activate constructor(
...
+            val rowsUpdated = entityManager.createQuery("""
+                UPDATE UtxoTransactionStatusEntity
+                SET status = :status, updated = :updatedAt
+                WHERE transactionId = :transactionId AND status = 'U'"""
+            )
+                .setParameter("transactionId", transactionId)
+                .setParameter("status", transactionStatus.value)
+                .setParameter("updatedAt", timestamp)
+                .executeUpdate()
+                .logResult("transaction status [$transactionId, ${transactionStatus.value}]")
```
chrisr3, line 559: This code is not equivalent to the code it is trying to replace. The original code first tries to `INSERT`, and if that fails it `UPDATE`s what is already there, all within a single atomic operation. Whereas the new code does the following:

- It invokes `find()` to fetch a row with the required primary key.
- If such a row exists, then it `UPDATE`s it.
- If the row _does not_ exist, then it attempts to `INSERT` it. But if this `INSERT` should fail with a key violation then the new code does nothing, whereas the original code would still have `UPDATE`d it instead.

Note that the `INSERT` _can_ indeed fail with the new code because there is a brief interval between `find()` returning `null` and performing the `INSERT` operation in which another worker process could also `INSERT` a row with this same key. It is not possible to use row-level locks to close this gap because there is not yet any row to lock.

A more accurate replacement would need to do the following:
- Attempt to `INSERT` a new row.
- If the `INSERT` fails due to a key violation then fetch the existing row (ideally with `SELECT FOR UPDATE`) and then `UPDATE` it.

IIRC a `SELECT FOR UPDATE` corresponds to what Hibernate calls "pessimistic locking".

relyafi: @chrisr3 the `find()` is only done in the case where the specified status is `U`, i.e. the only valid operation in this scenario is an insert - we'd never update a record to `U`. I agree the insert can fail, which would trigger an exception which is caught, which is analagous to `ON CONFLICT DO NOTHING`.

The update case does not do any find, it's an atomic operation.

For reference, the postgres docs indicate that this update pattern is similar to `ON CONFLICT DO UPDATE` when the tx isolation level is read committed: https://www.postgresql.org/docs/current/transaction-iso.html#XACT-READ-COMMITTED

chrisr3: If `transactionStatus` is `UNVERIFIED` but there's an existing row in the database _also_ with `transactionStatus = UNVERIFIED` then doesn't:
```
// Insert/update status. Update ignored unless: UNVERIFIED -> * | VERIFIED -> VERIFIED | INVALID -> INVALID
```
imply that you should still update?

The `find()` is a meaningless operation in any case because another process can always insert a row in the brief interval between receiving the `null` and executing the `INSERT`. You might as well always `INSERT` and then check whether it has succeeded.

My point about performing the `find()` before the `UPDATE` would be to ensure that you are updating the _actual_ current version of this row, e.g. if some other query has updated different columns on this table than you then your new update would accidentally clobber those changes.

And I don't see how your comparison with `ON CONFLICT DO UPDATE` is relevant here either because that concerns a single atomic database operation, whereas your usage pattern is spread across at least two.

relyafi: > If transactionStatus is UNVERIFIED but there's an existing row in the database also with transactionStatus = UNVERIFIED then doesn't:

> // Insert/update status. Update ignored unless: UNVERIFIED -> * | VERIFIED -> VERIFIED | INVALID -> INVALID
> imply that you should still update?

It's impossible to say what the original intent was, given it's not been documented anywhere. It's true that the timestamp won't be updated under the new model, but I suspect that was simply an artifact of the original approach rather than by design. These are supposedly idempotent operations, so I'd argue changing the record in this scenario is wrong.

> The find() is a meaningless operation in any case because another process can always insert a row in the brief interval between receiving the null and executing the INSERT. You might as well always INSERT and then check whether it has succeeded.

The plan is to rework this if we proceed to do a one-time lookup of the transaction record, and if found then we can assume the insert has already taken place (across all tables) since all the record inserts initially happen in a single transaction. In my opinion its better not to go causing needless transaction aborts. But this is simply a POC and there is much refinement still to do.

> My point about performing the find() before the UPDATE would be to ensure that you are updating the actual current version of this row, e.g. if some other query has updated different columns on this table than you then your new update would accidentally clobber those changes.

I think this shows a misunderstanding of what actually goes on when update occurs - if you read the Postgres documentation it tells you that the UPDATE operation will take into account any underlying updates that occur in parallel, so the WHERE clause will be re-evaluated in this scenario. There is no need for any `find()` in this scenario.

> And I don't see how your comparison with ON CONFLICT DO UPDATE is relevant here either because that concerns a single atomic database operation, whereas your usage pattern is spread across at least two.

Again, I think this shows a fundamental misunderstanding of what the update code is actually doing - it's clear from the Postgres documentation that this is atomic.

chrisr3: Re: the `UPDATE` -  OK, you are executing an _explicit_ UPDATE query here. I was thinking that you were merging a DTO (for some reason) since you couldn't be altering a "live" proxy class. Sure, the `UPDATE` step will be atomic.

relyafi: Yep. FWIW, I agree that _if_ we need to do an "upsert", then the current `ON CONFLICT DO UPDATE` would be required, since there is no atomic functionality available to Hibernate to do this. My focus here has been on seeing whether we need an upsert, or can always use an explicit insert or update depending on the context of the request.

What may scupper us is that @lankydan has found a scenario where we immediately insert an invalid status record, which means it could require either an insert or update. If we can't mitigate this by using the context of the API call, then that may force us to use an upsert approach for this.

components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
relyafi: @chrisr3 see https://github.com/corda/corda-runtime-os/pull/4385/files#diff-76df693940f1e402abd2fcf16bbde5b6334410c0492d3f84410733b754c53732R58 - is this what you were imagining?

PR #4386, opened 2023-07-31
CORE-15806 Re-creating entity manager factories causing memory leak of non-heap

kyriathar:
## Problem

On every VNode reconciliation run we create a new EMF per VNode DB connection (i.e. 1 EMF per VNode DB connection for all VNodes). The problem is, on new EMF creation, Hibernate creates and loads new proxy classes for the entities, even for the same entities, and these proxies accumulate in the class loader (memory leak). 

## Fix

Caches EMFs across VNode reconciliations to like so re-use existing EMFs and stop creating new ones every time which means entity proxies are now being created once per VNode and are re-used across reconciliations.

##

After this "fix" it means we should still have "duplicate" entities proxies across VNodes EMFs but the "duplicate" entities proxies across reconciliation runs are now resolved.

The effect of this "fix" was tested to be fixing the above memory leak as shown [here](https://github.com/corda/corda-runtime-os/pull/4364#discussion_r1277361956).


conal-smith (approved):
lgtm

simon-johnson-r3 (approved):

PR #4387, opened 2023-07-31
CORE-15574 - Add replays for exceptions on REST calls for e2e-tests

dimosr:
Adding retries for exceptions, so that tests can recover from transient failures, such as `HttpHostConnectException` or `SSLHandshakeException`. I scanned all the usages of `assertWithRetry` and converted to `assertWithRetryIgnoringExceptions` with the exceptions of cases where the operation might have side-effects if applied under transient network conditions that could later lead to test failures (e.g. initiation of registrations, flows etc.).

I also did a run of the e2e tests job integrated with the changes in this PR, available [here](https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda5-e2e-tests/detail/PR-178/3/pipeline/).

### DIFF
--- a/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/utils/ClusterUtils.kt
+++ b/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/utils/ClusterUtils.kt
@@ -55,7 +55,7 @@ fun ClusterBuilder.eventuallyUploadCpi(
 }
 
 fun ClusterBuilder.eventuallyCreateVirtualNode(cpiFileChecksum: String, x500Name: String): String {
-    val vNodeJson = assertWithRetryIgnoringExceptions {
+    val vNodeJson = assertWithRetry {
         timeout(retryTimeout)
         interval(retryInterval)
         command { vNodeCreate(cpiFileChecksum, x500Name) }
--- a/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/utils/ClusterUtils.kt
+++ b/applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/utils/ClusterUtils.kt
@@ -2,6 +2,7 @@ package net.corda.applications.workers.smoketest.utils
 
 import net.corda.e2etest.utilities.ClusterBuilder
 import net.corda.e2etest.utilities.assertWithRetry
+import net.corda.e2etest.utilities.assertWithRetryIgnoringExceptions
 import net.corda.e2etest.utilities.awaitVirtualNodeOperationStatusCheck
 import net.corda.rest.ResponseCode
 import org.assertj.core.api.Assertions
@@ -26,7 +27,7 @@ fun ClusterBuilder.eventuallyUploadCpi(
     Assertions.assertThat(requestId).withFailMessage(ERROR_IS_CLUSTER_RUNNING).isNotEmpty
 
     // BUG:  returning "OK" feels 'weakly' typed
-    val json = assertWithRetry {
+    val json = assertWithRetryIgnoringExceptions {
         timeout(retryTimeout)
         interval(retryInterval)
         command { cpiStatus(requestId) }
@@ -54,7 +55,7 @@ fun ClusterBuilder.eventuallyUploadCpi(
 }
 
 fun ClusterBuilder.eventuallyCreateVirtualNode(cpiFileChecksum: String, x500Name: String): String {
-    val vNodeJson = assertWithRetry {
+    val vNodeJson = assertWithRetryIgnoringExceptions {
         timeout(retryTimeout)
         interval(retryInterval)
         command { vNodeCreate(cpiFileChecksum, x500Name) }
--- a/testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
+++ b/testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
@@ -21,7 +21,7 @@ fun ClusterInfo.conditionallyUploadCordaPackage(
 }
 
 fun ClusterInfo.conditionallyUploadCpiSigningCertificate() = cluster {
-    val hasCertificateChain = assertWithRetry {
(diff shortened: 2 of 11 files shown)


testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MGMUtils.kt
```diff
@@ -334,7 +334,7 @@ fun ClusterInfo.activateMember(
     x500Name: String,
     serialNumber: Int? = null,
 ) = cluster {
-    assertWithRetry {
+    assertWithRetryIgnoringExceptions {
```
charlieR3: In the case of member activation, if the original attempt was successful but an error was returned through the REST API for some reason, I think retries will fail either due to the serial number no longer being the current version or the member status not being in the correct state for activation

charlieR3: Same for suspension, and maybe registration approve/decline also.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MGMUtils.kt
```diff
@@ -103,7 +103,7 @@ private fun ClusterInfo.createApprovalRuleCommon(
         "ruleRegex" to regex
     )
 
-    assertWithRetry {
+    assertWithRetryIgnoringExceptions {
         interval(1.seconds)
         command { post(url, ObjectMapper().writeValueAsString(payload)) }
```
charlieR3: This will fail on retry if the rule was successfully created the first time but the REST API returned an error. It will be considered a duplicate rule.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
```diff
@@ -159,7 +159,7 @@ fun ClusterInfo.createKeyFor(
     category: String,
     scheme: String
 ): String = cluster {
-    val keyId = assertWithRetry {
+    val keyId = assertWithRetryIgnoringExceptions {
         interval(1.seconds)
         command { createKey(tenantId, alias, category, scheme) }
```
charlieR3: create key and add soft hsm could fail on retry with the same parameters I think

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/utils/ClusterUtils.kt
```diff
@@ -54,7 +55,7 @@ fun ClusterBuilder.eventuallyUploadCpi(
 }
 
 fun ClusterBuilder.eventuallyCreateVirtualNode(cpiFileChecksum: String, x500Name: String): String {
-    val vNodeJson = assertWithRetry {
+    val vNodeJson = assertWithRetryIgnoringExceptions {
         timeout(retryTimeout)
         interval(retryInterval)
         command { vNodeCreate(cpiFileChecksum, x500Name) }
```
charlieR3: Vnode create could fail on retry if previously succeeded.

charlieR3 (commented):
LGTM but highlight a couple areas that might cause issues

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MGMUtils.kt
dimosr: I saw that for suspend and activate operations, the utils were already handling 409s, which is what I thought we returned in cases of stale serial numbers or a member being already in that status.

But, I'll admit it's hard to be 100% certain just by looking at the code as it spans multiple components and has a lot of logic (especially the registration handlers). So, I think it makes sense to err on the side of caution now and avoid retries here for now.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MGMUtils.kt
dimosr: ah ok, I thought we only performed uniqueness checks on the UUID, so it would just generate a duplicate rule with a different UUID and the same regex. But, I can see that we are explicitly checking for rules with the same regex and failing the operation. I'll revert.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
dimosr: Same here, I thought it would generate a new key pair. But, seems like the crypto service performs uniqueness checks on the alias indeed. I will revert.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/utils/ClusterUtils.kt
dimosr: that's probably an omission on my side, I thought I had removed it but maybe it slipped. I'll revert.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
```diff
@@ -32,7 +32,7 @@ fun ClusterInfo.conditionallyUploadCpiSigningCertificate() = cluster {
         it.code != ResponseCode.RESOURCE_NOT_FOUND.statusCode
     }
     if (!hasCertificateChain) {
-        assertWithRetryIgnoringExceptions {
+        assertWithRetry {
```
charlieR3: I think this one might have been OK. I originally had comments on the certificate upload, but I had a look at the code and it seems to merge the entity so I don't think reuploading the same cert will have a negative effect.

dimosr: ok I'll keep it then

charlieR3 (commented):
LGTM (although there is a build failure)

charlieR3 (dismissed):

charlieR3 (approved):

PR #4394, opened 2023-08-01
CORE-14685: Fix driver bundle resolution as a standalone utility.

chrisr3:
- Ensure that `corda-driver` depends upon the bundles it will need as a standalone utility for use outside of `corda-runtime-os`.
- Reduce the number of system API bundles to just `corda-base`, `corda-crypto*` and `corda-avro-schema`.
- Add missing Ledger and Notary API bundles to the public sandbox.
- Add missing ledger key metadata to the mock network.
- Ensure `corda-driver` cannot load broken `de.javakaffe.kryo-serializers` bundle.


yarivus (commented):
thx @chrisr3

PR #4400, opened 2023-08-02
CORE-14685: Add some Kotlin-friendly extensions to DriverDSL.

chrisr3:
Add some `inline` Kotlin extension functions so that the  Java `DriverDSL` is nicer to use from Kotlin.

Also compute the Corda API bundles to map into each sandbox rather than relying on a hard-coded list, which too easily becomes out-of-date. The selection criteria are:
- It is not a fragment bundle.
- It has the `Corda-Api` attribute in its manifest.
- It exports package matching `net.corda.v5.*`.

### DIFF
--- a/testing/driver/README.md
+++ b/testing/driver/README.md
@@ -134,7 +134,7 @@ public interface DriverDSL {
     @Nullable
     String runFlow(
         @NotNull VirtualNodeInfo virtualNodeInfo,
-        @NotNull Class<?> flowClass,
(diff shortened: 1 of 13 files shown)


mcgovc (approved):
LGTM

mag1c-48 (approved):
LGTM - build file only

yarivus (commented):
thx @chrisr3

PR #4402, opened 2023-08-02
CORE-15344 - Add lookup command

r3-arish:
## Desc

Add lookup command as per the design - ./corda-cli.sh network lookup

This PR introduces the following changes:

1. Adjusts the existing `members` subcommand by renaming it to `members` and adding a `--status` filter option. Also, it has group search through group lookup and holding identity. This allows users to filter members based on their status, with options for "ACTIVE" and "SUSPENDED". By default, only ACTIVE members are filtered. Only an MGM can view suspended members.

2. Adds a new `group-parameters` subcommand. This subcommand is a `RestCommand` that returns the group parameters. It provides information about the group ID, holding identity, and other relevant parameters.

These changes enhance the functionality of the Corda CLI network plugin by allowing users to perform member lookups and retrieve group parameters easily.

## Results
Screenshot 2023-08-02 at 11 47 44
Screenshot 2023-08-02 at 11 18 24
Screenshot 2023-08-02 at 11 20 03
Screenshot 2023-08-02 at 11 36 23
Screenshot 2023-08-02 at 11 39 37
Screenshot 2023-08-08 at 12 19 22


tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/GroupParameters.kt
```diff
@@ -0,0 +1,77 @@
...
+import net.corda.membership.rest.v1.types.RestGroupParameters
+import net.corda.v5.base.types.MemberX500Name
+import picocli.CommandLine
+import java.io.File
+
+@CommandLine.Command(
+    name = "group-parameters",
+    description = ["Lookup group parameters."]
+)
+class GroupParameters(private val output: Output = ConsoleOutput()) : RestCommand(), Runnable {
```
YashNabar:
```suggestion
class GroupParametersLookup(private val output: Output = ConsoleOutput()) : RestCommand(), Runnable {
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/GroupParameters.kt
```diff
@@ -0,0 +1,77 @@
...
+    private fun performGroupParametersLookup(): RestGroupParameters {
+        val holdingIdentity = getHoldingIdentity()
+        val result: RestGroupParameters = createRestClient(MemberLookupRestResource::class).use { client ->
+            val groupParametersProxy = client.start().proxy
+            groupParametersProxy.viewGroupParameters(holdingIdentity)
+        }
+        return result
+    }
+
+    private fun getHoldingIdentity(): String {
```
YashNabar: Can we move this function and `readDefaultGroup()` to a common utility? Looks like it's duplicated in multiple commands

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/Members.kt
```diff
@@ -7,10 +7,12 @@ import net.corda.membership.rest.v1.MemberLookupRestResource
...
 import net.corda.cli.plugins.network.output.Output
 import net.corda.cli.plugins.network.utils.PrintUtils.Companion.printJsonOutput
+import net.corda.v5.base.types.MemberX500Name
 import picocli.CommandLine
+import java.io.File
 
-@CommandLine.Command(name = "members-list", description = ["Shows the list of members on the network."])
-class MemberList(private val output: Output = ConsoleOutput()) : RestCommand(), Runnable {
+@CommandLine.Command(name = "members", description = ["Shows the list of members on the network."])
+class Members(private val output: Output = ConsoleOutput()) : RestCommand(), Runnable {
```
YashNabar:
```suggestion
class MemberLookup(private val output: Output = ConsoleOutput()) : RestCommand(), Runnable {
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/Members.kt
```diff
@@ -61,25 +71,55 @@ class MemberList(private val output: Output = ConsoleOutput()) : RestCommand(),
...
-                country
+                country,
+                status.orEmpty()
             ).members
         }
 
         return result
     }
 
+    private fun getHoldingIdentity(): String {
```
YashNabar: Same comment as above, would be better to define once and reuse

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/GroupParameters.kt
r3-arish: We can go three ways with this - either a utility function, abstract class or a base class with extension

YashNabar: If I'm not wrong, we're not expecting different implementations of these for different commands. So I think extracting them into a separate utils file should be sufficient.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/README.md
```diff
@@ -28,20 +32,27 @@ Alternatively, the following command line arguments can be used to define the st
...
 ./corda-cli.sh network groupPolicy --file="app/build/resources/src.yaml"
 ```
+
 Note:
+
 1. Only one of `memberNames` and `members` blocks may be present.
 2. Single endpoint is assumed for all members when `memberNames` is used.
-3. Endpoint information specified under `members` overrides endpoint information set at the root level. An error is thrown if endpoint information is not provided at all.
+3. Endpoint information specified under `members` overrides endpoint information set at the root level. An error is
+   thrown if endpoint information is not provided at all.
```
YashNabar: Can we undo these formatting changes? It's best to not wrap in markdown files so they can render as per the screen size (some editors don't ignore line breaks)

YashNabar (commented):
LGTM. Just a minor comment

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/README.md
r3-arish: Yes, I have turned off auto-reformatting in my IntelliJ

YashNabar (approved):

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/MemberLookup.kt
```diff
@@ -20,7 +21,15 @@ class MemberList(private val output: Output = ConsoleOutput()) : RestCommand(),
...
-        names = ["-cn"],
+        names = ["-g", "--group"],
+        arity = "1",
+        description = ["Group ID of holding identity performing the lookup. " +
+                "Required if running this command with X.500 name. Defaults to last created group."]
+    )
+    var group: String? = null
+
+    @CommandLine.Option(
+        names = ["-n"],
```
dimosr: That doesn't seem right.. There was an existing parameter `-cn` that was doing a different thing than what the new parameter `-n` is supposed to do. `-cn` was specifying the filter on the returned members list based on canonical name. `-n` is supposed to specify the name on behalf of which vnode to perform the lookup.

YashNabar (dismissed):
Adding another comment related to Dimos's

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/MemberLookup.kt
```diff
@@ -61,19 +70,28 @@ class MemberList(private val output: Output = ConsoleOutput()) : RestCommand(),
...
+        names = ["-s", "--status"],
+        arity = "1..*",
+        description = ["Status (\"ACTIVE\", \"SUSPENDED\") to filter members by. " +
+                "Multiple -s arguments may be provided. By default, only ACTIVE members are filtered. " +
+                "Only an MGM can view suspended members."]
+    )
+    var status: List<String>? = null
 
+    private fun performMembersLookup(): List<RestMemberInfo> {
+        val holdingIdentity = getHoldingIdentity(holdingIdentityShortHash, commonName, group)
```
YashNabar: We should be passing the name (`--name, -n`) here instead of the common name

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/MemberLookup.kt
r3-arish: Thanks. Will look into it. I mistook them as similar - will go through the design document again and make the required changes.

dimosr (dismissed):

YashNabar (approved):

dimosr (approved):

PR #4403, opened 2023-08-02
CORE-13501: Add findKeyAndId for wrapping key

anien:
This is a building block to be able to find the wrapping keys and rotate them.


dickon (approved):
LGTM

PR #4405, opened 2023-08-02
CORE-15574 - Introduce retries for REST calls in e2e test utils

dimosr:
Follow-up from https://github.com/corda/corda-runtime-os/pull/4387 adding retries on a few more requests that weren't included in the previous PR.

### DIFF
--- a/testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
+++ b/testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
@@ -133,8 +133,11 @@ fun ClusterInfo.getOrCreateVirtualNodeFor(
 fun ClusterInfo.getExistingCpi(
     cpiName: String
 ) = cluster {
-    cpiList()
-        .toJson().apply {
+    assertWithRetryIgnoringExceptions {
+        command { cpiList() }
+        condition { it.code == ResponseCode.OK.statusCode }
+        failMessage("Failed to list CPIs")
+    }.toJson().apply {
             assertThat(contains("cpis"))
         }["cpis"]
         .toList()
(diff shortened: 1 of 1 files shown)


testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
```diff
@@ -133,8 +133,11 @@ fun ClusterInfo.getOrCreateVirtualNodeFor(
...
     cpiName: String
 ) = cluster {
-    cpiList()
-        .toJson().apply {
+    assertWithRetryIgnoringExceptions {
+        command { cpiList() }
+        condition { it.code == ResponseCode.OK.statusCode }
+        failMessage("Failed to list CPIs")
+    }.toJson().apply {
             assertThat(contains("cpis"))
```
charlieR3: Unrelated to your change, but I'm not sure this assertion is doing what was intended..

charlieR3 (dismissed):
LGTM

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
dimosr: Not sure I understand.. Isn't it checking the `cpis` key exists before doing a `get` on it?

charlieR3: I think this would do that:
```suggestion
            assertThat(contains("cpis")).isTrue
```

Unless I'm misunderstanding but I think `assertThat(contains("cpis"))` is just creating an assertion object for a boolean but not actually doing any assertions so effectively it's doing nothing of use here.

Actually it's not that important because it would just fail on the next line anyway when accessing the field if it wasn't present.

dimosr: right, I see now. Yes, that makes sense. I can fix it since you spotted it, although I agree not a big deal.

charlieR3 (approved):

PR #4406, opened 2023-08-02
CORE-13501: First part of the rewrapWrappingKey function

anien:
This is not complete, the test does not contain assertion, but it's a good stepping stone towards creating the rewrapping function and the underlying changes for it.


anien:
As mentioned in the description, this is not complete, but it is a good stepping stone towards the rewrap function and shows the underlying changes.

dickon (approved):
LGTM

PR #4407, opened 2023-08-02
CORE-15919: Deduce Corda API bundles instead of using hard-coded lists.

chrisr3:
Remove the Corda API bundles from the sandboxes' lists of public "platform" bundles, because these lists are too difficult to maintain. Deduce which are Corda API bundles instead, according to these criteria:
- It is not a fragment bundle.
- It has the `Corda-Api` attribute in its manifest (which has already been implemented).
- It exports a package matching `net.corda.v5.*`.

Also remove all uses of the JVM `--illegal-access=warn` argument, because this is obsolete in Java 17.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -72,7 +72,6 @@ allprojects {
 
             doFirst {
                 systemProperty 'java.io.tmpdir', buildDir.absolutePath
-                jvmArgs '--illegal-access=deny'
             }
         }
     }
--- a/components/configuration/configuration-read-service-impl/test.bndrun
+++ b/components/configuration/configuration-read-service-impl/test.bndrun
@@ -4,9 +4,6 @@
 -runee: JavaSE-17
 -runtrace: true
 
--runvm: \
-    --illegal-access=deny
-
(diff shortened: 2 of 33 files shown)


ronanbrowne (approved):
LGTM - blt

josephzunigadaly (approved):

driessamyn (approved):
LGTM

pmck91 (approved):
LGTM

dimosr (approved):

vlajos (approved):
Ledger - ok.
I guess this change will make sandbox initialisation a bit slower, but I suppose the difference is negligible.

chrisr3:
> Ledger - ok. I guess this change will make sandbox initialisation a bit slower, but I suppose the difference is negligible.

Actually, it shouldn't because we do this only _once_ when we initialise the `SandboxGroupContextComponent`.

yarivus (approved):
LGTM

relyafi (approved):
Notary changes LGTM

PR #4411, opened 2023-08-02
CORE-13501: implement a crypto service method for rotating a single use of an unmanaged wrapping key

dickon:


anien (approved):
LGTM! 
Title might need a slight change as we have implemented the rewrapping too.

PR #4413, opened 2023-08-03
CORE-13501: update parent key on key rotation

dickon:
Improve test case, allow a second root wrapping key to be optionally specified in test fixture for `SoftCryptoService`.


anien (approved):
Nice test!

PR #4414, opened 2023-08-03
CORE-15933: Apply sandbox service filter to its platform services too.

chrisr3:
Apply `VirtualNodeContext.serviceFilter` when selecting both prototype services and platform services for a sandbox. It was only being applied to the prototype services before, which was creating inconsistent results. However, `serviceFilter` is currently always `null` (i.e. no filter to apply) and so no-one noticed :wink:.

Also remove a lot of useless dependencies from the sandbox code and its tests.


ronanbrowne (approved):
LGTM (build files)

yarivus (approved):
LGTM

simon-johnson-r3 (approved):

PR #4415, opened 2023-08-04
CORE-14685: Migrate HSQLDB-specific tokens into ledger-hsqldb module.

chrisr3:
_(Gaaaaah! I forgot I'd picked up these extra commits from the `release/os/5.1` branch already. But they still need to go into the feature branch.)_

Migrate the HSQLDB-specific tokens into the `:testing:ledger:ledger-hsqldb` module, and convert the parser's "post process" step into the converter's "pre-process" step.

This also means that the processed expression tokens are no longer passed through the validator.

All DBMS-specific functionality has now been migrated into the `VaultNamedQueryConverter` components, which are now genuine OSGi components.


relyafi (approved):
No ledger team review needed at this point as this is a feature branch, changes to be reviewed when merged into release branch

PR #4422, opened 2023-08-04
CORE-15964: Add full HSQLDB support to ledger.

chrisr3:
Create OSGi component `DatabaseProviderType` for selecting either Postgres or HSQLDB versions of DBMS-specific functionality. Currently this includes the following:
- Some non-ANSI native SQL operations for both Consensual and UTXO ledger.
- Custom queries using JSON formatted column data.

Corda selects Postgres by default. The `db-testkit` module will select either Postgres or HSQLDB, depending on whether it detects a `postgresPort` system property value.

Integration tests that were previously disabled for in-memory databases (i.e. HSQLDB) have now been re-enabled.

Note: HSQLDB support is provided _for unit and integration testing purposes only_. It should not be considered to be suitable for Production use.

---
Corda chooses the correct database-specific components using [OSGi service filtering](https://docs.osgi.org/javadoc/osgi.core/8.0.0/index.html?org/osgi/framework/Filter.html), which are LDAP-style rules applied to a service's properties.

Database-specific components are assigned a `@Reference` to a _particular_ `DatabaseTypeProvider` component:
```kotlin
@Component
class PostgresConsensualQueryProvider @Activate constructor(
    @Reference(target = "(database.type=postgres)")
    databaseTypeProvider: DatabaseTypeProvider
): ConsensualQueryProvider {
```
and
```kotlin
@Component
class HsqldbConsensualQueryProvider @Activate constructor(
    @Reference(target = "(database.type=hsqldb)")
    databaseTypeProvider: DatabaseTypeProvider
) : ConsensualQueryProvider {
```
By default, the only `DatabaseTypeProvider` component available in the system is this one:
```kotlin
@Component(service = [ DatabaseTypeProvider::class ], property = [ "database.type=postgres" ])
class DefaultDatabaseTypeProvider : DatabaseTypeProvider {
    override val databaseType: DatabaseType
        get() = POSTGRES
}
```
which can only satisfy the `@Reference` for `PostgresConsensualQueryProvider` and the other Postgres-specific components. However, `db-testkit` deactivates this default and activates an alternative which can be either Postgres or HSQLDB, depending on the value of `DbUtils.isInMemory`. The HSQLDB-specific components cannot be activated without a `DatabaseTypeProvider` that satisfies their `(database.type=hsqldb)` filter. 

The current assumption is that the choice of database is both global and immutable. It does not _need_ to be hard-coded to Postgres, although it currently is.

It would also be theoretically _possible_ for individual `PERSISTENCE` sandboxes to be able to select their own database type.

---
The `DatabaseType` enum is just a convenience so that you can determine which `DatabaseTypeProvider` has actually been selected more easily e.g.
```kotlin
init {
    LoggerFactory.getLogger(this::class.java).info("Activated for {}", databaseTypeProvider.databaseType)
}
```


yarivus (dismissed):
LGTM

buildSrc/src/main/groovy/corda.java-only.gradle
```diff
@@ -0,0 +1,7 @@
+pluginManager.withPlugin('org.jetbrains.kotlin.jvm') {
```
relyafi, line 6: Although this plugin is relatively straightforward, I think we should update the `buildSrc/README.md` for completeness and to document intended usecases.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Keywords.kt
```diff
@@ -4,230 +4,194 @@ interface Keyword : Token
...
     override fun equals(other: Any?): Boolean {
-        if (this === other) return true
-        if (javaClass != other?.javaClass) return false
-        return true
+        return (this === other) || (other is NotEquals)
     }
 
     override fun hashCode(): Int {
-        return javaClass.hashCode()
+        return this::class.java.hashCode()
```
relyafi: It sounds like these are the desired way of doing things going forward based on our discussion. I'd suggest getting buy in from @lankydan / @driessamyn and updating the [Kotlin code standards](https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4082204742/Kotlin+code+standards) page. This increases the chance that people will then automatically follow these and avoids unnecessarily bloating PRs with these changes.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/AbstractVaultNamedQueryConverterImpl.kt
```diff
@@ -0,0 +1,89 @@
...
+import net.corda.ledger.persistence.query.parsing.ParameterEnd
+import net.corda.ledger.persistence.query.parsing.PathReference
+import net.corda.ledger.persistence.query.parsing.PathReferenceWithSpaces
+import net.corda.ledger.persistence.query.parsing.RightParentheses
+import net.corda.ledger.persistence.query.parsing.Select
+import net.corda.ledger.persistence.query.parsing.SqlType
+import net.corda.ledger.persistence.query.parsing.Token
+import net.corda.ledger.persistence.query.parsing.Where
+
+abstract class AbstractVaultNamedQueryConverterImpl : VaultNamedQueryConverter {
```
relyafi, line 30: This feels over-complicated to me. We've added two new functions that now must be implemented (`preProcess` and `customConvert`), even if the implementation doesn't care about them (the Postgres implementation is already a case in point). And we've also prevented further manipulation of the `convert` function - it's a big assumption that different SQL implementations won't want to parse some of the "standard" tokens into different output strings.

Why force any such structure on an implementation, when all it really saves is the few lines we have in this class in `convert`? And even that is making assumptions about how a derived class will want to do conversion.

There is value in sharing the logic within `convert`, which is effectively shared ANSI SQL conversion, but it feels this could be achieved by having a single function, e.g. `convertToAnsiSql`, which could even return null to indicate it didn't match a supplied token. This could either be a protected function in this abstract class (with no other functions), or even just a standalone helper function. This allows common logic to be shared, whilst being un-opinionated on how a converter class performs its processing.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/AbstractVaultNamedQueryConverterImpl.kt
```diff
@@ -0,0 +1,89 @@
...
+            is GreaterThanEquals -> " >= "
+            is LessThan -> " < "
+            is LessThanEquals -> " <= "
+            is In -> " IN "
+            is IsNull -> " IS NULL "
+            is IsNotNull -> " IS NOT NULL "
+            is LeftParentheses -> "("
+            is RightParentheses -> ")"
+            is JsonKeyExists -> " \\?\\? "
+            is JsonCast -> "\\:\\:${token.value}"
```
relyafi: I'm guessing these aren't ANSI SQL conversions? If so, shouldn't these be part of the Postgres implementation?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Tokens.kt
```diff
@@ -2,38 +2,34 @@ package net.corda.ledger.persistence.query.parsing
 
 interface Token
 
-class LeftParentheses : Token {
+interface HasLeftParenthesis
+
+class LeftParentheses : Token, HasLeftParenthesis {
```
relyafi: This seems odd - the name sounds like that of a function rather than representing an "is-a" relationship, and is now asymmetric with `RightParantheses`.

The HSQLDB implementation should not have any impact on the generic set of tokens. The only usage I can find of this interface in the HSQLDB specific implementation. So I'd suggest that we could avoid the need to make changes to the non-test libraries by having a helper function `hasLeftParentheses` in the HSQLDB library, which simply checks the class is `LeftParantheses` or one of the HSQLDB specific types.

gradle.properties
```diff
@@ -106,7 +106,7 @@ paxJdbcVersion=1.5.3
 assertjVersion=3.24.2
 dom4jOsgiVersion = 2.1.3_1
 hamcrestVersion=2.2
-hsqldbVersion=2.7.1
+hsqldbVersion=2.7.2
```
relyafi: Is this necessary to work with these changes. If not, it should probably be separate to this PR?

libs/crypto/cipher-suite-impl/src/test/kotlin/net/corda/cipher/suite/impl/PlatformDigestServiceImplTest.kt
```diff
@@ -124,9 +125,9 @@ class PlatformDigestServiceImplTest {
                 "54FC9D3A6B3BEA1ECA8CC4E6BA1CF4DE78D8822B3EA724DE9D6C", hash.toString())
     }
 
+    @EnabledForJreRange(min = JAVA_11)
```
relyafi, line 128: The changes to this file look completely unrelated to this PR, and should be fixed separately?

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/DefaultDatabaseTypeProvider.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import net.corda.orm.DatabaseType
+import net.corda.orm.DatabaseType.POSTGRES
+import net.corda.orm.DatabaseTypeProvider
+import net.corda.orm.DatabaseTypeProvider.Companion.POSTGRES_TYPE
+import org.osgi.service.component.annotations.Component
+
+@Suppress("unused")
+@Component(service = [ DatabaseTypeProvider::class ], property = [ POSTGRES_TYPE ])
+class DefaultDatabaseTypeProvider : DatabaseTypeProvider {
```
relyafi, line 11: I'm not sure this is necessary? At present, a real Corda setup can hard-code to use the Postgres provider. Once multiple DBs are supported, it will need to be adapted to select an appropriate provider based on the vnode config. As vnodes must have a DB connection string, and the type can be inferred from that, I don't see the value in having a code-level default provider?

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/EntityManagerFactoryFactoryImpl.kt
```diff
@@ -58,8 +58,8 @@ class EntityManagerFactoryFactoryImpl(
     ): EntityManagerFactory {
         return create(
             persistenceUnitName,
-            entities.map { it.canonicalName },
-            entities.map { it.classLoader }.distinct(),
+            entities.map(Class<*>::getCanonicalName),
+            entities.map(Class<*>::getClassLoader).distinct(),
```
relyafi, line 62: Again changes in this file don't seem to have any relation to the PR? Cleanup work in areas we are not otherwise touching should be done separately to avoid bloating / hiding the intent of this PR.

libs/db/db-orm/src/main/kotlin/net/corda/orm/DatabaseType.kt
```diff
@@ -0,0 +1,18 @@
...
+    companion object {
+        const val DATABASE_TYPE = "database.type"
+        const val POSTGRES_TYPE = "$DATABASE_TYPE=postgres"
+        const val POSTGRES_TYPE_FILTER = "($POSTGRES_TYPE)"
+        const val HSQLDB_TYPE = "$DATABASE_TYPE=hsqldb"
+        const val HSQLDB_TYPE_FILTER = "($HSQLDB_TYPE)"
+    }
+
+    val databaseType: DatabaseType
+}
```
relyafi, line 18: I'm not sure this is adding value? The DB specific types can be computed where necessary from `database.type=${DatabaseType.value}`, and the "filter" constants are simply wrapping these in brackets?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/PostgresUtxoQueryProvider.kt
```diff
@@ -0,0 +1,105 @@
...
+import net.corda.orm.DatabaseTypeProvider
+import net.corda.orm.DatabaseTypeProvider.Companion.POSTGRES_TYPE_FILTER
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.Reference
+import org.slf4j.LoggerFactory
+
+@Suppress("unused")
+@Component
+class PostgresUtxoQueryProvider @Activate constructor(
```
relyafi, line 12: This arguably duplicates the purpose of `UtxoRepositoryImpl`, which is intended to abstract the database. However I can see some value in doing this as it avoids a load of repeated boiler plate Hibernate code in the different implementations.

I'm ok with this approach, however if we're going to do this then I think we should also migrate all of the read queries  as well. As it stands, read queries are formed within `UtxoRepositoryImpl`. These are DB specific too, for example MySQL uses `@` instead of `:` for parameters.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoQueryProvider.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    fun persistTransactionVisibleStates(consumed: Boolean): String
+
+    val persistTransactionSignature: String
+
+    val persistTransactionSource: String
+
+    val persistTransactionStatus: String
+
+    val persistSignedGroupParameters: String
```
relyafi: It feels a bit wrong to expose `val` attributes on an interface. Wouldn't it be more correct to have all of these as (mostly zero argument) functions?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/consensual/impl/ConsensualQueryProvider.kt
```diff
@@ -0,0 +1,20 @@
+package net.corda.ledger.persistence.consensual.impl
+
+import net.corda.ledger.common.data.transaction.TransactionStatus
+
+interface ConsensualQueryProvider {
+    companion object {
+        @JvmField
+        val UNVERIFIED = TransactionStatus.UNVERIFIED.value
+    }
```
relyafi: I appreciate this has just been moved, but I wonder if this is actually adding any meaningful value rather than just using the underlying value directly when needed?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoQueryProvider.kt
```diff
@@ -0,0 +1,28 @@
+package net.corda.ledger.persistence.utxo.impl
+
+import net.corda.ledger.common.data.transaction.TransactionStatus
+
+interface UtxoQueryProvider {
+    companion object {
+        @JvmField
+        val UNVERIFIED = TransactionStatus.UNVERIFIED.value
+    }
```
relyafi: I appreciate this has just been moved, but I wonder if this is actually adding any meaningful value rather than just using the underlying value directly when needed?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/consensual/impl/ConsensualQueryProvider.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    val persistTransaction: String
+
+    val persistTransactionComponentLeaf: String
+
+    val persistTransactionStatus: String
+
+    val persistTransactionSignature: String
+
+    val persistTransactionCpk: String
```
relyafi: Should these all be (zero argument) functions, given this is an interface?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoQueryProvider.kt
```diff
@@ -0,0 +1,28 @@
+package net.corda.ledger.persistence.utxo.impl
+
+import net.corda.ledger.common.data.transaction.TransactionStatus
+
+interface UtxoQueryProvider {
```
relyafi, line 12: Even though this is only internal API, I think it would be worth adding KDoc comments to our newly added interfaces, to assist those unfamiliar with the code, given the sheer number of different layers we have now.

relyafi (commented):
Although there's quite a few comments, I think the overall approach is sound and these are mostly about what is done where. The service filtering approach to selection looks like an elegant way of solving the selection problem.

Just a heads up that its entirely conceivable that in the medium term we will need to support different vnode vault DBs (and indeed other DB types) using different database implementations. The description alludes to this being possible in theory. Just wanted to confirm that the service filtering will work on a per-sandbox basis?

libs/db/db-orm/src/main/kotlin/net/corda/orm/DatabaseType.kt
chrisr3: The "added value" is in people not making typos when they write these things.

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/EntityManagerFactoryFactoryImpl.kt
chrisr3: The overhead of "cleanup work" in separate PRs has simply become too great ever since we stopped allowing NOTICK and insisted on everything having its own JIRA number.

These changes should be neither onerous nor controversial.

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/DefaultDatabaseTypeProvider.kt
chrisr3: > At present, a real Corda setup can hard-code to use the Postgres provider.

Except that our integration tests use HSQLDB locally and use Postgres on Jenkins. So hardcoding this to Postgres would be a regression. Also bear in mind that this PR has been carved out of my "driver" feature branch, where replacing the default Postgres implementation with HSQLDB is _the entire point_ of this change in the first place.

> Once multiple DBs are supported, it will need to be adapted to select an appropriate provider based on the vnode config.

I actually asked @simon-johnson-r3 about this, and he was under the impression that we were _not_ going to support individual sandboxes choosing their own database type. It sounds like different people have different understandings of what "mutiple DBs" implies :thinking:.

libs/crypto/cipher-suite-impl/src/test/kotlin/net/corda/cipher/suite/impl/PlatformDigestServiceImplTest.kt
chrisr3: I removed all uses of JUnit `Assumptions` as part of this PR.

gradle.properties
chrisr3: Why _not_ upgrade to the latest version of HSQLDB as part of this PR? What would be the point of implementing this functionality for anything _less_ than the latest version?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Tokens.kt
chrisr3: I added the `HasLeftParenthesis` interface because the new `HsqldbXXX` tokens represent _functions_, which means their expansions _by their very nature_ include a left parenthesis, e.g. `JsonFieldAsText(`, `JsonFieldAsObject(` and `HasJsonKey(`.

> and is now asymmetric with RightParantheses

When someone creates a token that maps to something like `xxxx)` or `)xxxx` then by all means, create a `HasRightParenthesis` interface too. But as of now, no such token exists.

> The HSQLDB implementation should not have any impact on the generic set of tokens.

The original set of token was created with just Postgres in mind, and it didn't consider the use of brackets in custom query expressions very deeply. There should be no issue with expanding the capabilities of these token to accommodate a broader usage.

The HSQLDB code counts left and right parentheses in order to compute "next" or "previous" expressions in the query expression, and this code is cleaner for only needing to spot either `RightParentheses`, or tokens that implement `HasLeftParenthesis`.

---
I _have_ currently resisted the temptation to fix the _egregious_ spelling mistake here that "Parentheses" is actually the _plural_ form! The singular is "Parenthesis".

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/AbstractVaultNamedQueryConverterImpl.kt
chrisr3: @lankydan, can you answer this? Didn't you say that you had investigated what Oracle's and MS Server's JSON support look like?

---
However, it might be instructive to examine these tokens a little more deeply...

Imagine a custom query expression `(a)::int`. Corda would initially parse that to:
```
LeftParentheses, PathReference('a'), RightParentheses, JsonCast('int')
```
which could be mapped trivially to Postgres-compatible SQL. However, HSQLDB would first need to transform that to:
```
HsqldbJsonCast, LeftParentheses, PathReference('a'), RightParentheses, As(), SqlType('int'), RightParentheses
```
before mapping the tokens to useful SQL. Other databases could easily need to perform this transformation too.

In cases like this, the original `JsonCast` token _simply has no meaning_ because its semantics are wrong :shrug:. Which presumably means that its mapping _could_ be moved into Postgres's `customConvert` function.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/AbstractVaultNamedQueryConverterImpl.kt
chrisr3: - We currently have exactly _**two**_ implementations.
- I am assuming that each token has a _constant_ mapping to SQL. Any database type that requires a different mapping for any token can apply the appropriate transformations in `preProcess` (e.g. swap that token with a new database-specific one, rearrange existing tokens around it, and even inject new tokens) and then provide that token's new SQL mapping in `customConvert`.

I would therefore expect a database to need to override both `preProcess` and `customConvert` unless its JSON handling were _exactly_ like Postgres's.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Keywords.kt
chrisr3: It is both relevant and improves the code, and so I do not agree with your use of the word "bloating".

chrisr3:
> Just a heads up that its entirely conceivable that in the medium term we will need to support different vnode vault DBs (and indeed other DB types) using different database implementations. The description alludes to this being possible in theory. Just wanted to confirm that the service filtering will work on a per-sandbox basis?

The `EntitySandboxServiceImpl` could set the `VirtualNodeContext.serviceFilter` to something other than `null`, which would restrict both the prototype and platform services included inside the `PERSISTENCE` sandbox. E.g. a filter requiring that _either_ the service has `database.type=postgres` _or_ has no `database.type` property at all.

There's _slightly_ more to it than that, but not much.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/consensual/impl/ConsensualQueryProvider.kt
chrisr3: > Should these all be (zero argument) functions, given this is an interface?

It's a Kotlin interface, and so these effectively _are_ zero-argument getter functions beneath the Kotlin "crust".

Curiously, if this were implemented instead as a Java interface of zero-argument getter functions then Kotlin would complain if you _didn't_ use them as properties.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoQueryProvider.kt
chrisr3: I'm guessing the point is that this needs to be a `String` rather than an `enum`.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/AbstractVaultNamedQueryConverterImpl.kt
relyafi: I'm not saying that it can't work the way it's currently been implemented, only that it seems to be unnecessarily complex. I think what I've suggested achieves the same end result whilst being a bit more flexible?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Tokens.kt
relyafi: Again though, I think what I'm suggesting achieves the same thing, without introducing this complication to the tokens?

gradle.properties
relyafi: >  What would be the point of implementing this functionality for anything less than the latest version?

This is a rhetorical question, right? We frequently don't use the latest versions. How often do you see people checking and bumping every version dependency every time they raise a PR?

On the off-chance this change introduces an issue, good luck tracking it down and finding a 50+ file PR that it's linked to, with zero explanation on the PR as to why we made the change. Any change in version also introduces additional risk that it could in theory cause a problem.

It's a common theme along with the unnecessary "cleanup" tasks that have been done as part of this PR - it's good practice to keep PRs focused on the task at hand, some cleanup may be justified if we're changing the same code in a file anyway but it should otherwise be avoided in my opinion.

This means diffs are smaller and easier to digest, both now and if someone comes along to understand the scope of the PR in future.

libs/crypto/cipher-suite-impl/src/test/kotlin/net/corda/cipher/suite/impl/PlatformDigestServiceImplTest.kt
relyafi: How does this change relate to the removal of `Assumptions`?

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/DefaultDatabaseTypeProvider.kt
relyafi: > Except that our integration tests use HSQLDB locally and use Postgres on Jenkins. So hardcoding this to Postgres would be a regression.

Is it not a case of making the "workers" use the Postgres provider, and the driver use HSQLDB, and for unit / integration tests, then the test library should select the appropriate one?

 > It sounds like different people have different understandings of what "mutiple DBs" implies 🤔. It sounds like different people have different understandings of what "mutiple DBs" implies 🤔.

We don't need to support them _yet_. So if there are quick wins that don't make that support more difficult in future, I think we should take those.

That said, even in the short term, we will need to support a single _configurable_ database type, which will require selection of something other than Postgres. And a default at the code level then doesn't make sense (it will likely be some form of config if there is a default).

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/EntityManagerFactoryFactoryImpl.kt
relyafi: So it can still be done, we just need a ticket to do the tidy up work under.

But probably the first step is to actually get buy in for agreed standards on these things, which then minimizes the amount of cleanup in future anyway.

I think this change is unjustified because we've not even had to make any other changes in the same file.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/PostgresUtxoQueryProvider.kt
chrisr3: > This arguably duplicates the purpose of `UtxoRepositoryImpl`, which is intended to abstract the database. However I can see some value in doing this as it avoids a load of repeated boiler plate Hibernate code in the different implementations.

It only provides the Postgres-specific non-ANSI SQL statements. It does not execute anything. The point is that `UtxoRepositoryImpl` receives a `UtxoQueryProvider` component that is appropriate to the required `database.type`.

> As it stands, read queries are formed within `UtxoRepositoryImpl`. These are DB specific too, for example MySQL uses `@` instead of `:` for parameters.

These are still _Hibernate_ native queries, and [named query parameters are a documented Hibernate feature](https://docs.jboss.org/hibernate/orm/5.6/userguide/html_single/Hibernate_User_Guide.html#sql-query-parameters). I am therefore sceptical that it should work any differently on MySQL.... :thinking:. I would still expect such differences to be concealed by the `Dialect`, even for native queries.

---
Hibernate native queries are not the same as raw JDBC queries. I wouldn't have needed to modify HSQLDB's `Dialect` using a `MetadataBuilderContributor` if they were.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/AbstractVaultNamedQueryConverterImpl.kt
chrisr3: HSQLDB does not implement the JSON operators that Postgres does, and must therefore generate a token stream where the tokens have both a different mapping and a different grammar. What greater flexibility than that can you be seeking here?

I think we are seeing the tokens differently: you seem to be assuming that we may want to map a given token to different text, depending on the database type, However, this assumes that each token's grammar remains constant across all database types too. Conversely, I am assuming that the tokens' text and grammar do not change, and that if a database type requires something different then it will substitute a new token with the correct semantics, adjusting the surrounding tokens too if necessary.

For Postgres:
```sql
WHERE (visible_state.custom_representation)->(a)->>(b)::int = :value
```
For HSQLDB:
```sql
WHERE CAST( JsonFieldAsText( JsonFieldAsObject( CAST(visible_state.custom_representation AS VARCHAR(32768)), ('a')), ('b')) AS int) = :value
```

I _really_ can't see that I've made _any_ "assumptions" about how different databases will want to rewrite this SQL, other than that it must always read from left to right.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Tokens.kt
chrisr3: But your suggestion adds unwarranted complexity to the HSQLDB code. The `HasLeftParenthesis` interface is a perfectly reasonable _semantic_ addition to the tokens.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/AbstractVaultNamedQueryConverterImpl.kt
relyafi: > I think we are seeing the tokens differently: you seem to be assuming that we may want to map a given token to different text, depending on the database type

I'm not assuming - both myself and @lankydan have said repeatedly that the tokens are in fact DB agnostic and represent tokens of the Corda custom query language, that's how it has been designed. We also said last week when we discussed the tokens that it is acceptable to have separate or specialised tokens for specific implementations to aid conversion, so long as these are private to that particular DB converter.

I still think that we haven't really addressed the substance of my original comment - what exactly are the challenges with following the suggested approach? It means that:

- An abstract class or a standalone helper function is defined which has a single protected function that _can_ be used by concrete implementations if desired

By contrast, the current approach:

- Locks down `convert` so that it cannot be used by concrete classes
- Forces concrete classes to implement `preProcess` and `customConvert` even if they don't care about these things, as evidenced already in `PostgresVaultNamedQueryConverter`.

Both solutions achieve the same thing, but to me it's a significant difference in terms of interface complexity.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Tokens.kt
relyafi: I'd rather introduce the complexity to a specific (and in this case test) implementation, rather than the general implementation. I'd also argue that it's no more complex than what has been done, it just makes the complexity specific rather than general.

If you wanted, we could even keep `HasLeftParenthesis` but make it specific to HSQLDB. That would technically be more correct ("has" implies composition and `LeftParenthesis` "is"), keeps the scope specific, and all it means is you need to check for either `LeftParenthesis` or `HasLeftParenthesis` in the HSQLDB code.

> The HasLeftParenthesis interface is a perfectly reasonable semantic addition to the tokens.

It is not a necessary one in the context of tokenising the Corda custom query language. To re-iterate what I said in an earlier comment, I'd focus on making changes in-line with how this has been designed, rather than how you would have designed it - which may have been an equally valid route to take, but it's not what we've done.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/PostgresUtxoQueryProvider.kt
relyafi: Ok, that's fair enough - I hadn't appreciated native queries weren't really native queries.

There is in my view a design decision to take as to which _interface_ is providing query functionality. Either:

- `UtxoRepository` is responsible for database access generally, in which case we would have a base (abstract) `AnsiSqlUtxoRepositoryImpl` and DB-specific overrides of this, implementing the write queries
- `UtxoQueryProvider` is responsible for providing queries generally, in which case we would have a base (abstract) `AnsiSqlUtxoQueryProvider`, and DB-specific overrides of this, implementing the write queries

Either way, this avoids splitting a functional task across two different interfaces, and in my view fits better with the SOLID "single-responsibility" principle.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoQueryProvider.kt
relyafi: Yes that makes sense, although it's only used in one query currently, so could be inlined. That said, it's conceivable this will end up used in multiple queries as the functionality evolves, so happy to leave as is.

Question: is `@JvmField` now the idiomatic way to remove generating getters and setters, because I think previously you were using `private` vals to achieve this? Or is using `private` still preferred, it's just we can't do that here because it's in an interface and that would stop derived classes from using it?

relyafi:
> > Just a heads up that its entirely conceivable that in the medium term we will need to support different vnode vault DBs (and indeed other DB types) using different database implementations. The description alludes to this being possible in theory. Just wanted to confirm that the service filtering will work on a per-sandbox basis?
> 
> The `EntitySandboxServiceImpl` could set the `VirtualNodeContext.serviceFilter` to something other than `null`, which would restrict both the prototype and platform services included inside the `PERSISTENCE` sandbox. E.g. a filter requiring that _either_ the service has `database.type=postgres` _or_ has no `database.type` property at all.
> 
> There's _slightly_ more to it than that, but not much.

Ok, so if it's at the persistence sandbox level, does that mean that all DBs (crypto, vault, uniqueness) must be the same type with this approach?

If so, I don't think that's a problem for us in the short term, nor should we jump through hoops to resolve that now. Just something we will need to keep in mind.

driessamyn:
Given that longer term we want to move away from OSGi for DI, do we really want to tie the DB type selection to OSGi specific features?

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/DefaultDatabaseTypeProvider.kt
driessamyn: > I actually asked @simon-johnson-r3 about this, and he was under the impression that we were _not_ going to support individual sandboxes choosing their own database type. It sounds like different people have different understandings of what "mutiple DBs" implies 🤔.

Not sure what the background of this is, but @relyafi is correct, that the type of DB will be VNode dependent once the "bring your own DB" feature is implemented.

relyafi:
> Given that longer term we want to move away from OSGi for DI, do we really want to tie the DB type selection to OSGi specific features?

I think this also ties in with my comments earlier about selecting based on (vnode) configuration. I think it should be possible to instead use a factory model which doesn't use OSGi, and uses either the Hibernate dialect (simple, but introduces Hibernate dependency) or the JDBC connection metadata (bit more involved, but still do-able) to select appropriate implementations. We could have different factory implementations:

- Corda workers: The "real" factory, currently only handles Postgres DBs, throws exception otherwise
- Test driver: Always selects HSQLDB (assuming that's all we support?)
- Unit / integration test: Has its own implementation based on the provisioned test DB (currently either HSQLDB or Postgres)

driessamyn:
> > Given that longer term we want to move away from OSGi for DI, do we really want to tie the DB type selection to OSGi specific features?
> 
> I think this also ties in with my comments earlier about selecting based on (vnode) configuration. I think it should be possible to instead use a factory model which doesn't use OSGi, and uses either the Hibernate dialect (simple, but introduces Hibernate dependency) or the JDBC connection metadata (bit more involved, but still do-able) to select appropriate implementations. We could have different factory implementations:
> 
>     * Corda workers: The "real" factory, currently only handles Postgres DBs, throws exception otherwise
> 
>     * Test driver: Always selects HSQLDB (assuming that's all we support?)
> 
>     * Unit / integration test: Has its own implementation based on the provisioned test DB (currently either HSQLDB or Postgres)

I agree that that would be a simpler, easier to understand, design. We don't _need_ to use OSGi just because it's available.
I'm hoping that soon we can look at removing OSGi from the workers that don't need a sandbox, and this includes the Crypto worker, which will also need multi-DB support.

chrisr3:
> > > Given that longer term we want to move away from OSGi for DI, do we really want to tie the DB type selection to OSGi specific features?

Our sandboxes cannot exist without OSGi.
The ledger cannot exist without sandboxes.

There simply _is_ no Corda 5 without OSGi, and trying to pretend otherwise is a ridiculous waste of everyone's time and effort.

driessamyn:
> > > > Given that longer term we want to move away from OSGi for DI, do we really want to tie the DB type selection to OSGi specific features?
> 
> Our sandboxes cannot exist without OSGi. The ledger cannot exist without sandboxes.
> 
> There simply _is_ no Corda 5 without OSGi, and trying to pretend otherwise is a ridiculous waste of everyone's time and effort.

The majority of our workers don't need a Sandbox, hence could work without a Sandbox.
The crypto worker is a good example. It needs DB access, but doesn't need a Sandbox or OSGi.

My main concern is that because we have OSGi in the stack doesn't mean we have to use it. I think the design @relyafi describes is simpler, for example, and would avoid leaking OSGi concerns everywhere.

chrisr3:
> The majority of our workers don't need a Sandbox, hence could work without a Sandbox. The crypto worker is a good example. It needs DB access, but doesn't need a Sandbox or OSGi.

The `flow-worker` and `db-worker` both need sandboxes, and also need crypto. We also need crypto _inside_ the sandboxes, which _requires_ that the crypto code understands OSGi too. OSGi services also allow different modules to "use" each other without coupling them strongly together. The sandboxes depend _heavily_ on this kind of functionality when they request all `PROTOTYPE` services that also advertise `UsedByFlow`, `UsedByVerification` and/or `UsedByPersistence`. Anything that _isn't_ OSGi is invisible to this process and hence _cannot_ be used inside a sandbox - and that includes _platform_ services too. AMQP and Kryo serializers are another good example. Do you think that everything with such a serializer is strongly coupled to either AMQP or Kryo? Do you **_want_** it to be?

Quasar is now an OSGi `WeavingHook` implemented inside an OSGi framework extension bundle (i.e. a fragment of the system bundle). This means that Quasar can now only instrument classes that _belong_ to OSGi bundles within an OSGi framework.

So when you talk about "moving away from OSGi for DI", I think you're really admitting that you don't really understand how much work the OSGi framework is quietly doing for us.

chrisr3:
> > > Just a heads up that its entirely conceivable that in the medium term we will need to support different vnode vault DBs (and indeed other DB types) using different database implementations. The description alludes to this being possible in theory. Just wanted to confirm that the service filtering will work on a per-sandbox basis?
> > 
> > 
> > The `EntitySandboxServiceImpl` could set the `VirtualNodeContext.serviceFilter` to something other than `null`, which would restrict both the prototype and platform services included inside the `PERSISTENCE` sandbox. E.g. a filter requiring that _either_ the service has `database.type=postgres` _or_ has no `database.type` property at all.
> > There's _slightly_ more to it than that, but not much.
> 
> Ok, so if it's at the persistence sandbox level, does that mean that all DBs (crypto, vault, uniqueness) must be the same type with this approach?
> 
> If so, I don't think that's a problem for us in the short term, nor should we jump through hoops to resolve that now. Just something we will need to keep in mind.

It would mean that everything _within a single sandbox_ would be expected to use the same database type, and that components that would expect to use different database types in different sandboxes would need to become `PROTOTYPE`s.

However, AFAIK the current requirement is _still_ only for a single database type that is determined at start-up.

relyafi:
> It would mean that everything _within a single sandbox_ would be expected to use the same database type, and that components that would expect to use different database types in different sandboxes would need to become `PROTOTYPE`s.
> 
> However, AFAIK the current requirement is _still_ only for a single database type that is determined at start-up.

Hmm, we might get away with it then, and there might be a route to making it work dynamically. You're right that a single DB type is the current requirement, but it definitely isn't a tenable position long term (think about a managed Corda cluster where people want to bring their own, disparate databases).

Changing to the sort of solution I suggested / Dries is advocating for would however I think be simpler, and would put the argument to bed once and for all - a factory approach will definitely work even in the face of this requirement, since it will simply instantiate the correct type based on the database connection information.

chrisr3:
> ..., but it definitely isn't a tenable position long term (think about a managed Corda cluster where people want to bring their own, disparate databases).

Doesn't Corda have a REST API for bridging configurations like that? Your "managed Corda cluster" assumes that everyone would have all JDBC drivers for everyone else's databases, and that someone with a different database who wished to join the cluster would need everyone else to upgrade their workers to include a new JDBC driver.

relyafi:
> Doesn't Corda have a REST API for bridging configurations like that? Your "managed Corda cluster" assumes that everyone would have all JDBC drivers for everyone else's databases, and that someone with a different database who wished to join the cluster would need everyone else to upgrade their workers to include a new JDBC driver.

Depends who we mean by everyone - the cluster operator needs to have all necessary JDBC drivers, yes. But a given virtual node operator would just access their virtual node via REST endpoints, and would have their own access setup for their own individual database(s)? Clearly vnode operators can't access other peoples databases.

I think it's more likely that we will see different DBs for different holding ids, as opposed to different DBs for different types for a given holding id. But I think even that might be possible - if we consider a C4 like setup, we may for example use Postgres for a vault DB, and Cockroach DB for a uniqueness DB (maybe a bad example as they're both Postgres compatible, but you get the idea).

These are all the sorts of questions that we need to think about as part of the overall delivery of multi database support. Hence if we have an alternative solution here that is a bit simpler, and also we know will just work, then that's one less thing to worry about.

driessamyn:
> > The majority of our workers don't need a Sandbox, hence could work without a Sandbox. The crypto worker is a good example. It needs DB access, but doesn't need a Sandbox or OSGi.
> 
> The `flow-worker` and `db-worker` both need sandboxes, and also need crypto. We also need crypto _inside_ the sandboxes, which _requires_ that the crypto code understands OSGi too. OSGi services also allow different modules to "use" each other without coupling them strongly together. The sandboxes depend _heavily_ on this kind of functionality when they request all `PROTOTYPE` services that also advertise `UsedByFlow`, `UsedByVerification` and/or `UsedByPersistence`. Anything that _isn't_ OSGi is invisible to this process and hence _cannot_ be used inside a sandbox - and that includes _platform_ services too. AMQP and Kryo serializers are another good example. Do you think that everything with such a serializer is strongly coupled to either AMQP or Kryo? Do you **_want_** it to be?
> 
> Quasar is now an OSGi `WeavingHook` implemented inside an OSGi framework extension bundle (i.e. a fragment of the system bundle). This means that Quasar can now only instrument classes that _belong_ to OSGi bundles within an OSGi framework.
> 
> So when you talk about "moving away from OSGi for DI", I think you're really admitting that you don't really understand how much work the OSGi framework is quietly doing for us.

There is a network boundary between the crypto processor and the flow processor. Nothing on the crypto processor side of that boundary needs OSGi, and OSGi leaking into it is hurting us because it means all its dependencies need to be OSGi bundles and it add a lot of friction to our development process.
It is my opinion that if something isn't needed in the Sandbox, it shouldn't be exposed to OSGi. I agree it's something we need to debate and agree on, and hopefully we'll get to that in the next few months.

driessamyn:
> > ..., but it definitely isn't a tenable position long term (think about a managed Corda cluster where people want to bring their own, disparate databases).
> 
> Doesn't Corda have a REST API for bridging configurations like that? Your "managed Corda cluster" assumes that everyone would have all JDBC drivers for everyone else's databases, and that someone with a different database who wished to join the cluster would need everyone else to upgrade their workers to include a new JDBC driver.

It's probably worth having a look at the Corda 5 topology.
Nobody has "eachother's" JDBC drivers. There is a cluster, and a cluster will need to provide JDBC drivers on all DB and Crypto workers, for all the DBs that Corda supports. 
The cluster admin will choose which DB type to use for the cluster DB, and during a VNode onboarding process, if a VNode operator chooses to use the "bring their own DB" feature, their connection details will inform which JDBC driver is used based on which type of DB they're bringing for their Vault, Crypto or Uniqueness DBs.

chrisr3:
> Changing to the sort of solution I suggested / Dries is advocating for would however I think be simpler, and would put the argument to bed once and for all - a factory approach will definitely work even in the face of this requirement, since it will simply instantiate the correct type based on the database connection information.

We already have `DbConnectionManager.getOrCreateEntityManagerFactory(...)` for providing `EntityManagerFactory` instances. What we _don't_ currently have is any means for services like `UtxoRepositoryImpl` to know (reliably) which kind of database the `EntityManager` they've been given corresponds to, e.g. for these APIs:
```kotlin
@Suppress("TooManyFunctions")
interface UtxoRepository {

    /** Retrieves transaction by [id] */
    fun findTransaction(
        entityManager: EntityManager,
        id: String
    ): SignedTransactionContainer?

    /** Retrieves transaction component leafs */
    fun findTransactionComponentLeafs(
        entityManager: EntityManager,
        transactionId: String
    ): Map<Int, List<ByteArray>>

    ...
}
```

I'm sure it would be possible to create something, but I'm not going to start solutionising now.

chrisr3:
> It's probably worth having a look at the Corda 5 topology. Nobody has "each other's" JDBC drivers. There is a cluster, and a cluster will need to provide JDBC drivers on all DB and Crypto workers, for all the DBs that Corda supports.

Have you considered how to upgrade any of the cluster's databases, if that upgrade would also require an updated JDBC driver?

relyafi:
> > Changing to the sort of solution I suggested / Dries is advocating for would however I think be simpler, and would put the argument to bed once and for all - a factory approach will definitely work even in the face of this requirement, since it will simply instantiate the correct type based on the database connection information.
> 
> We already have `DbConnectionManager.getOrCreateEntityManagerFactory(...)` for providing `EntityManagerFactory` instances. What we _don't_ currently have is any means for services like `UtxoRepositoryImpl` to know (reliably) which kind of database the `EntityManager` they've been given corresponds to, e.g. for these APIs:
> 
> ```kotlin
> @Suppress("TooManyFunctions")
> interface UtxoRepository {
> 
>     /** Retrieves transaction by [id] */
>     fun findTransaction(
>         entityManager: EntityManager,
>         id: String
>     ): SignedTransactionContainer?
> 
>     /** Retrieves transaction component leafs */
>     fun findTransactionComponentLeafs(
>         entityManager: EntityManager,
>         transactionId: String
>     ): Map<Int, List<ByteArray>>
> 
>     ...
> }
> ```
> 
> I'm sure it would be possible to create something, but I'm not going to start solutionising now.

I think this bit is the key - it should be entirely possible to determine the database type given access to the database connection manager, which we could either pass in directly (uniqueness already has direct access) or get at it from the EM / EMF. I've had a look online and this is just one way that might work: https://www.baeldung.com/jdbc-database-metadata#database-level-metadata

Here's how we did it in C4 for the notary: https://github.com/corda/enterprise-ha-notary/blob/master/backing-stores/src/main/kotlin/com/r3/hanotary/backingstores/BackingStoreFactory.kt (this just parsed the connection string IIRC, so maybe not the best way)

chrisr3:
> I think this bit is the key - it should be entirely possible to determine the database type given access to the database connection manager, which we could either pass in directly (uniqueness already has direct access) or get at it from the EM / EMF. I've had a look online and this is just one way that might work: https://www.baeldung.com/jdbc-database-metadata#database-level-metadata

Yeah, I saw that. That's _why_ I said "(reliably)".

relyafi:
> > It's probably worth having a look at the Corda 5 topology. Nobody has "each other's" JDBC drivers. There is a cluster, and a cluster will need to provide JDBC drivers on all DB and Crypto workers, for all the DBs that Corda supports.
> 
> Have you considered how to upgrade any of the cluster's databases, if that upgrade would also require an updated JDBC driver?

Not yet, that's another thing that will need to be considered and another problem that will need solving. Definitely beyond the scope of this work though, but just another case in point for why we don't need to create ourselves more problems, we have enough to worry about as-is :)

chrisr3:
> ..., and OSGi leaking into it is hurting us because it means all its dependencies need to be OSGi bundles and it add a lot of friction to our development process.

There's a saying about OSGi: It tells you when your baby is ugly, and most people don't take that news well.

Unfortunately, shooting the messenger doesn't mean your baby _stops_ being ugly :no_mouth:.

We should have solved most of the technical issues by now anyway. And I've seen what Corda looks like _without_ a DI framework too: It's called "Corda 4".... :scream:

chrisr3:
> Not yet, that's another thing that will need to be considered and another problem that will need solving. Definitely beyond the scope of this work though, but just another case in point for why we don't need to create ourselves more problems, we have enough to worry about as-is :)

The OSGi framework is a powerful tool, and neither PERSISTENCE sandboxes nor the ledger can exist without it. In which case, Corda 5 is stuck with it and we might as well use it properly instead.

But regardless, as of _right now_, Corda uses a single database type that is selected at start-up, and that database can be either Postgres or HSQLDB. All this PR is doing is providing HSQLDB support for those few aspects which are currently Postgres-specific. And yes, I am using OSGi for that because OSGi is capable of doing the brunt of the "heavy lifting" for me in a clean and reliable way.

This should be more than sufficient for our requirements _as they currently exist_. All this discussion about possible future requirements is really just noise. The crypto-worker is still using an OSGi framework, and it can be a fun and educational exercise for someone else to change that at some unspecified later date.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoQueryProvider.kt
chrisr3: > Question: is `@JvmField` now the idiomatic way to remove generating getters and setters, because I think previously you were using `private` vals to achieve this?

All `static final` fields on an `interface` _must_ be `public`. The JVM simply forbids anything else, and there's nothing Kotlin can do about it. `private val` can only be used for classes.

driessamyn:
> > It's probably worth having a look at the Corda 5 topology. Nobody has "each other's" JDBC drivers. There is a cluster, and a cluster will need to provide JDBC drivers on all DB and Crypto workers, for all the DBs that Corda supports.
> 
> Have you considered how to upgrade any of the cluster's databases, if that upgrade would also require an updated JDBC driver?

Simon is looking at upgrade generally, but I don't see how this one would be an issue. We'd be deploying new pods with the updated driver(s), then upgrading the DB.
This would work unless there is a DB driver that would not be backwards compatible, which would be very strange.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Tokens.kt
chrisr3: > I'd rather introduce the complexity to a specific (and in this case test) implementation, rather than the general implementation.

I would rather not have the complexity in the first place, and so have adjusted the tokens accordingly. The tokens are not perfect and I have subsequently found them lacking with regards to its handling of parentheses, which is also why I was forced to raise and fix CORE-15833. **Any** database that does not implement all four of our chosen JSON operators will have exactly the same issues as HSQLDB to either an equal or a lesser degree, and so I find your dismissal of HSQLDB as a "test" implementation to be patronising and condescending.

> > The HasLeftParenthesis interface is a perfectly reasonable semantic addition to the tokens.
> 
> It is not a necessary one in the context of tokenising the Corda custom query language. To re-iterate what I said in an earlier comment, I'd focus on making changes in-line with how this has been designed, rather than how you would have designed it - which may have been an equally valid route to take, but it's not what we've done.

I took the tokens as I found them and I extended them to work with HSQLDB. _**And I worked very hard to work within their existing implementation!**_ I cannot describe how angry I am that you would suggest otherwise, _over something as simple as the addition of a single semantic interface!_

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/AbstractVaultNamedQueryConverterImpl.kt
chrisr3: The current approach is called "the template method" pattern. (Perhaps you have heard of it?) It separates the parts that _don't_ need to change from the parts that are _expected_ to change.

- The algorithm for concatenating a sequence of tokens together into a text string is something that is _not_ expected to change across databases.
- The text mappings for the standard set of tokens are also not expected to change since their resulting output must always form valid SQL.

What _is_ expected to change is the sequence of tokens itself, i.e. which tokens are present and their ordering. And new tokens will also require new mappings to be defined.

HSQLDB is as different from Postgres is as possible to be: it does not support _any_ of the JSON operators that Postgres does. Whereas (e.g.) [MySQL8 supports only `->` and `->>`](https://dev.mysql.com/doc/refman/8.0/en/json-function-reference.html), meaning that expressions containing `HasJsonKey` and `JsonCast` tokens must be preprocessed into something that MySQL can accept.

It is not remotely surprising that Postgres doesn't need any new tokens or mappings since the tokens were written with Postgres in mind :roll_eyes:.

My experience with the tokens also suggests that I am the first person even to _try_ to use them for anything even remotely non-trivial. You cannot _possibly_ expect them to enter _actual Real World use_ without _any_ changes, can you?

Here's a thought: my `HsqldbJsonCast` token actually maps to ANSI `CAST(`, which means it could reasonably be moved out of `ledger-hsqldb`, thus making _two_ standard tokens with `HasLeftParenthesis`. I'm sure other databases will need to use `CAST(... AS ...)` in their mappings :no_mouth:.

> [B]oth myself and @lankydan have said repeatedly that the tokens are in fact DB agnostic and represent tokens of the Corda custom query language, that's how it has been designed. 

I haven't changed that either, so I don't see your point. But at the end of the day, the tokens are not someone's pretty toy, and I _do_ need to do something _useful_ with them :no_mouth:. And if I see a way that they can give me _more_ information about the custom query expression then I will obviously implement it.

> And we've also prevented further manipulation of the convert function - it's a big assumption that different SQL implementations won't want to parse some of the "standard" tokens into different output strings.

Really? _Show_ me a database where `Equals` would not map to `=`, or `And` to "AND". We _are_ writing SQL text here. But even then, the standard `Equals` token could be mapped to `FunkyEquals` in this _perverse_ database's pre-process step, with `FunkyEquals` mapping to whatever invalid SQL you care to choose in the `customConvert` function.

But I **must** point out - I have not changed how the initial set of tokens is parsed at all, although I have included an extra marker interface for any token with a left parenthesis, to make analysing the tokens easier. This is precisely the kind of addition that is likely to happen when someone tries to use a piece of code for the first time in a non-trivial scenario.

relyafi:
> But regardless, as of right now, Corda uses a single database type that is selected at start-up, and that database can be either Postgres or HSQLDB.

Technically right now, we're still hard coded to Postgres (for a real Corda worker), because we haven't implemented the configuration bit.

In practice, my expectation is the configuration bit will not in fact be a fixed configuration setting, the database type will simply be inferred from the DB connection. It should be perfectly possible to do this reliably - Hibernate manages to select a dialect automatically and reliably, and simply using this is one option.

The fact that we will still only support Postgres in practice is simply a consequence of the fact we won't yet be able to provision different DB types. That seems preferable to introducing an artificial configuration setting that will have to be deprecated later.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Tokens.kt
relyafi: I think probably the best thing to do on this comment (and perhaps a couple of the others that are in a similar vein) is to get @lankydan's input on his return from PTO. That does mean we'll have to wait a couple of weeks, but ultimately I think he's best placed to make a call on this, as the original designer he has the clearest understanding of the intent of this functionality and how it should be evolved.

yarivus (dismissed):
LGTM

yarivus (commented):
LGTM - but I'm not the one to approve ledger stuff

chrisr3:
> LGTM - but I'm not the one to approve ledger stuff

I've done the best I can with the tokens as currently implemented, but I suspect that all I've _really_ done here is highlight the tokens' implementation's underlying problems :no_mouth:.

yarivus (dismissed):
LGTM

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/AbstractVaultNamedQueryConverterImpl.kt
```diff
@@ -0,0 +1,93 @@
...
+import net.corda.ledger.persistence.query.parsing.RightParenthesis
+import net.corda.ledger.persistence.query.parsing.Select
+import net.corda.ledger.persistence.query.parsing.Token
+import net.corda.ledger.persistence.query.parsing.UnaryKeyword
+import net.corda.ledger.persistence.query.parsing.Where
+
+@Suppress("MemberVisibilityCanBePrivate")
+abstract class AbstractVaultNamedQueryConverterImpl : VaultNamedQueryConverter {
+
+    protected abstract fun writeCustom(output: StringBuilder, token: Token)
```
relyafi: I still don't understand why we're forcing concrete implementations to implement this? A sensible default would be to just throw an `IllegalArgumentException` (the else clause used in the concrete implementations).

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/Tokens.kt
relyafi: Closing, as the new approach means this is no longer relevant

relyafi (commented):
I think most of the major comments have now been addressed. I've marked my comments that have been addressed as resolved, so it's worth checking over those that are still open (these are now mostly minor).

In terms of major changes, I still feel we should do as @driessamyn suggested and not use OSGi for DB selection if we can avoid it. If we can just use a standard factory approach for this, I think that is likely to be simpler and more maintainable in the long run, based on what he said.

chrisr3:
> In terms of major changes, I still feel we should do as @driessamyn suggested and not use OSGi for DB selection if we can avoid it.

AFAIK the Dev-Ex team (@yarivus) wants to add the driver to Corda 5.1, so unless production databases other than Postgres are _also_ on the roadmap for Corda 5.1 then I wouldn't recommend adding anything more to this PR. Refactoring the UTXO custom query tokens has _already_ taken a month, and we simply do not have another month to spend re-inventing Corda's DB handling as well.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/parsing/converters/AbstractVaultNamedQueryConverterImpl.kt
chrisr3:
```kotlin
abstract class AbstractVaultNamedQueryConverterImpl : VaultNamedQueryConverter {

    protected open fun writeCustom(output: StringBuilder, token: Token) {
        throw IllegalArgumentException("Invalid token in expression - $token")
    }
```
So now each database must do something like:
```kotlin
    override fun writeCustom(output: StringBuilder, token: Token) {
        when(token) {
            is JsonField -> writeBinaryOperator(output, " -> ", token)
            is JsonArrayOrObjectAsText -> writeBinaryOperator(output, " ->> ", token)
            is JsonCast -> writeBinaryOperator(output, "\\:\\:", token)
            is JsonKeyExists -> writeBinaryOperator(output, " \\?\\? ", token)
            else ->
                super.writeCustom(output, token)
        }
    }
```

relyafi:
> > In terms of major changes, I still feel we should do as @driessamyn suggested and not use OSGi for DB selection if we can avoid it.
> 
> AFAIK the Dev-Ex team (@yarivus) wants to add the driver to Corda 5.1, so unless production databases other than Postgres are _also_ on the roadmap for Corda 5.1 then I wouldn't recommend adding anything more to this PR. Refactoring the UTXO custom query tokens has _already_ taken a month, and we simply do not have another month to spend re-inventing Corda's DB handling as well.

They are on the roadmap for 5.2, and we have a view to delivering these part of the changes for 5.1 so we don't have to do it as part of this PR, but it just means that this bit potentially becomes immediately redundant.

libs/crypto/cipher-suite-impl/src/test/kotlin/net/corda/cipher/suite/impl/PlatformDigestServiceImplTest.kt
chrisr3: Because I removed all lines like:
```kotlin
assumeTrue(JavaVersion.isVersionAtLeast(JavaVersion.JAVA_11))
```

dickon (dismissed):
LGTM for crypto change

khutsijabari:
Hi @aditisdesai, this is in scope for Corda 5.1, it is required as part of the Flow Testing Driver feature that is coupled with Corda 5.1. FYI, the Flow Testing Driver is currently being developed on feature branch off Corda Runtime OS and it will be merged into Corda 5.1 main branch: https://github.com/corda/corda-runtime-os/tree/feature/flow-testing

khutsijabari:
Hi @aditisdesai, I can confirm that this is scope for Corda 5.1. This is required by the Flow Testing Driver feature that is coupled with Corda 5.1 and has to be merged into the Corda Runtime OS main branch. The Flow Testing Driver feature is currently being developed on feature branch based off Corda Runtime OS main branch: https://github.com/corda/corda-runtime-os/tree/feature/flow-testing

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/DefaultDatabaseTypeProvider.kt
relyafi: Discussed that this can be refined and addressed in a subsequent change.

libs/db/db-orm-impl/src/main/kotlin/net/corda/orm/impl/EntityManagerFactoryFactoryImpl.kt
chrisr3: Actually I needed to add `configuration.extraProperties` into the `CustomPersistenceUnitInfo` in this file, because this is how I configure the Hibernate to use my `MetadataBuilderContributor`.

relyafi (approved):
LGTM

dickon (approved):
LGTM for crypto

kyriathar (approved):
OK (db)

mcgovc (approved):
LGTM - build files

PR #4423, opened 2023-08-07
CORE-15285: Change sampling logic to favour flow traces

ZawarR3:
The current configurability of traces is limited to a provided sample rate or unlimited sampling for all system requests, agnostic of their source or purpose. To reduce unnecessary data from being sampled, developers usually define a sample rate of their preference. However, this sample rate will be applied to all requests arbitrarily, meaning that the trace one is interested in might not get sampled.

When it comes to tracing as a whole, we are more interested in flow request sampling. With this change, we ensure that whatever sample rate is defined, is applied to all requests from URI endpoint `/flow` with method `POST,` as well as all other requests from the server. The change currently supports the following REST API versions: `v5_1,`, `v1`.

### DIFF
--- a/libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
+++ b/libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
@@ -7,10 +7,13 @@ import brave.baggage.BaggagePropagationConfig
 import brave.baggage.CorrelationScopeConfig
 import brave.context.slf4j.MDCScopeDecorator
 import brave.http.HttpRequest
+import brave.http.HttpRequestMatchers.methodEquals
 import brave.http.HttpRequestMatchers.pathStartsWith
+import brave.http.HttpRuleSampler
 import brave.http.HttpTracing
 import brave.propagation.B3Propagation
 import brave.propagation.ThreadLocalCurrentTraceContext
+import brave.sampler.Matchers.and
 import brave.sampler.RateLimitingSampler
 import brave.sampler.Sampler
 import brave.sampler.SamplerFunction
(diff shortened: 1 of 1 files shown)


libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
```diff
@@ -90,6 +85,25 @@ internal class BraveTracingService(serviceName: String, zipkinHost: String?, sam
         tracingBuilder.build().also(resourcesToClose::push)
     }
 
+    private fun sampler(samplesPerSecond: SampleRate): Sampler? = when (samplesPerSecond) {
```
josephzunigadaly: Let's make the return type non-nullable

```suggestion
    private fun sampler(samplesPerSecond: SampleRate): Sampler = when (samplesPerSecond) {
```

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
```diff
@@ -90,6 +85,25 @@ internal class BraveTracingService(serviceName: String, zipkinHost: String?, sam
...
+        is PerSecond -> {
+            logger.info("Tracing will sample ${samplesPerSecond.samplesPerSecond} requests per second")
+            RateLimitingSampler.create(samplesPerSecond.samplesPerSecond)
+        }
+
+        is Unlimited -> {
+            logger.info("Tracing will sample unlimited requests per second")
```
josephzunigadaly: We will get these log messages 3 times now.

It's a minor thing, but I'd pull the log statements out into the `tracing` lambda so we only see them once.

ZawarR3: You want me to move the log statements into the by lazy code block for tracing initialisation?

ZawarR3: like this?: ```when (samplesPerSecond) {
            is PerSecond -> {
                logger.info("Tracing will sample ${samplesPerSecond.samplesPerSecond} requests per second")
            }
            is Unlimited -> {
                logger.info("Tracing will sample unlimited requests per second")
            }
        }```

josephzunigadaly: yes please

libs/tracing/src/main/kotlin/net/corda/tracing/impl/BraveTracingService.kt
```diff
@@ -90,6 +94,23 @@ internal class BraveTracingService(serviceName: String, zipkinHost: String?, sam
...
+        }
+
+        is Unlimited -> {
+            Sampler.ALWAYS_SAMPLE
+        }
+    }
+
+    private val serverSampler: SamplerFunction<HttpRequest> = HttpRuleSampler.newBuilder()
+        .putRule(and(methodEquals("POST"), pathStartsWith("/api/v5_1/flow")), sampler(samplesPerSecond))
+        .putRule(and(methodEquals("POST"), pathStartsWith("/api/v1/flow")), sampler(samplesPerSecond))
```
josephzunigadaly, line 109: We need a better approach to identifying flow endpoints. This will stop matching in future releases.

I think we should merge this PR as it is and fix in a follow up PR.

josephzunigadaly (approved):

PR #4424, opened 2023-08-07
CORE-13376: remove unnecessary wrapping types for collections

Tom-Fitzpatrick:
Some endpoints which return a list of objects had wrapping Objects to contain the lists. This PR removes these wrapping objects and directly returns the list in order to have consistency across our endpoints. This was done using the guidelines here: [Versioning of Rest API](https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4472635529/Versioning+of+REST+API)

| Endpoint | Class | Method | Wrapper |  Wrapper contents | 
| ------------- | ------------- | ------------- | ------------- | ------------- |
| /flowclass/{holdingIdentityShortHash  | FlowClassRestResource | getStartableFlows | StartableFlowsResponse | flowClassNames: `List<String>` | 
| /flow/{holdingIdentityShortHash}  | FlowRestResource  | getMultipleFlowStatus | FlowStatusResponses | flowStatusResponses: `List<FlowStatusResponse>` | 
 | /virtualnode | VirtualNodeRestResource | getAllVirtualNodes | VirtualNodes | virtualNodes: `List<VirtualNodeInfo>` | 
 | /cpi | CpiUploadRestResource | getAllCpis | GetCPIsResponse | cpis: `List<CpiMetadata>` | 

Corresponding tests have been added to cover the deprecated endpoints here: https://github.com/corda/corda-e2e-tests/pull/186

Non-functional tests are being updated here: https://github.com/corda/corda-non-functional-test/pull/223

### DIFF
--- a/applications/workers/release/rest-worker/src/e2eTest/kotlin/net/corda/applications/workers/rest/utils/ClusterTestUtils.kt
+++ b/applications/workers/release/rest-worker/src/e2eTest/kotlin/net/corda/applications/workers/rest/utils/ClusterTestUtils.kt
@@ -94,7 +94,7 @@ fun E2eCluster.uploadCpi(
         with(client.start().proxy) {
             // Check if MGM CPI was already uploaded in previous run. Current validation only allows one MGM CPI.
             if (isMgm) {
-                getAllCpis().cpis.firstOrNull {
+                getAllCpisList().firstOrNull {
                     it.groupPolicy?.contains("CREATE_ID") ?: false
                 }?.let {
                     return it.cpiFileChecksum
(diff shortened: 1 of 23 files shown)


driessamyn:
Chatted to @Tom-Fitzpatrick off-line, but to summarise, I don't think this is a good change.
I actually thought Array was technically not allowed as a return type for `application/json`, but it turns out it is. However, I still don't think it's a good idea.
For example, consider today's return type is:
```json
{
  "cpis": [...]
}
```
tomorrow's return type could be:
```json
{
  "cpis": [...],
  "timestamp": "2022-08-09T11:11:11Z"
}
```
and the latter is not a breaking change. In contrast, if returning an array, it cannot be changed, without making it a breaking change.

Returning an array also means that some parsers choke unless you tell them to expect an array rather than an object (see https://stackoverflow.com/questions/5034444/can-json-start-with for discussion), which could be problematic for generic parsers.

Tom-Fitzpatrick:
After chatting with @driessamyn and @josephzunigadaly I agree that it makes more sense not to make this change and instead keep these wrapping objects for our responses. Will close the PR for now.

PR #4428, opened 2023-08-07
CORE-15477: add unit test for StateRef.parse()

jennyang-r3:
### Summary
This PR adds a unit test for `StateRef.parse()` method, testing if:
- valid StateRef is parsed correctly
- value with zero delimiter throws a correct exception
- value with missing index throws a correct exception
- any other malformed value throws transaction Id exception

**Test result:** https://gradle.dev.r3.com/s/tx6aoe4mz6k5m


jennyang-r3:
close since this test is written in API repo instead.

PR #4429, opened 2023-08-08
CORE-15582: Add encoding/transmission of function calls to an EVM Processor outside of a flow

fowlerrr:


components/flow/flow-service/src/main/kotlin/net/corda/flow/application/interop/AbiEncoder.kt
```diff
@@ -0,0 +1,13 @@
+package net.corda.flow.application.interop
+
+const val EVM_FUNCTIONS = "EvmFunctions"
+
+interface AbiEncoder {
+    // Encodes The Function Signature With Parameters
+    fun encodeFunctionSignature(contract: String = EVM_FUNCTIONS, method: String, params: List<Any>): String
+    fun encodeFunctionSignature(contract: String = EVM_FUNCTIONS, method: String, params: Map<String, Any>): String
```
yarivus: why do we need 2 versions of `encodeFunctionSignature` ?
I think we should strive to be simple (i.e. less options)

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/interop/AbiEncoderImpl.kt
```diff
@@ -0,0 +1,159 @@
...
+    val type: String,
+    val internalType: String,
+    val value: Any?,
+    val components: List<AbiContractFunctionInput>?,
+)
+
+data class AbiContractFunction(
+    val name: String?,
+    val inputs: List<AbiContractFunctionInput>,
+    val outputs: List<AbiContractFunctionInput>?,
```
yarivus: are the outputs of an input type ?!

yarivus (commented):
see some comments

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/interop/AbiEncoderImpl.kt
```diff
@@ -0,0 +1,159 @@
...
+                it
+            }
+        }
+        val outputFunction = function.copy(inputs = newParams)
+
+        return jacksonObjectMapper().writeValueAsString(outputFunction)
+    }
+
+    override fun <T> decodeFunctionResult(result: String, clazz: Class<T>): T {
+        return jacksonObjectMapper().readValue(result, clazz)
```
chrisr3: This scares me. Jackson's `ObjectMapper` remembers every `Class` that it deserialises like this, storing the information internally inside `static` collections. So if one of these `Class`es were to belong to a sandbox bundle then you would preventing the JVM from ever releasing that bundle's `ClassLoader`, i.e. you would have a memory leak.

`JsonMarshallingServiceImpl` contains a carefully configured `ObjectMapper`, and also has `PROTOTYPE` scope to ensure that each sandbox uses its own private `JsonMarshallingService` instance.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/interop/EvmServiceImpl.kt
```diff
@@ -0,0 +1,162 @@
...
+
+@Component(
+    service = [SigningService::class, UsedByFlow::class],
+    scope = PROTOTYPE
+)
+class EvmServiceImpl @Activate constructor(
+    private val contractAddress: String,
+    private val rpcUrl: String,
+    private val flowContextProperties: FlowContextProperties,
+    private val externalEventExecutor: ExternalEventExecutor,
```
chrisr3: This cannot work because an `@Activate` constructor's parameters must either be `@Reference`s or [Activation Objects](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#service.component-activation.objects).

To declare something as a `@Component` is to ask the OSGi framework to instantiate it for you. But how could the framework know the correct values for any of these constructor parameters?

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Create.kt
```diff
@@ -64,6 +64,7 @@ class Create(
     private val workersForProcessor = mapOf(
         "crypto" to listOf("crypto", "combined"),
         "db" to listOf("db", "combined"),
+        "evm" to listOf("evm", "combined"),
```
fowlerrr, line 95: Note, this will be part of a different change.  I don't think it belongs in this PR

components/flow/flow-service/src/test/kotlin/net/corda/flow/application/interop/EvmServiceImplTest.kt
```diff
@@ -0,0 +1,196 @@
...
+        val expectedRequest = EvmRequest(
+            "",
+            "from",
+            "to",
+            "rpcUrl",
+            Transaction(
+                "test",
+                TransactionOptions("1", "2", "3", "4"),
+                listOf(
+                    net.corda.data.interop.evm.request.Parameter("one", "INT64", "1"),
```
r3domfox: Is there a reason why some classes are fully qualified in some places (e.g. `net.corda.data.interop.evm.request.Parameter`) but not in others?

r3domfox: Consider using a typed parameter label for parameters, e.g.

```kotlin
data class RequestParameter<T>(name: String, evmType: Type) {
  // Ensures type agreement of parameter and value
  fun of(value: T): RequestParameterValue = RequestParameterValue(this, value)
}

// Extension function, if you want to be cute
fun Type.INT64.named(name: String): RequestParameter<Int> =
    RequestParamter<Int>(name, this)

fun Type.ADDRESS.named(name: String): RequestParameter<String> =
    RequestParameter<String>(name, this)

data class RequestParameterValue(val parameter: RequestParameter<Any>, val value: Any)

val parameterOne = Type.INT64.named("one")
val parameterTwo = Type.ADDRESS.named("two")

val result = service.call<String>(
            "test",
            "to",
            CallOptions("latest", "rpcUrl", "from"),
            parameterOne.of(1), parameterTwo.of("two") // vararg array of RequestParameterValue
        )
```

r3domfox (commented):
Suggestion for a typesafe way of supplying call parameters

components/flow/flow-service/src/test/kotlin/net/corda/flow/application/interop/EvmServiceImplTest.kt
fowlerrr: The reason for the fully qualified names is because the avro message and the user class share the same name.  Which is fine for the user since they never _see_ the avro, but makes this bit of code more awkward.

fowlerrr: I've updated the [API](https://github.com/corda/corda-api/pull/1188) to use types for the values of `Parameter`.  Have a look over there and let me know what you think.  It's not quite as far as you suggested but I think it's probably a good step there.

fowlerrr:
This work is outdated and the main body of effort is in the demo branch.  Closing as we'll make a new PR if we need one for this specific work

PR #2569, opened 2022-11-17
CORE-8135 Check the public key in the session certificate

williamvigorr3:
Check that the session certificate’s public key matches the one used by the peer in the handshake of the authentication protocol. If these differ then the certificate validation.

I tested this following the instructions here: https://github.com/corda/corda-runtime-os/pull/2286#issue-1403608712. i.e. used the fake CA tool to create a session certificate when deploying a dynamic network. Sending 100 message between two peers using the app simulator.

PR #4430, opened 2023-08-08
CORE-15350 Fix check for locally-hosted identities

YashNabar:
Changes the check for locally-hosted identities to include session key matching, to guard against scenarios in which a duplicate holding identity from another cluster attempts to register, but the MGM's decline message is instead routed to a matching holding identity which is hosted locally. 
Also changes the creation of locally-hosted identities in static networks to use the actual session key from `MemberInfo`, instead of a dummy key.

**E2E test**: 

https://github.com/corda/corda-e2e-tests/pull/188

**Manual test:**
1. Onboarded MGM and member ("O=Alice, L=London, C=GB") on cluster-A
2. Create vnode with same name "O=Alice, L=London, C=GB" on cluster-B, and attempt to register with MGM (cluster-A)
Screenshot 2023-08-10 at 11 26 30

Cluster-B Alice receives declined message as expected.

3. Lookup performed by cluster-A Alice:
Screenshot 2023-08-10 at 11 27 42


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/hosting/LinkManagerHostingMapImpl.kt
```diff
@@ -69,6 +72,17 @@ internal class LinkManagerHostingMapImpl(
     override fun isHostedLocally(identity: HoldingIdentity) =
         locallyHostedIdentityToIdentityInfo.containsKey(identity)
 
+    override fun isHostedLocally(member: MemberInfo?): Boolean {
```
williamvigorr3: Ideally we should make it clear as part of the API that this overload is making a more thorough check compared to isHostedLocally(identity: HoldingIdentity). Maybe call it `isHostedLocallyAndSessionKeyMatch` or something.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/hosting/LinkManagerHostingMapImpl.kt
```diff
@@ -69,6 +72,17 @@ internal class LinkManagerHostingMapImpl(
...
+        return if (member == null) {
+            false
+        } else {
+            locallyHostedIdentityToIdentityInfo[member.holdingIdentity]?.let { identityInfo ->
+                identityInfo.allSessionKeysAndCertificates.map { it.sessionPublicKey }
+                    .any { member.sessionInitiationKeys.contains(it) }
```
williamvigorr3: I think `.equals` doesn't work well on cryptographic keys. See https://github.com/corda/corda-runtime-os/pull/2569#discussion_r1025464004

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/hosting/LinkManagerHostingMapImpl.kt
```diff
@@ -69,6 +72,17 @@ internal class LinkManagerHostingMapImpl(
     override fun isHostedLocally(identity: HoldingIdentity) =
         locallyHostedIdentityToIdentityInfo.containsKey(identity)
 
+    override fun isHostedLocally(member: MemberInfo?): Boolean {
+        return if (member == null) {
+            false
```
williamvigorr3: I am not sure about this. If the member is not in the membership group reader we assume it's not locally hosted. Is this definitely what we want?  We could fall back to checking: `locallyHostedIdentityToIdentityInfo.containsKey(identity)` instead.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
```diff
@@ -134,11 +135,12 @@ internal class OutboundMessageProcessor(
         } catch (e: Exception) {
             return "destination '${destination.x500Name}' is not a valid X500 name: ${e.message}"
         }
+        val destinationMemberInfo = membershipGroupReaderProvider.lookup(cordaSource, cordaDestination, filter)
         return if (source.groupId != destination.groupId) {
             "group IDs do not match"
         } else if (!linkManagerHostingMap.isHostedLocally(cordaSource)) {
             "source ID is not locally hosted"
-        } else if (linkManagerHostingMap.isHostedLocally(cordaDestination)) {
+        } else if (linkManagerHostingMap.isHostedLocally(destinationMemberInfo)) {
```
williamvigorr3: I think it might be cleaner, to put the lookup in the `membershipGroupReaderProvider` inside the `LinkManagerHostingMap`. Something like:
```
    fun isHostedLocallySource(identity: HoldingIdentity): Boolean

    fun isHostedLocallyDest(identity: HoldingIdentity, membershipGroupReaderProvider: GroupReaderProvider): Boolean
```
Only `isHostedLocallyDest` would do a lookup in the group reader provider. The advantage of this is if the lookup fails then we can decide what that means about if an identity is locally hosted or not inside the `LinkManagerHostingMap` instead of passing in null which I find a bit confusing.

WDYT? I think if you did that you wouldn't need to change this file (part from using the two methods). I think the change to hear this file are Okay (I spent a while thinking about it) but the code is here quite subtle. The way here has the advantage of fewer lookups.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/staticnetwork/StaticMemberRegistrationService.kt
```diff
@@ -547,15 +547,15 @@ class StaticMemberRegistrationService(
             memberId.value,
             listOf(DUMMY_CERTIFICATE),
             HostedIdentitySessionKeyAndCert(
-                DUMMY_PUBLIC_SESSION_KEY,
+                keyEncodingService.encodeAsString(registeringMember.sessionInitiationKeys.first()),
```
williamvigorr3, line 550: Do we only support one session key in a static network?

williamvigorr3 (commented):
I left some comments.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/hosting/LinkManagerHostingMapImpl.kt
dimosr: Do we have any legitimate scenarios at the moment, where a destination might have an entry in the locally hosted list, but not one in the members list and we should be able to communicate with them?

If not (which I think might be the case), we could return false here. And in fact, in that case we could also make the `member` parameter non nullable which I think would make the interface more reasonable (e.g. providing the ability to perform a check for a null seems a bit confusing from an interface contract perspective).

That might also mean we don't need to worry about what you're discussing in https://github.com/corda/corda-runtime-os/pull/4430#discussion_r1290338146

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/staticnetwork/StaticMemberRegistrationService.kt
YashNabar: Yes, `StaticMemberRegistrationService` generates a single key pair per member

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/hosting/LinkManagerHostingMapImpl.kt
YashNabar: I was using the encoded format for comparison before, but switched to this as this seemed to be working okay. Can switch back if we think that's best

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/hosting/LinkManagerHostingMapImpl.kt
williamvigorr3: > Do we have any legitimate scenarios at the moment, where a destination might have an entry in the locally hosted list, but not one in the members list and we should be able to communicate with them?
> 
> If not (which I think might be the case), we could return false here. And in fact, in that case we could also make the `member` parameter non nullable which I think would make the interface more reasonable (e.g. providing the ability to perform a check for a null seems a bit confusing from an interface contract perspective).

I guess I just found it a bit confusing that it's asymmetric for source and dest. We can leave it as is from a functional perspective.

> That might also mean we don't need to worry about what you're discussing in [#4430 (comment)](https://github.com/corda/corda-runtime-os/pull/4430#discussion_r1290338146)

I guess it depends what the `OutboundMessageProcessor` looks like with added null checks if we move them out of here.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
dimosr: Between the suggestion here and the one [above](https://github.com/corda/corda-runtime-os/pull/4430#discussion_r1290174229), I personally think the one above might provide slightly better semantics as what kind of checks to perform based on whether something is a source or destination is currently a responsibility of the client (e.g. the outbound processor in this case, but assuming that other components use this they might potentially need to have different logic or they might not even have the concept of a source and destination).

So, clarifying what each `isHostedLocally` checks against with documentation and the reason for existence for two methods (and potentially with a slightly different name) seems like a cleaner interface.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/hosting/LinkManagerHostingMapImpl.kt
williamvigorr3: I think we should probably change it, to reduce risk of this breaking in the future.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
williamvigorr3: Maybe we could do [that](https://github.com/corda/corda-runtime-os/pull/4430/files#r1290174229) and add a comment to the `OutboundMessage` processor about why, we can't do the more thorough check for the source (it would be easy to change the code and break things). I think this is because when we (the source) first sends the unauthenticated message to the MGM, the source isn't in MembershipGroupReader (please correct me if I am wrong). Are there other cases?

This also means we don't need to couple the MembershipGroupReader to the LinkManagerHostingMap (which is a bit ugly).

YashNabar: Yes that was mainly for the unauthenticated message scenario when only the MGM (destination) is available in the reader, can't think of any other cases

williamvigorr3 (approved):
LGTM

YashNabar:
Multi-cluster pipeline with new test added in https://github.com/corda/corda-e2e-tests/pull/188: https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda-e2e-tests-multi-cluster-tests/detail/PR-188/4/pipeline

dimosr (approved):

PR #4432, opened 2023-08-08
CORE-13641 Try simplify test

kyriathar:
- moves common services & virtual node info to test class property level than having the same local in every test function
- simplifies test helper function to only do one thing (single responsibility principle)

### DIFF
--- a/components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
+++ b/components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
@@ -38,6 +38,7 @@ import net.corda.virtualnode.read.VirtualNodeInfoReadService
 import net.corda.virtualnode.toAvro
 import net.corda.virtualnode.toCorda
 import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.BeforeAll
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.TestInstance
@@ -117,6 +118,11 @@ class PersistenceExceptionTests {
         }
     }
 
+    @AfterEach
+    fun cleanUp() {
+        dbConnectionManager.stop()
+    }
+
     @Test
     fun `exception raised when cpks not present`() {
         val ignoredRequest = createEntityRequest()
@@ -146,8 +152,6 @@ class PersistenceExceptionTests {
         assertThat(response.error.errorType).isEqualTo(ExternalEventResponseErrorType.TRANSIENT)
         // The failure also captures the exception name.
         assertThat(response.error.exception.errorType).isEqualTo(CpkNotAvailableException::class.java.name)
-
-        dbConnectionManager.stop()
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
```diff
@@ -87,27 +93,33 @@ class PersistenceExceptionTests {
...
         lifecycle.accept(sandboxSetup) { setup ->
             virtualNode = setup.fetchService(timeout = 5000)
             cpiInfoReadService = setup.fetchService(timeout = 5000)
             cpkReadService = setup.fetchService(timeout = 5000)
             virtualNodeInfoReadService = setup.fetchService(timeout = 5000)
             responseFactory = setup.fetchService(timeout = 5000)
             currentSandboxGroupContext = setup.fetchService(timeout = 5000)
+
+            virtualNodeInfo = virtualNode.load(Resources.EXTENDABLE_CPB)
+            dbConnectionManager = FakeDbConnectionManager(
```
Sakpal, line 107: CURIOUS: @kyriathar do we need to close these connections once the tests are complete?

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
```diff
@@ -87,27 +93,33 @@ class PersistenceExceptionTests {
...
+                listOf(Pair(virtualNodeInfo.vaultDmlConnectionId, "animals-node")),
+                "PersistenceExceptionTests"
+            )
+            entitySandboxService =
```
Sakpal, line 111: CURIOUS: @kyriathar do we need to close delete these sandboxes once the tests are complete?

Sakpal (commented):
Looks good!

Can you please add description to the PR.

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
kyriathar: This is using in memory DB so I think we are fine. But even if we run against normal DB, it's not this test class responsibility to handle connections I think. That would be the services using the `DBConnectionManager` (`EntitySandboxService` in this case I believe?).

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
kyriathar: Not really sure "_close delete_" sandboxes can be done. This is an (OSGi) integration test, i.e. one-off process, so should be fine memory wise I believe .

Sakpal (dismissed):

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
driessamyn: I think it's still best practise to close the resources you create.
We shouldn't really make any assumption on the test execution lifecycle.

I would suggest closing all closable objects in the afterAll handler

kyriathar: Thanks, had another look. I believe the resource we creating in this test are the datasources through `FakeDbConnectionManager`, which we were already closing. Moved them to `AfterEach` handler to match previous behavior.

driessamyn (approved):
LGTM

PR #4433, opened 2023-08-08
CORE-13641 Adds tests exposing error on duplicate persistence request

kyriathar:
Adds tests around `EntityMessageProcessor` (it's responsible for processing Cordapp persistence requests) re-producing the following issues caused from retrying persistence requests (i.e. from processing duplicate persistence requests):

- duplicate `PersistEntities` request against table having PK (unique constraint) leading to PK violation.
- duplicate `PersistEntities` request against table not having unique constraint leading to duplicate entry(ies) in the DB.
- duplicate `MergeEntities` request leading to field updated statically being updated to the DB as well.

### DIFF
--- a/components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
+++ b/components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
@@ -298,7 +298,7 @@ class PersistenceExceptionTests {
     @Disabled("This test is disabled for now because currently we do execute duplicate persistence requests." +
             "It should be re-enabled after deduplication work is done in epic CORE-5909")
     @Test
-    fun `on duplicate persistence request don't execute it - statically updatable field isn't getting updated in DB`() {
+    fun `on duplicate persistence request don't execute it - statically updated field isn't getting updated in DB`() {
         createVersionedDogDb()
         val persistEntitiesRequest = createVersionedDogPersistRequest()
 
--- a/components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
+++ b/components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
@@ -41,6 +41,7 @@ import net.corda.virtualnode.toAvro
 import net.corda.virtualnode.toCorda
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertNull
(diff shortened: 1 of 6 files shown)


components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
```diff
@@ -80,6 +82,10 @@ class PersistenceExceptionTests {
         const val VERSIONED_DOGS_TABLE = "versioned-dogs.xml"
     }
 
+    @Suppress("JUnitMalformedDeclaration")
```
driessamyn, line 85: What's this suppression for?

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
```diff
@@ -80,6 +82,10 @@ class PersistenceExceptionTests {
...
+    @RegisterExtension
+    private val sandboxLifecycle = AllTestsLifecycle()
```
driessamyn, line 87: What does this do?

driessamyn (commented):
Generally looks ok, but isn't there a way for us to test this outside an OSGi test, using a regular DB integration test?

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
kyriathar: So this one seems to be the equivalent of `EachTestLifecycle` but runs once at `BeforeAll` and shuts down after all tests. 

It seems we have been re-fetching before each test those services (from `SandboxSetup`). That is fine, but then we would keep loading throughout the tests vnode info and cpk file hashes for the same CPB, so perhaps we could just keep these same throughout the test and initialize them at the beginning.

And then moved to `BeforeEach` services that need resetting and services that depend on those.

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
kyriathar: The compiler then complains that it should be defined at class level (since is is meant to be a before all).  However in our case `BeforeAll` is not a static function but a method, so `AllTestsLifecycle` could also be instance property in this case (just saw we would use it elsewhere too and took it from there).

driessamyn: Thanks for explaining. It probably suggests we're doing something a bit wrong generally, but that's out of scope of this change.

driessamyn (approved):
LGTM

PR #4434, opened 2023-08-08
CORE-12004 Adjust Member Onboarding Command

r3-arish:
## Summary
This PR modifies the onboard-member command in the Dynamic subcommand of the network command in the Corda CLI. The changes include updates to command parameters and options to align with the design requirements.

## Results
Screenshot 2023-08-08 at 16 49 48
Screenshot 2023-08-08 at 16 49 26

### DIFF
--- a/tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
+++ b/tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
@@ -9,6 +9,7 @@ import picocli.CommandLine.Option
 import java.io.File
 import java.security.MessageDigest
 import java.util.UUID
+import net.corda.cli.plugins.network.enums.MemberRole
 
 @Command(
     name = "onboard-member",
@@ -29,16 +30,16 @@ class OnBoardMember : Runnable, BaseOnboard() {
 
     @Option(
         names = ["--role", "-r"],
-        description = ["Member role, if any. Allowed values: 'notary'."]
+        description = ["Member role, if any. It is not mandatory to provide a role"],
     )
-    var role: String? = null
+    var role: MemberRole? = null
 
     @Option(
         names = ["--set", "-s"],
         description = ["Pass a custom key-value pair to the command to be included in the registration context. " +
-                "Specified as <key>=<value>. Multiple --set arguments may be provided."]
+                "Specified as <key>=<value>. Multiple --set arguments may be provided."],
     )
-    var customProperties: List<String> = emptyList()
+    var customProperties: Map<String, String> = emptyMap()
 
     @Option(
         names = ["--group-policy-file"],
@@ -52,7 +53,7 @@ class OnBoardMember : Runnable, BaseOnboard() {
 
     @Option(
         names = ["--cpi-hash"],
-        description = ["The CPI hash (Use either cpi-file, cpb-file or cpi-hash)."]
+        description = ["The CPI hash (Use either cpb-file or cpi-hash)."]
     )
     var cpiHash: String? = null
 
@@ -194,31 +195,41 @@ class OnBoardMember : Runnable, BaseOnboard() {
 
     override val registrationContext by lazy {
         val preAuth = preAuthToken?.let { mapOf("corda.auth.token" to it) } ?: emptyMap()
-        val roleProperty = if (role == "notary") mapOf("corda.roles.0" to role) else emptyMap()
-        val notaryServiceName = customProperties.firstNotNullOfOrNull { property ->
-            val key = property.substringBefore("=")
-            val value = property.substringAfter("=")
-            if (key == "corda.notary.service.name") value else null
-        }
-        val notaryProperties = if (role == "notary") {
-            mapOf(
-                "corda.notary.service.name" to notaryServiceName,
-                "corda.notary.service.flow.protocol.name" to "com.r3.corda.notary.plugin.nonvalidating",
-                "corda.notary.service.flow.protocol.version.0" to "1",
-                "corda.notary.keys.0.id" to notaryKeyId,
-                "corda.notary.keys.0.signature.spec" to "SHA256withECDSA"
-            )
-        } else {
-            emptyMap()
-        }
-
-        val endpoints: MutableMap<String, String> = mutableMapOf()
-
-        p2pGatewayUrls
-            .forEachIndexed { index, url ->
-                endpoints["corda.endpoints.$index.connectionURL"] = url
-                endpoints["corda.endpoints.$index.protocolVersion"] = "1"
+        val roleProperty = if (role != null) {
+                if (customProperties.containsKey("corda.notary.service.name")) {
+                    mapOf("corda.roles.0" to role!!.value)
+                } else {
+                    throw IllegalArgumentException("Cannot specify 'corda.notary.service.name'")
+                }
+            } else {
+                emptyMap()
             }
+
+        val notaryServiceName = customProperties["corda.notary.service.name"]
+        val notaryProperties = if (role != null) {
+                if (notaryServiceName == null) {
+                    throw IllegalArgumentException("Cannot specify 'corda.notary.service.name' when role is a value provided")
+                }
+                mapOf(
+                    "corda.notary.service.name" to notaryServiceName,
+                    "corda.notary.service.flow.protocol.name" to "com.r3.corda.notary.plugin.nonvalidating",
+                    "corda.notary.service.flow.protocol.version.0" to "1",
+                    "corda.notary.keys.0.id" to notaryKeyId,
+                    "corda.notary.keys.0.signature.spec" to "SHA256withECDSA"
+                )
+            } else {
+                emptyMap()
+            }
+
+        val endpoints: Map<String, String> = p2pGatewayUrls
+            .flatMapIndexed { index, url ->
+                    listOf(
+                        "corda.endpoints.$index.connectionURL" to url,
+                        "corda.endpoints.$index.protocolVersion" to "1"
+                    )
+            }
+            .toMap()
+
 
         val sessionKeys = mapOf(
             "corda.session.keys.0.id" to sessionKeyId,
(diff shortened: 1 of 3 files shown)


tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -11,30 +11,34 @@ import java.security.MessageDigest
...
-        names = ["--cpb-file"],
-        description = [
-            "Location of a CPB file (Use either cpi-file, cpb-file or cpi-hash).",
-            "Usage of --cpi-file or --cpi-hash is recommended. The CPI will be signed with default options."
-        ]
+        names = ["--role", "-r"],
+        description = ["Member role, if any. Allowed values: 'notary'."]
     )
-    var cpbFile: File? = null
+    var role: String? = null
```
yift-r3: Should this be an enum?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -11,30 +11,34 @@ import java.security.MessageDigest
...
+    var cpbFile: File? = null
 
     @Option(
-        names = ["--cpb-file"],
-        description = [
-            "Location of a CPB file (Use either cpi-file, cpb-file or cpi-hash).",
-            "Usage of --cpi-file or --cpi-hash is recommended. The CPI will be signed with default options."
-        ]
+        names = ["--role", "-r"],
+        description = ["Member role, if any. Allowed values: 'notary'."]
```
yift-r3: We should indicate that it's not a mandatory option.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -187,21 +188,48 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
+    private val notaryKeyId by lazy {
+        assignSoftHsmAndGenerateKey("NOTARY")
+    }
 
-        p2pGatewayUrls.mapIndexed { index, url ->
-            endpoints["corda.endpoints.$index.connectionURL"] = url
-            endpoints["corda.endpoints.$index.protocolVersion"] = "1"
+    override val registrationContext by lazy {
+        val preAuth = preAuthToken?.let { mapOf("corda.auth.token" to it) } ?: emptyMap()
+        val roleProperty = if (role == "notary") mapOf("corda.roles.0" to role) else emptyMap()
```
yift-r3: Maybe `if (role != null) `... (assuming we will have other roles in the future).

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -187,21 +188,48 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
+        val notaryServiceName = customProperties.firstNotNullOfOrNull { property ->
+            val key = property.substringBefore("=")
+            val value = property.substringAfter("=")
+            if (key == "corda.notary.service.name") value else null
```
yift-r3: If someone will use `--set 'corda.notary.service.name=hello=world'`?

We should validate that all the customProperties syntax is valid. I would use Map (see https://picocli.info/#_arrays_collections_maps)

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -187,21 +188,48 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
+        }
+        val notaryProperties = if (role == "notary") {
+            mapOf(
+                "corda.notary.service.name" to notaryServiceName,
```
yift-r3: If we are a notary role and the `notaryServiceName,` is null, should we throw an exception?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -187,21 +188,48 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
+                "corda.notary.keys.0.signature.spec" to "SHA256withECDSA"
+            )
+        } else {
+            emptyMap()
         }
 
-        mapOf(
+        val endpoints: MutableMap<String, String> = mutableMapOf()
+
+        p2pGatewayUrls
```
yift-r3: Instead of a mutable map, you can use `flatMap`.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: its just a single value ("notary") - having an enum for just one value wouldn't make sense

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: We can change it in the future if we have multiple roles - but for now we are just accepting "notary". Anything other than "notary" should not be accepted.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: It will still work, I have attached the screenshot above in which I tried the param `-s "corda.notary.service.name"="C=GB, L=London, O=Arish"`

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: We can do something like this `val notaryProperties = if (role == "notary" && notaryServiceName != null)`

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
yift-r3: If we add more values in the future, it makes sense to make it an enum now. This way, we can let picocli make the verification and the list of available values.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: What is the reasoning behind using the `flatMap` instead of `mutableMap`?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
yift-r3: That seems wrong.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: Thanks, I missed that.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
yift-r3: Picocli has better support for `key=value` collection than we can do. Why not use it?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
yift-r3: I don't think we support a notary without a service name.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
yift-r3: It's a good practice to avoid mutability.

yift-r3 (approved):

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -187,21 +189,55 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
     override val registrationContext by lazy {
-        val preAuth = if (preAuthToken != null) mapOf("corda.auth.token" to preAuthToken) else emptyMap()
-        val endpoints = mutableMapOf<String, String>()
+        val preAuth = preAuthToken?.let { mapOf("corda.auth.token" to it) } ?: emptyMap()
+        val notaryServiceName = customProperties["corda.notary.service.name"]
+        val roleProperty = if (role != null) {
+                if (notaryServiceName != null) {
+                    mapOf("corda.roles.0" to role!!.value)
+                } else {
+                    throw IllegalArgumentException("Cannot specify 'corda.notary.service.name'")
```
dimosr, line 203: Isn't this supposed to trigger when role = notary and notary service name is not specified? So, message should be instead something like:
```
throw IllegalArgumentException("When specifying a NOTARY role, you also need to specify a custom property for its name under 'corda.notary.service.name'.")
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -187,21 +189,55 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
+            "corda.session.keys.0.signature.spec" to "SHA256withECDSA"
+        )
+        val ledgerKeys = mapOf(
             "corda.ledger.keys.0.id" to ledgerKeyId,
-            "corda.ledger.keys.0.signature.spec" to "SHA256withECDSA",
-        ) + endpoints + preAuth
+            "corda.ledger.keys.0.signature.spec" to "SHA256withECDSA"
+        )
+
+        sessionKeys + ledgerKeys + endpoints + preAuth + roleProperty + notaryProperties
```
dimosr, line 240: Aren't `customProperties` supposed to also support population of extra custom metadata under the `ext.` prefix, which would be passed into the registration context as-is?

r3-arish: a follow up PR has been created to address the above mentioned comment: https://github.com/corda/corda-runtime-os/pull/4442/

PR #4436, opened 2023-08-09
CORE-13900 Throw more meaningful exception in toSignedTransaction() if the creator is not a signatory

vlajos:
Throw `CordaRuntimeException("Creator of the transaction does not own any signatory keys.")` instead of `TransactionNoAvailableKeysException` when TransactionBuilder's toSignedTransaction() gets called and the caller is not a signatory.


relyafi (approved):
LGTM

vlajos:
Closing in favour of https://github.com/corda/corda-runtime-os/pull/4440

PR #4442, opened 2023-08-10
CORE-12004 - Adjust Member Onboarding Plugin Command

r3-arish:
## Summary

This is a follow up PR that addresses the ability to add custom properties in the `registrationContext` for onboarding a member. 

## Results
Screenshot 2023-08-10 at 13 33 33
Screenshot 2023-08-10 at 13 33 50
Screenshot 2023-08-10 at 13 34 20
Screenshot 2023-08-10 at 13 42 06

## New Result
Screenshot 2023-08-11 at 13 45 25

### DIFF
--- a/tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
+++ b/tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
@@ -10,6 +10,19 @@ import java.io.File
 import java.security.MessageDigest
 import java.util.UUID
 import net.corda.cli.plugins.network.enums.MemberRole
+import net.corda.membership.lib.MemberInfoExtension.Companion.NOTARY_SERVICE_NAME
+import net.corda.membership.lib.MemberInfoExtension.Companion.NOTARY_SERVICE_PROTOCOL
+import net.corda.membership.lib.MemberInfoExtension.Companion.NOTARY_SERVICE_PROTOCOL_VERSIONS
+import net.corda.membership.lib.MemberInfoExtension.Companion.CUSTOM_KEY_PREFIX
+import net.corda.membership.lib.MemberInfoExtension.Companion.NOTARY_KEYS_ID
+import net.corda.membership.lib.MemberInfoExtension.Companion.NOTARY_KEY_SPEC
+import net.corda.membership.lib.MemberInfoExtension.Companion.ROLES_PREFIX
+import net.corda.membership.lib.MemberInfoExtension.Companion.URL_KEY
+import net.corda.membership.lib.MemberInfoExtension.Companion.PROTOCOL_VERSION
+import net.corda.membership.lib.MemberInfoExtension.Companion.PARTY_SESSION_KEYS_ID
+import net.corda.membership.lib.MemberInfoExtension.Companion.SESSION_KEYS_SIGNATURE_SPEC
+import net.corda.membership.lib.MemberInfoExtension.Companion.LEDGER_KEYS_ID
+import net.corda.membership.lib.MemberInfoExtension.Companion.LEDGER_KEY_SIGNATURE_SPEC
 
 @Command(
     name = "onboard-member",
@@ -32,7 +45,7 @@ class OnBoardMember : Runnable, BaseOnboard() {
         names = ["--role", "-r"],
         description = ["Member role, if any. It is not mandatory to provide a role"],
     )
-    var role: MemberRole? = null
+    var roles: Set<MemberRole> = emptySet()
 
     @Option(
         names = ["--set", "-s"],
@@ -195,79 +208,55 @@ class OnBoardMember : Runnable, BaseOnboard() {
 
     override val registrationContext by lazy {
         val preAuth = preAuthToken?.let { mapOf("corda.auth.token" to it) } ?: emptyMap()
-        
-        if (customProperties.size > 100) {
-            throw IllegalArgumentException("Cannot specify more than 100 key-value pairs")
-        }
-        
-        val notaryServiceName = customProperties["corda.notary.service.name"]
-      
-        customProperties = customProperties.filterKeys { it != "corda.notary.service.name" }
-        
+
         val extProperties = customProperties.filter {
             val key = it.key
             val value = it.value
-            val keyAndValueNotEmpty = (key.isNotEmpty()).and(value.isNotEmpty())
-            val keyStartsWithExt = key.startsWith("ext.")
-            val keyDoesNotEndWithExt = !key.endsWith("ext.")
-            val allChecks = keyAndValueNotEmpty.and(keyStartsWithExt).and(keyDoesNotEndWithExt)
+            val keyStartsWithExt = key.startsWith("$CUSTOM_KEY_PREFIX.")
 
-            if (!keyAndValueNotEmpty) {
-                throw IllegalArgumentException("Please specify a key or value, either cannot be empty")
-            }
-            if (!keyStartsWithExt) {
-                throw IllegalArgumentException("The key: $key has to start with `ext.` prefix")
-            }
-            if (!keyDoesNotEndWithExt) {
-                throw IllegalArgumentException("The key: $key cannot end with `ext.`")
-            }
             if (key.length > 128) {
                 throw IllegalArgumentException("The key length cannot exceed 128 characters")
             }
             if (value.length > 800) {
                 throw IllegalArgumentException("The value length cannot exceed 800 characters")
             }
-
-            allChecks
+            keyStartsWithExt
         }
-        val roleProperty = if (role != null) {
-                if (notaryServiceName != null) {
-                    mapOf("corda.roles.0" to role!!.value)
-                } else {
-                    throw IllegalArgumentException("Cannot specify 'corda.notary.service.name'")
-                }
-            } else {
-                emptyMap()
-            }
+        val roleProperty: Map<String, String> = roles.mapIndexed { index: Int, memberRole: MemberRole ->
+            "$ROLES_PREFIX.$index" to memberRole.value
+        }.toMap()
 
-        val notaryProperties = if (role != null) {
-                mapOf(
-                    "corda.notary.service.name" to notaryServiceName,
-                    "corda.notary.service.flow.protocol.name" to "com.r3.corda.notary.plugin.nonvalidating",
-                    "corda.notary.service.flow.protocol.version.0" to "1",
-                    "corda.notary.keys.0.id" to notaryKeyId,
-                    "corda.notary.keys.0.signature.spec" to "SHA256withECDSA"
-                )
-            } else {
+        val notaryProperties = if (roles.contains(MemberRole.NOTARY)) {
+            val notaryServiceName = customProperties[NOTARY_SERVICE_NAME] ?:
+            throw IllegalArgumentException("When specifying a NOTARY role, " +
+                    "you also need to specify a custom property for its name under $NOTARY_SERVICE_NAME.")
+            mapOf(
+                NOTARY_SERVICE_NAME to notaryServiceName,
+                NOTARY_SERVICE_PROTOCOL to "com.r3.corda.notary.plugin.nonvalidating",
+                NOTARY_SERVICE_PROTOCOL_VERSIONS.format("0") to "1",
+                NOTARY_KEYS_ID.format("0") to notaryKeyId,
+                NOTARY_KEY_SPEC.format("0") to "SHA256withECDSA"
+            )
+        } else {
             emptyMap()
-            }
+        }
 
         val endpoints: Map<String, String> = p2pGatewayUrls
             .flatMapIndexed { index, url ->
                     listOf(
-                        "corda.endpoints.$index.connectionURL" to url,
-                        "corda.endpoints.$index.protocolVersion" to "1"
+                        URL_KEY.format(index) to url,
+                        PROTOCOL_VERSION.format(index) to "1"
                     )
             }
             .toMap()
 
         val sessionKeys = mapOf(
-            "corda.session.keys.0.id" to sessionKeyId,
-            "corda.session.keys.0.signature.spec" to "SHA256withECDSA"
+            PARTY_SESSION_KEYS_ID.format(0) to sessionKeyId,
+            SESSION_KEYS_SIGNATURE_SPEC.format(0) to "SHA256withECDSA"
         )
         val ledgerKeys = mapOf(
-            "corda.ledger.keys.0.id" to ledgerKeyId,
-            "corda.ledger.keys.0.signature.spec" to "SHA256withECDSA"
+            LEDGER_KEYS_ID.format(0) to ledgerKeyId,
+            LEDGER_KEY_SIGNATURE_SPEC.format(0) to "SHA256withECDSA"
         )
 
         sessionKeys + ledgerKeys + endpoints + preAuth + roleProperty + notaryProperties + extProperties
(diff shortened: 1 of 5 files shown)


tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -195,7 +195,41 @@ class OnBoardMember : Runnable, BaseOnboard() {
 
     override val registrationContext by lazy {
         val preAuth = preAuthToken?.let { mapOf("corda.auth.token" to it) } ?: emptyMap()
+        
+        if (customProperties.size > 100) {
+            throw IllegalArgumentException("Cannot specify more than 100 key-value pairs")
+        }
```
YashNabar: I think it's fine to leave this out, and leave that validation to the platform

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -195,7 +195,41 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
+        
         val notaryServiceName = customProperties["corda.notary.service.name"]
+      
+        customProperties = customProperties.filterKeys { it != "corda.notary.service.name" }
+        
+        val extProperties = customProperties.filter {
```
YashNabar: I think something like this should be sufficient here -
```suggestion
        val extProperties = customProperties.filterKeys { it.startsWith("$CUSTOM_KEY_PREFIX.") }
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -195,7 +195,41 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
+
+            if (!keyAndValueNotEmpty) {
+                throw IllegalArgumentException("Please specify a key or value, either cannot be empty")
+            }
+            if (!keyStartsWithExt) {
+                throw IllegalArgumentException("The key: $key has to start with `ext.` prefix")
+            }
+            if (!keyDoesNotEndWithExt) {
+                throw IllegalArgumentException("The key: $key cannot end with `ext.`")
+            }
```
YashNabar: I don't think this is something we want to enforce

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -195,7 +195,41 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
+            val key = it.key
+            val value = it.value
+            val keyAndValueNotEmpty = (key.isNotEmpty()).and(value.isNotEmpty())
+            val keyStartsWithExt = key.startsWith("ext.")
+            val keyDoesNotEndWithExt = !key.endsWith("ext.")
+            val allChecks = keyAndValueNotEmpty.and(keyStartsWithExt).and(keyDoesNotEndWithExt)
+
+            if (!keyAndValueNotEmpty) {
+                throw IllegalArgumentException("Please specify a key or value, either cannot be empty")
+            }
```
YashNabar: We can remove this and the other conditions below - as long as it's a Map<String, String> (which I believe picocli will ensure it is) we can simply take the properties which have the prefix "ext." and add them to the context

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -195,7 +195,41 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
+            if (key.length > 128) {
+                throw IllegalArgumentException("The key length cannot exceed 128 characters")
+            }
+            if (value.length > 800) {
+                throw IllegalArgumentException("The value length cannot exceed 800 characters")
+            }
+
+            allChecks
+        }
         val roleProperty = if (role != null) {
```
YashNabar: We can have multiple roles, so we should change the type of this option to a Collection and change here to generate roleProperties similar to endpoints

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -195,7 +195,41 @@ class OnBoardMember : Runnable, BaseOnboard() {
 
     override val registrationContext by lazy {
         val preAuth = preAuthToken?.let { mapOf("corda.auth.token" to it) } ?: emptyMap()
+        
+        if (customProperties.size > 100) {
+            throw IllegalArgumentException("Cannot specify more than 100 key-value pairs")
+        }
+        
         val notaryServiceName = customProperties["corda.notary.service.name"]
```
YashNabar: We can reuse the constants from `MemberInfoExtension` in place of such keys. Also in other places

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: I followed this from here -> Note: Custom properties have a character limit of 128 for keys and 800 for values. A maximum of 100 such key-value pairs may be defined in the registration context. https://github.com/corda/corda-runtime-os/wiki/Member-Onboarding-(Dynamic-Networks)#build-registration-context

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: I am unsure about this point, as in users are allowed to specify multiple roles? what if someone enters multiple notary roles, do I need handle that scenario as well?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: I wanted to make sure the keys are not just `ext.`

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: Okay, got your point

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: I am not sure if picocli will handle empty values - what if the user specifies the key and the value is empty. Should we still send a pair with empty value?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: Should I use a companion object for specifying the constants in the same class? Or should I just create a separate constants class for this, the question is that I am not sure if we are going to be reusing these values elsewhere in the plugin classes.

YashNabar: You can reuse the ones from `MemberInfoExtension`, they are already defined there

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
YashNabar: Should be fine to pass empty strings, we don't have that validation at the platform level either

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -32,7 +45,7 @@ class OnBoardMember : Runnable, BaseOnboard() {
         names = ["--role", "-r"],
         description = ["Member role, if any. It is not mandatory to provide a role"],
     )
-    var role: MemberRole? = null
+    var roles: Set<MemberRole> = emptySet()
```
YashNabar, line 49: We should probably update the description to indicate that multiple roles may be specified

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -195,79 +208,55 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
-        
+
         val extProperties = customProperties.filter {
             val key = it.key
             val value = it.value
-            val keyAndValueNotEmpty = (key.isNotEmpty()).and(value.isNotEmpty())
-            val keyStartsWithExt = key.startsWith("ext.")
-            val keyDoesNotEndWithExt = !key.endsWith("ext.")
-            val allChecks = keyAndValueNotEmpty.and(keyStartsWithExt).and(keyDoesNotEndWithExt)
+            val keyStartsWithExt = key.startsWith("$CUSTOM_KEY_PREFIX.")
```
YashNabar: Curious what others think about these checks here. But I feel if we do want to keep the key value length checks we should also keep the check on number of fields. And in that case, applying them like this would probably be more readable -
```suggestion
val extProperties = customProperties.filterKeys { it.startsWith("$CUSTOM_KEY_PREFIX.") }.apply {
            if (size > 100) {
                throw IllegalArgumentException("The number of custom fields ($size) provided is larger than the maximum allowed ($MAX_CUSTOM_FIELDS).")
            }
            val keysTooLong = filterKeys { it.length > MAX_KEY_LENGTH }
            if (keysTooLong.isNotEmpty()) {
                throw IllegalArgumentException("The following keys exceed the maximum allowed length ($MAX_KEY_LENGTH): ${keysTooLong.keys}")
            }
            val valuesTooLong = filterValues { it.length > MAX_VALUE_LENGTH }
            if (valuesTooLong.isNotEmpty()) {
                throw IllegalArgumentException("Values of the following keys exceed the maximum allowed length ($MAX_VALUE_LENGTH): ${keysTooLong.keys}")
            }
        }
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -195,49 +208,58 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
+            val key = it.key
+            val value = it.value
+            val keyStartsWithExt = key.startsWith("$CUSTOM_KEY_PREFIX.")
+
+            if (key.length > 128) {
+                throw IllegalArgumentException("The key length cannot exceed 128 characters")
+            }
+            if (value.length > 800) {
+                throw IllegalArgumentException("The value length cannot exceed 800 characters")
             }
```
dimosr: What is the end-user experience in these scenarios? Are we going to end up with a stacktrace shown in the console? If so, we shouldn't be doing this. Any expected errors should lead to an error message shown in the console.

Stack traces are typically an indication of a bug.

dimosr: Btw, I think I've already raised the same issue with @r3-arish about our current `checkInvariant` checks. We should fix that as well (if we haven't raised a ticket, let's raise one now and prioritise it against any follow-up work).

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
dimosr: No strong opinion here, but if these are enforced by the platform already I don't think it's a big issue not enforcing them at the plugin level as well. They will lead to an invalid/declined registration anyway, which might require the user a few more steps to identify what went wrong but that's a minor issue given it's not a very likely scenario for a user. 

On the other hand, they lead to quite a bit of duplication and potential of drift against the rules in the platform that can change. This can be eliminated by having shared code between the plugin & the platform, but I think at this point we are probably starting to over-engineer things.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -195,49 +208,58 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
+            keyStartsWithExt
+        }
+        val roleProperty: Map<String, String> = roles.mapIndexed { index: Int, memberRole: MemberRole ->
+            "$ROLES_PREFIX.$index" to memberRole.value
+        }.toMap()
+
+        val notaryProperties = if (roles.contains(MemberRole.NOTARY)) {
+            val notaryServiceName = customProperties[NOTARY_SERVICE_NAME] ?:
+            throw IllegalArgumentException("When specifying a NOTARY role, " +
```
dimosr: same comment here on throwing an exception for invalid input.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
YashNabar: We are using `verifyAndPrintError` in some commands, which catches exceptions and prints the error message to console. We should use it here (and in other places where appropriate)

YashNabar: @dimosr I've raised a ticket [here](https://r3-cev.atlassian.net/browse/CORE-16269)

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -195,49 +209,49 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
-                emptyMap()
+        val roleProperty: Map<String, String> = roles.mapIndexed { index: Int, memberRole: MemberRole ->
+            "$ROLES_PREFIX.$index" to memberRole.value
+        }.toMap()
+
+        val extProperties = customProperties.filterKeys { it.startsWith("$CUSTOM_KEY_PREFIX.") }
+
+        val notaryProperties = if (roles.contains(MemberRole.NOTARY)) {
+            var notaryServiceName = ""
+            verifyAndPrintError {
```
YashNabar: We can probably just do this once in the `run()` method for the command

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
YashNabar: Discussed offline, addressed in [a799827](https://github.com/corda/corda-runtime-os/pull/4442/commits/a799827ecd09c6f8ef8b7d7370a0d7efb752246c)

YashNabar (approved):
LGTM - probably worth doing another round of tests after the recent changes

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
r3-arish: Screenshot 2023-08-14 at 14 12 59

dimosr: The fact that we show the exception here along with the message is a bit sub-optimal. We could probably fix that by introducing some explicit validation of the target URL at some early point.

dimosr (approved):
LGTM too - left a comment [here](https://github.com/corda/corda-runtime-os/pull/4442#discussion_r1293497776), but it's quite minor and not worth holding the PR for this.

PR #4443, opened 2023-08-11
ES-1123: Build the project using k8s pod rather than full EC2 instance

ronanbrowne:
Migrate to use cordaPipelineKubernetesAgent.groovy

- Functionally identical except for the fact we now build in a k8s pod, thus using less cloud resource as each build no longer needs a full EC2 instance
- No application-level change as part of this PR , this only effects infra this project is built on


yarivus (approved):
LGTM

PR #4444, opened 2023-08-11
CORE-16216 - Member is able to re-register when MGM is on 5.0 platform

nikinagy:
When MGM is on 5.0 and member is on 5.1 platform, the member can successfully re-register. Since re-registration is only available starting from 5.1, this scenario should be avoided and the re-registration attempt should be INVALID.

Testing:

1) Register MGM on 5.0 platform
2) Register Alice on 5.1 platform

Alice's view of initial registration request:
alice-registers
MGM's view of initial registration request:
mgms-view-of-alices-request

3) Alice tries to re-register

Alice sees request as invalid:
alice-re-registers
MGM won't receive the request:
mgm-wont-see-request

4) Alice tries to pretend initial registration by manually setting the serial

Alice sees request as invalid:
pretend-initial-registration

### DIFF
--- a/components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
+++ b/components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
@@ -108,7 +108,6 @@ import net.corda.v5.base.exceptions.CordaRuntimeException
 import net.corda.v5.base.types.MemberX500Name
 import net.corda.v5.base.versioning.Version
 import net.corda.v5.crypto.SignatureSpec
-import net.corda.v5.membership.MemberInfo
 import net.corda.virtualnode.HoldingIdentity
 import net.corda.virtualnode.read.VirtualNodeInfoReadService
 import net.corda.virtualnode.toAvro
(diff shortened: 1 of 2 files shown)


components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
```diff
@@ -466,6 +468,23 @@ class DynamicMemberRegistrationService @Activate constructor(
...
+         * we will mark their request as INVALID.
+         */
+        @Suppress("ComplexCondition")
+        private fun verifyReRegistrationIsEnabled(
+            submittedSerial: Long,
+            currentSerial: Long?,
+            mgmPlatformVersion: Int,
+        ) {
+            if ((submittedSerial > 0 || (currentSerial != null && currentSerial > 0)) && mgmPlatformVersion < 50100) {
+                throw InvalidMembershipRegistrationException("MGM is on 5.0 platform. " +
```
yift-r3: Should we be less explicit about the version number? Maybe something like `Re-registration is not supported with the current MGM version`? We might have 5.0.1...

dimosr: Keeping the message generic is a good idea, we could include the platform version (or maybe better the software version) of the mgm inside a parenthesis in the message, if we want to make it a bit more explanatory for the user.

nikinagy: addressed in [0b6113e](https://github.com/corda/corda-runtime-os/pull/4444/commits/0b6113e5dc2baaa31ff0be8cd374ba12962c1107)

yift-r3 (approved):

PR #4451, opened 2023-08-13
CORE-13893: Register sandbox eviction listener for StateAndRefCache.

chrisr3:
Register `onEviction` as a `FLOW` sandbox eviction listener, as otherwise it will never be invoked.


vlajos (approved):
LGTM

PR #4453, opened 2023-08-14
CORE-15179 - Fixed issue with the `getBalanceQuery` and the check for valid and invalid claims.

filipesoliveira:


chrisr3:
I see that the "tag" queries say:
```sql
"AND   token_tag ~ :$SQL_PARAMETER_TAG_FILTER"
```
What is `~`? I don't think that's ANSI SQL :thinking:. Is that a Postgres-specific operator for matching regular expressions? Because Corda isn't supposed to be Postgres-specific.

josephzunigadaly (dismissed):

relyafi:
> I see that the "tag" queries say:
> 
> ```sql
> "AND   token_tag ~ :$SQL_PARAMETER_TAG_FILTER"
> ```
> 
> What is `~`? I don't think that's ANSI SQL 🤔. Is that a Postgres-specific operator for matching regular expressions? Because Corda isn't supposed to be Postgres-specific.

I see this is in code that has already been merged (albeit very recently), but I agree with @chrisr3 - we should try and avoid any DB specific SQL wherever possible, if we can update this to be DB-agnostic that will save us a lot of pain, as otherwise this logic will need to be updated to have DB specific implementations in line with the other work we are doing for multi database support.

filipesoliveira:
@relyafi, the token logic must allow people to select multiple tokens with different tags. That operations has to be done at the DB level because it won't be practical to load all the tokens and do the filtering in-memory.

chrisr3:
> @relyafi, the token logic must allow people to select multiple tokens with different tags. That operations has to be done at the DB level because it won't be practical to load all the tokens and do the filtering in-memory.

But how will this work on other databases? Is there an ANSI solution to your problem?

One idea might for the token to be a "glob" that you can transform into something compatible with SQL's `LIKE`.

relyafi:
> @relyafi, the token logic must allow people to select multiple tokens with different tags. That operations has to be done at the DB level because it won't be practical to load all the tokens and do the filtering in-memory.

@filipesoliveira that's fine, but can it be done in a way which uses ANSI SQL? If there is no alternative to what's been done, we can deal with it, but it will introduce extra complexity as we will shortly end up having multiple implementations to maintain.

filipesoliveira:
There is no REGEX concept in ANSI SQL, unfortunately. From what I have found that feature is DB specific.

chrisr3:
> There is no REGEX concept in ANSI SQL, unfortunately. From what I have found that feature is DB specific.

Then don't use regular expressions? Regexs are _really_ complicated anyway, especially if all people really want is to match "tag*". I suspect [`LIKE`](https://www.w3schools.com/sql/sql_like.asp) is functional enough.

filipesoliveira:
I understand the concerns. But I cannot make that the decision because I am not fully aware of the use-cases. In the previous, implementation we were using regular expressions so I think it is best not to change the behaviour at least until @owenstanford is back.

chrisr3:
> so I think it is best not to change the behaviour at least until @owenstanford is back.

How about asking @blsemo?

---
BTW I have tried replacing
```sql
token_tag ~ :$SQL_PARAMETER_TAG_FILTER
```
with
```sql
REGEXP_MATCHES(token_tag, :$SQL_PARAMETER_TAG_FILTER)
```
for HSQLDB, and then executing my "driver" version of `TokenBalanceQueryTests`, but
```
ensure balance is equal to 2 but the balance including claimed tokens is 7
```
still fails with:
```
org.opentest4j.AssertionFailedError: 
expected: 7
 but was: 10
```
Are you sure that the E2E tests are fixed?

filipesoliveira:
It might be best to have @owenstanford involved, too.

chrisr3:
> It might best to have @owenstanford involved, too.

I'm not suggesting that @owenstanford and @blsemo have a cage fight. I'm sure this feature is understood by more than one person.

chrisr3:
According to [this](https://www.postgresql.org/docs/current/functions-matching.html), Postgres supports a function `regexp_like` that is equivalent to this `~` operator. Specifically, section 9.7.3.8:
> regexp_like(string, pattern) or string ~ pattern

Perhaps _at least_ use the `regexp_like` function here instead of the `~` operator? We can potentially teach other databases to map a function to something compatible by tweaking the Hibernate `Dialect` at runtime, but I don't know how to do that for operators.

Note that this _still_ assumes that other databases will support an `isMatching(string, regex)` type of function to map!

filipesoliveira:
> > It might best to have @owenstanford involved, too.
> 
> I'm not suggesting that @owenstanford and @blsemo have a cage fight. I'm sure this feature is understood by more than one person.

The suggestions being made are more than welcome. I will create a ticket so we capture this in the epic.

Ticket: https://r3-cev.atlassian.net/browse/CORE-15183

chrisr3 (changes requested):
Can you replace the `~` operator with Postgres's equivalent `REGEXP_LIKE` function please? Specifically:
```sql
REGEXP_LIKE(string, pattern)
```
This will allow us a least to _try_ to manage the database-specificity in the Hibernate `Dialect`. However, I'm not convinced that the E2E tests pass yet either.

chrisr3:
In `TokenCacheEventProcessor`:
```kotlin
val validClaims =
    nonNullableState.tokenClaims.filter { it.claimedTokenStateRefs == null || it.claimedTokenStateRefs.isEmpty() }
val invalidClaims =
    nonNullableState.tokenClaims.filterNot { it.claimedTokenStateRefs != null && !it.claimedTokenStateRefs.isEmpty() }
```
If I'm reading these lines correctly then the contents of `validClaims` and `invalidClaims` will be _exactly the same_, because you have implemented a double negative. Perhaps this instead?
```kotlin
val validClaims =
    nonNullableState.tokenClaims.filter { it.claimedTokenStateRefs.isNullOrEmpty() }
val invalidClaims = nonNullableState.tokenClaims - validClaims.toSet()
```
This removes a spurious warning about invalid claims being discarded.

filipesoliveira:
> In `TokenCacheEventProcessor`:
> 
> ```kotlin
> val validClaims =
>     nonNullableState.tokenClaims.filter { it.claimedTokenStateRefs == null || it.claimedTokenStateRefs.isEmpty() }
> val invalidClaims =
>     nonNullableState.tokenClaims.filterNot { it.claimedTokenStateRefs != null && !it.claimedTokenStateRefs.isEmpty() }
> ```
> 
> If I'm reading these lines correctly then the contents of `validClaims` and `invalidClaims` will be _exactly the same_, because you have implemented a double negative. Perhaps this instead?
> 
> ```kotlin
> val validClaims =
>     nonNullableState.tokenClaims.filter { it.claimedTokenStateRefs.isNullOrEmpty() }
> val invalidClaims = nonNullableState.tokenClaims - validClaims.toSet()
> ```
> 
> This removes a spurious warning about invalid claims being discarded.

Well spotted. I can see it now. It is easy to prove that the code is wrong by applying De Morgan's law to the boolean expression. I will fix that  too.

chrisr3:
You might also want to consider:
```diff
diff --git a/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/AvailableTokenServiceImpl.kt b/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/AvailableTokenServiceImpl.kt
index 22ed77fe5e..88ffd28fdc 100644
--- a/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/AvailableTokenServiceImpl.kt
+++ b/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/AvailableTokenServiceImpl.kt
@@ -6,16 +6,16 @@ import org.osgi.service.component.annotations.Activate
 import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
 import net.corda.db.connection.manager.DbConnectionManager
+import net.corda.db.schema.CordaDb
 import net.corda.flow.external.events.responses.exceptions.VirtualNodeException
 import net.corda.ledger.utxo.token.cache.entities.AvailTokenQueryResult
 import net.corda.ledger.utxo.token.cache.entities.CachedToken
 import net.corda.ledger.utxo.token.cache.repositories.UtxoTokenRepository
-import net.corda.orm.JpaEntitiesSet
 import net.corda.ledger.utxo.token.cache.entities.TokenBalance
 import net.corda.ledger.utxo.token.cache.entities.TokenPoolKey
+import net.corda.orm.JpaEntitiesRegistry
 import net.corda.virtualnode.read.VirtualNodeInfoReadService
 import net.corda.virtualnode.VirtualNodeInfo
-import org.slf4j.LoggerFactory
 
 @Component(
     service = [AvailableTokenService::class]
@@ -26,17 +26,14 @@ class AvailableTokenServiceImpl @Activate constructor(
     @Reference
     private val dbConnectionManager: DbConnectionManager,
     @Reference
+    private val jpaEntitiesRegistry: JpaEntitiesRegistry,
+    @Reference
     private val utxoTokenRepository: UtxoTokenRepository
 ) : AvailableTokenService, SingletonSerializeAsToken {
-
-    companion object {
-        private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
-    }
-
     override fun findAvailTokens(poolKey: TokenPoolKey, ownerHash: String?, tagRegex: String?): AvailTokenQueryResult {
         val virtualNode = getVirtualNodeInfo(poolKey)
 
-        val entityManagerFactory = createEntityManagerFactory(virtualNode)
+        val entityManagerFactory = getOrCreateEntityManagerFactory(virtualNode)
 
         return utxoTokenRepository.findTokens(entityManagerFactory.createEntityManager(), poolKey, ownerHash, tagRegex)
     }
@@ -48,7 +45,7 @@ class AvailableTokenServiceImpl @Activate constructor(
         claimedTokens: Collection<CachedToken>
     ): TokenBalance {
         val virtualNode = getVirtualNodeInfo(poolKey)
-        val entityManagerFactory = createEntityManagerFactory(virtualNode)
+        val entityManagerFactory = getOrCreateEntityManagerFactory(virtualNode)
 
         val totalBalance = utxoTokenRepository.queryBalance(entityManagerFactory.createEntityManager(), poolKey, ownerHash, tagRegex)
         val claimedBalance = claimedTokens.sumOf { it.amount }
@@ -57,10 +54,13 @@ class AvailableTokenServiceImpl @Activate constructor(
         return TokenBalance(availableBalance, totalBalance)
     }
 
-    private fun createEntityManagerFactory(virtualNode: VirtualNodeInfo) =
-        dbConnectionManager.createEntityManagerFactory(
+    private fun getOrCreateEntityManagerFactory(virtualNode: VirtualNodeInfo) =
+        dbConnectionManager.getOrCreateEntityManagerFactory(
             virtualNode.vaultDmlConnectionId,
-            JpaEntitiesSet.create("emptySet", emptySet()) // No entity required. The mapping will be manually done
+            jpaEntitiesRegistry.get(CordaDb.Vault.persistenceUnitName)
+                ?: throw IllegalStateException(
+                    "persistenceUnitName ${CordaDb.Vault.persistenceUnitName} is not registered."
+                )
         )
 
     private fun getVirtualNodeInfo(poolKey: TokenPoolKey) =
```
because creating a new `EntityManagerFactory` each time could be expensive and is definitely unnecessary. And you weren't destroying it after use either, which suggests a resource leak. Obviously, you won't _want_ to destroy it if you now fetch it from a pool instead :wink:.

filipesoliveira:
> 

This will be done in a different PR.

filipesoliveira:
> You might also want to consider:
> 
> ```diff
> diff --git a/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/AvailableTokenServiceImpl.kt b/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/AvailableTokenServiceImpl.kt
> index 22ed77fe5e..88ffd28fdc 100644
> --- a/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/AvailableTokenServiceImpl.kt
> +++ b/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/AvailableTokenServiceImpl.kt
> @@ -6,16 +6,16 @@ import org.osgi.service.component.annotations.Activate
>  import org.osgi.service.component.annotations.Component
>  import org.osgi.service.component.annotations.Reference
>  import net.corda.db.connection.manager.DbConnectionManager
> +import net.corda.db.schema.CordaDb
>  import net.corda.flow.external.events.responses.exceptions.VirtualNodeException
>  import net.corda.ledger.utxo.token.cache.entities.AvailTokenQueryResult
>  import net.corda.ledger.utxo.token.cache.entities.CachedToken
>  import net.corda.ledger.utxo.token.cache.repositories.UtxoTokenRepository
> -import net.corda.orm.JpaEntitiesSet
>  import net.corda.ledger.utxo.token.cache.entities.TokenBalance
>  import net.corda.ledger.utxo.token.cache.entities.TokenPoolKey
> +import net.corda.orm.JpaEntitiesRegistry
>  import net.corda.virtualnode.read.VirtualNodeInfoReadService
>  import net.corda.virtualnode.VirtualNodeInfo
> -import org.slf4j.LoggerFactory
>  
>  @Component(
>      service = [AvailableTokenService::class]
> @@ -26,17 +26,14 @@ class AvailableTokenServiceImpl @Activate constructor(
>      @Reference
>      private val dbConnectionManager: DbConnectionManager,
>      @Reference
> +    private val jpaEntitiesRegistry: JpaEntitiesRegistry,
> +    @Reference
>      private val utxoTokenRepository: UtxoTokenRepository
>  ) : AvailableTokenService, SingletonSerializeAsToken {
> -
> -    companion object {
> -        private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
> -    }
> -
>      override fun findAvailTokens(poolKey: TokenPoolKey, ownerHash: String?, tagRegex: String?): AvailTokenQueryResult {
>          val virtualNode = getVirtualNodeInfo(poolKey)
>  
> -        val entityManagerFactory = createEntityManagerFactory(virtualNode)
> +        val entityManagerFactory = getOrCreateEntityManagerFactory(virtualNode)
>  
>          return utxoTokenRepository.findTokens(entityManagerFactory.createEntityManager(), poolKey, ownerHash, tagRegex)
>      }
> @@ -48,7 +45,7 @@ class AvailableTokenServiceImpl @Activate constructor(
>          claimedTokens: Collection<CachedToken>
>      ): TokenBalance {
>          val virtualNode = getVirtualNodeInfo(poolKey)
> -        val entityManagerFactory = createEntityManagerFactory(virtualNode)
> +        val entityManagerFactory = getOrCreateEntityManagerFactory(virtualNode)
>  
>          val totalBalance = utxoTokenRepository.queryBalance(entityManagerFactory.createEntityManager(), poolKey, ownerHash, tagRegex)
>          val claimedBalance = claimedTokens.sumOf { it.amount }
> @@ -57,10 +54,13 @@ class AvailableTokenServiceImpl @Activate constructor(
>          return TokenBalance(availableBalance, totalBalance)
>      }
>  
> -    private fun createEntityManagerFactory(virtualNode: VirtualNodeInfo) =
> -        dbConnectionManager.createEntityManagerFactory(
> +    private fun getOrCreateEntityManagerFactory(virtualNode: VirtualNodeInfo) =
> +        dbConnectionManager.getOrCreateEntityManagerFactory(
>              virtualNode.vaultDmlConnectionId,
> -            JpaEntitiesSet.create("emptySet", emptySet()) // No entity required. The mapping will be manually done
> +            jpaEntitiesRegistry.get(CordaDb.Vault.persistenceUnitName)
> +                ?: throw IllegalStateException(
> +                    "persistenceUnitName ${CordaDb.Vault.persistenceUnitName} is not registered."
> +                )
>          )
>  
>      private fun getVirtualNodeInfo(poolKey: TokenPoolKey) =
> ```
> 
> because creating a new `EntityManagerFactory` each time could be expensive and is definitely unnecessary. And you weren't destroying it after use either, which suggests a resource leak. Obviously, you won't _want_ to destroy it if you now fetch it from a pool instead wink.

This will be done in a different PR.

PR: https://github.com/corda/corda-runtime-os/pull/4463

filipesoliveira:
> Can you replace the `~` operator with Postgres's equivalent `REGEXP_LIKE` function please? Specifically:
> 
> ```sql
> REGEXP_LIKE(string, pattern)
> ```
> 
> This will allow us a least to _try_ to manage the database-specificity in the Hibernate `Dialect`. However, I'm not convinced that the E2E tests pass yet either.

This will be done in a different PR.

filipesoliveira:
> > so I think it is best not to change the behaviour at least until @owenstanford is back.
> 
> How about asking @blsemo?
> 
> BTW I have tried replacing
> 
> ```sql
> token_tag ~ :$SQL_PARAMETER_TAG_FILTER
> ```
> 
> with
> 
> ```sql
> REGEXP_MATCHES(token_tag, :$SQL_PARAMETER_TAG_FILTER)
> ```
> 
> for HSQLDB, and then executing my "driver" version of `TokenBalanceQueryTests`, but
> 
> ```
> ensure balance is equal to 2 but the balance including claimed tokens is 7
> ```
> 
> still fails with:
> 
> ```
> org.opentest4j.AssertionFailedError: 
> expected: 7
>  but was: 10
> ```
> 
> Are you sure that the E2E tests are fixed?

The E2E tests need to be updated. (PR that fixes the test: https://github.com/corda/corda-e2e-tests/pull/189)

josephzunigadaly (approved):

chrisr3:
> This will be done in a different PR.

Actually, can you include `REGEXP_LIKE` in _this_ PR please? It is trivial, and its absence is blocking my efforts to use HSQLDB.

filipesoliveira:
> > This will be done in a different PR.
> 
> Actually, can you include `REGEXP_LIKE` in _this_ PR please? It is trivial, and its absence is blocking my efforts to use HSQLDB.

I have raised a new PR with that change: https://github.com/corda/corda-runtime-os/pull/4458.

chrisr3 (approved):
LGTM - but see my comments for #4458.

PR #4456, opened 2023-08-14
CORE-8657 - Added check that CPI version is valid

malachyb:
Prevents CLI users from creating a CPI where the CPI version contains only special characters.

### DIFF
--- a/tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/CreateCpiV2.kt
+++ b/tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/CreateCpiV2.kt
@@ -96,6 +96,8 @@ class CreateCpiV2 : Runnable {
 
         GroupPolicyValidator.validateGroupPolicy(groupPolicyString)
 
+        verifyCpiVersion()
+
         val cpbPath = cpbFileName?.let { requireFileExists(it) }
 
         // Check input Cpb file is indeed a Cpb
@@ -143,6 +145,15 @@ class CreateCpiV2 : Runnable {
             .trustedCerts(readCertificates(signingOptions.keyStoreFileName, signingOptions.keyStorePass))
             .build()
             .verify()
+    }
+
+    /**
+     * @throws IllegalArgumentException if Cpb version is invalid
+     */
+    private fun verifyCpiVersion() {
+        if (!cpiVersion.matches(".*[a-zA-Z0-9].*".toRegex())) {
+            throw IllegalArgumentException("CPI version should include at least one alphanumeric character")
+        }
     }
 
     /**
--- a/tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/CreateCpiV2.kt
+++ b/tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/CreateCpiV2.kt
@@ -150,9 +150,11 @@ class CreateCpiV2 : Runnable {
     /**
      * @throws IllegalArgumentException if Cpi version is invalid
      */
+
     public fun verifyCpiVersion(version: String) {
         require(version.matches("[0-9][a-zA-Z0-9\\.-]*".toRegex())) {
-            "CPI version should match the Maven version specification (see https://maven.apache.org/pom.html#version-order-specification for more information)"
+            "CPI version should match the Maven version specification " +
+                    "(see https://maven.apache.org/pom.html#version-order-specification for more information)"
         }
     }
 
--- a/tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/CreateCpiV2.kt
+++ b/tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/CreateCpiV2.kt
@@ -96,7 +96,7 @@ class CreateCpiV2 : Runnable {
 
         GroupPolicyValidator.validateGroupPolicy(groupPolicyString)
 
-        verifyCpiVersion()
+        require(cpiVersion.matches(".*[a-zA-Z0-9].*".toRegex())) { "CPI version should include at least one alphanumeric character" }
 
         val cpbPath = cpbFileName?.let { requireFileExists(it) }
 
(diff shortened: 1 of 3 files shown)


tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/CreateCpiV2.kt
```diff
@@ -145,6 +147,15 @@ class CreateCpiV2 : Runnable {
             .verify()
     }
 
+    /**
+     * @throws IllegalArgumentException if Cpb version is invalid
+     */
+    private fun verifyCpiVersion() {
```
kyriathar: Perhaps we could try inline his function using `require`?

tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/CreateCpiV2.kt
```diff
@@ -145,6 +147,15 @@ class CreateCpiV2 : Runnable {
             .verify()
     }
 
+    /**
+     * @throws IllegalArgumentException if Cpb version is invalid
```
kyriathar: I believe this should be "Cpi"?

kyriathar (commented):
Some minor comments

tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/CreateCpiV2.kt
```diff
@@ -96,6 +96,8 @@ class CreateCpiV2 : Runnable {
 
         GroupPolicyValidator.validateGroupPolicy(groupPolicyString)
 
+        require(cpiVersion.matches(".*[a-zA-Z0-9].*".toRegex())) { "CPI version should include at least one alphanumeric character" }
```
kyriathar: Decision pending here but, If we say we want move on with this solution (which satisfies ticket: CORE-8657) then maybe worth adding a comment here on why we 'd want at least one alphanumeric character (which could be to include the notion of versioning).

driessamyn: Maybe it makes sense to follow an existing version specification. For example, we could just adopt the maven version specication: https://maven.apache.org/pom.html#version-order-specification

tools/plugins/package/src/test/kotlin/net/corda/cli/plugins/packaging/CreateCpiV2Test.kt
```diff
@@ -273,4 +273,26 @@ class CreateCpiV2Test {
...
+
+    @Test
+    fun `cpi create tool aborts if cpi version is invalid`() {
+        val outputFile = Path.of(tempDir.toString(), CPI_FILE_NAME)
+        val errText = TestUtils.captureStdErr {
+            CommandLine(CreateCpiV2()).execute (
+                "--cpb=${cpbPath}",
+                "--group-policy=${testGroupPolicy}",
+                "--cpi-name=testCpi",
+                "--cpi-version=\$%#",
```
driessamyn: This isn't actually a great test and does not test what you say it does, it only tests that if you have a version of exactly `\$%#` it's considered invalid.

Think about TDD, if this test was written first, before the implementation, it would probably have been written different as you would specify all the invalid version formats and validate that they are rejected (red/green/refactor).

Furthermore, this isn't really a unit test, it's an end-to-end test, and this also highlights a smell. Following the single responsibility principle, you should be able to define the thing that validates the version numbers and test that independently with fast unit tests.
You could, for example, inject a "VersionChecker" class or function that can be tested separately for valid and invalid version variations.

tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/CreateCpiV2.kt
```diff
@@ -147,6 +147,15 @@ class CreateCpiV2 : Runnable {
             .verify()
     }
 
+    /**
+     * @throws IllegalArgumentException if Cpi version is invalid
+     */
+    public fun verifyCpiVersion(version: String) {
```
driessamyn, line 154: A bit nitpicky here, but this still violates the single responsibility principle. Needing to make a function on a class public just to test it is a bit of a smell.
Another option would have been to create the verify-er function separately (so it has a separate set of tests) and inject into into this class.

However, as said, it's very nitpicky and feel free to ignore ;)

tools/plugins/package/src/test/kotlin/net/corda/cli/plugins/packaging/CreateCpiV2Test.kt
```diff
@@ -276,23 +277,25 @@ class CreateCpiV2Test {
...
-                "--keystore=${testKeyStore}",
-                "--storepass=keystore password",
-                "--key=${SIGNING_KEY_1_ALIAS}",
-                "--sig-file=$CPI_SIGNER_NAME"
-            )
-        }
+        val createTool = CreateCpiV2()
+        assertFailsWith<IllegalArgumentException>(
+            block = {
+                createTool.verifyCpiVersion("\$%#")
```
driessamyn: This is still only testing that this exact sequence of invalid characters are present.

This test is also testing many things. Keep in mind that it will fail after the full failure. When you practice TDD and test-first development, you want as much granularity as possible so you know excactly what is failing, not just the first.

This would be a good candidate for a [parameterized test](https://www.baeldung.com/parameterized-tests-junit-5#first-impression) (or probably 2, one golden path, one exception path).

malachyb: For your first point, would you suggest adding more invalid strings or another approach?

driessamyn: I would suggest treating them all separate.
I.e. `1.$` is an invalid version but also `1.%`, which are 2 different tests.

For example:
```
@ParameterizedTest
@ValueSource("1.\", "1.$", "1.%", "1.#")
fun `when invalid version throw`() ...
```

(BTW, I know this is all bit nitpicky, feel free to ignore ;) )

tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/CreateCpiV2.kt
simon-johnson-r3: Given this is a custom "require" type function, there's already an existing pattern for this, have a look at `FileHelpers`.

I would add a `requireCpiVersionMatchesMavenSpecification` in that file (maybe think of a shorter name :-) )

PR #4458, opened 2023-08-14
CORE-15179 - Replaced the operator ~ with an equivalent so the sql query is also supported by `HSQLDB`

filipesoliveira:


components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/queries/impl/SqlQueryProviderTokens.kt
```diff
@@ -18,12 +18,12 @@ class SqlQueryProviderTokens : SqlQueryProvider {
 
     override fun getBalanceQuery(includeTagFilter: Boolean, includeOwnerFilter: Boolean): String {
         val tagFilter = if(includeTagFilter){
-            "AND   token_tag ~ :$SQL_PARAMETER_TAG_FILTER"
+            "AND REGEXP_MATCHES(token_tag, :$SQL_PARAMETER_TAG_FILTER)"
```
chrisr3: I don't think this will work. Postgres says that `REGEXP_LIKE` is equivalent to the `~` operator. The point is that I can configure HSQLDB's Hibernate `Dialect` to replace `REGEXP_LIKE` with `REGEXP_MATCHES` whenever it sees `REGEXP_LIKE` in the SQL.

But as far as Postgres is concerned, the SQL should _still_ use `REGEXP_LIKE`.

filipesoliveira: I haven't tested the code yet. I simply copied and pasted the suggestion made in the other PR:
image

chrisr3: Yes, but I was using HSQLDB there. `REGEXP_MATCHES` is the correct function for HSQLDB, but Postgres's `REGEXP_MATCHES` function does something different:

> The regexp_matches function returns a set of text arrays of matching substring(s) within matches of a POSIX regular expression pattern to a string.

I am asking you to replace Postgres's `~` operator with Postgres's equivalent `REGEXP_LIKE` function, and then I can handle HSQLDB-compatibility myself.

filipesoliveira: I see. I misunderstood the comment. I will make that change.

filipesoliveira:
The e2e tests are green with this change. So it should be fairly safe to merge this PR once all the checks are green.

josephzunigadaly (approved):

PR #4459, opened 2023-08-14
CORE-16264: Remove flow wakeup scheduler and the rebalance listener

JamesHR3:
**Problem description**

The flow engine will publish `Wakeup` events periodically to perform maintenance on flow checkpoints. This can result in swamping the flow event topic with events when the system is already overloaded. As a result, this mechanism for performing session timeouts and replays should be removed.

**Solution**

Removes the flow wakeup scheduler and the flow executor rebalance listener.

This will result in some behaviours being disabled (in particular session and external event retries). These will be re-established in a later PR using a different mechanism.

**Testing**

For this PR, testing is confined to ensuring the full suite of tests pass. A later PR will confirm that all `Wakeup` events are removed from the system under load scenarios once the rest of the `Wakeup` removal work has been completed.

**Other Changes**
- Flakiness can occur in the end to end tests when performing a config change. This happens if the system is unusually slow to process the change, leaving the REST endpoint unavailable for slightly longer than anticipated. The current end to end client code has a shorter timeout for this case than the rest of the interactions with the REST endpoint (owing to a different code path). This has been increased to move it in line with the timeout elsewhere.


JamesHR3:
Draft while further testing ongoing.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/config/SingleClusterTestConfigManager.kt
```diff
@@ -71,7 +72,7 @@ class SingleClusterTestConfigManager(
             if(newConfig != previousSourceConfig) {
                 updateConfig(newConfig, section)
```
emilybowe, line 74: does updateConfig also need the increased timeout that you've added for getConfig?

emilybowe (approved):
LGTM

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/config/SingleClusterTestConfigManager.kt
```diff
@@ -95,7 +96,7 @@ class SingleClusterTestConfigManager(
             )
             updateConfig(preTestConfig, section)
 
-            eventually {
+            eventually(duration = Duration.ofSeconds(30)) {
```
thiagoviana: Get the value (30) from a constant?

JamesHR3: Good catch, thanks!

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/config/SingleClusterTestConfigManager.kt
JamesHR3: Good question. Looking at it, the `updateConfig` method first tries to `getConfig`, but wraps it in an `assertWithRetryIgnoringExceptions` and so it will block for 1 minute waiting for the endpoint to be up already. As a result I think it's ok to leave this one.

jujoramos (approved):

Omar-awad (commented):
Looks good to me.

PR #4465, opened 2023-08-15
CORE-16228 Splitting FlowWorker into Flow and Verification workers

ben-millar:
### Overview:
The aim of this PR is to split out the `VerificationProcessor` from the `FlowWorker` into its own worker: a `VerificationWorker`. This will allow for more granular scaling of Corda clusters in response to increased traffic. This is part of a larger epic to create a more focused set of Corda workers; more detailed design decisions are being captured in the design document [HERE](https://github.com/corda/platform-eng-design/blob/bm/CORE-15877/core/corda-5/corda-5.0/worker-segragation/segregated-and-granular-worker-design.md).
### Related PRs:
https://github.com/corda/corda-api/pull/1214
https://github.com/corda/corda-shared-build-pipeline-steps/pull/1063
https://github.com/corda/corda-e2e-tests/pull/203 // Not to be merged
https://github.com/corda/corda-non-functional-test/pull/225
https://github.com/corda/corda-soak-test-deployment/pull/123


chrisr3:
> This will allow for more granular scaling of Corda clusters in response to increased traffic.

But if we _don't_ do this then deploying a single new worker pod gives you _both_ a new verification processor _and_ a new flow processor, and the processor with more actual work to do will naturally end up using more CPU time than the other. Isn't this a more efficient use of available computing resources? What you are doing here is contrary to what I was taught about J2EE deployments - that you deploy _all_ server processes to _every_ physical server in your cluster.

We are obviously using Cloud servers rather than physical hardware these days, but if you need 10 flow workers and only 5 verification workers then why launch 15 instances when you can launch only 10?

AFAICT all you are really doing here is increasing Corda's running costs.

ben-millar:
Hey @chrisr3, I've given an example in the design document [HERE](https://github.com/corda/platform-eng-design/blob/bm/CORE-15877/core/corda-5/corda-5.0/worker-segragation/segregated-and-granular-worker-design.md#why-should-we-change-it), but I think you get the gist of it.

We're talking about deploying a greater number of proportionally smaller workers; essentially we're trimming the fat and deploying only what we need. This does increase the total cost of ownership, particularly for smaller clusters, but are you suggesting that this totally negates any benefit? Kubernetes doesn't charge per-process, so we're just talking about the overhead of running another JVM process here.

chrisr3:
> Kubernetes doesn't charge per-process, so we're just talking about the overhead of running another JVM process here.

A JVM is not a lightweight thing; it has a considerable start-up cost, and expects to be long-running so that its Just-In-Time compiler can optimise the bytecode. And Kubernetes or no, running four JVMs is still going to require more resources than running a single JVM. There was actually an issue on Jenkins this week when two pipeline stages were switched to running in parallel, causing their dockerised Gradle daemons to collide. The resulting slow-downs and deadlocks were resolved by serialising those two stages again.

But I think you lost me when you made this claim about starting four extra instances of `FlowMapperService`:

> Clearly, this is an inefficient allocation of resources.

Err, more inefficient than starting _four extra JVMs_? What on Earth are these `FlowMapperService` instances doing when they're not running, apart from consuming a trivial amount of memory?

Omar-awad:
The main drivers for this segregation are:

 - Fine tune processes/services on the flow critical execution path individually.
	 - This includes the message bus configuration like Kafka topic partitions.
 - Being able to switch the transport method for these services individually.
	 - Some transports like HTTP might require additional pod level setup to support load balancing.
 - Reduce the blast radius of potential failures.
 - Improve our application insights and resource utilization tracking per process.

The objective is to have efficient utilization of each critical service assigned resources efficiently and being able to track/monitor its performance individually which is hard to do in the current setup. In addition to that, having many heavy processes running in one JVM even if not processing events they still add an overhead. Maybe we have to pay a marginal for extra JVM's but thats countered by having lighter weight processors and the benefits outlined above.

> There was actually an issue on Jenkins this week when two pipeline stages were switched to running in parallel, causing their dockerised Gradle daemons to collide. The resulting slow-downs and deadlocks were resolved by serialising those two stages again.

Please explain how does that relates to our case.

chrisr3:
> > There was actually an issue on Jenkins this week when two pipeline stages were switched to running in parallel, causing their dockerised Gradle daemons to collide. The resulting slow-downs and deadlocks were resolved by serialising those two stages again.
> 
> Please explain how does that relates to our case.

You seem to be assuming that putting lots of JVMs into Kubernetes separate pods means that they will always run independently, as if on separate servers. I don't think that's entirely true...

As for the rest...

My previous employment used a micro-service architecture, with everything communicating via synchronous HTTP. We discovered that HTTP(s) connections are expensive to create, and not particularly reliable either. Fortunately, that system only required _eventual_ consistency - we could recover from errors so long as all the messages were idempotent.

We also discovered that synchronous HTTP was not particularly compatible with high throughput, and that "auto-scaling" the cluster efficiently was a non-trivial problem.

As far as I was aware here, we only needed to use Kafka _less_, e.g. by routing messages locally where possible instead of sending them across the Kafka bus, rather than opening a whole new can of worms called HTTP.

Omar-awad:
>You seem to be assuming that putting lots of JVMs into Kubernetes separate pods means that they will always run independently, as if on separate servers. I don't think that's entirely true...

It's fine. We will be able to monitor the performance of each container and tune it accordingly. I don't see the problem with sharing CPU's time as long they are separate processes.

> We also discovered that synchronous HTTP was not particularly compatible with high throughput, and that "auto-scaling" the cluster efficiently was a non-trivial problem. As far as I was aware here, we only needed to use Kafka  _less_, e.g. by routing messages locally where possible instead of sending them across the Kafka bus, rather than opening a whole new can of worms called HTTP.

Our objective is not to deliver a final solution rather a stepping stone in the journey. HTTP as any other tech has its own limitations but it's suitable for the first few milestones we are aiming for which is few hundred transactions per second. As we go in the journey there are more design changes coming to tackle performance and scalability limits. Today we are only focusing on the topology and in later releases we will address the transport layer in more details.

ben-millar:
>As far as I was aware here, we only needed to use Kafka less, e.g. by routing messages locally where possible instead of sending them across the Kafka bus, rather than opening a whole new can of worms called HTTP.

I don't think that's correct, actually. In our testing, we found the best improvements not from reducing Kafka messages (which actually had very little effect IIRC, at least when isolated to the flow stuff) but by enforcing some prioritisation on the messages. I.e., if there's some work we can do right now, we should do it rather than pushing it into a Kafka queue to pick up later. This is partially what we're doing now with the synchronous RPC calls.

chrisr3:
> > As far as I was aware here, we only needed to use Kafka less, e.g. by routing messages locally where possible instead of sending them across the Kafka bus, rather than opening a whole new can of worms called HTTP.
> 
> I.e., if there's some work we can do right now, we should do it rather than pushing it into a Kafka queue to pick up later. 

Err, yes? When I say "routing messages locally" I _mean_ handling it immediately within the same JVM rather than pushing it across either the Kakfa bus, or even HTTP for that matter. Which does require that the JVM also be hosting all the relevant handlers, of course.

Omar-awad:
>Err, yes? When I say "routing messages locally" I _mean_ handling it immediately within the same JVM rather than pushing it across either the Kakfa bus, or even HTTP for that matter. Which does require that the JVM also be hosting all the relevant handlers, of course.

That also is covered in the initiative. There is a separate design review for removing scheduled wake up events and process them in the same process without going over-the-wire.

chrisr3:
> > Err, yes? When I say "routing messages locally" I _mean_ handling it immediately within the same JVM rather than pushing it across either the Kafka bus, or even HTTP for that matter. Which does require that the JVM also be hosting all the relevant handlers, of course.
> 
> That also is covered in the initiative. There is a separate design review for removing scheduled wake up events and process them in the same process without going over-the-wire.

How are these two initiatives compatible? You cannot handle events within the same process if you are simultaneously moving all the different handlers into separate processes.

chrisr3:
> I don't see the problem with sharing CPU's time as long they are separate processes.

The problem on Jenkins appeared to involve the two pods fighting each other and degrading each other's network performance. I don't think sharing CPU time was the issue.

> Our objective is not to deliver a final solution rather a stepping stone in the journey. HTTP as any other tech has its own limitations but it's suitable for the first few milestones we are aiming for which is few hundred transactions per second. As we go in the journey there are more design changes coming to tackle performance and scalability limits. Today we are only focusing on the topology and in later releases we will address the transport layer in more details.

But today's road must still stretch beyond that first milestone and be capable of actually _reaching_ that final destination, as otherwise it's like we're trying to reach the moon by climbing a tree. And HTTP _does_ look like a tree to me.

blsemo:
I am not convinced by this. Please see my comment on the JIRA why.

Omar-awad:
> I am not convinced by this. Please see my comment on the JIRA why.
I replied. The longer term vision meets your concerns.

Omar-awad:
This PR looks good to me.

PR #4466, opened 2023-08-15
CORE-14685: Support for UTXO token cache.

chrisr3:
Add `TokenCacheProcessor` to handle flows that issue and consume UTXO tokens.

The token cache uses Postgres-specific `REGEXP_LIKE` in its SQL query, which Hibernate must map to `REGEXP_MATCHES` when using HSQLDB.

Ultimately, Corda will need to decide whether it should continue to use regular expressions here, or switch to something more compatible with ANSI SQL `LIKE`.


wzur-r3 (approved):
LGTM for changes of Gradle files

yarivus (commented):
thx @chrisr3

PR #4467, opened 2023-08-15
CORE-16269 - Error handling has been improved for plugin commands

r3-arish:
This PR addresses error handling by introducing the `verifyAndPrintError` lambda function in `run` blocks of several classes such as `OnboardMgm`, `AllowClientCertificate`, `GenerateGroupPolicy`, `GroupParametersLookup` and `MemberLookup`. 

This change will prevent any stack trace being shown to the end-user.

Also, the `verifyAndPrintError` function has been modified to address this comment [here](https://github.com/corda/corda-runtime-os/pull/4442#discussion_r1293497776)

## Result
Screenshot 2023-08-15 at 14 26 17
Screenshot 2023-08-15 at 14 34 33
Screenshot 2023-08-15 at 15 29 50
Screenshot 2023-08-15 at 15 29 34

## New Result (Pretty Printing)
Screenshot 2023-08-15 at 16 45 01


tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMgm.kt
```diff
@@ -179,26 +180,29 @@ class OnboardMgm : Runnable, BaseOnboard() {
     }
 
     override fun run() {
-        println("This sub command should only be used in for internal development")
-        println("On-boarding MGM member $x500Name")
+        verifyAndPrintError {
+            println("This sub command should only be used in for internal development")
```
YashNabar: Do we still want this here?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/GenerateGroupPolicy.kt
```diff
@@ -49,15 +50,17 @@ class GenerateGroupPolicy(private val output: GroupPolicyOutput = ConsoleGroupPo
...
+        verifyAndPrintError {
+            val objectMapper = jacksonObjectMapper()
+            val groupPolicy = generateGroupPolicyContent()
 
-        // add pretty printer and override indentation to make the nested values look better and the file more presentable
-        val pp = DefaultPrettyPrinter()
-        pp.indentArraysWith(DefaultIndenter.SYSTEM_LINEFEED_INSTANCE)
+            // add pretty printer and override indentation to make the nested values look better and the file more presentable
+            val pp = DefaultPrettyPrinter()
+            pp.indentArraysWith(DefaultIndenter.SYSTEM_LINEFEED_INSTANCE)
```
YashNabar: Can't we use `printJsonOutput()` here as well?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/utils/PrintUtils.kt
```diff
@@ -27,17 +28,27 @@ class PrintUtils {
...
                 else -> throw IllegalArgumentException("Unsupported output type")
             }
         }
 
         fun verifyAndPrintError(action: () -> Unit) {
             try {
                 action()
+            } catch (e: InternalErrorException) {
+                if (e.localizedMessage.contains("URI does not specify a valid host name")) {
+                    System.err.println("Error: Invalid host name")
```
YashNabar: Why is this particular error message a special case?

dimosr: I suspect this came as a reaction to [my comment](https://github.com/corda/corda-runtime-os/pull/4442#discussion_r1293497776) on the previous PR. But, I think the right way to do this is to perform explicit validation that the target is a well-formed URL, instead of letting the HTTP request fail and parse the exception message.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMgm.kt
r3-arish: The print statement?

YashNabar: Yes, not sure why it's only printed for specific commands. We should either print it everywhere (in which case we should fix the typo there) or remove it entirely

dimosr: I think that might have been added back when we started adding capabilities that made the tool easier to use for us (e.g. the native k8s capabilities to auto-discover things), but not necessarily for all use cases. Given we'll have a general designation for the plugin on what it's suitable for, we could remove this now to avoid causing confusion.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/utils/PrintUtils.kt
r3-arish: Yes, now I have moved the explicit validation of the `targetUrl` in the `RestClientUtils`.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
```diff
@@ -253,7 +253,6 @@ class OnBoardMember : Runnable, BaseOnboard() {
 
     override fun run() {
         verifyAndPrintError {
-            println("This sub command should only be used in for internal development")
```
dimosr: There are more instances of this. Search for `This sub command should only be used` and you can find them.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/utils/PrintUtils.kt
```diff
@@ -27,8 +27,12 @@ class PrintUtils {
             pp.indentArraysWith(DefaultIndenter.SYSTEM_LINEFEED_INSTANCE)
 
             val jsonString = objectMapper.writer(pp).writeValueAsString(result)
+            val formattedString = jsonString
+                .replace("\\n", "")
+                .replace("\\","")
+                .replace("\"\"", "")
```
dimosr, line 33: what is the motivation behind these replacements?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/utils/PrintUtils.kt
```diff
@@ -37,7 +41,7 @@ class PrintUtils {
             try {
                 action()
             } catch (e: Exception) {
-                System.err.println("Error: ${e.message}")
+                printJsonOutput(e.localizedMessage, ConsoleOutput())
```
dimosr, line 48: why do we use `printJsonOutput` here? That seems like a plain string, not a json object.

tools/plugins/plugins-rest/src/main/kotlin/net/corda/cli/plugins/common/RestClientUtils.kt
```diff
@@ -86,12 +88,16 @@ object RestClientUtils {
...
         logger.info("Re-throwing", ex)
         throw ex
     }
 
-    fun ignore(ex: ResourceAlreadyExistsException) {
-        logger.debug("Ignoring", ex)
+    private fun validateTargetUrl(url: String) {
+        try {
+            URL(url)
+        } catch (e: Exception) {
```
dimosr: Given there's a more specific exception thrown for invalid URLs (`MalformedURLException`), I think it's better to catch this instead.

r3-arish: Okay, I will look into that particular exception

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/utils/PrintUtils.kt
r3-arish: Sometimes, the error response is a JSON

r3-arish: 260765299-3015aadb-2e91-4454-b946-f25c3bf4df90
@dimosr

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/utils/PrintUtils.kt
r3-arish: The error response sent by the REST endpoints have escape characters in them, it looks a bit messy - thats the reason why I used those replacements

r3-arish: Screenshot 2023-08-16 at 16 19 50
@dimosr

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/utils/PrintUtils.kt
dimosr: Can you share some more details on where exactly this error is coming from, what is the scenario that triggers this and also what is the actual exception type in that case? It's not obvious from the screenshot.

dimosr: hmm, I think I know what's happening here.. the `RemoteClient` in `rest-client` [is automatically converting ](https://github.com/corda/corda-runtime-os/blob/a7b50f89a5cb81c8acae99264ebe74cd47522f5f/libs/rest/rest-client/src/main/kotlin/net/corda/rest/client/connect/remote/RemoteClient.kt#L118)any non-200 codes into exceptions. In this case, a 409 is converted into a `ResourceAlreadyExistsException` with the payload of the body as the message of the exception.

dimosr: I am in two minds. One one hand, we could handle these exceptions and throw an exception with a better message. However, we'll still need the parsing as a belt and braces approach, since there can always be a place where we forgot to check for some error case.

So, I think we can keep the parsing there. But, we should add a comment to clarify why we do this (and same for the escaping above).

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/utils/PrintUtils.kt
```diff
@@ -27,17 +27,27 @@ class PrintUtils {
...
                 else -> throw IllegalArgumentException("Unsupported output type")
             }
         }
 
+        /**
+         * This function [verifyAndPrintError] is present to address the issue
+         * of the RemoteClient in rest-client automatically converts any non-200 codes into exceptions.
+         * In this case, a 409 is converted into a ResourceAlreadyExistsException
+         * with the payload of the body as the message of the exception.
+         */
```
dimosr: This comment should go below at the point we print the exception message as json (e.g. before line 50), not in this function, since this function has a larger purpose which is to catch any exceptions that escape during execution of a command regardless of whether they are coming from the RestClient or not.

I would also remove the second part:
```
 * In this case, a 409 is converted into a ResourceAlreadyExistsException
 * with the payload of the body as the message of the exception.
```
as this can apply to other error codes, not just 409. 409 was just the screenshot you shared.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnBoardMember.kt
dimosr: @r3-arish this hasn't been addressed yet.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/utils/PrintUtils.kt
r3-arish: Understood, will address it

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/utils/PrintUtils.kt
```diff
@@ -37,7 +41,11 @@ class PrintUtils {
             try {
                 action()
             } catch (e: Exception) {
-                System.err.println("Error: ${e.message}")
+                /**
+                 * This ia present to address the issue of the RemoteClient in
+                 * rest-client automatically converts any non-200 codes into exceptions.
+                 */
```
dimosr:
```suggestion
                /**
                 * This is present to address the issue of the RemoteClient in
                 * rest-client automatically converting any non-200 codes into exceptions with the response body as message.
                 */
```

dimosr (commented):
just a minor typo - otherwise LGTM

dimosr (approved):

YashNabar (approved):

PR #4471, opened 2023-08-16
CORE-10386 Re-execute named query until result set is filled

nkovacsx:
### Overview

Change the main named query executor logic to the following:
1. Execute the query for the first time and filter the results.  If the `limit`(page size) is filled up, return the results.
2. If not, then execute the query once again with the original `limit` and repeat until either
- the page size is reached after filtering 
- we run out of records from the database
- we reach the maximum number of retries (currently 5)

### Test plan

Tested using the e2e test cases from https://github.com/corda/corda-e2e-tests/pull/195 and verified the logs to trace that the logic works as expected.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/external/events/FindAllExternalEventFactory.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/external/events/FindAllExternalEventFactory.kt
@@ -37,8 +37,7 @@ class FindAllExternalEventFactory: ExternalEventFactory<FindAllParameters, Entit
 
         return ResultSetExecutor.Results(
             serializedResults = response.results,
-            numberOfRowsFromQuery = numberOfRowsFromQuery,
-            newOffset = numberOfRowsFromQuery
+            numberOfRowsFromQuery = numberOfRowsFromQuery
         )
     }
 }
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/external/events/FindAllExternalEventFactory.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/external/events/FindAllExternalEventFactory.kt
@@ -33,9 +33,12 @@ class FindAllExternalEventFactory: ExternalEventFactory<FindAllParameters, Entit
     }
 
     override fun resumeWith(checkpoint: FlowCheckpoint, response: EntityResponse): ResultSetExecutor.Results {
+        val numberOfRowsFromQuery = response.metadata.items.single { it.key == "numberOfRowsFromQuery" }.value.toInt()
+
         return ResultSetExecutor.Results(
             serializedResults = response.results,
-            numberOfRowsFromQuery = response.metadata.items.single { it.key == "numberOfRowsFromQuery" }.value.toInt()
+            numberOfRowsFromQuery = numberOfRowsFromQuery,
+            newOffset = numberOfRowsFromQuery
         )
     }
 }
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/external/events/NamedQueryExternalEventFactory.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/external/events/NamedQueryExternalEventFactory.kt
@@ -34,9 +34,12 @@ class NamedQueryExternalEventFactory : ExternalEventFactory<NamedQueryParameters
     }
 
     override fun resumeWith(checkpoint: FlowCheckpoint, response: EntityResponse): ResultSetExecutor.Results {
+        val numberOfRowsFromQuery = response.metadata.items.single { it.key == "numberOfRowsFromQuery" }.value.toInt()
+
         return ResultSetExecutor.Results(
             serializedResults = response.results,
-            numberOfRowsFromQuery = response.metadata.items.single { it.key == "numberOfRowsFromQuery" }.value.toInt()
+            numberOfRowsFromQuery = numberOfRowsFromQuery,
+            newOffset = numberOfRowsFromQuery
         )
     }
 }
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/external/events/NamedQueryExternalEventFactory.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/external/events/NamedQueryExternalEventFactory.kt
@@ -38,8 +38,7 @@ class NamedQueryExternalEventFactory : ExternalEventFactory<NamedQueryParameters
 
         return ResultSetExecutor.Results(
             serializedResults = response.results,
-            numberOfRowsFromQuery = numberOfRowsFromQuery,
-            newOffset = numberOfRowsFromQuery
+            numberOfRowsFromQuery = numberOfRowsFromQuery
         )
     }
 }
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
@@ -18,7 +18,7 @@ data class ResultSetImpl<R> internal constructor(
     init {
         // Ideally this will never happen, but we keep this check in here for safety
         require(offset >= 0) {
-            "Offset cannot be negative or zero"
+            "Offset cannot be negative"
         }
         require(limit > 0) {
             "Limit cannot be negative or zero"
@@ -43,18 +43,8 @@ data class ResultSetImpl<R> internal constructor(
         }
         val (serializedResults, numberOfRowsFromQuery) = resultSetExecutor.execute(serializedParameters, offset)
 
-        // Here have the number of rows we fetched from the database (before filtering)
-        // then we check if the remainder is 0 when diving the number of rows fetched with the limit
-        // This will:
-        // 1. yield false only if we don't have a next page to fetch
-        // Examples:
-        // `limit` = 5, `numberOfRowsFromQuery` = 9, `hasNext = 9 % 5 == 0` => `hasNext = 4 == 0` == false
-        // `limit` = 5, `numberOfRowsFromQuery` = 4, `hasNext = 4 % 5 == 0` => `hasNext = 4 == 0` == false
-        // 2. yield true if we can't know for sure if there's a next page
-        // Example:
-        // `limit` = 5, `numberOfRowsFromQuery` = 5, `hasNext = 5 % 5 == 0` => `hasNext =  0 == 0` == true
         hasNext = numberOfRowsFromQuery != 0  // If there are no rows left there are no more pages to fetch
-                && numberOfRowsFromQuery % limit == 0
+                && serializedResults.size == limit // If the current page is full, it means we might have more records, so we go and check
         offset += numberOfRowsFromQuery
         results = serializedResults.map { serializationService.deserialize(it.array(), resultClass) }
         return results
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
@@ -31,9 +31,9 @@ data class ResultSetImpl<R> internal constructor(
         if (!hasNext()) {
             throw NoSuchElementException("The result set has no more pages to query")
         }
-        val (serializedResults, numberOfRowsFromQuery) = resultSetExecutor.execute(serializedParameters, offset)
+        val (serializedResults, numberOfRowsFromQuery, newOffset) = resultSetExecutor.execute(serializedParameters, offset)
         hasNext = limit in 1..numberOfRowsFromQuery
-        offset += limit
+        offset = newOffset
         results = serializedResults.map { serializationService.deserialize(it.array(), resultClass) }
         return results
     }
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
@@ -31,9 +31,20 @@ data class ResultSetImpl<R> internal constructor(
         if (!hasNext()) {
             throw NoSuchElementException("The result set has no more pages to query")
         }
-        val (serializedResults, numberOfRowsFromQuery, newOffset) = resultSetExecutor.execute(serializedParameters, offset)
-        hasNext = limit in 1..numberOfRowsFromQuery
-        offset = newOffset
+        val (serializedResults, numberOfRowsFromQuery) = resultSetExecutor.execute(serializedParameters, offset)
+
+        // Here have the number of rows we fetched from the database (before filtering)
+        // then we check if the remainder is 0 when diving the number of rows fetched with the limit
+        // This will:
+        // 1. yield false only if we don't have a next page to fetch
+        // Examples:
+        // `limit` = 5, `numberOfRowsFromQuery` = 9, `hasNext = 9 % 5 == 0` => `hasNext = 4 == 0` == false
+        // `limit` = 5, `numberOfRowsFromQuery` = 4, `hasNext = 4 % 5 == 0` => `hasNext = 4 == 0` == false
+        // 2. yield true if we can't know for sure if there's a next page
+        // Example:
+        // `limit` = 5, `numberOfRowsFromQuery` = 5, `hasNext = 5 % 5 == 0` => `hasNext =  0 == 0` == true
+        hasNext = numberOfRowsFromQuery != 0 && numberOfRowsFromQuery % limit == 0
+        offset += numberOfRowsFromQuery
         results = serializedResults.map { serializationService.deserialize(it.array(), resultClass) }
         return results
     }
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
@@ -15,6 +15,16 @@ data class ResultSetImpl<R> internal constructor(
     private val resultSetExecutor: ResultSetExecutor<R>
 ) : PagedQuery.ResultSet<R> {
 
+    init {
+        // Ideally this will never happen, but we keep this check in here for safety
+        require(offset >= 0) {
+            "Offset cannot be negative or zero"
+        }
+        require(limit > 0) {
+            "Limit cannot be negative or zero"
+        }
+    }
+
     private var results: List<R> = emptyList()
     private var hasNext: Boolean = true
 
@@ -44,7 +54,6 @@ data class ResultSetImpl<R> internal constructor(
         // Example:
         // `limit` = 5, `numberOfRowsFromQuery` = 5, `hasNext = 5 % 5 == 0` => `hasNext =  0 == 0` == true
         hasNext = numberOfRowsFromQuery != 0  // If there are no rows left there are no more pages to fetch
-                && limit != 0 // If the limit is 0 it means we fetch everything in one go so there are no pages left
                 && numberOfRowsFromQuery % limit == 0
         offset += numberOfRowsFromQuery
         results = serializedResults.map { serializationService.deserialize(it.array(), resultClass) }
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
@@ -43,7 +43,9 @@ data class ResultSetImpl<R> internal constructor(
         // 2. yield true if we can't know for sure if there's a next page
         // Example:
         // `limit` = 5, `numberOfRowsFromQuery` = 5, `hasNext = 5 % 5 == 0` => `hasNext =  0 == 0` == true
-        hasNext = numberOfRowsFromQuery != 0 && numberOfRowsFromQuery % limit == 0
+        hasNext = numberOfRowsFromQuery != 0  // If there are no rows left there's no more pages to fetch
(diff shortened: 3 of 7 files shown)


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -31,6 +31,9 @@ class VaultNamedQueryExecutorImpl(
         const val UTXO_TX_COMPONENT_TABLE = "utxo_transaction_component"
         const val TIMESTAMP_LIMIT_PARAM_NAME = "Corda_TimestampLimit"
 
+        // TODO Should this be configurable or hard-coded?
```
relyafi: Should take a decision on this now, and raise (and reference) a ticket if we do want to be configurable but aren't doing yet. I'd be tempted to leave it fixed for now.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -75,17 +75,73 @@ class VaultNamedQueryExecutorImpl(
         // Return the filtered/transformed/collected (if present) result to the caller
         return EntityResponse.newBuilder()
             .setResults(collectedResults.map { ByteBuffer.wrap(serializationService.serialize(it).bytes) })
-            .setMetadata(KeyValuePairList(listOf(KeyValuePair("numberOfRowsFromQuery", contractStateResults.size.toString()))))
+            // TODO Should `numberOfRowsFromQuery` be the number of rows before or after filtering?
```
relyafi: What's the intended use of this metadata? If it's for the benefit of the caller, I'd say after (the in-memory filtering is a black box, anything on the API should reflect the end result, not the internal processing).

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -75,17 +75,73 @@ class VaultNamedQueryExecutorImpl(
...
+            // If we can't fetch more states we just return the result set as-is
+            if (contractStateResults.isEmpty())
+                break
+
+            // Increase the internal offset, so we don't fetch the same results multiple times
+            internalOffset += contractStateResults.size
+
+            // Apply filters (if there's any) and add the filtered results to the final result set
+            filteredResults.addAll(contractStateResults.filter {
+                vaultNamedQuery.filter?.filter(it, deserializedParams) ?: true
```
relyafi: How common do we expect filters to be? If no filter is a relatively common case, might want to do the check first and skip the filter step here if so (means doing the check once, rather than for each element returned).

nkovacsx: Good point actually!

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -126,12 +124,15 @@ class VaultNamedQueryExecutorImpl(
...
 
             // Apply filters (if there's any) and add the filtered results to the final result set
-            filteredResults.addAll(contractStateResults.filter {
-                vaultNamedQuery.filter?.filter(it, deserializedParams) ?: true
-            })
+            vaultNamedQuery.filter?.let { vaultNamedQueryFilter ->
+                filteredResults.addAll(contractStateResults.filter {
+                    vaultNamedQueryFilter.filter(it, deserializedParams)
+                })
+            }
```
relyafi: Doesn't this now return no results if we don't have a filter?

relyafi (approved):

relyafi (dismissed):

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -31,6 +31,8 @@ class VaultNamedQueryExecutorImpl(
         const val UTXO_TX_COMPONENT_TABLE = "utxo_transaction_component"
         const val TIMESTAMP_LIMIT_PARAM_NAME = "Corda_TimestampLimit"
 
+        const val RESULT_SET_FILL_RETRY_LIMIT = 5
```
lankydan, line 34: Maybe `RESULT_SET_REPEAT_QUERY_LIMIT` might be better?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -75,17 +74,77 @@ class VaultNamedQueryExecutorImpl(
...
+     *
+     * This logic will be repeated until either:
+     * - The size of the filtered results reaches the "page size"
+     * - We run out of data to fetch from the database
+     * - We reach the number of retries ([RESULT_SET_FILL_RETRY_LIMIT])
+     *
+     * If any of these conditions happen, we just return the result set as-is without filling
+     * up the "page size".
+     */
+    private fun filterResultsAndFillPageSize(
```
lankydan, line 103: Thanks for adding this comment, I was going to add a comment asking you to do this (which I then deleted 😄).

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
lankydan: The `FILL` part is quite good though. Maybe `RESULT_SET_FILL_PAGE_REQUERY_LIMIT` gives all the context we'd need?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -75,17 +74,77 @@ class VaultNamedQueryExecutorImpl(
         // Return the filtered/transformed/collected (if present) result to the caller
         return EntityResponse.newBuilder()
             .setResults(collectedResults.map { ByteBuffer.wrap(serializationService.serialize(it).bytes) })
-            .setMetadata(KeyValuePairList(listOf(KeyValuePair("numberOfRowsFromQuery", contractStateResults.size.toString()))))
```
lankydan, line 78: The code a few lines up:

```kotlin
val collectedResults = vaultNamedQuery.collector?.collect(
            filteredAndTransformedResults,
            deserializedParams
        )?.results?.filterNotNull() ?: filteredAndTransformedResults
```

Originally I was thinknig that we could also decide to stop querying if the results of running a `VaultNamedQueryCollector` returned `isDone == true`, but the more I think about it, I feel that the collector doesn't really fit correctly. Mainly that the `resultSet` it takes in would need to contain the rows of previous query executions to really "collect" things (while still allowing the query to be re-executed).

I'm ok with not refactoring the collection part into the requerying logic you've added. We can always come back to this in the future if we decide to add it, or we might possibly deprecate the collector interface since I'm not really happy with it when I reflect on it.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -75,17 +74,77 @@ class VaultNamedQueryExecutorImpl(
...
+                offsetOverride = internalOffset
+            )
+
+            // If we have no filter, there's no need to continue the loop
+            if (vaultNamedQuery.filter == null)
+                return contractStateResults
+
+            // If we can't fetch more states we just return the result set as-is
+            if (contractStateResults.isEmpty())
+                break
```
lankydan, line 135: I'm not _that_ keen on the `break`, but I'm not sure if any other way of writing it would be any better.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -75,17 +74,77 @@ class VaultNamedQueryExecutorImpl(
...
+    ): List<StateAndRef<ContractState>> {
+        val filteredResults = mutableListOf<StateAndRef<ContractState>>()
+
+        var internalOffset = request.offset
+        var currentRetry = 0
+
+        while (filteredResults.size < request.limit && currentRetry < RESULT_SET_FILL_RETRY_LIMIT) {
+            ++currentRetry
+
+            log.debug("Executing try: $currentRetry, fetched ${filteredResults.size} number of results so far.")
```
lankydan: Can we move this down to `trace` for now.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -75,17 +74,77 @@ class VaultNamedQueryExecutorImpl(
...
+
     /**
      * A function that fetches the contract states that belong to the given transaction IDs. The data stored in the
      * component table will be deserialized into contract states using component groups.
      */
     private fun fetchStateAndRefs(
         request: FindWithNamedQuery,
-        whereJson: String?
+        whereJson: String?,
+        offsetOverride: Int? = null
```
lankydan: I think change `offsetOverride` to `offset` and always pass it into `fetchStateAndRefs`.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -75,17 +74,77 @@ class VaultNamedQueryExecutorImpl(
...
-            .setMetadata(KeyValuePairList(listOf(KeyValuePair("numberOfRowsFromQuery", contractStateResults.size.toString()))))
+            .setMetadata(KeyValuePairList(listOf(KeyValuePair("numberOfRowsFromQuery", filteredAndTransformedResults.size.toString()))))
             .build()
     }
 
+    /**
+     * This function will fetch a given number ("page size") of records from the database
+     * (amount defined by [request]'s limit field).
+     *
+     * After fetching those records the in-memory filter will be applied. If the filtering
```
lankydan, line 88: I think we're missing some data being passed back to the flow here.

If we're changing the offset under the hood, then we need to communicate this back to the flow. If we don't do this then when we query for the next page (from the flow side) then we're going to re-query data that we've already been through on the previous iteration.

This will relate to the changes that @relyafi has or is going to propose around querying (don't know where things are atm). We'll either need to pass an offset to the flow side (I introduced a metadata map in the avro response to make adding these changes easier), or update the timestamps we're using (but I don't think we've switched to that yet).

It _might_ be possible to the `numberOfRowsFromQuery` that I originally added to infer an offset, but that will need confirming.

I don't think you hit this scenario in your tests as it looks like you hit the page size eventually from one call to `query.execute`.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
nkovacsx: Yeah I thought about this too thanks!

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
nkovacsx: Yeah same, this is more of a "Java-ish" code with the break/return in the while loop but even though I spent some time thinking about this couldn't really come up with a better solution.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/external/events/FindAllExternalEventFactory.kt
```diff
@@ -33,9 +33,12 @@ class FindAllExternalEventFactory: ExternalEventFactory<FindAllParameters, Entit
     }
 
     override fun resumeWith(checkpoint: FlowCheckpoint, response: EntityResponse): ResultSetExecutor.Results {
+        val numberOfRowsFromQuery = response.metadata.items.single { it.key == "numberOfRowsFromQuery" }.value.toInt()
```
lankydan, line 36: I think we'lll need a comment somewhere about why this works.

Let me write down what I think happens based on this and @nkovacsx can you confirm that is what you wanted?

- We do the querying and repeat querying on the database worker side.
- Each time we re-query (to fill the limit/page depending on in-memory filtering) we increase `numberOfRowsFromQuery`. So if we re-query x2 and the limit is 5 then we'll get a value for `numberOfRowsFromQuery` between 5-10. 10 == there could be more, <10 == there isn't another page.
- We then update the `hasNext` and `newOffset` based on this information.
- We can use `numberOfRowsFromQuery` as the `newOffset` because:
  - If we got less than the limit then we stop querying.
  - If we got equal to the limit then we updated the new offset by an amount == to the limit.
  - If we re-queried to fill the page, then we'll get a value higher than the limit, so setting `numberOfRowsFromQuery` as the offset makes sense because we want to jump to the last row we actually queried.

I don't think `hasNext = limit in 1..numberOfRowsFromQuery` is completely right. If we re-query to fill the page and reach the end of the actual data in the database, then we'll request another page from the flow side because we don't know that we reached the end. So say the limit is 5, we re-query to fill the page after filtering, which gives us a total of 9 rows returned in total (before filtering). `hasNext = 5 in 1..9` is true. But we know that there is no more data.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/external/events/FindAllExternalEventFactory.kt
```diff
@@ -33,9 +33,11 @@ class FindAllExternalEventFactory: ExternalEventFactory<FindAllParameters, Entit
...
```
lankydan, line 36: We can probably revert the changes to this file based on the recent changes.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/external/events/NamedQueryExternalEventFactory.kt
```diff
@@ -34,9 +34,11 @@ class NamedQueryExternalEventFactory : ExternalEventFactory<NamedQueryParameters
     }
 
     override fun resumeWith(checkpoint: FlowCheckpoint, response: EntityResponse): ResultSetExecutor.Results {
+        val numberOfRowsFromQuery = response.metadata.items.single { it.key == "numberOfRowsFromQuery" }.value.toInt()
```
lankydan, line 37: Can probably revert.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
```diff
@@ -15,6 +15,16 @@ data class ResultSetImpl<R> internal constructor(
     private val resultSetExecutor: ResultSetExecutor<R>
 ) : PagedQuery.ResultSet<R> {
 
+    init {
+        // Ideally this will never happen, but we keep this check in here for safety
+        require(offset >= 0) {
+            "Offset cannot be negative or zero"
+        }
```
lankydan: Check doesn't match the message.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
```diff
@@ -32,8 +42,20 @@ data class ResultSetImpl<R> internal constructor(
             throw NoSuchElementException("The result set has no more pages to query")
         }
         val (serializedResults, numberOfRowsFromQuery) = resultSetExecutor.execute(serializedParameters, offset)
-        hasNext = limit in 1..numberOfRowsFromQuery
-        offset += limit
+
+        // Here have the number of rows we fetched from the database (before filtering)
```
lankydan: Specify "post-filtering in the database worker" or something along those lines.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
```diff
@@ -32,8 +42,20 @@ data class ResultSetImpl<R> internal constructor(
...
+        // then we check if the remainder is 0 when diving the number of rows fetched with the limit
+        // This will:
+        // 1. yield false only if we don't have a next page to fetch
+        // Examples:
+        // `limit` = 5, `numberOfRowsFromQuery` = 9, `hasNext = 9 % 5 == 0` => `hasNext = 4 == 0` == false
+        // `limit` = 5, `numberOfRowsFromQuery` = 4, `hasNext = 4 % 5 == 0` => `hasNext = 4 == 0` == false
+        // 2. yield true if we can't know for sure if there's a next page
+        // Example:
+        // `limit` = 5, `numberOfRowsFromQuery` = 5, `hasNext = 5 % 5 == 0` => `hasNext =  0 == 0` == true
```
lankydan: Show the 10 example as it shows we've covered and understood that scenario.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/ResultSetImpl.kt
```diff
@@ -32,8 +42,20 @@ data class ResultSetImpl<R> internal constructor(
...
+        hasNext = numberOfRowsFromQuery != 0  // If there are no rows left there are no more pages to fetch
```
lankydan, line 46: I think put the condition on one line, I don't think we need the comment since the condition is pretty clear now.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -72,11 +73,78 @@ class VaultNamedQueryExecutorImpl(
             deserializedParams
         )?.results?.filterNotNull() ?: filteredAndTransformedResults
 
-        // Return the filtered/transformed/collected (if present) result to the caller
+        // Return the filtered/transformed/collected (if present) result and the offset to continue the paging from to the caller
         return EntityResponse.newBuilder()
             .setResults(collectedResults.map { ByteBuffer.wrap(serializationService.serialize(it).bytes) })
-            .setMetadata(KeyValuePairList(listOf(KeyValuePair("numberOfRowsFromQuery", contractStateResults.size.toString()))))
-            .build()
+            .setMetadata(KeyValuePairList(listOf(
```
lankydan, line 79: Can we calculate the `numberOfRowsFromQuery` into a variable then pass to here. Hopefully we can then keep the formatting of this line tidy.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -72,11 +73,78 @@ class VaultNamedQueryExecutorImpl(
...
+
+            // Apply filters (if there's any) and add the filtered results to the final result set
+            filteredResults.addAll(contractStateResults.filter {
+                vaultNamedQuery.filter.filter(it, deserializedParams)
+            })
+        }
+
+        return Pair(
+            filteredResults.take(request.limit), // Make sure we don't return more than the limit even if the list has overflown
+            internalOffset - (filteredResults.size - request.limit) // subtract the extra size we removed from the results from the offset
```
lankydan: We could do `.coerseAtLeast(request.offset)` if we're worried about making any mistakes. This also means the other early return could be removed (but I'm fine leaving it in).

We don't have to do it, I'll let you make the decision.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -212,4 +221,9 @@ class VaultNamedQueryExecutorImpl(
             throw NullParameterException(msg)
         }
     }
+
+    data class FilterResult(
```
lankydan: make this class private.

lankydan (commented):
One small comment but otherwise looks good to me.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
nkovacsx: good point, modified

lankydan (approved):

LWogan (approved):
LGTM

nkovacsx:
We discussed the new logic and @lankydan 's outstanding comments on a huddle a week or so ago. Marking them as addressed now.

PR #4473, opened 2023-08-16
CORE-13935 added check if ExternalEventStateType is OK when resending external event, set it to RETRY and increment the retry count

ciaran238:
### Context
Currently it is possible that an external event can be retried indefinitely if the external event response doesn't include an error (This is the behaviour being seen in the ticket, when there are no DbWorkers). This means the external event state status doesn't get set to RETRY and the retry count doesn't get incremented.

### Solution
In the `getEventToSend` method I have implemented a check when the event is being retried, to see if the external event state type is `OK`. If it is `OK`, it gets set to `RETRY`, an appropriate exception is set and the retry count is incremented.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
@@ -3,6 +3,7 @@ package net.corda.flow.external.events.impl
 import net.corda.avro.serialization.CordaAvroDeserializer
 import net.corda.avro.serialization.CordaAvroSerializationFactory
 import net.corda.avro.serialization.CordaAvroSerializer
+import net.corda.data.ExceptionEnvelope
 import net.corda.data.flow.event.external.ExternalEvent
 import net.corda.data.flow.event.external.ExternalEventResponse
 import net.corda.data.flow.event.external.ExternalEventResponseErrorType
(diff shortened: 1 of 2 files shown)


components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
```diff
@@ -169,6 +170,12 @@ class ExternalEventManagerImpl(
                     "Resending external event request ${externalEventState.requestId} which was last sent at " +
                             externalEventState.eventToSend.timestamp
                 }
+                if (externalEventState.status.type == ExternalEventStateType.OK) {
+                    externalEventState.status.exception =
```
LWogan, line 174: What would happen if we left externalEventState.status.exception unset/set to null?

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
```diff
@@ -169,6 +170,12 @@ class ExternalEventManagerImpl(
                     "Resending external event request ${externalEventState.requestId} which was last sent at " +
                             externalEventState.eventToSend.timestamp
                 }
+                if (externalEventState.status.type == ExternalEventStateType.OK) {
```
LWogan, line 173: We'd need a unit test change to go along with this

LWogan (commented):
Going into RETRY mode && incrementing the retry counter when we are retrying make sense to me!

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
ciaran238: It throws a null pointer exception in some places where that value is read

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
```diff
@@ -169,6 +170,12 @@ class ExternalEventManagerImpl(
...
+                    externalEventState.status.exception =
+                        ExceptionEnvelope("Unreceived Error", "Event retried but did not receive an error, Ensure all workers are running")
```
LWogan: The text reads a little funky. For errorType strings its better to to have a string with no spaces. Something like "NoResponse". And adjust the error message to "Received no response for external event request, ensure all workers are running"

ciaran238: 👍

LWogan (dismissed):
LGTM!

LWogan (approved):

PR #4475, opened 2023-08-16
CORE-14685: Driver DSL now remembers its virtual nodes to support lookups.

chrisr3:
Store the `VirtualNodeInfo` objects that `startNodes` returns, so that `DriverDSL` can support a look-up API:
- `VirtualNodeInfo nodeFor(String groupName, MemberX500Name member)`
- `Map<MemberX500Name, VirtualNodeInfo> nodesFor(String groupName)`

This also allows us to simplify the `group()` API.

### DIFF
--- a/testing/driver/driver-testing/src/integrationTest/java/net/corda/testing/driver/tests/FlowDriverJavaTest.java
+++ b/testing/driver/driver-testing/src/integrationTest/java/net/corda/testing/driver/tests/FlowDriverJavaTest.java
@@ -1,0 +1,104 @@
+package net.corda.testing.driver.tests;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.r3.corda.demo.mandelbrot.CalculateBlockFlow;
+import com.r3.corda.demo.mandelbrot.RequestMessage;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Stream;
+import net.corda.testing.driver.AllTestsDriver;
+import net.corda.testing.driver.DriverNodes;
+import net.corda.v5.base.types.MemberX500Name;
+import net.corda.virtualnode.VirtualNodeInfo;
+import org.jetbrains.annotations.NotNull;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.Timeout;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.ArgumentsProvider;
+import org.junit.jupiter.params.provider.ArgumentsSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
+
+@Timeout(value = 5, unit = MINUTES)
+@TestInstance(PER_CLASS)
+class FlowDriverJavaTest {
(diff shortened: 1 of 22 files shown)


testing/driver/driver-testing/src/integrationTest/java/net/corda/testing/driver/tests/FlowDriverJavaTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+@Timeout(value= 5, unit = MINUTES)
+@TestInstance(PER_CLASS)
+class FlowDriverJavaTest {
+    private static final Logger LOGGER = LoggerFactory.getLogger(FlowDriverJavaTest.class);
+    private static final String MANDELBROT_APPLICATION = "mandelbrot";
+
+    private static final MemberX500Name ALICE = MemberX500Name.parse("CN=Alice, OU=Testing, O=R3, L=London, C=GB");
+    private static final MemberX500Name BOB = MemberX500Name.parse("CN=Bob, OU=Testing, O=R3, L=San Francisco, C=US");
+    private final Set<VirtualNodeInfo> virtualNodes = new LinkedHashSet<>();
```
tlawson3: I'm not sure I see the value of having this, as you can't really retrieve a given virtual node.

In this example, you add the started virtual nodes to it, but then never use it

testing/driver/driver-testing/src/integrationTest/java/net/corda/testing/driver/tests/FlowDriverJavaTest.java
```diff
@@ -0,0 +1,103 @@
...
+            LOGGER.info("VirtualNode({}): {}", vNode.getHoldingIdentity().getX500Name(), vNode)
+        );
+        LOGGER.info("{} and {} started successfully", ALICE.getCommonName(), BOB.getCommonName());
+    }
+
+    @ParameterizedTest
+    @ArgumentsSource(RequestProvider.class)
+    void testMandelbrotFlow(RequestMessage request) {
+        final String aliceResult = driver.let(dsl ->
+            dsl.runFlow(dsl.nodeFor(MANDELBROT_APPLICATION, ALICE), CalculateBlockFlow.class, () ->
```
tlawson3: I understand the need to specify which CorDapp you're using when you're running in runtime-os, but in a single CorDapp project there would only be the current "workflows" CorDapp, so is it possible to omit this information each time I want to run a flow?

What if we were to have access to the DSL `node(virtualNodeInfo: VirtualNodeInfo` and pass a given `VirtualNodeInfo` ?

testing/driver/driver-testing/src/integrationTest/java/net/corda/testing/driver/tests/FlowDriverJavaTest.java
chrisr3: Good thinking, Batman! :bat:

testing/driver/driver-testing/src/integrationTest/java/net/corda/testing/driver/tests/FlowDriverJavaTest.java
chrisr3: The problem is that the DSL has no idea how many CorDapps may exist on the classpath, and the "unique key" for each `VirtualNodeInfo` must always be ("X500Name", "application group name"). We must also identify the correct `VirtualNodeInfo` in order to run a flow in the first place, and so the goal of the `nodeFor()` API is to allow people to identify their vNode without needing to rummage through the results of `startNodes()`.

Of course, we _could_ do something like:
```java
driver.let(dsl ->
    dsl.group("workflow", group -> {
        group.runFlow(ALICE, MyFlow.class, () -> jsonMapper.writeValueAsString(request))
    })
)
```
except that `driver.let()` can still only return a single value.

> What if we were to have access to the DSL node(virtualNodeInfo: VirtualNodeInfo and pass a given VirtualNodeInfo ?

Isn't that effectively what we already have? Of course, if you _know_ you only have a single CorDapp on your classpath then the list returned by `startNodes()` shouldn't need much searching.

tlawson3 (dismissed):

seanbrereton (approved):
lgtm - blt

yarivus (commented):
thx @chrisr3

PR #4476, opened 2023-08-16
CORE-16330 - Scheduler Processor

driessamyn:
- Hibernate mappings for the Scheduler DB table plus the Scheduler DB API.
- Implementation of the Scheduler and Schedule.
- Implementation of a Scheduler Processor.
- Integration of the Scheduler Processor with the Combined worker.

Associated API PR: https://github.com/corda/corda-api/pull/1218

**NOTE: this has processor has not yet been integrated with the DB Worker as when run currently it will generate warnings due to the hardcoded test task schedule to publish to a non-existing `telephone` topic. However integrating should be exactly like the changes in the Combined worker, and pretty straight forward hopefully.**


ronanbrowne (dismissed):
BLT files - LGTM , although I presume not fully ready as API version is still an alpha

driessamyn:
> BLT files - LGTM , although I presume not fully ready as API version is still an alpha

Thanks @ronanbrowne , yes, I was hoping to get small changes merged, but I'm continuing to add bits now so I'm not held up. I'll ping you when ready.

driessamyn:
Validation when run in the combined worker:

```
2023-08-18 16:14:40.855 [lifecycle-coordinator-2] ERROR net.corda.messagebus.db.persistence.DBAccess {} - Error while trying to get topic partition map. Transaction has been rolled back.
net.corda.messaging.api.exception.CordaMessageAPIFatalException: Cannot find topic telephone
	at net.corda.messagebus.db.persistence.DBAccess$getTopicPartitionMapFor$topicEntry$1.invoke(DBAccess.kt:123) ~[?:?]
```

(test schedule uses a non-existing topic called `telephone`).

processors/scheduler-processor/src/main/kotlin/net/corda/processors/scheduler/impl/SchedulerProcessorImpl.kt
```diff
@@ -0,0 +1,143 @@
...
+        // TODO - CORE-16331: plug in config
+        //::configurationReadService,
+    )
+    private val lifecycleCoordinator =
+        coordinatorFactory.createCoordinator<SchedulerProcessorImpl>(dependentComponents, ::eventHandler)
+
+    // now just hardcoding schedulers here until CORE-16331 is picked up, when we should take this from config
+    private val schedules = listOf(
+        // example schedule, delete when we have a real one
+        Schedule("say-hello", 60, "telephone"),
```
LWogan: Looks like these are being sent to the bus and failing in the combined worker tests as the db bus has no telephone topic

driessamyn: Thanks @LWogan .
This was was a warning previously, but this was indeed the reason for the failure.
For now I've commented out the schedule.

.run/Combined Worker Local.run.xml
```diff
@@ -4,7 +4,8 @@
     <option name="VM_PARAMETERS" value="-Dco.paralleluniverse.fibers.verifyInstrumentation=true" />
     <option name="PROGRAM_PARAMETERS" value="--instance-id=0 -mbus.busType=DATABASE -spassphrase=password -ssalt=salt -spassphrase=password -ssalt=salt -ddatabase.user=user -ddatabase.pass=password -ddatabase.jdbc.url=jdbc:postgresql://localhost:5432/cordacluster -ddatabase.jdbc.directory=$ProjectFileDir$/applications/workers/release/combined-worker/drivers -rtls.crt.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/server.crt -rtls.key.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/server.key -rtls.ca.crt.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/ca-chain-bundle.crt" />
     <option name="WORKING_DIRECTORY" value="$ProjectFileDir$" />
-    <option name="ALTERNATIVE_JRE_PATH" />
+    <option name="ALTERNATIVE_JRE_PATH_ENABLED" value="true" />
+    <option name="ALTERNATIVE_JRE_PATH" value="azul-17" />
```
mbrkic-r3: Is this needed since it was working so far without setting this value? Could this specific value cause issue for other developers (considering also different OS-es)?

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
```diff
@@ -189,6 +192,7 @@ class CombinedWorker @Activate constructor(
         restProcessor.start(config)
         linkManagerProcessor.start(config)
         gatewayProcessor.start(config)
+        schedulerProcessor.start(config)
```
mbrkic-r3, line 195: It should be stopped as well (in the `shutdown()`)

components/scheduler/src/main/kotlin/net/corda/components/scheduler/impl/SchedulerEventHandler.kt
```diff
@@ -0,0 +1,101 @@
...
+        )
+    }
+
+    private fun onRegistrationStatusChangeEvent(event: RegistrationStatusChangeEvent, coordinator: LifecycleCoordinator) {
+        if (event.status == LifecycleStatus.UP) {
+            logger.info("Starting task scheduling")
+            coordinator.updateStatus(LifecycleStatus.UP)
+            triggerAndScheduleNext(coordinator)
+        } else {
+            // TODO Revise below actions in case of an error from sub services (DOWN vs ERROR)
```
mbrkic-r3: Was this revised, could we remove TODO?

.run/Combined Worker Local.run.xml
driessamyn: Sorry. I committed this by accident. Undid the change now.

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
driessamyn: Added. Thanks!

components/scheduler/src/main/kotlin/net/corda/components/scheduler/impl/SchedulerEventHandler.kt
driessamyn: Removed.

libs/scheduler/scheduler-datamodel/src/main/kotlin/net/corda/libs/scheduler/datamodel/SchedulerLockImpl.kt
```diff
@@ -0,0 +1,33 @@
...
+import net.corda.libs.scheduler.datamodel.db.TaskSchedulerLog
+import net.corda.libs.scheduler.datamodel.db.TaskSchedulerLogEntityRepository
+import java.time.temporal.ChronoUnit
+import javax.persistence.EntityManager
+
+class SchedulerLockImpl(
+    override val taskName: String,
+    override val schedulerId: String,
+    private val em: EntityManager,
+    private val logEntityRepository: TaskSchedulerLogEntityRepository = TaskSchedulerLogEntityRepository()
```
mbrkic-r3, line 12: Repositories should depend on domain objects, not the other way around. We should be able to create domain objects without reaching to infrastructure level. We should be able to retrieve domain objects using different implementations of repositories (i.e. not bind them to specific implementation e.g. by depending on EntityManager).

driessamyn: I'm not sure I understand.
Is your issue with the name or the pattern?

This class represents a lock on the scheduler which can either be updated or just released.

If I injected a `get or initialise log` function instead of a repository. Would that an improvement? If so, why?
Happy to fix up naming, but what is structurally wrong about this class? I think it adheres to all SOLID principals.

mbrkic-r3: Assuming this is a domain entity and not a service, my issue is with the pattern. From DDD point of view entities should not reference repositories (here are few links discussing that: [link1](https://stackoverflow.com/questions/5694241/ddd-the-rule-that-entities-cant-access-repositories-directly), [link2](https://stackoverflow.com/questions/14020196/domain-driven-design-and-ioc-dependency-injection/), [link3](https://stackoverflow.com/questions/827670/is-it-ok-for-entities-to-access-repositories), [link4](https://stackoverflow.com/questions/14175931/do-you-inject-a-repository-into-domain-objects)).

In general, injecting repositories into entities violates Single Responsibility Principle because it adds to them responsibility of communication with infrastructure.

[Here](https://enterprisecraftsmanship.com/posts/domain-model-purity-completeness/) is comparison of different approaches, each has some advantages an disadvantages. It also addresses your question about injecting function instead of repository.

To me this looks like a better candidate for a domain service that would have functions `acquireLock` and `releaseLock` and that would be stateless.

components/scheduler/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+    implementation 'net.corda:corda-config-schema'
+
+    implementation platform("net.corda:corda-api:$cordaApiVersion")
+    implementation 'org.jetbrains.kotlin:kotlin-osgi-bundle'
+
+    implementation project(":components:configuration:configuration-read-service")
+    implementation project(":components:db:db-connection-manager")
+    implementation project(":libs:configuration:configuration-core")
+    implementation project(':libs:lifecycle:lifecycle')
+    implementation project(":libs:lifecycle:lifecycle-impl")
```
mbrkic-r3: Could this be `runtimeOnly`?

components/scheduler/src/main/kotlin/net/corda/components/scheduler/impl/SchedulerEventHandler.kt
```diff
@@ -0,0 +1,100 @@
...
+                publisher.publish(schedule.taskName, schedule.scheduleTriggerTopic)
+            } else {
+                logger.debug { "Skipping publishing task scheduler for ${schedule.taskName} " +
+                        "because it has only been ${it.secondsSinceLastScheduledTrigger} " +
+                        "since the last trigger." }
+            }
+        }
+        scheduleNext(coordinator)
+    } catch (e: Throwable) {
+        logger.warn("Task scheduling for ${schedule.taskName} failed. Terminating Scheduler", e)
```
mbrkic-r3, line 77: This would also happen when DB is temporarily unreachable and would completely stop scheduling, right? Could we make it more resilient, e.g. if error is recoverable (e.g. temporary issue with network or database not ready due to restart) we continue retrying (after some time)?

libs/scheduler/scheduler-datamodel/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+    implementation 'javax.persistence:javax.persistence-api'
+    implementation 'net.corda:corda-db-schema'
+    implementation 'org.jetbrains.kotlin:kotlin-osgi-bundle'
+    implementation project(':libs:db:db-core')
+
+    testImplementation "org.mockito:mockito-core:$mockitoVersion"
+    testImplementation "org.mockito.kotlin:mockito-kotlin:$mockitoKotlinVersion"
+
+    integrationTestImplementation project(':libs:db:db-admin')
+    integrationTestImplementation project(':libs:db:db-admin-impl')
```
mbrkic-r3, line 21: Could "*-impl" dependencies be  `integrationTestRuntimeOnly `?

libs/scheduler/scheduler-datamodel/src/main/kotlin/net/corda/libs/scheduler/datamodel/db/internal/TaskSchedulerLogEntity.kt
```diff
@@ -0,0 +1,67 @@
...
+    override var schedulerId: String,
+
+    // Updated by the DB.
+    @Column(name = "last_scheduled", insertable = false, updatable = false)
+    override var lastScheduled: Instant = Instant.MIN,
+
+    // not managed by Hibernate, but still set as part of the get Query.
+    // NOTE: using util.Date because that is what Hibernate maps CURRENT_TIMESTAMP to (java.sql.Timestamp).
+    @Transient
+    var dbNow: java.util.Date = java.util.Date(Long.MIN_VALUE)
```
mbrkic-r3, line 47: It might be better that current time is provided by JRE instead of DB, so it's consistent with the rest of the application.

libs/scheduler/scheduler-datamodel/src/main/kotlin/net/corda/libs/scheduler/datamodel/SchedulerLock.kt
```diff
@@ -0,0 +1,28 @@
+package net.corda.libs.scheduler.datamodel
+
+/**
+ * Represents a lock on a given Task, allowing to find out when the last time was one was triggered.
+ * This lock must be released by calling `updateAndRelease` or `release`.
```
mbrkic-r3: There is no `release` function (also mentioned in the next sentence).

libs/scheduler/scheduler-datamodel/src/main/kotlin/net/corda/libs/scheduler/datamodel/SchedulerLockImpl.kt
driessamyn: So I'm still not really sure what exactly the problem is with the pattern.
I.e. why is this an anti-pattern, or what are the problems with the appproach?
I don't think our code base generally follows a DDD approach, so why is this relevant?
As far as I can see, the class has a single responsibility, is easy to understand and test.

I could change this to a service that has an explicit acquire and release lock, but in my opinion it would just make the API more error prone.
Given that this is a closable object, the useage pattern (i.e. always close/release the lock) is easily implied from the structure.

components/scheduler/src/main/kotlin/net/corda/components/scheduler/impl/SchedulerEventHandler.kt
driessamyn: Yes, but I think that is correct because, if I understand the lifecycle coordinator correctly, the DB connection manager will go down (once there is a DB issue), then signal back up when it recovers.

libs/scheduler/scheduler-datamodel/src/main/kotlin/net/corda/libs/scheduler/datamodel/db/internal/TaskSchedulerLogEntity.kt
driessamyn: This is as per design.
I specifically chose to use the DB clock to avoid issues with clock drift. 

Given that this is quite time sensitive (e.g. seconds rather than minutes hours), if the timestamp would be updated buy the application, any clock drift  between the different hosts could result in triggers being missed.
Given that we have a central resource here, I don't see any downsides in using it.

libs/scheduler/scheduler-datamodel/src/main/kotlin/net/corda/libs/scheduler/datamodel/SchedulerLock.kt
driessamyn: Good spot. This was left over from an earlier implementation where I had an explicit release, but I don't think there is any point.
Updated.

libs/scheduler/scheduler-datamodel/build.gradle
driessamyn: `db-admin-impl` is needed for the liquibase migration as part of the test.
`db-orm-impl` is needed to create an EMF instance.


(this integration test isn't and doesn't need to be, an OSGi test)

components/scheduler/src/main/kotlin/net/corda/components/scheduler/impl/SchedulerEventHandler.kt
mbrkic-r3: I was not aware we added periodical DB connection checks to DB connection manager. so I guess this covers it.

libs/scheduler/scheduler-datamodel/build.gradle
mbrkic-r3: Is my understanding wrong that `integrationTestRuntimeOnly` would add it to runtime classpath when running the test?

driessamyn: Yes, but we need it in the compile config too as we make a reference to the implementation classes because we don't have access to the OSGi DI.

libs/scheduler/scheduler-datamodel/src/main/kotlin/net/corda/libs/scheduler/datamodel/db/internal/TaskSchedulerLogEntity.kt
mbrkic-r3: Hosts should be NTP synchronized anyway (to avoid all kinds of time related issues).

libs/scheduler/scheduler-datamodel/src/main/kotlin/net/corda/libs/scheduler/datamodel/SchedulerLockImpl.kt
mbrkic-r3: This class is pretty simple so not a big deal here, but if this becomes an example followed by other domain entities, which may be more complex, that would have multiple repositories injected, handle transactions in constructors, I think it will soon become clear that due to this pattern such classes have multiple responsibilities, are hard to test and make our code hard to manage.

libs/scheduler/scheduler-datamodel/src/main/kotlin/net/corda/libs/scheduler/datamodel/db/internal/TaskSchedulerLogEntity.kt
driessamyn: I agree it would be best practice, but I don't think we state in our system requirements that they should.
I don't really see a downside in using the DB time in this case though.

mbrkic-r3 (dismissed):
LGTM

wzur-r3 (approved):
LGTM for Gradle configuration files

mbrkic-r3 (approved):

JamesHR3 (approved):

PR #4477, opened 2023-08-16
CORE-16266: Allow the flow processor to run the fiber multiple times as part of the same event execution

JamesHR3:
**Problem Description**
The flow engine can in some circumstances publish events back to its own input topic. This happens if the user code is executed and it is discovered that all the required information to run the flow again is present in the checkpoint after the fiber has suspended. This causes a large number of spurious events on the `flow.event` topic, which could be avoided by just running the fiber again immediately.

This is the first part of the solution, which allows the flow pipeline to run the fiber multiple times as part of processing the same event. The second part will take the `Wakeup` event out the system entirely.

**Solution**
Adjust the flow pipeline to repeatedly evaluate the checkpoint after executing the user code and doing any post-processing on the outputs. In cases where the user code can immediately run again, this process will ensure that that happens.

As part of this solution, the flow pipeline has been split into two classes. Longer term it is probably worth splitting the pre-processing step out into its own class and removing the full pipeline class entirely (instead chaining the classes for the individual steps in the `FlowEventProcessorImpl`), but that is out of scope for this PR. The new class is responsible for executing the user code, doing any processing of the checkpoint on output, and establishing whether the flow is eligible to be run again.

- Move the stages responsible for executing user code and handling any results to their own class.
- Add extra logic to the flow runner to prevent repeatedly starting a flow if it needs to run multiple times on a start event.

**Testing**
Testing is confined to getting the unit and end to end tests to pass. A later PR will ensure that no `Wakeup` events remain in the system once other changes have been committed. (Note that this PR does not remove `Wakeup` events directly - it just makes this possible.)

Note that the integration tests have been temporarily disabled. Fixing these is likely to be a significant effort, and a future PR will also require integration changes. Therefore all integration test fixes will be carried out in this second PR. This PR is targeted at a feature branch for this reason.

**Other changes**
No other changes.


components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowExecutionPipelineStage.kt
```diff
@@ -0,0 +1,106 @@
...
+) {
+
+    private companion object {
+        private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass.name)
+    }
+
+    fun runFlow(context: FlowEventContext<Any>, timeout: Long) : FlowEventContext<Any> {
+        var currentContext = context
+        var continuation = flowReady(context)
+        while (continuation != FlowContinuation.Continue) {
```
jujoramos, line 49: Nitpick, and probably not to be addressed by this `PR`, but this condition is quite counter intuitive. I had to go and check what `FlowContinuation.Continue` represents to be able to understand why the code checks that continuation is literally different from `Continue` to enter the `while` loop 😓:


```
sealed class FlowContinuation {

    /**
     * [Continue] specifies that a flow should continue receiving and processing events and should not run or resume its fiber.
     */
    object Continue : FlowContinuation()
...
}
```

jujoramos (approved):

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowExecutionPipelineStage.kt
JamesHR3: I also think `Continue` is a bit misleading. Maybe it should be something like `Skip`? Thoughts welcome.

jujoramos: What about `Suspend`?, wouldn't it make sense as that's what we'll actually do once the flow can't proceed without waiting for something else to happen?.

sprintengine: `runOrContinue()` could also be renamed `runOrSkip`

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowExecutionPipelineStage.kt
```diff
@@ -0,0 +1,106 @@
...
+            else -> throw FlowFatalException("Invalid ${FlowIORequest::class.java.simpleName} returned from flow fiber")
+        }
+    }
+
+    private fun updateContext(output: FlowIORequest<*>, context: FlowEventContext<Any>) : FlowEventContext<Any> {
+        // This [uncheckedCast] is required to remove the [out] from the [FlowRequestHandler] that is extracted from the map.
+        // The [out] cannot be kept as it leaks onto the [FlowRequestHandler] interface eventually leading to code that cannot compile.
+        @Suppress("unchecked_cast")
+        val handler = flowRequestHandlers[output::class.java] as? FlowRequestHandler<FlowIORequest<*>>
+            ?: throw FlowFatalException("${output::class.qualifiedName} does not have an associated flow request handler")
```
sprintengine, line 122: I assume external events always stop the loop? There is 1 case where a ExternalEventWaitingForHandler can return a `FlowContinuation.Run`, and that's when it has already received the result. My assumption here is we don't care because we will never make external requests from within this executeFlow loop, so we'll never get any external event results within our loop, thus we don't need to make "checkpoint.externalEventState" a list.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/FlowEventPipeline.kt
```diff
@@ -31,34 +30,13 @@ interface FlowEventPipeline {
...
-     * @return The updated pipeline instance.
-     */
-    fun setWaitingFor(): FlowEventPipeline
-
-    /**
-     * Performs [FlowIORequest] post-processing on the pipeline.
-     *
-     * @return The updated pipeline instance.
+     * @param timeout Time in milliseconds to wait before timing out running the fiber. Note that this applies per
+     * fiber execution, so this function may block for longer than this if the fiber is run multiple times.
```
LWogan, line 36: This has the potential to cause issues if not handled with care. The timeout is a calculated value: 0.75 of the message pattern processor timeout. This is to ensure the flow engine handles the timeout internally rather than the message processor. if multiple flow executions are executed in one processor loop then this restriction is no longer applied,

conal-smith (approved):
LGTM

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/impl/FlowEventPipelineImplTest.kt
```diff
@@ -192,7 +151,14 @@ class FlowEventPipelineImplTest {
...
 
-    @ParameterizedTest(name = "runOrContinue runs a flow when {0} is returned by the FlowWaitingForHandler with suspend result")
+    @Test
+    fun `execute flow invokes the execute flow pipeline stage`() {
+        val pipeline = buildPipeline()
+        pipeline.executeFlow(RUN_OR_CONTINUE_TIMEOUT)
+        verify(mockFlowExecutionPipelineStage).runFlow(inputContext, RUN_OR_CONTINUE_TIMEOUT)
+    }
+
+/*    @ParameterizedTest(name = "runOrContinue runs a flow when {0} is returned by the FlowWaitingForHandler with suspend result")
```
LWogan: is this meant to be commented out? will it be put back after a subsequent change?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/FlowEventPipeline.kt
LWogan: After speaking about it in a call this is a known limitation to be refactored later when integrating with the new subscription

LWogan (commented):
Looks ok to me so far.

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/impl/FlowEventPipelineImplTest.kt
JamesHR3: I intended to use this to make sure I'd covered the same cases in the unit tests for the new class I added. Good catch though, I don't think I've got a complete list of tests yet. I'll go ahead and tidy up the other set of unit tests to make sure the same cases are covered, and then remove this.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowExecutionPipelineStage.kt
JamesHR3: We won't publish anything until we exit the loop, so I don't think we can get a response back from something generated in this flow execution. If we received the right event as part of the pre-process step then we'll execute the flow of course, but that should only happen on the first iteration I think.

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/impl/FlowEventPipelineImplTest.kt
JamesHR3: I've removed this now.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/FlowEventPipeline.kt
JamesHR3: I've left a comment about this in the code.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowExecutionPipelineStage.kt
JamesHR3: I'm going to leave this out of this PR, but I'm happy to raise a follow-up for this.

jujoramos: 👍

LWogan (approved):
LGTM

PR #4481, opened 2023-08-17
CORE-14685: Only cache VirtualNodeInfo for members, not notaries.

chrisr3:
The `VirtualNodeInfo`'s unique key is actually the `HoldingIdentity`, but the DSL only knows about the X500Name and the application group name. This is fine for the members, but doesn't work for the notaries. But this is also fine, because the DSL isn't supposed to be handling the Notary nodes anyway.


tlawson3 (approved):

yarivus (commented):
thx @chrisr3

PR #4482, opened 2023-08-17
CORE-16184: SessionClose refactor

emilybowe:
SessionClose refactor as part of CORE-15757/flow-session-refactor 
Now a requireClose boolean value is used. 
If a party sets requireClose to be true it will wait to receive a close (but will not ScheduleClose) and will cleanup.
If requireClose is false then after a configurable amount of time it will ScheduleClose, cleanup, and will not wait for Acks.


components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
```diff
@@ -30,7 +34,7 @@ class CloseSessionsRequestHandler @Activate constructor(
 
     override fun getUpdatedWaitingFor(context: FlowEventContext<Any>, request: FlowIORequest.CloseSessions): WaitingFor {
         val sessionsToClose = try {
-            getSessionsToClose(context.checkpoint, request)
+            getSessionsToClose(context.checkpoint, request, initiatingIdentity = true)
```
LWogan: You don't know at this point if it was the initiator or initiated who called close here. We can suspend for both cases.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
```diff
@@ -67,9 +88,63 @@ class CloseSessionsRequestHandler @Activate constructor(
...
+         * else if i am initiating party i.e i dont end with INITIATED_SESSION_ID_SUFFIX
+        if requireClose == false
+        - dont return this session in the list
+        else if requireClose == true
+        - do return this session in the list
+
+         && filter out sessions in status ERROR or CLOSED
+         */
+
+        val sessions = request.sessions.toList()
```
LWogan: the kdoc comments i added above here is logic you need to apply to each sessionId in this list.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
```diff
@@ -67,9 +88,63 @@ class CloseSessionsRequestHandler @Activate constructor(
...
+        val statusToFilterOut = setOf(SessionStateType.ERROR, SessionStateType.CLOSED)
+
+        val sessionsInErrorOrClosed = flowSessionManager.getSessionsWithStatuses(checkpoint, sessions, statusToFilterOut)
+        sessions - sessionsInErrorOrClosed.map { it.sessionId }
```
LWogan: This operation isn't assigned to any field so it has no affect. Needs to be something like:
val filteredSessionsByStatus = sessions - sessionsInErrorOrClosed.map { it.sessionId }

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
```diff
@@ -67,9 +88,63 @@ class CloseSessionsRequestHandler @Activate constructor(
...
+
+        val requiredCloseSessions = flowSessionManager.getSessionsByRequireClose(checkpoint, sessions, requireClose = true)
+        sessions - requiredCloseSessions.map { it.sessionId }
```
LWogan: same problem here.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
```diff
@@ -67,9 +88,63 @@ class CloseSessionsRequestHandler @Activate constructor(
...
+
+        return if(initiatingIdentity) {
```
LWogan: Same comment as earlier. We need to work out if we are the initiating identity per sessionId in this handler. In reality due to how the flow engin/session APIs are written we are always one or the other for all sessions in the list, but we shouldn't code based on that here.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/sessions/FlowSessionManager.kt
```diff
@@ -237,4 +237,35 @@ interface FlowSessionManager {
...
+     * If false is provided get the sessions for which requireClose is false
+     * If true is provided get the sessions for which requireClose is true
+     * @param checkpoint The checkpoint to check states within
+     * @param sessionIds The sessions to check
+     * @return The list of states that have requireClose set to the value provided
+     */
+    fun getSessionsByRequireClose(
+        checkpoint: FlowCheckpoint,
+        sessionIds: List<String>,
+        requireClose: Boolean
```
LWogan: requireClose will be set per session, so you can't have a boolean on this API.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/sessions/FlowSessionManager.kt
```diff
@@ -237,4 +237,35 @@ interface FlowSessionManager {
...
+     *
+     * @param checkpoint The flow's [FlowCheckpoint].
+     * @param sessionIds The session ids to check the status of.
+     * @param status The acceptable status the sessions can have.
+     *
+     * @return A list of [SessionState]s that have a [SessionState.status] of [status].
+     *
+     * @throws []FlowSessionStateException] If a session does not exist within the flow's [FlowCheckpoint].
+     */
+    fun getSessionsWithStatus(
```
LWogan: Theres already a getSessionsWithStatus in this class

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/processor/SessionCloseProcessorSend.kt
```diff
@@ -103,28 +117,29 @@ class SessionCloseProcessorSend(
...
-            logger.trace { "Currently in CLOSING. Changing to WAIT_FOR_FINAL_ACK. nextSeqNum: $nextSeqNum, adding this event to send " +
-                    "${sessionEvent.sequenceNum}, $sessionId" }
-            // Doesn't go to closed until ack received
-            sessionState.apply {
-                sendEventsState.lastProcessedSequenceNum = nextSeqNum
-                sendEventsState.undeliveredMessages = sessionState.sendEventsState.undeliveredMessages.plus(sessionEvent)
+    ) : SessionState {
+        val requireClose = sessionState.requireClose
+        var status = sessionState.status
+        return if (isInitiatedIdentity(sessionEvent) && status !== SessionStateType.CLOSED) {
```
LWogan: this class looks correct!

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/processor/SessionCloseProcessorSend.kt
```diff
@@ -103,28 +117,29 @@ class SessionCloseProcessorSend(
...
+                status = SessionStateType.CLOSED
+                sessionState
             }
-        }
-        else -> {
-            val errorMessage = "Tried to send SessionClose on key $key and sessionId $sessionId, session status is " +
-                    "$currentState. Current SessionState: $sessionState."
+        } else {
+            val errorMessage =
+                "Tried to send SessionClose on key $key and sessionId $sessionId, session status is " +
```
LWogan: need to update this error to say something about the fact its trying to send a close when requireClose is set to false. It shouldn't ever happen in reality.

LWogan (commented):
I think a few of my comments are in classes that are still in progress but wanted to make sure we course correct early.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
emilybowe: I added the -= if you're happy with that:
e.g.         `sessions -=  sessionsInErrorOrClosed.map { it.sessionId }`

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/sessions/FlowSessionManager.kt
emilybowe: The boolean is there so I know if when I'm filtering the list I'm interested in returning for requireClose = true or requireClose = false
Is that use case ok?

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/processor/SessionCloseProcessorSend.kt
```diff
@@ -32,6 +33,19 @@ class SessionCloseProcessorSend(
         private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
     }
 
+    /**
```
LWogan: remove these pseudo code comments throughout the code if they're implemented

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/processor/SessionCloseProcessorSend.kt
```diff
@@ -103,28 +117,29 @@ class SessionCloseProcessorSend(
...
+                status = SessionStateType.CLOSED
+                sessionState
             }
-        }
-        else -> {
-            val errorMessage = "Tried to send SessionClose on key $key and sessionId $sessionId, session status is " +
-                    "$currentState. Current SessionState: $sessionState."
+        } else {
+            val errorMessage =
+                "Tried to send SessionClose when requireClose is set to false. Key: $key sessionId: $sessionId, session status is " +
```
LWogan: this error message isn't correct based on the If condition.  The issue would be an initiating party trying to send a close message which is not allowed in the protocol or the status is not correct for sending a close

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/processor/SessionCloseProcessorSend.kt
```diff
@@ -103,28 +117,29 @@ class SessionCloseProcessorSend(
...
+        var status = sessionState.status
+        return if (isInitiatedIdentity(sessionEvent) && status !== SessionStateType.CLOSED) {
+            if (requireClose) {
+                sessionState.apply {
+                    logger.trace {
+                        "Sending SessionClose and setting status to CLOSED. nextSeqNum: $nextSeqNum, adding this event to send " +
+                                "${sessionEvent.sequenceNum}, $sessionId"
+                    }
+                    status = SessionStateType.CLOSED
+                    sendEventsState.lastProcessedSequenceNum = nextSeqNum
```
LWogan, line 97: We also need to set session event sequence number

sessionEvent.sequenceNum = nextSeqNum

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/processor/SessionCloseProcessorSend.kt
LWogan: I think if the status was ERROR this would also be a condition where we shouldn't try send an close.
Change the second condition to something like 

``status !in listOf(SessionStateType.ERROR, SessionStateType.CLOSED)``

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
```diff
@@ -49,10 +50,48 @@ class CloseSessionsRequestHandler @Activate constructor(
...
+                if(requireCloseFalse.isNotEmpty()) {
+                    flowSessionManager.updateStatus(checkpoint, requireCloseFalse, SessionStateType.CLOSED)
+                }
+
+                if (requireCloseTrue.isNotEmpty()) {
+                    flowSessionManager.updateStatus(checkpoint, requireCloseFalse, SessionStateType.CLOSING)
+                }
+            }
+
+            //TODO - @Lorcan Can I delete this?
```
LWogan: no

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
```diff
@@ -49,10 +50,48 @@ class CloseSessionsRequestHandler @Activate constructor(
...
+            val initiatingSessions = initiatingAndInitiated.first
+            val initiatedSessions = initiatingAndInitiated.second
+
+            // if I am the initiated party i.e ends with INITIATED_SESSION_ID_SUFFIX
+            if(initiatedSessions.isNotEmpty()) {
+                checkpoint.putSessionStates(flowSessionManager.sendCloseMessages(checkpoint, initiatedSessions, Instant.now()))
+            }
+
+            /**
+             *  if i am initiating party i.e i dont end with INITIATED_SESSION_ID_SUFFIX
```
LWogan: remove this comment

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
```diff
@@ -67,9 +106,50 @@ class CloseSessionsRequestHandler @Activate constructor(
...
+        //filter out RequireClose false
+        val requireCloseSessions = flowSessionManager.getRequireCloseTrueAndFalse(checkpoint, filteredSessions)
+
+        return requireCloseSessions.first
+    }
+
+    private fun getSessionsToClose(request: FlowIORequest.CloseSessions): List<String> {
+        /**
+         * else if i am initiating party i.e i dont end with INITIATED_SESSION_ID_SUFFIX
+        if requireClose == false
```
LWogan: remove this

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/sessions/impl/Constants.kt
```diff
@@ -0,0 +1,4 @@
+package net.corda.flow.pipeline.sessions.impl
+
+const val INITIATED = "initiated"
```
LWogan: are these being used?

LWogan (commented):
Some initial comments

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
```diff
@@ -49,10 +50,48 @@ class CloseSessionsRequestHandler @Activate constructor(
...
+
+                val requireCloseTrueAndFalse = flowSessionManager.getRequireCloseTrueAndFalse(checkpoint, initiatingSessions)
+                val requireCloseTrue = requireCloseTrueAndFalse.first
+                val requireCloseFalse = requireCloseTrueAndFalse.second
+
+                if(requireCloseFalse.isNotEmpty()) {
+                    flowSessionManager.updateStatus(checkpoint, requireCloseFalse, SessionStateType.CLOSED)
+                }
+
+                if (requireCloseTrue.isNotEmpty()) {
```
LWogan: If the initiating party calls close() and the they have already received a close from the other side, then the current status will be CLOSING. In that case we would want to update those session states to CLOSED instead of CLOSING. 

The logic in this class is starting to get a bit confusing with multiple calls to the flow session manager.

Given this logic needs to be replicated to SubFlowFinishedRequestHandler, i think moving the logic from line 78 -> 91 to its own service might help make the code more maintainable

emilybowe: handled here: closeSessionsAlreadyInClosing(checkpoint, initiatingSessions)
and I've split out the logic

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/service/CloseSessionService.kt
```diff
@@ -0,0 +1,74 @@
...
+
+            val requireCloseTrueAndFalse = flowSessionManager.getRequireCloseTrueAndFalse(checkpoint, initiatingSessions)
+            val requireCloseTrue = requireCloseTrueAndFalse.first
+            val requireCloseFalse = requireCloseTrueAndFalse.second
+
+            if (requireCloseFalse.isNotEmpty()) {
+                flowSessionManager.updateStatus(checkpoint, requireCloseFalse, SessionStateType.CLOSED)
+            }
+
+            if (requireCloseTrue.isNotEmpty()) {
```
LWogan: in closeSessionsAlreadyInClosing you set sessions that were in CLOSING to CLOSED, but then here you would reset them back to CLOSING. If the status is ERROR/CLOSED you dont want to update the session state as its already  terminated. 


Try find the cleanest way possible to handle the possible scenarios for initiating flows are
- status == ERROR/CLOSED-> do nothing, sessions are already terminated
- requireClose == true && status == CREATED/CONFIRMED -> go to CLOSING
- requireClose == true && status == CLOSING -> go to CLOSED
- requireClose == false -> go to CLOSED.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/service/CloseSessionService.kt
```diff
@@ -0,0 +1,74 @@
...
+import net.corda.data.flow.state.session.SessionStateType
+import net.corda.flow.pipeline.sessions.FlowSessionManager
+import net.corda.flow.state.FlowCheckpoint
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.Reference
+import java.time.Instant
+
+@Component(service = [CloseSessionService::class])
+class CloseSessionService @Activate constructor(
```
LWogan, line 12: no kdocs on the class or any of the methods

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/service/CloseSessionService.kt
```diff
@@ -0,0 +1,74 @@
...
+        //filter out states
+        val filteredSessions = filterOutErrorAndClosed(checkpoint, initiatingSessions)
+
+        //filter out RequireClose false
+        val requireCloseSessions = flowSessionManager.getRequireCloseTrueAndFalse(checkpoint, filteredSessions)
+
+        return requireCloseSessions.first
+    }
+
+    fun getSessionsForPostProcess(sessionsToClose: List<String>, checkpoint: FlowCheckpoint) {
```
LWogan: this is called get but doesn't return anything. it should be named something like handleCloseForSessions or event closeSessions. (im not the best with names)

LWogan (commented):
Couple more comments

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/helper/Utils.kt
```diff
@@ -0,0 +1,15 @@
+package net.corda.flow.pipeline.handlers.requests.helper
+
+import net.corda.session.manager.Constants
+
+/**
+ * check if an event is sent by Initiating identity
+ * @param sessionEvent input session event to get identity info from
+ * @return [Boolean] is this event sent by the Initiating identity
+ */
+fun isInitiatingIdentity(sessionId: String): Boolean {
```
LWogan: Theres a common helper method in :libs:flows:flow-utils which you can use instead of adding this new method. Beware it is the inverse of your function, returning true for initiated, not initiating. so youll need to account for that

fun isInitiatedParty(sessionEvent: SessionEvent) : Boolean {
    return sessionEvent.sessionId.contains(INITIATED_SESSION_ID_SUFFIX)
}

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
```diff
@@ -116,8 +117,12 @@ class FlowSessionImpl(
 
     @Suspendable
     override fun close() {
-        fiber.suspend(FlowIORequest.CloseSessions(setOf(sourceSessionId)))
-        log.trace { "Closing session: $sourceSessionId" }
+        val flowCheckpoint = flowFiberService.getExecutingFiber().getExecutionContext().flowCheckpoint
+        val sessionState = flowCheckpoint.getSessionState(sourceSessionId)
+        if(checkNotNull(sessionState!!.status) != SessionStateType.CLOSED) {
```
LWogan: If session state was null here you'd throw a null pointer exception so the checkNotNull isn't doing anything functional. 

It should be impossible for the session state to be null at this point. If it is null and we suspend the close request handler will detect it and throw an appropriate exception. So im ok with checking status against closed where status can be null i.e

```
val sessionState = flowFiberService.getExecutingFiber().getExecutionContext().flowCheckpoint.getSessionState(sourceSessionId)
        if(sessionState?.status != SessionStateType.CLOSED) {
            fiber.suspend(FlowIORequest.CloseSessions(setOf(sourceSessionId)))
            log.trace { "Closing session: $sourceSessionId" }
        }

File: components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/sessions/impl/FlowSessionManagerImplTest.kt [hunk 23]
```diff
@@ -52,6 +53,7 @@ import java.nio.ByteBuffer
 import java.time.Instant
 import java.util.stream.Stream
 
+@Disabled //todo - core-15757
```
LWogan: you need to enable these tests

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/processor/SessionCloseProcessorSend.kt
```diff
@@ -85,37 +80,44 @@ class SessionCloseProcessorSend(
...
-            logger.trace { "Currently in CLOSING. Changing to WAIT_FOR_FINAL_ACK. nextSeqNum: $nextSeqNum, adding this event to send " +
-                    "${sessionEvent.sequenceNum}, $sessionId" }
-            // Doesn't go to closed until ack received
-            sessionState.apply {
-                sendEventsState.lastProcessedSequenceNum = nextSeqNum
-                sendEventsState.undeliveredMessages = sessionState.sendEventsState.undeliveredMessages.plus(sessionEvent)
+        } else {
+            val status = sessionState.status
+            val errorMessage: String = if (status in listOf(SessionStateType.ERROR, SessionStateType.CLOSED)) {
+                "Tried to send SessionClose when status is not correct for sending a close. " +
```
LWogan: I think you can write a single error message that conveys all the necessary information required for for the log without the conditionals.

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/processor/helper/SessionProcessorHelper.kt
```diff
@@ -140,4 +141,13 @@ fun setErrorState(
...
         )
     }
+}
+
+/**
+ * check if an event is sent by Initiated or Initiating identity
+ * @param sessionEvent input session event to get identity info from
+ * @return [Boolean] is this event sent by the Initiated identity
+ */
+fun isInitiatedIdentity(sessionEvent: SessionEvent): Boolean {
```
LWogan: delete and reuse the new flow-utils method

LWogan (commented):
Couple more comments. Address all the comments before you merge from release again as there will be a headache merge to do when the flow session refactor pr is merged today. Having your branch ready to go makes it marginally easier for you.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/helper/Utils.kt
emilybowe: do you object to this:

```
package net.corda.flow.utils

const val INITIATED_SESSION_ID_SUFFIX = "-INITIATED"

fun isInitiatedIdentity(sessionId: String) : Boolean {
    return sessionId.contains(INITIATED_SESSION_ID_SUFFIX)
}

fun isInitiatingIdentity(sessionId: String) : Boolean {
    return !sessionId.contains(INITIATED_SESSION_ID_SUFFIX)
}
```

I think it's easier to read isInitiated / isInitiating when I was writing this but I'll change it if you feel it's better to remove the method. I did change all the usages so they are all coming from that one utils file.

LWogan: yeah thats ok with me.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/service/CloseSessionService.kt
```diff
@@ -0,0 +1,94 @@
...
+        return requireCloseTrueSessions
+    }
+
+    /**
+     * Executes the close logic for the given [sessions].
+     * @param checkpoint The checkpoint.
+     * @param sessions List of sessions IDs.
+     * @return List of sessions that are not CLOSED or ERROR
+     */
+    fun handleCloseForSessions(checkpoint: FlowCheckpoint, sessions: List<String>): List<String> {
```
LWogan: This method does not need to return anything. it just updates the session states.

LWogan (commented):
One more comment and then i think we're good to merge!

LWogan (approved):
LGTM

PR #4446, opened 2023-08-11
CORE-15757 WIP:some initial changes to setup the feature branch with avro objects to begin development

LWogan:
https://github.com/corda/platform-eng-design/pull/560

https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4552491009/Corda+Scalable+Topology+Design.

https://github.com/corda/corda-api/pull/1208

Theres a lot of logic i intend to change here. so please contact via slack rather than leave comments

Other WIP PRS:
Session init refactor + miscellaneous tickets: https://github.com/corda/corda-runtime-os/pull/4483
Session Close refactor: https://github.com/corda/corda-runtime-os/pull/4482

PR #4483, opened 2023-08-17
CORE-16174 session init refactor, removal of heartbeating and resends

LWogan:
Design: https://github.com/corda/platform-eng-design/blob/lorcan/CORE-15873/session-redesign/core/corda-5/corda-5.0/flow-worker/flow-sessions-new.md

PR is suitable for review,

https://github.com/corda/corda-api/pull/1212

The code style/structure/patterns have been minimally changed.

SessionInit nullable field added to SessionData object. Populated for each data message sent to the initiated party until they respond with a data message. Initiator sessionState status will remain in CREATED until they receive a data message back (CONFIRMED)
We cannot be sure which SessionData will arrive first to the counterparty due to no order guarantees in the p2p layer. SessionInit object contains the information needed to start a flow on the other side, as well as the user and platform context data. This allows the checkpoint to be created regardless of which data message arrives first. 

Removed restriction from initiating flow whereby they could not proceed past the first send without receiving something back.

contextSessionProperties is piggybacked onto existing data messages when it is not explicitly requested. It is moved to the SessionEvent object rather than the SessionConfirm object.

SessionInit & Confirm exchange is kept for when session.getCounterpartyInfo is called before any messages are exchanged.

Pure Acks removed

Heartbeating is removed

Session timeout logic left in place. Additional work required to integrate with the scheduler

Message resends logic removed. 

session.close() is "hacked" to immediately move the session state to CLOSED. proper implementation to be done as part of https://github.com/corda/corda-runtime-os/pull/4482

Additional changes were made in this initial pr to add the new initiateFlow API to define whether a close message is requested. https://github.com/corda/corda-runtime-os/pull/4446
I think ill update the SessionState to remove the explicit requireClose field and leave it as an entry in the session context map.


ronanbrowne (approved):
LGTM (Jenkins file / properties file changes only reviewed )

Jenkinsfile
```diff
@@ -8,8 +8,8 @@ cordaPipelineKubernetesAgent(
     publishPreTestImage: true,
     publishHelmChart: true,
     e2eTestName: 'corda-runtime-os-e2e-tests',
-    runE2eTests: false,
-    combinedWorkere2eTests: false,
```
JamesHR3, line 12: Presumably you'll need to revert these changes?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
```diff
@@ -85,6 +85,18 @@ class SessionEventHandler @Activate constructor(
         return context
     }
 
+    private fun getSessionInitIfPresent(sessionEvent: SessionEvent?): SessionInit? {
```
JamesHR3, line 88: Think you can rewrite this to be a little clearer:

```kotlin
return when (val payload = sessionEvent?.payload) {
    is SessionInit -> payload
    is SessionData -> payload.sessionInit
    else -> null
}
```

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
```diff
@@ -46,7 +44,7 @@ class CloseSessionsRequestHandler @Activate constructor(
     }
 
     override fun postProcess(context: FlowEventContext<Any>, request: FlowIORequest.CloseSessions): FlowEventContext<Any> {
-        val checkpoint = context.checkpoint
+        /*val checkpoint = context.checkpoint
```
JamesHR3: Should this commented out code be removed?

testing/cpbs/test-cordapp/src/main/kotlin/com/r3/corda/testing/smoketests/flow/InitiatedSmokeTestFlow.kt
```diff
@@ -5,12 +5,18 @@ import net.corda.v5.application.flows.ResponderFlow
...
 @InitiatedBy(protocol = "smoke-test-protocol")
 class InitiatedSmokeTestFlow : ResponderFlow {
 
+    private companion object {
+        private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
+    }
+
     @Suspendable
     override fun call(session: FlowSession) {
+        log.info("Starting initiated flow")
```
JamesHR3, line 19: Do we still want this log?

JamesHR3 (commented):
I've assumed the TODOs are for a follow up in this feature branch. If not they should probably be resolved too.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
```diff
@@ -85,6 +85,18 @@ class SessionEventHandler @Activate constructor(
...
+        if (sessionEvent == null) return null
+        val payload = sessionEvent.payload
+        return if (payload is SessionInit) {
+            payload
```
sprintengine: so SessionInit will still be sent when flow calls session.getCounterpartyInfo before any messages have been exchanged, would the counterparty info fit better in a SessionData so we can entirely remove SessionInit from existing on its own? (probably in a future change)

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionInitHelper.kt
```diff
@@ -0,0 +1,88 @@
...
+        sessionInit: SessionInit,
+        flowConfig: SmartConfig,
+        instant: Instant
+    ): SessionInitOutputs {
+        return if (messageDirection == MessageDirection.INBOUND) {
+            val flowId = generateFlowId()
+            sessionInit.flowId = flowId
+            SessionInitOutputs(
+                flowId,
+                Record(Schemas.Flow.FLOW_EVENT_TOPIC, flowId, FlowEvent(flowId, sessionEvent))
```
owenstanford: for consistency is there a reason not to use the record factory for all records?

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionEventExecutor.kt
```diff
@@ -22,21 +24,34 @@ class SessionEventExecutor(
...
     private companion object {
         private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
     }
 
     override fun execute(): FlowMapperResult {
-        return if (flowMapperState == null) {
+        val payload = sessionEvent.payload
+        val sessionInit = getInitPayload(payload)
+        return if (flowMapperState == null && sessionInit != null) {
+            sessionInitHelper.processSessionInit(sessionEvent, sessionInit, flowConfig, instant)
```
sprintengine: Should we also be recording metrics here for session inits? I.e. move 

```
            CordaMetrics.Metric.FlowMapperCreationCount.builder()
                .withTag(CordaMetrics.Tag.FlowEvent, sessionInit::class.java.name)
                .build().increment()
```
into the sessionInitHelper

Jenkinsfile
LWogan: this change is the revert. they were disabled in the feature branch

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
LWogan: yeah, theres a comment below. Separate task to do this properly.

testing/cpbs/test-cordapp/src/main/kotlin/com/r3/corda/testing/smoketests/flow/InitiatedSmokeTestFlow.kt
LWogan: Yeah i think its useful. there's logs on the initiating side but none this side.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
LWogan: CounterpartyInfo is stored in sessionContextProperties which is now set at SessionEvent level. 

It wouldnt fit in SessionData as you can ask for CounterpartyInfo when no data is being exchanged,

Theres other fields in SessionInit which are only sent 1 direction from initiator to initiated so I think we'll need that avro object going forward

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionInitHelper.kt
LWogan: I've only just merged the record factory pr into this one. I may need to ammend this.

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionEventExecutor.kt
LWogan: good spot!

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionInitHelper.kt
LWogan: updated

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionEventExecutor.kt
```diff
@@ -22,21 +24,34 @@ class SessionEventExecutor(
...
+    private val instant: Instant,
+    private val sessionInitHelper: SessionInitHelper
+) : FlowMapperEventExecutor {
 
     private companion object {
         private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
     }
 
     override fun execute(): FlowMapperResult {
-        return if (flowMapperState == null) {
```
simon-johnson-r3, line 33: This is a matter of style preference I suppose, but Kotlin would let you write this more concisely, and specifically without the double check for `flowMapperState` being null which adds a bit of extra challenge to read:

```
override fun execute() = if (flowMapperState == null) {
  getInitPayload(sessionEvent.payload)?.let { sessionInit->
    sessionInitHelper.processSessionInit(sessionEvent, sessionInit, flowConfig, instant)
  } ?: handleNullState()
} else {
  processOtherSessionEvents(flowMapperState)
}

private fun getInitPayload(payload: Any) = when (payload) {
    is SessionInit -> payload
    is SessionData -> payload.sessionInit
    else -> null
}
```

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionEventExecutor.kt
```diff
@@ -79,6 +94,7 @@ class SessionEventExecutor(
                 FlowMapperResult(null, listOf())
             }
             FlowMapperStateType.CLOSING -> {
+                //todo - CORE-15757/ CORE-16184
```
simon-johnson-r3: There are lots of debates about whether todo's in code are are good idea. Firstly I would say this should be `// TODO` which highlights the comment correctly in the IDE, but anyway, given this isn't really anything to do with this PR but instead a pointer in the code for some other developer to follow for some different ticket later on - perhaps just add a github link to this code line under the relevant tickets instead and delete the comment?

Not only would that alleviate the need for a discussion, but it would help those developers track the code down without having to rely on this.

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionInitHelper.kt
```diff
@@ -0,0 +1,85 @@
...
+import net.corda.metrics.CordaMetrics
+import java.time.Instant
+
+/**
+ * Helper class to process session events which contain a SessionInit field/payload
+ */
+class SessionInitHelper(private val recordFactory: RecordFactory) {
+
+    /**
+     * Process a [sessionEvent] and [sessionInit] payload to producer a flow mapper state
```
simon-johnson-r3: typo: producer -> produce

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionInitHelper.kt
```diff
@@ -0,0 +1,85 @@
...
+import net.corda.flow.mapper.factory.RecordFactory
+import net.corda.libs.configuration.SmartConfig
+import net.corda.messaging.api.records.Record
+import net.corda.metrics.CordaMetrics
+import java.time.Instant
+
+/**
+ * Helper class to process session events which contain a SessionInit field/payload
+ */
+class SessionInitHelper(private val recordFactory: RecordFactory) {
```
simon-johnson-r3: I find it odd to read in the code that a class called `helper` is instantiated. Not least because you instantiate it on behalf of its users and pass it around.

It's used by SessionEventExecutor and SessionInitExecutor and only has 1 public method, and it's only stateful purpose is to store `recordFactory`.

Could you just make `processSessionInit` a publicly accessible (static) function which takes `recordFactory` as a parameter as well, and call it direct in SessionEventExecutor and SessionInitExecutor (passing those the `recordFactory` on construction)?

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionInitHelper.kt
```diff
@@ -0,0 +1,85 @@
...
+                flowConfig,
+                instant
+            )
+
+        return FlowMapperResult(
+            FlowMapperState(flowId, null, FlowMapperStateType.OPEN),
+            listOf(outputRecord)
+        )
+    }
+
```
simon-johnson-r3: Again a minor style thing, but Kotlin allows you reduce functions which only return into aliases like:
```
    private fun getSessionInitOutputs(
        messageDirection: MessageDirection?,
        sessionEvent: SessionEvent,
        sessionInit: SessionInit,
        flowConfig: SmartConfig,
        instant: Instant
    ) = if (messageDirection == MessageDirection.INBOUND) {
```

I would consider this optional though.

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionInitHelper.kt
```diff
@@ -0,0 +1,85 @@
...
+            sessionInit.flowId = null
+
+            SessionInitOutputs(
+                tmpFLowEventKey,
+                recordFactory.forwardEvent(sessionEvent, instant, flowConfig, sessionEvent.messageDirection)
+            )
+        }
+    }
+
+    data class SessionInitOutputs(
```
simon-johnson-r3: should this be private?

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
```diff
@@ -41,38 +40,34 @@ class FlowSessionImpl(
...
+        val counterPartyFlowInfo = getFlowInfoFromSessionContext()
         return if (counterPartyFlowInfo != null) {
             counterPartyFlowInfo
         } else {
             val request = FlowIORequest.CounterPartyFlowInfo(getSessionInfo())
             fiber.suspend(request)
             //If we are able to receive counterparty info this means the session initiation has been completed.
             setSessionConfirmed()
-            getCounterpartySessionContext() ?: throw CordaRuntimeException(
+            getFlowInfoFromSessionContext() ?: throw CordaRuntimeException(
```
simon-johnson-r3, line 50: Am I reading this correctly, you seem to have called this function already to set `counterPartyFlowInfo`?

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
```diff
@@ -41,38 +40,34 @@ class FlowSessionImpl(
...
-        } else {
-            null
-        }
+        val sessionProperties = sessionState?.sessionProperties
+        return getFlowInfoFromSessionProps(sessionProperties)
     }
 
-    private fun getFlowInfoFromSessionProps(counterpartySessionProperties: KeyValuePairList): FlowInfo {
-        val props = KeyValueStore(counterpartySessionProperties)
+    private fun getFlowInfoFromSessionProps(sessionProperties: KeyValuePairList?): FlowInfo? {
```
simon-johnson-r3: We seem to have evolved to a state where we have 2 helper functions which are only ever called from each other in one place? Perhaps this could be refactored a bit?

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
```diff
@@ -41,38 +40,34 @@ class FlowSessionImpl(
...
+        if (sessionProperties == null) return null
```
simon-johnson-r3: this line is only here because line 61 allows nulls (the other call point already does a null check itself): `val sessionProperties = sessionState?.sessionProperties`

I wonder if the code can be refactored a little to avoid the need for it and make the two points consistently check for null?

My motivation here is that there are multiple return points in this method already.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
```diff
@@ -68,11 +69,10 @@ class SessionEventHandler @Activate constructor(
 
         // Null is returned if duplicate [SessionInit]s are received
         val nextSessionEvent = sessionManager.getNextReceivedEvent(updatedSessionState)
-        val nextSessionPayload = nextSessionEvent?.payload
-
         if (!checkpoint.doesExist) {
-            if (nextSessionPayload is SessionInit) {
-                createInitiatedFlowCheckpoint(context, nextSessionPayload, nextSessionEvent, updatedSessionState)
+            val sessionInit = getSessionInitIfPresent(nextSessionEvent)
```
simon-johnson-r3, line 73: I think this could do with a bit of refactoring. You pass `nextSessionEvent` to a function, and then on the next line ask if it's null. If it was null the function returns null anyway.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
```diff
@@ -113,13 +121,17 @@ class SessionEventHandler @Activate constructor(
...
             when {
-                result.isSuccess -> sendConfirmMessage(
-                    result.getOrNull(),
-                    requestedProtocolName,
-                    initiatorVersionsSupported,
-                    context,
-                    sessionId
-                )
+                result.isSuccess -> {
+                    if (sessionEvent.payload is SessionInit) {
```
simon-johnson-r3, line 125: what if it isn't a SessionInit? Is it correct that case is not handled?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/SubFlowFinishedRequestHandler.kt
```diff
@@ -52,8 +51,16 @@ class SubFlowFinishedRequestHandler @Activate constructor(
         val hasNoSessionsOrAllClosed = try {
             val sessionsToClose = getSessionsToClose(checkpoint, request)
 
+            //TODO - CORE-15757 / CORE-16184 do this properly
```
simon-johnson-r3: I don't think we can let this comment go into the open source repo. This is aside from whether we consider TODO's ok or not, we shouldn't let people believe we've committed code which doesn't work properly. Presumably you actually mean, this code works fine, but a different implementation is perhaps coming later. But that is potentially true of all code.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/SubFlowFinishedRequestHandler.kt
```diff
@@ -52,8 +51,16 @@ class SubFlowFinishedRequestHandler @Activate constructor(
...
             checkpoint.putSessionStates(flowSessionManager.sendCloseMessages(checkpoint, sessionsToClose, Instant.now()))
-
+            val sessionStates = sessionsToClose.mapNotNull { sessionToClose ->
+                checkpoint.sessions.find {
+                    sessionToClose == it.sessionId
+                }
+            }.onEach {
```
simon-johnson-r3: do you need a second loop here, could you not do a find and set the property in a single iteration?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
```diff
@@ -64,7 +62,18 @@ class CloseSessionsRequestHandler @Activate constructor(
             context.copy(outputRecords = context.outputRecords + listOf(record))
         } else {
             context
+        }*/
+
+        //TODO CORE-15757 / CORE-16184 - implement close properly
```
simon-johnson-r3: Same as above, let's not try to erroneously suggest we submit not working code.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/sessions/CloseSessionsRequestHandler.kt
```diff
@@ -64,7 +62,18 @@ class CloseSessionsRequestHandler @Activate constructor(
...
+        val checkpoint = context.checkpoint
+        val sessionsToClose = getSessionsToClose(checkpoint, request)
+        context.checkpoint.sessions.onEach {
+            if (sessionsToClose.contains(it.sessionId))
```
simon-johnson-r3: If you want to split an `if` statement on multiple lines it's dangerous to not include braces. It's a common mistake that the next developer comes along an adds something at line 73 which they think is encapsulated within the `if` and isn't.

Personally I'd put all `if` code in braces to prevent this from ever happening, but at least please put it on one line. If it doesn't fit then revert to braces

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/sessions/impl/FlowSessionManagerImpl.kt
```diff
@@ -78,17 +79,34 @@ class FlowSessionManagerImpl @Activate constructor(
...
         )
     }
 
+    override fun generateSessionState(
+        checkpoint: FlowCheckpoint,
+        sessionId: String,
+        x500Name: MemberX500Name,
+        sessionProperties: KeyValuePairList,
+        instant: Instant
+    ): SessionState {
```
simon-johnson-r3, line 100: Another style choice, but consider using `=`

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/SessionManagerImpl.kt
```diff
@@ -49,14 +52,34 @@ class SessionManagerImpl @Activate constructor(
...
     ): SessionState {
         return sessionEventProcessorFactory.createEventToSendProcessor(key, event, sessionState, instant, maxMsgSize).execute()
     }
 
+    override fun generateSessionState(
+        sessionId: String,
+        contextSessionProperties: KeyValuePairList,
+        counterparty: HoldingIdentity,
+        instant: Instant,
+    ): SessionState {
```
simon-johnson-r3: Consider `=`

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/SessionManagerImpl.kt
```diff
@@ -98,70 +119,93 @@ class SessionManagerImpl @Activate constructor(
         identity: HoldingIdentity,
     ): Pair<SessionState,
             List<SessionEvent>> {
-        val messagesToReturn = getMessagesToSendAndUpdateSendState(sessionState, instant)
+        val sessionEvents = getMessagesToSend(sessionState)
+        val messagesToReturn= handleSessionTimeouts(sessionState, config, instant, sessionEvents, identity)
```
simon-johnson-r3: typo missing space before =

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/SessionManagerImpl.kt
```diff
@@ -98,70 +119,93 @@ class SessionManagerImpl @Activate constructor(
...
      * @return Messages to send
      */
-    private fun getMessagesToSendAndUpdateSendState(
+    private fun getMessagesToSend(
         sessionState: SessionState,
-        instant: Instant,
     ): List<SessionEvent> {
-        val sessionEvents = filterSessionEvents(sessionState, instant)
+        val sessionEvents = sessionState.sendEventsState.undeliveredMessages
         if (sessionEvents.isEmpty()) return emptyList()
```
simon-johnson-r3: in the current incarnation of the implementation if `sessionEvents` is empty, that is an empty list of session events, so that is not a special case and I think this line is not needed?

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/processor/SessionDataProcessorReceive.kt
```diff
@@ -37,12 +40,16 @@ class SessionDataProcessorReceive(
 
     override fun execute(): SessionState {
         val sessionId = sessionEvent.sessionId
-        return if (sessionState == null) {
+        val sessionInit = payload.sessionInit
```
simon-johnson-r3, line 43: I think this code is saying we should handle either a sessionState or a sessionInit, and I think it's the same if you do the simpler:
```
if (sessionState != null) {
  getInboundDataEventResult(sessionState, sessionId)
} else if (sessionInit != null) {
  val newSessionState = sessionInitProcessorReceive.execute()
  getInboundDataEventResult(newSessionState, sessionId)
} else {
  ...
}
```
but please check this!

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/processor/SessionDataProcessorSend.kt
```diff
@@ -108,4 +109,8 @@ class SessionDataProcessorSend(
         (copy.payload as SessionData).payload = chunk
         return copy
     }
+
+    private fun isInitiatedParty(sessionEvent: SessionEvent) : Boolean {
```
simon-johnson-r3: this logic appears in two places, consider making this a public helper is that's simple

simon-johnson-r3 (changes requested):

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionInitExecutor.kt
```diff
@@ -24,10 +19,10 @@ class SessionInitExecutor(
     private val sessionEvent: SessionEvent,
     private val sessionInit: SessionInit,
     private val flowMapperState: FlowMapperState?,
-    private val sessionEventSerializer: CordaAvroSerializer<SessionEvent>,
     private val flowConfig: SmartConfig,
     private val recordFactory: RecordFactory,
-    private val instant: Instant
+    private val instant: Instant,
+    private val sessionInitHelper: SessionInitHelper,
```
JamesHR3: Thinking about it - do you need `SessionInitExecutor` at all? We had this because we wanted quite different handling between session init messages and session data messages. However, we're now trying to do the same handling in both the "normal" session event handler and this session init handler. Can we just use a single session event handler to do this? Does that let you simplify the code in the mapper elsewhere?

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
JamesHR3: The `suspend` in the middle will cause this to return a different result when calling it a second time (assuming the response from the counterparty had the relevant information in).

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowGlobalPostProcessorImpl.kt
```diff
@@ -105,6 +105,7 @@ class FlowGlobalPostProcessorImpl @Activate constructor(
          * If the counterparty doesn't exist in our network, don't send our queued messages yet.
          * If we've also exceeded the [SESSION_MISSING_COUNTERPARTY_TIMEOUT_WINDOW], throw a [FlowPlatformException]
          */
+        //TODO - CORE-15757 should we fail immediately here instead?
```
JamesHR3: Did you make any progress on answering this question?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
```diff
@@ -58,18 +59,25 @@ class FlowRunnerImpl @Activate constructor(
...
             is StartFlow -> startFlow(context, receivedEvent)
             is SessionEvent -> {
-                val payload = receivedEvent.payload
+                val payload = getInitPayload(receivedEvent)
                 if (payload is SessionInit) {
-                    startInitiatedFlow(context, payload)
+                    startInitiatedFlow(context, payload, receivedEvent)
                 } else {
                     resumeFlow(context, flowContinuation)
                 }
```
JamesHR3: Isn't this now effectively a `null` check on `payload`?

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/sessions/impl/FlowSessionImpl.kt
simon-johnson-r3: Ok, the first time I guess we issue a request, from then on it's in the flow checkpoint.

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionEventExecutor.kt
LWogan: im merging to my feature branch, not release so i think this is  matter of personal preference?

The close logic is being worked on by a different developer in a branch from the feature so i added comments to help them locate the places they need to make sure to update more easily. The code is actively changing so I think GitHub links in JIRA tickets could go stale quickly.

Ive updated to TODO from todo.
I agree they shouldn't go into release though.

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionInitHelper.kt
LWogan: Ill make it a proper service and inject it where its required.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
LWogan: if its not a SessionInit, then theres nothing to do here. We only respond with a confirm if weve successfully initialized the checkpoint and the session event payload was a SessionInit. This can only happen when the user calls getCounterpartyInfo before any sends or receives are executed.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/SubFlowFinishedRequestHandler.kt
LWogan: This is going into a feature branch as part of multiple prs. Emily is working on this functionality here https://github.com/corda/corda-runtime-os/pull/4482

removed "do this properly"

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/SubFlowFinishedRequestHandler.kt
LWogan: Updated but note that this is being written here https://github.com/corda/corda-runtime-os/pull/4482. The close handlers are not being worked on in this pr

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowGlobalPostProcessorImpl.kt
LWogan: I'm updating this to fail the flow immediately as we no longer support retries

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionInitExecutor.kt
LWogan: I've refactored this a bit per Simons comments. Its not massively different to what you reviewed though. Ill have a think on it.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
LWogan: Not sure I follow. Its checking to see if this is the first event received (which can be an Init or Data). If it is, extract out the SessionInit object (to be used to get user/platform context properties from)

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/executor/SessionEventExecutor.kt
LWogan: After some discussions yesterday it looks like the close logic changes might not make it into our next release scheduled for friday so im going to update this pr to remove the TODOs

simon-johnson-r3: > im merging to my feature branch, not release so i think this is matter of personal preference?

Sorry I didn't notice that. I was summoned to this review in slack.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
JamesHR3: `getInitPayload` returns `SessionInit?`, so `if (payload is SessionInit)` is now equivalent to `if (payload != null)`, which might be clearer.

simon-johnson-r3 (approved):

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
LWogan: applied

PR #4487, opened 2023-08-17
CORE-14685: Validate members for startNodes(), and throw exception if flow ends without any result.

chrisr3:
- Add extra validation for the DriverDSL's `startNodes` function.
- Throw `FlowErrorException` if we run out of flow events before the flow completes.

### DIFF
--- a/testing/driver/driver-testing/src/test/java/net/corda/testing/driver/tests/DriverJavaTest.java
+++ b/testing/driver/driver-testing/src/test/java/net/corda/testing/driver/tests/DriverJavaTest.java
@@ -1,0 +1,64 @@
+package net.corda.testing.driver.tests;
+
+import java.util.List;
+import java.util.Set;
+import net.corda.testing.driver.DriverNodes;
+import net.corda.testing.driver.EachTestDriver;
+import net.corda.v5.base.types.MemberX500Name;
+import net.corda.virtualnode.VirtualNodeInfo;
+import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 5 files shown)


testing/driver/driver-testing/src/test/java/net/corda/testing/driver/tests/DriverJavaTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.jupiter.api.Timeout;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
+import static java.util.concurrent.TimeUnit.MINUTES;
+
+@Timeout(value = 5, unit = MINUTES)
+@TestInstance(PER_CLASS)
```
tlawson3, line 21: As you're using an `EachTestDriver` do you still need this annotation?

chrisr3: Hmm, interesting question... :thinking:. Probably not "need", but that doesn't mean that `EachTestDriver` shouldn't still _work_...

---
Ah, this annotation allows my `@BeforeAll` method not to be `static`.

yarivus (approved):
LGTM

PR #4491, opened 2023-08-18
CORE-13624 Service side idempotency id for persistence requests

kyriathar:
- uses the deduplication table (idempotency table) to deduplicate `PersistEntities` (coming from `PersistenceService.persist`)
- moves `EntityManager.transaction` around specific `EntityRequest`s to allow more fine-grained error control
- changed `PersistenceExceptionTests` to:
  -  use `BeforeEach` handler as sandbox context needs resetting
  - removed `VirtualNodeService` and used `VirtualNodeLoader` instead in test class, because it seems `VirtualNodeService` creates (in memory) DB with name X but then we would also use `FakeDbConnectionManager` which creares a different (in memory) DB with name Y, so different DBs.

*** Please note *** if merged now this PR it means the deduplication tables will keep growing since no clean up mechanism is yet in place.

api PR: [#1222](https://github.com/corda/corda-api/pull/1222)


components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
```diff
@@ -80,20 +82,26 @@ class PersistenceExceptionTests {
...
+        private const val X500_NAME = "CN=Testing, OU=Application, O=R3, L=London, C=GB"
+        fun generateHoldingIdentity() = createTestHoldingIdentity(X500_NAME, UUID.randomUUID().toString())
     }
 
-    @Suppress("JUnitMalformedDeclaration")
     @RegisterExtension
-    private val sandboxLifecycle = AllTestsLifecycle()
+    private val beforeEachLifecycle = EachTestLifecycle()
+
+//    private lateinit var virtualNodeService: VirtualNodeService
```
simon-johnson-r3: commented out code?

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
```diff
@@ -114,38 +124,42 @@ class PersistenceExceptionTests {
     ) {
         logger.info("Setup test (test Directory: $testDirectory)")
         sandboxSetup.configure(bundleContext, testDirectory)
-        sandboxLifecycle.accept(sandboxSetup) {
-            virtualNode = sandboxSetup.fetchService(timeout = 5000)
+        // the below code block runs before every test
```
simon-johnson-r3, line 132: what is the thinking behind this over assigning them in @BeforeEach ?

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
```diff
@@ -309,21 +319,41 @@ class PersistenceExceptionTests {
...
     }
 
+    @Test
+    fun `should distinguish duplicate persistence request from actual error in persistence request`() {
+        createDogDb()
+        val dogId = UUID.randomUUID()
+        val persistEntitiesRequest = createDogPersistRequest(dogId)
+
+        val record1 = processor.onNext(listOf(Record(TOPIC, UUID.randomUUID().toString(), persistEntitiesRequest)))
+        assertNull(((record1.single().value as FlowEvent).payload as ExternalEventResponse).error)
```
simon-johnson-r3: consider creating a helper like:
`assertIsExternalEventResponseWithoutError(record1)`

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityMessageProcessor.kt
```diff
@@ -54,9 +61,21 @@ class EntityMessageProcessor(
...
     private val responseFactory: ResponseFactory,
     private val payloadCheck: (bytes: ByteBuffer) -> ByteBuffer,
+    private val requestsIdsRepository: RequestsIdsRepository = RequestsIdsRepositoryImpl()
 ) : DurableProcessor<String, EntityRequest> {
     private companion object {
         val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+
+        const val SQL_PK_VIOLATION_CODE = "23505"
+
+        fun isDuplicateRequest(e: PersistenceException): Boolean =
```
simon-johnson-r3: Major: I would strongly recommend not trying to parse an assert message, which is not part of a managed api, and instead simply catch exceptions from each of the persist lines individually

I suggest something like this:
```
is PersistEntities -> {
    val requestId = UUID.fromString(request.flowExternalEventContext.requestId)
    
    val entityResponse = em.transaction {
        withDeduplicationCheck(requestId, it) {
          persistenceServiceInternal.persist(serializationService, it, entityRequest)
        }
    }
    responseFactory.successResponse(
        request.flowExternalEventContext,
        entityResponse
    )
}
// you don't need the try/catch any more
```
With the helper function:

```
private fun withDeduplicationCheck(requestId: UUID, entityManager: EntityManager, block:() -> EntityResponse): EntityResponse {
    try {
        requestsIdsRepository.persist(requestId, entityManager)
    } catch (e: PersistenceException) {
        // A persistence exception thrown in the de-duplication check means we have already performed the operation and
        // can therefore treat the request as successful
        return EntityResponse(emptyList(), KeyValuePairList(emptyList()))
    }
    return block()
}
```

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityMessageProcessor.kt
```diff
@@ -137,53 +156,102 @@ class EntityMessageProcessor(
...
+                    }
+                    responseFactory.successResponse(
+                        request.flowExternalEventContext,
+                        entityResponse
+                    )
+                } catch (e: PersistenceException) {
+                    if (isDuplicateRequest(e)) {
+                        responseFactory.successResponse(
+                            request.flowExternalEventContext,
+                            EntityResponse(emptyList(), KeyValuePairList(emptyList()))
```
simon-johnson-r3: The merge function returns an entity response with information filled in here. Can you check that isn't important?

(I know this will change the nature of my suggestion above, but I think that's still a good solution, just might require some refinement)

components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
```diff
@@ -144,7 +145,14 @@ class EntitySandboxServiceImpl @Activate constructor(
...
 
         // Create the JPA entity set to pass into the EMF
-        val entitiesSet = JpaEntitiesSet.create(virtualNode.vaultDmlConnectionId.toString(), entityClasses)
+        val persistenceRequestIdEntityClass =
+            PersistenceRequestIdEntity::class.java
+
+        val entitiesSet =
+            JpaEntitiesSet.create(
+                virtualNode.vaultDmlConnectionId.toString(),
+                entityClasses + persistenceRequestIdEntityClass
```
simon-johnson-r3: I'm not entirely convinced JPA is helping you here. The INSERT statement for the request Id would be trivial to write, only needs to exists in one place, and means you don't need to worry about registering the Hibernate class like this. That said I'm not strongly opposed to it either.

simon-johnson-r3 (changes requested):

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
kyriathar: So, these used to be in a `@BeforeEach` originally, then changed them [here](https://github.com/corda/corda-runtime-os/pull/4433/files#diff-1e11121a06b09f4342b336f579cd1a1746cdc152c25b9004c091cbc588efcc18R117) to before all thinking that the same `virtualNodeInfo` and `cpkFileHashes` are re-used across tests since we are loading the same CPB. However, that caused an issue where it would cache sandbox context and re-use it, where we need to create new one here for each test, so put them back.

kyriathar: Actually they were in a `@BeforeAll` function but registered in a `net.corda.testing.sandboxes.lifecycle.EachTestLifecycle` which runs before each test.

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
kyriathar: Removed

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
kyriathar: Done

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityMessageProcessor.kt
kyriathar: Good point! Changed to use `EntityManager.flush` instead, so we can get the PK constraint violation on persisting the request id rather than on committing the transaction.

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityMessageProcessor.kt
kyriathar: Actually good catch. Merge duplicate-found path needs aligning with normal one to be made idempotent.

components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
kyriathar: Replaced entity with native query.

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
simon-johnson-r3: Ok it appears `EachTestLifecycle` is a registered JUnit extension which handles the initializing and closing of sanboxes. You need to put this in a method because you OSGi inject the service which it targets.

i think that it's a bit odd to set this up in a `@BeforeAll` but I understand why now at least so let's not worry so much about it.

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityMessageProcessor.kt
```diff
@@ -137,56 +144,114 @@ class EntityMessageProcessor(
...
+        return when (val entityRequest = request.request) {
+            is PersistEntities -> {
+                val requestId = UUID.fromString(request.flowExternalEventContext.requestId)
+                val entityResponse = withDeduplicationCheck(
+                    requestId,
+                    em,
+                    block = {
+                        persistenceServiceInternal.persist(serializationService, it, entityRequest)
+                    },
+                    onDuplication = {
```
simon-johnson-r3, line 154: I would switch wound `onDuplication` and `block`, that way you can do

```
withDeduplicationCheck(...,  onDuplication = {}) {
    persistenceServiceInternal.persist(serializationService, it, entityRequest)
}
```
which reads more naturally for a "with" function.

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityMessageProcessor.kt
```diff
@@ -137,56 +144,114 @@ class EntityMessageProcessor(
...
                     request.flowExternalEventContext,
                     persistenceServiceInternal.deleteEntitiesByIds(
                         serializationService,
                         it,
                         entityRequest
                     )
                 )
+            }
+
+            is MergeEntities -> {
```
simon-johnson-r3, line 185: I thought we were going to drop the dedeuplication check for merges?

Also I'm not sure returning `EntityResponse(entityRequest.entities` will work, don't you need to return the results of the last operation?

simon-johnson-r3 (dismissed):

simon-johnson-r3 (dismissed):

davidcurrie (dismissed):
LGTM

davidcurrie (approved):

simon-johnson-r3 (approved):

PR #4494, opened 2023-08-20
CORE-16389: Do not reuse AMQP SerializationOutput or DeserializationInput objects.

chrisr3:
In Corda 4, the `SerializationScheme` object creates a new `SerializationOutput` / `DeserializationInput` object for each serialization / deserialization request. However, this behaviour has been lost in the translation to Corda 5, with the result that `SerializationService` is currently non-reentrant.

Modify the `SerializationServiceImpl` to create a new `SerializationOutput` or `DeserializationInput` object, in the same way that `SerializationScheme` would have.

In actual practice, we would expect both serialization and deserialization to use the same `SerializerFactory`. However, it is convenient to allow them to be different here for the sake of testing.

### DIFF
--- a/simulator/runtime/src/main/kotlin/net/corda/simulator/runtime/serialization/BaseSerializationService.kt
+++ b/simulator/runtime/src/main/kotlin/net/corda/simulator/runtime/serialization/BaseSerializationService.kt
@@ -8,8 +8,6 @@ import net.corda.internal.serialization.SerializationContextImpl
 import net.corda.internal.serialization.SerializationServiceImpl
 import net.corda.internal.serialization.amqp.DefaultDescriptorBasedSerializerRegistry
 import net.corda.internal.serialization.amqp.DescriptorBasedSerializerRegistry
-import net.corda.internal.serialization.amqp.DeserializationInput
-import net.corda.internal.serialization.amqp.SerializationOutput
 import net.corda.internal.serialization.amqp.SerializerFactory
 import net.corda.internal.serialization.amqp.SerializerFactoryBuilder
 import net.corda.internal.serialization.amqp.amqpMagic
@@ -63,10 +61,12 @@ class BaseSerializationService(
                 schemeMetadata,
                 context.sandboxGroup as SandboxGroup
             )
-            val output = SerializationOutput(factory)
-            val input = DeserializationInput(factory)
 
-            return SerializationServiceImpl(output, input, context)
+            return SerializationServiceImpl(
+                outputFactory = factory,
+                inputFactory = factory,
+                context
+            )
         }
 
         private class SimSandboxGroup(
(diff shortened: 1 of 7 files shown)


simulator/runtime/src/main/kotlin/net/corda/simulator/runtime/serialization/BaseSerializationService.kt
```diff
@@ -63,10 +61,8 @@ class BaseSerializationService(
                 schemeMetadata,
                 context.sandboxGroup as SandboxGroup
             )
-            val output = SerializationOutput(factory)
-            val input = DeserializationInput(factory)
 
-            return SerializationServiceImpl(output, input, context)
+            return SerializationServiceImpl(factory, factory, context)
```
dickon: It would be clearer to use named parameters here, since there's 2 paramters of the same type, and at first it looks weird.

dickon:
It's not clear to me why we need this change, beyond "Corda 4 did it that way". Also, I'm not sure if these objects are expensive to create, and so this might have a noticeable performance impact.

chrisr3:
> It's not clear to me why we need this change, beyond "Corda 4 did it that way". Also, I'm not sure if these objects are expensive to create, and so this might have a noticeable performance impact.

The `DeserializationInput` and `SerializationOutput` objects both contain internal "state" that cannot be shared. This means that `SerializationService` cannot be used concurrently because its functions are non-reentrant. This is a "gotcha".

dickon:
> > It's not clear to me why we need this change, beyond "Corda 4 did it that way". Also, I'm not sure if these objects are expensive to create, and so this might have a noticeable performance impact.
> 
> The `DeserializationInput` and `SerializationOutput` objects both contain internal "state" that cannot be shared. This means that `SerializationService` cannot be used concurrently because its functions are non-reentrant. This is a "gotcha".

Ah yes, as in fact the bug report states, my bad.

testing/test-serialization/src/main/kotlin/net/corda/internal/serialization/amqp/helper/TestSerializationService.kt
```diff
@@ -36,11 +34,9 @@ class TestSerializationService {
             cipherSchemeMetadata: CipherSchemeMetadata
         ) : SerializationService {
             val factory = getTestDefaultFactoryNoEvolution(registerMoreSerializers, cipherSchemeMetadata)
-            val output = SerializationOutput(factory)
-            val input = DeserializationInput(factory)
             val context = testSerializationContext
 
-            return SerializationServiceImpl(output, input, context)
+            return SerializationServiceImpl(factory, factory, context)
```
dickon: Here as well for the named parameters (not blocking).

dickon (dismissed):
LGTM (minor comment on named parameters in one more place, not a blocker)

JamesHR3 (approved):

vlajos (approved):
It looks good from the ledger point of view.

simon-johnson-r3 (approved):

yarivus (commented):
thx @chrisr3

PR #4497, opened 2023-08-21
CORE-16409 Fixing logic in CordaPublisherImplTest to create correct number of threads in pool

ben-millar:
### Context
Tests in `CordaPublisherImplTest.kt` were failing intermittently with the error:
```
org.mockito.exceptions.verification.MoreThanAllowedActualInvocations: 	
Wanted at most 100 times but was 101
```
Upon investigation, I noticed the following logic in the test:
```
val numThreads = 100
val futures = (0..numThreads).map { CompletableFuture<Unit>() }
```
`IntRange` based for loops in Kotlin are, by default, inclusive of the start and end values meaning that `(0..numThreads)` will run for 101 iterations; 0 to 100 inclusive.
### Fix
I changed the creation of the `futures` list as follows, ensuring exactly 100 futures are created:
```
val futures = List(numThreads) { CompletableFuture<Unit>() }
```


mbrkic-r3 (approved):
LGTM

PR #4498, opened 2023-08-21
CORE-9975: Removed unnecessary recoverable and merged functions in sign operation

malachyb:


dickon (approved):
LGTM

dickon:
> LGTM

This is good, I agree the `recoverable` check is pointless, and I'm glad we are collapsing down some ot the many layers of `sign()`.

PR #4499, opened 2023-08-21
CORE-16266: Remove remaining flow Wakeup events

JamesHR3:
**Problem Description**

The flow engine will publish `Wakeup` events on the `flow.event` topic to signal to itself that a flow is eligible to be run again but the current pipeline execution has moved past the point of executing user code. This leads to a lot of extra traffic on the `flow.event` topic, along with message amplification for each input event. To address this, a previous PR allowed the flow pipeline to execute user code up until the point the required data for it to continue is no longer present. However, the events themselves were still being published out the pipeline.

A flow may retry an event if a transient exception is encountered. This typically happens if some data the flow requires has not yet been distributed to the worker it is executing on. If this occurs, the flow will only attempt to replay the original event if it receives a `Wakeup` event. The retry mechanism previously relied on the retry scheduler to do this, and so keeping this feature requires a change.

**Solution**
Remove the instances where `Wakeup` events are published. This primarily involves changing request handlers, which are responsible for dealing with the request the user has made out of flow code.

The solution for retries is to publish the event back to the flow event topic to be processed again later. This mechanism should be sufficient to get past transient problems with running a flow. If this fails enough times, a fatal error will occur as before. Note that this requires no scheduling, and therefore no `Wakeup` event is required. All `Wakeup` events can therefore be removed from the system.

**Testing**
All unit tests and end to end tests passing.

This PR leaves the integration tests disabled. Fixing these will be done separately. Note that some integration test changes are required to keep the code compiling.

**Other changes**
Some changes needed to be made to the patterns library integration tests to keep these compiling too.


JamesHR3:
This branch will need rebasing once other PRs have been merged.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowExecutionPipelineStage.kt
```diff
@@ -0,0 +1,106 @@
...
+import net.corda.flow.pipeline.exceptions.FlowFatalException
+import net.corda.flow.pipeline.handlers.requests.FlowRequestHandler
+import net.corda.flow.pipeline.handlers.waiting.FlowWaitingForHandler
+import net.corda.flow.pipeline.runner.FlowRunner
+import org.slf4j.LoggerFactory
+import java.nio.ByteBuffer
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+
+internal class FlowExecutionPipelineStage(
```
LWogan: A class level kdoc might be nice here.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
```diff
@@ -53,11 +55,18 @@ class FlowRunnerImpl @Activate constructor(
...
+                if (waitingFor is WaitingForStartFlow) {
+                    startFlow(context, receivedEvent)
+                } else {
+                    resumeFlow(context, flowContinuation)
+                }
+            }
             is SessionEvent -> {
                 val payload = receivedEvent.payload
-                if (payload is SessionInit) {
+                if (payload is SessionInit && waitingFor is WaitingForSessionInit) {
```
LWogan: Was this change necessary?

LWogan (commented):
Looks ok to me so far.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowExecutionPipelineStage.kt
JamesHR3: This was introduced in the previous PR and I've addressed it there.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/runner/impl/FlowRunnerImpl.kt
JamesHR3: Yes. We need to decide whether to run based on state rather than event now that we may execute the fiber multiple times for the same event input.

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/ReceiveAcceptanceTest.kt
```diff
@@ -38,7 +37,6 @@ class ReceiveAcceptanceTest : FlowServiceTestBase() {
         @JvmStatic
         fun wakeupAndSessionAck(): Stream<Arguments> {
```
emilybowe: rename? to sessionAck for example

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/SendAcceptanceTest.kt
```diff
@@ -88,11 +88,6 @@ class SendAcceptanceTest : FlowServiceTestBase() {
...
         }
 
-        `when` {
-            wakeupEventReceived(FLOW_ID1)
-                .suspendsWith(FlowIORequest.FlowFailed(Exception()))
-        }
-
         then {
             expectOutputForFlow(FLOW_ID1) {
                 flowResumedWithError<CordaRuntimeException>()
```
emilybowe, line 93: As discussed offline, the 'thens' won't work and the tests will need to be changed more.

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/SubFlowFinishedAcceptanceTest.kt
```diff
@@ -33,7 +33,6 @@ class SubFlowFinishedAcceptanceTest : FlowServiceTestBase() {
         @JvmStatic
         fun wakeupAndSessionAck(): Stream<Arguments> {
```
emilybowe: remove wakeup from function name

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/requests/SendExternalMessageRequestHandler.kt
```diff
@@ -79,16 +78,10 @@ class SendExternalMessageRequestHandler @Activate constructor(
...
-                WakeupEvent()
-            )
-        )
-
-        if (externalMessage != null) {
-            records.add(externalMessage)
+        val records = if (externalMessage != null) {
+            listOf(externalMessage)
+        } else {
+            listOf()
```
emilybowe: nit: maybe emptyList to be deliberate

LWogan (approved):
LGTM

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventPipelineImpl.kt
```diff
@@ -46,7 +46,13 @@ internal class FlowEventPipelineImpl(
          * should re-write the event the pipeline should process the event to be retried, in place of the default
```
emilybowe: this sentence is a little unclear

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/factory/FlowRecordFactory.kt
```diff
@@ -18,7 +17,7 @@ interface FlowRecordFactory {
      * Creates a generic [FlowEvent] record
      *
      * @param flowId The id of the flow generating or receiving the event.
-     * @param payload The instance of the specific flow event to be carried by the record. Valid types are [StartFlow], [Wakeup],
+     * @param payload The instance of the specific flow event to be carried by the record. Valid types are [StartFlow],
      * [SessionEvent].
```
emilybowe: nit:  comma -> and 
e.g. The instance of the specific flow event to be carried by the record. Valid types are [StartFlow] **and** [SessionEvent].

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImplTest.kt
```diff
@@ -100,17 +101,21 @@ class FlowEventExceptionProcessorImplTest {
...
         val flowStatusUpdateRecord = Record("", key, flowStatusUpdate)
+        val flowId = "f1"
+        val flowEventRecord = Record("", flowId, FlowEvent(flowId, ExternalEventResponse()))
+        whenever(flowCheckpoint.flowId).thenReturn(flowId)
         whenever(flowCheckpoint.currentRetryCount).thenReturn(1)
         whenever(flowMessageFactory.createFlowRetryingStatusMessage(flowCheckpoint)).thenReturn(flowStatusUpdate)
         whenever(flowRecordFactory.createFlowStatusRecord(flowStatusUpdate)).thenReturn(flowStatusUpdateRecord)
         whenever(flowCheckpoint.doesExist).thenReturn(true)
         whenever(flowCheckpoint.flowKey).thenReturn(key)
+        whenever(flowRecordFactory.createFlowEventRecord(flowId, Wakeup())).thenReturn(flowEventRecord)
```
emilybowe: is this to check your 'hack' as you mentioned, works for retries? or why is wakeup here?

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/impl/FlowEventPipelineImplTest.kt
```diff
@@ -84,31 +86,25 @@ class FlowEventPipelineImplTest {
...
-                Arguments.of(FlowContinuation.Error(Exception()))
-            )
-        }
-    }
-
     @Test
     fun `eventPreProcessing with no retry calls the event handler`() {
-        val pipeline = buildPipeline()
+        val context = buildFlowEventContext<Any>(checkpoint, ExternalEventResponse("foo"))
+        val pipeline = buildPipeline(context)
```
emilybowe: why not use the default context?

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/runner/FlowRunnerImplTest.kt
```diff
@@ -280,7 +279,7 @@ class FlowRunnerImplTest {
     @Test
     fun `other event types resume existing flow`() {
         val flowContinuation = FlowContinuation.Run()
-        val context = buildFlowEventContext<Any>(flowCheckpoint, Wakeup())
+        val context = buildFlowEventContext<Any>(flowCheckpoint, net.corda.data.flow.event.external.ExternalEventResponse())
```
emilybowe, line 282: do you want this fully qualified?

emilybowe (approved):
LGTM, left some comments

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/ReceiveAcceptanceTest.kt
JamesHR3: Good point. These are disabled for now, so it won't be caught in testing just yet.

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImplTest.kt
JamesHR3: Yeah, it's currently there to test the retry behaviour. However, I think we may take a different approach for this, so expect a change shortly.

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/impl/FlowEventPipelineImplTest.kt
JamesHR3: Good catch, I'd missed a change here. Updated.

components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/runner/FlowRunnerImplTest.kt
JamesHR3: This does need to be fully qualified because we're using two types named the same thing in the same file. `ExternalEventResponse` is also used as a `WaitingFor` state.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventPipelineImpl.kt
JamesHR3: It should now have been removed.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/factory/FlowRecordFactory.kt
JamesHR3: I've updated this to say we could also get `ExternalEventResponses`.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventPipelineImpl.kt
```diff
@@ -41,39 +41,14 @@ internal class FlowEventPipelineImpl(
...
-                // No Wakeup events should be in the system if the flow is not in retrying state, so this is now a fatal
-                // error.
-                log.warn("Flow received a Wakeup event when not in retrying state.")
-                throw FlowFatalException("Flow received a Wakeup event when not in retrying state.")
-            }
-            log.debug(
-                "Flow is in retry state, using retry event " +
-                        "${context.checkpoint.retryEvent.payload::class.qualifiedName} for the pipeline processing."
-            )
-            context.copy(
```
LWogan, line 54: If we're getting rid of this permanently then we should remove the failedEvent field from PipelineState avro and the associated FlowCheckpoint code with it. I'm ok with doing that as a follow up.

LWogan (approved):
LGTM

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventPipelineImpl.kt
JamesHR3: Let's do this as a follow up.

PR #4501, opened 2023-08-21
CORE-16319 state manager changes

conal-smith:
The following [design](https://github.com/corda/platform-eng-design/blob/8cc7b58d8d38cf54da2030e51e5b04c7ed2c62ea/core/corda-5/corda-5.0/flow-worker/state-manager.md) describes the state manager.

This set of changes relies on these [corda-api changes](https://github.com/corda/corda-api/pull/1223).

- StateManagerService component that creates a StateManager implementation based on runtime dependency.
image

- StateManagerService exposes StateManager API.

- This iteration uses the cluster db to run the [liquibase migration scripts](https://github.com/corda/corda-api/pull/1223/files#diff-f181b9ecc6f0504c1e96c099af7ca271ec3ca1be7c453035fde937e1d9c288dc) and this db connection information is provided into the flow worker.


libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -0,0 +1,20 @@
+package net.corda.libs.statemanager.api
+
+/**
+ * The [StateManager] provides functions to manage states to the underlying storage mechanism.
+ */
+interface StateManager : AutoCloseable {
```
sprintengine, line 8: I removed some of the functions here, we can add them back later, with some implementations with the repository.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -0,0 +1,84 @@
...
+    private val serializerFactory: CordaAvroSerializationFactory,
+    private val serializers: MutableMap<Class<*>, CordaAvroSerializer<*>> = ConcurrentHashMap(),
+    private val deserializers: MutableMap<Class<*>, CordaAvroDeserializer<*>> = ConcurrentHashMap(),
+) : StateManager {
+
+    private val objectMapper: ThreadLocal<ObjectMapper> = object : ThreadLocal<ObjectMapper>() {
+        override fun initialValue() = ObjectMapper()
+    }
+
+    override fun <S : Any> get(clazz: Class<S>, keys: Set<String>): Map<String, State<S>> {
```
sprintengine: we must provide a class since many different states could be persisted and we need to have serializers / deserializers for each of them.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -0,0 +1,84 @@
...
+
+class StateManagerImpl(
+    private val stateManagerRepository: StateManagerRepository,
+    private val entityManagerFactory: EntityManagerFactory,
+    private val serializerFactory: CordaAvroSerializationFactory,
+    private val serializers: MutableMap<Class<*>, CordaAvroSerializer<*>> = ConcurrentHashMap(),
+    private val deserializers: MutableMap<Class<*>, CordaAvroDeserializer<*>> = ConcurrentHashMap(),
+) : StateManager {
+
+    private val objectMapper: ThreadLocal<ObjectMapper> = object : ThreadLocal<ObjectMapper>() {
```
sprintengine: Not sure this is entirely necessary, I think ObjectMapper is thread safe anyway.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/dto/StateDto.kt
```diff
@@ -0,0 +1,14 @@
+package net.corda.libs.statemanager.impl.dto
+
+import java.time.Instant
+
+/**
+ * DTO for states in the state manager.
+ */
+data class StateDto(
+    val key: String,
+    val state: ByteArray?,
```
sprintengine: This is nullable, it is different from the design, maybe we will null out the state here like we currently do.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/dto/StateDto.kt
```diff
@@ -0,0 +1,14 @@
...
+    var version: Int,
+    var metadata: String? = null,
+    var modifiedTime: Instant
```
sprintengine: I want the db to generate this timestamp.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/factory/StateManagerFactoryImpl.kt
```diff
@@ -0,0 +1,44 @@
...
+@Component(service = [StateManagerFactory::class])
+class StateManagerFactoryImpl @Activate constructor(
+    @Reference(service = EntityManagerFactoryFactory::class)
+    private val entityManagerFactoryFactory: EntityManagerFactoryFactory,
+    @Reference(service = CordaAvroSerializationFactory::class)
+    private val avroSerializationFactory: CordaAvroSerializationFactory,
+) : StateManagerFactory {
+
+    override fun create(config: SmartConfig): StateManager {
+        val dataSource = PostgresDataSourceFactory().create(
```
sprintengine: This use of Postgres and EntityManagerFactoryFactory feels like it should be abstracted behind another service, but I didn't have time to do this.

This implementation is tied to postgres through this datasource and also the entities. (JSONB is a postgres type). See https://r3-cev.atlassian.net/browse/CORE-16663 .

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/model/v1_0/StateEntity.kt
```diff
@@ -0,0 +1,45 @@
...
+import net.corda.db.schema.DbSchema
+
+@Entity
+@Table(name = DbSchema.STATE_MANAGER_TABLE)
+internal class StateEntity(
+    @Id
+    @Column(name = "key", length = 255)
+    val key: String,
+
+    @Column(name = "state", columnDefinition = "BLOB")
```
sprintengine: `@Lob` causes an exception when reading from this.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateManagerRepositoryImpl.kt
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    // hibernate 5 does not support inserting a String to a jsonb column type out of the box. Native query with casting is also used
+    // in the ledger. It means, however, this does not work with HSQLDB. Thus, integration tests are skipped when using this in-memory
+    // database.
+    private fun persistWithNativeQuery(entityManager: EntityManager, it: StateDto) {
+        entityManager.createNativeQuery(
+            """
+                INSERT INTO ${DbSchema.STATE_MANAGER_TABLE}
```
sprintengine: I'm sure we can persist many entities here in one query. We might not want to do that though. We might also want to return the keys of the states that were successfully persisted, and the error messages of the states that failed to persist. This iteration is simple.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateManagerRepositoryImpl.kt
```diff
@@ -0,0 +1,71 @@
...
+                VALUES (
+                    :key,
+                    :state,
+                    :version,
+                    CAST(:metadata as JSONB),
+                    CURRENT_TIMESTAMP
```
sprintengine: I think this works across a variety of databases, but this implementation is seemingly more and more tied to postgres anyway.

processors/flow-processor/build.gradle
```diff
@@ -71,5 +72,7 @@ dependencies {
     runtimeOnly project(":libs:sandbox-internal")
     runtimeOnly project(":libs:schema-registry:schema-registry-impl")
     runtimeOnly project(":libs:serialization:serialization-kryo")
-
+    runtimeOnly project(":libs:state-manager:state-manager-impl")
+    runtimeOnly project(':libs:db:db-orm-impl')
+    runtimeOnly "com.sun.activation:javax.activation:$activationVersion"
```
sprintengine, line 76: This dependency is required.

processors/flow-processor/src/main/kotlin/net/corda/processors/flow/internal/FlowProcessorImpl.kt
```diff
@@ -57,7 +58,9 @@ class FlowProcessorImpl @Activate constructor(
     @Reference(service = GroupPolicyProvider::class)
     private val groupPolicyProvider: GroupPolicyProvider,
     @Reference(service = MembershipQueryClient::class)
-    private val membershipQueryClient: MembershipQueryClient
+    private val membershipQueryClient: MembershipQueryClient,
+    @Reference(service = StateManagerService::class)
+    private val stateManagerService: StateManagerService,
```
sprintengine, line 57: Here we introduce state manager component into flow processor.

tools/plugins/db-config/src/main/kotlin/net/corda/cli/plugins/dbconfig/Spec.kt
```diff
@@ -99,7 +99,8 @@ class Spec(private val config: SpecConfig = SpecConfig()) : Runnable {
             "net/corda/db/schema/config/db.changelog-master.xml",
             "net/corda/db/schema/messagebus/db.changelog-master.xml",
             "net/corda/db/schema/rbac/db.changelog-master.xml",
-            "net/corda/db/schema/crypto/db.changelog-master.xml"
+            "net/corda/db/schema/crypto/db.changelog-master.xml",
+            "net/corda/db/schema/statemanager/db.changelog-master.xml"
```
sprintengine, line 103: The state manager schema is deployed with the other db schema via cli plugin. This is done by helm

.run/Combined Worker Local.run.xml
```diff
@@ -2,7 +2,7 @@
...
+    <option name="PROGRAM_PARAMETERS" value="-SstateManagerType=DATABASE -Sdatabase.user=user -Sdatabase.pass=password -Sdatabase.jdbc.url=jdbc:postgresql://localhost:5432/cordacluster --instance-id=0 -mbus.busType=DATABASE -spassphrase=password -ssalt=salt -spassphrase=password -ssalt=salt -ddatabase.user=user -ddatabase.pass=password -ddatabase.jdbc.url=jdbc:postgresql://localhost:5432/cordacluster -ddatabase.jdbc.directory=$ProjectFileDir$/applications/workers/release/combined-worker/drivers -rtls.crt.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/server.crt -rtls.key.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/server.key -rtls.ca.crt.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/ca-chain-bundle.crt" />
```
sprintengine: run config

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
```diff
@@ -118,11 +118,13 @@ class CombinedWorker @Activate constructor(
         val databaseConfig = PathAndConfig(BootConfig.BOOT_DB, params.databaseParams)
         val cryptoConfig = PathAndConfig(BootConfig.BOOT_CRYPTO, createCryptoBootstrapParamsMap(params.hsmId))
         val restConfig = PathAndConfig(BootConfig.BOOT_REST, params.restParams)
+        val stateManagerConfig = PathAndConfig(BootConfig.BOOT_STATE_MANAGER, params.stateManagerParams)
```
sprintengine, line 137: state manager is now boot config.

applications/workers/release/flow-worker/src/main/kotlin/net/corda/applications/workers/flow/FlowWorker.kt
```diff
@@ -91,4 +98,7 @@ class FlowWorker @Activate constructor(
 private class FlowWorkerParams {
     @Mixin
     var defaultParams = DefaultWorkerParams()
+
+    @CommandLine.Option(names = ["-S", "--${BootConfig.BOOT_STATE_MANAGER}"], description = ["Parameters for the state manager."])
+    var stateManagerParams = emptyMap<String, String>()
```
sprintengine, line 107: state manager is now boto config, I don't think I tested this out fully, I'm not sure if we deploy without config if the worker will give an error.

charts/corda-lib/templates/_worker.tpl
```diff
@@ -269,6 +269,13 @@ spec:
           - "-ddatabase.pool.keepaliveTimeSeconds={{ .clusterDbConnectionPool.keepaliveTimeSeconds }}"
           - "-ddatabase.pool.validationTimeoutSeconds={{ .clusterDbConnectionPool.validationTimeoutSeconds }}"
           {{- end }}
+          {{- if $optionalArgs.stateManagerAccess }}
```
sprintengine: Whichever workers define "StateManagerAccess" true in the workers.yaml, wll get these properties.

charts/corda/templates/workers.yaml
```diff
@@ -4,7 +4,9 @@
 {{- include "corda.worker" ( list $ .Values.workers.db "db"
   ( dict "clusterDbAccess" true )
 ) }}
-{{- include "corda.worker" ( list $ .Values.workers.flow "flow" ) }}
+{{- include "corda.worker" ( list $ .Values.workers.flow "flow"
```
sprintengine, line 7: flow worker takes statemanager command line

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -0,0 +1,48 @@
...
+     * @return states that could not be deleted due to mismatch versions.
+     */
+    fun <S : Any> delete(clazz: Class<S>, keys: Set<String>): Map<String, State<S>>
+
+    /**
+     * Retrieve all states that were last updated between [start] and [finish] times.
+     *
+     * @return states that were last updated between [start] and [finish] times.
+     */
+    fun <S : Any> getUpdatedBetween(clazz: Class<S>, start: Instant, finish: Instant): Map<String, State<S>>
```
sprintengine, line 47: Afaik @LWogan also needs an API that does the opposite of this, getStatesNotUpdatedBetween, we can add this in a later PR.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    override fun findUpdatedBetween(
+        entityManager: EntityManager,
+        start: Instant,
+        finish: Instant
+    ): Collection<StateEntity> {
+        return entityManager
+            .createNamedQuery(QUERY_STATES_BY_UPDATED_TIMESTAMP_NAME.trimIndent(), StateEntity::class.java)
+            .setParameter(STAR_TIMESTAMP, start)
```
sprintengine, line 97: minor - typo

conal-smith (commented):
lgtm, I can't approve as I created the PR

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
```diff
@@ -128,11 +131,13 @@ class CombinedWorker @Activate constructor(
         val databaseConfig = PathAndConfig(BootConfig.BOOT_DB, params.databaseParams)
         val cryptoConfig = PathAndConfig(BootConfig.BOOT_CRYPTO, createCryptoBootstrapParamsMap(params.hsmId))
         val restConfig = PathAndConfig(BootConfig.BOOT_REST, params.restParams)
-        val config = getBootstrapConfig(
+        val stateManagerConfig = PathAndConfig(BootConfig.BOOT_STATE_MANAGER, params.stateManagerParams)
```
JamesHR3, line 137: The construction here for setting up the state manager config feels a bit off. We create the path and config, and then later if it's set correctly we modify the config that we've just created here? Why not set the state manager config up correctly once (perhaps in a helper somewhere?) and then pass the fully formed config into `getBootstrapConfig`?

applications/workers/release/flow-worker/src/main/kotlin/net/corda/applications/workers/flow/FlowWorker.kt
```diff
@@ -92,4 +99,10 @@ class FlowWorker @Activate constructor(
 private class FlowWorkerParams {
     @Mixin
     var defaultParams = DefaultWorkerParams()
-}
\ No newline at end of file
+
+    @CommandLine.Option(
```
JamesHR3, line 103: Might we need this in other workers than combined and flow? I think we probably will - we should put this in the shared worker parameters somewhere.

applications/workers/worker-common/src/test/kotlin/net/corda/applications/workers/workercommon/internal/BootstrapConfigTest.kt
```diff
@@ -29,8 +29,8 @@ class BootstrapConfigTest {
     private val extraParamsMap = listOf(
         PathAndConfig("fred", mapOf("age" to "12", "hair" to "none"))
     )
-    private val file1 = Path.of(this::class.java.classLoader.getResource("test1.properties").toURI())
-    private val file2 = Path.of(this::class.java.classLoader.getResource("test2.properties").toURI())
+    private val file1 = Path.of(this::class.java.classLoader.getResource("test1.properties")!!.toURI())
```
JamesHR3, line 32: How come you need the `!!` when you didn't before? Wouldn't you have got the same NPE either way?

charts/corda/values.yaml
```diff
@@ -225,6 +225,9 @@ bootstrap:
           key: ""
   # Configuration for database bootstrap
   db:
+    # -- Indicates whether DB bootstrap is enabled as part of installation
+    enabled: true
```
JamesHR3, line 229: Would you mind explaining what this change is and why it's needed?

components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/ConfigProcessor.kt
```diff
@@ -2,6 +2,7 @@ package net.corda.configuration.read.impl
 
 import com.typesafe.config.ConfigFactory
 import com.typesafe.config.ConfigRenderOptions
+import java.util.concurrent.ConcurrentHashMap
```
JamesHR3, line 5: Might as well revert this one - there's no changes here.

components/state-manager/build.gradle
```diff
@@ -0,0 +1,20 @@
+plugins {
+    id 'corda.common-library'
+    id 'corda.common-publishing'
+}
+
+description 'Corda State Manager Service'
```
JamesHR3, line 6: I think we should remove this component entirely and let things that require the state manager use the library and supply the config themselves. This extra layer of services isn't really adding anything, and could result in subtle bugs if dependent services access the state manager this provides at the wrong time across a config change.

libs/configuration/configuration-merger/src/main/kotlin/net/corda/libs/configuration/merger/impl/ConfigMergerImpl.kt
```diff
@@ -27,4 +27,4 @@ class ConfigMergerImpl @Activate constructor(
         val bootDBParamsConfig = bootConfig.getConfigOrEmpty(BOOT_DB)
         return bootDBParamsConfig.withFallback(updatedDbConfig)
     }
-}
\ No newline at end of file
+}
```
JamesHR3, line 30: Can we get rid of this whitespace change too?

libs/configuration/configuration-merger/build.gradle
```diff
@@ -14,9 +14,10 @@ dependencies {
...
-    implementation project(':libs:configuration:configuration-core')
-    implementation project(':libs:messaging:message-bus')
-    implementation "com.typesafe:config:$typeSafeConfigVersion"
-    implementation 'org.jetbrains.kotlin:kotlin-osgi-bundle'
     implementation 'org.slf4j:slf4j-api'
+    implementation 'org.jetbrains.kotlin:kotlin-osgi-bundle'
+    implementation "com.typesafe:config:$typeSafeConfigVersion"
+    implementation project(':libs:messaging:message-bus')
+    implementation project(':libs:configuration:configuration-core')
+    implementation project(':libs:state-manager:state-manager-api')
```
JamesHR3, line 22: I don't think you use this change any more - can this be reverted as well?

libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/configuration/DbBusConfigMergerImpl.kt
```diff
@@ -1,33 +1,44 @@
...
     override fun getMessagingConfig(bootConfig: SmartConfig, messagingConfig: SmartConfig?): SmartConfig {
-        val updatedMessagingConfig = messagingConfig?: SmartConfigImpl.empty()
+        var updatedMessagingConfig = messagingConfig ?: SmartConfigImpl.empty()
+
+        bootConfig.getConfigOrEmpty(BootConfig.BOOT_STATE_MANAGER).entrySet().forEach { entry ->
+            updatedMessagingConfig = updatedMessagingConfig.withValue(
+                "${BootConfig.BOOT_STATE_MANAGER}.${entry.key}",
+                fromAnyRef(bootConfig.getString("${BootConfig.BOOT_STATE_MANAGER}.${entry.key}"))
+            )
+        }
```
JamesHR3, line 28: Isn't this something of a config merge operation? Do the normal config merge utilities not work here?

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/config/KafkaConfigMergerImpl.kt
```diff
@@ -4,20 +4,23 @@ import com.typesafe.config.ConfigValueFactory
...
 import org.osgi.service.component.annotations.Component
+import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
 @Component(service = [BusConfigMerger::class])
 class KafkaConfigMergerImpl : BusConfigMerger {
 
     private companion object {
-        val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+        val logger: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
```
JamesHR3, line 23: I don't think you should need to declare the type of this.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/config/KafkaConfigMergerImpl.kt
```diff
@@ -37,6 +40,16 @@ class KafkaConfigMergerImpl : BusConfigMerger {
...
 
+        logger.debug { "Looping through State Manager Boot Configuration" }
+        val stateManagerBootConfig = bootConfig.getConfigOrEmpty(BOOT_STATE_MANAGER).entrySet()
+        stateManagerBootConfig.forEach { entry ->
+            logger.debug { "Entry key: ${entry.key}" }
+            updatedMessagingConfig = updatedMessagingConfig.withValue(
+                "$BOOT_STATE_MANAGER.${entry.key}",
+                ConfigValueFactory.fromAnyRef(bootConfig.getString("$BOOT_STATE_MANAGER.${entry.key}"))
+            )
+        }
```
JamesHR3, line 51: Doing this in two different places is suggesting to me that we haven't got this quite right. A utility for merging the state manager config in would be better, but we might be able to tidy this up further with a little thought.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -0,0 +1,48 @@
...
+ * The [StateManager] provides functions to manage states within the underlying persistent storage.
+ */
+interface StateManager : AutoCloseable {
+
+    /**
+     * Create [states] into the underlying storage.
+     * Control is only returned to the caller once all [states] have been updated and replicas of the underlying
+     * persistent storage, if any, are synced.
+     */
+    fun <S : Any> create(clazz: Class<S>, states: Set<State<S>>)
```
JamesHR3, line 15: What should happen if any of the states conflict with something already present in the store? Does it overwrite the old one, does that one fail, or do they all fail to get set?

How might the client establish what happened?

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Get all states referenced by [keys].
+     * Only states that have been successfully committed and distributed within the underlying persistent
+     * storage are returned.
+     */
+    fun <S : Any> get(clazz: Class<S>, keys: Set<String>): Map<String, State<S>>
```
JamesHR3, line 22: What happens if the key is not present in the store? Do we not return a value for that key, or do we throw an exception?

What if we do have a value but it's the wrong type `S`?

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Update [states] within the underlying storage.
+     * Control is only returned to the caller once all [states] have been updated and replicas of the underlying
+     * persistent storage, if any, are synced.
+     *
+     * @return states that could not be updated due to mismatch versions.
+     */
+    fun <S : Any> update(clazz: Class<S>, states: Set<State<S>>): Map<String, State<S>>
```
JamesHR3, line 31: This returns only those states **not** updated, correct? That might be a little surprising.

What happens if one state fails to update due to some problem with the backing store? Do we fail everything or just add the failed one to the map?

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Delete all states referenced by [keys] from the underlying storage.
+     * Control is only returned to the caller once all states have been deleted and replicas of the underlying
+     * persistent storage, if any, are synced.
+     *
+     * @return states that could not be deleted due to mismatch versions.
+     */
+    fun <S : Any> delete(clazz: Class<S>, keys: Set<String>): Map<String, State<S>>
```
JamesHR3, line 40: What happens if we fail to delete one of them due to failures in the backing store?

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -0,0 +1,48 @@
+package net.corda.libs.statemanager.api
+
+import java.time.Instant
+
+/**
+ * The [StateManager] provides functions to manage states within the underlying persistent storage.
+ */
+interface StateManager : AutoCloseable {
```
JamesHR3, line 8: Could we accept `Collection<T>` in this API rather than `Set<T>`?

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManagerFactory.kt
```diff
@@ -0,0 +1,17 @@
+package net.corda.libs.statemanager.api
+
+import net.corda.libs.configuration.SmartConfig
+
+/**
+ * Factory for creating type
+ */
+interface StateManagerFactory {
```
JamesHR3, line 8: Minor detail, but is the client expected to get one of these via an OSGi injected service?

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -0,0 +1,118 @@
...
+import javax.persistence.EntityManagerFactory
+
+@Suppress("ForbiddenComment")
+class StateManagerImpl(
+    private val stateRepository: StateRepository,
+    private val entityManagerFactory: EntityManagerFactory,
+    private val serializerFactory: CordaAvroSerializationFactory,
+    private val serializers: MutableMap<Class<*>, CordaAvroSerializer<*>> = ConcurrentHashMap(),
+    private val deserializers: MutableMap<Class<*>, CordaAvroDeserializer<*>> = ConcurrentHashMap(),
+) : StateManager {
```
JamesHR3, line 25: This could do with some logging even if it's at the debug/trace level.

JamesHR3 (commented):
I've not reviewed the full thing yet, but this is a start.

It feels like this could really be two PRs - one with the library API and implementation and another to plug it in elsewhere. That would make reviewing this simpler.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/factory/StateManagerFactoryImpl.kt
```diff
@@ -0,0 +1,74 @@
...
+        val pass = config.getString(JDBC_PASS)
+        val jdbcUrl = config.getString(JDBC_URL)
+        val maxPoolSize = config.getInt(JDBC_POOL_MAX_SIZE)
+        val minPoolSize = config.getIntOrDefault(JDBC_POOL_MIN_SIZE, maxPoolSize)
+        val idleTimeout = config.getInt(JDBC_POOL_IDLE_TIMEOUT_SECONDS).toLong().run(Duration::ofSeconds)
+        val maxLifetime = config.getInt(JDBC_POOL_MAX_LIFETIME_SECONDS).toLong().run(Duration::ofSeconds)
+        val keepAliveTime = config.getInt(JDBC_POOL_KEEP_ALIVE_TIME_SECONDS).toLong().run(Duration::ofSeconds)
+        val validationTimeout = config.getInt(JDBC_POOL_VALIDATION_TIMEOUT_SECONDS).toLong().run(Duration::ofSeconds)
+
+        val dataSource = HikariDataSourceFactory().create(
```
JamesHR3, line 48: How is this done in the DB code in the rest of the project? Can we leverage work already done?

applications/workers/release/flow-worker/src/main/kotlin/net/corda/applications/workers/flow/FlowWorker.kt
sprintengine: I had thought about putting it into the shared worker parameters, but since we won't have state manager in _all_ workers, I opted to set this up on a per-worker basis.

EDIT - I notice not all the "default worker params" are used in all the workers, so this could be renamed "SharedWorkerParams" rather than implying all params are used in every worker. I added state-manager there as it'll mean fewer code change to add support for stateManager in a worker.

jujoramos:
> It feels like this could really be two PRs - one with the library API and implementation and another to plug it in elsewhere. That would make reviewing this simpler.

Agreed, 💯... Will spend some time today trying to split the `PR` into two and applying other review comments.

applications/workers/worker-common/src/test/kotlin/net/corda/applications/workers/workercommon/internal/BootstrapConfigTest.kt
jujoramos: It is not needed, I've just added it to get rid of the IntelliJ warning 😓. Will revert the change.

charts/corda/values.yaml
jujoramos: It is not needed, it just made sense to have it at the top before all other properties (they won't be used if `enabled` is set as `false`). Will revert the change.

components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/ConfigProcessor.kt
jujoramos: 👍

components/state-manager/build.gradle
jujoramos: 👍

libs/configuration/configuration-merger/src/main/kotlin/net/corda/libs/configuration/merger/impl/ConfigMergerImpl.kt
jujoramos: 👍

libs/configuration/configuration-merger/build.gradle
jujoramos: 👍

libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/configuration/DbBusConfigMergerImpl.kt
jujoramos: Will revert, thanks for pointing it out (@conalsmith-r3 : guess this is where you'll need to pick up user + pass from the command line and pass it into the configuration?).

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/config/KafkaConfigMergerImpl.kt
jujoramos: I've just added it to get rid of the IntelliJ warning 😓. Will revert the change.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/config/KafkaConfigMergerImpl.kt
jujoramos: 👍

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
jujoramos: Only the ones that didn't fail will be persisted, and the ones that failed will be returned to the user with an actual cause (updated [here](https://github.com/corda/corda-runtime-os/pull/4602/commits/6a5fc921c6b39f3a9aa8c29a9a9425bbd74f3f51)).

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
jujoramos: > What happens if the key is not present in the store? Do we not return a value for that key, or do we throw an exception?

None, the returned map simply won't contain the state.

> What if we do have a value but it's the wrong type S?

Haven't thought about this yet, guess the state manager should have an injectable error handler for serialisation / deserialisation or something similar 🤔

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
jujoramos: Updated [here](https://github.com/corda/corda-runtime-os/pull/4602/commits/6a5fc921c6b39f3a9aa8c29a9a9425bbd74f3f51).

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
jujoramos: The state manager doesn't serialise / deserialise anymore, it deals with byte arrays only.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/factory/StateManagerFactoryImpl.kt
jujoramos: We should entirely get rid of Hibernate and the EMF and use JDBC directly instead (already created a ticket and referenced it within the split `PR`).

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
jujoramos: The state manager doesn't deal with serialisers/deserialisers anymore (implemented in https://github.com/corda/corda-runtime-os/pull/4602).

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
jujoramos: > This returns only those states not updated, correct? That might be a little surprising.

Guess it goes against the "standards" a little bit, doesn't it... 🤔 

> What happens if one state fails to update due to some problem with the backing store? Do we fail everything or just add the failed one to the map?

The whole batch is halted in that case, the only case on which individual states are returned is when the optimistic locking check fails.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
jujoramos: The whole batch is halted in that case (single transaction), the only case on which individual states are returned is when the optimistic locking check fails.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
jujoramos: 👍

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
sprintengine: Perfect! +1

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
sprintengine: I just opted to reuse existing mechanism to build the config, albeit it's pretty unintuitive. I'll add a parameter to the `getBootstrapConfig` to take a list of configs, and create the config for `stateManager` and pass it in. That should add a bit of clarity around this code.

EDIT - I refactored this code a bit, removing the PathAndConfig entirely.

PR #4502, opened 2023-08-22
CORE-16230 Split UniquenessProcessor into its own worker

mbrkic-r3:
Reverted to 5.1 branch of corda-shared-build-pipeline-steps (branch mbrkic-r3/CORE-16230/uniqueness-worker was used temporarily to minimize build risks until it was merged to 5.1)


mbrkic-r3:
Used wrong branch, closing

PR #4504, opened 2023-08-22
CORE-9569 Remove Direct Access to Consumers From StateAndEventConsumer

Cap-j:
Remove public access to `eventConsumer` and `stateConsumer` from `StateAndEventConsumer`.


Cap-j:
Tests failing with tryProcessBatchOfEvents() - refactoring going deeper than initially expected

PR #4506, opened 2023-08-22
CORE-9975: Allow `sign` operation to take a key category in context argument

malachyb:


libs/messaging/messaging-impl/src/test/kotlin/net/corda/messaging/publisher/CordaPublisherImplTest.kt
```diff
@@ -256,7 +256,7 @@ class CordaPublisherImplTest {
     fun testBatchPublishWithMultipleThreads() {
         val publisher = createPublisher(true)
         val numThreads = 100
-        val futures = (0..numThreads).map { CompletableFuture<Unit>() }
+        val futures = List(numThreads) { CompletableFuture<Unit>() }
```
LWogan: This change already exists in release 5.1 from this merge. https://github.com/corda/corda-runtime-os/pull/4497/files

I think you need to merge down from release branch

malachyb: Merge from release done, not sure how that ended up there. Clean now

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/SigningWrappedSpec.kt
```diff
@@ -11,6 +11,7 @@ import java.security.PublicKey
  * @param keyMaterialSpec The spec for the wrapped key.
  * @param keyScheme The scheme for the key used for signing operation.
  * @param signatureSpec The signature spec to use for signing, such as SHA256withECDSA, etc.
+ * @param category The category of the key, if specified
```
dickon: add .

dickon: thanks, done.

dickon (approved):
LGTM

dickon:
In particular, now that the API is frozen I think a `cateogry` key in the `context` `Map<String,String>` parameter  is the best way to communicate a category.

PR #4507, opened 2023-08-22
CORE-16431: Make sure the MemberSynchronisationServiceImpl constructors have different number of arguments

yift-r3:
The issue is the same as https://github.com/corda/corda-runtime-os/pull/3128 - and the solution is also the same.

### DIFF
--- a/components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
+++ b/components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
@@ -69,20 +69,9 @@ import java.util.UUID
 import java.util.concurrent.TimeUnit
 
 @Component(service = [SynchronisationService::class])
-@Suppress("LongParameterList")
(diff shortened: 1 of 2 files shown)


williamvigorr3 (dismissed):
LGTM

components/membership/synchronisation-impl/src/main/kotlin/net/corda/membership/impl/synchronisation/MemberSynchronisationServiceImpl.kt
```diff
@@ -112,30 +101,46 @@ class MemberSynchronisationServiceImpl internal constructor(
...
-        MerkleTreeGenerator(
-            merkleTreeProvider,
-            cordaAvroSerializationFactory,
-        ),
-        UTCClock(),
-        membershipPersistenceClient,
-        groupParametersFactory,
+    )
+
+    internal data class Services(
```
dimosr, line 134: should we add a comment here, so that future readers can easily understand what's the purpose of this without digging through history?

dimosr (dismissed):
LGTM - just a minor comment 

Also, make sure the eventual commit from the merge has a good commit message rather than just linking to the previous PR as in the description.

dimosr (approved):

PR #4511, opened 2023-08-22
CORE-6192 Async web client to be used as the RPC client 

pmck91:


owenstanford (commented):
It would be good to have a little background in the PR description.
I think this needs to be a lib rather than component.

components/web/http-client/src/main/kotlin/net/corda/web/client/CordaHttpClientImpl.kt
```diff
@@ -0,0 +1,32 @@
...
+import java.net.http.HttpResponse
+import java.util.concurrent.Executors
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.future.await
+import org.osgi.service.component.annotations.Component
+
+@Component(service = [CordaHttpClient::class])
+class CordaHttpClientImpl : CordaHttpClient {
+
+    private val threadPoolSize = 10  // Modify based on requirements
```
owenstanford, line 15: assume this is to limit the max number of concurrent (executing) requests, in which case it should probably be configurable by the user(s) of the client, and probably should have a configurable default.

pmck91: I will add that in today!

pmck91:
> It would be good to have a little background in the PR description. I think this needs to be a lib rather than component.

of course, its only in draft stage at the moment Ill update the description once its ready for review :)

owenstanford:
> I think this needs to be a lib rather than component.

Thinking about this a little more my view is that we shouldn't preemptively be creating lib code unless we have a demonstrable reuse cases elsewhere, if the only use case for this utility is in the messaging code, it should just be a package in the module that needs it. If we find a reuse case in the future it can be pulled up into it's own lib.

thiagoviana:
> > It would be good to have a little background in the PR description. I think this needs to be a lib rather than component.
> 
> of course, its only in draft stage at the moment Ill update the description once its ready for review :)

Draft or not, I'd suggest putting a description now to make the life of the reviewers easier, as we're actually expecting people to review it.

components/web/http-client/src/main/kotlin/net/corda/web/client/CordaHttpClient.kt
```diff
@@ -0,0 +1,9 @@
+package net.corda.web.client
+
+import java.net.URL
+
+interface CordaHttpClient {
+
+    suspend fun post(url: URL, payload: ByteArray) : ByteArray
```
mbrkic-r3, line 7: How do we handle errors?

components/web/http-client/src/main/kotlin/net/corda/web/client/CordaHttpClientImpl.kt
mbrkic-r3: Will client support retries?

components/web/http-client/src/main/kotlin/net/corda/web/client/CordaHttpClientImpl.kt
```diff
@@ -0,0 +1,32 @@
...
+        .build()
+
+    override suspend fun post(url: URL, payload: ByteArray): ByteArray {
+        return coroutineScope {
+            val request = HttpRequest.newBuilder()
+                .uri(url.toURI())
+                .POST(HttpRequest.BodyPublishers.ofByteArray(payload))
+                .build()
+
+            client.sendAsync(request, HttpResponse.BodyHandlers.ofByteArray()).await().body()
```
mbrkic-r3, line 29: What do we get with async calls considering we immediately block (wait for result)?

driessamyn: I agree. The async API should be exposed to the consumer in my opinion.

components/web/http-client/src/main/kotlin/net/corda/web/client/CordaHttpClientImpl.kt
```diff
@@ -0,0 +1,32 @@
...
+import java.net.http.HttpClient
+import java.net.http.HttpRequest
+import java.net.http.HttpResponse
+import java.util.concurrent.Executors
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.future.await
+import org.osgi.service.component.annotations.Component
+
+@Component(service = [CordaHttpClient::class])
+class CordaHttpClientImpl : CordaHttpClient {
```
driessamyn, line 13: This class is so "thin", i.e. doesn't really add anything to the out-the-box `HttpClient` that I don't think it warrants its own library/module.
Can this not just be rolled into the patterns library client?

PR #4514, opened 2023-08-23
ES-1184: Ensure correct CLI version used in release builds 

ronanbrowne:
If in a GA or an RC build, we should be using the version of the CLI set in `gradle.properties` i.e. 5.1.0-HC03, for example,

Existing logic ended up setting `baseImageTag` as `unstable-5.1.0-HC03` in the above example which is incorrect, for release branch builds / PRs the existing logic is correct but for release builds we need to lose the unstable prefix and just take `cliVersion`

Note, for HCs RELEASE_TYPE is set to RC also, this env var is set by CI server


ronanbrowne:
@seanbrereton Fix for the issue we observed during corda-runtime-os internal release last week

yarivus (approved):
LGTM

PR #4515, opened 2023-08-23
CORE-16556: Removing NoopTracingService class

ZawarR3:
Previously when configuring tracing for any given worker, if no Zipkin host URL was provided, the `TracingService` would be configured to be `NoopTracingService`,  however now that tracing is available even without a Zipkin server, there is no need for the `NoopTracingService` class since everything is being handled by the `BraveTracingService`. This change removes the `NoopTracingService` as it is no longer needed.


josephzunigadaly (approved):

chrisr3:
How is this compatible with unit testing?

chrisr3 (changes requested):
Unit tests and integration tests _must_ require a `NoopTracingService`! How can they not?

You can't possibly be expecting Gradle builds running either locally or on Jenkins to need `BraveTracingService` can you?

I thought I'd requested that the tracing library be separated into a _"waffer theen"_ API and an implementation, whereas this PR seems to be going in the diametrically opposite direction.

PR #4519, opened 2023-08-24
CORE-11675 remove the hard coded avro serializers from the producer and consumer builders

ciaran238:
### Context
The kafka and DB producer/consumer builders create concreate instances of the avro serialisers/deserialisers rather than using the factory. 
### Changes
This PR changes the builders so they use the factory to create serializers.
It also removes the implementation of `Serializer<T>` from `CordaAvroSerializerImpl` and `Deserializer<Any>` from `CordaAvroDeserializerImpl`

### DIFF
--- a/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/serialization/CordaDBAvroSerializerImpl.kt
+++ b/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/serialization/CordaDBAvroSerializerImpl.kt
@@ -14,17 +14,16 @@ class CordaDBAvroSerializerImpl<T : Any>(
         private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
     }
 
-    override fun serialize(data: T?): ByteArray? {
+    override fun serialize(data: T): ByteArray? {
 
         try {
             return when (data) {
                 is String -> data.encodeToByteArray()
                 is ByteArray -> data
-                null -> null
                 else -> schemaRegistry.serialize(data).array()
             }
         } catch (ex: Throwable) {
-            val message = "Failed to serialize instance of class type ${data?.javaClass?.name}"
+            val message = "Failed to serialize instance of class type ${data::class.java.name} containing $data"
 
             onError?.invoke(message.toByteArray())
             log.error(message, ex)
--- a/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/serialization/CordaDBAvroSerializerImpl.kt
+++ b/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/serialization/CordaDBAvroSerializerImpl.kt
@@ -14,16 +14,17 @@ class CordaDBAvroSerializerImpl<T : Any>(
         private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
     }
 
-    override fun serialize(data: T): ByteArray? {
+    override fun serialize(data: T?): ByteArray? {
 
         try {
             return when (data) {
-                is String -> (data as String).encodeToByteArray()
+                is String -> data.encodeToByteArray()
                 is ByteArray -> data
+                null -> null
                 else -> schemaRegistry.serialize(data).array()
             }
         } catch (ex: Throwable) {
-            val message = "Failed to serialize instance of class type ${data::class.java.name} containing $data"
+            val message = "Failed to serialize instance of class type ${data!!::class.java.name} containing $data"
 
             onError?.invoke(message.toByteArray())
             log.error(message, ex)
--- a/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/serialization/CordaDBAvroSerializerImpl.kt
+++ b/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/serialization/CordaDBAvroSerializerImpl.kt
@@ -24,7 +24,7 @@ class CordaDBAvroSerializerImpl<T : Any>(
                 else -> schemaRegistry.serialize(data).array()
             }
         } catch (ex: Throwable) {
-            val message = "Failed to serialize instance of class type ${data!!::class.java.name} containing $data"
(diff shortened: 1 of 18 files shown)


libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/serialization/CordaDBAvroSerializerImpl.kt
```diff
@@ -14,16 +14,17 @@ class CordaDBAvroSerializerImpl<T : Any>(
...
             return when (data) {
-                is String -> (data as String).encodeToByteArray()
+                is String -> data.encodeToByteArray()
                 is ByteArray -> data
+                null -> null
                 else -> schemaRegistry.serialize(data).array()
             }
         } catch (ex: Throwable) {
-            val message = "Failed to serialize instance of class type ${data::class.java.name} containing $data"
+            val message = "Failed to serialize instance of class type ${data!!::class.java.name} containing $data"
```
josephzunigadaly: We should avoid the `!!` operator here. We don't want to throw a null pointer exception while building our error message.

I'd also remove `$data` from the log message.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/serialization/CordaAvroSerializerImpl.kt
```diff
@@ -33,29 +32,22 @@ class CordaAvroSerializerImpl<T : Any>(
...
                 is String -> stringSerializer.serialize(null, data)
                 is ByteArray -> data
+                null -> null
                 else -> {
                     schemaRegistry.serialize(data).array()
                 }
             }
         } catch (ex: Throwable) {
-            val message = "Failed to serialize instance of class type ${data::class.java.name} containing $data"
+            val message = "Failed to serialize instance of class type ${data!!::class.java.name} containing $data"
```
josephzunigadaly: Same with this one

libs/serialization/serialization-avro/src/main/kotlin/net/corda/avro/serialization/CordaAvroSerializer.kt
```diff
@@ -11,6 +11,6 @@ interface CordaAvroSerializer<T> {
      * @param data The object to be serialized.
      * @return The serialized byte stream for transfer across the message bus or null if unsuccessful.
      */
-    fun serialize(data: T): ByteArray?
+    fun serialize(data: T?): ByteArray?
```
simon-johnson-r3: I don't like the fact we've polluted our Kotlin api to shoehorn this in as a Kafka `Serializer<>`, I suggest we revert this and resolve the null issue in the adaptor passed to `KafkaProducer`.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/builder/KafkaCordaProducerBuilderImpl.kt
```diff
@@ -83,10 +82,12 @@ class KafkaCordaProducerBuilderImpl @Activate constructor(
             if (currentBundle != null) {
                 Thread.currentThread().contextClassLoader = currentBundle.adapt(BundleWiring::class.java).classLoader
             }
+            val keySerializer = cordaAvroSerializationFactory.createAvroSerializer<Any>(onSerializationError)
+            val valueSerializer = cordaAvroSerializationFactory.createAvroSerializer<Any>(onSerializationError)
             KafkaProducer(
                 kafkaProperties,
-                CordaAvroSerializerImpl(avroSchemaRegistry, onSerializationError),
-                CordaAvroSerializerImpl(avroSchemaRegistry, onSerializationError)
```
simon-johnson-r3, line 89: Could we consider having an extension function for `CordaAvroSerializationFactory` which returns you a Kafka `Serializer<>`?

That function would return you an adaptor instead of defining one inline here. And it could deal with the possibility of being told to serialize null (see comment in `CordaAvroSerializer`. The old implementation simply returned null if it received null, so you could handle that as a special case.

Here it would look like `cordaAvroSerializationFactory.createAvroBasedKafkaSerializer(...)` i.e.
```
KafkaProducer(
                kafkaProperties,
                cordaAvroSerializationFactory.createAvroBasedKafkaSerializer(onSerializationError)
                cordaAvroSerializationFactory.createAvroBasedKafkaSerializer(onSerializationError)
                ...
```

I think at the point of use that is quite expressive. What we have now is a little esoteric. It's probably fine to define that extension function in this file.

libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/serialization/CordaDBAvroSerializerImpl.kt
```diff
@@ -14,16 +14,17 @@ class CordaDBAvroSerializerImpl<T : Any>(
         private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
     }
 
-    override fun serialize(data: T): ByteArray? {
+    override fun serialize(data: T?): ByteArray? {
```
simon-johnson-r3: I assume you could revert this if you deal with the null in the adaptor as suggested in other comments.

simon-johnson-r3 (changes requested):

ben-millar (approved):
For the serializer tests, I think I'd rather check the values too rather than just checking that the result isn't null, but I'll leave that up to you.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/builder/KafkaCordaProducerBuilderImpl.kt
ciaran238: I implemented this but have done so in a separate interface, I think it makes sense to separate it from the CordaAvroSerializationFactory

josephzunigadaly (approved):

simon-johnson-r3 (approved):

PR #4522, opened 2023-08-24
CORE-16612: Force quasar to ignore metrics packages when instrumenting code

JamesHR3:
**Problem Description**
Sometimes when running flows, an exception can occur inside the metrics library indicating that an object is `null`, despite this not being possible from the compiler's perspective. This is happening because Quasar is instrumenting the metrics library, and incorrectly restoring the metrics object on the stack after a suspension. There is no suspendable code inside the metrics library, and so this instrumentation is not required.

**Solution**
Use the `@QuasarIgnoreAllPackages` annotation to force Quasar to not instrument these packages.

**Testing**
No new unit tests required. The change has been run in the test environment where the original problem was observed and this change addresses the issue.

### DIFF
--- a/libs/metrics/build.gradle
+++ b/libs/metrics/build.gradle
@@ -10,6 +10,7 @@ dependencies {
 
     compileOnly 'org.osgi:osgi.core'
     compileOnly 'org.osgi:osgi.annotation'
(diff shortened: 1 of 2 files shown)


jujoramos (approved):

conal-smith (approved):
lgtm

ronanbrowne (approved):
LGTM

PR #4524, opened 2023-08-24
CORE-16215 Enhance db utils sql server support

PranushanR3:
This PR is adds support for SQL server. SQL server support has been added to the db-testkit and should now be able to support connections to a SQL server for testing. The DB type that will be used will be decided via the properties set in the gradle.properties file. Connections have been tested using the existing tests. Postgres and HSQL connections work as expected and tests pass. The SQL server connection works in the tests however the tests fail due a syntax error and I believe if syntax is correct, SQL server will work like the Postgres and HSQL DB.


build.gradle
```diff
@@ -251,6 +251,10 @@ allprojects {
         inputs.property('postresUser', postgresUser)
         inputs.property('postgresPassword', postgresPassword)
 
+        inputs.property('sqlPort',sqlPort)
+        inputs.property('sqlUser',sqlUser)
+        inputs.property('sqlPassword',sqlPassword)
```
blsemo: Does `sql` stand for SqlServer here? If yes, that is a bit confusing, as it could also mean generic sql database. Probably better to either use `mssqlPort` or `sqlserverPort`.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtilsAbstract.kt
```diff
@@ -0,0 +1,100 @@
...
+import net.corda.orm.DbEntityManagerConfiguration
+import net.corda.orm.DdlManage
+import net.corda.orm.EntityManagerConfiguration
+import net.corda.test.util.LoggingUtils.emphasise
+
+import org.slf4j.Logger
+import java.sql.Connection
+
+
+abstract class DbUtilsAbstract {
```
blsemo: The name is rather unconventional. Should either be `AbstractDbUtils` or `DbUtilsBase`.

blsemo (commented):
Could we structure the change so that all the tests use the abstract DbUtils class, and we inject the implementation at runtime - so that testing everything against a different database does not require changing the code everywhere?

build.gradle
```diff
@@ -264,6 +268,9 @@ allprojects {
         systemProperty 'postgresPassword', postgresPassword
         systemProperty 'org.slf4j.simpleLogger.showDateTime', true
         systemProperty 'org.slf4j.simpleLogger.dateTimeFormat', 'yyyy-MM-dd HH:mm:ss:SSS Z'
+        systemProperty 'sqlPort', sqlPort
+        systemProperty 'sqlUser',sqlUser
+        systemProperty 'sqlPassword',sqlPassword
```
relyafi: I think these should match the properties set earlier.
```suggestion
        systemProperty 'mssqlPort', sqlPort
        systemProperty 'mssqlUser',sqlUser
        systemProperty 'mssqlPassword',sqlPassword
```

gradle.properties
```diff
@@ -155,6 +156,14 @@ postgresDb=
 postgresUser=
 postgresPassword=
 
+# Setting these properties
+# SQL server has a password requirement, please make sure all passwords contain uppercase and lowercase letters
+# as well as base 10 digits
+# The user should be set to sa as SQL server defaults to this when starting up
```
relyafi: I think it needs to be made clearer that these are still related to the above DB integration test settings, and are mutually exclusive with the Postgres settings.

libs/db/db-core/src/main/kotlin/net/corda/db/core/SQLDataSourceFactory.kt
```diff
@@ -0,0 +1,6 @@
+package net.corda.db.core
+class SQLDataSourceFactory(
```
relyafi:
```suggestion
class MsSqlDataSourceFactory(
```

relyafi (changes requested):
I'm a bit concerned at the number of files that have had to change in this PR, and the fact that pretty much all tests are now using a `PostgresDbUtils` library. Ultimately, all of these tests should work with SQL server, and indeed any other arbitrary future database we add.

With that in mind, I would expect us to maintain a single `DbUtils` library, which instantiates the correct connection based on which properties were set. Setting both postgres as sql server properties at the same time should cause failure, and this can be validated at runtime. As it stands, it's not clear to me how we would run the tests with SQL server.

If we go with this approach, then I think this PR becomes a much smaller change.

I also think there needs to be a clearer description on the PR, and notes on how we tested.

simon-johnson-r3:
A summary of my thoughts on this change is that we've ended up with a design with a class hierarchy to manage just a single string with the driver class defined in it, but where we have a genuine and significant change in functional behaviour (in DbUtils) we have gone for conditional code based on booleans.

It feels like we've done this the wrong way round, that the data source factory is pretty generic and can be a single class with a driver injected when instantiated, and the DbUtils is the place where the real changes are needed and that should be split into a more OO approach?

For example, in DbUtils we've ended up with this:
```
fun createDataSource(

  if (!isPostgres) {
    val factory = PostgresDataSourceFactory()
  
  else {
    val factory = SQLDataSourceFactory()
```
So we've wrapped an OO based factory with a non OO factory method. All that code is doing is choosing a driver string.

Much of what is here now is probably emergent design centred around the fact we have hard coded Postgres support in the past. I would suggest instead of worrying about what's in the code now, start by drawing out what we want the class hierarchy and factory scheme to look like for its client code.

simon-johnson-r3 (changes requested):
See previous comment.

simon-johnson-r3:
Could you switch the title of the PR, the ticket is `CORE-16215`

libs/db/db-core/src/main/kotlin/net/corda/db/core/SQLDataSourceFactory.kt
relyafi: > I think this comment is still relevant, would also rename file to reflect updated class name

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -4,105 +4,143 @@ import com.typesafe.config.Config
...
 import net.corda.orm.DdlManage
 import net.corda.orm.EntityManagerConfiguration
-import net.corda.schema.configuration.DatabaseConfig
 import net.corda.test.util.LoggingUtils.emphasise
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import java.sql.Connection
+import net.corda.db.core.PostgresDataSourceFactory
+import net.corda.db.core.SQLDataSourceFactory
+import net.corda.schema.configuration.DatabaseConfig
```
relyafi: Not sure why we changed the existing import ordering? I'd put them in order as this aids maintainability. You can also use the "optimize imports" option by right clicking on the file in the project explorer in IntelliJ, and it will apply the correct ordering for you.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -4,105 +4,143 @@ import com.typesafe.config.Config
...
+        else if(!isMSSQL){
+            getPropertyNonBlank("mssqlPassword", "password")
+        }
+        else{
+            ""
+        }
     @Suppress("LongParameterList")
     fun getEntityManagerConfiguration(
-        inMemoryDbName: String,
+        inMemoryDbName : String,
```
relyafi: Unnecessary (and inconsistent) change?

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -4,105 +4,143 @@ import com.typesafe.config.Config
...
         dbPassword: String? = null,
         schemaName: String? = null,
         createSchema: Boolean = false,
         showSql: Boolean = true,
         rewriteBatchedInserts: Boolean = false
-    ): EntityManagerConfiguration {
-        val port = System.getProperty(POSTGRES_PORT_PROPERTY)
-        return if (!port.isNullOrBlank()) {
-            val ds = createPostgresDataSource(dbUser, dbPassword, schemaName, createSchema, rewriteBatchedInserts)
+    ):EntityManagerConfiguration{
```
relyafi:
```suggestion
    ):EntityManagerConfiguration {
```

Again, unnecessary (and incorrect) whitespace change

(This looks to be a recurring theme in the PR - please review all instances and revert where necessary)

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -4,105 +4,143 @@ import com.typesafe.config.Config
...
+    fun getDatabase() =
+        if (!isPostgres){
+            getPropertyNonBlank("postgresDb", "postgres")
+        }
+        else if (!isMSSQL){
+            getPropertyNonBlank("sqlDb", "sql")
+        }
+        else {
+            ""
+        }
```
relyafi: This approach works, but it may become unwieldy as we add more databases.

Ultimately we know what each database type needs to supply, it's effectively the information that each function returns in this object. I wonder if it might be cleaner and more scalable to define an interface which specifies each of these functions, have a different implementation for each DB type, and then have the `DBUtils` construct one of these classes once as part of initialisation?

That means that the implementation for each DB type would be separated into different classes / files, and we'd only need to do the if statement once on initialisation here. All the functions could then just delegate to the specific DB class.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -4,105 +4,143 @@ import com.typesafe.config.Config
...
-            jdbcUrl = if (rewriteBatchedInserts) {
-                "$jdbcUrl?currentSchema=$schemaName&reWriteBatchedInserts=true"
-            } else {
-                "$jdbcUrl?currentSchema=$schemaName"
+        if (!isPostgres){
+            val port = System.getProperty(POSTGRES_PORT_PROPERTY)
+            val host = getPropertyNonBlank(POSTGRES_HOST_PROPERTY,"localhost")
+            val factory = PostgresDataSourceFactory()
+            val db = getDatabase()
+            var jdbcurl = "jdbc:postgresql://$host:$port/$db"
```
relyafi:
```suggestion
            var jdbcUrl = "jdbc:postgresql://$host:$port/$db"
```

Use camel-case for `jdbcUrl`. Previously it was like this, so not sure why it was changed. There's a few instances of this,  they should all be corrected.

gradle.properties
```diff
@@ -155,6 +156,17 @@ postgresDb=
 postgresUser=
 postgresPassword=
 
+# DB integration tests
+# NOTE: Setting these properties
+# SQL server has a password requirement, please make sure all passwords contain uppercase and lowercase letters
+# as well as base 10 digits
+# The user should be set to sa as SQL server defaults to this when starting up
```
relyafi: This suggests the property needs setting to this. But we already default in the code, so I don't think this is true?

What are we saying here? If we are saying that the user _must_ always be sa, then there's no point in having the property. If it _could_ be different (which I suspect it could be if someone overrode the name when provisioning the container), then it doesn't seem any different from Postgres, and doesn't need to be explicitly called out?

gradle.properties
```diff
@@ -155,6 +156,17 @@ postgresDb=
 postgresUser=
 postgresPassword=
 
+# DB integration tests
```
relyafi:
```suggestion
# DB integration tests (MS SQL)
```

Would also make it clear in these comments that only one set of DB-specific properties should be set. I'd also update the Postgres comments to reflect the above, and update / remove line 152 (which is now inaccurate)

relyafi (commented):
This looks like a much more scalable solution now. I still think there's room to improve the maintainability - see specific comments.

There's also quite a few unnecessary changes to whitespace, variable names, and import ordering. I've commented where I've spotted these, but please review and make sure that these are corrected, and ensure that new code is also following the standard whitespace conventions.

Finally, the general comment from my last review still applies - there needs to be a clearer description on the PR, and notes on how we tested. We should confirm that HSQLDB, Postgres and MS SQL all run correctly following these changes.

libs/db/db-core/src/main/kotlin/net/corda/db/core/BaseDataSourceFactory.kt
```diff
@@ -2,17 +2,19 @@ package net.corda.db.core
 
 import java.time.Duration
 
-class PostgresDataSourceFactory(
-    private val datasourceFactory: DataSourceFactory = HikariDataSourceFactory()
-) {
+class BaseDataSourceFactory(
```
relyafi: Is `BaseDataSourceFactory` still an accurate name, given this is now just the singular implementation (after the latest changes)? I wonder if this should just be `DataSourceFactory` (maybe too generic?) or `CordaDataSourceFactory`?

build.gradle
```diff
@@ -264,6 +268,11 @@ allprojects {
         systemProperty 'postgresPassword', postgresPassword
         systemProperty 'org.slf4j.simpleLogger.showDateTime', true
         systemProperty 'org.slf4j.simpleLogger.dateTimeFormat', 'yyyy-MM-dd HH:mm:ss:SSS Z'
+        systemProperty 'mssqlHost', mssqlHost
+        systemProperty 'mssqlDb', mssqlDb
+        systemProperty 'mssqlPort', mssqlPort
+        systemProperty 'mssqlUser',mssqlUser
+        systemProperty 'mssqlPassword',mssqlPassword
```
relyafi:
```suggestion
        systemProperty 'mssqlUser', mssqlUser
        systemProperty 'mssqlPassword', mssqlPassword
```

libs/db/db-core/src/main/kotlin/net/corda/db/core/BaseDataSourceFactory.kt
blsemo:
```suggestion
class DataSourceFactory(
```
Not actually a base class anymore.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -14,47 +15,67 @@ import org.slf4j.Logger
...
-    private val logger: Logger = LoggerFactory.getLogger(this::class.java)
-
-    val isInMemory = System.getProperty(POSTGRES_PORT_PROPERTY).isNullOrBlank()
+interface DbUtilsHelper {
 
-    /**
-     * Returns Postgres DB name
-     */
-    fun getPostgresDatabase() = getPropertyNonBlank("postgresDb", "postgres")
+    val isInMemory: Boolean
```
blsemo: I think the `inMemory` should be a third helper implementation, not a special case on the postgres one.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -126,29 +139,166 @@ object DbUtils {
...
     }
+}
+
+class SQLServerHelper : DbUtilsHelper {
+    companion object {
+        private const val MSSQL_HOST_PROPERTY = "mssqlHost"
+        private const val MSSQL_PORT_PROPERTY = "mssqlPort"
+    }
+
+    override val isInMemory = System.getProperty(MSSQL_PORT_PROPERTY).isNullOrBlank()
```
blsemo: Shouldn't this always be `false` for MS SQL?

blsemo (commented):
I don't think it's quite there yet.
I would imagin HSQL to have its own helper, not be a sub-case of Postgres. That would also allow to clean up the interface and thus the other implementations.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -126,29 +211,135 @@ object DbUtils {
                 .withValue(DatabaseConfig.DB_USER, ConfigValueFactory.fromAnyRef(user))
                 .withValue(DatabaseConfig.DB_PASS, ConfigValueFactory.fromAnyRef(password))
         } else {
-            // in memory
             return ConfigFactory.empty()
                 .withValue(DatabaseConfig.JDBC_DRIVER, ConfigValueFactory.fromAnyRef("org.hsqldb.jdbc.JDBCDriver"))
                 .withValue(DatabaseConfig.JDBC_URL, ConfigValueFactory.fromAnyRef("jdbc:hsqldb:mem:$inMemoryDbName"))
                 .withValue(DatabaseConfig.DB_USER, ConfigValueFactory.fromAnyRef(user))
                 .withValue(DatabaseConfig.DB_PASS, ConfigValueFactory.fromAnyRef(password))
```
blsemo: I think the else branch should be an error now - when using HSQL, we should not use the PostgresHelper.

libs/db/db-core/src/main/kotlin/net/corda/db/core/DBBaseDataSourceFactory.kt
```diff
@@ -2,17 +2,19 @@ package net.corda.db.core
 
 import java.time.Duration
 
-class PostgresDataSourceFactory(
+class DBBaseDataSourceFactory(
```
blsemo: Do we actually still need this class? All it does is capturing the HikariDataSourceFactory, so maybe a free standing `createDatasource()` method would do?

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -126,29 +211,135 @@ object DbUtils {
...
-        return if (value.isNullOrBlank()) {
-            defaultValue
-        } else {
-            value
+class HSQLHelper : DbUtilsHelper {
+    override fun isInMemory(): Boolean = true
+
+    override fun getDatabase(): String = ""
+
+    override fun getAdminUser() = if (isInMemory()) "sa" else getPropertyNonBlank("postgresUser","postgres")
```
blsemo: You can hardcode this one to `sa` - it's always in memory.
Similar for a lot of other occurences in this file - whenever `isInMemory` is constant, you can remove the extra if checks.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -126,29 +211,135 @@ object DbUtils {
...
+    override fun getEntityManagerConfiguration(
+        inMemoryDbName: String,
+        dbUser: String?,
+        dbPassword: String?,
+        schemaName: String?,
+        createSchema: Boolean,
+        showSql: Boolean,
+        rewriteBatchedInserts: Boolean
+    ): EntityManagerConfiguration {
+        logger.info("Using in-memory (HSQL) DB".emphasise())
```
blsemo: Should probably check that `inMemoryDbName` is not blank.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -126,29 +211,135 @@ object DbUtils {
...
+            .withValue(DatabaseConfig.DB_USER, ConfigValueFactory.fromAnyRef(user))
+            .withValue(DatabaseConfig.DB_PASS, ConfigValueFactory.fromAnyRef(password))
+    }
+
+}
+
+object DbUtils {
+
+    val isNotPostgres = System.getProperty("postgresPort").isNullOrBlank()
+    val isNotMSSQL = System.getProperty("mssqlPort").isNullOrBlank()
```
blsemo:
```suggestion
    val isPostgres = !System.getProperty("postgresPort").isNullOrBlank()
    val isMSSQL = !System.getProperty("mssqlPort").isNullOrBlank()
```

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -126,29 +211,135 @@ object DbUtils {
...
+
+    var utilsHelper: DbUtilsHelper = when {
+        !isNotMSSQL -> SQLServerHelper()
```
blsemo: Should probably check that not MsSql and Postgres have both been set.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -126,29 +211,135 @@ object DbUtils {
...
+    }
+
+}
+
+object DbUtils {
+
+    val isNotPostgres = System.getProperty("postgresPort").isNullOrBlank()
+    val isNotMSSQL = System.getProperty("mssqlPort").isNullOrBlank()
+
+    var utilsHelper: DbUtilsHelper = when {
```
blsemo: Could this be `val`? Shouldn't change in the lifetime of the process.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
blsemo, line 1: Ideally split this file into separate files per class. Also, please sort out the formatting - there are a few cases of double empty lines, and a few functions have no empty lines between them.

blsemo (commented):
Getting there, but I think it needs a bit of extra clean-up.

libs/db/db-core/src/main/kotlin/net/corda/db/core/DBBaseDataSourceFactory.kt
PranushanR3: Left this the same as this class is still being used in other places other than the DbUtills. Changing this would require refactoring those other files such as `PostgresEntityMangerConfigurationTest. `

blsemo: Hmm, but you still need to touch all those places because you changed the name of the class - I still don't see the point of this class in the current form.

libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/persistence/DbUtils.kt
```diff
@@ -21,7 +22,8 @@ internal fun EntityManagerFactoryFactory.create(
     val dbSource = if (jdbcUrl == null || jdbcUrl.contains("hsqldb", ignoreCase = true)) {
         InMemoryDataSourceFactory().create(DB_MESSAGE_BUS)
     } else {
-        PostgresDataSourceFactory().create(
+        DBBaseDataSourceFactory(HikariDataSourceFactory()).create(
```
blsemo:
```suggestion
        DBBaseDataSourceFactory().create(
```
No point adding the default arg here.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/PostgresHelper.kt
```diff
@@ -0,0 +1,97 @@
...
+import net.corda.db.core.DBBaseDataSourceFactory
+import net.corda.orm.DbEntityManagerConfiguration
+import net.corda.orm.DdlManage
+import net.corda.orm.EntityManagerConfiguration
+import net.corda.schema.configuration.DatabaseConfig
+import net.corda.test.util.LoggingUtils.emphasise
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+class PostgresHelper : DbUtilsHelper{
```
blsemo: These helpers look like they have a lot of overlap - could you check if we can have a base class that implements the shared bits, and only specialise the bits that are genuinely different?

blsemo (commented):
Nearly there, a few more remarks...

blsemo (dismissed):
Looks good to me.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/AbstractDBHelper.kt
```diff
@@ -0,0 +1,97 @@
...
+            val jdbcUrlCopy = "$jdbcUrl?currentSchema=$schemaName"
+            return ConfigFactory.empty()
+                .withValue(DatabaseConfig.JDBC_URL, ConfigValueFactory.fromAnyRef(jdbcUrlCopy))
+                .withValue(DatabaseConfig.DB_USER, ConfigValueFactory.fromAnyRef(user))
+                .withValue(DatabaseConfig.DB_PASS, ConfigValueFactory.fromAnyRef(password))
+        }
+        return ConfigFactory.empty()
+            .withValue(DatabaseConfig.JDBC_URL, ConfigValueFactory.fromAnyRef(jdbcUrl))
+            .withValue(DatabaseConfig.DB_USER, ConfigValueFactory.fromAnyRef(user))
+            .withValue(DatabaseConfig.DB_PASS, ConfigValueFactory.fromAnyRef(password))
```
blsemo: Slighly more kotlin-y (and less code duplication)
```suggestion
        val currentJdbcUrl = if(!schemaName.isNullOrBlank()) "$jdbcUrl?currentSchema=$schemaName" else jdbcUrl
        return ConfigFactory.empty()
            .withValue(DatabaseConfig.JDBC_URL, ConfigValueFactory.fromAnyRef(currentJdbcUrl))
            .withValue(DatabaseConfig.DB_USER, ConfigValueFactory.fromAnyRef(user))
            .withValue(DatabaseConfig.DB_PASS, ConfigValueFactory.fromAnyRef(password))
```

blsemo (dismissed):
LGTM

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/AbstractDBHelper.kt
```diff
@@ -0,0 +1,95 @@
...
+import net.corda.db.core.createDataSource
+import net.corda.orm.DbEntityManagerConfiguration
+import net.corda.orm.DdlManage
+import net.corda.orm.EntityManagerConfiguration
+import net.corda.schema.configuration.DatabaseConfig
+import net.corda.test.util.LoggingUtils.emphasise
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+abstract class AbstractDBHelper : DbUtilsHelper{
```
relyafi:
```suggestion
abstract class AbstractDBHelper : DbUtilsHelper {
```

relyafi: This class may benefit from a KDoc description explaining its purpose.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/AbstractDBHelper.kt
```diff
@@ -0,0 +1,95 @@
...
+        dbUser: String?,
+        dbPassword: String?,
+        schemaName: String?,
+        createSchema: Boolean,
+        rewriteBatchedInserts: Boolean
+    ): CloseableDataSource {
+        val user = dbUser ?: getAdminUser()
+        val password = dbPassword ?: getAdminPassword()
+
+        if(!schemaName.isNullOrBlank()){
```
relyafi:
```suggestion
        if (!schemaName.isNullOrBlank()) {
```

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/AbstractDBHelper.kt
```diff
@@ -0,0 +1,95 @@
...
+            if(createSchema){
```
relyafi:
```suggestion
            if (createSchema) {
```

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/AbstractDBHelper.kt
```diff
@@ -0,0 +1,95 @@
...
+        inMemoryDbName: String,
+        dbUser: String?,
+        dbPassword: String?,
+        schemaName: String?
+    ): Config {
+        val user = dbUser ?: getAdminUser()
+        val password = dbPassword ?: getAdminPassword()
+        val currentJdbcUrl = if(!schemaName.isNullOrBlank()){
+            "$jdbcUrl?currentSchema=$schemaName"
+        }else{
```
relyafi:
```suggestion
        val currentJdbcUrl = if (!schemaName.isNullOrBlank()) {
            "$jdbcUrl?currentSchema=$schemaName"
        } else {
```

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtilsHelper.kt
```diff
@@ -0,0 +1,60 @@
+package net.corda.db.testkit
+
+import com.typesafe.config.Config
+import net.corda.db.core.CloseableDataSource
+import net.corda.orm.EntityManagerConfiguration
+import java.sql.Connection
+
+interface DbUtilsHelper {
```
relyafi: This may also benefit from a KDoc.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtilsHelper.kt
```diff
@@ -0,0 +1,60 @@
...
+            defaultValue
+        } else {
+            value
+        }
+    }
+
+    fun Connection.createSchema(schemaName: String?) {
+        requireNotNull(schemaName)
+        this.use { conn ->
+            conn.prepareStatement("CREATE SCHEMA IF NOT EXISTS $schemaName;").execute()
```
relyafi: Is this SQL database agnostic?

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/HSQLHelper.kt
```diff
@@ -0,0 +1,72 @@
...
+    override fun getEntityManagerConfiguration(
+        inMemoryDbName: String,
+        dbUser: String?,
+        dbPassword: String?,
+        schemaName: String?,
+        createSchema: Boolean,
+        showSql: Boolean,
+        rewriteBatchedInserts: Boolean
+    ): EntityManagerConfiguration {
+        if(inMemoryDbName.isBlank()){
```
relyafi:
```suggestion
        if (inMemoryDbName.isBlank()) {
```

relyafi (commented):
PR description / test notes broadly looks good now. Appreciate that we won't get a successful run of the full tests with SQL server as we do have further changes needed to support this, but I think we should at least try and get one successful test case from our existing set of tests to show we were able to connect / read / write to SQL server.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtilsHelper.kt
PranushanR3: Just looked it up and seems like it needs a different statement. It should be more along the line of 
```
IF NOT EXISTS (SELECT schema_name FROM information_schema.schemata WHERE schema_name = 'YourSchemaName')
BEGIN
    EXEC('CREATE SCHEMA YourSchemaName')
END
```
Will try get this changed today before I head out

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtilsHelper.kt
```diff
@@ -0,0 +1,63 @@
...
+    ): CloseableDataSource
+
+    fun createConfig(
+        inMemoryDbName: String = "",
+        dbUser: String? = null,
+        dbPassword: String? = null,
+        schemaName: String? = null
+    ): Config
+
+    fun getPropertyNonBlank(key: String, defaultValue: String): String {
```
simon-johnson-r3: This method sticks out here. I understand you want a generic helper method to do this, but I would separate those out from the polymorphic helpers on a per db level.

simon-johnson-r3 (changes requested):

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/HelperUtils.kt
```diff
@@ -0,0 +1,20 @@
...
+import java.sql.Connection
+
+fun getPropertyNonBlank(key: String, defaultValue: String): String {
+    val value = System.getProperty(key)
+    return if (value.isNullOrBlank()) {
+        defaultValue
+    } else {
+        value
+    }
+}
```
relyafi: I wonder if this should be wrapped in an `object` to provide a namespace? The function name alone feels generic enough it could clash with other things. Or maybe these should just be moved to the existing `DbUtils` file in the testkit? This doesn't necessarily feel specific to our helpers.

simon-johnson-r3 (dismissed):

wzur-r3 (dismissed):
LGTM for Gradle configuration files

blsemo:
I managed to run the test `EntitiesInBundlesTest` against Sqlserver, ensuring it uses the correct database and schema as specified by the test.
image

libs/db/osgi-integration-tests/build.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
     integrationTestRuntimeOnly "org.ops4j.pax.jdbc:pax-jdbc-hsqldb:$paxJdbcVersion"
     integrationTestRuntimeOnly "org.hsqldb:hsqldb:$hsqldbVersion"
     integrationTestRuntimeOnly "org.postgresql:postgresql:$postgresDriverVersion"
+    integrationTestRuntimeOnly "com.microsoft.sqlserver:mssql-jdbc:11.2.3.jre17"
```
nkovacsx: Shouldn't we have a property for this driver version instead of just hardcoding it?

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/PostgresHelper.kt
```diff
@@ -22,4 +24,11 @@ class PostgresHelper : AbstractDBHelper() {
 
     override val driverClass = "org.postgresql.Driver"
 
+    override fun createSchema(connection: Connection, schemaName: String): Pair<String, String> {
+        connection.use { conn ->
+            conn.prepareStatement("CREATE SCHEMA IF NOT EXISTS $schemaName;").execute()
```
nkovacsx: do we need to do this for Postgres? Shouldn't this just be a noop?
I know it will only be executed if it doesn't exist but still a code overhead.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/AbstractDBHelper.kt
```diff
@@ -60,6 +61,8 @@ abstract class AbstractDBHelper : DbUtilsHelper {
         return DbEntityManagerConfiguration(ds,showSql,true, DdlManage.NONE)
     }
 
+    abstract fun createSchema(connection: Connection, schemaName: String): Pair<String, String>
```
nkovacsx: Maybe this should be called something more explicit as it is quite confusing why a schema creation would return a `Pair<String, String>`. Probably `createSchemaAndUser` or `initialiseSchemaAndUser`?

libs/db/osgi-integration-tests/build.gradle
nkovacsx: I was looking at an older version sorry, I see this is fixed now.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/AbstractDBHelper.kt
```diff
@@ -0,0 +1,114 @@
...
+            }
+            val jdbcUrlCopy = if (rewriteBatchedInserts) {
+                "$jdbcUrl?currentSchema=$schemaName&reWriteBatchedInserts=true"
+            } else {
+                "$jdbcUrl?currentSchema=$schemaName"
+            }
+            logger.info("Using URL $jdbcUrlCopy".emphasise())
+            return createDataSource(driverClass,jdbcUrlCopy, credentials.first, credentials.second)
+        }
+        logger.info("Using URL $jdbcUrl".emphasise())
```
nkovacsx: Probably don't need to log this twice.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/PostgresHelper.kt
blsemo: What should be a no op here?

relyafi: Yeah I think we still need to create a schema, we were doing this before right? It's just that MS SQL has different syntax?

relyafi (dismissed):
LGTM

build.gradle
```diff
@@ -251,6 +251,10 @@ allprojects {
         inputs.property('postresUser', postgresUser)
         inputs.property('postgresPassword', postgresPassword)
 
+        inputs.property('mssqlPort', mssqlPort)
```
mbrkic-r3: Are we going to use all types of databases or just one of them (in which case we could use same properties e.g. `dbType`, `dbUser`, etc)?

libs/db/db-core/src/test/kotlin/PostgresEntityManagerConfigurationTest.kt
```diff
@@ -14,10 +14,12 @@ class PostgresEntityManagerConfigurationTest {
             on { create(any(), any(), any(), any(), any(), any(),any(), any(), any(), any(), any(), any()) } doReturn (mock())
         }
 
-        PostgresDataSourceFactory(dataSourceFactory).create(
+        createDataSource(
```
mbrkic-r3, line 17: Why don't we use `PostgresHelper.createDataSource()` here?

libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/persistence/DbUtils.kt
```diff
@@ -21,7 +21,8 @@ internal fun EntityManagerFactoryFactory.create(
     val dbSource = if (jdbcUrl == null || jdbcUrl.contains("hsqldb", ignoreCase = true)) {
         InMemoryDataSourceFactory().create(DB_MESSAGE_BUS)
     } else {
-        PostgresDataSourceFactory().create(
+        createDataSource(
```
mbrkic-r3, line 24: Why don't we use `PostgresHelper.createDataSource()` here?

libs/messaging/db-topic-admin-impl/src/main/kotlin/net/corda/libs/messaging/topic/factory/DBTopicUtilsFactory.kt
```diff
@@ -44,7 +44,8 @@ class DBTopicUtilsFactory @Activate constructor(
         } else {
             val username = config.getString(USER)
             val pass = config.getString(PASS)
-            PostgresDataSourceFactory().create(
+            createDataSource(
```
mbrkic-r3, line 47: Why don't we use `PostgresHelper.createDataSource()` here?

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -1,154 +1,60 @@
...
+    private val isPostgres = !System.getProperty("postgresPort").isNullOrBlank()
+    private val isMSSQL = !System.getProperty("mssqlPort").isNullOrBlank()
 
-    val isInMemory = System.getProperty(POSTGRES_PORT_PROPERTY).isNullOrBlank()
-
-    /**
-     * Returns Postgres DB name
-     */
-    fun getPostgresDatabase() = getPropertyNonBlank("postgresDb", "postgres")
+    private val utilsHelper: DbUtilsHelper = when {
```
mbrkic-r3: If only one should be set then it would be better to have generic parameters (that includes DB type).

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/AbstractDBHelper.kt
```diff
@@ -0,0 +1,127 @@
...
+                        jdbcUrl,
+                        user,
+                        password,
+                        maximumPoolSize = 1
+                    ).connection, schemaName, user, password
+                )
+            }
+
+            if (rewriteBatchedInserts) {
+                "$jdbcUrl?currentSchema=$schemaName&reWriteBatchedInserts=true"
```
mbrkic-r3: Isn't this Postgres specific?

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/HelperUtils.kt
```diff
@@ -0,0 +1,10 @@
+package net.corda.db.testkit.dbutilsimpl
+
+fun getPropertyNonBlank(key: String, defaultValue: String): String {
```
mbrkic-r3: Do we need this function, could we use `System.getProperty(key, defaultValue)` instead?

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/SQLServerHelper.kt
```diff
@@ -0,0 +1,54 @@
...
+    override fun createSchemaAndLogin(
+        connection: Connection,
+        schemaName: String,
+        user: String,
+        password: String
+    ): Pair<String, String> {
+        val schemaUser = "user_$schemaName"
+        val schemaPassword = "password${schemaName}123(!)"
+        connection.use { conn ->
+                conn.prepareStatement("IF NOT EXISTS ( SELECT  * FROM sys.schemas WHERE   name = N'$schemaName' )\n" +
```
mbrkic-r3: This might be more readable by using multiline string and `trimIndent()`.

libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/persistence/DbUtils.kt
blsemo: I don't think we want to use code from Testkit here? This is main, i.e. potentially prod code.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/AbstractDBHelper.kt
blsemo: Good point - that whole method needs more looking at.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
blsemo: I am not sure I understand the comment? What should be a generic parameter on what?

libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/persistence/DbUtils.kt
mbrkic-r3: Right, I've missed that. Wouldn't be better to have a class for Postgres that can be used for prod code packages so it's clear this doesn't work with all database types?

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
mbrkic-r3: e.g. instead of having properties "postgresPort" and "mssqlPort" to have generic property "dbPort", and then additionally, property "dbType" based on which we choose the helper below.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/HelperUtils.kt
blsemo: I've tried it - but it behaves differnetly - it returns a blank string if the property is defined without value rather than the default.

libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/persistence/DbUtils.kt
blsemo: We had that - but the only diff for different databases was just the driver class/jdbc URL, so having a differnent class literally wrapping two strings for each DB was deemed too repetitive.

gradle.properties
```diff
@@ -149,12 +150,17 @@ systemLambdaVersion=1.2.1
 #  `integrationTest` stage.
 # See libs/db/osgi-integration-tests/src/integrationTest/kotlin/net/corda/db/test/osgi/EntitiesInBundlesTest.kt
 # for an example of one of these tests.
-# The above test will fall back to HSQL (in-memory) when these properties are left empty.
-postgresHost=
-postgresPort=
-postgresDb=
-postgresUser=
-postgresPassword=
+# Only one set of these properties need to be set, either the postgres details or the mssql details. If neither set it falls back to HSQL (in-memory)
```
relyafi: This comment needs updating to reflect the new properties

components/uniqueness/backing-store-impl/src/backingStoreBenchmark/README.md
```diff
@@ -13,8 +13,9 @@ These tests can be run by executing the Gradle `backingStoreBenchmark` task, wit
...
 the following properties:
 
-- `postgresPort`: Setting this will run the tests against a Postgres DB instance, instead of using 
-   HSQLDB (the default). It is strongly recommended that you use Postgres DB, as HSQLDB is not a 
+- `databaseType`: Setting this will run the tests against the configured database type instance,
+   instead of using HSQLDB (the default). Currently supported options are POSTGRES and HSQL.
+   It is strongly recommended that you use Postgres DB, as HSQLDB is not a 
    production grade database, and is unlikely to provide realistic results. Additional Postgres
    properties, such as user, password etc. can also be specified. These are the same set of
    properties as used by other database based tests in this repo.
```
relyafi:
```suggestion
   production grade database, and is unlikely to provide realistic results. Additional
   properties, such as user, password etc. can also be specified. These are the same set of
   properties as used by other database based tests in this repo.
```

components/ledger/ledger-persistence/src/integrationTest/kotlin/net/corda/ledger/persistence/utxo/tests/UtxoLedgerMessageProcessorTests.kt
```diff
@@ -70,7 +70,7 @@ import java.util.UUID
  *
  *     docker run --rm --name test-instance -e POSTGRES_PASSWORD=password -p 5432:5432 postgres
  *
- *     gradlew integrationTest -PpostgresPort=5432
+ *     gradlew integrationTest -PdatabaseType=POSTGRES [-PdatabaseAdminPassword=<password> -PdatabaseAdminUser=<user>]
```
relyafi: This isn't an exhaustive list of arguments - might be better to just refer to the optional arguments defined in the gradle config?

libs/db/readme.md
```diff
@@ -50,17 +50,30 @@ Start container like so:
...
+   are `HSQL` (the default) and `POSTGRES`. There is also experimental support 
+   for MS SQL server (`MSSQL`).
+- `-PdatabasePort`: connection port for database servers. This will default to the
+   standard port for the chose DB server.
+- `-PdatabaseHost`: database host to connect to. Will default to `localhost`
+- `-PdatabaseAdminUser` and `-PdatabaseAdminPassword`: defaults depend on database type
+   chosen. This should match the respective docker command line examples.
+- `-PdatabaseName`: name of the database instance to use. Defaults depend on 
+   the chosen database type.
+
```
relyafi: This content looks good, but we might need to review the document structure, as this currently sits under a Postgres heading, and this content is not Postgres specific.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -1,154 +1,64 @@
...
+        DatabaseType.MSSQL -> SQLServerHelper()
+        DatabaseType.POSTGRES -> PostgresHelper()
+        DatabaseType.HSQL -> HSQLHelper()
+    }
 
-    /**
-     * Returns DB admin user
-     */
-    fun getAdminUser() = if (isInMemory) "sa" else getPropertyNonBlank("postgresUser", "postgres")
+    val isInMemory = utilsHelper.isInMemory()
```
relyafi: Does this need to be a DB specific attribute, or is this implied by being `DatabaseType.HSQL`?

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/HelperUtils.kt
relyafi: @blsemo think this still applies?

relyafi (commented):
Some minor comments.

Also, there may be follow-up actions required on merging this PR, as we are breaking backwards compatibility by removing the postgres specific system properties. We should probably put out a message to engineers to tell people about this change, and this will also likely require updating of Jenkinsfiles so the CI builds don't break.

driessamyn (dismissed):
LGTM

wzur-r3 (dismissed):
LGTM

blsemo:
Closing to re-open clean PR against 5.2

PR #4525, opened 2023-08-24
ES-1230: Publish second Combined Worker jar with Kafka bus enabled

mcgovc:
Add new maven publication for `kafkaCombinedWorker` which is the kafak message bus implementation of the combined worker, only publish this if the property `busImpl` is present.

Change also needed in `corda.common-app.gradle` to ensure we do not clobber the original combined worker as with out this both would be re published with the kafka implementation 

Associated pipeline change - https://github.com/corda/corda-shared-build-pipeline-steps/pull/1107

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,4 @@
-@Library('corda-shared-build-pipeline-steps@connelm/ES-1230/publish-kafka-combined-worker') _
+@Library('corda-shared-build-pipeline-steps@5.1') _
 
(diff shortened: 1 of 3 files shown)


ronanbrowne (dismissed):
LGTM

yarivus (dismissed):
LGTM

mcgovc (commented):
LGTM - on changes made by @ronanbrowne

yarivus (approved):
LGTM

PR #4532, opened 2023-08-25
ES-597: updates for preinstall job

wzur-r3:
* added an option to overwrite `serviceAccount` for pre-install job, that is required to enable that job on an Openshift cluster
* added an option to disable creation of a `serviceAccount` for pre-install job, that is required to enable that job on an Openshift cluster, which comes with such `serviceAccount` pre-configured
* updated `values.yaml` with default values for newly created options
* updated schema for newly created options
* removed trailing spaces

### DIFF
--- a/charts/corda-lib/templates/_bootstrap.tpl
+++ b/charts/corda-lib/templates/_bootstrap.tpl
@@ -1,7 +1,7 @@
 {{/*
  Create the name of the service account to use for Preinstall Checks
  */}}
-{{- define "corda.bootstrapPreinstall.ServiceAccountName" -}}
+{{- define "corda.bootstrapPreinstallServiceAccountName" -}}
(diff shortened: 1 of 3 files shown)


charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -1,3 +1,14 @@
+{{/*
+ Create the name of the service account to use for Preinstall Checks
+ */}}
+{{- define "corda.bootstrapPreinstall.ServiceAccountName" -}}
```
davidcurrie: Nit: we tend to just use a single period to separate the chart designator from the rest of the function name:
```suggestion
{{- define "corda.bootstrapPreinstallServiceAccountName" -}}
```

davidcurrie (dismissed):
one nit on the function name but otherwise LGTM

davidcurrie (approved):
LGTM

PR #4534, opened 2023-08-28
CORE-16653: Re-enable the start flow acceptance tests

JamesHR3:
**Problem description**

As part of the removing wakeup events work, the acceptance tests were temporarily disabled, as fixing these was likely to involve a lot of code churn. These need to be re-enabled. The first set to look at is the start flow acceptance tests.

**Solution**

The acceptance test infrastructure requires some changes to support potentially running the fiber multiple times:

- Allow multiple flow IO requests to be declared as output from the fiber.
- Change the fake fiber to output each declared IO request in turn, so that multiple runs of the fiber can be simulated.
- Remove references to feeding wakeup events into the flow pipeline.
- Remove references to expecting wakeup events to be returned from the pipeline.

Note that in some cases the test behaviour also changes, due to the fiber now running through multiple steps in a single test.

**Testing**

The start flow acceptance tests should now pass.

### DIFF
--- a/components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/StartFlowTest.kt
+++ b/components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/StartFlowTest.kt
@@ -15,7 +15,6 @@ import org.osgi.test.junit5.service.ServiceExtension
 
 @ExtendWith(ServiceExtension::class)
 @Execution(ExecutionMode.SAME_THREAD)
-@Disabled
 class StartFlowTest : FlowServiceTestBase() {
 
     @Test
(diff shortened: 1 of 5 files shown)


components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/StartFlowTest.kt
```diff
@@ -106,32 +97,27 @@ class StartFlowTest : FlowServiceTestBase() {
...
 
-        then {
-            expectOutputForFlow(FLOW_ID1) {
-                noFlowEvents()
-                checkpointHasRetry(2)
-                flowFiberCacheDoesNotContainKey(BOB_HOLDING_IDENTITY, REQUEST_ID1)
-            }
-        }
-
         // Now fix the issue and expect the wake-up event to
```
jujoramos, line 105: Nitpick: we don't have wake ups anymore.

jujoramos (approved):
LGTM

LWogan (approved):
Makes sense to me

PR #4536, opened 2023-08-29
CORE-16653: Re-enable further acceptance tests

JamesHR3:
**Problem description**

The acceptance tests were disabled as part of removing `Wakeup` events from the flow pipeline. These need to be re-enabled.

**Solution**

This PR fixes the external event, flow finished and flow failed acceptance tests. This leaves just acceptance tests that depend on the session functionality to re-enable. As there are changes being made elsewhere to change how the session layer works, it is likely that these tests will need some collaboration to enable.

**Testing**

Newly enabled acceptance tests pass.

### DIFF
--- a/components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/SendAcceptanceTest.kt
+++ b/components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/SendAcceptanceTest.kt
@@ -176,7 +176,7 @@ class SendAcceptanceTest : FlowServiceTestBase() {
 
     @Test
     @Disabled
-    fun `Calling 'send' on a session that is confirmed sets a wakeup event and data message`() {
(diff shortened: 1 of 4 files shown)


LWogan (approved):
LGTM

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/tests/SendAcceptanceTest.kt
```diff
@@ -171,6 +175,7 @@ class SendAcceptanceTest : FlowServiceTestBase() {
     }
 
     @Test
+    @Disabled
     fun `Calling 'send' on a session that is confirmed sets a wakeup event and data message`() {
```
jujoramos: Can we remove `wakeup` from the method name?.

jujoramos (approved):

PR #4538, opened 2023-08-29
CORE-8976: added notary match check

jennyang-r3:
## Overview
This PR added a check if the notary in tx is valid notary that is the same as the current notary in the network.
### Test Plan
The unit test `NonValidatingNotaryServerFlowImpl` covers this scenario.

**Manual test**
I passed notary "O=AnotherNotaryService, L=London, C=GB" and tweaked `notaryRepresentative`(notary vnode) in `NonValidatingNotaryClientFlowImpl` with wrong notary as below:
 ```kotlin        
val session = flowMessaging.initiateFlow(MemberX500Name.parse("CN=Notary-afbe5058-36ad-4636-8f79-0296741dee9s, OU=Application, O=R3, L=London, C=GB"))
```
then ran MultipleNotariesTests.`Multiple Notaries - sanity check of multiple notaries by creating two states with different notaries for each()` where two notaries are available in the network.

I was able to see the expected error message in the log:
```
2023-09-01 14:47:35.769 [single-threaded-scheduled-executor-1] WARN  com.r3.corda.notary.plugin.nonvalidating.server.NonValidatingNotaryServerFlowImpl {corda.client.id=0de2a179-31e8-4392-b5d5-825bab5e90db-INITIATED, flow.id=e83c0c7d-b848-4555-a17c-04eb29d49217, session.event.id=0de2a179-31e8-4392-b5d5-825bab5e90db-INITIATED, vnode.id=EF5CF6FA0F4D} - **Error while processing request from client. Cause: java.lang.IllegalArgumentException: Given notary service = O=AnotherNotaryService, L=London, C=GB is invalid **
```

### DIFF
--- a/notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
+++ b/notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
@@ -78,7 +78,7 @@ class NonValidatingNotaryServerFlowImpl() : ResponderFlow {
 
             val txDetails = validateRequest(requestPayload)
 
-            getCurrentNotaryAndValidateNotary(txDetails)
+            validateTransactionNotaryAgainstCurrentNotary(txDetails)
 
             if (logger.isTraceEnabled) {
                 logger.trace("Received notarization request for transaction {}", txDetails.id)
@@ -126,14 +126,15 @@ class NonValidatingNotaryServerFlowImpl() : ResponderFlow {
      * This function will validate selected notary is valid notary to notarize.
      * */
     @Suspendable
-    private fun getCurrentNotaryAndValidateNotary(txDetails: NonValidatingNotaryTransactionDetails) {
-        val currentNotaryMemberProvidedCtx = memberLookup.myInfo().memberProvidedContext
-        val currentNotaryServiceName = currentNotaryMemberProvidedCtx.parse(NOTARY_SERVICE_NAME, MemberX500Name::class.java)
+    private fun validateTransactionNotaryAgainstCurrentNotary(txDetails: NonValidatingNotaryTransactionDetails) {
+        val currentNotaryServiceName = memberLookup
+            .myInfo()
+            .memberProvidedContext
+            .parse(NOTARY_SERVICE_NAME, MemberX500Name::class.java)
 
-        val payloadNotaryServiceName = txDetails.notaryName
-
-        require(currentNotaryServiceName == payloadNotaryServiceName) {
-            "Given notary service = $payloadNotaryServiceName is invalid"
+        require(currentNotaryServiceName == txDetails.notaryName) {
+            "Notary service on the transaction ${txDetails.notaryName} does not match the notary service represented" +
+                    " by this notary virtual node (${currentNotaryServiceName})"
         }
     }
 
--- a/notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
+++ b/notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
@@ -137,6 +137,11 @@ class NonValidatingNotaryServerFlowImpl() : ResponderFlow {
     @Suspendable
     private fun extractParts(requestPayload: NonValidatingNotarizationPayload): NonValidatingNotaryTransactionDetails {
         val filteredTx = requestPayload.transaction as UtxoFilteredTransaction
+
+        if (filteredTx.notaryKey != requestPayload.notaryKey) {
+            throw IllegalStateException("Unvalidated notary was used")
+        }
+
         // The notary component is not needed by us but we validate that it is present just in case
         requireNotNull(filteredTx.notaryName) {
             "Notary name component could not be found on the transaction"
--- a/notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
+++ b/notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
@@ -7,10 +7,12 @@ import com.r3.corda.notary.plugin.nonvalidating.api.NonValidatingNotarizationPay
 import net.corda.v5.application.flows.CordaInject
 import net.corda.v5.application.flows.InitiatedBy
 import net.corda.v5.application.flows.ResponderFlow
+import net.corda.v5.application.membership.MemberLookup
 import net.corda.v5.application.messaging.FlowSession
 import net.corda.v5.application.uniqueness.model.UniquenessCheckResultSuccess
 import net.corda.v5.base.annotations.Suspendable
 import net.corda.v5.base.annotations.VisibleForTesting
+import net.corda.v5.base.types.MemberX500Name
 import net.corda.v5.ledger.common.transaction.TransactionSignatureService
 import net.corda.v5.ledger.utxo.StateAndRef
 import net.corda.v5.ledger.utxo.StateRef
@@ -19,6 +21,7 @@ import net.corda.v5.ledger.utxo.transaction.filtered.UtxoFilteredTransaction
 import net.corda.v5.ledger.utxo.uniqueness.client.LedgerUniquenessCheckerClientService
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
+import java.security.PublicKey
 
 /**
  * The server-side implementation of the non-validating notary logic.
@@ -30,6 +33,9 @@ class NonValidatingNotaryServerFlowImpl() : ResponderFlow {
 
     private companion object {
         private val logger: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+
+        const val NOTARY_KEYS = "corda.notary.keys"
+        const val NOTARY_SERVICE_NAME = "corda.notary.service.name"
     }
 
     @CordaInject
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt [hunk 1]
```diff
@@ -30,6 +32,8 @@ class NonValidatingNotaryServerFlowImpl() : ResponderFlow {
 
     private companion object {
         private val logger: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+
+        const val NOTARY_SERVICE_NAME = "corda.notary.service.name"
```
relyafi, line 36: Maybe makes sense to use the existing definition in `MemberInfoExtension` rather than duplicating?

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
```diff
@@ -111,6 +122,21 @@ class NonValidatingNotaryServerFlowImpl() : ResponderFlow {
...
+     * This function will validate selected notary is valid notary to notarize.
+     * */
+    @Suspendable
+    private fun getCurrentNotaryAndValidateNotary(txDetails: NonValidatingNotaryTransactionDetails) {
+        val currentNotaryMemberProvidedCtx = memberLookup.myInfo().memberProvidedContext
+        val currentNotaryServiceName = currentNotaryMemberProvidedCtx.parse(NOTARY_SERVICE_NAME, MemberX500Name::class.java)
+
+        val payloadNotaryServiceName = txDetails.notaryName
+
+        require(currentNotaryServiceName == payloadNotaryServiceName) {
```
relyafi:
```suggestion
        val currentNotaryServiceName = memberLookup
            .myInfo()
            .memberProvidedContext
            .parse(NOTARY_SERVICE_NAME, MemberX500Name::class.java)

        require(currentNotaryServiceName == txDetails.notaryName) {
```

Removes two unnecessary temporary variables that are only used once, and aren't adding much value imo.

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
```diff
@@ -111,6 +122,21 @@ class NonValidatingNotaryServerFlowImpl() : ResponderFlow {
...
+            "Given notary service = $payloadNotaryServiceName is invalid"
```
relyafi:
```suggestion
            "Notary service on the transaction $payloadNotaryServiceName does not match the notary service represented by this notary virtual node (${currentNotaryServiceName})"
```

Current message is a bit misleading (it may well be a valid notary service). This text is more accurate and provides a more specific / helpful error.

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/test/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImplTest.kt
```diff
@@ -43,6 +45,8 @@ import java.time.Instant
 class NonValidatingNotaryServerFlowImplTest {
 
     private companion object {
+        const val NOTARY_SERVICE_NAME = "corda.notary.service.name"
```
relyafi, line 48: Maybe makes sense to use the existing definition in MemberInfoExtension rather than duplicating?

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
```diff
@@ -111,6 +122,21 @@ class NonValidatingNotaryServerFlowImpl() : ResponderFlow {
         }
     }
 
+    /**
+     * This function will validate selected notary is valid notary to notarize.
+     * */
+    @Suspendable
+    private fun getCurrentNotaryAndValidateNotary(txDetails: NonValidatingNotaryTransactionDetails) {
```
blsemo: I don't like the name of this function - it's quite long but still doesn't tell me what the function does. So either something short (`checkNotaryName`) or very descriptive (`validateTransactionNotaryAgainstCurrentNotary`)

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
blsemo: Should this also checknotary keys?

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/test/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImplTest.kt
```diff
@@ -322,7 +329,12 @@ class NonValidatingNotaryServerFlowImplTest {
             on { size } doReturn 1
         }
 
-        // 2. Check if any filtered transaction data should be overwritten
+        // 2. Get current notary and parse its data
+        whenever(mockMemberLookup.myInfo()).thenReturn(notaryInfo)
+        whenever(mockMemberLookup.myInfo().memberProvidedContext).thenReturn(memberProvidedContext)
```
blsemo: Wouldn't it make sense to do the mocking on the objects that are defined already anyway? That way, adding or removing a temporary val won't randomly break a test.
```suggestion
        whenever(notaryInfo.memberProvidedContext).thenReturn(memberProvidedContext)
```

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/test/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImplTest.kt
jennyang-r3: Likewise, that is also not available in this module

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/test/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImplTest.kt
jennyang-r3: Right, that makes sense to me.

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
jennyang-r3: @nkovacsx and @relyafi talked about it earlier today. IIUC checking the notary name would be enough since it would fail later if it went to the wrong notary. This is just a sanity check afaik

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
jennyang-r3: This constant is not importable in this module, would you suggest add the dependency to this module?

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/test/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImplTest.kt
relyafi: Can that be resolved by adding the appropriate module dependency in the gradle file?

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/main/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImpl.kt
relyafi: Hmm, my first thought was yes, but this is a `runtime-os` module. So that could cause issues if we choose to split the plugin out into its own repo in future, because we may not publish that API externally.

I'd say leave it as-is for now.

notary-plugins/notary-plugin-non-validating/notary-plugin-non-validating-server/src/test/kotlin/com/r3/corda/notary/plugin/nonvalidating/server/NonValidatingNotaryServerFlowImplTest.kt
relyafi: Ok, feel free to drop.

lankydan (approved):
LGTM

relyafi (approved):

PR #4540, opened 2023-08-29
CORE-16388 Removal of old 'InteropAliasProcessor' and redundant supporting classes

ac101m:
# Changes
- Delete interop alias processor.
- Add comment for future maintainers to (currently disabled) tests.


ac101m:
@szymonsztuka This ended up being very simple, I can't see any other references to the deleted code, let me know if there is anything you think I have missed.

components/interop/interop-service/src/integrationTest/kotlin/net/corda/interop/service/integration/InteropServiceIntegrationTest.kt
```diff
@@ -164,9 +162,6 @@ class InteropServiceIntegrationTest {
         val publisher = publisherFactory.createPublisher(PublisherConfig("client1"), bootConfig)
         // Test config updates don't break Interop Service
         republishConfig(publisher)
-        // Need to publish at least 2 identities in order to trigger InteropAliasProcessor to create synthetic aliases
-        publisher.publish(listOf(createHostedAliasIdentity(sourceIdentity.toCorda())))
```
szymonsztuka, line 168: I would maybe leave the hit (as the test is turn off now) for future fixing the test: `//The original test published 2 hosted identities to automatically publish derived members of interoperability group (for topics MemberInfo and Hosted Identities).` or some description in that ticket to see what was removed

szymonsztuka (approved):

components/interop/interop-service/src/integrationTest/kotlin/net/corda/interop/service/integration/InteropServiceIntegrationTest.kt
ac101m: Good idea

szymonsztuka (approved):

PR #4541, opened 2023-08-29
CORE-16434 - Create multi-version tests for membership groups - without the platform upgrade

nikinagy:
Added new clusterInfo for 5.0 cluster deployment used in e2e testing. Had to add deprecated endpoints so that we can use them in the e2e tests.
Wired in REST API version information to ClusterInfo, so that we can easily decide what endpoint to use to for different deployments.


testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterBuilder.kt
```diff
@@ -18,7 +18,7 @@ import java.time.Instant
 class ClusterBuilder {
     
     internal companion object {
-        val REST_API_VERSION_PATH = RestApiVersion.C5_1.versionPath
+        var REST_API_VERSION_PATH = ""
```
yift-r3, line 21: Do we want this in the companion object?

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterBuilder.kt
```diff
@@ -88,17 +92,33 @@ class ClusterBuilder {
...
     fun importCertificate(resourceName: String, usage: String, alias: String) =
-        uploadCertificateResource("/api/$REST_API_VERSION_PATH/certificate/cluster/$usage", resourceName, alias)
-
-    fun importCertificate(file: File, usage: String, alias: String) =
-        uploadCertificateFile("/api/$REST_API_VERSION_PATH/certificate/cluster/$usage", file, alias)
+        if (REST_API_VERSION_PATH == RestApiVersion.C5_0.versionPath) {
+            // Used to test RestApiVersion.C5_0 CertificateRestResource, remove after LTS
+            deprecatedImportCertificate(resourceName, usage, alias)
+        } else {
+            uploadCertificateResource("/api/$REST_API_VERSION_PATH/certificate/cluster/$usage", resourceName, alias)
```
yift-r3: Maybe we should have some kind of private function `RestApiVersion.certificatePath()` that returns wither `certificate` or `certificates` to avoid all the duplications?

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterInfo.kt
```diff
@@ -80,18 +83,29 @@ data class P2PEndpointInfo(
  */
 object ClusterAInfo : ClusterInfo() {
     override val id = "A"
+    override val restApiVersion = RestApiVersion.C5_1
```
yift-r3: Should we control this from the environment variable as well?

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterBuilder.kt
nikinagy: Yes, as it is being used in other utils too eg: MembershipUtils.kt

yift-r3: So, changing this will change all those utilities? This seems wrong. 

(Maybe we should fix it in another pull request)

williamvigorr3: Before we could quickly bump this when we change versions. I think it is good to have a default defined in one place, with the latest version. We could move it to the constructor, with a default argument. Multi Version Testing is a special case.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterInfo.kt
nikinagy: addressed in [6714e40](https://github.com/corda/corda-runtime-os/pull/4541/commits/6714e40473b6eba753e7f3df06dea431ee78d5e1)

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterBuilder.kt
nikinagy: addressed in [6714e40](https://github.com/corda/corda-runtime-os/pull/4541/commits/6714e40473b6eba753e7f3df06dea431ee78d5e1)

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterBuilder.kt
nikinagy: I added the default to ClusterInfo in my latest commit. It can be manipulated via setting env variables.

yift-r3: I would still prefer to move it to be a member. 

(But I understand it's not part of this change)

yift-r3 (approved):
LGTM

PR #4543, opened 2023-08-30
CORE-16623 : Changes for p2p parameters

SaurabhR3:
1. Added new method getP2PParameters() in GroupPolicyProvider.kt
2. Changes in link manager to replace the getGroupPolicy() call with getP2PParameters()
3. Changes in GroupPolicyExt.kt for networkType and protocolModes
4. Fixed UTs
5. Reverted removed restrictions in GroupPolicy interface

### DIFF
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/common/MessageConverter.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/common/MessageConverter.kt
@@ -177,13 +177,13 @@ class MessageConverter {
             message: InboundUnauthenticatedMessage,
             source: HoldingIdentity,
             destMemberInfo: MemberInfo,
-            groupPolicy: GroupPolicy,
+            networkType: NetworkType,
         ): LinkOutMessage {
             return createLinkOutMessage(
                 message,
                 source,
                 destMemberInfo,
-                groupPolicy.networkType,
+                networkType,
                 )
         }
 
@@ -232,10 +232,10 @@ class MessageConverter {
                         "which is not in the network map. The message was discarded.")
                 return null
             }
-            val groupPolicy = groupPolicyProvider.getGroupPolicy(source)
-            if (groupPolicy == null) {
+            val p2pParams = groupPolicyProvider.getP2PParameters(source)
+            if (p2pParams == null) {
                 logger.warn(
-                    "Could not find the group info in the " +
+                    "Could not find the p2p parameters in the " +
                         "GroupPolicyProvider for our identity = $source. The message was discarded."
                 )
                 return null
@@ -245,7 +245,7 @@ class MessageConverter {
                 result,
                 source,
                 destMemberInfo,
-                groupPolicy.networkType,
+                p2pParams.networkType,
             )
         }
 
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/common/MessageConverter.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/common/MessageConverter.kt
@@ -15,7 +15,6 @@ import net.corda.data.p2p.crypto.AuthenticatedEncryptedDataMessage
 import net.corda.membership.grouppolicy.GroupPolicyProvider
 import net.corda.membership.lib.MemberInfoExtension.Companion.endpoints
 import net.corda.membership.lib.MemberInfoExtension.Companion.holdingIdentity
-import net.corda.membership.lib.grouppolicy.GroupPolicy
 import net.corda.membership.read.MembershipGroupReaderProvider
 import net.corda.p2p.crypto.protocol.ProtocolConstants
 import net.corda.p2p.crypto.protocol.api.AuthenticatedEncryptionSession
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/delivery/InMemorySessionReplayer.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/delivery/InMemorySessionReplayer.kt
@@ -140,7 +140,7 @@ internal class InMemorySessionReplayer(
             return
         }
 
-        val networkType = groupPolicyProvider.getGroupPolicy(messageReplay.sessionCounterparties.ourId)?.networkType
+        val networkType = groupPolicyProvider.getP2PParameters(messageReplay.sessionCounterparties.ourId)?.networkType
         if (networkType == null) {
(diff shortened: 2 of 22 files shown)


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/delivery/InMemorySessionReplayer.kt
```diff
@@ -140,7 +140,7 @@ internal class InMemorySessionReplayer(
             return
         }
 
-        val networkType = groupPolicyProvider.getGroupPolicy(messageReplay.sessionCounterparties.ourId)?.networkType
+        val networkType = groupPolicyProvider.getP2PParameters(messageReplay.sessionCounterparties.ourId)?.networkType
```
szymonsztuka, line 143: I think we should create new interface `P2PParametersProvider` with just a single method `getP2pParameters`, as it looks a bit illogical here and in other cases that the object is called `groupPolicy` and has many methods but we in this context are using it only for P2pP parameters.
Saying that the actual implementation class (`GroupPolicyProviderImpl`) can stay the same and extend from 2 interfaces  (the existing `GroupPolicyProvider` and the new one `GroupP2PParametersProvider`).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/common/MessageConverter.kt
```diff
@@ -232,10 +231,10 @@ class MessageConverter {
                         "which is not in the network map. The message was discarded.")
                 return null
             }
-            val groupPolicy = groupPolicyProvider.getGroupPolicy(source)
-            if (groupPolicy == null) {
+            val p2pParams = groupPolicyProvider.getP2PParameters(source)
+            if (p2pParams == null) {
                 logger.warn(
-                    "Could not find the group info in the " +
+                    "Could not find the p2p parameters in the " +
```
szymonsztuka, line 237: I think all changes to warning in the logs in this PR are very tricky, because anyway we always need to fetch internally `"GroupPolicy"` as a box holding our data and something which exists standalone in CPI files/on Kafka queues.  When we just log  warning "p2pParameters are not found" what does it mean?  

- Have we found Group Policy but that group policy doesn't have p2pParameters

-  or we couldn't find a group policy at all. this is fundamental difference. 

I think we know that in 99% (?) the reason is lack of a group policy. I would suggest to revert this and all other warnings to saying as before  `"Could not find the group info in the..."`, maybe (?) adding something that it was while trying to retrieve p2p parameters.  Or could we at this level distinguish when we has lacked group policy or just group policy missed p2p params or had mangled params?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/delivery/InMemorySessionReplayer.kt
jmacmahonr3: No, this actually goes against the Network team's intensions. They want to split current group policy into two subsections which are `MgmPolicy `& `P2pParams` which are obtainable from the same interface. We've done enough of the work to suit us but they will likely do the rest in time

szymonsztuka: Thanks for the explanation, it would be better to separate "mgm policy" and "p2p params" concerns into a distinct interfaces so, the consumer has more narrow interface scope, but it the Network team prefers the the other solution that's fine.

jmacmahonr3: When they offer us money to do it i'll prioritize it ;)

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/common/MessageConverter.kt
SaurabhR3: Actually we are calling api `getP2PParameters()` so here at consumer layer, important thing to log is `p2pParameters are not found`, in order to know the reason we need to check the logs in `getP2PParameters()` method. 
Inside method we are calling another method `getGroupPolicy()` -> `parseGroupPolicy()` this method logs warning if metadata is not available from where we read group policy for holding identity.

If the group policy against holding identity not found then we search for interop group policy and if that is not found then we logs another warning for `interop group policy not found` 

So I think at the api consumer level it is correct information to log and if we need to know the reason then we need to deep dive in the next level log and we can find the reason.

I think if we revert the logging then it does not sound good that at consumer level we are fetching p2p parameters and log a warning "Could not find the group info in the...". 

Another important thing I need check that is there any log watcher for these waning messages, if there is any then certainly we need to revert this warning or change the log watcher.

components/membership/group-policy-impl/src/main/kotlin/net/corda/membership/impl/grouppolicy/GroupPolicyProviderImpl.kt
```diff
@@ -102,6 +103,21 @@ class GroupPolicyProviderImpl @Activate constructor(
         listeners.put(name, listener)?.stop()
     }
 
+    override fun getP2PParameters(holdingIdentity: HoldingIdentity): GroupPolicy.P2PParameters? {
+        val groupPolicy = getGroupPolicy(holdingIdentity)
+        if (groupPolicy == null ) {
+            logger.warn("Could not get the group policy for holding identity [${holdingIdentity}]. Checking interop group policy.")
```
jmacmahonr3, line 109: Update this logger message

components/membership/group-policy-impl/src/main/kotlin/net/corda/membership/impl/grouppolicy/GroupPolicyProviderImpl.kt
```diff
@@ -102,6 +103,21 @@ class GroupPolicyProviderImpl @Activate constructor(
...
+            val interopGroupPolicy: String? = interopGroupPolicyReadService.getGroupPolicy(holdingIdentity.groupId)
+            if (interopGroupPolicy == null) {
+                logger.warn("Could not get interop group policy for holding identity [${holdingIdentity}].")
+                return null
+            }
+            return groupPolicyParser.parseInteropGroupPolicy(interopGroupPolicy).p2pParameters
+        } else {
```
jmacmahonr3, line 116: Is the indentation off here? It's confusing to read.

jmacmahonr3 (changes requested):
Have we done a diff against 5.1 to make sure we've restored all unwanted modification to grouppolicy and parsing?

PR #4544, opened 2023-08-30
CORE-16436: Add tests for Member Lookup Command

r3-arish:
This PR introduces tests for the network plugin Lookup command. The tests would target the Combined Worker (similar to the existing smoke-tests). The tests are invoking the plugin classes directly.

## Test Results
Screenshot 2023-09-04 at 17 03 23


tools/plugins/network/build.gradle
```diff
@@ -37,6 +37,8 @@ dependencies {
     testImplementation "org.pf4j:pf4j:${pf4jVersion}"
     testImplementation "org.junit.jupiter:junit-jupiter:${junit5Version}"
     testImplementation "com.github.stefanbirkner:system-lambda:1.2.1"
+    testImplementation project(path: ':testing:e2e-test-utilities')
```
YashNabar:
```suggestion
    testImplementation project(':testing:e2e-test-utilities')
```

tools/plugins/network/build.gradle
```diff
@@ -37,6 +37,8 @@ dependencies {
...
+    testImplementation 'ch.qos.logback:logback-classic:1.2.3'
```
YashNabar: Leftover from testing maybe?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -202,7 +202,7 @@ abstract class BaseOnboard : Runnable, RestCommand() {
         }
     }
 
-    protected val holdingId: String by lazy {
+    val holdingId: String by lazy {
```
YashNabar: We shouldn't change this for the test. We can construct the holding identity in the tests using the name and group ID if required

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,89 @@
...
+@Retention(AnnotationRetention.RUNTIME)
+annotation class SkipInitialization
+
+class MemberLookupTest {
+
+    private lateinit var outputStub: OutputStub
+    private lateinit var memberLookup: MemberLookup
+    private lateinit var onboardMgm: OnboardMgm
+
+    @BeforeEach
```
YashNabar: This can be a `@BeforeAll` so we are not onboarding for every test case. We should also include a couple of members here

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        onboardMgm.name = "O=Mgm, L=London, C=GB"
+        onboardMgm.run()
+    }
+
+    @Test
+    @SkipInitialization
+    fun `test member lookup command with status filter`() {
+        memberLookup.holdingIdentityShortHash = onboardMgm.holdingId
+        memberLookup.status = listOf("ACTIVE", "SUSPENDED")
+        memberLookup.run()
```
YashNabar: You can use `CommandLine.execute()` for invoking the command (also in other places). You could refer to the package plugin tests for this

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
+        onboardMgm.name = "O=Mgm, L=London, C=GB"
+        onboardMgm.run()
+    }
+
+    @Test
+    @SkipInitialization
+    fun `test member lookup command with status filter`() {
+        memberLookup.holdingIdentityShortHash = onboardMgm.holdingId
+        memberLookup.status = listOf("ACTIVE", "SUSPENDED")
```
YashNabar: We should test these cases separately and check we get the expected output e.g. filtering on SUSPENDED should return no results

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        memberLookup.run()
+
+        val mgmContext = outputStub.printedOutput?.get(0)?.get("mgmContext")
+        assertEquals("ACTIVE", mgmContext?.get("corda.status")?.asText() ?: "")
+    }
+
+    @Test
+    @SkipInitialization
+    fun `test member lookup command with X500 name`() {
```
YashNabar: We should clarify in the name that this test case exercises the use of the default group ID (from file).
We should also add a test case where the group ID is provided by the user

YashNabar (commented):
Left some initial comments. Can you add more test cases around the different command line options we have e.g. filtering based on X.500 attributes? (I have also suggested a few in the comments below)

tools/plugins/network/build.gradle
```diff
@@ -37,6 +37,8 @@ dependencies {
     testImplementation "org.pf4j:pf4j:${pf4jVersion}"
     testImplementation "org.junit.jupiter:junit-jupiter:${junit5Version}"
     testImplementation "com.github.stefanbirkner:system-lambda:1.2.1"
+    testImplementation project(':testing:e2e-test-utilities')
+    testImplementation 'ch.qos.logback:logback-classic:1.2.9'
```
charlieR3: What is this dependency used for? Is it new for the runtime os repo?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/enums/MemberStatus.kt
```diff
@@ -0,0 +1,8 @@
+package net.corda.cli.plugins.network.enums
+
+import net.corda.membership.lib.MemberInfoExtension.Companion.MEMBER_STATUS_ACTIVE
+import net.corda.membership.lib.MemberInfoExtension.Companion.MEMBER_STATUS_SUSPENDED
+enum class MemberStatus(val value: String) {
+    ACTIVE(MEMBER_STATUS_ACTIVE),
+    SUSPENDED(MEMBER_STATUS_SUSPENDED)
+}
```
charlieR3: Do you need to introduce this enum? Could you use the constants `MEMBER_STATUS_ACTIVE` and `MEMBER_STATUS_SUSPENDED` directly instead?

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,215 @@
...
+    companion object {
+        private lateinit var memberLookup: MemberLookup
+        private lateinit var onboardMgm: OnboardMgm
+        private lateinit var outputStub: OutputStub
+
+        private val CLI_PARAMS = arrayOf(
+            "--target=${DEFAULT_CLUSTER.rest.uri}",
+            "--user=${DEFAULT_CLUSTER.rest.user}",
+            "--password=${DEFAULT_CLUSTER.rest.password}",
+            "--insecure=${true}"
```
charlieR3:
```suggestion
            "--insecure=true"
```

charlieR3: Probably don't need the string interpolation here.

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,215 @@
...
+        )
+
+        private val X500 = arrayOf(
```
charlieR3: Could alternatively be `MemberX500Name.parse("CN=Alice, OU=R3 Test, O=Mgm, L=London, ST=Tottenham, C=GB")`

And then below you can reference the specific property rather than relying on the order of elements in the array. e.g. `X500.toString()`, `X500.commonName`, `X500.organizationUnit`, etc.
That would be easier for anyone looking at this test in the future

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,215 @@
...
+            "-h=${HoldingIdentityUtils.getHoldingIdentity(
+                null,
+                onboardMgm.name,
+                null
+            )}",
+            *CLI_PARAMS
+        )
+
+        val mgmContext = outputStub.printedOutput?.get(0)?.get("mgmContext")
+        assertEquals(ACTIVE.value, mgmContext?.get(MemberInfoExtension.STATUS)?.asText() ?: "")
```
charlieR3: Is ` ?: ""` necessary? If it's null the assertion should fail anyway, which is what we want right?

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,215 @@
...
+            *CLI_PARAMS
+        )
+
+        assertEquals(JsonNodeFactory.instance.arrayNode(), outputStub.printedOutput)
+    }
+
+    @Test
+    fun `test member lookup command with Common Name (CN)`() {
+        CommandLine(memberLookup).execute(
+            "-cn=${X500[0].split("=")[1]}",
```
charlieR3: I think this could benefit from my comment above in regards to representing the name as a `MemberX500Name` type.

`X500.commonName` would be much clearer I think

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
charlieR3: Same comment on all the tests below

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
charlieR3: I think this would align better with naming conventions
```suggestion
        private val x500 = arrayOf(
```

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,215 @@
...
+            "-h=${HoldingIdentityUtils.getHoldingIdentity(
+                null,
+                onboardMgm.name,
+                null
+            )}",
+            *CLI_PARAMS
+        )
+
+        val memberContext = outputStub.printedOutput?.get(0)?.get("memberContext")
+        assertEquals(onboardMgm.name, memberContext?.get(MemberInfoExtension.PARTY_NAME)?.asText() ?: "")
```
charlieR3: It could be useful to add an extension function since this is repeated a lot. Something like: 

```
fun OutputStub.getFirstPartyName(): String? {
    return printedOutput?.get(0)?.get("memberContext")?.get(MemberInfoExtension.PARTY_NAME)?.asText()
}
```
Then each test could have
```
assertEquals(onboardMgm.name, outputStub.getFirstPartyName())
```

tools/plugins/network/build.gradle
r3-arish: I was getting a logger error, adding dependency fixed it.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/enums/MemberStatus.kt
r3-arish: anticipating for using this enum type for the status variable in the `MemberLookup` class.

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
r3-arish: this is truly amazing, didn't know we had it - I was actually thinking of making one in the `utils`

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -13,45 +13,46 @@ import net.corda.cli.plugins.network.utils.HoldingIdentityUtils
...
     }
 
+    @BeforeEach
+    fun beforeEach() {
+        outputStub = OutputStub()
+        memberLookup = MemberLookup(outputStub)
+    }
+
     @Test
+    @Order(1)
```
charlieR3: I'm wondering now do you need a defined order for this? I think they should be able to run in any order and if so the Order annotation could be removed

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -200,11 +203,14 @@ class MemberLookupTest {
...
-        assertEquals(onboardMgm.name, memberContext?.get(MemberInfoExtension.PARTY_NAME)?.asText() ?: "")
+        assertEquals(onboardMgm.name, outputStub.getFirstPartyName())
+    }
+
+    private fun OutputStub.getFirstPartyName(): String? {
+        return printedOutput?.get(0)?.get("memberContext")?.get(MemberInfoExtension.PARTY_NAME)?.asText()
     }
 
-    private class OutputStub : Output {
+    class OutputStub : Output {
```
charlieR3: Could this stay as private?

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,221 @@
...
+        private lateinit var onboardMgm: OnboardMgm
+
+        private val CLI_PARAMS = arrayOf(
+            "--target=${DEFAULT_CLUSTER.rest.uri}",
+            "--user=${DEFAULT_CLUSTER.rest.user}",
+            "--password=${DEFAULT_CLUSTER.rest.password}",
+            "--insecure=true"
+        )
+
+        private val x500 = MemberX500Name.parse(
```
YashNabar:
```suggestion
        private val mgm = MemberX500Name.parse(
```

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,221 @@
...
+            "CN=Alice, OU=R3 Test, O=Mgm, L=London, ST=Tottenham, C=GB"
+        )
+
+        @BeforeAll
+        @JvmStatic
+        fun setup() {
+            onboardMgm = OnboardMgm()
```
YashNabar: Should we onboard a member as well? I don't have a strong opinion on this

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,221 @@
...
+    @Test
+    @Order(1)
+    fun `test member lookup command with status filter with ACTIVE`() {
+        CommandLine(memberLookup).execute(
+            "--status=${ACTIVE}",
+            "-h=${HoldingIdentityUtils.getHoldingIdentity(
+                null,
+                onboardMgm.name,
+                null
+            )}",
```
YashNabar: We can do this once under `setup` and reuse it

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,221 @@
...
+            *CLI_PARAMS
+        )
+
+        assertEquals(onboardMgm.name, outputStub.getFirstPartyName())
+    }
+
+    @Test
+    @Order(9)
+    fun `test member lookup command with X500 name with default groupId from file`() {
```
YashNabar: Can we add a test where group ID is passed in?

tools/plugins/network/src/test/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
r3-arish: I am not sure about this, it would be pretty easy to onboard a member. What would the outcome of this be?

williamvigorr3 (dismissed):
LGTM Once @YashNabar is also happy.

YashNabar (commented):
LGTM. Can you include a screenshot of the test passing locally on the Combined Worker? (since it's not part of the PR gates yet)

tools/plugins/network/src/smokeTest/kotlin/net/corda/cli/plugins/network/MemberLookupTest.kt
```diff
@@ -0,0 +1,233 @@
...
+import org.junit.jupiter.api.BeforeEach
+import java.io.File
+
+class MemberLookupTest {
+    companion object {
+        private lateinit var memberLookup: MemberLookup
+        private lateinit var outputStub: OutputStub
+        private lateinit var onboardMgm: OnboardMgm
+
+        private val CLI_PARAMS = arrayOf(
```
YashNabar: Passing these in as `CLI_PARAMS[0]`, `CLI_PARAMS[1]` is not very readable. You could simply do -
```suggestion
        private val targetUrl = DEFAULT_CLUSTER.rest.uri
```
and when you execute the commands -
```
        <command>.execute(
            "-h=$holdingIdentity",
            "--target=$targetUrl",
```

YashNabar (approved):
LGTM

tools/plugins/network/build.gradle
```diff
@@ -10,6 +10,38 @@ ext {
     releasable = false
 }
 
+sourceSets {
```
williamvigorr3, line 13: I don't think these tests are running in Jenkins (correct me if I am wrong). I think they should run here: https://ci02.dev.r3.com/job/Corda5/job/corda-runtime-os-combined-worker-e2e-tests/job/PR-4544/4/testReport/.
We need to work out why.

williamvigorr3 (commented):
The implementation of the tests looks good.

tools/plugins/network/build.gradle
r3-arish: Yash had mentioned that we will set up the job in Jenkins later

YashNabar: @williamvigorr3 we will need to set up a pipeline for this, it will most likely be separate from the other combined worker tests since we don't want it to be a PR gate yet, so something similar to the multi-cluster pipeline. But we can do that separately, the goal of this change is to have some initial tests which run against a combined worker locally

williamvigorr3 (approved):
LGTM

PR #4548, opened 2023-08-30
CORE-13119 Stable paging of ledger custom queries

relyafi:
### Summary

Previously, when using paging with a custom query, it was possible for custom queries to miss rows of data if any of the fields in the query where clause changed. This would most commonly occur if the query was using the `consumed` field as part of the where clause, and one or more of the states in a previously retrieved page were consumed prior to the final page of the query being returned. See the custom query stable paging [design document](https://github.com/corda/platform-eng-design/blob/master/core/corda-5/corda-5.0/ledger/core-13119-custom-query-stable-paging/index.md) for worked examples and further background.

This PR resolves this problem by introducing `StableResultSetImpl` and `StableResultSetExecutor` classes. Unlike the existing offset-based implementations, these rely on a query extending the user-supplied `WHERE` clause with additional fields, which act as a resumption point in subsequent pages. These additional fields must form a monotonically increasing, unique key on the rows. Each page returned specifies a resumption point which contains the values of these fields from the final row returned in the page, which are then used as a lower bound for the query when executing the next page.

This approach guarantees a stable ordering, because even if the fields in the user-supplied where clause change between pages, the resumption point remains fixed. It is also likely to be more performant compared to the previous offset approach when there are many pages of data to return, because the database does not need to scan through all previously seen records before starting to return the latest page of data.

Only custom queries have been adapted to use this new approach. The existing `ResultSetImpl` and `ResultSetExecutor` have been renamed to `OffsetResultSetImpl` and `OffsetResultSetExecutor` respectively, to make their behavior unambiguous. All other queries continue to use this approach for the time being. This approach continues to be safe so long as the fields specified in the query where clause are invariant for the lifetime of the query, however where possible, we should gradually transition other functionality over to use the stable implementations as well, due to the aforementioned performance benefits.

### Backwards compatibility implications

These changes modify the Avro messages sent between the flow processor which makes query requests, and ledger persistence processor, which processes and responds to the requests. See https://github.com/corda/corda-api/pull/1229 for these changes.

As Corda 5.1 does not support rolling upgrades, this is not anticipated to have an impact on backwards compatibility as an entire cluster will be brought down and restarted with the new version of the software. There will therefore be no old messages using the 5.0 Avro format in-flight on restart.

As mentioned earlier, only custom query functionality has been updated to use this approach. All other queries continue to use the old offset based query method, and are therefore unaffected.

### Testing

Extensive changes have had to be made to tests both in this PR, and in the E2E test repository (see https://github.com/corda/corda-e2e-tests/pull/220), as these changes have resulted in some changes in behavior. Following these changes, all automated test cases pass.


components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/StableResultSetImpl.kt
```diff
@@ -0,0 +1,60 @@
...
+            throw IllegalStateException("The query returned too many results")
+        }
+
+        results = serializedResults.map { serializationService.deserialize(it.array(), resultClass) }
+
+        if (nextResumePoint != null && nextResumePoint == resumePoint) {
+            throw IllegalStateException("Infinite query detected; resume point has not been updated")
+        } else {
+            resumePoint = nextResumePoint
+        }
```
lankydan: We could use `check` here and some of the if statements above where we're throwing `IlegalStateException`s, especially this one where the else block will only be executed if the exception isn't thrown anyway.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -36,6 +38,59 @@ class VaultNamedQueryExecutorImpl(
         val log = LoggerFactory.getLogger(VaultNamedQueryExecutorImpl::class.java)
     }
 
+    /*
+     * Captures data passed back and forth between this query execution and the caller in a flow
+     * processor to enable subsequent pages to know where to resume from. Data is opaque outside
+     * this class.
+     */
+    @CordaSerializable
+    data class ResumePoint(
```
lankydan, line 52: Are we just focusing on our ordering needs and excluding the custom ordering for now?

lankydan: Should we make the comment more clear that this isn't in the data module because it is not serialized anywhere else?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -36,6 +38,59 @@ class VaultNamedQueryExecutorImpl(
...
+        val created: Instant,
+        val txId: String,
+        val leafIdx: Int
+    )
+
+    /*
+     * Stores query results following processing / filtering, in a form ready to return to th
```
lankydan: `th` -> `the`

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -36,6 +38,59 @@ class VaultNamedQueryExecutorImpl(
...
+    }
+
+    /*
+     * Stores a set of raw query data returned from a single database query invocation. To support
+     * paging, this not only returns the raw query data, but also a `more` flag to indicate whether
+     * another page of data is available.
+     */
+    private data class RawQueryResults(
+        val results: List<RawQueryData>,
+        val more: Boolean
```
lankydan: `hasMore`? `hasMorePages`? Not super fussed about this one though, I'm sure you considered it.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -57,28 +112,32 @@ class VaultNamedQueryExecutorImpl(
...
         // Fetch and filter the results and try to fill up the page size then map the results
         // mapNotNull has no effect as of now, but we keep it for safety purposes
-        val (fetchedRecords, numberOfRowsFromQuery) = filterResultsAndFillPageSize(
+        val (fetchedRecords, resumePoint, numberOfRowsReturned) = filterResultsAndFillPageSize(
             request,
             vaultNamedQuery,
             deserializedParams
         )
 
+        log.trace("Fetched ${fetchedRecords.size} records in this page " +
```
lankydan: I think we should include the name of the query in the log line.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -97,86 +156,116 @@ class VaultNamedQueryExecutorImpl(
...
         deserializedParams: Map<String, Any>
-    ): FilterResult {
-        val filteredResults = mutableListOf<StateAndRef<ContractState>>()
+    ): ProcessedQueryResults {
+        val filteredRawData = mutableListOf<RawQueryData>()
 
         var currentRetry = 0
         var numberOfRowsFromQuery = 0
+        var currentResumePoint = request.resumePoint?.let {
+            serializationService.deserialize(request.resumePoint.array(), ResumePoint::class.java)
```
lankydan: should be able to do `deserialize<ResumePoint<(request.resumePoint.array()`, you'll need to find the right import though.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -97,86 +156,116 @@ class VaultNamedQueryExecutorImpl(
...
+        }
 
-        while (filteredResults.size < request.limit && currentRetry < RESULT_SET_FILL_RETRY_LIMIT) {
+        while (filteredRawData.size < request.limit && currentRetry < RESULT_SET_FILL_RETRY_LIMIT ) {
             ++currentRetry
 
-            log.trace("Executing try: $currentRetry, fetched ${filteredResults.size} number of results so far.")
+            log.trace("Executing try: $currentRetry, fetched ${filteredRawData.size} number of results so far.")
```
lankydan: Can you add the query name (I know this isn't really your change here)

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -97,86 +156,116 @@ class VaultNamedQueryExecutorImpl(
...
+        resumePoint: ResumePoint?
+    ): RawQueryResults {
 
         validateParameters(request)
 
         @Suppress("UNCHECKED_CAST")
-        return entityManagerFactory.transaction { em ->
+        val resultList = entityManagerFactory.transaction { em ->
+
+            val resumePointExpr = resumePoint?.let {
```
lankydan, line 285: As a theoretical, what would happen if the user specified the `created` time in their where clause? Would that cause a conflict with our code? Do we want to prevent that?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -97,86 +156,116 @@ class VaultNamedQueryExecutorImpl(
...
+                " AND ((tc_output.created > :created) OR " +
+                "(tc_output.created = :created AND tc_output.transaction_id > :txId) OR " +
+                "(tc_output.created = :created AND tc_output.transaction_id = :txId AND tc_output.leaf_idx > :leafIdx))"
```
lankydan, line 288: Is the leaf index needed here in case we return one leaf and not the other? Maybe a comment would be useful here, because I thought it was odd that we're doing a `>` on `leaf_idx`.

lankydan: Maybe the same for transaction_id, but when i think about it some more, it does make sense. So maybe the comment just makes things clearer.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/StableResultSetImpl.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    @Suspendable
+    override fun next(): List<R> {
+        if (!firstExecution && !hasNext()) {
+            throw NoSuchElementException("The result set has no more pages to query")
+        }
+
+        val (serializedResults, nextResumePoint) = resultSetExecutor.execute(
+            serializedParameters,
+            if (firstExecution) null else resumePoint
```
blsemo: Couldn't that just be `resumePoint`? It will always be null on first execution.

components/flow/flow-service/src/test/kotlin/net/corda/flow/application/persistence/query/ResultSetFactoryImplTest.kt
blsemo, line 1: Shouldn't that also include a test for the StableResultSet now?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
blsemo:
```suggestion
     * Stores query results following processing / filtering, in a form ready to return to the
```

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -57,28 +112,32 @@ class VaultNamedQueryExecutorImpl(
...
             deserializedParams
         )
 
+        log.trace("Fetched ${fetchedRecords.size} records in this page " +
+                "(${numberOfRowsReturned - fetchedRecords.size} records filtered)")
+
         val filteredAndTransformedResults = fetchedRecords.mapNotNull {
             vaultNamedQuery.mapper?.transform(it, deserializedParams) ?: it
         }
-
```
blsemo: Keep this blank line for readability?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -97,86 +156,116 @@ class VaultNamedQueryExecutorImpl(
...
-                    filteredResults.add(contractStateResult)
+                if (vaultNamedQuery.filter.filter(result.stateAndRef, deserializedParams)) {
+                    filteredRawData.add(result)
                 }
 
-                if (filteredResults.size >= request.limit) {
-                    return FilterResult(
-                        results = filteredResults,
-                        numberOfRowsFromQuery = numberOfRowsFromQuery
+                if (filteredRawData.size >= request.limit) {
```
blsemo, line 232: Why is this `>=`? The filtered data size should never be greater than the limit, or did I miss something here?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -97,86 +156,116 @@ class VaultNamedQueryExecutorImpl(
...
+                    // Page filled. We need to set the resume point based on the final filtered
+                    // result (as we may be throwing out additional records returned by the query).
+                    //
+                    // There are more results if either we didn't get through all the results
+                    // returned by the query invocation, or if the query itself indicated there are
+                    // more results to return.
+                    val moreResults = rawResults.results.size > filteredRawData.size || rawResults.more
```
blsemo: Is that correct? `rawResults` is the current fetch from the database, which should be of size `request.limit` unless it's smaller as there aren't any more records. `filteredRawData.size` is >= `request.limit` here, so I am not sure whether `rawresults.results.size` can ever be greater than `filteredRawData.size` at this point.
Don't we really want to check if we have looked at all the results in the current `rawResults`? That probably requires a counter how many rows have been filtered from the current result.

blsemo (commented):
I have a few questions around the resume point logic.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -187,28 +276,38 @@ class VaultNamedQueryExecutorImpl(
...
+                        $resumePointExpr
                         AND visible_states.created <= :$TIMESTAMP_LIMIT_PARAM_NAME
                         ORDER BY tc_output.created, tc_output.transaction_id, tc_output.leaf_idx
                 """,
-                Tuple::class.java
-            )
+                Tuple::class.java)
+
+            if (resumePoint != null) {
+                log.debug("Query is resuming from $resumePoint")
```
lankydan: Should be `debug { }` I also wonder if `trace {}` is better here?

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/query/StableResultSetImpl.kt
```diff
@@ -0,0 +1,60 @@
...
+    private var results: List<R> = emptyList()
+    private var resumePoint: ByteBuffer? = null
+    private var firstExecution = true
+
+    override fun getResults(): List<R> {
+        return results
+    }
+
+    override fun hasNext(): Boolean {
+        return resumePoint != null
```
lankydan, line 30: Should we add a comment that `resumePoint == null` means that the query has reached the final page? I think we either do that, or we make a `ResumePoint` avro object that specifies its state more explicitly, but that is more work.

lankydan (commented):
Left some comments, but I think logically it makes sense. If all of the tests that @nkovacsx wrote then we're obviously in a good place with working code.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
relyafi: I believe this is existing code (from @nkovacsx recent changes), I agree I think just an equals check would be sufficient, but left it as I felt it wouldn't be harmful and is a bit more defensive if for whatever reason the code breaks in future (unlikely though that seems).

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
relyafi: > Are we just focusing on our ordering needs and excluding the custom ordering for now?

Yes. I imagine we will need to revisit this in future and change it so that the resume point and order by expressions share field / value pairs that can be specified by custom queries.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
relyafi: I think it's just stylistic, happy to change it to `hasMore`

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
relyafi: Good question, I think our solution should not preclude user specified created times. I don't see any conflicts with this - even if the user gave a created time in conflict with the resume point time, it should just return 0 results and there be no more data.

I do think though it's worth writing an additional test case as part of the E2E PR, just to confirm this works correctly.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
relyafi: Yes exactly - we could be mid-way through states for a transaction where multiple states in the transaction match the query when we end a page.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
nkovacsx: Yeah I agree with @lankydan, `more` is more like a "non-boolean" naming for me.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -97,86 +156,116 @@ class VaultNamedQueryExecutorImpl(
...
+                    )
+                }
             }
 
-            contractStateResults.forEach { contractStateResult ->
+            rawResults.results.forEach { result ->
                 ++numberOfRowsFromQuery
-                if (vaultNamedQuery.filter.filter(contractStateResult, deserializedParams)) {
-                    filteredResults.add(contractStateResult)
+                if (vaultNamedQuery.filter.filter(result.stateAndRef, deserializedParams)) {
```
nkovacsx, line 227: This is not your change but I was wondering if we should rename `filter` to something else. This chain looks a bit hideous here: `filter.filter()`.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
nkovacsx: Yeah I just added `>` for extra safety.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
relyafi: @lankydan is that necessary? The code is single threaded with no suspension, and we print the query name at a higher log level on line 101. So if trace logging is enabled, we should see these lines together in the logs (if you filter by class name - admittedly a worker with multiple active processors may muddy the waters).

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
relyafi: Same as earlier comment - we could but think it's redundant?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
relyafi: I think the difficulty is that it's probably the function call, not property that could be named better - I'm not sure what else we would call an instance of `VaultNamedQueryFilter` - `vaultNamedQuery.queryFilter.filter` for example isn't really any better.

Unfortunately `VaultNamedQuery` is public API, so we can't do anything there - ideally we would do something like rename `filter(...)` to `execute(...)` or something.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
relyafi: Yeah I think you're right. I've updated to check if the current result (in the context of the for each loop) is the final raw result.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
blsemo: You'd hope that even if you have multiple processors running, something like thread or flow id in the MDC would allow you to easily tie together relevant loglines.

relyafi: Yes, there will be that context so you could piece things together. Closing as in Dan's absence, @nkovacsx is happy with this too.

josephzunigadaly: Trace ID inside MDC is always enabled in 5.1 onwards. So there will be a unique ID tying all related log messages together, even across workers.

blsemo (dismissed):
LGTM

blsemo (approved):
Re-approved

LWogan (approved):
lgtm

simon-johnson-r3 (approved):

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
relyafi: Addressed by https://github.com/corda/corda-e2e-tests/pull/236

PR #4550, opened 2023-08-30
CORE-16803: Refactor TokenCacheEventProcessorFactory as an OSGi component.

chrisr3:
Refactor complex wiring code for `TokenCacheEventProcessor` into `TokenCacheEventProcessorFactory` and then declare this factory to be an OSGi component, so that other components (e.g. testing frameworks) can create `TokenCacheEventProcessor` instances without breaking encapsulation.


owenstanford (approved):
looks ok to me

PR #4551, opened 2023-08-30
CORE-15808: Remove 'releasable=true' from tracing/build.gradle

ZawarR3:
This change simply removes the 'releasable=true' extension from the Gradle build file for the tracing module


josephzunigadaly (approved):

ronanbrowne (approved):
LGTM

PR #4552, opened 2023-08-30
CORE-16560 Harden interop identity registry to avoid desynchronization with kafka topic

ac101m:
# Changes
- Harden `InteropIdentityRegistryView` against the possibility of desynchronization with topics.
- Refactor registry methods for ease of use.
- Add exception class for registry state errors.
- Remove `equals()` and `hashCode()` overrides for `InteropIdentity` class, as they are no longer required.
- Remove `InteropIdentityTests`, no longer required because of above change.


jmacmahonr3 (approved):
LGTM

components/interop/interop-identity-registry-service-impl/src/main/kotlin/net/corda/interop/identity/registry/impl/InteropIdentityRegistrySet.kt
```diff
@@ -0,0 +1,24 @@
+package net.corda.interop.identity.registry.impl
+
+import net.corda.crypto.core.ShortHash
+import net.corda.interop.core.InteropIdentity
+
+/**
+ * Set-like container for InteropIdentities which keys off of short hash and ignores other properties.
```
szymonsztuka, line 7: The data structure is "domain" agnostic, but in fact is used in a specific context, I would like to explicitly say at least in `get` method description, short hash of what is used?

components/interop/interop-identity-registry-service-impl/src/main/kotlin/net/corda/interop/identity/registry/impl/InteropIdentityRegistryViewImpl.kt
```diff
@@ -5,118 +5,134 @@ import net.corda.interop.core.InteropIdentity
...
-    private val byGroupId = HashMap<String, HashSet<InteropIdentity>>()
-    private val byVirtualNodeShortHash = HashMap<ShortHash, HashSet<InteropIdentity>>()
-    private val byShortHash = HashMap<ShortHash, InteropIdentity>()
-    private val myIdentities = HashMap<String, InteropIdentity>()
-    private val byApplicationName = HashMap<String, InteropIdentity>()
-    private val byFacadeId = HashMap<String, HashSet<InteropIdentity>>()
+    private val ownedIdentities = HashMap<String, InteropIdentityRegistrySet>()
+    private val byGroupId = HashMap<String, InteropIdentityRegistrySet>()
+    private val byVirtualNodeShortHash = HashMap<ShortHash, InteropIdentityRegistrySet>()
+    private val byApplicationName = HashMap<String, InteropIdentityRegistrySet>()
```
szymonsztuka, line 24: In most cases when using auxiliary maps like `byApplicationName` a calling code needs just a set and is not interested of a mapping from  hashes, hence you have a lot of `.toSet()` which flattens data structure. What about in auxiliary data structures we don't use `InteropIdentityRegistry` and just `Set<InteropIdentity>` , maybe even `TreeSet` is fine ? And the usage is simpler then, from:
``
Collections.unmodifiableSet(byApplicationName[applicationName]?.toSet() ?: emptySet())
``
to:
``Collections.unmodifiableSet(byApplicationName[applicationName] ?: emptySet())``

components/interop/interop-identity-registry-service-impl/src/main/kotlin/net/corda/interop/identity/registry/impl/InteropIdentityRegistryViewImpl.kt
```diff
@@ -5,118 +5,134 @@ import net.corda.interop.core.InteropIdentity
...
+    override fun getIdentities(): Set<InteropIdentity> =
+        Collections.unmodifiableSet(interopIdentities.toSet())
+
+    override fun getIdentitiesByGroupId(groupId: String): Set<InteropIdentity> =
+        Collections.unmodifiableSet(byGroupId[groupId]?.toSet() ?: emptySet())
 
-    override fun getIdentitiesByGroupId(): Map<String, Set<InteropIdentity>> =
-        Collections.unmodifiableMap(byGroupId)
+    override fun getIdentityWithShortHash(shortHash: ShortHash): InteropIdentity? =
+        interopIdentities.get(shortHash)
```
szymonsztuka, line 106: As spoke offline, this method is used by REST service only, REST doesn't need to be extremely performant (unlike usage in FlowProcessor / InteropProcessor), what about just using a set and filtering to get the right entry, we could get ring on one data structure then `interopIdentities`.

szymonsztuka (commented):
General comment, all those auxiliary HashMaps  in`InteropIdentityRegistryViewImpl` will be stored also in a fiber, in general maps used exclusively by REST or serives for  adding /removing entries can be just replaced by single set and filer/find, they can be slow, but agree anything used by interopProcessor/FlowWorker (aka runtime flow processing) should use hash maps, be fast. This is maybe longer term goal to have specialized `RegistryView`. for REST, for InteropProcessor and for FlowWorker (they can reuse code):
My understanding who uses each method (and it's backing  HashMap):
// REST  + InteropIdentityProcessor 
getIdentities()

//Interop Processor
getIdentitiesByGroupId

//REST
getIdentityWithShortHash

// REST + Flow Worker
getIdentitiesByApplicationName

// flow worker only?
getIdentityWithApplicationName

//interop processor - in the future own worker
getOwnedIdentity

components/interop/interop-identity-registry-service-impl/src/main/kotlin/net/corda/interop/identity/registry/impl/InteropIdentityRegistryViewImpl.kt
ac101m: `java.util.Set` relies upon the `equals()` and `hashCode()` functions, which we'd have to override. Overriding `equals()` means we can't use == to compare interop identity objects. This was the rationale for the `InteropIdentityRegistrySet` class.

ronanbrowne (approved):
LGTM - blt files

PR #4557, opened 2023-08-31
CORE-16623 : Sync syntax with release 5.1

SaurabhR3:


jmacmahonr3 (approved):
LGTM

gradle.properties
```diff
@@ -46,7 +46,7 @@ bouncycastleVersion=1.73
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.0.0.xx-SNAPSHOT to pick up maven local published copy
 #cordaApiVersion=5.1.0.xxx-SNAPSHOT
-cordaApiVersion=5.1.0-INTEROP.15-beta+
+cordaApiVersion=5.1.0-INTEROP.15-alpha-1693465318193
```
ac101m: Don't forget to update once API pr is merged!

ac101m (changes requested):

ac101m (approved):
LGTM

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
```diff
@@ -135,23 +137,18 @@ internal class OutboundMessageProcessor(
...
-        if (!linkManagerHostingMap.isHostedLocally(cordaSource)) {
-            return "source ID is not locally hosted"
-        }
-        // Perform a stricter locally-hosted check on the destination, which includes session key matching, to guard
-        // against scenarios such as when a message is being sent to a duplicate holding identity on another cluster,
-        // but is instead routed to a matching holding identity which is hosted locally.
-        membershipGroupReaderProvider.lookup(cordaSource, cordaDestination, filter)?.let {
-            if (linkManagerHostingMap.isHostedLocallyAndSessionKeyMatch(it)) {
-                return validateCanMessage(cordaSource, cordaDestination, outbound = true, inbound = true)
-            }
```
ac101m: I think it would be wise to leave this code in, commented out with a comment, rather than to delete it entirely. Will be helpful to whoever picks up the related ticket to fix this later, and will also provide context to people merging in the future.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
```diff
@@ -196,48 +193,38 @@ internal class OutboundMessageProcessor(
             return emptyList()
         }
 
-        val destinationMemberInfo = membershipGroupReaderProvider.lookup(
+        val destMemberInfo = membershipGroupReaderProvider.lookup(
```
ac101m: Undo rename, we shouldn't be refactoring stuff in link manager!

ac101m (changes requested):
Tested locally, looks to be working!
But I think the diff is unnecessarily large, with some re-ordered added/deleted things that didn't need changing.
I think it's probably best to reverse these changes and keep the diff as small as possible. And also to not to delete any code outright here, but rather comment it out and add an explanation, so that context is maintained and we can resolve future merges correctly.

szymonsztuka (changes requested):
The last change in e82eeec12db2c892bac1b812b24da271cbdd6c4a to find a temp workaround is wrong as it alters the platform behaviour and is too invasive. I think the only change should be to modify behavior of `isHostedLocallyAndSessionKeyMatch` method for the case it's processes interop based Memebr info, we know  when it does then it throws an exception, so we should catch this exception, the existing method:
```
  override fun isHostedLocallyAndSessionKeyMatch(member: MemberInfo) =
        locallyHostedIdentityToIdentityInfo[member.holdingIdentity]?.let { identityInfo ->
            member.sessionInitiationKeys.any { memberSessionKey ->
                identityInfo.allSessionKeysAndCertificates.any {
                    it.sessionPublicKey.encoded.contentEquals(memberSessionKey.encoded)
                }
            }
        } ?: false
```
Should be wrapped with try catch:
```
  override fun isHostedLocallyAndSessionKeyMatch(member: MemberInfo) =
        try {
        locallyHostedIdentityToIdentityInfo[member.holdingIdentity]?.let { identityInfo ->
            member.sessionInitiationKeys.any { memberSessionKey ->
                identityInfo.allSessionKeysAndCertificates.any {
                    it.sessionPublicKey.encoded.contentEquals(memberSessionKey.encoded)
                }
            }
        } ?: false
        } catch (exp: Exception) { false }
```
Not sure if we shout return true or false when exception is caught, but this should be the only change I think.

szymonsztuka:
Regarding my last comment (https://github.com/corda/corda-runtime-os/pull/4557#pullrequestreview-1609963860) - the workaround proposal I think it should return true in case of an exception is cought.

PR #4559, opened 2023-08-31
CORE-16585: Fix re-registration on vNode upgrade

BenYip123:
This PR fixes re-registration by getting the serial from the correct field.

- Query for only `APPROVED` registration requests
- Get previous serial from `serial` field
- Attach serial to all re-registration requests

### DIFF
--- a/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
+++ b/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
@@ -238,14 +238,15 @@ internal class VirtualNodeUpgradeOperationHandler(
                     } catch (e: ContextDeserializationException) {
                         logger.warn(
                             "Could not deserialize previous registration context for ${holdingIdentity.shortHash}. " +
-                                    "Re-registration won't be attempted."
+                                    "Re-registration will not be attempted."
                         )
                     }
 
                 is MembershipQueryResult.Failure ->
                     logger.warn(
-                        "Failed to query for an APPROVED previous request: ${registrationRequest.errorMsg}. " +
-                                "Re-registration won't be attempted."
+                        "Failed to query for an APPROVED previous registration request for ${holdingIdentity.shortHash}: " +
+                                "${registrationRequest.errorMsg}. " +
+                                "Re-registration will not be attempted."
                     )
             }
         }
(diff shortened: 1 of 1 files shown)


nikinagy (dismissed):
LGTM

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -213,22 +214,20 @@ internal class VirtualNodeUpgradeOperationHandler(
...
         ) {
-            val membershipGroupReader = membershipGroupReaderProvider.getGroupReader(holdingIdentity)
-            val x500Name = membershipGroupReader.owningMember
+            val x500Name = membershipGroupReaderProvider.getGroupReader(holdingIdentity).owningMember
             val registrationRequest = membershipQueryClient
-                .queryRegistrationRequests(holdingIdentity, x500Name, limit = 1)
+                .queryRegistrationRequests(holdingIdentity, x500Name, listOf(APPROVED), 1)
             try {
-                val registrationContext = (registrationRequest as MembershipQueryResult.Success)
+                val registrationRequestDetails = (registrationRequest as MembershipQueryResult.Success)
```
josephzunigadaly: We should avoid the use of `as` here. It will throw an exception when the type is not `Success`. It's better to add an if statement that checks the type and skip this block of code when it is not `Success`.

josephzunigadaly (dismissed):

nikinagy (commented):
Shouldn't we handle somehow if the querying is not successful?

BenYip123:
> Shouldn't we handle somehow if the querying is not successful?

Would it be better to keep using `as` in this case and then just catch the exception if this is the behaviour we want?

nikinagy:
> > Shouldn't we handle somehow if the querying is not successful?
> 
> Would it be better to keep using `as` in this case and then just catch the exception if this is the behaviour we want?

I think you can just add an else branch here and log a warning message in case of failure. Eg similar to what have in DistributeMemberInfoActionHandler.kt line 84-85

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -213,30 +214,39 @@ internal class VirtualNodeUpgradeOperationHandler(
...
+                        logger.warn(
+                            "Could not deserialize previous registration context for ${holdingIdentity.shortHash}. " +
+                                    "Re-registration won't be attempted."
+                        )
+                    }
+
+                is MembershipQueryResult.Failure ->
+                    logger.warn(
+                        "Failed to query for an APPROVED previous request: ${registrationRequest.errorMsg}. " +
+                                "Re-registration won't be attempted."
```
nikinagy: One last comment: I think we should put here whose re-registration won't be attempted. It would help a lot to include the holding identity's short hash in the warning message. So that ideally the users/clients can see that the re-registration attempt failed and can try to submit it manually for that particular member.

nikinagy (commented):
LGTM

nikinagy (approved):

PR #4560, opened 2023-08-31
CORE-16808: Add a Cloudwatch metrics registry

JamesHR3:
**Problem statement**

Corda can provide a large volume of metrics data when run at scale. This can lead to challenges with running monitoring infrastructure. Addressing this requires both a review of the monitoring stack and of the data that Corda is exporting to ensure this is sensible. In the short term, Corda needs to integrate with a monitoring system that copes with the volume of data being exported, in order for us to do some initial testing.

**Solution**

Introduce an additional Cloudwatch metrics registry alongside the Prometheus one Corda already uses. This allows Corda to push metrics data to a Cloudwatch instance provided one is set up correctly in the environment Corda is currently running in.

This is a very rough solution to this problem. In particular, the mechanism for providing the AWS SDK in the OSGi framework is very blunt and does not follow the pattern we've established for working with other non-OSGi dependencies. In the short term this is probably ok. Longer term a more fully fledged metrics infrastructure is likely needed, which will remove the need for this change.

Ideally it should be possible to disable this behaviour, or change the frequency with which metrics are pushed. This will be addressed in the future.

**Testing**

Deployed into an environment where Cloudwatch is enabled and proved that metrics are visible in Cloudwatch.

**Other changes**

None.


applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/internal/WorkerMonitorImpl.kt
```diff
@@ -39,17 +44,35 @@ internal class WorkerMonitorImpl @Activate constructor(
...
     private val prometheusRegistry: PrometheusMeterRegistry = PrometheusMeterRegistry(PrometheusConfig.DEFAULT)
+    private val cloudwatchConfig = object : CloudWatchConfig {
+        override fun get(key: String): String? {
+            return null
+        }
+
+        override fun namespace(): String {
+            return CORDA_NAMESPACE
+        }
+    }
```
jujoramos: Out of curiosity: how does this fixed namespace part work?, wouldn't having this hard-coded cause issues when having multiple Corda deployments in the same Kubernetes cluster?, shouldn't this be obtained as an environmnet variable or something similar (using the Kubernetes Downward API probably?)?.

jujoramos (dismissed):

applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/internal/WorkerMonitorImpl.kt
JamesHR3: It's not the kubernetes namespace, it's the namespace to write Corda's metrics under in Cloudwatch. At the moment this means we'll write all metrics under the same namespace.

That is quite ugly though and could lead to data getting mixed up, so actually that's not a great idea. I think you're right and we should use the kubernetes namespace to separate out this data. Let me look into that.

jujoramos (dismissed):

davidcurrie:
What happens if the customer doesn't have AWS/CloudWatch? Are they going to see a stack of errors? Is there a performance overhead in having registries that are not being used?

JamesHR3:
> What happens if the customer doesn't have AWS/CloudWatch? Are they going to see a stack of errors? Is there a performance overhead in having registries that are not being used?

As it stands yes, but this is more to unblock an immediate need and some refinement will come later. Hard to answer the second question right now because of said immediate need. Happy to catch up with you separately if you have some concerns about that.

mcgovc (dismissed):
LGTM - build changes

thiagoviana (dismissed):
LGTM

charts/corda-lib/templates/_worker.tpl
```diff
@@ -159,6 +159,13 @@ spec:
               fieldRef:
                 apiVersion: v1
                 fieldPath: metadata.namespace
+          - name: ENABLE_CLOUDWATCH
+            value:
+              {{- if eq $.Values.serviceAccount.name "cloudwatch-writer" }}
```
davidcurrie, line 164: Eventually, I'd prefer to see this exposed explicitly under the `metrics` stanza in the overrides but I'll take this as a step in the right direction!

davidcurrie (approved):

jujoramos (approved):

mag1c-48 (approved):
LGTM - BLT files only

PR #4563, opened 2023-08-31
CORE-15304 Categorise persistence exceptions

lankydan:
In the flow and ledger persistence code, categorise the persistence
exceptions that occur so that we can handle them appropriately.

`PersistenceException` and `SQLException` will now default to be
platform errors (meaning they return an error to user code) rather
than transient errors which led to unnecessary retries. For a subset
of exceptions, transient or fatal errors will be returned depending on
the exceptions.


blsemo (dismissed):
LGTM

vlajos (approved):

simon-johnson-r3 (approved):

PR #4567, opened 2023-09-01
CORE-16178 update to latest corda-api version and test against old initiate flow api with context builder

LWogan:
Address merge conflicts due to flow-mapper moving a new worker while flow-mapper lib moved to component.

Update a flow test to use the old API ` FlowSession initiateFlow(@NotNull MemberX500Name x500Name, @NotNull FlowContextPropertiesBuilder flowContextPropertiesBuilder);`  to ensure backwards compatibility


simulator/runtime/src/main/kotlin/net/corda/simulator/runtime/messaging/ConcurrentFlowMessaging.kt
```diff
@@ -76,6 +76,13 @@ class ConcurrentFlowMessaging(
         return doInitiate(x500Name, flowContextPropertiesBuilder)
     }
 
+    override fun initiateFlow(
+        x500Name: MemberX500Name,
+        flowContextPropertiesBuilder: FlowContextPropertiesBuilder
+    ): FlowSession {
+        return initiateFlow(x500Name, true, flowContextPropertiesBuilder)
```
JamesHR3, line 83: Just to check - when we have the `requireClose` functionality implemented, are we expecting the API without the `requireClose` parameter to default it to `true` or `false`?

JamesHR3 (approved):

JamesHR3 (approved):

simulator/runtime/src/main/kotlin/net/corda/simulator/runtime/messaging/ConcurrentFlowMessaging.kt
JamesHR3: Verified we're starting with this `true`.

LWogan: true was agreed in design meeting as the starting point.

JamesHR3 (approved):

PR #4570, opened 2023-09-01
CORE-14685: Separate notary service X500 name from its member X500 name.

chrisr3:
An actual notary service should be provided by one or more "worker" virtual nodes, none of whose X500 names matches the notary service's X500 name.

Adjust the driver's mock network to provide a separate X500 name for each notary service's "worker" virtual node, to make it more realistic.

### DIFF
--- a/testing/driver/src/main/java/net/corda/testing/driver/node/EmbeddedNodeService.java
+++ b/testing/driver/src/main/java/net/corda/testing/driver/node/EmbeddedNodeService.java
@@ -13,8 +13,35 @@ import java.time.Duration;
 import java.util.Map;
 import java.util.Set;
 
+import static java.util.Objects.requireNonNull;
+
 @DoNotImplement
 public interface EmbeddedNodeService {
+    /**
+     * This tag is appended to the {@code commonName} field
+     * of each Notary service's {@link MemberX500Name} to
+     * generate the {@link MemberX500Name} for its vNode.
+     */
+    String NOTARY_WORKER_TAG = "(Worker)";
+
+    /**
+     * @param x500Name The {@link MemberX500Name} for a notary service.
+     * @return The {@link MemberX500Name} for a notary service's virtual node.
+     */
+    @NotNull
+    static MemberX500Name toNotaryWorkerName(@NotNull MemberX500Name x500Name) {
+        requireNonNull(x500Name, "x500Name cannot be null");
+        final String commonName = x500Name.getCommonName();
+        return new MemberX500Name(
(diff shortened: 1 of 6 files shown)


testing/driver/src/main/java/net/corda/testing/driver/node/EmbeddedNodeService.java
```diff
@@ -13,8 +13,35 @@
 import java.util.Map;
 import java.util.Set;
 
+import static java.util.Objects.requireNonNull;
+
 @DoNotImplement
 public interface EmbeddedNodeService {
```
khutsijabari, line 19: Why create this as an interface when it contains an implementation already?

testing/driver/src/main/java/net/corda/testing/driver/node/EmbeddedNodeService.java
```diff
@@ -13,8 +13,35 @@
...
+    /**
+     * This tag is appended to the {@code commonName} field
+     * of each Notary service's {@link MemberX500Name} to
+     * generate the {@link MemberX500Name} for its vNode.
+     */
+    String NOTARY_WORKER_TAG = "(Worker)";
```
tlawson3, line 25: Why is this required, is there a change to the Corda platform which does this?
Can't say I'm a fan: I give my notary one x500Name and it modifies it to be slightly different, no doubt I'll struggle to look it up later.

testing/driver/src/main/java/net/corda/testing/driver/node/EmbeddedNodeService.java
chrisr3: > Why create this as an interface when it contains an implementation already?

I've created a `static` method on this interface so that it can be invoked both _inside_ and _outside_ the driver's OSGi framework. Otherwise I'd have needed to duplicate it.

testing/driver/src/main/java/net/corda/testing/driver/node/EmbeddedNodeService.java
chrisr3: > Why is this required, is there a change to the Corda platform which does this? Can't say I'm a fan: I give my notary one x500Name and it modifies it to be slightly different, no doubt I'll struggle to look it up later.

The Corda platform actually _does_ work this way already, and I just hadn't realised. So I've fixed my network to be more realistic.

I don't think CorDapp developers are really supposed to interact directly with the notary "worker" nodes in practice. They're just supposed to use the notary's service X500 name.

mcgovc (approved):
LGTM

PR #4573, opened 2023-09-02
CORE-16858 session initiation and context propagation issues

LWogan:
Fix issue where session context properties were not being exchanged properly.

Also fix an issue where calling receive as the first operation did not send a SessionInit message

This PR does not address the issue where it is not possible to exchange session properties after the first send when no data messages are received. i.e call send and then call getCounterPartyInfo https://r3-cev.atlassian.net/browse/CORE-16860

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/sessions/impl/FlowSessionManagerImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/sessions/impl/FlowSessionManagerImpl.kt
@@ -117,7 +117,8 @@ class FlowSessionManagerImpl @Activate constructor(
             checkpoint,
             sessionId,
             payload = SessionConfirm(),
-            instant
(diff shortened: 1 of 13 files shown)


components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/sessions/impl/FlowSessionManagerImpl.kt
```diff
@@ -117,7 +117,8 @@ class FlowSessionManagerImpl @Activate constructor(
             checkpoint,
             sessionId,
             payload = SessionConfirm(),
-            instant
+            instant,
+            contextSessionProperties = contextSessionProperties
```
chrisr3: This looks odd, and a bit unnecessary seeing as the parameter name matches the parameter value.

ben-millar (dismissed):

ben-millar (approved):

PR #4574, opened 2023-09-03
CORE-16808: Create AWS CloudWatch SDK bundle for temporary use.

chrisr3:
Wrap the AWS CloudWatch APIs as one big OSGi bundle so that we can remove them from the system bundle.

~I haven't tested that it can still upload metrics to AWS, and we almost certainly _don't_ want to distribute this artifact even if it _does_ work.~
@JamesHR3 reports that this PR works with CloudWatch correctly, but this AWS SDK bundle is _still_ just a temporary artifact.


JamesHR3 (approved):

mcgovc (approved):
LGTM - build file changes

PR #4575, opened 2023-09-04
CORE-15562 NPE fixes

vlajos:
A set of fixes from 5.1
CORE-15562 Add QuasarIgnoreAllPackages to net.corda.internal.serialization.amqp
CORE-16798: Quasar to ignore ledger-utxo-data

### DIFF
--- a/libs/ledger/ledger-utxo-data/build.gradle
+++ b/libs/ledger/ledger-utxo-data/build.gradle
@@ -8,6 +8,7 @@ description 'Corda ledger UTXO data'
 dependencies {
     compileOnly 'org.osgi:osgi.annotation'
     compileOnly 'org.osgi:org.osgi.service.component.annotations'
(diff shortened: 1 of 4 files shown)


libs/metrics/src/main/java/net/corda/metrics/package-info.java
```diff
@@ -1,6 +1,6 @@
 @Export
-@QuasarIgnorePackage
+@QuasarIgnoreAllPackages
```
chrisr3: You don't need this. The original annotation is correct because this bundle only _has_ one package to ignore.

chrisr3 (changes requested):
The metrics change is unnecessary.

libs/metrics/src/main/java/net/corda/metrics/package-info.java
vlajos: 👍  I have removed that commit.

chrisr3 (approved):
LGTM

PR #4578, opened 2023-09-04
CORE-15939: NOT FOR REVIEW - Add support for Rekey and Rewrap bus processors, add REST APIs

anien:
This change adds two building blocks for key rotation

- It adds first draft for REST APIs that support initiating the key rotation of the root key and to check the status of the current key rotation in progress.
- It also adds the rekey and rewrap bus processors to handle new kafka messages.
- `CryptoRekeyBusProcessor` handles messages coming from the REST Worker about the key rotation
- `CryptoRewrapBusProcessor` handles individuals key rotations (rewraps) of the key.

The idea is that user post a request to Corda saying it wants to rotate a key. Rest Worker then publishes a message to Kafka (first new message) that key rotation is initiated. Crypto Worker picks up this message and finds out how many keys are required to rotate (rewrap with the new wrapping key) - this is done by `CryptoRekeyBusProcessor`. For each of these keys it posts a message to Kafka with individual key request. This is a second new message. Messages are picked up by `CryptoRewrapBusProcessor` which then do the individual rewrap of the key.

Design doc: [Key rotation](https://github.com/corda/platform-eng-design/blob/7abc01be6c2fb1ece3bb239f1864ea806fe6326e/core/corda-5/corda-5.0/crypto-worker/crypto-wrapping-keys-autorotation.md#startkeyrotationcommand-kafka-message)

API PR:[1232](https://github.com/corda/corda-api/pull/1232)


dickon:
I've been through this to some extent and I think it's ready to go to the feature branch. Unfortunately to get it working I had to add a few commits of my own, which we'll need to get reviewed at some point. Since it is DRAFT and green at the time of writing I can't merge the PR so I'll simply force-push the head  to the feature branch.

PR #4583, opened 2023-09-05
CORE-16901: Do not require a wakeup event when platform errors are encountered

JamesHR3:
**Problem statement**

Platform errors are thrown in the flow pipeline if an error is encountered when calling an API, and are used as a signal that information about the failure should be returned to the user code. Previously when this happened, the exception triggered the flow pipeline to abort processing of the inbound event and record the error on the checkpoint. The flow was then woken up by the scheduler at some later point, which would trigger the platform error to be passed back to the user code.

With the removal of the wakeup scheduler, this mechanism no longer works. However, as part of the same change, the flow pipeline is also capable of running the user code through multiple fiber suspension points in one run. This means that for errors with requests coming out the user code, the fiber can immediately be re-run with an error returned.

**Solution**

Catch platform exceptions in the `FlowExecutionPipelineStage`, convert them to error flow continuations, and return the error to the user directly. This deals with the majority of platform exceptions currently in the system. There are two exceptions:

- The `CounterpartyFlowInfoWaitingForHandler` can throw a platform exception. As a waiting for handler, the job of this class is to evaluate the current state of the checkpoint alongside the input event, and determine which flow continuation to return. In this case, the same condition can be handled by immediately returning a flow continuation instead, rather than throwing the excpetion.
- The `FlowGlobalPostProcessor` may throw a platform exception if the counterparty to a session does not exist. This is inherently dangerous, as the flow could wake up from calling any API (or even exiting a subflow) with an error for something entirely unrelated to what the flow was doing. However, the check must be done here as it is possible for a counterparty to cease to exist after the flow has requested a message be sent (for example if the member is suspended). Consequently, changing this to a fatal exception is the most consistent approach.

Other structure for handling platform errors, such as the checks in the `WaitingForWakeupHandler`, and the checkpoint pending platform error, can be removed.

**Testing**

Executed this branch against the `FlowTests` e2e suite and achieved a passing run.


jujoramos (approved):
LGTM

jujoramos (dismissed):

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowGlobalPostProcessorImpl.kt
```diff
@@ -109,7 +109,7 @@ class FlowGlobalPostProcessorImpl @Activate constructor(
             if (doesCheckpointExist) {
                 log.debug { "$msg. Throwing FlowPlatformException" }
```
emilybowe: should this be updated to Throwing FlowFatalException?

LWogan: debug line here needs updating

LWogan: although maybe it should be deleted? not sure it adds value

emilybowe (dismissed):
LGTM

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowGlobalPostProcessorImpl.kt
```diff
@@ -109,7 +109,7 @@ class FlowGlobalPostProcessorImpl @Activate constructor(
...
                 checkpoint.putSessionState(sessionState)
-                throw FlowPlatformException(msg)
+                throw FlowFatalException(msg)
```
LWogan, line 112: The user code is no longer given the chance to catch this exception. Are we sure we're ok with that?

JamesHR3: The issue here is that catching this from a user perspective is very difficult, as this could be thrown from any input event and therefore after any suspension. That means we might suddenly throw an exception back to the user from a surprising place, so unless the entire flow is wrapped in a `try` block I think catching this is very hard.

LWogan: If the Platform exception was thrown somewhere like FlowSessionManagerImpl.sendSessionMessageToExistingSession it could be picked up in the post processing step.

I'm not fully against a fatal exception though.

JamesHR3: It would probably be better to put this check in the send/send and receive handlers. However, that still doesn't help with the code as written today. The output messages get generated in the global post processor, so they could be produced by any event input.

Perhaps we could treat that as an enhancement?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowGlobalPostProcessorImpl.kt
JamesHR3: Fair question. I've just updated it for now as it's debug.

LWogan (approved):
LGTM

PR #4584, opened 2023-09-05
CORE-13225 - Make publisher reference atomic

bpaunescu:
Some tests (currently observed e2e tests only) seem to fail (rarely) due to what looks like a Kafka producer fencing issue. The initial attempt to resolve this issue, which impacts test stability, made the publisher reference atomic (Corda publishers are wrappers over standard Kafka producers) to ensure that resetting the publisher would ensure the new instance was visible in every thread using it, avoiding having two producer references with the same transactional ID. This fix proved insufficient over time as more failures were reported. This is because other publishers would fail 

Recent analysis reveals that there is a quite specific order of publishers which cause the issue. Meaning, if the first publisher is fixed, the next one would eventually cause the same issue. For example, logs from failures from the last 5 months show that the issue is always caused first by **VirtualNodeInfoWriterComponent**. A fix there would eventually show that the **CpiInfoWriterComponent** will fail. Fixing that as well simply makes the issue appear in **ChunkDbWriter**.

**Proposed solution** - ensure **all references** are atomic by moving initial fix to the Corda publisher code;

Log statements from when the issue started in VirtualNodeInfoWriterComponent

`[pod/corda-bob-db-worker-7cf4cd766f-6k69g/corda-bob-db-worker] 14:45:06.712 [lifecycle-coordinator-41] INFO  net.corda.messaging.publisher.CordaPublisherImpl {} - Producer clientId VIRTUAL_NODE_INFO_WRITER, transactional true, failed to send, resetting producer
[pod/corda-bob-db-worker-7cf4cd766f-6k69g/corda-bob-db-worker] net.corda.messaging.api.exception.CordaMessageAPIProducerRequiresReset: Error occurred beginning transaction for CordaKafkaProducer with clientId VIRTUAL_NODE_INFO_WRITER`

Log statements from when the issue started in CpiInfoWriterComponent

`[pod/corda-bob-db-worker-5bfc868499-6m7d2/corda-bob-db-worker] 08:34:47.945 [lifecycle-coordinator-57] INFO  net.corda.messaging.publisher.CordaPublisherImpl {} - Producer clientId CPI_INFO_WRITER, transactional true, failed to send, resetting producer
[pod/corda-bob-db-worker-5bfc868499-6m7d2/corda-bob-db-worker] net.corda.messaging.api.exception.CordaMessageAPIProducerRequiresReset: Error occurred beginning transaction for CordaKafkaProducer with clientId CPI_INFO_WRITER`

Log statements from when the issue started in ChunkDbWriter

`[pod/corda-bob-db-worker-744b8fd4d9-j65pn/corda-bob-db-worker] 13:27:07.877 [durable processing thread cpi.chunk.writer-cpi.upload] INFO  net.corda.messaging.publisher.CordaPublisherImpl {} - Producer clientId chunk-writer, transactional true, failed to send, resetting producer
[pod/corda-bob-db-worker-744b8fd4d9-j65pn/corda-bob-db-worker] net.corda.messaging.api.exception.CordaMessageAPIProducerRequiresReset: Error occurred beginning transaction for CordaKafkaProducer with clientId chunk-writer`

**Observations**: 

- issues seems to exclusively happen only during reconciliation process
- root cause is still unknown (logs don't show any other errors before); services going up and down a lot during reconciliation might have something to do with it
- something in Kafka throws an `IllegalStateException` when a producer tries to **begin** a transaction; our code handles this by resetting the published (close existing producer -> blocks until all messages are sent; create new instance with the same transactional ID)

### DIFF
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaPublisherImpl.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaPublisherImpl.kt
@@ -42,7 +42,8 @@ internal class CordaPublisherImpl(
         private const val QUEUE_SIZE = 200
     }
 
-    private val cordaProducer: AtomicReference<CordaProducer?> = AtomicReference(cordaProducerBuilder.createProducer(producerConfig, config.messageBusConfig))
+    private val cordaProducer: AtomicReference<CordaProducer?> = AtomicReference(cordaProducerBuilder
+        .createProducer(producerConfig, config.messageBusConfig))
 
     private data class Batch(val records: List<Record<*, *>>, val future: CompletableFuture<Unit>)
     private val queue = ArrayBlockingQueue<Batch>(QUEUE_SIZE)
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaPublisherImpl.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaPublisherImpl.kt
@@ -18,6 +18,7 @@ import org.slf4j.LoggerFactory
 import java.nio.ByteBuffer
 import java.util.concurrent.ArrayBlockingQueue
 import java.util.concurrent.CompletableFuture
+import java.util.concurrent.atomic.AtomicReference
 import java.util.concurrent.locks.ReentrantLock
 import kotlin.concurrent.withLock
 
@@ -41,7 +42,7 @@ internal class CordaPublisherImpl(
         private const val QUEUE_SIZE = 200
     }
 
-    private var cordaProducer = cordaProducerBuilder.createProducer(producerConfig, config.messageBusConfig)
+    private val cordaProducer: AtomicReference<CordaProducer?> = AtomicReference(cordaProducerBuilder.createProducer(producerConfig, config.messageBusConfig))
 
     private data class Batch(val records: List<Record<*, *>>, val future: CompletableFuture<Unit>)
     private val queue = ArrayBlockingQueue<Batch>(QUEUE_SIZE)
@@ -118,7 +119,7 @@ internal class CordaPublisherImpl(
         records.toCordaProducerRecords().forEach {
             val fut = CompletableFuture<Unit>()
             futures.add(fut)
-            cordaProducer.send(it) { ex ->
+            cordaProducer.get()?.send(it) { ex ->
                 setFutureFromResponse(ex, fut, it.topic)
             }
         }
@@ -134,7 +135,7 @@ internal class CordaPublisherImpl(
         recordsWithPartitions.forEach { (partition, record) ->
             val fut = CompletableFuture<Unit>()
             futures.add(fut)
-            cordaProducer.send(record, partition) { ex ->
+            cordaProducer.get()?.send(record, partition) { ex ->
                 setFutureFromResponse(ex, fut, record.topic)
             }
         }
@@ -187,9 +188,9 @@ internal class CordaPublisherImpl(
         val future = CompletableFuture<Unit>()
         try {
             tryWithSingleRecoveryAttempt {
-                cordaProducer.beginTransaction()
-                block(cordaProducer)
-                cordaProducer.commitTransaction()
+                cordaProducer.get()?.beginTransaction()
+                block(cordaProducer.get()!!)
+                cordaProducer.get()?.commitTransaction()
                 future.complete(Unit)
             }
         } catch (ex: Exception) {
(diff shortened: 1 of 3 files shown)


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaPublisherImpl.kt
```diff
@@ -187,9 +189,9 @@ internal class CordaPublisherImpl(
         val future = CompletableFuture<Unit>()
         try {
             tryWithSingleRecoveryAttempt {
-                cordaProducer.beginTransaction()
-                block(cordaProducer)
-                cordaProducer.commitTransaction()
```
simon-johnson-r3, line 192: Is this just about ensuring the value is propagated over multiple threads' views of the variable?

An issue here is that nothings appears to be done atomically so to speak. `cordaProducer` at line 193 might have been set to null since the `beginTransaction` call in line 192 because the gets are done as separate non atomic calls.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaPublisherImpl.kt
```diff
@@ -282,15 +284,16 @@ internal class CordaPublisherImpl(
...
+        cordaProducer.set(cordaProducerBuilder.createProducer(producerConfig, config.messageBusConfig))
     }
 
     /**
      * If an exception is thrown whilst closing, swallow the error to avoid double exceptions.
      */
     private fun closeProducerAndSuppressExceptions() {
         try {
-            cordaProducer.close()
+            cordaProducer.get()?.close()
```
simon-johnson-r3, line 295: Same here, another thread scheduled in at this point (between lines 295 and 296) would be operating on a closed `cordaProducer` before this thread set is to null, so it appears to be half a solution. Perhaps there's some piece of information I'm missing?

fowlerrr: I think these two lines could be resolved by
```suggestion
            cordaProducer.getAndSet(null)?.close()
```

fowlerrr: Though, in hindsight, is there any use to making it nullable?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaPublisherImpl.kt
fowlerrr: Maybe just add 

```
synchronized(cordaProducer) {
   ...
}
```

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaPublisherImpl.kt
simon-johnson-r3: Well it does offer limited protection against operating on a closed producer, except at line 193.

I think if we believe this class should be thread safe, we should resolve it everywhere. But I'm not sure why making this thread safe fixes the issue at the moment because I don't know enough about it.

fowlerrr: The problem, IIRC, is that we are closing and recreating a publisher, but some thread will, inevitably, access the publisher in the meantime.  And so Kafka ring-fences the underlying producer.  Then, when we get that error, it stops the re-creation too.  So I _think_ we end up stuck from that point.  Syncronising all access to the producer (especially around re-creation) should fix this, I think.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaPublisherImpl.kt
bpaunescu: @simon-johnson-r3 yes, the fix was first done months ago in a specific publisher/writer. Now I've just moved it in the generic publisher. It shouldn't be a problem if the publisher is null at that point, the exception is handled. It's more important to ensure that a change in is propagated and we avoid having two instances with the same ID resulting in a producer fencing issue.

@fowlerrr The entire method is already synchronized.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/CordaPublisherImpl.kt
bpaunescu: I agree with @fowlerrr that having a lock on producer should fix the issue, but I'm worried about performance decrease. Currently we don't have an environment to test performance before and after changes.

PR #4587, opened 2023-09-05
CORE-16879 - use max retry window duration 

conal-smith:
**Problem statement**

The flow engine will raise `FlowTransientException` if something that the flow requires to run (e.g. the CPKs, the VNode info) is not currently available. The assumption is that this data will become available by waiting some amount of time, and the original event will succeed processing if replayed some time later.

Previously this was implemented using the scheduler, which was removed. Instead, the replayed event is now published back to the input topic. This means that in low load scenarios (such as some test cases), the event is replayed much more rapidly than before. This can cause failures as the retry count limit is hit much faster, often before the required data has arrived at the flow engine.

**Solution**

Terminate the retry loop based on elapsed time rather than retry count.

This should allow enough time to pass in all circumstances for the required data to become available.

**Testing**

Passing runs of unit and integration tests. E2E tests run and some failing tests are shown to improve.

API change - https://github.com/corda/corda-api/pull/1236/files


components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
```diff
@@ -77,13 +78,15 @@ class FlowEventExceptionProcessorImpl @Activate constructor(
...
 
-            /** If we have reached the maximum number of retries then we escalate this to a fatal
-             * exception and DLQ the flow
-             */
-            if (flowCheckpoint.currentRetryCount >= maxRetryAttempts) {
+            fun isRetryWindowExpired(firstFailureTimestamp: Instant?) =
+                Duration.between(firstFailureTimestamp, Instant.now()) >= maxRetryWindowDuration
+
+            /** If the retry window has expired then we escalate this to a fatal exception and DLQ the flow */
+            if (flowCheckpoint.firstFailureTimestamp != null && isRetryWindowExpired(flowCheckpoint.firstFailureTimestamp)) {
```
jujoramos: Nitpick: checking for `flowCheckpoint.firstFailureTimestamp != null` directly within `isRetryWindowExpired()` method would make the code easier to read.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
```diff
@@ -77,13 +78,12 @@ class FlowEventExceptionProcessorImpl @Activate constructor(
...
-             * exception and DLQ the flow
-             */
-            if (flowCheckpoint.currentRetryCount >= maxRetryAttempts) {
+            /** If the retry window has expired then we escalate this to a fatal exception and DLQ the flow */
+            if (retryWindowExpired(flowCheckpoint.firstFailureTimestamp)) {
                 return@withEscalation process(
                     FlowFatalException(
-                        "Execution failed with \"${exception.message}\" after $maxRetryAttempts retry attempts.",
+                        "Execution failed with \"${exception.message}\" after " +
+                                "${flowCheckpoint.currentRetryCount} retry attempts in a retry window of $maxRetryWindowDuration.",
```
emilybowe, line 90: I'd put a unit in the log message as well since I presume maxRetryWindowDuration will just print a number.

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/context/FlowServiceTestContext.kt
```diff
@@ -103,10 +103,10 @@ class FlowServiceTestContext @Activate constructor(
         FlowConfig.EXTERNAL_EVENT_MESSAGE_RESEND_WINDOW to 500000L,
         FlowConfig.SESSION_TIMEOUT_WINDOW to 500000L,
         FlowConfig.SESSION_FLOW_CLEANUP_TIME to 30000,
-        FlowConfig.PROCESSING_MAX_RETRY_ATTEMPTS to 5,
         FlowConfig.PROCESSING_MAX_FLOW_SLEEP_DURATION to 60000,
         FlowConfig.PROCESSING_MAX_RETRY_DELAY to 16000,
         FlowConfig.PROCESSING_FLOW_CLEANUP_TIME to 30000,
+        FlowConfig.PROCESSING_MAX_RETRY_WINDOW_DURATION to 300000,
```
emilybowe, line 109: nit - some of these are long? should they all be? or none of them..? 🤔

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/context/FlowServiceTestContext.kt
```diff
@@ -71,7 +71,7 @@ import org.osgi.service.component.annotations.Reference
 import org.slf4j.LoggerFactory
 import java.nio.ByteBuffer
 import java.time.Instant
-import java.util.UUID
+import java.util.*
```
emilybowe: wildcard import here in case this was added by accident

sprintengine: Good spot, thanks!

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
sprintengine: I think a duration prints out like "PT30S" for 30 seconds, "PT1H" for 1 hour, "PT0.2S" for 200ms. It might be a bit ugly. I was considering putting `Duration.between(firstFailureTimestamp, Instant.now())` in there instead of the max, but it makes it harder to make an assertion on this error message in the test.

JamesHR3: I think this gives enough context to leave in for now. I think alternatives are probably not worth the effort.

jujoramos (dismissed):
LGTM

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
```diff
@@ -52,13 +54,16 @@ class FlowEventExceptionProcessorImpl @Activate constructor(
 ) : FlowEventExceptionProcessor {
 
     private companion object {
-        val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
+        private val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
+        private const val DEFAULT_MAX_RETRY_WINDOW_DURATION_MS = 30000L
```
emilybowe: nit - could make a val that's 5 minutes in MS and use this here rather than 30000 - might be easier for someone reading

LWogan (dismissed):
LGTM

emilybowe (dismissed):
LGTM

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
JamesHR3: We had a bit of a race condition with this one :) good catch!

LWogan (approved):
lgtm

thiagoviana (approved):

PR #4589, opened 2023-09-05
CORE-15958 - Registration should fail when using EDDSA for scheme ED25519 for session keys

nikinagy:
The p2p e2e session protocol does not support EDDSA at the moment, it only supports RSA and ECDSA. Every request that tries to use any other scheme apart from RSA and ECDSA for session keys should be marked as invalid.

**Testing**
1) Generated session key using CORDA.EDDSA.ED25519 scheme and tried to register
Screenshot 2023-09-06 at 09 49 41
Screenshot 2023-09-06 at 15 20 15

2) Generated session key using CORDA.ECDSA.SECP256R1 scheme, but defined EdDSA as signature spec in the registration context
Screenshot 2023-09-06 at 09 53 21

### DIFF
--- a/components/membership/membership-p2p/src/main/kotlin/net/corda/membership/p2p/helpers/KeySpecExtractor.kt
+++ b/components/membership/membership-p2p/src/main/kotlin/net/corda/membership/p2p/helpers/KeySpecExtractor.kt
@@ -33,7 +33,7 @@ class KeySpecExtractor(
             SM2_CODE_NAME to SignatureSpecs.SM2_SM3,
             SPHINCS256_CODE_NAME to SignatureSpecs.SPHINCS256_SHA512,
         )
-        private val validSpecsNames = mapOf(
+        private val validSpecsNamesForSessionKeys = mapOf(
             ECDSA_SECP256K1_CODE_NAME to listOf(
                 SignatureSpecs.ECDSA_SHA256,
                 SignatureSpecs.ECDSA_SHA384,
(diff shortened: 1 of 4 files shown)


components/membership/membership-p2p/src/main/kotlin/net/corda/membership/p2p/helpers/KeySpecExtractor.kt
```diff
@@ -46,18 +46,29 @@ class KeySpecExtractor(
...
-        fun CryptoSigningKey.validateSpecName(specName: String) {
-            val validSpecs = validSpecsNames[this.schemeCodeName]
-                ?: throw IllegalArgumentException("Could not identify spec for key scheme ${this.schemeCodeName}.")
-            if (!validSpecs.contains(specName)) {
+        fun CryptoSigningKey.validateSpecName(specName: String, type: KeySpecType = KeySpecType.OTHER) {
+            val validSpecs = if (type == KeySpecType.SESSION) {
+                requireNotNull(validSpecsNamesForSessionKeys[this.schemeCodeName]) {
+                    "Invalid key scheme is used for session key or " +
+                            "could not identify spec for key scheme ${this.schemeCodeName}. The following " +
+                            "schemes should be used when generating session keys: ${validSpecsNamesForSessionKeys.keys}"
```
YashNabar: Could this be shorter? It doesn't fit in the `reason` field of the registration request at the moment

components/membership/membership-p2p/src/main/kotlin/net/corda/membership/p2p/helpers/KeySpecExtractor.kt
```diff
@@ -46,18 +46,29 @@ class KeySpecExtractor(
...
+                }
+            } else {
+                requireNotNull(validSpecsNames[this.schemeCodeName]) {
+                    "Could not identify spec for key scheme ${this.schemeCodeName}."
+                }
+            }
+            require(validSpecs.contains(specName)) {
```
williamvigorr3, line 70: I think if you use require, you don't need the illegal argument exception, inside the lambda.

williamvigorr3 (commented):
I left a comment. Otherwise LGTM.

components/membership/membership-p2p/src/main/kotlin/net/corda/membership/p2p/helpers/KeySpecExtractor.kt
nikinagy: ah yeah, wrong copy paste, good catch!

YashNabar (approved):
LGTM

williamvigorr3 (approved):
LGTM

PR #4590, opened 2023-09-05
CORE-16919: Move tracing implementation into :libs:tracing-impl module.

chrisr3:
Migrate Brave tracing implementation into `:libs:tracing-impl` module. Allow `:libs:tracing` to load Brave implementation using `java.util.ServiceLoader`.

Modules that need to use tracing only need to include `:libs:tracing`, which defaults to the `NoopTracingService` implementation for the sake of unit and integration testing. The `:libs:tracing-impl` module has been added to all workers.

Note that we must include the SPI-Fly framework extension for `java.util.ServiceLoader` to operate correctly inside the OSGi framework.

---
Also note that Liquibase uses `java.util.ServiceLoader`, which means that _every_ DB-related worker needs to include the SPI-Fly framework extension regardless.


libs/tracing-impl/src/main/kotlin/net/corda/tracing/brave/BraceTracingServiceFactory.kt
```diff
@@ -0,0 +1,27 @@
+package net.corda.tracing.brave
+
+import aQute.bnd.annotation.spi.ServiceProvider
+import net.corda.tracing.TracingService
+import net.corda.tracing.TracingServiceFactory
+import net.corda.v5.base.exceptions.CordaRuntimeException
+
+@Suppress("unused")
+@ServiceProvider(TracingServiceFactory::class)
+class BraceTracingServiceFactory : TracingServiceFactory {
```
mbrkic-r3: Typo, it should be `BraveTracingServiceFactory` (and file renamed)

owenstanford (commented):
What is the issue with the current implementation? The Noop classes are not required anymore and scheduled for deletion so we don't really want to depend on them if we don't have to. All the current unit tests etc. work with the current implementation.

libs/tracing-impl/src/main/kotlin/net/corda/tracing/brave/BraceTracingServiceFactory.kt
```diff
@@ -0,0 +1,27 @@
...
+    override fun create(serviceName: String, zipkinHost: String?, samplesPerSecond: String?): TracingService {
```
mbrkic-r3: It would be better that `samplesPerSecond` is of type `SampleRate` and parsing below (with reference to "`--trace-samples-per-second`" which doesn't belong here) is done when processing worker's arguments (using `picocli` custom type converter).

chrisr3:
> What is the issue with the current implementation? The Noop classes are not required anymore and scheduled for deletion so we don't really want to depend on them if we don't have to. All the current unit tests etc. work with the current implementation.

Surely the current unit tests etc _require_ the Noop classes? I'm struggling to understand how you can remove them, unless you're saying that we don't have any test coverage for code paths which include tracing?

My purpose here is to remove Jetty and Javalin as dependencies from _everything_ which includes `:lib:tracing`.

libs/tracing-impl/src/main/kotlin/net/corda/tracing/brave/BraceTracingServiceFactory.kt
chrisr3: I believe `SampleRate` is a `brave` type, and so adding `brave` as a dependency to `:libs:tracing` would defeat the entire point of this PR, which is to _remove_ `brave` from `:libs:tracing`.

mbrkic-r3: It's our class: `net.corda.tracing.impl.SampleRate`

chrisr3: So they are.

However, I see that they've also been explicitly marked as `internal` and `sealed`, which makes me think that they aren't intended to be exposed as public API.

I don't need to make this change in order to remove Javalin and Jetty as dependencies.

mbrkic-r3: Yes, this might be considered as technical debt and done in a separate PR, tagging @josephzunigadaly to decide/create a Jira ticket for this if needed.

mbrkic-r3 (approved):
LGTM

mcgovc (approved):
LGTM - build file changes

owenstanford:
> > What is the issue with the current implementation? The Noop classes are not required anymore and scheduled for deletion so we don't really want to depend on them if we don't have to. All the current unit tests etc. work with the current implementation.
> 
> Surely the current unit tests etc _require_ the Noop classes? I'm struggling to understand how you can remove them, unless you're saying that we don't have any test coverage for code paths which include tracing?
> 
> My purpose here is to remove Jetty and Javalin as dependencies from _everything_ which includes `:lib:tracing`.

Yes, you're correct we can't remove them now we have split the module into API + Impl. 
We can look again at this as part of other tech dept and future changes around tracing.

owenstanford (approved):

williamvigorr3 (approved):
LGTM (network team files).

PR #4592, opened 2023-09-06
CORE-14025 Check if transaction present in the database before requesting in backchain resolution

nkovacsx:
### Overview

Adding a new request to the UTXO ledger layer to fetch transaction that are not invalid. This means they have verified or unverified status.

This is needed when doing a back-chain resolution. We don't want to fetch transactions that are already present in our database. We just want to fetch their IDs though, no need to deserialise the whole object.

The back-chain receiver flow will now check if any of the transactions we want to fetch are already in the database. If yes, then we'll just not request those transactions.

```
existingTransactionsInDb.addAll(utxoLedgerPersistenceService.findExistingNotInvalidTransactionIds(
    // Make sure we don't fetch the same transaction multiple times
    (transactionsToRetrieve - existingTransactionsInDb)
))

// Remove the transaction from the "to retrieve" set if they are in our DB
transactionsToRetrieve.removeIf { existingTransactionsInDb.contains(it) }
```

### Test Plan

Added new unit test scenarios to V1/V2 flow tests and making sure the current e2e tests are passing.


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
```diff
@@ -69,6 +71,24 @@ class UtxoRepositoryImpl @Activate constructor(
...
+    override fun findExistingNotInvalidTransactionIds(
+        entityManager: EntityManager,
+        transactionIds: List<String>
+    ): List<SecureHash> {
+        return entityManager.createNativeQuery(
+            """
+                SELECT id,
+                FROM {h-schema}utxo_transaction
+                WHERE id IN (:transactionIds) 
+                AND status <> :invalid""",
```
lankydan: Is `<>` mean not equal? I thought it was `!=` in sql

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1Test.kt
```diff
@@ -70,6 +70,39 @@ class TransactionBackchainReceiverFlowV1Test {
...
+     * This test is simulating a scenario where we want to fetch 2 transactions but we already have one of them
+     * in our database. In that case it will be removed from the `transactionsToRetrieve` list and will not be
+     * requested.
+     */
+    @Test
+    fun `transaction will not be requested if it is present in the database`() {
+        whenever(utxoLedgerPersistenceService.findExistingNotInvalidTransactionIds(any()))
+            .thenReturn(listOf(TX_ID_1))
+
+        whenever(session.sendAndReceive(eq(List::class.java), any())).thenReturn(
```
lankydan: Is this mock needed?

lankydan (commented):
Generally looks fine to me, left a few comments.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
vlajos: Did the status already get moved into utxo_transaction?

nkovacsx: Good point, I meant to use `utxo_transaction_status` here.

nkovacsx: @lankydan AFAIK both `<>` and `!=` are the same basically, I just prefer `<>` for some reason.

relyafi: Technically, it looks like using `<>` is actually more correct, in so far as that is how not equals is defined in the ANSI standard (http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt):

   `<not equals operator> ::= <>`

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1Test.kt
nkovacsx: nope, good point.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/request/handlers/UtxoFindNotInvalidTransactionIdsRequestHandler.kt
blsemo, line 1: The filename does not quite match the implementation - this looks for ids in a given set of statuses - not invalid is just the use case we use it for in this PR, but nothing in this class mandates this.

nkovacsx: thanks, fixed.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -72,7 +73,17 @@ class TransactionBackchainReceiverFlowV1(
...
         while (transactionsToRetrieve.isNotEmpty()) {
+            existingTransactionsInDb.addAll(utxoLedgerPersistenceService.findTransactionIdsWithStatuses(
+                // Make sure we don't fetch the same transaction multiple times
+                (transactionsToRetrieve - existingTransactionsInDb),
+                listOf(UNVERIFIED, VERIFIED)
+            ))
+
+            // Remove the transaction from the "to retrieve" set if they are in our DB
+            transactionsToRetrieve.removeIf { existingTransactionsInDb.contains(it) }
             val batch = transactionsToRetrieve.take(batchSize)
```
vlajos, line 84: I think `batch` can be empty now.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -72,7 +73,17 @@ class TransactionBackchainReceiverFlowV1(
...
+
         while (transactionsToRetrieve.isNotEmpty()) {
+            existingTransactionsInDb.addAll(utxoLedgerPersistenceService.findTransactionIdsWithStatuses(
+                // Make sure we don't fetch the same transaction multiple times
+                (transactionsToRetrieve - existingTransactionsInDb),
+                listOf(UNVERIFIED, VERIFIED)
+            ))
+
+            // Remove the transaction from the "to retrieve" set if they are in our DB
+            transactionsToRetrieve.removeIf { existingTransactionsInDb.contains(it) }
```
vlajos: Could this work?
```suggestion
            transactionsToRetrieve.removeAll ( existingTransactionsInDb )
```

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -72,7 +73,17 @@ class TransactionBackchainReceiverFlowV1(
 
         val batchSize = ledgerConfig.getInt(BACKCHAIN_BATCH_CONFIG_PATH)
 
+        val existingTransactionsInDb = mutableSetOf<SecureHash>()
```
lankydan: I don't think we need `existingTransactionsInDb` since the this code should handle that scenario:

```kotlin
private fun addUnseenDependenciesToRetrieve(
        retrievedTransaction: UtxoSignedTransaction,
        sortedTransactionIds: TopologicalSort,
        transactionsToRetrieve: LinkedHashSet<SecureHash>
    ) {
        if (retrievedTransaction.id !in sortedTransactionIds.transactionIds) {
            retrievedTransaction.dependencies.let { dependencies ->
                val unseenDependencies = dependencies - sortedTransactionIds.transactionIds
                log.trace {
                    val ignoredDependencies = dependencies - unseenDependencies
                    "Backchain resolution of $initialTransactionIds - Adding dependencies for transaction ${retrievedTransaction.id} " +
                            "dependencies: $unseenDependencies to transactions to retrieve. Ignoring dependencies: $ignoredDependencies " +
                            "as they have already been seen."
                }
                sortedTransactionIds.add(retrievedTransaction.id, dependencies)
                transactionsToRetrieve.addAll(unseenDependencies)
            }
        }
    }
```

Double check that to be sure.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -72,7 +73,17 @@ class TransactionBackchainReceiverFlowV1(
...
+
         while (transactionsToRetrieve.isNotEmpty()) {
+            existingTransactionsInDb.addAll(utxoLedgerPersistenceService.findTransactionIdsWithStatuses(
```
lankydan: I was just thinking if there is an advantage to returning the statuses of each transaction, rather than filtering by status. This would be similar to the `persistIfDoesNotExist` method used a bit further down in this class.

My thinking is that if a transaction is invalid, then we might as well just stop backchain resolution there.

Also makes me realise that `persistIfDoesNotExist` doesn't handle the scenario of invalid transactions.

In fact we have this code:

```kotlin
val (status, summaries) = serializationService.deserialize<Pair<String?, List<CordaPackageSummary>>>(it.array())
                when (status) {
                    null -> TransactionExistenceStatus.DOES_NOT_EXIST
                    "U" -> TransactionExistenceStatus.UNVERIFIED
                    "V" -> TransactionExistenceStatus.VERIFIED
                    else -> throw IllegalStateException("Invalid status $status")
                } to summaries
```

Which is going to fail if a transaction already exists in the database as invalid when resolving a transaction. Luckily the invalid code path only really happens with malicious actors, but it should still work.

I think we should raise a separate ticket being its a bigger change than what @nkovacsx has done here. @blsemo @relyafi @vlajos what do you think?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
lankydan: I agree, I guess we can stop receiving transactions at that point. Or do a `continue` to go round the loop?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoPersistenceService.kt
```diff
@@ -25,6 +26,17 @@ interface UtxoPersistenceService {
...
+    /**
+     * Find transactions with the given [transactionIds] that are present in the persistence context
+     * and has a status that is in the [statuses] list.
+     *
+     * @param transactionIds IDs of transactions to find.
+     * @param statuses Statuses the transactions should have in order to be fetched.
+     *
+     * @return A list of the transaction IDs found.
+     */
+    fun findTransactionIdsWithStatus(transactionIds: List<String>, statuses: List<String>): List<SecureHash>
```
relyafi: Should `statuses` be a list of `String`, or `TransactionStatus`? The latter approach prevents illegal values being supplied, and `String` already has an extension function to convert to this.

(Looks like this comment may be relevant to a few of our other functions too)

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
nkovacsx: I suppose if the batch is empty there's no point looping anymore? We can just jump straight to sending `TransactionBackchainRequestV1.Stop`. What do you think @lankydan ?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
blsemo: I am in favour of having a `getTransactionStatus( transactionIds: Set<SecureHash>) : Map<SecureHash, TransactionStatus>`, and then use that to do the logic on the flow side.
That might also be useful beyond this specific use case.

relyafi: > Which is going to fail if a transaction already exists in the database as invalid when resolving a transaction. Luckily the invalid code path only really happens with malicious actors, but it should still work.

This sounds like a bug, so yes agree that sounds like a separate ticket.

> I was just thinking if there is an advantage to returning the statuses of each transaction, rather than filtering by status. This would be similar to the persistIfDoesNotExist method used a bit further down in this class.

If our use case simply wants to get everything, that makes sense - we could drop the `statuses` argument and return status in the return value. That would simplify (and probably marginally speed up) the DB query. This would come at the cost of hurting performance if another API user genuinely did just want a subset based on state, as we're back to in-memory filtering territory. We could always add another API for that though.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
lankydan: Discussed with @nkovacsx and we think the behaviour he added is right. We'll add an extra test to confirm that the transactions in the db don't go into the `TopologicalSort` output.

lankydan: Nvm he already added a test, so all good.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
lankydan: I'll raise a bug ticket to address this.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
lankydan: i think so.

lankydan:
Annoyingly, I thought of another issue with our current solution.

Scenario:
- Two flows are resolving the same backchain.
- Flow A is a bit faster in getting through the backchain and is doing the most of the persisting of unverified transactions.
- Flow B is a bit slower and is hitting our new code which sees that it already has the unverified transactions, so it doesn't request them from the peer.
- Flow B skips requesting the transactions from peers, as it already has them as unverified, and finishes its receiving step of backchain resolution and moves onto verification. It is important to point out here that skipping the receiving step means (based on our current implementation) that we also don't request the dependencies of the transactions we already see as unverified. But, even though we see them as unverified, we might not have yet requested them in Flow A and we cannot guarantee that Flow A has or will finish successfully.
- As Flow B has moved onto verification it will verify whatever transactions it did need to request and then verify the new transaction.

This means we can end up verifying a new transaction without verifying its whole backchain. 

In fact, if we hit a malicious transaction and this race-condition simultaneously, we could successfully verify an invalid transaction. Even if Flow A did find an invalid transaction in the backchain while it was resolving, if Flow B has moved to the verifying step before Flow A, then it is too late.

@blsemo @relyafi @vlajos @nkovacsx can you all go over what I've written and check that it's plausible?

vlajos:
> Annoyingly, I thought of another issue with our current solution.
> 
> Scenario:
> 
>     * Two flows are resolving the same backchain.
> 
>     * Flow A is a bit faster in getting through the backchain and is doing the most of the persisting of unverified transactions.
> 
>     * Flow B is a bit slower and is hitting our new code which sees that it already has the unverified transactions, so it doesn't request them from the peer.
> 
>     * Flow B skips requesting the transactions from peers, as it already has them as unverified, and finishes its receiving step of backchain resolution and moves onto verification. It is important to point out here that skipping the receiving step means (based on our current implementation) that we also don't request the dependencies of the transactions we already see as unverified. But, even though we see them as unverified, we might not have yet requested them in Flow A and we cannot guarantee that Flow A has or will finish successfully.
> 
>     * As Flow B has moved onto verification it will verify whatever transactions it did need to request and then verify the new transaction.
> 
> 
> This means we can end up verifying a new transaction without verifying its whole backchain.
> 
> In fact, if we hit a malicious transaction and this race-condition simultaneously, we could successfully verify an invalid transaction. Even if Flow A did find an invalid transaction in the backchain while it was resolving, if Flow B has moved to the verifying step before Flow A, then it is too late.
> 
> @blsemo @relyafi @vlajos @nkovacsx can you all go over what I've written and check that it's plausible?

Yes. It sounds plausible to me.
Also, dependencies may include the related signed group parameters, not just the input/reference state's transactions.
If the signed group parameters stay missing, then B may move the transaction to Invalid, if A already downloaded the Tx, but the related group parameters have not arrived yet.

Maybe we could modify the code to bypass requesting Txs only if they are already fully Verified in the database since the unverified ones may not be fully downloaded with all of their dependencies?

blsemo:
Good catch, Dan - it seems we can't stop resolving at an unverified transaction. We don't need to request that transaction, but we need to keep walking the chain of dependencies (potentially fetching none of them), just in case the resolution that run before failed.

nkovacsx:
`We don't need to request that transaction, but we need to keep walking the chain of dependencies` --  I don't think this is possible. We only fetch the ID from the database so there's no way to get its dependencies without retrieving the transaction itself. @blsemo @vlajos or did I miss something?

vlajos:
> `We don't need to request that transaction, but we need to keep walking the chain of dependencies` -- I don't think this is possible. We only fetch the ID from the database so there's no way to get its dependencies without retrieving the transaction itself. @blsemo @vlajos or did I miss something?

I think @blsemo means "request the transaction **from the initiator participant**"

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -212,6 +202,51 @@ class TransactionBackchainReceiverFlowV1(
...
+        existingTransactionIdsInDb: MutableMap<SecureHash, TransactionStatus>,
+        transactionsToRetrieve: LinkedHashSet<SecureHash>
+    ) {
+        // Fetch the existing transaction with Verified/Unverified status from the DB and store them
+        existingTransactionIdsInDb.putAll(
+            utxoLedgerPersistenceService.findTransactionIdsAndStatuses(
+                // Make sure we don't fetch the same transaction multiple times
+                (transactionsToRetrieve - existingTransactionIdsInDb.keys)
+            ).filterValues {
+                // We throw away transactions that are "INVALID"
```
vlajos: There was an idea to abort back-chain resolution if any dependencies were invalid.
Or did this land in a ticket?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -191,6 +202,51 @@ class TransactionBackchainReceiverFlowV1(
...
+            } else {
+                log.trace {
+                    "Transaction with ID $transactionId is present in the database with unverified status " +
+                            "but could not fetch the signed transaction object and its dependencies."
+                }
+            }
+        }
+        // Once we added the unverified transactions' dependencies into the "to retrieve" set
+        // we can remove those from the set as we don't need to get those from the initiator
+        transactionsToRetrieve.removeIf { existingTransactionIdsInDb[it] == UNVERIFIED }
```
blsemo: You have to loop around at this point - even if you can also remove the dependencies, the dependencies of dependencies need to be looked at and so on. If you come out of this function with an empty batch, the loop in the calling function will be broken, and we might not resolve/verify some transactions.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -212,6 +202,51 @@ class TransactionBackchainReceiverFlowV1(
...
+            if (transactionFromDb != null) {
+                // Add the dependencies to the "to retrieve" set
+                transactionsToRetrieve.addAll(transactionFromDb.dependencies)
+            } else {
+                log.trace {
+                    "Transaction with ID $transactionId is present in the database with unverified status " +
+                            "but could not fetch the signed transaction object and its dependencies."
+                }
+            }
+        }
```
vlajos, line 283: The signed group parameters referenced in the tx's metadata may need to be requested if it is unknown to the current participant.
Probably the easiest is to keep the txs with unknown, signed group parameters in the transationsToRerieve var.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -191,6 +202,51 @@ class TransactionBackchainReceiverFlowV1(
...
+        existingTransactionIdsInDb: MutableMap<SecureHash, TransactionStatus>,
+        transactionsToRetrieve: LinkedHashSet<SecureHash>
+    ) {
+        // Fetch the existing transaction with Verified/Unverified status from the DB and store them
+        existingTransactionIdsInDb.putAll(
+            utxoLedgerPersistenceService.findTransactionIdsAndStatuses(
+                // Make sure we don't fetch the same transaction multiple times
+                (transactionsToRetrieve - existingTransactionIdsInDb.keys)
+            ).filterValues {
+                // We throw away transactions that are "INVALID"
```
blsemo: Is that the correct behaviour? Sholdn't we mark a transaction as invalid if any dependency is invalid?

blsemo (commented):
Correct me if I am wrong, but I think this would still fall short if the backchain is several layers deep, and the faster flow has fetched but not verified more than one, but not all layers?
Also, do we need to handle invalid transactions?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -139,19 +154,19 @@ class TransactionBackchainReceiverFlowV1(
...
+
+        val (groupParametersHash, groupParameters) = fetchGroupParametersAndHashForTransaction(retrievedTransaction)
+
+        if (groupParameters != null) {
+            return
+        }
 
         if (utxoLedgerGroupParametersPersistenceService.find(groupParametersHash) != null) {
             return
         }
```
vlajos: I think this is not needed anymore

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -191,6 +206,76 @@ class TransactionBackchainReceiverFlowV1(
...
+    ) {
+
+        val transactionsFromDb = utxoLedgerPersistenceService.findTransactionIdsAndStatuses(
+            // Make sure we don't fetch the same transaction multiple times
+            (transactionsToRetrieve - existingTransactionIdsInDb.keys)
+        )
+
+        // Check if we have any invalid transactions. If yes, we can't continue the back-chain resolution.
+        val invalidTransactions = transactionsFromDb.filterValues { it == INVALID }.keys
+        if (invalidTransactions.isNotEmpty()) {
```
vlajos:
```suggestion
        if (transactionsFromDb.any { it.value == INVALID }) {
```

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -191,6 +206,76 @@ class TransactionBackchainReceiverFlowV1(
...
+            throw InvalidBackchainException("Found the following invalid transaction(s) during back-chain resolution: " +
+                    "$invalidTransactions. Back-chain resolution cannot be continued.")
+        }
+
+        // Fetch the existing transaction with Verified/Unverified status from the DB and store them
+        existingTransactionIdsInDb.putAll(
+            transactionsFromDb.filterValues {
+                it == VERIFIED || it == UNVERIFIED
```
vlajos: Can there be anything else in that variable here?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -191,6 +206,76 @@ class TransactionBackchainReceiverFlowV1(
...
+                // but only if we "know" its group parameters, otherwise we need to keep it there
+                if (fetchGroupParametersAndHashForTransaction(transactionFromDb).second != null) {
+                    transactionsToRetrieve.remove(transactionId)
+                }
+            } else {
+                log.trace {
+                    "Transaction with ID $transactionId is present in the database with unverified status " +
+                            "but could not fetch the signed transaction object and its dependencies."
+                }
+            }
```
blsemo: This loop will go through the tx in the list to fetch and check if those are in the DB, but it uncoditionally adds their depencies without checking if they are in the DB, in turn, so they would end up in the batch and get fetched whether they're required or not.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
nkovacsx: I wanted to do this first but this way we lose the transaction ids for the log line. Not sure if that's a problem though.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
nkovacsx: As I see `TransactionStatus` is an enum:
```
enum class TransactionStatus(val value: String) {
    INVALID("I"),
    UNVERIFIED("U"),
    VERIFIED("V");
...
}
```

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
nkovacsx: Yeah but the main loop iterates through `transactionsToRetrieve`:
`while (transactionsToRetrieve.isNotEmpty()) {` (line 82) and it will check the same conditions for each transactions. Including the dependencies.

nkovacsx: `transactionsToRetrieve` keeps growing as we go through dependencies and the same checks etc. will run for each transaction that we put in there.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
vlajos: Good point.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
vlajos: This `filterValues` does not do anything here since if there were any invalids, we do not reach this point.

blsemo (dismissed):
LGTM

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
nkovacsx: good point

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -189,6 +201,85 @@ class TransactionBackchainReceiverFlowV1(
...
+                transactionsFromDb.filterValues {
+                    it == VERIFIED || it == UNVERIFIED
+                }
+            )
+
+            // Remove the transaction from the "to retrieve" and "to check" collections if they are in our DB,
+            // and they are verified
+            listOf(transactionsToCheck, transactionsToRetrieve).forEach { collection ->
+                collection.removeIf { existingTransactionIdsInDb[it] == VERIFIED }
+            }
```
lankydan: I think I'd prefer it was:

```kotlin
transactionsToCheck.removeIf { existingTransactionIdsInDb[it] == VERIFIED }
transactionsToRetrieve.removeIf { existingTransactionIdsInDb[it] == VERIFIED }
```

I think it's clearer this way. I was wondering why we were putting it into a list etc.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -208,4 +299,6 @@ class TransactionBackchainReceiverFlowV1(
         result = 31 * result + session.hashCode()
         return result
     }
-}
\ No newline at end of file
+}
+
+class InvalidBackchainException(message: String) : CordaRuntimeException(message, null)
```
lankydan: Move this to its own file imo.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -189,6 +201,85 @@ class TransactionBackchainReceiverFlowV1(
...
+            )
+
+            // Remove the transaction from the "to retrieve" and "to check" collections if they are in our DB,
+            // and they are verified
+            listOf(transactionsToCheck, transactionsToRetrieve).forEach { collection ->
+                collection.removeIf { existingTransactionIdsInDb[it] == VERIFIED }
+            }
+
+            val newTransactionsToCheck = mutableListOf<SecureHash>()
+            transactionsToCheck.forEach { transactionId ->
```
lankydan, line 226: This logic works.

It does highlight how there is still more optimisation work to be done, probably with the outcome of more logic being moved to the database worker side to reduce the back and forth between the flow and database worker.

We call`findTransactionIdsAndStatuses` at the start of the while loop:

```kotlin
while (transactionsToCheck.isNotEmpty()) {
            val transactionsFromDb = utxoLedgerPersistenceService.findTransactionIdsAndStatuses(transactionsToCheck)
```

And then for the ones that are not verified we then fetch the transaction contents individually:

```kotlin
if (existingTransactionIdsInDb[transactionId] != null) {
                    // Fetch the transaction object from the database, so we can get the dependencies
                    val transactionFromDb = utxoLedgerPersistenceService.findSignedTransaction(
                        transactionId,
                        UNVERIFIED
                    )
```

Logically this is all fine, but as the original point of this change is to improve performance there is still a way to go.

On the upside, on the happy path where transaction chains haven't been received or have already been verified, this solution should work better than what we had before (skipping the session communication).

I feel like we might have a ticket for doing more work on the database worker already, but I'm not 100% sure.

Doing this work on the database side will save us a lot of back-and-forths between workers. What the code is doing, from my understanding, is going through a transaction's dependencies and then finding the dependencies' dependencies and so on, until we either hit a verified transaction, the end of a chain, or a transaction that we don't have locally and need from a peer.

We could achieve this on the database worker by going through a transaction's dependency chain until we hit a point that says, "we need to get this transaction from a peer" or "we hit the end of the chain / verified transaction". We then return to the flow only with transactions that need to be retrieved from a peer.

Nandor mentioned that he had a look at using recursion to simplify the solution (which is a common solution for tree traversal). I felt that it wasn't appropriate for our current implementation as the stack could start to build up as we hold transaction objects in the stack; however, I realised we should be able to keep the transactions out of the stack using private methods which lookup a transaction and only return its dependencies.

This is also code that could be moved to the database worker.

Because it feels like there is more work to do, I suggest we go with what we've got here but then start a new change that continues this work (while Nandor has it in his head).

lankydan: @blsemo @vlajos @relyafi @nkovacsx what do you think?

vlajos: It sounds reasonable to me.
Assuming the current solution is correct.
Last time I checked, I did not see how the topological sort is populated with the bypassed items or why it is OK to leave them out from the topological sort. (But this might have been fixed already/maybe I just overlooked something.)

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -189,6 +201,85 @@ class TransactionBackchainReceiverFlowV1(
...
+                        // Add the dependencies to the "to retrieve" set (we might remove them later)
+                        transactionsToRetrieve.addAll(transactionFromDb.dependencies)
+                        // and we also need to check them
+                        newTransactionsToCheck.addAll(transactionFromDb.dependencies)
+
+                        // Once we added the unverified transactions' dependencies into the "to retrieve" set
+                        // we can remove those from the set as we don't need to get those from the initiator
+                        // but only if we "know" its group parameters, otherwise we need to keep it there
+                        if (version == TransactionBackChainResolutionVersion.V1) {
+                            transactionsToRetrieve.remove(transactionId)
```
vlajos: Last week's news, but 5.0 compatibility modes are not needed anymore.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
lankydan: Yeah, @nkovacsx reminded me to double check that part, let me take a look.

lankydan: Ok, I think @vlajos is correct.

We should be adding all the unverified transactions that we already have in the database to the topological sort. If we do not do this, then while we'll receive the dependencies for transactions in the backchain, we will not verify the ones that exist in the database as unverified, thus meaning that the transaction we're receiving the backchain for could move to verified even though one of its dependencies (or dependencies' dependencies) might be invalid.

You'll only catch this in an E2E or integration test. We don't have integration tests around this, and setting up this scenario in an E2E test can be a pain. So you probably haven't seen it but it could happen.

I think it would go after this line?

```
// Remove the transaction from the "to retrieve" and "to check" collections if they are in our DB,
            // and they are verified
            listOf(transactionsToCheck, transactionsToRetrieve).forEach { collection ->
                collection.removeIf { existingTransactionIdsInDb[it] == VERIFIED }
            }
```

Meaning that all unverified transactions are added to the topological sort. But after thinking about it again if the `TopologicalSort` line looks like `sortedTransactionIds.add(retrievedTransaction.id, dependencies)` then we're missing the parent transaction id that is required for the sorting.

This code then seems related:

```
// Add the dependencies to the "to retrieve" set (we might remove them later)
transactionsToRetrieve.addAll(transactionFromDb.dependencies)
// and we also need to check them
newTransactionsToCheck.addAll(transactionFromDb.dependencies)
```

But the comment of "we might remove them later" makes me unsure if this is correct.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
nkovacsx: Good point, thanks!

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
nkovacsx: ok

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
nkovacsx: Will wait for @vlajos ' PR to be merged (the one that removes the backwards compatibility code) then rebase this PR.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -182,6 +194,94 @@ class TransactionBackchainReceiverFlowV1(
         }
     }
 
+    @Suppress("NestedBlockDepth")
+    private fun handleExistingTransactionsAndTheirDependencies(
+        existingTransactionIdsInDb: MutableMap<SecureHash, TransactionStatus>,
+        transactionsToRetrieve: LinkedHashSet<SecureHash>,
+        sortedTransactionIds: TopologicalSort
+    ) {
+        var transactionsToCheck = transactionsToRetrieve.toMutableList()
```
vlajos, line 203: Could we subtract here the `existingTransactionIdsInDb`s ? Those are supposed to be processed already in the main loop's previous iteration.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -182,6 +194,94 @@ class TransactionBackchainReceiverFlowV1(
...
+            if (invalidTransactions.isNotEmpty()) {
+                throw InvalidBackchainException(
+                    "Found the following invalid transaction(s) during back-chain resolution: " +
+                            "$invalidTransactions. Back-chain resolution cannot be continued."
+                )
+            }
+
+            // Store the Verified/Unverified transactions from the database
+            existingTransactionIdsInDb.putAll(
+                transactionsFromDb.filterValues {
```
vlajos: Do we need this filter?
We cannot have any INVALID ones here anymore.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -182,6 +194,94 @@ class TransactionBackchainReceiverFlowV1(
...
+                        // we can remove the parent transaction from the set as we don't need to get that from
+                        // the initiator but only if we know its group parameters
+                        if (fetchGroupParametersAndHashForTransaction(transactionFromDb).second != null) {
+                            transactionsToRetrieve.remove(transactionId)
+                        }
+                    } else {
+                        log.trace {
+                            "Transaction with ID $transactionId is present in the database with unverified status " +
+                                    "but could not fetch the signed transaction object and its dependencies."
+                        }
```
vlajos: Is not this the `else` of `if (transactionFromDb != null) {` ?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -182,6 +194,117 @@ class TransactionBackchainReceiverFlowV1(
...
+
+            val newTransactionsToCheck = mutableListOf<SecureHash>()
+            transactionsToCheck.forEach { transactionId ->
+                if (existingTransactionIdsInDb[transactionId] != null) {
+                    // Fetch the transaction object from the database, so we can get the dependencies
+                    // Also fetch its status because it might have changed since the last time
+                    val (transactionFromDb, status) = utxoLedgerPersistenceService.findSignedLedgerTransactionWithStatus(
+                        transactionId,
+                        UNVERIFIED
+                    ) ?: throw InvalidBackchainException("Transaction with ID $transactionId not exist.")
```
vlajos: This `throw` looks a  bit strong to me.
If the transaction is not available in the database, we could just treat it like it was not there from the beginning, we will request it from another party anyway.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1.kt
```diff
@@ -182,6 +194,117 @@ class TransactionBackchainReceiverFlowV1(
...
+                            throw InvalidBackchainException(
+                                "Found the following invalid transaction during back-chain resolution: " +
+                                        "$transactionId. Backchain resolution cannot be continued."
+                            )
+                        }
+
+                        UNVERIFIED -> {
+                            // If the status is still UNVERIFIED we can continue as expected
+                            if (transactionFromDb == null) {
+                                throw InvalidBackchainException(
```
vlajos: Same here.

vlajos (commented):
LGTM, apart from the Gradle conflict

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1Test.kt
```diff
@@ -72,21 +80,733 @@ class TransactionBackchainReceiverFlowV1Test {
...
+            StateRef(TX_ID_3, 0)
+        ))
+        whenever(retrievedTransaction1.referenceStateRefs).thenReturn(emptyList())
+        whenever(retrievedTransaction1.metadata).thenReturn(tx1Metadata)
+
+        // TX3
+        whenever(utxoLedgerPersistenceService.findSignedLedgerTransactionWithStatus(eq(TX_ID_3), eq(UNVERIFIED)))
+            .thenReturn(Pair(retrievedTransaction3, UNVERIFIED))
+
+        whenever(retrievedTransaction3.id).thenReturn(TX_ID_1)
```
blsemo: Is that correct?
```suggestion
        whenever(retrievedTransaction3.id).thenReturn(TX_ID_3)
```

blsemo (commented):
Looks ok to me, but I think there is an ID wrong in one of the unit tests.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1Test.kt
```diff
@@ -72,21 +80,733 @@ class TransactionBackchainReceiverFlowV1Test {
...
+            eq(TransactionBackchainRequestV1.Get(setOf(TX_ID_1)))
+        )
+    }
+
+    /**
+     * This test is simulating a scenario where the transaction's status was originally UNVERIFIED in the database,
+     * but then it changed to VERIFIED while the flow was in flight.
+     */
+    @Test
+    fun `if transaction status changed to verified from unverified it will not be retrieved`() {
```
lankydan, line 777: Just to double-check, the same behaviour would apply to any of the dependencies of a transaction?

lankydan (dismissed):
LGTM, the comments are really helpful because this code has a lot of edge-cases that need handling.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1Test.kt
nkovacsx: good point thanks!

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainReceiverFlowV1Test.kt
nkovacsx: yeah but I added another test case 😄

lankydan (dismissed):

lankydan (dismissed):
Approving based on previous approval.

vlajos (dismissed):

lankydan (approved):

vlajos (approved):

blsemo (approved):
LGTM

PR #4593, opened 2023-09-06
CORE-16911 Fix TLS certificate verification in case of multiple trust roots

YashNabar:
In a scenario where the MGM is onboarded with multiple TLS trust roots, we verify a TLS certificate against all trust roots from the group policy while creating locally-hosted identities. Before this change, if verification failed for one of the roots with `SignatureException`, locally-hosted identity creation failed without checking the other trust roots. This change fixes the issue by handling such `SignatureException`s.

Tested using a multi-cluster setup deployed using [app-simulator](https://github.com/corda/corda-runtime-os/tree/release/os/5.1/applications/tools/p2p-test/app-simulator) scripts. MGM onboarded with two TLS trust roots:
- `corda.group.trustroot.tls.0` - random certificate causing `SignatureException`
- `corda.group.trustroot.tls.1` - actual root certificate used to sign the clusters' TLS certificates


components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/HostedIdentityEntryFactory.kt
```diff
@@ -269,6 +270,8 @@ internal class HostedIdentityEntryFactory(
                     true
                 } catch (e: InvalidKeyException) {
                     false
+                } catch (e: SignatureException) {
+                    false
```
yift-r3, line 274: Can you add a unit test?

components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/HostedIdentityEntryFactory.kt
```diff
@@ -267,7 +267,7 @@ internal class HostedIdentityEntryFactory(
                 try {
                     certificate.verify(rootCertificate.publicKey)
                     true
-                } catch (e: InvalidKeyException) {
+                } catch (e: GeneralSecurityException) {
```
williamvigorr3: Are you sure this is what we want to do? I think some of these exceptions can be thrown if the certificate can never be valid e.g. its encoding is completely incorrect. So we should probably reject the request in that case. Probably we want to just catch `InvalidKeyException` and `SignatureException` (as @YashNabar was doing before).

williamvigorr3 (approved):
LGTM

PR #4599, opened 2023-09-06
CORE-15391 Add support for custom checkpoint state

owenstanford:
Add support for custom state to be stored in the checkpoint. This has been added to allow external flow event handlers to store local state that can span flow suspensions. As part of this change the state used for metrics has been migrated to use the new custom state.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/state/impl/FlowCheckpointImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/state/impl/FlowCheckpointImpl.kt
@@ -1,8 +1,11 @@
 package net.corda.flow.state.impl
 
+import com.fasterxml.jackson.databind.ObjectMapper
 import java.nio.ByteBuffer
 import java.time.Instant
 import net.corda.data.ExceptionEnvelope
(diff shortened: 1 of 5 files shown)


components/flow/flow-service/src/main/kotlin/net/corda/flow/state/impl/FlowCheckpointImpl.kt
```diff
@@ -26,6 +29,19 @@ class FlowCheckpointImpl(
     instantProvider: () -> Instant
 ) : FlowCheckpoint {
 
+    init {
+        if (checkpoint.customState == null) {
```
LWogan, line 40: Should we have an initCustomeState() api, similar to initFlowState() instead, and call it in the CheckpointInitializer?

owenstanford: We need a catch-all for the upgrade edge condition, in theory we could read an older version of the checkpoint and bypass the initFlowState(), I'll add a comment highlighting this

LWogan (approved):
lgtm

PR #4600, opened 2023-09-06
CORE-16982: Configurable Plugin Timeout

jujoramos:
Make the timeout used by the create and delete topics plugin
configurable so the users of the chart can fine tune the flag when
required.


tools/plugins/topic-config/src/main/kotlin/org/apache/kafka/clients/admin/AmdinExt.kt
davidcurrie, line 1: `AmdinExt` -> `AdminExt`

davidcurrie (changes requested):

tools/plugins/topic-config/src/main/kotlin/org/apache/kafka/clients/admin/AmdinExt.kt
jujoramos: 👍

davidcurrie (approved):
LGTM

PR #4601, opened 2023-09-07
CORE-16177 Creating generic MessageProducer for Kafka/DB/RPC

ben-millar:
### Context:
Today, Corda has external event producers which are capable of publishing messages to either a Kafka or a DB message bus, and the implementation of these producers has been done with Kafka in mind.

As part of the [multi-source event mediator](https://r3-cev.atlassian.net/browse/CORE-15754) implementation, we will have a new subscription which should be able to send messages (NB: non-transactionally, as of today) to endpoints which may be Kafka, DB or RPC; the subscription code should be agnostic of the implementation of these endpoints, and so we now require a more generic messaging system which isn't tied to Kafka.

### Changes:
- A new `MessageProducer` interface has been created which exposes `send()` methods similar to the existing `CordaProducer` interface, but which takes a new more generic `CordaMessage` rather than the `CordaProducerRecord` which was based on Kafka records. Additionally, no support for transactions has been added.
- A generic `CordaMessage` data class has been created which has a single `val payload: T?`, with any additional properties being stored in a mutable map, allowing us to encode Kafka, DB and RPC destined messages within.
- A `MessageProducerImpl` has been created which implements the `MessageProducer` interface, wrapping the `CordaProducer` allowing for non-transactional sending of messages over Kafka or a DB message bus.
- **Important**: This PR also introduces the `kotlinx.coroutines` library into Corda, allowing us to use the full suite of coroutine functionality provided by JetBrains (previously we were using `kotlin.coroutines` which contains only the core components).

### Notes:
For speed of development, new classes have been created for this functionality (which may act as interfaces over existing classes); this prevents us from having to refactor existing code and from potentially breaking unrelated code which uses the existing infrastructure. In the future, we may wish to come back and coalesce these two implementations.


libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/MessageProducer.kt
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    /**
+     * Asynchronously sends a generic [CordaMessage], and invoke the provided callback when the message has been acknowledged.
+     * If the developer would like this to be a blocking call, this can be achieved through the callback.
+     *
+     * @param message The message to send.
+     * @param callback A user-supplied callback to execute when teh record has been successfully sent or an error has occurred.
+     */
+    fun send(message: CordaMessage<*>, callback: Callback?)
```
mbrkic-r3: We might want to use `CompletableFuture` instead of callbacks. Kafka producer is async while RPC client is supposed to be sync. Multi-source subscription should handle producers in the same way, so I guess we need to support only one type here. Better choice would probably be async, as we can always use it in a sync way if we want (which might not work as intended for RPC if RPC client won't support it).

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/MessageProducer.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Send a batch of [CordaMessage] instances to their respective destinations. These should all be of the same type
+     * (E.g. Kafka, DB, RPC).
+     *
+     * @param messages the list of [CordaMessage] to be sent.
+     */
+    fun sendMessages(messages: List<CordaMessage<*>>)
```
mbrkic-r3: Batch sending might improve performance, but I'm not sure it will be supported by RPC client. Multi-source subscription should handle producers in the same way, so they should all implement it. RPC producer might still initially implement this method by sending single requests via RPC client, I guess.
Note that this method should also return something.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/KafkaMessageProducerImpl.kt
```diff
@@ -0,0 +1,223 @@
...
+        try {
+            block()
+            getOrCreateBatchPublishTracing(config.clientId).complete()
+        } catch (ex: Exception) {
+            getOrCreateBatchPublishTracing(config.clientId).abort()
+            handleException(ex, operation)
+        }
+    }
+
+    private fun handleException(ex: Exception, operation: String) {
```
mbrkic-r3: We would like to handle recoverable exceptions here, including those where KafkaClient has to be recreated, but that can be left for later when we put more focus on resiliency.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/KafkaMessageProducerImpl.kt
```diff
@@ -0,0 +1,223 @@
...
+                "CordaKafkaProducer failed to close producer safely. This can be observed when there are " +
+                        "no reachable brokers. ClientId: ${config.clientId}", ex
+            )
+        } finally {
+            (producerMetricsBinder as? AutoCloseable)?.close()
+        }
+    }
+}
+
+fun CordaMessage<*>.toCordaProducerRecord() : CordaProducerRecord<*, *> {
```
mbrkic-r3: This should be private

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/KafkaMessageProducerImpl.kt
ben-millar: Good point, I've added a @TODO to the code here.

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/CordaMessage.kt
```diff
@@ -0,0 +1,72 @@
...
+data class CordaMessage<T: Any>(
+    val payload: T?,
+    val props: MutableMap<String, Any> = mutableMapOf()
+) {
+    /**
+     * Adds a new property to the internal storage.
+     *
+     * @param property A key value [Pair] to insert into our additional properties map.
+     */
+    fun addProperty(property: Pair<String, Any>) {
```
mbrkic-r3: Do we get something with wrapping key and value into `Pair`?

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/MessageProducer.kt
```diff
@@ -0,0 +1,32 @@
+package net.corda.messagebus.api.producer
+
+import kotlinx.coroutines.Deferred
+
+interface MessageProducer : AutoCloseable {
+    /**
+     * Defines the callback for post-send events. If there was an exception it will be provided on this callback.
+     */
+    fun interface Callback {
```
mbrkic-r3: Why do we need callback if we are going to use `Deferred`?

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/MessageProducer.kt
```diff
@@ -0,0 +1,32 @@
...
+        fun onCompletion(exception: Exception?)
+    }
+
+    /**
+     * Asynchronously sends a generic [CordaMessage], and invoke the provided callback when the message has been acknowledged.
+     * If the developer would like this to be a blocking call, this can be achieved through the callback.
```
mbrkic-r3: How is callback related to blocking call?

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/MessageProducer.kt
```diff
@@ -0,0 +1,32 @@
...
+     *
+     * @param message The [CordaMessage] to send.
+     * @return [Deferred] instance representing the asynchronous computation result, or null if the destination doesn't
+     * provide a response.
+     * */
+    fun send(message: CordaMessage<*>, callback: Callback?) : Deferred<*>?
```
mbrkic-r3: Response should also be a (deferred) `CordaMessage`
We should also consider how we would return/handle error messages that are coming back (that will be more relevant for RPC Producer so maybe we'll need to adjust interface when we start implementing that).

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/MessageProducer.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Send a batch of [CordaMessage] instances to their respective destinations. These should all be of the same type
+     * (E.g. Kafka, DB, RPC).
+     *
+     * @param messages the list of [CordaMessage] to be sent.
+     * @return List of [Deferred] instances representing the asynchronous computation results, or null if the
+     * destination doesn't provide a response.
+     */
+    fun sendMessages(messages: List<CordaMessage<*>>) : List<Deferred<*>>?
```
mbrkic-r3: Response should also be a list of (deferred) `CordaMessage`s

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/KafkaMessageProducerImpl.kt
```diff
@@ -0,0 +1,231 @@
...
+import org.apache.kafka.common.errors.ProducerFencedException
+import org.apache.kafka.common.errors.TimeoutException
+import org.apache.kafka.common.errors.UnsupportedForMessageFormatException
+import org.apache.kafka.common.errors.UnsupportedVersionException
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+class KafkaMessageProducerImpl(
+    private val config: ResolvedProducerConfig,
+    private val producer: Producer<Any, Any>,
```
mbrkic-r3: Wouldn't it be easier to reuse `CordaProducer`, which is already configured and has metrics implemented? That would also cover DB version of the Producer. See https://github.com/corda/corda-runtime-os/pull/4616/files#diff-56439326c80558141e01bca33e27cee75bb223543e5e9c7b6092fb4f3817b380

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/KafkaMessageProducerImpl.kt
```diff
@@ -0,0 +1,231 @@
...
+    override fun send(message: CordaMessage<*>, callback: MessageProducer.Callback?): Deferred<*>? {
+        val headers = message.getProperty<Headers>("headers")
+        val partition = message.getPropertyOrNull<Int>("partition")
+
+        getOrCreateBatchPublishTracing(clientId).begin(listOf(headers))
+        tryWithCleanupOnFailure("send single message") {
+            sendMessage(message, partition, callback)
+        }
+
+        return null
```
mbrkic-r3: Seems this is not implemented (returning `Deferred` response)

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/KafkaMessageProducerImpl.kt
```diff
@@ -0,0 +1,231 @@
...
+            it.getProperty<Headers>("headers")
+        })
+        tryWithCleanupOnFailure("send multiple message") {
+            messages.forEach { message ->
+                val partition = message.getPropertyOrNull<Int>("partition")
+                sendMessage(message, partition)
+            }
+        }
+
+        return null
```
mbrkic-r3: Seems this is not implemented (returning `Deferred` responses)

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/KafkaMessageProducerImpl.kt
```diff
@@ -0,0 +1,231 @@
...
+                    ctx.errorAndFinish(ex)
+                } else {
+                    ctx.finish()
+                }
+            }
+        }
+    }
+
+    override fun send(message: CordaMessage<*>, callback: MessageProducer.Callback?): Deferred<*>? {
+        val headers = message.getProperty<Headers>("headers")
```
mbrkic-r3: Since Strings like "headers" are used multiple times in a file it's better to define them as constants.

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/MessageProducer.kt
ben-millar: Discussed offline; will remove.

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/MessageProducer.kt
ben-millar: Outdated comment, I've updated this.

wzur-r3 (dismissed):
LGTM for files owned by BLT

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/CordaMessage.kt
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @param key The key of the property to retrieve.
+     * @return The property associated with the given key, or null if not found.
+     */
+    fun getPropertyOrNull(key: String) : Any? {
+        return props[key]
+    }
+
+    /**
+     * Retrieves a property of a specific type from the additional properties map, returning null if not found or if the cast fails.
```
simon-johnson-r3: It throws if the cast fails.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/MessageProducerImpl.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    private companion object {
+        private val log: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+    }
+
+    override fun send(message: CordaMessage<*>): Deferred<CordaMessage<*>?> =
+        CompletableDeferred<CordaMessage<*>>().apply {
+            producer.send(message.toCordaProducerRecord()) { ex ->
+                ex?.let { completeExceptionally(ex) }
+            }
```
simon-johnson-r3: Don't you need a `complete` here?

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/MessageProducerImpl.kt
```diff
@@ -0,0 +1,57 @@
...
+            )
+        } finally {
+            (producerMetricsBinder as? AutoCloseable)?.close()
+        }
+    }
+}
+
+private fun CordaMessage<*>.toCordaProducerRecord() : CordaProducerRecord<*, *> {
+    return CordaProducerRecord(
+        topic = this.getProperty<String>("topic"),
```
simon-johnson-r3: This knowledge would seemingly be required outside this class, should it be defined publicly somewhere?

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/MessageProducer.kt
```diff
@@ -0,0 +1,14 @@
...
+
+interface MessageProducer : AutoCloseable {
+    /**
+     * Asynchronously sends a generic [CordaMessage], and returns any result/error through a [Deferred] response.
+     *
+     * @param message The [CordaMessage] to send.
+     * @return [Deferred] instance representing the asynchronous computation result, or null if the destination doesn't
+     * provide a response.
+     * */
+    fun send(message: CordaMessage<*>) : Deferred<CordaMessage<*>?>
```
simon-johnson-r3: You can only usefully use a `Deferred` in a `CoroutineScope` where you would `await` completion. I expect this will very quickly ripple through Corda, where previously we passed callbacks though to async methods, now we will need to manage coroutine scopes in a number of places. Are you sure this is what you want?

libs/messaging/kafka-message-bus-impl/src/test/kotlin/net/corda/messagebus/kafka/producer/MessageProducerImplTest.kt
```diff
@@ -0,0 +1,76 @@
...
+            message.getProperty<String>("topic"),
+            message.getProperty("key"),
+            message.payload
+        )
+
+        verify(cordaProducer).send(eq(expected), any())
+    }
+
+    @Test
+    fun testSendWithError() {
```
simon-johnson-r3: You never test send with success, you only test the underlying `send` was called. I think if you wrote such a test you'd find it failed.

simon-johnson-r3 (changes requested):

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/MessageProducerImpl.kt
ben-millar: The `Deferred` return is primarily to be used for the `RPC` implementation; for Kafka and DB (handled by this impl) there'll never be a return value. We just return exceptions (previously this was done via a callback).

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/MessageProducerImpl.kt
ben-millar: Currently it's only required in the class. In a future PR, we'll connect the multi-source subscription pattern to this producer and at that stage it would make sense to extract this functionality so it can be used by both. (that's my thinking, anyway)

libs/messaging/kafka-message-bus-impl/src/test/kotlin/net/corda/messagebus/kafka/producer/MessageProducerImplTest.kt
ben-millar: I'm checking that `send()` was called on the underlying `CordaProducer` (which is contained within this `MessageProducer`), and that the `CordaMessage` was successfully converted into the expected, kafka-compatible, `CordaProducerRecord`. At that stage, the logic is tested by the `CordaProducer` tests.

When we're using this `MessageProducer` in anger (as part of the multi-source subscription), we'll be able to implement proper integration and E2E tests on this logic.

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/MessageProducer.kt
ben-millar: As part of the performance redesign, external event calls are going to become blocking, sequential calls so this is intentional. In terms of managing the coroutine scopes, my thinking was that they would be tied to the `Lifecycle` component of the relevant producer, e.g.:

```
RPCMessageProducerImpl(...) : MessageProducer {
    private val job: Job = Job()
    private val coroutineScope: CoroutineScope = CoroutineScope(Dispatchers.IO + job)

    ...
    override fun close() { 
        job.cancel() // When we close, we cancel all coroutines within our scope
    }
}
```

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/producer/CordaMessage.kt
ben-millar: I've updated the comment, good catch! This (I.e., returning null on a failing cast) was the previous behaviour, but I felt that behaviour made things unclear to the developer.

simon-johnson-r3: For the record I agree on the behaviour. You'd call this if you expected the property to be not set. But you'd still expect someone hadn't put an unexpected type in there.

simon-johnson-r3:
Myself and @ben-millar discussed this offline. We didn't yet reach a conclusion, but I'll post some of the important point here so there's a record.

At this level of abstraction I should be able to do this:
```
myProducer.send(someMessage)
```
without caring what the underlying comms layer is. That is after all the very reason we abstract. But it seems like here I'd need to do:
```
class MyClass {

  // This needs to be called in a coroutine scope, but in non-RPC cases it doesn't act like a coroutine at all
  fun something(...) {
    ...
    if (RPCComms) {
      var result = myProducer.send(someMessage).await()
      process(results)
    } else {
      // Still need to call await to trigger any throws
      myProducer.send(someMessage).await()
    }
  }

  // For non blocking comms, we get result back here
  override fun onNext(result) {
    process(results)
  }

  fun processResult(result) {
    // common result processing code
  }
}
```

The challenge here is that in the non blocking producer case, in our current codebase, the code which handles the response can be in a completely different place to the one which sends the message.

Feels like the correct point of abstraction is not at the producer at all but at a higher level, e.g. for RPC, with the sender: https://github.com/corda/corda-runtime-os/blob/a189d56f9f5a14b30e6bd29a5192b26cad15c49f/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/factory/CordaPublisherFactory.kt#L67

At this level the pattern is the same no matter what kind of comms infrastructure is in play. It would be very easy to make a HTTP sender which completed that future before returning, unlike the Kafka one which does it some time later. There is no change to the contract of this class.

I'm not sure if it is the rpc pattern you are wanting to replace. If it's another one, the same principle would apply, and I think we should try and find a higher level of abstraction which meets the send/response type interface instead of making http feel like Kafka at low level - because the underlying patterns are totally different at this point.

I also wouldn't conflate the db implementation into this. That is very much a Kafka emulation over db. I don't see it as a third thing so much as a specialisation of message bus.

mbrkic-r3:
I see classes were left in wrong packages and I guess this is causing confusion. One of my previous comments was

> Wouldn't it be easier to reuse CordaProducer, which is already configured and has metrics implemented? That would also cover DB version of the Producer. See https://github.com/corda/corda-runtime-os/pull/4616/files#diff-56439326c80558141e01bca33e27cee75bb223543e5e9c7b6092fb4f3817b380

That was addressed, but I should have been more specific and clarify it was also supposed to be moved out of the Kafka module since this PR should not reference Kafka at all. We have `CordaProducer` which is abstraction above Kafka and DB implementations. With `MessageProducer` we are now introducing abstraction level above that. So `MessageProducerImpl` should be actually renamed to `MessageBusProducer` and we'll also have `RpcProducer` (that will use `RpcClient`).

Sending messages will now support request-reply messaging pattern, and it's usage will be similar to this:
```
var result = myProducer.send(someMessage)
if (myProducer.isRequestRepy()) {   // or:   if (result.message != null)
  process(result.message) 
}
```

Function `send` will actually return `Deferred<Result>` (or `CompletableFuture`, if `Deferred` doesn't bring any advantages, we would appreciate comments on that as we don't have experience with Kotlin coroutines). There is no `onNext()` function (new RPC patterns is not related in any way with existing RPC over Kafka). I'm not sure whether this resolves issues related to blocking/nonblocking usage?

wzur-r3 (dismissed):
LGTM for files owned by BLT

driessamyn:
> Myself and @ben-millar discussed this offline. We didn't yet reach a conclusion, but I'll post some of the important point here so there's a record.
> 
> At this level of abstraction I should be able to do this:
> 
> ```
> myProducer.send(someMessage)
> ```
> 
> without caring what the underlying comms layer is. That is after all the very reason we abstract. But it seems like here I'd need to do:
> 
> ```
> class MyClass {
> 
>   // This needs to be called in a coroutine scope, but in non-RPC cases it doesn't act like a coroutine at all
>   fun something(...) {
>     ...
>     if (RPCComms) {
>       var result = myProducer.send(someMessage).await()
>       process(results)
>     } else {
>       // Still need to call await to trigger any throws
>       myProducer.send(someMessage).await()
>     }
>   }
> 
>   // For non blocking comms, we get result back here
>   override fun onNext(result) {
>     process(results)
>   }
> 
>   fun processResult(result) {
>     // common result processing code
>   }
> }
> ```
> 
> The challenge here is that in the non blocking producer case, in our current codebase, the code which handles the response can be in a completely different place to the one which sends the message.
> 
> Feels like the correct point of abstraction is not at the producer at all but at a higher level, e.g. for RPC, with the sender:
> 
> https://github.com/corda/corda-runtime-os/blob/a189d56f9f5a14b30e6bd29a5192b26cad15c49f/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/factory/CordaPublisherFactory.kt#L67
> 
> At this level the pattern is the same no matter what kind of comms infrastructure is in play. It would be very easy to make a HTTP sender which completed that future before returning, unlike the Kafka one which does it some time later. There is no change to the contract of this class.
> 
> I'm not sure if it is the rpc pattern you are wanting to replace. If it's another one, the same principle would apply, and I think we should try and find a higher level of abstraction which meets the send/response type interface instead of making http feel like Kafka at low level - because the underlying patterns are totally different at this point.
> 
> I also wouldn't conflate the db implementation into this. That is very much a Kafka emulation over db. I don't see it as a third thing so much as a specialisation of message bus.

I think this make sense.

I do actually wonder what the advantage is of this abstraction? A message bus send is so very different from the request/response pattern that layering an abstraction over both at the same time is a bit odd.
What benefit are we gaining by this?

mbrkic-r3:
More clarifications:

- While functions for creation of `Publisher` and `RPCSender` are in the same `CordaPublisherFactory` they create objects which implement completely different interfaces (and should most likely be in a different factories). One is just for sending messages without possibility to receive reply (and it's pretty much message bus, even Kafka specific) and the other is for the request-reply messaging pattern.
- If we don't have a layer of abstractions above those two, this is exactly what we would have to do:
```
 if (myProducer is RPCComms) {
       var result = myProducer.send(someMessage).await()
       process(results)
 } else {
       myProducer.send(someMessage).await()
 }

```
- Having such layer of abstraction was one of the requirements for Multi-Source Event Mediator (https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4552491009/Corda+Scalable+Topology+Design#Multi-Source-Subscription-Pattern - it's called "Publisher" there):
> The publisher responsible of processing the output records can be of many types. The subscription will use Kafka publishers for Kafka events and RPC publisher for RPC events, etc…
- Advantage of abstraction is to have Multi-Source Event Mediator implementation that can work with any type of "Publisher", without need to have dedicated logic/handle special cases for each type
- This PR needs further changes to fix issues mentioned in my previous comment

mbrkic-r3:
I'll take over this PR (since Ben is on vacation) and make required changes, I'm just wondering whether we'll go with Kotlin coroutine types or use `CompletableFutures` instead.
@simon-johnson-r3 what is your recommendation?

mbrkic-r3:
@simon-johnson-r3 @driessamyn 
Refactored, please re-review.

buildSrc/src/main/groovy/corda.common-library.gradle
```diff
@@ -22,9 +22,26 @@ configurations {
     configureEach {
         resolutionStrategy {
             dependencySubstitution {
-                substitute module('antlr:antlr') using project(':libs:antlr')
-                substitute module('de.javakaffee:kryo-serializers') using project(':libs:serialization:kryo-serializers')
-                substitute module('software.amazon.awssdk:cloudwatch') using project(':libs:awssdk')
+                substitute module('antlr:antlr') using variant(project(':libs:antlr')) {
```
simon-johnson-r3: heads up, you'll get a conflict here now

simon-johnson-r3:
@mbrkic-r3 

> I'm just wondering whether we'll go with Kotlin coroutine types or use CompletableFutures instead.
@simon-johnson-r3 what is your recommendation?

Personally I have no problems with using Kotlin coroutines as a general principle. However I would ask here what we get from them? A completable future is the means by which we do things now, and I wonder whether we want to change that pattern? It seems to offer us the same type of api contract without enforcing coroutines across the entire of the message patterns library (because that is the inevitable effect of this at least, possible propagating even further outside them)

Not least, we should consider if we want to present Kotlin coroutines to our entire developer group. Again, I'd not be opposed to that in cases where we see very clear benefits. But I'm not really convinced this is one of them.

simon-johnson-r3:
@mbrkic-r3 My initial concerns around this PR have been addressed (largely the point about completing on success).

My only concern now is the one about the correct place of abstraction, which @driessamyn also commented on here https://github.com/corda/corda-runtime-os/pull/4601#issuecomment-1722882117

I have to say, I'm not close to this design at all. What you have here would work if you want a "fire and forget" type of producer towards multiple sources. If that's what you want, then I guess this is it. I was kind of expecting the switch to some RPC system at a higher level of abstraction though, but perhaps there is the need for this, I just don't know.

mbrkic-r3:
> Not least, we should consider if we want to present Kotlin coroutines to our entire developer group. Again, I'd not be opposed to that in cases where we see very clear benefits. But I'm not really convinced this is one of them.

@driessamyn what's your call on this?

mbrkic-r3:
> My only concern now is the one about the correct place of abstraction, which @driessamyn also commented on here https://github.com/corda/corda-runtime-os/pull/4601#issuecomment-1722882117

This was shortly discussed offline, it seems the only issue is naming, as "producer" is used in other packages and has a different meaning.

thiagoviana (approved):

simon-johnson-r3 (approved):

ronanbrowne (approved):
LGTM - blt

wzur-r3 (approved):
LGTM for Gradle files

PR #4602, opened 2023-09-07
CORE-16319: State Manager API & Implementation

jujoramos:
Add initial API and basic implementation for the State Manager.


jujoramos:
Split from https://github.com/corda/corda-runtime-os/pull/4501.
Corda API changes: https://github.com/corda/corda-api/pull/1240.

libs/state-manager/state-manager-impl/src/test/kotlin/net/corda/libs/statemanager/impl/StateManagerImplTest.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private val stateManager = StateManagerImpl(
+        stateRepository = stateRepository,
+        entityManagerFactory = entityManagerFactory,
+        serializerFactory = serializerFactory,
+    )
+
+    @Test
+    fun `get state creates tx and uses repository`() {
```
sprintengine: we prob should add a few more unit tests for this class and the repository functions, but we do have the integration test coverage

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    override fun findUpdatedBetween(
+        entityManager: EntityManager,
+        start: Instant,
+        finish: Instant
+    ): Collection<StateEntity> {
+        return entityManager
+            .createNamedQuery(QUERY_STATES_BY_UPDATED_TIMESTAMP_NAME.trimIndent(), StateEntity::class.java)
+            .setParameter(STAR_TIMESTAMP, start)
```
sprintengine: minor - typo

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/Metadata.kt
```diff
@@ -0,0 +1,40 @@
...
+        Long::class.java,
+        Float::class.java,
+        Double::class.java,
+        Character::class.java
+    )
+
+    private fun isPrimitiveOrBoxedValue(value: V): Boolean {
+        return when {
+            value::class.java.isPrimitive -> true
+            supportedBoxedPrimitives.contains(value::class.java) -> true
```
sprintengine: should we use `supportedBoxedPrimitives.any { it.isAssignableFrom(value::class.java) }` instead?

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
jujoramos: 👍

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/State.kt
```diff
@@ -0,0 +1,34 @@
...
+    val state: S,
+
+    /**
+     * Identifier for the state.
+     */
+    val key: String,
+
+    /**
+     * Arbitrary Map of primitive types that can be used to store and query extra data associated with the state
+     * that must generally not be part of the state itself.
```
driessamyn: Just curious, why is the guidance that it should not be part of the state?
May it not be acceptable to extract something from the state into metadata if we know it must be query-able, for example?

libs/state-manager/state-manager-impl/build.gradle
```diff
@@ -0,0 +1,42 @@
+plugins {
+    id 'corda.common-publishing'
+    id 'corda.common-library'
+}
+description 'State Manager Implementation'
```
driessamyn: This is one example where I think it is justified to have an separate module for the implementation (not sure why we do that everywhere else and collect 250+ modules on the way), but shouldn't we indicate in this module that this is the postgres implementation? I.e if this is `state-manager-postgres`, we may one day have a `state-manager-redis` and/or `state-manager-coherence`, for example.

libs/state-manager/state-manager-impl/build.gradle
```diff
@@ -0,0 +1,42 @@
...
+    integrationTestImplementation project(':testing:db-testkit')
+    integrationTestImplementation project(':testing:test-utilities')
+
+    integrationTestRuntimeOnly "org.hsqldb:hsqldb:$hsqldbVersion"
+    integrationTestRuntimeOnly "org.postgresql:postgresql:$postgresDriverVersion"
+}
+
+tasks.named('jar', Jar) {
+    bnd '''\
+DynamicImport-Package: org.hibernate.proxy
```
driessamyn: Do we need to use Hibernate for this?

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -0,0 +1,137 @@
...
+import net.corda.libs.statemanager.api.StateManager
+import net.corda.libs.statemanager.impl.model.v1.StateEntity
+import net.corda.libs.statemanager.impl.repository.StateRepository
+import net.corda.orm.utils.transaction
+import org.slf4j.LoggerFactory
+import java.time.Instant
+import java.util.concurrent.ConcurrentHashMap
+import javax.persistence.EntityManagerFactory
+
+@Suppress("ForbiddenComment")
```
driessamyn: Can we fix the comment rather than suppress the warning?

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -0,0 +1,137 @@
...
+class StateManagerImpl(
```
driessamyn: Similar to the comment in the API repo. Would it be worth having a group/namespace/partition argument so that we can have multiple logical state managers that can be partitions separately?

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/State.kt
```diff
@@ -0,0 +1,34 @@
...
+import java.time.Instant
+
+/**
+ * A state managed via the state manager.
+ */
+data class State<S>(
+    /**
+     * The typed state itself.
+     */
+    val state: S,
```
driessamyn: maybe `value` would be a better? (to avoid state.state)

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -0,0 +1,137 @@
...
+    private val stateRepository: StateRepository,
+    private val entityManagerFactory: EntityManagerFactory,
+    private val serializerFactory: CordaAvroSerializationFactory,
+    private val serializers: MutableMap<Class<*>, CordaAvroSerializer<*>> = ConcurrentHashMap(),
+    private val deserializers: MutableMap<Class<*>, CordaAvroDeserializer<*>> = ConcurrentHashMap(),
```
driessamyn: I'm in 2 minds about this, so not saying it's wrong. But is it right that the state manager is also responsible for serialisation (and hence has an opaque/implicit requirement that it can only store objects that have avro serialisers registered)?
It feels a bit wrong and may it makes more sense for value to always be a binary blob so that the consumer of the state manager can be responsible for serialisation.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/model/v1/StateEntity.kt
```diff
@@ -0,0 +1,94 @@
...
+const val STATE_ID = "state"
+const val VERSION_ID = "version"
+const val METADATA_ID = "metadata"
+const val START_TIMESTAMP = "startTime"
+const val FINISH_TIMESTAMP = "finishTime"
+
+// TODO-[CORE-16663]: make the database provider pluggable.
+// Hibernate 5 does not support inserting a String to a jsonb column type out of the box. Native query with casting is also used
+// in the ledger. It means, however, this does not work with HSQLDB.
+@NamedNativeQuery(
```
driessamyn: As above, I'm really not convinced we should pay the overhead of Hibernate in this case, especially if we use native queries.
Queries will generally be simple and performance and tunability is very important in this case.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/Metadata.kt
jujoramos: 👍

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/State.kt
jujoramos: Bad working on my part, thanks 👍.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/State.kt
jujoramos: 👍

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/State.kt
jujoramos: I meant "wording", not "working" 😅

libs/state-manager/state-manager-impl/build.gradle
jujoramos: > not sure why we do that everywhere else and collect 250+ modules on the way 

💯 


> shouldn't we indicate in this module that this is the postgres implementation? I.e if this is `state-manager-postgres`, we may one day have a `state-manager-redis` and/or `state-manager-coherence`, for example.

I'd go for `state-manager-db-impl` or something similar, as to indicate that the persistent storage is a database system. I don't know if we want to include `postgres`, though, we might want to support other databases in the short term future as the rest of the platform does. I doubt customers will use database system `X` for the Corda platform and database system `Y` for just the state manager.

libs/state-manager/state-manager-impl/build.gradle
jujoramos: Similar to my previous comment, Hibernate is included just as an abstraction layer between the state manager and the chosen RDBMS.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
jujoramos: The forbidden comments are all `TODO`s, actually 😅 . Guess we can try to provide a quick implementation of the missing methods instead of suppressing the warning, the idea was to have the `TODO`s for a short period of time only (not more than 1 or 2 weeks).

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
jujoramos: Yeah, I'm also sitting on the fence for this one... I'll remove the serailizers for the time being, we can always put them back if needed.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/model/v1/StateEntity.kt
jujoramos: I agree but, as before, defaulting to a single database backend won't allow customers to customise this part as they'll be able to do for the rest of the platform; not sure that's what we want.

libs/state-manager/state-manager-impl/build.gradle
driessamyn: That's fair. Happy with `db-impl`. 👍

libs/state-manager/state-manager-impl/build.gradle
driessamyn: But what's the point, especially if we are using native queries?
It brings a fair amount of overhead (EMFs are expensive for example) and less flexibility for very little benefit. In this case, I suspect even the SQL queries will be the same for each DB type unless we want to explicitly tune them for a particular DB, in which case we'd loose the  benefit of Hibernate anyway.
Longer term, I think we should discuss removing Hibernate anywhere we don't have user schemas (where we have a good case for JPA) as we don't really use it as an ORM, just as a SQL abstraction layer, and it's very expensive for the value in that case.
In think for this library, it would be easy to not have Hibernate as it's quite separate from the rest of the persistence management.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
driessamyn: I agree that that's a questionable rule, but I always worry when we have temporary suppression as it's easy to forget about them.
If we don't agree with the rule, we should change the rule rather than suppress it everywhere (maybe a topic to bring to the weekly meeting).

BTW. the rule actually triggers `TODO:` but not `TODO -` for example. 🤷

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/model/v1/StateEntity.kt
driessamyn: We don't allow customers to customise this anyway. We only support a given set of DBs (currently 1, postgres). 
Work is underway to support an additional, but I believe that hibernate will in this instance give is very little over standard JDBC. We will need a facility to customise queries per DB dialect anyway because we use native queries elsewhere that need to vary between DB dialects.

libs/state-manager/state-manager-impl/build.gradle
jujoramos: Due to the existing tooling around Hibernate within the Corda platform, and in an attempt to speed up the first deliverable of the State Manager, it was agreed offline to keep using Hibernate in the short term and create tickets to entirely remove it in favour of direct JDBC access.

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
jujoramos: 👍

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/model/v1/StateEntity.kt
jujoramos: Due to the existing tooling around Hibernate within the Corda platform, and in an attempt to speed up the first deliverable of the State Manager, it was agreed offline to keep using Hibernate in the short term and create tickets to entirely remove it in favour of direct JDBC access.

libs/state-manager/state-manager-impl/src/test/kotlin/net/corda/libs/statemanager/impl/StateManagerImplTest.kt
jujoramos: Done

libs/state-manager/state-manager-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
jujoramos: It might be worth, sure, but I'm not sure how it will work at the implementation level; guess I'll need to read some about DB partitioning and check how it fits here.

driessamyn: I was more thinking of having it as a logical partition for now, which would allow us to tune things later.
I.e. we don't need to have the implementation on day 1, but if it is in the API, then it allows us to tune behind the scenes.
I think most of the potential implementations support a kind of partitioning. E.g. [Postgresql](https://www.postgresql.org/docs/current/ddl-partitioning.html), [Redis](https://redis-doc-test.readthedocs.io/en/latest/topics/partitioning/)

jujoramos: I've been talking with @Omar-awad and @vinir3 today, we are a bit tight in terms of time so we think is best to go with we have right now and add the extra functionality later. Hope that's okay with you?.

driessamyn: Sure. It was just a suggestion. 
The only downside of not doing it now is that adding it later on will require an (internal) API change.

jujoramos: Agreed 👍 . We'll get back to this before the actual release, though, so we should be fine.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/State.kt
```diff
@@ -0,0 +1,58 @@
...
+ * A state managed via the state manager.
+ */
+data class State(
+    /**
+     * Identifier for the state.
+     */
+    val key: String,
+
+    /**
+     * The typed state itself.
```
sprintengine: minor - this doc needs updated

jujoramos: 👍

conal-smith (commented):
LGTM once we merge corda-api

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -0,0 +1,132 @@
...
+
+        // Only primitive types are supported as part of the state metadata
+        val nativeType = when (value) {
+            is String -> "text"
+            is Number -> "numeric"
+            is Boolean -> "boolean"
+            else -> throw IllegalArgumentException("Unsupported Type: ${value::class.java.simpleName}")
+        }
+        val query = entityManager.createNativeQuery(
+            "SELECT * FROM ${DbSchema.STATE_MANAGER_TABLE} s WHERE (s.metadata->>'$key')::::$nativeType $comparison '$value'",
```
driessamyn: We should not have a SELECT * if using a Native Query, but list the fields required.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/model/v1/StateEntity.kt
```diff
@@ -0,0 +1,94 @@
...
+)
+
+@NamedNativeQuery(
+    name = UPDATE_STATE_QUERY_NAME,
+    query = """
+        UPDATE ${DbSchema.STATE_MANAGER_TABLE} SET
+        key = :$KEY_ID, value = :$VALUE_ID, version = version + 1, metadata = CAST(:$METADATA_ID as JSONB), modified_time = CURRENT_TIMESTAMP AT TIME ZONE 'UTC'
+        WHERE key = :$KEY_ID
+    """
+)
```
driessamyn, line 45: I'm not sure how nicely this manual version increment is going to play with Hibernate's optimistic locking.
Maybe not a problem for now, but keep an eye on it, and another reason to maybe move away from hibernate (or completely adopt it).

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/State.kt
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    override fun hashCode(): Int {
+        var result = key.hashCode()
+        result = 31 * result + value.contentHashCode()
+        result = 31 * result + version
+        result = 31 * result + metadata.hashCode()
+        result = 31 * result + modifiedTime.hashCode()
+
+        return result
```
driessamyn, line 62: Why do these need to be overriden given it is a data class?

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/model/v1/StateEntity.kt
jujoramos: We'll keep an eye (there's already a ticket to thoroughly test and improve the optimistic locking part).

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
jujoramos: 👍

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/State.kt
jujoramos: Due to the `ByteArray` property, according to the IDE, it's recommended to override `hashCode()` and `equals()` in this case, even for data classes (I haven't checked the actual reasons for the recommendation, though).

driessamyn: I'm guessing this is the reason: https://blog.jetbrains.com/kotlin/2015/09/feedback-request-limitations-on-data-classes/

Worth understanding and adding a comment as otherwise somebody may just delete them.

gradle.properties
```diff
@@ -46,7 +46,7 @@ commonsTextVersion = 1.10.0
 bouncycastleVersion=1.73
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.0.0.xx-SNAPSHOT to pick up maven local published copy
-cordaApiVersion=5.1.0.19-beta+
+cordaApiVersion=5.1.0.19-alpha-1694441708944
```
wzur-r3: please make sure this is updated back to beta

wzur-r3 (changes requested):

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Create [states] into the underlying storage.
+     * Control is only returned to the caller once all [states] have been updated and replicas of the underlying
+     * persistent storage, if any, are synced.
+     *
+     * @param states Collection of states to be persisted.
+     * @return States that could not be created on the persistent storage, along with the actual reason for the failures.
+     */
+    fun create(states: Collection<State>): Map<String, Exception>
```
JamesHR3: Worth being explicit that this isn't transactional and is best effort (assuming that's the case) - we'll try every state in the collection regardless, and we'll store anything that succeeds even if some fail.

Probably worth the same description for the other APIs on this interface (assuming that's true).

What's the behaviour if there is already a state on the key? Does it fail or does it overwrite the old one?

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Update [states] within the underlying storage.
+     * Control is only returned to the caller once all [states] have been updated and replicas of the underlying
+     * persistent storage, if any, are synced.
+     * The operation is transactional, either all [states] are updated or none is.
+     *
+     * @param states Collection of states to be updated.
+     * @return States that could not be updated due to mismatch versions.
+     */
+    fun update(states: Collection<State>): Map<String, State>
```
JamesHR3, line 54: Interesting that this is transactional. Does that mean if any of the states have mismatched versions none of them will be updated? If that's the case I don't think the return type makes sense any more - it'll either be the input collection or empty. If it's not the case then I think the comment is a bit misleading.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Delete all [states] from the underlying storage.
+     * Control is only returned to the caller once all states have been deleted and replicas of the underlying
+     * persistent storage, if any, are synced.
+     * The operation is transactional, either all [states] are deleted or none is.
+     *
+     * @param states Collection of states to be deleted.
+     * @return States that could not be deleted due to mismatch versions.
+     */
+    fun delete(states: Collection<State>): Map<String, State>
```
JamesHR3, line 73: Same comment here as for the update case.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -0,0 +1,454 @@
...
+                it.assertThat(loadedEntity.value).isEqualTo((stateContent(i, key).toByteArray()))
+                it.assertThat(objectMapper.toMetadata(loadedEntity.metadata))
+                    .containsExactlyInAnyOrderEntriesOf(metadataContent(i, key))
+            }
+        }
+    }
+
+    @ParameterizedTest
+    @ValueSource(ints = [1, 10])
+    fun canCreateBasicStates(stateCount: Int) {
```
JamesHR3, line 135: Minor, but for test cases you can use function names wrapped in backticks to make the test case names easier to read, e.g.

```kotlin
fun `can create basic states`(stateCount: Int) {
}
```

Sometimes the output of parameterized tests is named in an unhelpful way by default, so you might want to consider changing the display name. See https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-display-names.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -0,0 +1,135 @@
...
+    private fun State.toPersistentEntity(): StateEntity =
+        StateEntity(key, value, objectMapper.writeValueAsString(metadata), version, modifiedTime)
+
+    private fun StateEntity.fromPersistentEntity() =
+        State(key, value, version, metadata.toMetadataMap(), modifiedTime)
+
+    private fun String.toMetadataMap() =
+        objectMapper.readValue(this, object : TypeReference<Metadata>() {})
+
+    @VisibleForTesting
```
JamesHR3: I'm not really a fan of `@VisibleForTesting`, it usually implies that the tests are not really testing the class contract and are verifying some detail of the implementation. 

In this case you could write tests that verify you call update/delete on the set you're expecting to be matched and return the mismatched ones. Then you're free to change how you calculate which items are which without invalidating the unit test.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    override fun update(states: Collection<State>): Map<String, State> {
+        entityManagerFactory.transaction { em ->
+            val (updatable, mismatchVersions) = checkVersionAndPrepareEntitiesForPersistence(states, em)
+            stateRepository.update(em, updatable)
+
+            return mismatchVersions.also {
+                if (it.isNotEmpty()) {
+                    logger.warn("Optimistic locking check failed for States ${it.entries.joinToString()}")
```
JamesHR3: `warn` might be a little strong for these, but `info` seems appropriate to me. Do we anticipate that a locking failure is an error the user might need to worry about? It seems to me more likely to occur if there's a race between processes and they need to retry, which doesn't seem to be an issue from a user perspective.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -0,0 +1,134 @@
...
+            .setParameter(VERSION_ID, state.version)
+            .setParameter(METADATA_ID, state.metadata)
+            .executeUpdate()
+    }
+
+    private fun findByKeys(
+        entityManager: EntityManager,
+        keys: Collection<String>
+    ): List<StateEntity> {
+        val results = keys.chunked(50) { chunkedKeys ->
```
JamesHR3: The 50 here should probably be a constant.

Is this something we may want to tune?

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -0,0 +1,134 @@
...
+            else -> throw IllegalArgumentException("Unsupported Type: ${value::class.java.simpleName}")
+        }
+        val query = entityManager.createNativeQuery(
+            "SELECT s.key, s.value, s.metadata, s.version, s.modified_time " +
+                    "FROM ${DbSchema.STATE_MANAGER_TABLE} s " +
+                    "WHERE (s.metadata->>'$key')::::$nativeType $comparison '$value'",
+            StateEntity::class.java
+        )
+
+//        Should we use this instead of casting to the specific type (::text, ::numeric, ::boolean)?
```
JamesHR3: Would be good to resolve this question and either put this code in or remove it.

gradle.properties
jujoramos: Yes, once the api changes are merged I'll revert this back.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -140,7 +140,7 @@ class StateManagerIntegrationTest {
         assertThat(stateManager.create(states)).isEmpty()
         softlyAssertPersistedStateEntities(
             (1..stateCount),
-            { _, _ -> -1 },
+            { _, _ -> 0 },
```
jujoramos: This should use `State.VERSION_INITIAL_VALUE` now, shouldn't it?

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
jujoramos: KDocs updated.

> What's the behaviour if there is already a state on the key? Does it fail or does it overwrite the old one?

It fails and returns the error to the caller as part of the result Map.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
jujoramos: KDocs updated.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
jujoramos: KDocs updated.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
jujoramos: You're right, changed the log level to `info` 👍

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
jujoramos: 👍

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
jujoramos: > For test cases you can use function names wrapped in backticks to make the test case names easier to read, e.g.

I know, but I like the old way of squashing all words together to make our livers harder :-).

> Sometimes the output of parameterized tests is named in an unhelpful way by default,

Yeah, I didn't consider it necessary for this scenario. Made the change, anyways 👍

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
jujoramos: Removed.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
jujoramos: This seemed like a magic number to me as well but forgot to remove it. Unless @conalsmith-r3 has any objections, I'd vote for removing it and getting it back in the future if needed.

wzur-r3 (approved):
LGTM for files owned by BLT

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -39,14 +39,10 @@ class StateRepositoryImpl : StateRepository {
         entityManager: EntityManager,
         keys: Collection<String>
     ): List<StateEntity> {
-        val results = keys.chunked(50) { chunkedKeys ->
-            entityManager
+        return entityManager
```
sprintengine, line 42: We need to keep an eye on this query, afaik IN clauses with large list of keys could be inefficient, hence why I went with chunked approach initially. Let's go with this but keep an eye out.

conal-smith (approved):

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
jujoramos: 👍

JamesHR3 (approved):

PR #4603, opened 2023-09-07
CORE-16992: Set to Corda managed schema if using cluster DB or DDL provided

josephzunigadaly:
* `isPlatformManagedDb` will be set to true if we are using the cluster DB or if DDL connection string provided
* Invert the if statement to reduce the number of NOT operations in use and make the method easier to understand

### DIFF
--- a/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDbFactoryImpl.kt
+++ b/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDbFactoryImpl.kt
@@ -88,30 +88,29 @@ internal class VirtualNodeDbFactoryImpl(
         ddlConfig: String?,
         dmlConfig: String?
     ): VirtualNodeDb {
-        val connectionsProvided = !dmlConfig.isNullOrBlank()
+        val usingClusterDb = dmlConfig.isNullOrBlank()
+
         val dbConnections =
-            if (connectionsProvided) {
+            if (usingClusterDb) {
+                mapOf(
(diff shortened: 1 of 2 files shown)


components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDbFactoryImpl.kt
```diff
@@ -88,30 +88,29 @@ internal class VirtualNodeDbFactoryImpl(
...
+                } else mapOf(
+                    Pair(DDL, ddlConfig?.let { createConnection(dbType, holdingIdentityShortHash, DDL, ddlConfig) }),
+                    Pair(DML, dmlConfig?.let { createConnection(dbType, holdingIdentityShortHash, DML, dmlConfig) })
                 )
             }
+
+        val ddlProvided = ddlConfig?.isNotBlank() == true
         return VirtualNodeDbImpl(
-            !connectionsProvided,
+            usingClusterDb || ddlProvided,
```
Tom-Fitzpatrick: If we pass "none" for dml and ddl uniqueness connnection strings then our dbConnections are both null and we set the managed schema to be true, which is what we want. 

However, if "none" gets passed for dml and `null` for ddl for the uniqueness strings, our dbConnections are both null, but the managed schema gets set to false here right? Is this correct behavior? 

I'm not sure if this scenario makes much sense, but I thought I'd check as we're not explicitly testing it yet

josephzunigadaly: Good point! I'll add some tests to cover that.

Tom-Fitzpatrick (approved):
LGTM

PR #4609, opened 2023-09-08
CORE-16584 - Add MGM verification of member signature on registration context

nikinagy:
Perform MGM-side signature verification of member signature on the member-provided context on receiving a registration request. We should verify the signature using the key associated with the key ID specified in the request, before persisting the request or publishing MemberInfo .

On verification failure we will discard the registration request.


components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/QueueRegistrationHandler.kt
```diff
@@ -24,14 +28,21 @@ import net.corda.virtualnode.toCorda
...
     private val clock: Clock,
     private val membershipPersistenceClient: MembershipPersistenceClient,
     cordaAvroSerializationFactory: CordaAvroSerializationFactory,
+    signatureVerificationService: SignatureVerificationService,
+    private val keyEncodingService: KeyEncodingService,
     private val p2pRecordsFactory: P2pRecordsFactory = P2pRecordsFactory(
         cordaAvroSerializationFactory,
         clock,
     ),
+    private val verifier: Verifier = Verifier(
```
williamvigorr3, line 43: Nit: `private val verifier = Verifier(` (I think).

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/QueueRegistrationHandler.kt
```diff
@@ -132,6 +158,13 @@ internal class QueueRegistrationHandler(
         )
     }
 
+    private fun KeyValuePairList.getSessionKeys() =
+        this.items.filter { keyValuePair ->
+            keyValuePair.key.startsWith(SESSION_KEYS) && keyValuePair.key.endsWith(".pem")
```
williamvigorr3: Again very much a nit: Can you use `PARTY_SESSION_KEYS_PEM` instead of hardcoding the `.pem` here. Incase we decide to change the key format.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/QueueRegistrationHandler.kt
```diff
@@ -91,6 +102,21 @@ internal class QueueRegistrationHandler(
...
+            verifier.verify(
+                context.getSessionKeys(),
+                command.memberRegistrationRequest.memberContext.signature,
+                command.memberRegistrationRequest.memberContext.signatureSpec,
+                command.memberRegistrationRequest.memberContext.data.array(),
+            )
+        } catch (e: Exception) {
+            logger.warn("Signature verification failed. Discarding the registration with ID `$registrationId`. " +
+                    "Reason: ${e.message}")
+            return emptyList()
```
williamvigorr3, line 118: Observation: I guess we made the conscious choice not to persist the `RegistrationRequest` with status failed. Incase someone sends lots of spurious requests to try and use all the disk space on the MGM's DB.

williamvigorr3 (commented):
Looks good. I left some minor comments.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/QueueRegistrationHandler.kt
nikinagy: this can't be done, discussed offline

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/QueueRegistrationHandler.kt
nikinagy: Plus, I don't think its good idea to persist a potentially malicious/modified request

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/QueueRegistrationHandler.kt
```diff
@@ -132,6 +159,13 @@ internal class QueueRegistrationHandler(
         )
     }
 
+    private fun KeyValuePairList.getSessionKeys() =
+        this.items.filter { keyValuePair ->
```
kyriathar: Maybe `keyValuePair` could be changed to reflect what it holds conceptually.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/handler/mgm/QueueRegistrationHandler.kt
```diff
@@ -132,6 +159,13 @@ internal class QueueRegistrationHandler(
...
+            keyValuePair.key.startsWith(SESSION_KEYS) && keyValuePair.key.endsWith(KEYS_PEM_SUFFIX)
+        }.map {  filtered ->
```
kyriathar: Same as above for `filtered`.

libs/crypto/crypto-impl/build.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
 
     implementation project(":libs:configuration:configuration-core")
     implementation project(':libs:layered-property-map')
+    implementation project(':libs:membership:membership-common')
```
williamvigorr3: I think we should avoid adding a dependency on `membership` to `crypto`. It feels like this could lead to circular dependencies, as crypto lower level. 

Probably you can get away with it because `membership-common` doesn't have much in it. I still don't think it's a good idea.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/converter/PublicKeyConverter.kt
```diff
@@ -25,8 +26,8 @@ class PublicKeyConverter @Activate constructor(
     /**
      * Select the single element in case the structure is like 'corda.ledger.keys.1'
      */
-    override fun convert(context: ConversionContext): PublicKey? = if (context.value("pem") != null) {
-        context.value("pem")?.let {
+    override fun convert(context: ConversionContext): PublicKey? = if (context.value(KEYS_PEM_SUFFIX) != null) {
```
williamvigorr3: Maybe add `KEYS_PEM_SUFFIX` to the companion object of `PublicKeyConverter` instead (to solve the dependencies).

kyriathar (commented):
LGTM crypto wise (2 minor comments other than crypto).

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/converter/PublicKeyConverter.kt
nikinagy: I will just revert it for now then. Since it would require extra review from crypto and does not seem like an urgent matter.

libs/crypto/crypto-impl/build.gradle
kyriathar: Good catch.

nikinagy: Reverted crypto changes

williamvigorr3 (approved):
LGTM

PR #4610, opened 2023-09-08
CORE-9975: Allow signing operation to use new key category parameters

malachyb:


LWogan (dismissed):
Looks ok to me from a flow perspective. Don't forget to update the API version to a beta release before merging.

relyafi (commented):
I'm not familiar with this code, so might be I overlooked something, but it looks to me like this is only capturing the requested key type? Where are the changes to the logic to use the parameter to filter which keys the signing logic actually uses - does this code already exist?

Also, I think we should probably have additional test cases to prove that the filtering is working. For example, asking the signing service to sign when we provide it with both a ledger and notary key, specifying one or the other types, and verifying we only got the one signature back with the correct type. The PR should also have a populated description summarizing what we've done and why.

malachyb:
Yes sorry, I added that in a previously merged change along with some testing (https://github.com/corda/corda-runtime-os/commit/980f9a5a6b28dff82e37d58384881f82727171b4)

relyafi:
> Yes sorry, I added that in a previously merged change along with some testing ([980f9a5](https://github.com/corda/corda-runtime-os/commit/980f9a5a6b28dff82e37d58384881f82727171b4))

Thanks, it looks like that PR covers the bits that I thought were missing from this PR. I would still consider adding an additional test case (which I don't think we have in that PR) to cover what I mentioned earlier:

_asking the signing service to sign when we provide it with both a ledger and notary key, specifying one or the other types, and verifying we only got the one signature back with the correct type_

relyafi (dismissed):
Change LGTM, I would recommend adding an additional test case (see earlier comment)

davidcurrie (dismissed):
LGTM

PR #4615, opened 2023-09-11
CORE-17030 - Fix up SyncRPCSubscriptionlog level when response payload is null

driessamyn:
Expected log level is warning in this case as errors are reserved for cases where the processor cannot continue.


emilybowe (approved):
LGTM

PR #4620, opened 2023-09-11
CORE-17051: Ensure pipeline context is updated with intermediate flow results

JamesHR3:
**Problem statement**

The flow engine was recently updated to execute user code through multiple suspension points when processing a single event. When it does this, a new user code execution stage keeps track of the intermediate context. However, if an error occurs during processing then the context used to handle the error will be the one from before any fiber code was executed. This means that in some circumstances the flow engine incorrectly thinks the checkpoint is yet to be initialized, and so it will not publish a status update for the flow.

**Solution**

Ensure the flow pipeline is informed of context changes as the fiber is executed. This means that if an error is thrown, the most up-to-date `FlowEventContext` will be used to handle the error.

**Testing**

Passing full e2e test run.

**Other changes**

- Updated a misleading log line about being unable to update status. This can happen if the flow is not failing.
- Updated a misleading comment. The old comment was describing behaviour that is no longer present.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowExecutionPipelineStage.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowExecutionPipelineStage.kt
@@ -43,15 +43,22 @@ internal class FlowExecutionPipelineStage(
      *
      * @param context The current flow context.
      * @param timeout Timeout for an individual flow execution.
+     * @param notifyContextUpdate A function to signal intermediate context updates to the caller.
+     *                            Useful for error scenarios.
      * @return FlowEventContext Updated context for the flow.
      */
-    fun runFlow(context: FlowEventContext<Any>, timeout: Long) : FlowEventContext<Any> {
+    fun runFlow(
+        context: FlowEventContext<Any>,
+        timeout: Long,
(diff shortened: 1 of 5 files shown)


components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowExecutionPipelineStage.kt
```diff
@@ -43,15 +43,22 @@ internal class FlowExecutionPipelineStage(
      *
      * @param context The current flow context.
      * @param timeout Timeout for an individual flow execution.
+     * @param notifyContextUpdate A function to signal intermediate context updates to the caller.
+     *                            Useful for error scenarios.
```
emilybowe, line 47: nit - is the spacing on this a bit weird? it looks tabbed, even looking at the full file for me

emilybowe (approved):
LGTM

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowExecutionPipelineStage.kt
JamesHR3: This was just to line the text up with the rest of the parameter description.

PR #4623, opened 2023-09-12
ES-1256 Expose pod and container security contexts in Helm chart

jacob-scott3141:
This enables the security contexts to be unset so that they can be automatically populated in an OpenShift environment.

### DIFF
--- a/charts/corda-lib/templates/_helpers.tpl
+++ b/charts/corda-lib/templates/_helpers.tpl
@@ -74,7 +74,7 @@ Container security context
 {{- if not .Values.dumpHostPath }}
 {{- with .Values.containerSecurityContext }}
 securityContext:
-  {{ . | toYaml }}
+  {{ . | toYaml | nindent 2}}
 {{- end }}
 {{- end }}
 {{- end }}
--- a/charts/corda-lib/templates/_helpers.tpl
+++ b/charts/corda-lib/templates/_helpers.tpl
@@ -72,14 +72,10 @@ Container security context
 */}}
 {{- define "corda.containerSecurityContext" -}}
 {{- if not .Values.dumpHostPath }}
+{{- with .Values.containerSecurityContext }}
 securityContext:
-  runAsUser: 10001
-  runAsGroup: 10002
-  allowPrivilegeEscalation: false
-  readOnlyRootFilesystem: true
-  capabilities:
-    drop:
-      - "ALL"
+  {{ . | toYaml }}
+{{- end }}
 {{- end }}
 {{- end }}
 
--- a/charts/corda-lib/templates/_worker.tpl
+++ b/charts/corda-lib/templates/_worker.tpl
@@ -106,7 +106,7 @@ spec:
       {{- if and ( not $.Values.dumpHostPath ) ( not .profiling.enabled ) }}
       {{- with $.Values.podSecurityContext }}
       securityContext:
-        {{ . | toYaml }}
+        {{ . | toYaml | nindent 8 }}
       {{- end }}
       {{- end }}
       {{- include "corda.imagePullSecrets" $ | nindent 6 }}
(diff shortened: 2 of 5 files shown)


charts/corda-lib/templates/_worker.tpl
```diff
@@ -104,10 +104,10 @@ spec:
         {{- include "corda.workerSelectorLabels" ( list $ $worker ) | nindent 8 }}
     spec:
       {{- if and ( not $.Values.dumpHostPath ) ( not .profiling.enabled ) }}
+      {{- with .Values.podSecurityContext }}
```
davidcurrie: I think you need to use `$.Values` here (which is why the build is failing)

charts/corda/values.yaml
```diff
@@ -20,6 +20,31 @@ image:
...
+containerSecurityContext:
+  # -- specify what user ID that processes will run with
+  runAsUser: 10001
+  # -- specify what group ID that processes will run with
+  runAsGroup: 10002
+  # -- enable privilege escalation
+  allowPrivilegeEscalation: false
+  # -- mount the container's root filesystem as read-only
+  readOnlyRootFilesystem: true
+  # -- linux capabilities for users
```
davidcurrie: only leaf doc comments should have `--`

charts/corda/values.schema.json
```diff
@@ -121,6 +121,128 @@
                 "Always"
             ]
         },
+        "containerSecurityContext": {
```
davidcurrie, line 124: I think there are some extra values here that we don't currently use but should still allow the user to set. Take a look at the schema here, for example:
https://github.com/instrumenta/kubernetes-json-schema/blob/master/v1.7.4/securitycontext.json
https://github.com/instrumenta/kubernetes-json-schema/blob/master/v1.7.4/podsecuritycontext.json

charts/corda/values.schema.json
```diff
@@ -121,6 +121,128 @@
...
+                        "add": {
+                            "type": "array",
+                            "default": [],
+                            "title": "capabilities to add"
+                        }
+                    }
+                }
+            }
+        },
+        "podSecurityContext": {
```
davidcurrie, line 271: Not all of these fields exist for the pod security context (e.g. `capabilities` can only be specified at the container level). Take a look at the schema referenced above.

charts/corda/values.schema.json
```diff
@@ -121,6 +121,128 @@
...
+                        true
+                    ]
+                },
+                "capabilities": {
+                    "type": "object",
+                    "default": {},
+                    "title": "linux capabilities for users",
+                    "properties": {
+                        "drop": {
+                            "type": "array",
```
davidcurrie: We should be say what the type of the items in the array are e.g. https://github.com/instrumenta/kubernetes-json-schema/blob/master/v1.7.4/capabilities.json#L7-L12

davidcurrie (changes requested):

charts/corda/values.schema.json
jacob-scott3141: Do you know what version we're using? The schema for podsecuritycontext doesn't have a runAsGroup in this version

davidcurrie: So, we currently say `1.23`. Unfortunately that site stops at 1.18 but that at least gets you a bit closer:
https://github.com/instrumenta/kubernetes-json-schema/blob/master/v1.18.1/podsecuritycontext.json

The definitive source is probably the API docs: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#security-context

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -72,14 +72,10 @@ Container security context
...
+{{- with .Values.containerSecurityContext }}
 securityContext:
-  runAsUser: 10001
-  runAsGroup: 10002
-  allowPrivilegeEscalation: false
-  readOnlyRootFilesystem: true
-  capabilities:
-    drop:
-      - "ALL"
+  {{ . | toYaml }}
```
davidcurrie: need another `nindent` here

davidcurrie (commented):
LGTM - can you verify that, as per the ticket, you can set the two values to `nil` and then nothing gets included in the deployments? Then, as discussed, we should make the `corda` changes to `corda-enterprise` first and then we can merge this PR with the `corda-lib` updates in.

davidcurrie (approved):
LGTM

PR #4624, opened 2023-09-12
CORE-17154 - boot the flow and combined workers with the state manager

conal-smith:
This change follows on from [4602](https://github.com/corda/corda-runtime-os/pull/4602) to enable deployment of workers with state manager configuration in boot config.

API change - https://github.com/corda/corda-api/pull/1242


applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
```diff
@@ -121,21 +124,24 @@ class CombinedWorker @Activate constructor(
...
-        val cryptoConfig = PathAndConfig(BootConfig.BOOT_CRYPTO, createCryptoBootstrapParamsMap(params.hsmId))
-        val restConfig = PathAndConfig(BootConfig.BOOT_REST, params.restParams)
-        val config = getBootstrapConfig(
+
+        var config = getBootstrapConfig(
             secretsServiceFactoryResolver,
             params.defaultParams,
             configurationValidatorFactory.createConfigValidator(),
-            listOf(databaseConfig, cryptoConfig, restConfig),
+            listOf(
```
sprintengine, line 155: This change addresses comment here - https://github.com/corda/corda-runtime-os/pull/4501#discussion_r1317060119

After this change, running combined worker, I can see the configs in the ConfigProcessor cache:

flow:
```
{"event":{"maxRetries":5,"messageResendWindow":60000},"processing":{"cleanupTime":600000,"maxFlowSleepDuration":900000,"maxRetryAttempts":1,"maxRetryDelay":16000,"maxRetryWindowDuration":1},"session":{"cleanupTime":600000,"p2pTTL":300000,"timeout":1800000}}
```

messaging:
```
{"bus":{"busType":"DATABASE","dbProperties":{"jdbcUrl":"jdbc:postgresql://localhost:5432/cordacluster?currentSchema=MESSAGEBUS","pass":"password","user":"user"}},"instanceId":"0","maxAllowedMessageSize":972800,"publisher":{"closeTimeout":600,"transactional":true},"stateManager":{"database":{"jdbc":{"url":"jdbc:postgresql://localhost:5432/cordacluster"},"pass":"password","pool":{"idleTimeoutSeconds":"120","keepAliveTimeSeconds":"0","maxLifetimeSeconds":"1800","maxSize":"5","minSize":"1","validationTimeoutSeconds":"5"},"user":"user"},"type":"DATABASE"},"subscription":{"commitRetries":3,"pollTimeout":500,"processorRetries":3,"processorTimeout":15000,"subscribeRetries":3,"threadStopTimeout":10000},"topicPrefix":""}
```

corda-db
```
{"bus":{"busType":"DATABASE"},"database":{"jdbc":{"directory":"C:\\workspace/corda-runtime-os/applications/workers/release/combined-worker/drivers","url":"jdbc:postgresql://localhost:5432/cordacluster?currentSchema=CONFIG","url_messagebus":"jdbc:postgresql://localhost:5432/cordacluster?currentSchema=MESSAGEBUS"},"pass":"password","user":"user"}}
```

crypto
```
{"caching":{"expireAfterAccessMins":{"default":60},"maximumSize":{"default":10000}},"hsm":{"defaultWrappingKey":"root1","retrying":{"attemptTimeoutMills":20000,"maxAttempts":3},"wrappingKeys":[{"algorithm":"PBKDF2WithHmacSHA256","alias":"root1","passphrase":"*****","salt":"*****"}]},"retrying":{"maxAttempts":{"default":3},"waitBetweenMills":{"default":[200]}}}
```

applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/DefaultWorkerParams.kt
```diff
@@ -70,4 +70,10 @@ class DefaultWorkerParams(healthPortOverride: Int = WORKER_MONITOR_PORT) {
     @Option(names = ["--trace-samples-per-second"], description = ["Number of request traces to sample per second, " +
             "defaults to 1 sample per second. Set to \"unlimited\" to record all samples"])
     var traceSamplesPerSecond: String? = null
+
+    @Option(
+        names = ["-S", "--${BootConfig.BOOT_STATE_MANAGER}"],
+        description = ["Configuration for the state manager."]
+    )
+    var stateManagerParams = emptyMap<String, String>()
```
sprintengine, line 78: Addresses comment here - https://github.com/corda/corda-runtime-os/pull/4501#discussion_r1317061411

In a recent change I put state manager params into the default worker params. I noticed that this means we can handle this by default rather than supplying state manager params in as "extra params" now. I'll update this.

applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/WorkerHelpers.kt
```diff
@@ -74,13 +118,8 @@ class WorkerHelpers {
             secretsServiceFactoryResolver: SecretsServiceFactoryResolver,
             defaultParams: DefaultWorkerParams,
             validator: ConfigurationValidator,
-            extraParams: List<PathAndConfig> = emptyList(),
+            extraConfigs: List<Config> = emptyList(),
         ): SmartConfig {
-            val extraParamsMap = extraParams
```
sprintengine, line 79: we now pass in built configs instead of "extraParams" so we no longer need to perform this step key mapping step.

The list of extra config is merged on top of the config constructed in this function, i.e. the builtConfig is fallback config upon which the extra configs are applied.

This refactoring was done before I moved stateManager config into default worker params and handle it differently. I could pull this refactoring out of this PR.

libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/configuration/DbBusConfigMergerImpl.kt
```diff
@@ -1,33 +1,44 @@
...
 import org.osgi.service.component.annotations.Component
 
 @Component(service = [BusConfigMerger::class])
 class DbBusConfigMergerImpl : BusConfigMerger {
 
     override fun getMessagingConfig(bootConfig: SmartConfig, messagingConfig: SmartConfig?): SmartConfig {
-        val updatedMessagingConfig = messagingConfig?: SmartConfigImpl.empty()
+        var updatedMessagingConfig = messagingConfig ?: SmartConfigImpl.empty()
+
+        bootConfig.getConfigOrEmpty(BootConfig.BOOT_STATE_MANAGER).entrySet().forEach { entry ->
```
sprintengine, line 24: I'll extract this out to its own interface and implementation supplied by statemanager-db-impl.

EDIT - I think we should leave this as is for now. I tried a couple of approaches where the state-manager implementation provides the function to merge the config via an OSGI component, but these approaches aren't correct as they introduce a runtime dependency on state manager for every worker to boot which is not what we want. 

Some more thought could be put in to come up with alternative solution. If we swap the state-manager implementation, the new implementation should provide the way to merge the configs. 

EDIT 2 - Actually, in reality all this "implementation" is really doing is taking the config from the command line arguments, and prepending with a "stateManager." prefix. Maybe all we need here is a helper function that takes a "bootKey" and a "newKey" that takes out config from bootConfig and puts into the main config with the newKey...

libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/configuration/StateManagerConfigMerger.kt
```diff
@@ -0,0 +1,16 @@
...
+
+    /**
+     * Return a new [SmartConfig] with the state manager configuration from the [bootConfig] merged with any state manager config from
+     * the existing [messagingConfig].
+     *
+     * @param bootConfig configuration object containing boot config
+     * @param messagingConfig existing configuration object containing messaging config, or null
+     * @return SmartConfig object containing the state manager configuration.
+     */
+    fun getStateManagerConfig(bootConfig: SmartConfig, messagingConfig: SmartConfig?) : SmartConfig
```
sprintengine: introduction of this interface allows us to address this comment - https://github.com/corda/corda-runtime-os/pull/4501#discussion_r1317092310

EDIT - this change implies a runtime dependency on a state manager implementation in all workers, so this is not correct approach

wzur-r3 (dismissed):
LGTM for Gradle files

libs/state-manager/state-manager-db-impl/build.gradle
```diff
@@ -22,6 +22,7 @@ dependencies {
     implementation project(':libs:db:db-orm')
     implementation project(':libs:db:db-core')
     implementation project(":libs:configuration:configuration-core")
+    implementation project(":libs:messaging:message-bus")
```
sprintengine: This dependency is so we can implement the `StateManagerConfigMerger`.

Tbh this dependency doesn't feel 100% correct to me, maybe an alternative is to put the interface into configuration-merger.

.run/Combined Worker Local.run.xml
```diff
@@ -2,7 +2,8 @@
...
+    <!-- TODO-[CORE-16419]: isolate StateManager database from the Cluster database -->
+    <option name="PROGRAM_PARAMETERS" value="--instance-id=0 -mbus.busType=DATABASE -spassphrase=password -ssalt=salt -spassphrase=password -ssalt=salt -ddatabase.user=user -ddatabase.pass=password -ddatabase.jdbc.url=jdbc:postgresql://localhost:5432/cordacluster -ddatabase.jdbc.directory=$ProjectFileDir$/applications/workers/release/combined-worker/drivers -Stype=DATABASE -Sdatabase.user=user -Sdatabase.pass=password -Sdatabase.jdbc.url=jdbc:postgresql://localhost:5432/cordacluster -rtls.crt.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/server.crt -rtls.key.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/server.key -rtls.ca.crt.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/ca-chain-bundle.crt" />
```
driessamyn: Is this using `-S` for state manager config? 
I think that's super confusing given `-s` is for the secrets config.
Can we not pick another shortcode?

tools/plugins/db-config/src/main/kotlin/net/corda/cli/plugins/dbconfig/Spec.kt
```diff
@@ -99,7 +99,8 @@ class Spec(private val config: SpecConfig = SpecConfig()) : Runnable {
             "net/corda/db/schema/config/db.changelog-master.xml",
             "net/corda/db/schema/messagebus/db.changelog-master.xml",
             "net/corda/db/schema/rbac/db.changelog-master.xml",
-            "net/corda/db/schema/crypto/db.changelog-master.xml"
+            "net/corda/db/schema/crypto/db.changelog-master.xml",
```
driessamyn, line 102: Not directly related to this PR, but there is an outstanding issue that this is not supporting all the DBs being deployed to their own instance. 
This is even more important for the state manager, which I think should NEVER be deployed into the DB that also hosts the vnode DBs, for example.
And our default deployment scripts must ensure that the users don't overlap as that would weaken our sandbox story.

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
```diff
@@ -146,6 +151,17 @@ class CombinedWorker @Activate constructor(
         val dbAdminPassword = if (config.getConfig(BootConfig.BOOT_DB).hasPath(DatabaseConfig.DB_PASS))
             config.getConfig(BootConfig.BOOT_DB).getString(DatabaseConfig.DB_PASS) else "password"
 
+        // Default pool settings for State Manager
```
sprintengine: I thought we could rely on the schema `state-manager-db-properties` (or corda.boot.json) to have these defaults set, this doesn't appear to be happening when I remove these pool settings.

EDIT - it's because the when we create the bootConfig "stateManager.database" object the "pool" object is null and our "ApplyDefaultsStrategy" in ConfigurationValidatorImpl only applies defaults to attributes that are missing, it doesn't seem to add a new object where it is missing.

.run/Combined Worker Local.run.xml
sprintengine: In terms of short options, the following letters are unavailable:

h V i t M n p m s a f d 

That leaves b c e f g h i j k l o r q u w x y z.

I'm not sure which of these sounds more appropriate, unless we just get rid of the single character short option?

tools/plugins/db-config/src/main/kotlin/net/corda/cli/plugins/dbconfig/Spec.kt
sprintengine: We have https://r3-cev.atlassian.net/browse/CORE-16419 to deploy an isolated db for state manager. For now we are deploying the state manager into the cluster db as a temporary solution to speed up development. I'll add a comment to that task about not sharing the same users as those used to access cluster db

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/config/KafkaConfigMergerImpl.kt
```diff
@@ -4,20 +4,24 @@ import com.typesafe.config.ConfigValueFactory
...
 import org.osgi.service.component.annotations.Component
+import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
 @Component(service = [BusConfigMerger::class])
 class KafkaConfigMergerImpl : BusConfigMerger {
 
     private companion object {
-        val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+        val logger: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
```
sprintengine: I think this was changed due to compiler warning without it... making this field private also removes the compiler warning, even though the companion object is already private.

.run/Combined Worker Local.run.xml
JamesHR3: We're adding a bunch of mandatory extra config to start the combined worker. This might make it harder for developers to work with it, especially for those not starting it from IntelliJ.

We may separately want to consider our story for starting the combined worker up.

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
JamesHR3: If we can get rid of this then great, but if not I'd prefer it to be pulled into its own function that returns a `SmartConfig`, and allow that to be passed into `getBootstrapConfig`. That means the `config` variable in this function no longer needs to be `var`.

applications/workers/release/flow-worker/build.gradle
```diff
@@ -50,4 +50,10 @@ dependencies {
         exclude group: 'org.apache.felix'
         exclude group: 'org.osgi'
     }
+
+    // This puts the jdbc driver into the docker image in the /opt/jdbc-driver folder
+    // this folder can contain many jdbc drivers (and DataSourceFactory provider bundles).
+    // Postgres doesn't need a DataSourceFactory provider bundle (e.g. pax-jdbc), because
+    // the postgres devs have written their own and it's in this jar (PGDataSourceFactory).
+    dockerImageJdbc "org.postgresql:postgresql:$postgresDriverVersion"
```
JamesHR3: Is this because the flow worker is the only place we've integrated the state manager? Do we need to do this yet?

.run/Combined Worker Local.run.xml
driessamyn: Y. Maybe we stick with the long option?

driessamyn: For the combined worker, couldn't we just default to a HSQL (in-memory) database?

sprintengine: I'm not sure HSQL works very well with the JSONB type and JSON queries. For combined worker I think I'd prefer to default the stateManager config to the cluster db parameters if no stateManager config is provided, how does that sound?

e.g. something like this:

```
    // For combined worker, stateManager config defaults to the database parameters if not provided
    private fun createOrDeriveStateManagerConfig(stateManagerParams: Map<String, String>, dbConfig: Config): Config {
        return if (stateManagerParams.isEmpty()) {
            ConfigFactory.empty()
                .withValue(BOOT_STATE_MANAGER_TYPE, ConfigValueFactory.fromAnyRef("DATABASE"))
                .withValue(BOOT_STATE_MANAGER_JDBC_URL, ConfigValueFactory.fromAnyRef(dbConfig.getString(DatabaseConfig.JDBC_URL)))
                .withValue(BOOT_STATE_MANAGER_DB_USER, ConfigValueFactory.fromAnyRef(dbConfig.getString(DatabaseConfig.DB_USER)))
                .withValue(BOOT_STATE_MANAGER_DB_PASS, ConfigValueFactory.fromAnyRef(dbConfig.getString(DatabaseConfig.DB_PASS)))
        } else {
            createConfigFromParams(BootConfig.BOOT_STATE_MANAGER, stateManagerParams)
        }
    }
```

driessamyn: I don't think that's a great argument.
Ultimately everything needs to work with HSQL because it's our default local test target. We'll run into the same issue with integration tests.

sprintengine: Yeah unfortunately we skip our integration tests when in-mem db is used - using `DbUtils.assumeFalse`. Afaik the build machine runs with postgres, and developers can run the tests by setting the [postgres properties](https://github.com/corda/corda-runtime-os/blob/3e407e7716881a245f24b8d2fdd07f65d2e038da/gradle.properties#L153-L156). 

It's obviously not ideal, that's why I created https://r3-cev.atlassian.net/browse/CORE-16663 and suggested to either abstract the repository layer to a db-specific implementation or else rename this implementation "state-manager-postgres". Although I'm not sure this is the correct approach either - just swapping to a different implementation for local tests. With our current JPA named query approach I wasn't able to get HSQL to work with JSONB, maybe we can find a solution that uses stored procs and selects the query based on database type, though I haven't done that before.

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
sprintengine: Yeah unfortunately we can't supply this into getBootstrapConfig as an "extraConfig" the way it stands because these fields aren't defined in the `corda.boot.json` schema, that function validates the config against that schema. 

For db worker and database params (which also has pool properties which aren't supplied in the boot config), it gets the defaults from the corda.db.json schema in `DataSourceFactoryHelper` and applies them at the point in time when it creates the [datasourceFactory](https://github.com/corda/corda-runtime-os/blob/eaee01430026c6d7ab77f547a4cbdc5445e93a16/components/db/db-connection-manager-impl/src/main/kotlin/net/corda/db/connection/manager/impl/DataSourceFactoryHelper.kt#L29). Note these defaults are not set in the config merger.

If we do something similar to that, we can apply the defaults from the schema in the StateManagerFactoryImpl. Although it's not entirely clear why we shouldn't apply defaults at the config merger.. maybe to do with avoiding having these defaults in the [config cache](https://github.com/corda/corda-runtime-os/blob/ab61bfc87ea2ad0b59e0cd1d7cb0937dc9632acb/components/configuration/configuration-read-service-impl/src/main/kotlin/net/corda/configuration/read/impl/ConfigProcessor.kt#L33).

I'll try this tomorrow.

applications/workers/release/flow-worker/build.gradle
sprintengine: Yes, we can implement this as part of the [flow worker integration task](https://r3-cev.atlassian.net/browse/CORE-16320). I'll link it.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManagerFactory.kt
```diff
@@ -8,10 +8,10 @@ import net.corda.libs.configuration.SmartConfig
...
-     * Create a state manager from the given [config].
+     * Create a state manager from the given [messagingConfig].
      *
-     * @param config for the state manager to connect to underlying storage mechanism.
-     * @return a state manager created from the given [config].
+     * @param messagingConfig containing the state manager to connect to underlying storage mechanism.
+     * @return a state manager created from the given [messagingConfig].
      */
-    fun create(config: SmartConfig): StateManager
+    fun create(messagingConfig: SmartConfig): StateManager
```
sprintengine, line 16: I'm not really that happy having to provide messagingConfig to this function. It's because all our keys (`MessagingConfig.StateManager.JDBC_*`) start from messagingConfig as the root (they are "stateManager.database.jdbc...").

Alternative solution that allows caller of the factory to provide stateManager config is:
- in the StateManagerConfigProvider: `return stateManagerConfig.withFallback(defaultPoolConfig.getConfig(MessagingConfig.StateManager.STATE_MANAGER))`
- in the stateManagerFactory create a new config that wraps the main config with a "stateManager" e.g.

```
val conf = ConfigFactory.empty()
            .withValue(MessagingConfig.StateManager.STATE_MANAGER, fromAnyRef(smConfigWithFallback))
```
But this feels a bit hacky, if it's even possible..

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
```diff
@@ -195,6 +214,47 @@ class CombinedWorker @Activate constructor(
...
+    /**
+     * Combined worker parameter for state manager's JDBC URL should be the schemaless database URL because the combined worker sets up
+     * schemas itself. However, Corda processors all expect the JDBC URL in the config to point to the config schema
+     * directly, so the name of that schema must be added to the params that are used to create the config.
+     */
+    private fun prepareStateManagerConfig(stateManagerConfig: Config): Config {
+        // add the state manager schema to the JDBC URL.
+        return stateManagerConfig.withValue(
+            BOOT_STATE_MANAGER_JDBC_URL,
+            fromAnyRef("${stateManagerConfig.getString(BOOT_STATE_MANAGER_JDBC_URL)}?currentSchema=$STATE_MANAGER_SCHEMA_NAME")
```
sprintengine, line 241: I plugged a dummy state manager into flow engine and needed the schema to get it to work.

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
```diff
@@ -195,6 +214,47 @@ class CombinedWorker @Activate constructor(
...
+        return if (stateManagerParams.isEmpty()) {
+            ConfigFactory.empty()
+                .withValue(BOOT_STATE_MANAGER_TYPE, fromAnyRef(DEFAULT_BOOT_STATE_MANAGER_TYPE))
+                .withValue(BOOT_STATE_MANAGER_JDBC_URL, fromAnyRef(dbConfig.getString(BOOT_JDBC_URL)))
+                .withValue(BOOT_STATE_MANAGER_DB_USER, fromAnyRef(dbConfig.getString(BOOT_JDBC_USER)))
+                .withValue(BOOT_STATE_MANAGER_DB_PASS, fromAnyRef(dbConfig.getString(BOOT_JDBC_PASS)))
+        } else {
+            createConfigFromParams(BootConfig.BOOT_STATE_MANAGER, stateManagerParams)
+        }
+    }
```
sprintengine, line 259: when no state manager cmd options are provided, derive the state manager config from the db params for now.

charts/corda-lib/templates/_worker.tpl
```diff
@@ -276,6 +277,22 @@ spec:
           - "-ddatabase.pool.keepaliveTimeSeconds={{ .clusterDbConnectionPool.keepaliveTimeSeconds }}"
           - "-ddatabase.pool.validationTimeoutSeconds={{ .clusterDbConnectionPool.validationTimeoutSeconds }}"
           {{- end }}
+          {{- /* TODO-[CORE-16419]: isolate StateManager database from the Cluster database */ -}}
+          {{- if $optionalArgs.stateManagerDbAccess }}
+          - "--stateManager type=DATABASE"
```
sprintengine: These spaces could be causing an issue with the current [cluster deployment failure](https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda-runtime-os-e2e-tests/detail/PR-4624/11/pipeline/).

Also, we've removed the postgres driver from the flow worker in a recent commit, it could also be causing an issue.

davidcurrie: As per Slack, I suspect you need to split these over two lines e.g.
```
          - "--stateManager"
          - "database.user=$(DB_CLUSTER_USERNAME)"
```

charts/corda/values.schema.json
```diff
@@ -1258,6 +1258,106 @@
                                 }
                             }]
                         },
+                        "stateManager": {
+                            "type": "object",
+                            "default": {},
+                            "title": "State-manager db configuration",
```
davidcurrie: nit: inconsistent use of "state-manager" vs "state manager"

charts/corda/values.yaml
```diff
@@ -363,6 +363,50 @@ bootstrap:
...
+        idleTimeoutSeconds: 120
+        # -- maximum time (in seconds) a connection can stay in the pool, regardless if it has been idle or has been recently used; If a connection is in-use and has reached "maxLifetime" timeout, it will be removed from the pool only when it becomes idle
+        maxLifetimeSeconds: 1800
+        # -- interval time (in seconds) in which connections will be tested for aliveness; Connections which are no longer alive are removed from the pool; A value of 0 means this check is disabled
+        keepaliveTimeSeconds: 0
+        # -- maximum time (in seconds) that the pool will wait for a connection to be validated as alive
+        validationTimeoutSeconds: 5
+      # the username configuration
+      username:
+        # -- the username, defaults to a value randomly-generated on install
```
davidcurrie: I know this is copied from above but my suspicion is that all of the comments are lying! The default username is the one given here.

davidcurrie (commented):
Helm chart changes LGTM, barring a few nits and resolving the issue with the argument passing.

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
```diff
@@ -118,24 +132,29 @@ class CombinedWorker @Activate constructor(
...
-        // Add the config schema to the JDBC URL in the params so that any processors which need the JDBC URL are using
-        // the config schema.
-        params.addSchemaToJdbcUrl("CONFIG")
-        params.addDatabaseParam(DatabaseConfig.JDBC_URL + "_messagebus", dbUrl + "?currentSchema=MESSAGEBUS")
+
+        val dbConfig = createConfigFromParams(BootConfig.BOOT_DB, params.databaseParams)
+        val stateManagerConfig = createOrDeriveStateManagerConfig(params.defaultParams.stateManagerParams, dbConfig)
+
+        val preparedDbConfig = prepareDbConfig(dbConfig)
+        val preparedStateManagerConfig = prepareStateManagerConfig(stateManagerConfig)
```
jujoramos, line 143: Considering that the combined worker is supposed to be used for development purposes only, can we get away with just using the cluster database for the state manager _**always**_ instead of allowing it to be parametrisable?. That would simplify the `PR` and the implementation itself, and also will prevent adding new startup parameters to the combined worker.

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
```diff
@@ -195,6 +214,47 @@ class CombinedWorker @Activate constructor(
...
+    /**
+     * Combined worker parameter for JDBC URL should be the schemaless database URL because the combined worker sets up
+     * schemas itself. However, Corda processors all expect the JDBC URL in the config to point to the config schema
+     * directly, so the name of that schema must be added to the params that are used to create the config.
+     */
+    private fun prepareDbConfig(dbConfig: Config): Config {
+        val tempJdbcUrl = dbConfig.getString(BOOT_JDBC_URL)
+        return dbConfig
+            .withValue(BOOT_JDBC_URL, fromAnyRef("$tempJdbcUrl?currentSchema=$CONFIG_SCHEMA_NAME"))
+            .withValue(BOOT_JDBC_URL + "_messagebus", fromAnyRef("$tempJdbcUrl?currentSchema=$MESSAGEBUS_SCHEMA_NAME"))
```
JamesHR3: Does the string `BOOT_JDBC_URL + "_messagebus"` have a constant somewhere?

charts/corda-lib/templates/_worker.tpl
jujoramos: @conalsmith-r3 : you could try adding a short alias within the `DefaultWorkerParams.kt` for the state manager and replace all startup parameters with `-SparamName=paramValue`:

```
    @Option(
        names = ["-S", "--${BootConfig.BOOT_STATE_MANAGER}", ]
        description = ["Configuration for the state manager."]
    )
```

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/configuration/StateManagerConfigProvider.kt
```diff
@@ -0,0 +1,24 @@
...
+            .withValue(MessagingConfig.StateManager.JDBC_DRIVER, fromAnyRef("org.postgresql.Driver"))
+            .withValue(MessagingConfig.StateManager.JDBC_PERSISTENCE_UNIT_NAME, fromAnyRef("corda-state-manager"))
+            .withValue(MessagingConfig.StateManager.JDBC_POOL_MIN_SIZE, fromAnyRef(1))
+            .withValue(MessagingConfig.StateManager.JDBC_POOL_MAX_SIZE, fromAnyRef(5))
+            .withValue(MessagingConfig.StateManager.JDBC_POOL_IDLE_TIMEOUT_SECONDS, fromAnyRef(Duration.ofMinutes(2).toSeconds()))
+            .withValue(MessagingConfig.StateManager.JDBC_POOL_MAX_LIFETIME_SECONDS, fromAnyRef(Duration.ofMinutes(30).toSeconds()))
+            .withValue(MessagingConfig.StateManager.JDBC_POOL_KEEP_ALIVE_TIME_SECONDS, fromAnyRef(Duration.ZERO.toSeconds()))
+            .withValue(
+                MessagingConfig.StateManager.JDBC_POOL_VALIDATION_TIMEOUT_SECONDS, fromAnyRef(Duration.ofSeconds(5).toSeconds())
+            )
```
jujoramos: Any reasoning behind setting the default values in the code instead of using the helm chart as before?

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
sprintengine: Not that I could find, "_messagebus" seems only used when db message bus is used, and this constant ought to be shared between CombinedWorker and DbBusConfigMergerImpl. I don't know exactly where this should go, but for now think I'll extract it to the companion object so at least it's more visible.

applications/workers/worker-common/src/main/resources/net/corda/applications/workers/workercommon/boot/corda.boot.json
```diff
@@ -192,6 +192,59 @@
...
+              ]
+            },
+            "pass": {
+              "description": "Admin password to use by the state manager when connecting to the database used as the persistent storage",
+              "type": [
+                "string",
+                "null"
+              ]
+            },
+            "additionalProperties": false
```
sprintengine: I think we're missing the pool configurations here, which allows helm chart to supply these configs and allow the boot config validation to pass:

```
database.pool.maxSize
database.pool.minSize
database.pool.idleTimeoutSeconds
database.pool.maxLifetimeSeconds
database.pool.keepAliveTimeSeconds
database.pool.validationTimeoutSeconds
```

jujoramos: I don't recall having these properties defined within the boot scheme before either... 🤔. I also can't seem to find where the pipeline is failing for this `PR` now , so I might be looking at the wrong thing here :-/.

sprintengine: There looks like a bug with our [boot config validation](https://github.com/corda/corda-runtime-os/blob/ed98b4553d997c73a5eb3a49f26f7100dabe4e22/applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/WorkerHelpers.kt#L167), as I can provide random parameters via command line and they pass validation against `corda.boot.json`. That's why we can provide these database pool config (and state manager pool config) without adding them to corda.boot.json.

EDIT - this is to do with the "additionalProperties" attribute. It's not clear if it's intentional to allow workers with this version of the schema for backwards compatibility to pass validation if it receives startup config for new properties at certain levels (e.g. boot.db.database level). 

I will restrict stateManager config from additionalProperties at all levels except the boot.stateManager level which may allow a redis configuration in future.

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
```diff
@@ -220,8 +223,20 @@ class CombinedWorker @Activate constructor(
...
+            .withValue(MessagingConfig.StateManager.JDBC_PERSISTENCE_UNIT_NAME, fromAnyRef("corda-state-manager"))
+            .withValue(MessagingConfig.StateManager.JDBC_POOL_MIN_SIZE, fromAnyRef(1))
+            .withValue(MessagingConfig.StateManager.JDBC_POOL_MAX_SIZE, fromAnyRef(5))
+            .withValue(MessagingConfig.StateManager.JDBC_POOL_IDLE_TIMEOUT_SECONDS, fromAnyRef(Duration.ofMinutes(2).toSeconds()))
+            .withValue(MessagingConfig.StateManager.JDBC_POOL_MAX_LIFETIME_SECONDS, fromAnyRef(Duration.ofMinutes(30).toSeconds()))
+            .withValue(MessagingConfig.StateManager.JDBC_POOL_KEEP_ALIVE_TIME_SECONDS, fromAnyRef(Duration.ZERO.toSeconds()))
+            .withValue(
+                MessagingConfig.StateManager.JDBC_POOL_VALIDATION_TIMEOUT_SECONDS, fromAnyRef(Duration.ofSeconds(5).toSeconds())
+            )
+        val stateManagerConfigWithFallback = stateManagerConfig.withFallback(defaultConfig)
```
sprintengine, line 237: Helm template and values.yaml provides defaults for the pool config for other workers, here we apply defaults for the combined worker as it does not use helm. These config are validated against the corda.boot.json schema.

EDIT - I tried getting the boot config to apply the defaults but it wouldn't do it for me, even when setting the object as required. We give the jsonSchemaFactory a schemaValidatorConfig that can apply defaults for missing properties but it doesn't seem to do it for missing objects.

applications/workers/worker-common/src/main/resources/net/corda/applications/workers/workercommon/boot/corda.boot.json
```diff
@@ -32,55 +32,55 @@
...
       "properties": {
-        "database" : {
-              "description": "Database params",
+        "database": {
+          "description": "Database params",
+          "properties": {
+            "jdbc": {
+              "description": "The JDBC database info",
               "properties": {
-                "jdbc": {
```
sprintengine, line 38: this is just reformatting, I didn't change the "additionalProperties" boolean in these objects.

applications/workers/worker-common/src/main/resources/net/corda/applications/workers/workercommon/boot/corda.boot.json
```diff
@@ -221,7 +221,22 @@
                     "string",
                     "null"
                   ]
-                }
+                },
```
sprintengine, line 224: actual changes in this file start here

applications/workers/worker-common/src/main/resources/net/corda/applications/workers/workercommon/boot/corda.boot.json
```diff
@@ -221,7 +221,22 @@
...
+                "driver": {
+                  "description": "The JDBC driver to use by the state manager when using database as the persistent storage",
+                  "type": [
+                    "string",
+                    "null"
+                  ]
+                },
+                "persistenceUnitName": {
+                  "description": "The persistence unit name to use by the state manager when using database as the persistent storage",
```
sprintengine, line 233: when we remove hibernate we won't use a persistence unit, however, since nobody is using state manager, we can remove this config as part of that task (it is probably a breaking change)

applications/workers/worker-common/src/main/resources/net/corda/applications/workers/workercommon/boot/corda.boot.json
```diff
@@ -239,13 +254,70 @@
...
               ]
             },
-            "additionalProperties": false
-          }
-        },
-        "additionalProperties": false
-      }
-    },
-    "additionalProperties": false
+            "pool": {
```
sprintengine, line 257: I added the pool config, no additional properties are allowed in this config.

applications/workers/worker-common/src/main/resources/net/corda/applications/workers/workercommon/boot/corda.boot.json
```diff
@@ -239,13 +254,70 @@
...
+                  ]
+                }
+              },
+              "additionalProperties": false
+            }
+          },
+          "additionalProperties": false
+        }
+      },
+      "additionalProperties": true
```
sprintengine, line 319: explicitly allowing additional properties at the boot.stateManager level - compatibility if we add "redis" config in future?

charts/corda-lib/templates/_worker.tpl
```diff
@@ -300,7 +300,7 @@ spec:
           - "--stateManager"
           - "database.pool.maxLifetimeSeconds={{ .stateManagerDbConnectionPool.maxLifetimeSeconds }}"
           - "--stateManager"
-          - "database.pool.keepAliveTimeSeconds={{ .stateManagerDbConnectionPool.keepAliveTimeSeconds }}"
+          - "database.pool.keepaliveTimeSeconds={{ .stateManagerDbConnectionPool.keepaliveTimeSeconds }}"
```
conalsmith-r3, line 303: I noticed we've introduced slightly inconsistent naming for this property, I just want to keep it consistent

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
sprintengine: I think now that stateManagerParams are in the defaultWorkerParams, they already apply to the combinedWorker.

We can run combined worker OK without providing any extra params. but now have the option to change the config

.run/Combined Worker Local.run.xml
```diff
@@ -2,7 +2,8 @@
...
+    <!-- TODO-[CORE-16419]: isolate StateManager database from the Cluster database -->
+    <option name="PROGRAM_PARAMETERS" value="--instance-id=0 -mbus.busType=DATABASE -spassphrase=password -ssalt=salt -spassphrase=password -ssalt=salt -ddatabase.user=user -ddatabase.pass=password -ddatabase.jdbc.url=jdbc:postgresql://localhost:5432/cordacluster -ddatabase.jdbc.directory=$ProjectFileDir$/applications/workers/release/combined-worker/drivers --stateManager type=DATABASE --stateManager database.user=user --stateManager database.pass=password --stateManager database.jdbc.url=jdbc:postgresql://localhost:5432/cordacluster -rtls.crt.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/server.crt -rtls.key.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/server.key -rtls.ca.crt.path=$ProjectFileDir$/applications/workers/release/combined-worker/tls/rest/ca-chain-bundle.crt" />
```
jujoramos: Nitpick: I think this change is not needed anymore as the state manager will pick up default db config if nothing is set?.

applications/workers/release/combined-worker/src/main/kotlin/net/corda/applications/workers/combined/CombinedWorker.kt
jujoramos: Yeah, got it, was just trying to say that we don't need this at the moment as the combined worker will always use a single database for everything.

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -510,8 +510,9 @@ a second init container to execute the output SQL to the relevant database
...
   {{- include "corda.bootstrapResources" . | nindent 2 }}
   {{- include "corda.containerSecurityContext" . | nindent 2 }}
+  {{- /* TODO-[CORE-16419]: isolate StateManager database from the Cluster database */ -}}
   {{- if eq .name "db" }}
-  args: [ 'database', 'spec', '-g', 'config:{{ .Values.db.cluster.schema }},rbac:{{ .Values.bootstrap.db.rbac.schema }},crypto:{{ .Values.bootstrap.db.crypto.schema }}', '-c', '-l', '/tmp', '--jdbc-url', 'jdbc:{{ include "corda.clusterDbType" . }}://{{ required "A db host is required" .Values.db.cluster.host }}:{{ include "corda.clusterDbPort" . }}/{{ include "corda.clusterDbName" . }}', '-u', $(PGUSER), '-p', $(PGPASSWORD) ]
+  args: [ 'database', 'spec', '-g', 'config:{{ .Values.db.cluster.schema }},rbac:{{ .Values.bootstrap.db.rbac.schema }},crypto:{{ .Values.bootstrap.db.crypto.schema }},stateManager:{{ .Values.bootstrap.db.stateManager.schema }}', '-c', '-l', '/tmp', '--jdbc-url', 'jdbc:{{ include "corda.clusterDbType" . }}://{{ required "A db host is required" .Values.db.cluster.host }}:{{ include "corda.clusterDbPort" . }}/{{ include "corda.clusterDbName" . }}', '-u', $(PGUSER), '-p', $(PGPASSWORD) ]
```
jujoramos: Nitpick: You are already using the schema from `.Values.bootstrap.db.stateManager` , so I think you can remove the `TODO` now.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/config/KafkaConfigMergerImpl.kt
```diff
@@ -37,6 +40,16 @@ class KafkaConfigMergerImpl : BusConfigMerger {
...
+        logger.debug { "Looping through State Manager Boot Configuration" }
+        val stateManagerBootConfig = bootConfig.getConfigOrEmpty(BOOT_STATE_MANAGER).entrySet()
+        stateManagerBootConfig.forEach { entry ->
+            logger.debug { "Entry key: ${entry.key}" }
+            updatedMessagingConfig = updatedMessagingConfig.withValue(
+                "${MessagingConfig.StateManager.STATE_MANAGER}.${entry.key}",
+                ConfigValueFactory.fromAnyRef(bootConfig.getString("$BOOT_STATE_MANAGER.${entry.key}"))
+            )
+        }
+
```
jujoramos, line 52: I think we need to address this one before merging: https://github.com/corda/corda-runtime-os/pull/4501#discussion_r1317092310

applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/WorkerHelpers.kt
```diff
@@ -33,9 +32,26 @@ enum class BusType {
...
+            "database.pass",
+            "--stateManager.database.pass",
+            "-spassphrase",
+            "-msasl.jaas.config"
+        )
+
+        /**
+         * Define the paths of any boot config that must be treated as integers for boot config json validation.
+         */
+        private val BOOT_CONFIG_INTEGER_PATHS = setOf(
```
sprintengine, line 47: I introduced this because now I'm validating the boot schema properly, particularly the state manager pool config, and PicoCLI is giving us a Map<String, String>, so when we create a config, we have to convert these to Integers. This applies to any worker that tries to add state manager config.

sprintengine: I need to add a unit test for this logic

.run/Combined Worker Local.run.xml
```diff
@@ -2,6 +2,7 @@
   <configuration default="false" name="Combined Worker Local" type="JarApplication">
     <option name="JAR_PATH" value="$PROJECT_DIR$/applications/workers/release/combined-worker/build/bin/corda-combined-worker-5.1.0.0-SNAPSHOT.jar" />
     <option name="VM_PARAMETERS" value="-Dco.paralleluniverse.fibers.verifyInstrumentation=true" />
+    <!-- TODO-[CORE-16419]: isolate StateManager database from the Cluster database -->
```
jujoramos: Nitpick: remove this TODO.

JamesHR3 (dismissed):

jujoramos (dismissed):
LGTM

charts/corda-lib/templates/_worker.tpl
sprintengine: An earlier discussion decided to get rid of the short param as it was confusing with secrets config (`-s`)

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/config/KafkaConfigMergerImpl.kt
sprintengine: Will follow up with a change to address this

jujoramos (approved):
LGTM

conal-smith:
Squash commit message:

```
CORE-17154 - Enable state manager to be configured as part of worker deployment.
- New `--stateManager` boot option for all workers to configure a state manager.
- State manager configuration is optional for the combined worker and defaults to the cluster database.
- Worker helm template supports deployment of workers with state manager config, reading configuration from values.yaml.
- state manager boot configuration provided at worker startup is validated against boot config json schema.
- STATE_MANAGER schema is deployed as part of application db setup.
- Refactor getBootstrapConfig so we can supply extraConfig, and we can supply a list of keys in the config that must be converted to Integer before boot config schema validation.

Co-authored-by: Juan José Ramos Cassella <jujoramos@gmail.com>
```

wzur-r3 (approved):
LGTM for Gradle files

filipesoliveira (approved):

kyriathar (approved):
OK (db changes)

davidcurrie (approved):
LGTM for @corda/cluster-management

PR #4625, opened 2023-09-12
CORE-17062 - Token Selection - Update how an instance of `TokenClaim` is created

filipesoliveira:
Before, a `TokenClaim` instance was created using the default constructor plus the apply method. This strategy should be avoided because the defaults specified in the Avro schema are not applied to the fields that are not explicitly set inside the apply method. All objects are set to null if they are not explicitly set which causes issues when data is serialised to Json and the Avro schema does not accept a null value.


owenstanford (approved):
LGTM

PR #4626, opened 2023-09-12
CORE-16965: Proof of Action Lock State (Part 1)

julia-filipczak:


testing/cpbs/atomic-swap-workflows/src/main/kotlin/com/r3/corda/atomic/swap/workflows/CreateLockFlow.kt
```diff
@@ -0,0 +1,151 @@
...
+
+            val assetWithCompositeKey = constructLockedAsset(inputState, newOwnerInfo)
+
+            val outputState = inputState.withNewOwner(
+                assetWithCompositeKey.owner,
+                listOf(ownerInfo.ledgerKeys[0], newOwnerInfo.ledgerKeys[0])
+//                listOf(assetWithCompositeKey.owner)
+            )
+
+            log.info("KEYS PLEASE lockState.participants: ${lockState.participants}")
```
szymonsztuka: Remove all statements `log.info("KEYS PLEASE`... used when troubleshooting.

testing/cpbs/atomic-swap-workflows/src/main/kotlin/com/r3/corda/atomic/swap/workflows/CreateLockFlow.kt
```diff
@@ -0,0 +1,151 @@
...
+
+            val lockState = LockState(
+                inputState.owner,
+                newOwnerInfo.ledgerKeys[0],
+                inputState.assetName,
+                inputState.assetId,
+                listOf(inputState.owner, newOwnerInfo.ledgerKeys[0])
+            )
+
+            val assetWithCompositeKey = constructLockedAsset(inputState, newOwnerInfo)
```
szymonsztuka: `assetWithCompositeKey` is not used as a whole and it's an unnecessary intermediate object, consider in-lining this code here:
```
val compositeKey = compositeKeyGenerator.create(
            listOf(
                CompositeKeyNodeAndWeight(asset.owner, 1),
                CompositeKeyNodeAndWeight(newOwner.ledgerKeys.single(), 1)
            ), 1
        )

        outputState.withNewOwner(compositeKey, listOf(asset.owner, newOwner.ledgerKeys[0]))
```

testing/cpbs/atomic-swap-workflows/src/main/kotlin/com/r3/corda/atomic/swap/workflows/CreateLockFlow.kt
```diff
@@ -0,0 +1,151 @@
...
+
+            val outputState = inputState.withNewOwner(
+                assetWithCompositeKey.owner,
+                listOf(ownerInfo.ledgerKeys[0], newOwnerInfo.ledgerKeys[0])
+//                listOf(assetWithCompositeKey.owner)
```
szymonsztuka: Remove unused code

testing/cpbs/atomic-swap-workflows/src/main/kotlin/com/r3/corda/atomic/swap/workflows/UnlockFlow.kt
```diff
@@ -0,0 +1,106 @@
...
+            val txBuilder = ledgerService.createTransactionBuilder()
+
+                .setNotary(stateAndRefAsset.state.notaryName)
+                .setTimeWindowBetween(Instant.now(), Instant.now().plusMillis(Duration.ofDays(1).toMillis()))
+                .addInputState(stateAndRefLock.ref)
+                .addInputState(stateAndRefAsset.ref)
+                .addOutputState(newState)
+                .addCommand(LockContract.LockCommands.Unlock(true))
+                .addCommand(AssetContract.AssetCommands.Transfer())
+//                .addSignatories(inputLockState.participants)// alice, bob
```
szymonsztuka: Remove unused code

testing/cpbs/atomic-swap-workflows/src/main/kotlin/com/r3/corda/atomic/swap/workflows/UnlockFlow.kt
```diff
@@ -0,0 +1,106 @@
...
+                .addSignatories(newState.participants)//bob
```
szymonsztuka: remove `//bob`

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/AssetContract.kt
```diff
@@ -0,0 +1,70 @@
...
+        class Create: AssetCommands
+        class Transfer: AssetCommands
+    }
+
+
+    override fun verify(transaction: UtxoLedgerTransaction) {
+        val command = transaction.getCommands(AssetCommands::class.java).singleOrNull()  ?: throw CordaRuntimeException(
+            REQUIRE_SINGLE_COMMAND
+        )
+        // Switches case based on the command
```
szymonsztuka: Remove all comemnd in this calls, they are duplicating exactly the code e.g.:
```
// Rules applied only to transactions with the Transfer Command.
            is AssetCommands.Transfer -> {
```

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/AssetContract.kt
```diff
@@ -0,0 +1,70 @@
...
+import net.corda.v5.base.exceptions.CordaRuntimeException
+import net.corda.v5.ledger.utxo.Command
+import net.corda.v5.ledger.utxo.Contract
+import net.corda.v5.ledger.utxo.transaction.UtxoLedgerTransaction
+
+
+class AssetContract: Contract {
+
+    internal companion object {
+        const val REQUIRE_SINGLE_COMMAND = "Requires a single command."
```
szymonsztuka: I would inline all text into the contract code, it makes the contract code more readable.

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/AssetContract.kt
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    // Helper function to allow writing constraints in the Corda 4 '"text" using (boolean)' style
+    private infix fun String.using(expr: Boolean) {
+        if (!expr) throw CordaRuntimeException("Failed requirement: $this")
+    }
+
+    // Helper function to allow writing constraints in '"text" using {lambda}' style where the last expression
+    // in the lambda is a boolean.
+    private infix fun String.using(expr: () -> Boolean) {
```
szymonsztuka: I think this method is not used so we can remove it

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/AssetContract.kt
```diff
@@ -0,0 +1,70 @@
...
+        )
+        // Switches case based on the command
+        when(command) {
+            // Rules applied only to transactions with the Create Command.
+            is AssetCommands.Create -> {
+                REQUIRES_ZERO_INPUTS using (transaction.inputContractStates.isEmpty())
+                REQUIRES_ONE_OUTPUT using (transaction.outputContractStates.size == 1)
+                REQUIRES_ONE_ASSET_OUTPUT using (transaction.getOutputStates(Asset::class.java).size == 1)
+
+                val asset = transaction.getOutputStates(Asset::class.java)[0]
```
szymonsztuka: In all those cases instead of index access `[0]`we can use method `.first()`

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
```diff
@@ -0,0 +1,55 @@
...
+
+            is LockCommands.Unlock -> {
+                val input = transaction.getInputStates(LockState::class.java).singleOrNull()
+                    ?: throw CordaRuntimeException("Can't find lock state for verification")
+                "Unlock takes one command input of type Boolean and the value should be true." using
+                        (command.bool == input.bool)
+                "There should be one input state as the lock state needs to be consumed" using
+                        (transaction.getInputStates(LockState::class.java).size == 1)
+                "There should be one output state as an unlocked asset state needs to be created" using
+                        (transaction.outputContractStates.size == 1)
```
szymonsztuka, line 41: The unlock command should allow only a specific "actor" to unlock an asset.

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
```diff
@@ -0,0 +1,55 @@
...
+                throw CordaRuntimeException("Command not allowed.")
+            }
+        }
+    }
+
+    private infix fun String.using(expr: Boolean) {
+        if (!expr) throw CordaRuntimeException("Failed requirement: $this")
+    }
+
+    private infix fun String.using(expr: () -> Boolean) {
```
szymonsztuka: I think this is not in use.

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/states/Asset.kt
```diff
@@ -0,0 +1,28 @@
...
+import net.corda.v5.ledger.utxo.BelongsToContract
+import java.security.PublicKey
+
+@BelongsToContract(AssetContract::class)
+
+data class Asset (
+    private val owner : PublicKey,
+    val assetName: String,
+    val assetId: String,
+    private val participants: List<PublicKey>) : OwnableState {
```
szymonsztuka: This is lowest priority, I think we should make this constructor private one, and add another constructor which would set `participants = listOf(owner)`

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/states/Asset.kt
```diff
@@ -0,0 +1,28 @@
...
+    override fun getParticipants(): List<PublicKey> {
+        return participants
+    }
+
+    override fun getOwner(): PublicKey {
+        return owner
+    }
+
+    fun withNewOwner(newOwner: PublicKey, newParticipants: List<PublicKey>): Asset {
+        return Asset(newOwner, assetName, assetId, newParticipants)
```
szymonsztuka, line 25: This is lowest priority, we should derived participants like that:
we should retain the previous owner in participants list and then add the new owner.
In general here I think participants should be always the owner and eventually former owner of the asset, we can compute it internally no need to force user to put it in parameters.

testing/cpbs/atomic-swap-workflows/src/main/kotlin/com/r3/corda/atomic/swap/workflows/CreateLockFlow.kt
```diff
@@ -0,0 +1,151 @@
...
+
+                .setNotary(stateAndRef.state.notaryName)
+                .setTimeWindowBetween(Instant.now(), Instant.now().plusMillis(Duration.ofDays(1).toMillis()))
+                .addInputState(stateAndRef.ref)
+                .addOutputState(lockState)
+                .addOutputState(outputState)
+                .addCommand(LockContract.LockCommands.Lock())
+                .addCommand(AssetContract.AssetCommands.Transfer())
+                .addSignatories(lockState.participants)
+                .addSignatories(assetWithCompositeKey.owner)
```
szymonsztuka: I'm curious that this type of lock works as it doesn't use encumbrances.

szymonsztuka (changes requested):

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/AssetContract.kt
```diff
@@ -0,0 +1,51 @@
...
+
+            is AssetCommands.Transfer -> {
+                "The transaction requires one Asset output." using (transaction.getOutputStates(Asset::class.java).size == 1)
+                "The transaction requires one Asset input." using (transaction.getInputStates(Asset::class.java).size == 1)
+
+                val input = transaction.getInputStates(Asset::class.java).first()
+
+                val output = transaction.getOutputStates(Asset::class.java).first()
+                "Owner must change in this transaction." using (input.owner != output.owner)
+            }
```
szymonsztuka, line 41: Should we add the check that (at least) input.owner should sign the transaction, or maybe even old and new owners should sign the transaction. This is in fact business dependent, but we could keep consistent (as in `AssetCommands.Create` we do check who sign, so maybe in `AssetCommands.Transfer` we should have such type of a check as well.)

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
```diff
@@ -0,0 +1,52 @@
...
+            is LockCommands.Unlock -> {
+                val input = transaction.getInputStates(LockState::class.java).singleOrNull()
+                    ?: throw CordaRuntimeException("Can't find lock state for verification")
+                "Unlock takes one command input of type Boolean and the value should be true." using
+                        (command.bool == input.bool)
+                "There should be one input state as the lock state needs to be consumed" using
+                        (transaction.getInputStates(LockState::class.java).size == 1)
+                "There should be one output state as an unlocked asset state needs to be created" using
+                        (transaction.outputContractStates.size == 1)
+                "The signer of the unlock should be the new owner of the Asset state." using (outputState.participants == transaction.signatories)
```
szymonsztuka: There is a single "actor" who can unlock only, the other actor cant unlock. Also only one "actor" do sign unlock (in business terms only one actor will unlock).

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/states/LockState.kt
```diff
@@ -0,0 +1,24 @@
...
+import net.corda.v5.ledger.utxo.BelongsToContract
+import net.corda.v5.ledger.utxo.ContractState
+import java.security.PublicKey
+
+@BelongsToContract(LockContract::class)
+data class LockState (
+    val creator : PublicKey,
+    val receiver: PublicKey,
+    val assetName: String,
+    val assetId: String,
```
szymonsztuka, line 12: I don't think we need `assetId` and `assetName`, at least in the most plain version of lock. They may serve some "compliance" purpose / vault search but they absolutely are not used in the contract logic.

szymonsztuka (changes requested):

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
julia-filipczak: In this case `outputState.participants` is a single actor.

szymonsztuka: The contract code has no such constraint (how many participants are in `participants` collections), and more other `LockCommands.Lock` suggests there may be  2 participants.
You are inferring some extra knowledge looking at code written in flows but a contract code needs to be independent and prevent people of writing wrong flows (flows creating wrong transactions).

szymonsztuka (changes requested):

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
```diff
@@ -0,0 +1,55 @@
...
+        class Lock : LockCommands
+        class Unlock(val bool: Boolean) : LockCommands
+    }
+
+    override fun verify(transaction: UtxoLedgerTransaction) {
+
+        val command = transaction.getCommands(LockCommands::class.java).singleOrNull()
+            ?: throw CordaRuntimeException("Requires a single command.")
+
+        val outputState = transaction.getOutputStates(Asset::class.java).first()
```
szymonsztuka: I don't think the Lock contract should have access or require code of Asset class. Can we built `LockContract` which is agnostic to an object (states) it's locking or depends on as minimally as possible? 
For example what if we what to use our Lock with "Orange" state or "Bushel of Wheat" State ? Then we would need to write new Lock code with `val outputState = transaction.getOutputStates("Bushel of Wheat"::class.java).first()`

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
```diff
@@ -0,0 +1,55 @@
...
+                val input = transaction.getInputStates(LockState::class.java).singleOrNull()
+                    ?: throw CordaRuntimeException("Can't find lock state for verification")
+                "Unlock takes one command input of type Boolean and the value should be true." using
+                        (command.bool == input.bool)
+                "There should be one input state as the lock state needs to be consumed" using
+                        (transaction.getInputStates(LockState::class.java).size == 1)
+                "There should be one output state as an unlocked asset state needs to be created" using
+                        (transaction.outputContractStates.size == 1)
+                "The signer of the unlock should be the new owner of the Asset state." using
+                        (outputState.owner == transaction.signatories)
```
szymonsztuka: Here we just check that the new owner has signed it, but I think there are specific condition who can be the new owner. The contract code doesn't state it, doesn't check it, so a thief "actor"  (aka not Bob or alice) can assigned themself to be an owner of the state and added themself as signatory.

szymonsztuka (changes requested):

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
```diff
@@ -0,0 +1,60 @@
...
+
+            is LockCommands.Unlock -> {
+                val input = transaction.getInputStates(LockState::class.java).singleOrNull()
+                    ?: throw CordaRuntimeException("Can't find lock state for verification")
+                "Unlock takes one command input of type Boolean and the value should be true." using
+                        (command.bool == input.bool)
+                "There should be one input state as the lock state needs to be consumed" using
+                        (transaction.getInputStates(LockState::class.java).size == 1)
+                "There should be one output state as an unlocked asset state needs to be created" using
+                        (transaction.outputContractStates.size == 1)
```
szymonsztuka, line 41: And I would also check that Lock is not longer in output states, like? :
`(transaction.getOnputStates(LockState::class.java).size == 0)`

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
```diff
@@ -0,0 +1,60 @@
...
+                "The new owner should not be the old owner" using (outputState.owner != input.creator)
+                "The new owner should be the same as the receiver of the lock state" using
+                        (outputState.owner == input.receiver)
+                "There should be one signer for the unlock command." using
+                        (transaction.signatories.size == 1)
+                "The signer of the unlock should be the new owner of the Asset state." using
+                        (outputState.owner == transaction.signatories.first())
```
szymonsztuka: Check this in similar way as in Asset - via checking if a collection contain an element.

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/states/Asset.kt
szymonsztuka: Due to some Kotlin wrnings as discoverd by @julia-filipczak  (and it probably know problem https://discuss.kotlinlang.org/t/data-class-with-private-constructor/25599) let's skip that refactoring I was proposing.

testing/cpbs/atomic-swap-workflows/src/main/kotlin/com/r3/corda/atomic/swap/workflows/CreateLockFlow.kt
```diff
@@ -0,0 +1,139 @@
...
+            )
+
+        } catch (e: Exception) {
+            log.warn("Failed to process utxo flow for request body '$requestBody' because: '${e.message}'")
+            throw e
+        }
+    }
+
+    private fun constructCompositeKey(asset: Asset, newOwner: MemberInfo): PublicKey {
+        // Build composite key
```
szymonsztuka: Remove unnecessary comment + maybe more compact code:
```
    private fun constructCompositeKey(asset: Asset, newOwner: MemberInfo): PublicKey =
         compositeKeyGenerator.create(
            listOf(
                CompositeKeyNodeAndWeight(asset.owner, 1),
                CompositeKeyNodeAndWeight(newOwner.ledgerKeys.single(), 1)
            ), 1
        )
```

szymonsztuka (changes requested):

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/states/Asset.kt
szymonsztuka: Ignore this comment - see https://github.com/corda/corda-runtime-os/pull/4626#discussion_r1346147857

szymonsztuka (approved):

PR #4627, opened 2023-09-12
CORE-17074: Upgrade Kotlin Metadata 0.6.2 -> 0.7.0.

chrisr3:
Kotlin Metadata 0.7.0 is a breaking change from 0.6.2, but thankfully it's just an internal dependency.

Also upgrade ASM 9.4 -> 9.5 for `-conditionalpackage` dependency.

### DIFF
--- a/libs/kotlin-reflection/src/main/kotlin/net/corda/kotlin/reflect/types/KotlinMutableProperty1.kt
+++ b/libs/kotlin-reflection/src/main/kotlin/net/corda/kotlin/reflect/types/KotlinMutableProperty1.kt
@@ -1,14 +1,5 @@
 package net.corda.kotlin.reflect.types
 
-import kotlinx.metadata.Flag.Function.IS_INFIX
-import kotlinx.metadata.Flag.Function.IS_INLINE
-import kotlinx.metadata.Flag.Function.IS_OPERATOR
-import kotlinx.metadata.Flag.Function.IS_SUSPEND
-import kotlinx.metadata.Flag.Property.IS_EXTERNAL
-import kotlinx.metadata.Flags
-import kotlinx.metadata.KmProperty
-import kotlinx.metadata.jvm.getterSignature
-import kotlinx.metadata.jvm.setterSignature
 import java.lang.reflect.Field
 import java.lang.reflect.Method
 import java.util.Objects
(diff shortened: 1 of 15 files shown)


wzur-r3 (approved):
LGTM for files owned by BLT

libs/kotlin-reflection/src/main/kotlin/net/corda/kotlin/reflect/types/KotlinMutableProperty1.kt
```diff
@@ -18,37 +9,45 @@ import kotlin.reflect.KType
...
+        setterSignature = kmProperty.setterSignature?.toSignature(declaringClass.classLoader),
         javaField = null,
         javaGetter = null,
         javaSetter = null,
         instanceClass = declaringClass
     )
 
     override val isAbstract: Boolean
-        get() = isAbstract(javaGetter, javaSetter, kmProperty.flags)
+        get() = isAbstract(javaGetter, javaSetter, kmProperty.setter!!.modality)
```
dickon, line 42: Can we assume isAbstract is false if kmProepty.setter is null, to avoid the !!? Also in the other cases

dickon (approved):
A minor comment on use of !!, but otherwise LGTM, let's keep up to date.

libs/kotlin-reflection/src/main/kotlin/net/corda/kotlin/reflect/types/KotlinMutableProperty1.kt
chrisr3: The `setter` cannot be `null` for a mutable property; that's what makes it mutable:
```kotlin
    /**
     * Attributes of the setter of this property.
     * Attributes can be retrieved with extension properties, such as [KmPropertyAccessorAttributes.visibility] or [KmPropertyAccessorAttributes.isNotDefault].
     *
     * Returns null if setter is absent, i.e. [KmProperty.isVar] is false.
     */
    var setter: KmPropertyAccessorAttributes? = if (this.hasSetter) KmPropertyAccessorAttributes(setterFlags) else null
        set(new) {
            this.hasSetter = new != null
            field = new
        }
```
where:
```kotlin
fun <T, V> Class<*>.createKotlinProperty1(kmProperty: KmProperty): KotlinProperty1<T, V> {
    return if (kmProperty.isVar) {
        KotlinMutableProperty1(kmProperty, declaringClass = this)
    } else {
        KotlinProperty1(kmProperty, declaringClass = this)
    }
}
```

rick-r3 (approved):

PR #4628, opened 2023-09-12
CORE-16614: index utxo_visible_transaction_state.consumed

jennyang-r3:
This PR indexed `consumed` field of `utxo_visible_transaction_state` table.


jennyang-r3:
Closing this PR because the expected change is in api repo instead of runtime.

PR #4630, opened 2023-09-12
CORE-17065: Ensure session messages get routed in the flow mapper correctly

JamesHR3:
**Problem statement**

The flow mapper contains logic to determine where to forward session messages to depending on where they have come from and who the recipient is. Inbound messages should be forwarded to the local flow engine, while outbound messages are forwarded to either the mapper itself (if the recipient is local to the sender) or to the P2P layer (if it isn't). Previously this check for whether the session event recipient was local or remote was incorrectly using the sender's identity rather than the recipients.

The code for generating forwarded session events also appears to have a bug whereby messages sent to remote peers would lose the sequence number of the outbound message.

**Solution**

Use the existing `getSourceAndDestinationIdentity` function to get the correct destination identity. Additionally, modify the `RecordFactoryImpl` to make the division of responsibility clearer. This class is now responsible for generating the forwarded session record, given the source record. All other classes in the mapper implementation now use this to generate forwarded session records.

Previously the division of responsibility was less clear, which could cause situations where the session events were generated inconsistently. This also allows a lot of utility functions to be better encapsulated by a single class, reducing the need for the mapper helpers.

**Testing**

- Full passing run of the multi-cluster e2e tests.

**Other changes**

On flow transient errors currently there is no logging indicating why the flow cannot proceed immediately. An info level log has been added to indicate what the flow is waiting for in order to proceed.


jujoramos (approved):

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/RecordFactoryImpl.kt
```diff
@@ -34,84 +42,156 @@ class RecordFactoryImpl @Activate constructor(
...
+            Schemas.Flow.FLOW_MAPPER_EVENT_TOPIC -> Pair(MessageDirection.INBOUND, toggleSessionId(sourceEvent.sessionId))
+            Schemas.Flow.FLOW_EVENT_TOPIC -> Pair(MessageDirection.INBOUND, sourceEvent.sessionId)
+            else -> Pair(MessageDirection.OUTBOUND, sourceEvent.sessionId)
+        }
+        val sequenceNumber = if (newPayload is SessionError) null else sourceEvent.sequenceNum
+        val sessionEvent = SessionEvent(
+            newDirection,
+            timestamp,
+            sessionId,
+            sequenceNumber,
```
sprintengine, line 107: So now, when we create outbound sessionEvent record we set the sequence number - whereas beforehand it was always null?

conal-smith (approved):
LGTM

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/RecordFactoryImpl.kt
JamesHR3: I think we potentially had a problem where if the outbound record was going to a remote cluster, we would have washed off the sequence number (incorrectly). We would never have hit this problem because the original bug would prevent us from hitting that code path anyway.

emilybowe (approved):
LGTM, discussed offline 👍

PR #4577, opened 2023-09-04
CORE-16195 Uniqueness RPC external event processor

mbrkic-r3:
Uniqueness RPC external event processor

PR #4470, opened 2023-08-15
CORE-16193 - HTTP RPC subscription and webserver 

pmck91:
This PR will add an RPC subscription component that will allow inter-worker communication via RPC over HTTP. 

To achieve this, I have used the webserver already set up by the HealthMonitor, and have also uncoupled the dependence on Javalin. Both the health monitor and RPC subscription are agnostic of server implementation, but I have created a Javalin Implementation for them to use.

This is the server side of the RPC pattern, each worker will need to define their own exposed RPC methods and use this component to load the relevant endpoints into the web server.

PR #4631, opened 2023-09-12
CORE-16195 - uniqueness-rpc-processor

driessamyn:
- Integration of a HTTP RPC server into the Uniqueness processor.
- Smoketests include a test for the service outside the scope of a flow.
- Bunch of fixes/changes to the JavalinServer implementation.
- No integration of a client for this service yet, this will come in a subsequent change.

--
NOTE:
Supersedes  https://github.com/corda/corda-runtime-os/pull/4577 
Commits squashed to clean up history, original commits by @LWogan and @mbrkic-r3


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
```diff
@@ -245,6 +261,28 @@ internal class StateAndEventSubscriptionImpl<K : Any, S : Any, E : Any>(
...
+                        .uri(URI.create("http://localhost:7000/uniquenessChecker"))
+                        .headers("Content-Type", "application/octet-stream")
+                        .POST(HttpRequest.BodyPublishers.ofByteArray(payload))
+                        .build()
+                    val response = httpClient.send(request, HttpResponse.BodyHandlers.ofByteArray())
+                    val responseBody: ByteArray = response.body()
+                    @Suppress("UNCHECKED_CAST")
+                    val responseEvent = avroDeserializer.deserialize(responseBody) as? E
+                    log.info("Got a response: $responseEvent")
+                    Record(record.topic, record.key, responseEvent)
```
emilybowe: Think this has to be Flow Event topic possibly? Miljenko made a note but may not have pushed it

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
```diff
@@ -40,25 +39,19 @@ class JavalinServer(
 
     override fun start(port: Int) {
         if (server != null) {
```
mbrkic-r3: Might use `check()` instead

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
```diff
@@ -74,42 +67,60 @@ class JavalinServer(
...
-    override fun stop() {
-        coordinator.updateStatus(LifecycleStatus.DOWN)
+    private fun stopServer() {
         server?.stop()
         server = null
+    }
+
+    private fun restartServer() {
+        // restart server without marking the component down.
+        val port = server?.port()?:throw java.lang.IllegalStateException("Cannot restart a non-existing server")
```
mbrkic-r3: It might be easier to read if `check()` is performed above with `server` instead of `port`

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
```diff
@@ -74,42 +67,60 @@ class JavalinServer(
...
+    override fun stop() {
+        coordinator.updateStatus(LifecycleStatus.DOWN)
+        stopServer()
         coordinator.stop()
     }
 
     override fun registerEndpoint(endpoint: Endpoint) {
-        registerEndpointInternal(endpoint)
+        // register immediately when the server has been started
+        if(null != server) registerEndpointInternal(endpoint)
```
mbrkic-r3, line 107: We should see whether we can enforce formatting / reformat code automatically after it's pushed so we have consistent formatting across the codebase (in this file it's not consistent space between `if` and `(`, `server != null` vs `null != server`).

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
```diff
@@ -40,25 +39,19 @@ class JavalinServer(
 
     override fun start(port: Int) {
         if (server != null) {
-            throw CordaRuntimeException("The Javalin webserver is already initialized")
+            throw IllegalStateException("The Javalin webserver is already initialized")
```
mbrkic-r3: Tests that check for this exception need to be updated

driessamyn: Y. This is WIP, just trying to get to the point the thing actually starts.
I suspect a lot of tests will need to be updated, and some added as there must have been quite a lot of gaps.

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
driessamyn: Simon has previously suggested we introduce an automatic linter. I'm fully supportive.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/utils/UniquenessAssertions.kt
```diff
@@ -0,0 +1,36 @@
...
+import net.corda.test.util.time.AutoTickTestClock
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Assertions.assertInstanceOf
+import java.time.Instant
+
+/**
+ * Uniqueness check related assertions for use in tests
+ */
+@Suppress("SpreadOperator", "TooManyFunctions")
+object UniquenessAssertions {
```
mbrkic-r3: We don't need to copy this, we can add it as dependency instead:
`smokeTestImplementation project(":testing:uniqueness:uniqueness-utilities")`

applications/workers/workers-smoketest/build.gradle
```diff
@@ -114,25 +122,29 @@ tasks.register('smokeTest', Test) {
     testClassesDirs = project.sourceSets["smokeTest"].output.classesDirs
     classpath = project.sourceSets["smokeTest"].runtimeClasspath
 
-    def combinedWorker = project.getProperties().getOrDefault("isCombinedWorker",false)
+    // TODO - switch default around so this can be run locally.
+    //  Also refactor property names to remove "health" now that they are more general purpose.
```
driessamyn, line 126: This will be addressed in a separate PR (CORE-17202).

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
```diff
@@ -245,6 +261,32 @@ internal class StateAndEventSubscriptionImpl<K : Any, S : Any, E : Any>(
 
         commitTimer.recordCallable {
             producer.beginTransaction()
+
+            // HACK for testing only
```
relyafi: This looks odd, this isn't test specific code. If this is a temporary change needed to facilitate our current implementation, I'd leave it as a TODO / FIXME with an associated Jira ticket reference that will implement an improved solution.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/UniquenessChecker.kt
```diff
@@ -0,0 +1,186 @@
...
+import java.time.Instant
+import java.time.LocalDate
+import java.time.ZoneOffset
+import java.util.UUID
+
+/**
+ * Tests for the UniquenessChecker RPC service
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class UniquenessChecker {
```
relyafi: This feels a bit too broad a name and doesn't follow the naming precedent of other smoke test classes. I'd suggest something like `UniquenessCheckerRPCTests` or `UniquenessCheckerRPCSmokeTests`.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/UniquenessChecker.kt
```diff
@@ -0,0 +1,186 @@
...
+        registerStaticMember(bobHoldingId)
+        registerStaticMember(charlieHoldingId)
+        registerStaticMember(notaryHoldingId, NOTARY_SERVICE_X500)
+    }
+
+    @Test
+    fun `when call service with valid payload return idempotently`() {
+        val url = "${System.getProperty("uniquenessWorkerHealthHttp")}api/5.1/uniqueness-checker"
+
+        logger.warn("uniqueness url: $url")
```
relyafi: If this is needed at all, shouldn't it be on `info` not `warn`?

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/UniquenessChecker.kt
```diff
@@ -0,0 +1,186 @@
...
+        assertThat(notaryActualHoldingId).isEqualTo(notaryHoldingId)
+
+        registerStaticMember(aliceHoldingId)
+        registerStaticMember(bobHoldingId)
+        registerStaticMember(charlieHoldingId)
+        registerStaticMember(notaryHoldingId, NOTARY_SERVICE_X500)
+    }
+
+    @Test
+    fun `when call service with valid payload return idempotently`() {
```
relyafi: I'm not sure this test case is proving what it states? We're only sending the request once, so isn't this just proving a trivial success case? I think you want to send twice and verify the exact same response if you want to prove idempotency.

I'd also recommend a second test scenario triggering a failure scenario and proving idempotency here too, since the failure code path through the uniqueness checker is different from the success case. A good way of doing this might be to send a request with a very near time window upper bound, increment the clock past this, and then send the exact same response again. A "new" request would fail the check, but the old one should simply retrieve the original result. It should be possible to manipulate the clock to do this given we're using a test clock.

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/UniquenessCheckRpcMessageProcessor.kt
```diff
@@ -0,0 +1,38 @@
...
+ * events response API.
+ */
+class UniquenessCheckRpcMessageProcessor(
+    private val uniquenessChecker: UniquenessChecker,
+    private val externalEventResponseFactory: ExternalEventResponseFactory,
+    override val requestClass: Class<UniquenessCheckRequestAvro>,
+    override val responseClass: Class<FlowEvent>,
+) : SyncRPCProcessor<UniquenessCheckRequestAvro, FlowEvent> {
+
+    override fun process(request: UniquenessCheckRequestAvro): FlowEvent {
```
relyafi, line 21: This implies that we are artificially limiting the RPC interface to process a single request at a time at present. But is there any value in doing this? I appreciate that for the time being, RPC calls will be synchronous and only issue one request at a time, but we could have the server side support a list of requests, and just get the client to invoke with a single request at a time?

This would save us having to revisit this interface and having to worry about backwards compatibility in future.

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/UniquenessCheckRpcMessageProcessor.kt
```diff
@@ -0,0 +1,38 @@
...
+        val result = uniquenessChecker.processRequests(listOf(request))
```
relyafi: This could be inlined on like 24 as this is only used once.

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/UniquenessCheckRpcMessageProcessor.kt
```diff
@@ -0,0 +1,38 @@
...
+
+        return result.map { (request, response) ->
+            if (response.result is UniquenessCheckResultUnhandledExceptionAvro) {
+
+                externalEventResponseFactory.platformError(
+                    request.flowExternalEventContext,
+                    (response.result as UniquenessCheckResultUnhandledExceptionAvro).exception
+                )
+
+
```
relyafi: Redundant whitespace

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/UniquenessCheckRpcMessageProcessor.kt
relyafi, line 21: This function appears to be missing the tracing logic that is present in the (Kafka based) `UniquenessCheckMessageProcessor`. Are we intentionally considering tracing out of scope for now, and if so, should we add a TODO / FIXME with a Jira ticket that will re-introduce this functionality?

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/UniquenessCheckRpcMessageProcessor.kt
```diff
@@ -0,0 +1,38 @@
...
+            } else {
+                externalEventResponseFactory.success(request.flowExternalEventContext, response)
+            }
+        }.first().value!!
```
relyafi: If we do adopt my earlier suggestion then this will need to return all results. Otherwise, this code block should probably enforce exactly one result via `single()`, either here or on line 24.

components/uniqueness/uniqueness-checker-impl/src/test/kotlin/net/corda/uniqueness/checker/impl/UniquenessCheckRpcMessageProcessorTests.kt
```diff
@@ -0,0 +1,82 @@
...
+import net.corda.messaging.api.records.Record
+import net.corda.uniqueness.checker.UniquenessChecker
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.mockito.kotlin.any
+import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.verify
+
+class UniquenessCheckRpcMessageProcessorTests {
```
relyafi, line 19: I'd suggest that the test cases in this suite should include adaptations of those in `MessageBusIntegrationTests`, as this is effectively testing at the same level as those tests. This would give us strong confidence that the new interface behaves identically to the old one. This will likely need doing anyway at the point the Kafka based message bus implementation is removed, as we won't want to lose the associated test cases.

relyafi (commented):
Mostly minor comments. As mentioned in one of the specific comments, I'd strongly consider making the server API support a batch of requests, even if the client will only send a single request at a time at present.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/UniquenessChecker.kt
```diff
@@ -0,0 +1,186 @@
...
+        assertThat(deserializedExternalEventResponse).isNotNull
+        UniquenessAssertions.assertStandardSuccessResponse(deserializedExternalEventResponse!!, testClock)
+    }
+
+    private val testClock = AutoTickTestClock(Instant.MAX, Duration.ofSeconds(1))
+
+    /**
+     * Returns a random set of bytes
+     */
+    private fun randomBytes(): ByteArray {
```
mbrkic-r3: Might be replaced with `Random.nextBytes(16)`?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/StateAndEventSubscriptionImpl.kt
mbrkic-r3: These changes were not supposed to get merged, they need to be removed.

applications/workers/workers-smoketest/build.gradle
```diff
@@ -114,25 +122,29 @@ tasks.register('smokeTest', Test) {
...
+            project.getProperties().getOrDefault("cryptoWorkerHealthHttp", combinedWorker ? combinedWorkerAddress : "http://localhost:7001/")
     systemProperty "restWorkerHealthHttp",
-            project.getProperties().getOrDefault("restWorkerHealthHttp",combinedWorker ? null : "http://localhost:7002/")
+            project.getProperties().getOrDefault("restWorkerHealthHttp", combinedWorker ? combinedWorkerAddress : "http://localhost:7002/")
     systemProperty "flowWorkerHealthHttp",
-            project.getProperties().getOrDefault("flowWorkerHealthHttp",combinedWorker ? null : "http://localhost:7003/")
+            project.getProperties().getOrDefault("flowWorkerHealthHttp", combinedWorker ? combinedWorkerAddress : "http://localhost:7003/")
+    systemProperty "flowMapperWorkerHealthHttp",
+            project.getProperties().getOrDefault("flowMapperWorkerHealthHttp", combinedWorker ? combinedWorkerAddress : "http://localhost:7004/")
+    systemProperty "flowMapperWorkerHealthHttp",
```
emilybowe: think this should be `systemProperty verificationWorkerHealthHttp`

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/UniquenessCheckRpcMessageProcessor.kt
mbrkic-r3: Tracing was moved to `SyncRPCSubscriptionImpl.kt`

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/UniquenessChecker.kt
```diff
@@ -0,0 +1,186 @@
...
+
+        registerStaticMember(aliceHoldingId)
+        registerStaticMember(bobHoldingId)
+        registerStaticMember(charlieHoldingId)
+        registerStaticMember(notaryHoldingId, NOTARY_SERVICE_X500)
+    }
+
+    @Test
+    fun `when call service with valid payload return idempotently`() {
+        val url = "${System.getProperty("uniquenessWorkerHealthHttp")}api/5.1/uniqueness-checker"
```
mbrkic-r3: Should we get the version from `PlatformInfoProvider.kt`?

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/UniquenessChecker.kt
mbrkic-r3: The main idea of this test was just to verify that RPC endpoint is working, that we are accepting requests and returning back responses. So I think we should just rename the test. Client side is not finished yet, and whole RPC stuff won't be used until Multisource-Event Mediator is implemented, so I think we don't need to test everything/whole Uniqueness functionality at this time since RPC endpoint won't be used yet.

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/UniquenessCheckRpcMessageProcessor.kt
mbrkic-r3: This could definitely be a future improvement, but I would leave it for another PR, this one is big enough already.

relyafi: Is it a big change? I'd just be worried if we do that then it'll fall into the category of change that will never be considered important enough to do, until we need it, which will then likely require an API version upgrade. Arguably the batch support is less of a change from what we already have.

components/uniqueness/uniqueness-checker-impl/src/test/kotlin/net/corda/uniqueness/checker/impl/UniquenessCheckRpcMessageProcessorTests.kt
mbrkic-r3: This is just a unit test (with mocks) for `UniquenessCheckRpcMessageProcessor`. Tests similar to those in `MessageBusIntegrationTests` should probably also be integration tests, e.g. `UniquenessRpcIntegrationTests` (that replaces `MessageBusIntegrationTests` once we are ready for that).

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/UniquenessCheckRpcMessageProcessor.kt
mbrkic-r3: I meant that PR is big enough already to add additional methods and tests. If we just want to replace existing method, it might be a small change, and it might make sense to do it now if that would otherwise be a backward-incompatible change in the future. For initial version of Multi-source Event Mediator we didn't plan to batch RPC requests.

relyafi: I think that would be worth investigating, as we already have batching support available within the existing uniqueness checker functionality so it wouldn't be too onerous to do a quick sanity check the batch interface works. Understand it won't be used by the client in the first instance (i.e. it would always use a batch size of 1).

emilybowe: I've created a ticket to track it https://r3-cev.atlassian.net/browse/CORE-17182

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/UniquenessChecker.kt
mbrkic-r3: Do we still use the same port for health and other functionality? I see [this](https://github.com/corda/corda-runtime-os/pull/4470#discussion_r1311559501) comment is still not resolved?

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/UniquenessCheckRpcMessageProcessor.kt
mbrkic-r3: @Omar-awad @driessamyn Please advise.

ronanbrowne (dismissed):
LGTM - build files only reviewed

dimosr (dismissed):
approving for @corda/corda-platform-network-team (only import changes)

josephzunigadaly (dismissed):
Approving for REST owned files only

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/UniquenessChecker.kt
emilybowe: for comment 1, discussed offline as this would need to be injected as a service it looks like and might be non-trivial. Smaller improvement added for now to use a constant in utils with ticket created for later https://r3-cev.atlassian.net/browse/CORE-17187

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/UniquenessCheckRpcMessageProcessor.kt
driessamyn: It's valid feedback from @relyafi ,  but I agree that deferring this change is the right thing to do. 
Personally, I think we should have a think about which batching implementations work best. Batching could, for example, happen completely within the Uniqueness Service, i.e. it could batch up requests to process together, rather, or as well as, implementing this on the API.
We should test and validate this I think before doing the implementation.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/UniquenessChecker.kt
driessamyn: > Do we still use the same port for health and other functionality? I see [this](https://github.com/corda/corda-runtime-os/pull/4470#discussion_r1311559501) comment is still not resolved?

Yes, the decision was to prefer a single web server until there is a reason not to.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/UniquenessChecker.kt
driessamyn: Sorry, that is my fault. I initially added `idempotently` to the test name as I thought it would be worth validating this (and add as a regression).

I agree that we should add additional tests though. Let's raise a JIRA so we backfill the test suite with negative and idempotent tests.

relyafi (approved):
Uniqueness changes LGTM, subject to the topic of batching being revisited as discussed.

components/uniqueness/uniqueness-checker-impl/src/main/kotlin/net/corda/uniqueness/checker/impl/UniquenessCheckRpcMessageProcessor.kt
emilybowe: great thanks. I've linked this thread and the message further down where @relyafi mentions changing single() to returning all results in the [ticket](https://r3-cev.atlassian.net/browse/CORE-17182 ) for this. I'll mark this resolved if no one objects.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/UniquenessChecker.kt
emilybowe: great thanks for clarity on comment 2, marking this thread resolved now

mbrkic-r3 (approved):
LGTM

ronanbrowne (approved):

dimosr (approved):
re-approving for @corda/corda-platform-network-team files

josephzunigadaly (approved):
Re-approve for REST owned files

PR #4635, opened 2023-09-13
CORE-16330 Scheduler fixes

kyriathar:
Adding a topic for the `Scheduler` allowed us to go past the error: "_net.corda.messaging.api.exception.CordaMessageAPIFatalException: Cannot find topic X_" and then revealed the below issues:

1. we were not using `SchedulerLock.updateAndRelease` which also commits the scheduler lock's DB transaction and so we would get: "_Caused by: java.sql.SQLTransientConnectionException: HikariP not available, request timed out after X ms._" after a while, because we were closing the entity manager without committing the transaction first.
2. we were not persisting the scheduled task to `task_scheduler_log` DB table. 

* changes `SchedulerLock.updateAndRelease` to only update the DB log entry and moves the releasing of the DB log entry (committing the DB transaction) to the `SchedulerLock.close` method.
* persists the scheduled task to `task_scheduler_log` DB table.

### DIFF
--- a/components/scheduler/src/main/kotlin/net/corda/components/scheduler/impl/SchedulerEventHandler.kt
+++ b/components/scheduler/src/main/kotlin/net/corda/components/scheduler/impl/SchedulerEventHandler.kt
@@ -63,13 +63,15 @@ class SchedulerEventHandler(
     }
 
     private fun triggerAndScheduleNext(coordinator: LifecycleCoordinator) = try {
-        schedulerLog.getLastTriggerAndLock(schedule.taskName, schedulerName).use {
-            if (it.secondsSinceLastScheduledTrigger >= schedule.scheduleIntervalInSeconds) {
+        schedulerLog.getLastTriggerAndLock(schedule.taskName, schedulerName).use { schedulerLock ->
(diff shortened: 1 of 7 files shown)


components/scheduler/src/main/kotlin/net/corda/components/scheduler/impl/SchedulerEventHandler.kt
```diff
@@ -63,13 +63,15 @@ class SchedulerEventHandler(
     }
 
     private fun triggerAndScheduleNext(coordinator: LifecycleCoordinator) = try {
-        schedulerLog.getLastTriggerAndLock(schedule.taskName, schedulerName).use {
-            if (it.secondsSinceLastScheduledTrigger >= schedule.scheduleIntervalInSeconds) {
+        schedulerLog.getLastTriggerAndLock(schedule.taskName, schedulerName).use { schedulerLock ->
```
simon-johnson-r3, line 66: This `use` suggests `close` is called on `schedulerLock`? In that case it would seem to be reasonable to make close responsible for release and make `updateAndRelease` just `update`?

driessamyn (approved):
LGTM

PR #4640, opened 2023-09-14
CORE-14624: Made helm charts more flexible to allow for enterprise configuration more easily.

malachyb:
Do not propagate config salt and passphrase when using Vault.
 
Split out setting crypto DB username and password, so that if a consuming chart is passing password via Vault they do not need to include the password unnecesarily in the environment.

### DIFF
--- a/charts/corda-lib/templates/_bootstrap.tpl
+++ b/charts/corda-lib/templates/_bootstrap.tpl
@@ -256,7 +256,8 @@ spec:
             {{- include "corda.rbacDbUserEnv" . | nindent 12 }}
             {{- include "corda.clusterDbEnv" . | nindent 12 }}
             {{- include "corda.restApiAdminSecretEnv" . | nindent 12 }}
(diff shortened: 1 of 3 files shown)


davidcurrie (commented):
It is hard to comment without seeing the corresponding changes proposed for Enterprise but are the differences really conditional on Enterprise vs OS or are they conditional on the use of Vault? Enterprise != Vault

dickon:
> It is hard to comment without seeing the corresponding changes proposed for Enterprise but are the differences really conditional on Enterprise vs OS or are they conditional on the use of Vault? Enterprise != Vault

So that's the deep issue. My thinking was that for now there's no reason to use Vault without Enterprise, and we may never support Enterprise without Vault. However, I think is marginal, and we could well make the flag to be about Vault. The downside of that we might have lots of flags in long run.

I can walk you through the Enterprise side changes.

davidcurrie (approved):
LGTM

PR #4644, opened 2023-09-14
CORE-17058 - Add minVersion for network endpoints introduced in 5.1

dimosr:
Setting `minVersion` to 5.1 for network endpoints that were introduced in 5.1 and didn't have one, so they were falling back to the default (5.0).

In order to identify the endpoints, I did a diff between the v1 baseline files in the branches `release/os/5.0` and `release/os/5.1`


williamvigorr3 (approved):
LGTM - Checked the endpoints were added into 5.1 and not 5.0. Using git blame and checking the Jiras (CORE-9519 + CORE-12417).

josephzunigadaly (approved):

PR #4645, opened 2023-09-14
CORE-17114 Pre-install check improvements

davidcurrie:
A few improvements to the pre-install check:

- Handle `null` when looking up secret
- Output single-line resource check info messages
- Always check both memory and CPU resources
- Update developer `values.yaml` and `values-prereqs.yaml` to disable the preinstall check as they do not specify resources (as is the preference in a low resource environment)


jacob-scott3141 (approved):
LGTM

PR #4646, opened 2023-09-14
CORE-16181 Initial commit to show PoC

ben-millar:


libs/messaging/message-bus/build.gradle
```diff
@@ -9,9 +9,21 @@ dependencies {
 
     implementation platform("net.corda:corda-api:$cordaApiVersion")
     implementation 'org.jetbrains.kotlin:kotlin-osgi-bundle'
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:$kotlinCoroutinesVersion"
```
chrisr3: This should be `compileOnly` because you are absorbing the jar's contents and so neither need nor want the original jar to be present at runtime.

libs/messaging/messaging-impl/build.gradle
```diff
@@ -31,6 +31,7 @@ dependencies {
     implementation 'net.corda:corda-topic-schema'
     implementation 'net.corda:corda-config-schema'
     implementation 'org.jetbrains.kotlin:kotlin-osgi-bundle'
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:$kotlinCoroutinesVersion"
```
chrisr3: Again, you don't want the original `kotlinx-coroutines-core-jvm` to be present at runtime because you are absorbing Kotlin Coroutines into `:libs:messaging:message-bus`.

Gradle's `implementation` scope puts this dependency on this module's compile classpath, and on the runtime classpath for anything that depends on this module. Since `kotlinx-coroutines-core-jvm` has no OSGi metadata of its own, it serves no purpose inside the workers other than bloat.

libs/messaging/message-bus/build.gradle
```diff
@@ -9,9 +9,21 @@ dependencies {
...
+tasks.named('jar', Jar) {
+    bundle {
+        bnd '''\
+-conditionalpackage: kotlinx.coroutines*
+Export-Package: kotlinx.coroutines*
+Import-Package: !android*, *
+'''
+    }
+}
+
```
chrisr3: This will only absorb the packages matching `kotlinx.coroutines*` that Bnd knows that `message-bus` _itself_ will need. (And those that _those_ packages need, etc.) It also doesn't address any _potential_ Intellectual Property concerns involved with stuffing JetBrains' copyrighted classes into `message-bus` - you still need to check that the licence terms allow you to do this.

Repackaging Kotlin Coroutines into its own bundle would allow other code to use it without becoming entangled with `message-bus`, and so would be a better approach.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/producer/RPCMessageProducerImpl.kt
```diff
@@ -0,0 +1,63 @@
...
+import kotlinx.coroutines.async
+import kotlinx.coroutines.delay
+import net.corda.data.messaging.RPCResponse
+import net.corda.messagebus.api.producer.CordaMessage
+import net.corda.messagebus.api.producer.MessageProducer
+import net.corda.metrics.CordaMetrics
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+class RPCMessageProducerImpl(
```
chrisr3: You need to prove that this works as expected inside an OSGi framework. How are you doing this?

chrisr3 (commented):
For your POC, you also need to show that Kotlin Coroutines still work as expected when used _inside an OSGi framework_.

libs/messaging/message-bus/build.gradle
ben-millar: It's covered under the `Apache License 2.0`: https://github.com/Kotlin/kotlinx.coroutines/blob/master/LICENSE.txt

PR #4648, opened 2023-09-14
CORE-17074: Include JetBrains attribution for kotlin-reflection library.

chrisr3:
Include attribution for the `kotlinx-metadata-jvm` library into the published POM for `kotlin-reflection`.

This library uses Apache Licence 2.0.


ronanbrowne (approved):
LGTM

PR #4650, opened 2023-09-14
CORE-17115: Send error messages back to the counterparty when data is received and the mapper state is null

JamesHR3:
**Problem statement**

The flow mapper may receive a session event from a counterparty containing just data with no init information. If this happens and the mapper does not know about the session, then this is an error that should be returned to the counterparty. However, at the moment this error is forwarded directly to the local flow engine.

**Solution**

Add a method to the `RecordFactory` to send errors back, rather than forwarding them onwards, and invoke that from the session event handler in this situation. This keeps the logic for generating records to send in the `RecordFactory`, while still ensuring that the session error is sent in the correct direction.

**Testing**

- New unit and integration tests.

### DIFF
--- a/components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/RecordFactoryImpl.kt
+++ b/components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/RecordFactoryImpl.kt
@@ -38,14 +38,14 @@ class RecordFactoryImpl @Activate constructor(
     private val sessionEventSerializer = cordaAvroSerializationFactory.createAvroSerializer<SessionEvent> { }
 
     override fun forwardError(
-        sessionEvent: SessionEvent,
+        sourceEvent: SessionEvent,
         exceptionEnvelope: ExceptionEnvelope,
         instant: Instant,
         flowConfig: SmartConfig,
         flowId: String
     ): Record<*, *> {
         return buildSessionRecord(
-            sessionEvent,
+            sourceEvent,
             SessionError(
                 exceptionEnvelope
             ),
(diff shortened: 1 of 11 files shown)


components/flow/flow-mapper-impl/src/test/kotlin/net/corda/flow/mapper/impl/RecordFactoryImplTest.kt
```diff
@@ -164,9 +174,12 @@ internal class RecordFactoryImplTest {
 
     @Test
     fun `forwardEvent returns record for same cluster`() {
+        val timestamp = Instant.now()
         val sessionEvent = SessionEvent(
             MessageDirection.OUTBOUND,
-            Instant.now(), SESSION_ID, 1,
+            timestamp,
+            SESSION_ID,
+            1,
```
emilybowe: I would change this to a SEQUENCE_NUMBER and there's a few places where the msg refers to the session ID being 1 but I think it's the seq number?

emilybowe (dismissed):
LGTM

components/flow/flow-mapper-impl/src/main/kotlin/net/corda/flow/mapper/impl/RecordFactoryImpl.kt
```diff
@@ -112,6 +137,10 @@ class RecordFactoryImpl @Activate constructor(
         )
         return when (outputTopic) {
             Schemas.Flow.FLOW_EVENT_TOPIC -> {
+                if (flowId == null) {
+                    throw IllegalArgumentException("Flow ID is required to forward an event back to the flow event" +
```
sprintengine, line 141: What handles this exception, or will it bring down entire subscription and worker? A quick pass looks like it gets caught in the S&E subscription when tryProcessBatchOfEvents is called, and throws `CordaMessageAPIFatalException`, and takes down the worker?

```
                        throw CordaMessageAPIFatalException(
                            "Failed to process records from topic $eventTopic, group ${config.group}, " +
                                    "producerClientId ${config.clientId}. " +
                                    "Fatal error occurred.", ex
                        )
```

JamesHR3: Good point - I've updated the `SessionEventExecutor` to catch this and log a warning.

conal-smith (approved):
LGTM

PR #4652, opened 2023-09-14
CORE-15820 - Added the tests so deprecated warnings are ignored

filipesoliveira:
This change is required to be able to merge the changes in the API that made `net.corda.v5.ledger.utxo.observer.UtxoLedgerTokenStateObserver` deprecated. Keep in mind that warnings are treated as errors which aborts the build if deprecated code is used.

### DIFF
--- a/components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
+++ b/components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
@@ -26,7 +26,6 @@ import net.corda.sandboxgroupcontext.service.registerCustomJsonSerializers
 import net.corda.utilities.debug
 import net.corda.v5.crypto.SecureHash
 import net.corda.v5.ledger.utxo.ContractState
-import net.corda.v5.ledger.utxo.observer.UtxoLedgerTokenStateObserver
 import net.corda.v5.ledger.utxo.query.VaultNamedQueryFactory
 import net.corda.v5.ledger.utxo.query.json.ContractStateVaultJsonFactory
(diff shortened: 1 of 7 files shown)


owenstanford (approved):
LGTM

components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/internal/EntitySandboxServiceImpl.kt
```diff
@@ -189,9 +188,12 @@ class EntitySandboxServiceImpl @Activate constructor(
         }
     }
 
+    @Suppress("DEPRECATION")
     private fun requireSingleObserverToState(
-        tokenStateObserverMap: Map<Class<ContractState>, List<UtxoLedgerTokenStateObserver<ContractState>>>
-    ): Map<Class<ContractState>, UtxoLedgerTokenStateObserver<ContractState>?> {
+        tokenStateObserverMap: Map<Class<
+                ContractState>,
+                List<net.corda.v5.ledger.utxo.observer.UtxoLedgerTokenStateObserver<ContractState>>>
```
kyriathar, line 195: minor: this bit could do with some better formatting I believe.

lankydan (approved):
Changes look fine but the PR title makes it sound like its doing something else than adding `@Supress` and moving the package names to be inline.

kyriathar (approved):
OK(db changes). A minor comment only but can be addressed later.

chrisr3:
I'm almost afraid to ask, but.... What is the policy for deprecating API classes from Corda 5.0? How long do we guarantee that CorDapps compiled for earlier version of Corda will continue to work?

filipesoliveira:
THIS PR IS NO LONGER NEEDED!

PR #4653, opened 2023-09-15
CORE-13620 Idempotency id table housekeeping

kyriathar:
Introduces housekeeping functionality for the deduplication table as a `Scheduler` task.

- adds `DELETE` query for "old" request ids.
- adds clean up functionality in a Kafka consumer (`DurableProcessor`) to consume from the `Scheduler`.
- adds the `Scheduler` to the `DBWorker`.

**Testing:**
- the `DELETE` query was tested it works against both in memory and postgres DBs at integration level.
- the cluster was deployed and vnodes were created then db-worker logs were monitored to witness the execution of the scheduled task and also monitored for potential errors.

api PR: https://github.com/corda/corda-api/pull/1249

### DIFF
--- a/libs/virtual-node/virtual-node-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/RequestsIdsRepositoryTest.kt
+++ b/libs/virtual-node/virtual-node-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/RequestsIdsRepositoryTest.kt
@@ -8,11 +8,11 @@ import net.corda.libs.virtualnode.datamodel.repository.RequestsIdsRepository
 import net.corda.libs.virtualnode.datamodel.repository.RequestsIdsRepositoryImpl
 import net.corda.orm.impl.EntityManagerFactoryFactoryImpl
 import net.corda.orm.utils.transaction
-import net.corda.orm.utils.use
 import org.junit.jupiter.api.AfterAll
 import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.TestInstance
 import java.util.UUID
@@ -82,6 +82,8 @@ class RequestsIdsRepositoryTest {
             requestsIdsRepository.persist(requestId1, em)
         }
 
+        Thread.sleep(1)
+
         entityManagerFactory.createEntityManager().transaction { em ->
             requestsIdsRepository.persist(requestId2, em)
         }
@@ -95,7 +97,8 @@ class RequestsIdsRepositoryTest {
         assertTrue(request1Time < request2Time)
     }
 
-    // Might want to Disable this test as it adds significant time overhead to the test pipeline
+    @Disabled("Disabling due to its time overhead of 2 seconds. " +
+            "The test, however, is valid to assert `requestsIdsRepository.deleteRequestsOlderThan` works")
     @Test
     fun `deletes older requests`() {
         val requestId1 = UUID.randomUUID()
--- a/libs/virtual-node/virtual-node-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/RequestsIdsRepositoryTest.kt
+++ b/libs/virtual-node/virtual-node-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/RequestsIdsRepositoryTest.kt
@@ -8,7 +8,9 @@ import net.corda.libs.virtualnode.datamodel.repository.RequestsIdsRepository
 import net.corda.libs.virtualnode.datamodel.repository.RequestsIdsRepositoryImpl
 import net.corda.orm.impl.EntityManagerFactoryFactoryImpl
 import net.corda.orm.utils.transaction
+import net.corda.orm.utils.use
 import org.junit.jupiter.api.AfterAll
+import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertTrue
 import org.junit.jupiter.api.Test
@@ -54,6 +56,15 @@ class RequestsIdsRepositoryTest {
     fun cleanup() {
         dbConfig.close()
         entityManagerFactory.close()
+    }
+
+    @AfterEach
+    fun cleanUpAfterEach() {
+        entityManagerFactory.createEntityManager().transaction {
+            it.createNativeQuery(
+                "DELETE FROM ${DbSchema.VNODE_PERSISTENCE_REQUEST_ID_TABLE}"
+            ).executeUpdate()
+        }
     }
 
     private val requestsIdsRepository: RequestsIdsRepository = RequestsIdsRepositoryImpl()
--- a/libs/virtual-node/virtual-node-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/RequestsIdsRepositoryTest.kt
+++ b/libs/virtual-node/virtual-node-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/RequestsIdsRepositoryTest.kt
@@ -79,6 +79,7 @@ class RequestsIdsRepositoryTest {
         assertTrue(request1Time < request2Time)
     }
 
+    // Might want to Disable this test as it adds significant time overhead to the test pipeline
     @Test
     fun `deletes older requests`() {
         val requestId1 = UUID.randomUUID()
--- a/libs/virtual-node/virtual-node-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/RequestsIdsRepositoryTest.kt
+++ b/libs/virtual-node/virtual-node-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/RequestsIdsRepositoryTest.kt
@@ -6,6 +6,7 @@ import net.corda.db.schema.DbSchema
 import net.corda.db.testkit.DbUtils
 import net.corda.libs.virtualnode.datamodel.repository.RequestsIdsRepository
 import net.corda.libs.virtualnode.datamodel.repository.RequestsIdsRepositoryImpl
+import net.corda.orm.EntityManagerConfiguration
 import net.corda.orm.impl.EntityManagerFactoryFactoryImpl
 import net.corda.orm.utils.transaction
 import org.junit.jupiter.api.AfterAll
@@ -21,10 +22,8 @@ import javax.persistence.EntityManagerFactory
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class RequestsIdsRepositoryTest {
 
-    private val dbConfig = run {
-        //System.setProperty("postgresPort", "5432")
-        DbUtils.getEntityManagerConfiguration(this::class.java.simpleName)
-    }
+    @Suppress("JoinDeclarationAndAssignment")
+    private val dbConfig: EntityManagerConfiguration
 
     private val entityManagerFactory: EntityManagerFactory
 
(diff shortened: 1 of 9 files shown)


processors/db-processor/src/main/kotlin/net/corda/processors/db/internal/DBProcessorImpl.kt
```diff
@@ -262,20 +270,40 @@ class DBProcessorImpl @Activate constructor(
...
+                DeduplicationTableCleanUpProcessor(
+                    dbConnectionManager,
+                    virtualNodeInfoReadService,
+                    RequestsIdsRepositoryImpl()
+                ),
+                messagingConfig,
+                null
+            )
+        }.also {
+            it.start()
```
simon-johnson-r3: minor, you don't need an also for a one line operation on `it` you can call `start` directly on the return value

processors/db-processor/src/main/kotlin/net/corda/processors/db/internal/schedule/DeduplicationTableCleanUpProcessor.kt
```diff
@@ -0,0 +1,59 @@
...
+    override val valueClass: Class<ScheduledTaskTrigger>
+        get() = ScheduledTaskTrigger::class.java
+
+    override fun onNext(events: List<Record<String, ScheduledTaskTrigger>>): List<Record<*, *>> {
+        // TODO Add metric around it?
+        log.info("Cleaning up deduplication table for all vnodes")
+        val startTime = System.nanoTime()
+        virtualNodeInfoReadService.getAll().forEach {
+            log.debug { "Cleaning up deduplication table for vnode: ${it.holdingIdentity.shortHash}" }
+            val emf = dbConnectionManager.createEntityManagerFactory(
```
simon-johnson-r3: given that leaving this unclosed would be a bad thing and you can't use it after it's closed I'd run these lines together, e.g.
```
 dbConnectionManager.createEntityManagerFactory(...).use {
}
```
then it can't be misused easily in future changes

processors/db-processor/src/main/kotlin/net/corda/processors/db/internal/schedule/DeduplicationTableCleanUpProcessor.kt
```diff
@@ -0,0 +1,59 @@
...
+                it.vaultDmlConnectionId,
+                // We don't really want to make use of any entities here.
+                object : JpaEntitiesSet {
```
simon-johnson-r3, line 58: I think this illustrates we should create a means to do this without using hibernate!

I mean in the future, not as part of this change

processors/scheduler-processor/src/main/kotlin/net/corda/processors/scheduler/impl/SchedulerProcessorImpl.kt
```diff
@@ -64,6 +65,7 @@ class SchedulerProcessorImpl @Activate constructor(
     // now just hardcoding schedulers here until CORE-16331 is picked up, when we should take this from config
     private val schedules = listOf<Schedule>(
         // example schedule, delete/replace when we have a real one, uncomment for testing
```
simon-johnson-r3: can remove the comments now as per the comments

libs/virtual-node/virtual-node-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/RequestsIdsRepositoryTest.kt
```diff
@@ -0,0 +1,145 @@
...
+            requestsIdsRepository.persist(requestId2, em)
+        }
+
+        val storedRequestIds = getStoredRequestIds()
+        assertEquals(2, storedRequestIds.size)
+        assertEquals(requestId1, storedRequestIds[0].first)
+        assertEquals(requestId2, storedRequestIds[1].first)
+        val request1Time = storedRequestIds[0].second
+        val request2Time = storedRequestIds[1].second
+        assertTrue(request1Time < request2Time)
```
simon-johnson-r3, line 99: I guess this doesn't happen, but if the entities were persisted really quickly this might fail - is it worth introducing a sleep of the minimum unit of change of the db time stamp (e.g. 1 millisecond)?

libs/virtual-node/virtual-node-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/RequestsIdsRepositoryTest.kt
```diff
@@ -0,0 +1,145 @@
...
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+import java.util.UUID
+import javax.persistence.EntityManagerFactory
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class RequestsIdsRepositoryTest {
+
+    private val dbConfig = run {
```
simon-johnson-r3: run isn't doing anything for you here, `this` is the `RequestsIdsRepositoryTest` instance with or without it

libs/virtual-node/virtual-node-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/RequestsIdsRepositoryTest.kt
```diff
@@ -0,0 +1,145 @@
...
+        val storedRequestIds = getStoredRequestIds()
+        assertEquals(2, storedRequestIds.size)
+        assertEquals(requestId1, storedRequestIds[0].first)
+        assertEquals(requestId2, storedRequestIds[1].first)
+        val request1Time = storedRequestIds[0].second
+        val request2Time = storedRequestIds[1].second
+        assertTrue(request1Time < request2Time)
+    }
+
+    // Might want to Disable this test as it adds significant time overhead to the test pipeline
```
simon-johnson-r3: I would definitely do this

processors/scheduler-processor/src/main/kotlin/net/corda/processors/scheduler/impl/SchedulerProcessorImpl.kt
```diff
@@ -64,6 +65,7 @@ class SchedulerProcessorImpl @Activate constructor(
...
+        Schedule("clean-up-deduplication-table", 120, VIRTUAL_NODE_DEDUPLICATION_TABLE_CLEAN_UP_TOPIC)
 //        Schedule("say-hello", 60, "telephone"),
 //        Schedule("say-goodbye", 600, "telephone"),
```
driessamyn: You can remove these as well now. I just left them in there to show how it's used.

libs/virtual-node/virtual-node-datamodel/src/integrationTest/kotlin/net/corda/libs/configuration/datamodel/tests/RequestsIdsRepositoryTest.kt
kyriathar: I just wanted its initialization to be grouped together with `//System.setProperty("postgresPort", "5432")`, indeed I am not making use of `this`.

simon-johnson-r3: `System.setProperty("postgresPort", "5432")` should go in `init` surely?

kyriathar: Done

simon-johnson-r3 (dismissed):

ronanbrowne (dismissed):
lgtm - blt files

processors/db-processor/src/main/kotlin/net/corda/processors/db/internal/schedule/DeduplicationTableCleanUpProcessor.kt
```diff
@@ -0,0 +1,67 @@
...
+                            }
+                        ).use { emf ->
+                            emf.createEntityManager().transaction { em ->
+                                // TODO The below interval needs to be made configurable
+                                requestsIdsRepository.deleteRequestsOlderThan(120, em)
+                            }
+                        }
+                    }
+                val cleanUpTime = Duration.ofNanos(System.nanoTime() - startTime)
+                log.info("Cleaning up deduplication table for all vnodes COMPLETED in ${cleanUpTime.toMillis()} ms")
```
davidcurrie: I'd prefer to see this below `info` level given it's going to be output every 120 seconds.

processors/db-processor/src/main/kotlin/net/corda/processors/db/internal/schedule/DeduplicationTableCleanUpProcessor.kt
```diff
@@ -0,0 +1,67 @@
...
+        // TODO Add metric around it?
+        events
+            .filter {
+                val taskName = it.key
+                taskName == DEDUPLICATION_TABLE_CLEAN_UP_TASK
+            }.forEach { _ ->
+                log.info("Cleaning up deduplication table for all vnodes")
+                val startTime = System.nanoTime()
+                virtualNodeInfoReadService.getAll()
+                    .forEach {
```
davidcurrie: should there be some error handling here? if it fails on one virtual node, should we continue on to others...

kyriathar: Not sure. I think currently if it fails then our message pattern library (whatever calls `DurableProcessor.onNext`) will close the subscription and won't continue to other vnodes. Errors I can think of that could happen here could be unable to connect to the vnode's DB and then something's wrong with the vnode's schema.

processors/db-processor/src/main/kotlin/net/corda/processors/db/internal/schedule/DeduplicationTableCleanUpProcessor.kt
kyriathar: Changed to `debug`

processors/db-processor/src/main/kotlin/net/corda/processors/db/internal/schedule/DeduplicationTableCleanUpProcessor.kt
davidcurrie: In a future where the vnode tables may be in separate DBs, the chances of one failing where others might succeed will increase, so I think we should be catching errors here, logging, and continuing to process. The implication of this is probably that we won't retry until the next scheduled occurrence, but I think that's also okay.

kyriathar: Good catch, will address

processors/db-processor/src/main/kotlin/net/corda/processors/db/internal/schedule/DeduplicationTableCleanUpProcessor.kt
```diff
@@ -0,0 +1,74 @@
...
+                        get() = emptySet()
+                }
+            ).use { emf ->
+                emf.createEntityManager().transaction { em ->
+                    // TODO The below interval needs to be made configurable
+                    requestsIdsRepository.deleteRequestsOlderThan(120, em)
+                }
+            }
+        } catch (e: Exception) {
+            log.info("Cleaning up deduplication table for vnode: ${virtualNodeInfo.holdingIdentity.shortHash} FAILED", e)
```
davidcurrie: should be `error`

kyriathar: Discussed offline. Changed to `warn`.

davidcurrie (dismissed):
LGTM

simon-johnson-r3 (dismissed):

simon-johnson-r3 (approved):

davidcurrie (approved):

ronanbrowne (approved):

PR #4655, opened 2023-09-15
CORE-16615: introduce unconsumed visible state by type query

jennyang-r3:
This PR includes updates such as:
- introduce `findUnconsumedVisibleStatesByExactType` that finds unconsumed states by their exact type unlike `findUnconsumedVisibleStatesByExactType` that finds its subclass typed states.
- filter data by type in db query rather than in-memory filter

**Test Plan**
- Replaced flow`UtxoQueryWithEncumberedStatesFlow` used in `Utxo Ledger - findUnconsumedStatesByType` to use `findUnconsumedStatesByExactType` instead of `findUnconsumedStatesByType`. The test Passes.
- Added `FindUnconsumedStatesByExactTypeExternalEventFactoryTest`

### DIFF
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
@@ -25,9 +25,15 @@ interface UtxoRepository {
         transactionId: String
     ): Map<Int, List<ByteArray>>
 
-    /** Retrieves transaction component leafs related to visible unspent states */
+    /** Retrieves transaction component leafs related to visible unspent states and subclass states */
     fun findUnconsumedVisibleStatesByType(
         entityManager: EntityManager
+    ):  List<UtxoTransactionOutputDto>
+
+    /** Retrieves transaction component leafs related to visible unspent states */
+    fun findUnconsumedVisibleStatesByExactType(
+        entityManager: EntityManager,
+        stateClassType: String
     ):  List<UtxoTransactionOutputDto>
 
     /** Retrieves transaction component leafs related to specific StateRefs */
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
@@ -27,8 +27,7 @@ interface UtxoRepository {
 
     /** Retrieves transaction component leafs related to visible unspent states */
     fun findUnconsumedVisibleStatesByType(
-        entityManager: EntityManager,
-        stateClassType: String
+        entityManager: EntityManager
     ):  List<UtxoTransactionOutputDto>
 
     /** Retrieves transaction component leafs related to specific StateRefs */
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
@@ -25,12 +25,12 @@ interface UtxoRepository {
         transactionId: String
     ): Map<Int, List<ByteArray>>
 
-    /** Retrieves transaction component leafs related to visible unspent states by type and subtypes */
+    /** Retrieves transaction component leaves related to visible unspent states and subclass states.*/
     fun findUnconsumedVisibleStatesByType(
         entityManager: EntityManager
     ):  List<UtxoTransactionOutputDto>
 
-    /** Retrieves transaction component leafs related to visible unspent states */
+    /** Retrieves transaction component leaves related to visible unspent states */
     fun findUnconsumedVisibleStatesByExactType(
         entityManager: EntityManager,
         stateClassType: String
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
@@ -27,7 +27,8 @@ interface UtxoRepository {
 
     /** Retrieves transaction component leafs related to visible unspent states */
     fun findUnconsumedVisibleStatesByType(
-        entityManager: EntityManager
+        entityManager: EntityManager,
+        stateClassType: String
     ):  List<UtxoTransactionOutputDto>
 
     /** Retrieves transaction component leafs related to specific StateRefs */
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
@@ -25,7 +25,7 @@ interface UtxoRepository {
         transactionId: String
     ): Map<Int, List<ByteArray>>
 
-    /** Retrieves transaction component leafs related to visible unspent states and subclass states and subclass states */
+    /** Retrieves transaction component leafs related to visible unspent states by type and subtypes */
     fun findUnconsumedVisibleStatesByType(
         entityManager: EntityManager
     ):  List<UtxoTransactionOutputDto>
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
@@ -25,7 +25,7 @@ interface UtxoRepository {
         transactionId: String
     ): Map<Int, List<ByteArray>>
 
-    /** Retrieves transaction component leafs related to visible unspent states and subclass states */
+    /** Retrieves transaction component leafs related to visible unspent states and subclass states and subclass states */
     fun findUnconsumedVisibleStatesByType(
         entityManager: EntityManager
     ):  List<UtxoTransactionOutputDto>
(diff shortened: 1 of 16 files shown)


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
```diff
@@ -25,11 +25,17 @@ interface UtxoRepository {
         transactionId: String
     ): Map<Int, List<ByteArray>>
 
-    /** Retrieves transaction component leafs related to visible unspent states */
+    /** Retrieves transaction component leafs related to visible unspent states and subclass states */
     fun findUnconsumedVisibleStatesByType(
         entityManager: EntityManager
     ):  List<UtxoTransactionOutputDto>
 
+    /** Retrieves transaction component leafs related to visible unspent states */
```
relyafi:
```suggestion
    /** Retrieves transaction component leafs related to visible unspent states, excluding subclass states */
```

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -46,6 +46,32 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
...
+            JOIN {h-schema}utxo_transaction_output AS tx_o
+                ON tx_o.transaction_id = tc_output.transaction_id
+                AND tx_o.leaf_idx = tc_output.leaf_idx
+            JOIN {h-schema}utxo_transaction_status AS ts
+                ON ts.transaction_id = tx_o.transaction_id
+            WHERE (tx_o.type = :type    
+                AND rts.consumed IS NULL
+                AND ts.status = :verified)
+            ORDER BY tc_output.created, tc_output.transaction_id, tc_output.leaf_idx"""
+            .trimIndent()
```
relyafi, line 73: Observation - there is a lot of duplication here, but on balance I think this way is better than adding complexity trying to share the string between this function and `findUnconsumedVisibleStatesByType`

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
```diff
@@ -106,6 +106,24 @@ class UtxoRepositoryImpl @Activate constructor(
             }
     }
 
+    override fun findUnconsumedVisibleStatesByExactType(
+        entityManager: EntityManager,
+        stateClassType: String
+    ): List<UtxoTransactionOutputDto> {
+        return entityManager.createNativeQuery(queryProvider.findUnconsumedVisibleStatesByExactType, Tuple::class.java)
+            .setParameter("verified", TransactionStatus.VERIFIED.value)
+            .setParameter("type", stateClassType)
```
relyafi: There is a slightly stronger argument to have one function serving both queries at this level, given `stateClassType` could be an optional parameter that when specified triggers the exact type behavior. But I think this current approach is also justifiable.

simulator/runtime/src/main/kotlin/net/corda/simulator/runtime/ledger/utxo/SimUtxoLedgerService.kt
```diff
@@ -129,6 +130,10 @@ class SimUtxoLedgerService(
         return stateAndRefs
     }
 
+    override fun <T : ContractState> findUnconsumedStatesByExactType(type: Class<T>): List<StateAndRef<T>> {
+        TODO("Not implemented yet")
```
relyafi, line 134: Feels like it does need implementing, otherwise this would be a barrier to using the simulator to test a CorDapp that uses this API?

relyafi (commented):
Broadly LGTM, couple of minor comments.

There's quite a bit of code duplication with the existing find function that returns all sub-types, but on balance I think the simplicity of what has been done is better than trying to share code and increase complexity in this instance.

simulator/runtime/src/main/kotlin/net/corda/simulator/runtime/ledger/utxo/SimUtxoLedgerService.kt
jennyang-r3: Not entirely sure but I think I heard a simulator is not used from somewhere. Correct me if I'm wrong. wdyt @vlajos?

relyafi: The issue is there's been a few test frameworks / "simulators", so the question is which one we're talking about! I _think_ this is the latest one, and was only developed a month or two ago.

vlajos: The original method has an implementation in this Sim. So, probably, it is not too complicated to duplicate that following the story of the current changes.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
```diff
@@ -25,11 +25,17 @@ interface UtxoRepository {
         transactionId: String
     ): Map<Int, List<ByteArray>>
 
-    /** Retrieves transaction component leafs related to visible unspent states */
+    /** Retrieves transaction component leafs related to visible unspent states and subclass states and subclass states */
```
blsemo:
```suggestion
    /** Retrieves transaction component leaves related to visible unspent states and subclass states.*/
```

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
```diff
@@ -25,11 +25,17 @@ interface UtxoRepository {
...
     fun findUnconsumedVisibleStatesByType(
         entityManager: EntityManager
     ):  List<UtxoTransactionOutputDto>
 
+    /** Retrieves transaction component leafs related to visible unspent states */
```
blsemo:
```suggestion
    /** Retrieves transaction component leaves related to visible unspent states */
```

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
lankydan: `by type and subtypes` (or subclasses)

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
```diff
@@ -106,6 +106,24 @@ class UtxoRepositoryImpl @Activate constructor(
...
+            .resultListAsTuples()
+            .map { t ->
+                UtxoTransactionOutputDto(
+                    t[0] as String, // transactionId
+                    t[1] as Int, // leaf ID
+                    t[2] as ByteArray, // outputs info data
+                    t[3] as ByteArray // outputs data
+                )
```
lankydan:
```kotlin
.resultListAsTuples()
            .map { t ->
                UtxoTransactionOutputDto(
                    t[0] as String, // transactionId
                    t[1] as Int, // leaf ID
                    t[2] as ByteArray, // outputs info data
                    t[3] as ByteArray // outputs data
                )
            }
```

Shows up a few times, is it worth extracting it out into a method/extension function?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -46,6 +46,32 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
...
+                AND tc_output.leaf_idx = tc_output_info.leaf_idx
+                AND tc_output.group_idx = ${UtxoComponentGroup.OUTPUTS.ordinal}
+            JOIN {h-schema}utxo_transaction_output AS tx_o
+                ON tx_o.transaction_id = tc_output.transaction_id
+                AND tx_o.leaf_idx = tc_output.leaf_idx
+            JOIN {h-schema}utxo_transaction_status AS ts
+                ON ts.transaction_id = tx_o.transaction_id
+            WHERE (tx_o.type = :type    
+                AND rts.consumed IS NULL
+                AND ts.status = :verified)
```
lankydan: Do we need the brackets?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
relyafi: Sounds like we might want something like this: https://github.com/corda/corda-runtime-os/commit/1e0f12237db15ebaf26c0fbd801b52063e41c8b0#diff-3c569291eb64d40d8e699643529119565dbae3a7a1f2e2809ea30cc05fe265e5R71

In which case, wonder if it's worth deferring until we implement paging for this, where we will need resume point information too? Might be the referenced code can be extracted and shared.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoRepositoryImpl.kt
```diff
@@ -90,57 +90,56 @@ class UtxoRepositoryImpl @Activate constructor(
             .mapToComponentGroups(UtxoComponentGroupMapper(transactionId))
     }
 
-    override fun findUnconsumedVisibleStatesByType(
-        entityManager: EntityManager
-    ): List<UtxoTransactionOutputDto> {
-        return entityManager.createNativeQuery(queryProvider.findUnconsumedVisibleStatesByType, Tuple::class.java)
-            .setParameter("verified", TransactionStatus.VERIFIED.value)
-            .resultListAsTuples()
+    private fun mapToUtxoTransactionOutputDto(queryObj: Query): List<UtxoTransactionOutputDto> {
```
lankydan: Can we move this towards the bottom of the class and make it an extension function? I think it makes things easier to read.

simulator/runtime/src/main/kotlin/net/corda/simulator/runtime/ledger/utxo/SimUtxoLedgerService.kt
blsemo: I think not implemented is fine here.

blsemo (commented):
Some kDoc fixes required, but looks fine to me otherwise

lankydan (dismissed):

blsemo (dismissed):
LGTM

blsemo (approved):

PR #4656, opened 2023-09-15
CORE-17123: Fix dependency substitution for jars wrapped as OSGi bundles.

chrisr3:
Ensure that Gradle's dependency substitution selects the `JAR` artifact for each of the modules that wraps a third-party jar as an OSGi bundle.

Also clean up their respective `build.gradle` scripts to make them all consistent.

### DIFF
--- a/buildSrc/src/main/groovy/corda.common-library.gradle
+++ b/buildSrc/src/main/groovy/corda.common-library.gradle
@@ -22,9 +22,21 @@ configurations {
     configureEach {
         resolutionStrategy {
             dependencySubstitution {
-                substitute module('antlr:antlr') using project(':libs:antlr')
-                substitute module('de.javakaffee:kryo-serializers') using project(':libs:serialization:kryo-serializers')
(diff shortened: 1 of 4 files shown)


wzur-r3 (approved):
LGTM

dickon (commented):
beyond my gradle knowledge to know what this stuff is remotely doing.

chrisr3:
> beyond my gradle knowledge to know what this stuff is remotely doing.

The `corda.common-library` plugin replaces references to these third-party jars with a reference to a `project` module instead, where the project module has converted the jar into a working OSGi bundle. However, by default, a `project` dependency is going to refer to its unpackaged `.class` files rather than its final `.jar` file. This is a Bad Thing :tm: for OSGi because the OSGi metadata we need only exists in the jar's manifest, and these particular `project`s don't _have_ any source files in the first place!

I have therefore adjusted the substitution rules also to include the `LibraryElement.JAR` attribute, which prevents Gradle from making this mistake.

simon-johnson-r3 (approved):

kyriathar (approved):

PR #4657, opened 2023-09-15
CORE-16177: Align Kotlin coroutines with other wrapped bundles.

chrisr3:
Ensure that Corda always uses our "wrapped" Kotlin Coroutines OSGi bundle. The smoke tests can continue to use the original unwrapped jar because these run outside an OSGi framework.


wzur-r3 (approved):
LGTM

PR #4662, opened 2023-09-18
CORE-16319 - Fix time sensitive windows nightly build failure

conal-smith:
Testing nightly windows build here - https://ci02.dev.r3.com/job/Corda5/job/Nightlys/job/Corda-Runtime-OS/job/Windows/job/PR-4662/


chrisr3:
`System.currentTimeMillis()` is not suitable for calculating elapsed time accurately.

conal-smith:
> `System.currentTimeMillis()` is not suitable for calculating elapsed time accurately.

Yeah it was in attempt to cut off some of the precision, but not the correct approach. The correct way to resolve this is to fetch the persisted times of the entities we want to use for the interval and use these in the `stateManager.getUpdatedBetween(startTime, finishTime)` API. These start and finish times are inclusive in the query.

JamesHR3 (approved):

PR #4663, opened 2023-09-18
CORE-17173: Improve documentation for SandbxGroupContextServiceImpl.

chrisr3:
Update KDocs for `SandboxGroupContextServiceImpl` and its associated classes.

Also remove some technical debt.

### DIFF
--- a/components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
+++ b/components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
@@ -554,7 +554,7 @@ class SandboxGroupContextServiceImpl @Activate constructor(
 
                 if (!injectable.isByConstructor) {
                     logger.warn("{} must only use constructor injection - IGNORED", injectable)
-                    injectable.broken()
+                    injectable.asBroken()
                 } else if (sandboxRequirements.isNotEmpty()) {
                     // Collect this service's requirements so that
                     // we can examine all requirements afterwards.
@@ -572,7 +572,7 @@ class SandboxGroupContextServiceImpl @Activate constructor(
                         )?.also { svc ->
                             closeables.addFirst(svc)
                             iter.remove()
-                        } ?: run(injectable::broken)
+                        } ?: run(injectable::asBroken)
                     }
                 }
             }
@@ -704,7 +704,7 @@ class SandboxGroupContextServiceImpl @Activate constructor(
                 val nonInjectable = ServiceDefinition(description, serviceFilter).withServiceReferences(serviceIndex)
                 if (!nonInjectable.isByConstructor) {
                     logger.warn("{} must only use constructor injection - IGNORED", nonInjectable)
-                    nonInjectables[serviceRef] = nonInjectable.broken()
+                    nonInjectables[serviceRef] = nonInjectable.asBroken()
                     null
                 } else if (nonInjectable.sandboxReferences.isEmpty() && serviceFilter == null) {
                     // This service doesn't use any of our prototypes, and we don't
@@ -713,7 +713,7 @@ class SandboxGroupContextServiceImpl @Activate constructor(
                     sourceContext.getServiceObjects(serviceRef)
                         ?.let(::registerNonInjectableSandboxService)
                         ?: run {
-                            nonInjectables[serviceRef] = nonInjectable.broken()
+                            nonInjectables[serviceRef] = nonInjectable.asBroken()
                             null
                         }
                 } else {
--- a/components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
+++ b/components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
@@ -240,7 +240,8 @@ class SandboxGroupContextServiceImpl @Activate constructor(
      * the sandbox. We assume that the OSGi isolation hooks protect us from
      * finding any pre-existing services inside the sandbox itself.
      *
-     * Identify which of these services should be registered with the OSGi framework.
+     * Identify which of these services should be registered with the OSGi framework
+     * as "injectable" services, i.e. candidates for `@CordaInject`.
      */
     @Suppress("ComplexMethod")
     private fun createSandboxServiceContext(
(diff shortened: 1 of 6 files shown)


components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
```diff
@@ -240,7 +240,8 @@ class SandboxGroupContextServiceImpl @Activate constructor(
      * the sandbox. We assume that the OSGi isolation hooks protect us from
      * finding any pre-existing services inside the sandbox itself.
      *
-     * Identify which of these services should be registered with the OSGi framework.
+     * Identify which of these services should be registered with the OSGi framework
+     * as "injectable" services.
```
kyriathar: Would it be helpful to elaborate here on which services we consider "injectable" (or non "injectable") for future reference?

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
```diff
@@ -536,6 +540,10 @@ class SandboxGroupContextServiceImpl @Activate constructor(
             }
         }
 
+        /**
+         * Iterate over [injectables], registering those which have no requirements to satisfy.
+         * We will also identify any [nonInjectables] which the [injectables] require.
```
kyriathar, line 545: Might be worth giving having a definiton of `nonInjectables` here

simon-johnson-r3 (dismissed):

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
```diff
@@ -677,25 +701,26 @@ class SandboxGroupContextServiceImpl @Activate constructor(
         private fun addNonInjectable(serviceRef: ServiceReference<*>, closeables: Deque<AutoCloseable>): Boolean {
             var modified = false
             serviceComponentRuntime.getComponentDescriptionDTO(serviceRef)?.let { description ->
-                val nonInjectable = ServiceDefinition(description, serviceFilter).initialise(serviceIndex)
-                if (nonInjectable.sandboxReferences.isEmpty()) {
-                    // This service doesn't use any of our prototypes, which means that
+                val nonInjectable = ServiceDefinition(description, serviceFilter).withServiceReferences(serviceIndex)
+                if (!nonInjectable.isByConstructor) {
+                    logger.warn("{} must only use constructor injection - IGNORED", nonInjectable)
+                    nonInjectables[serviceRef] = nonInjectable.broken()
```
kyriathar: nit: maybe we could rename to `flagBroken` to make a bit more sense?

kyriathar (commented):
Thank you @chrisr3. A few minor comments

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
chrisr3: It _is_ setting a flag, but putting "flag" in the name is _horribly_ reminiscent of MS "Hungarian Notation".

The actual flag is called "`isBroken`". This line is adding a "broken" non-injectable `ServiceDefinition` object into the map.

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
chrisr3: Non-injectable => not suitable for `@CordaInject`

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
chrisr3: "Injectable" -> suitable for `@CordaInject`.

kyriathar (dismissed):

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
simon-johnson-r3: I think the usual convention here would be `setBroken()`?

simon-johnson-r3: I think a more idiomatic Kotlin way of doing this would be:
```
nonInjectables[serviceRef] = nonInjectable.apply {
  broken = true
}
```
but I see the broken function already exists, so it's arguably not part of this PR anyway.

simon-johnson-r3: ah, `broken` is private, possibly then:
```
nonInjectables[serviceRef] = nonInjectable.apply {
  setBroken()
}
```

simon-johnson-r3 (dismissed):

components/virtual-node/sandbox-group-context-service/src/main/kotlin/net/corda/sandboxgroupcontext/service/impl/SandboxGroupContextServiceImpl.kt
chrisr3: I'll use `asBroken()` because it returns itself. A `setXXX()` method would usually return `Unit`.

chrisr3: The point of the `broken()` function is to set the `broken` flag such that it cannot be unset, so it's not a conventional "setter" method.

kyriathar (approved):

PR #4664, opened 2023-09-19
CORE-16620 Implement interop identity delete endpoint

ac101m:
# Changes
- Add delete endpoint
- Add suspend and enable endpoints
- Fix an error with interop identity deletion in the interop processor
- Modify interop identity lookup process to ignore suspended identities


jmacmahonr3:
I think this approach is fine but i'd be eager to see what the the impact on deleting an identity mid flow is ... I imagine we'd break things pretty badly. It's an area we'd want to test

processors/rest-processor/src/integrationTest/resources/swaggerBaseline-v1.json
```diff
@@ -1230,6 +1230,106 @@
         }
       }
     },
+    "/interop/{holdingidentityshorthash}/delete/identity/{interopidentityshorthash}" : {
```
vkolomeyko: New REST APIs should not be added to v1 as this is representation of C5.0 GA release.
I.e. this file should remain unchanged.

Please see my proposed changes to the annotations of newly created methods.

libs/interop/interop-rest/src/main/kotlin/net/corda/libs/interop/endpoints/v1/InteropRestResource.kt
```diff
@@ -49,6 +50,56 @@ interface InteropRestResource : RestResource {
...
 
+    /**
+     * Suspend an interop identity so that it may no longer be used for lookups at the flow level.
+     */
+    @HttpGET(
+        path = "{holdingIdentityShortHash}/suspend/identity/{interopIdentityShortHash}",
+        title = "Suspend interop identity.",
+        description = "Suspends an interop identity. Prevents more lookups at the flow level without" +
+                      "disrupting active flow sessions.",
+        responseDescription = "Response entity with the status of the request."
```
vkolomeyko, line 63:
```suggestion
        responseDescription = "Response entity with the status of the request.",
        minVersion = RestApiVersion.C5_1
```

libs/interop/interop-rest/src/main/kotlin/net/corda/libs/interop/endpoints/v1/InteropRestResource.kt
```diff
@@ -49,6 +50,56 @@ interface InteropRestResource : RestResource {
...
+    ): ResponseEntity<String>
+
+    /**
+     * Enable a previously disabled interop identity so that it may no longer be used for lookups at the flow level.
+     */
+    @HttpGET(
+        path = "{holdingIdentityShortHash}/enable/identity/{interopIdentityShortHash}",
+        title = "Enable interop identity.",
+        description = "Enables a suspended interop identity so that it is available for lookups at the flow level.",
+        responseDescription = "Response entity with the status of the request."
```
vkolomeyko, line 79:
```suggestion
        responseDescription = "Response entity with the status of the request.",
        minVersion = RestApiVersion.C5_1
```

libs/interop/interop-rest/src/main/kotlin/net/corda/libs/interop/endpoints/v1/InteropRestResource.kt
```diff
@@ -49,6 +50,56 @@ interface InteropRestResource : RestResource {
...
+
+    /**
+     * Endpoint to remove an interop identity.
+     * Only suspended identities can be deleted.
+     */
+    @HttpDELETE(
+        path = "{holdingIdentityShortHash}/delete/identity/{interopIdentityShortHash}",
+        title = "Delete a suspended interop identity.",
+        description = "Delete an interop identity.",
+        responseDescription = "Response entity with the status of the request."
```
vkolomeyko, line 96:
```suggestion
        responseDescription = "Response entity with the status of the request.",
        minVersion = RestApiVersion.C5_1
```

libs/interop/interop-rest/src/main/kotlin/net/corda/libs/interop/endpoints/v1/types/InteropIdentityResponse.kt
```diff
@@ -10,5 +10,6 @@ data class InteropIdentityResponse(
     val facadeIds: List<FacadeId>,
     val applicationName: String,
     val endpointUrl: String,
-    val endpointProtocol: String
+    val endpointProtocol: String,
```
vkolomeyko, line 13: Change to this type will have an effect on `v1` REST API. I think we do need to create an almost identical data container for C5.1 onwards.

vkolomeyko (changes requested):

processors/rest-processor/src/integrationTest/resources/swaggerBaseline-v1.json
ac101m: Yup, we've been kicking this can down the road for some time now. We have been adding interop endpoints since before the 5.0GA and have yet to update our feature branch.

I'll apply your suggestions for the new endpoints introduced here, but fixing this issue for our other endpoints is out of scope for this PR. I've created a [separate ticket](https://r3-cev.atlassian.net/browse/CORE-17530) to move all our existing endpoints to 5.1+ and reset the 5.0 baseline to match that in `release/os/5.1`.

libs/interop/interop-rest/src/main/kotlin/net/corda/libs/interop/endpoints/v1/types/InteropIdentityResponse.kt
ac101m: As previously stated, this is going into an interop feature branch where (despite erroneously being added to the 5.0 baseline), all our APIs are unreleased.

processors/rest-processor/src/integrationTest/resources/swaggerBaseline-v1.json
ac101m: On reflection, I think this will be quicker to do all at once. Will raise a separate PR for this.

vkolomeyko (approved):
Fine to merge into team's branch.

components/interop/interop-identity-write-service-impl/src/main/kotlin/net/corda/interop/identity/write/impl/HostedIdentityProducer.kt
```diff
@@ -23,37 +23,58 @@ class HostedIdentityProducer(private val publisher: AtomicReference<Publisher?>)
         private val DUMMY_PUBLIC_SESSION_KEY = this::class.java.getResource("/dummy_session_key.pem")?.readText()
     }
 
+    fun clearHostedInteropIdentity(interopIdentityShortHash: ShortHash) {
+        if (publisher.get() == null) {
```
szymonsztuka: We could use a variable:
```
val pub = publisher.get()
```
And the we don't need to cast to not-null (`!!`):
```
val futures = pub.publish(listOf(record))
```
Same suggestion for all new similar code in this PR.

ac101m: Done

szymonsztuka (approved):

jmacmahonr3 (approved):
LGTM

PR #4666, opened 2023-09-19
CORE-17166 Add cpiSummary.filecheckSum to DigitalSignatureMetadata in TransactionSignatureService

vlajos:
CORE-17166 Adding cpiSummary.filecheckSum to DigitalSignatureMetadata in TransactionSignatureService, because that is also required to identify the running CPI.


relyafi (approved):
Code changes LGTM. Discussed with @vlajos if there are any backwards compatibility concerns with adding the new metadata field (trying to verify a 5.0.x signature on a 5.1 worker) but apparently there shouldn't be any issues with this.

It may be worth adding a test scenario in the E2E test repo to cover this case.

PR #4669, opened 2023-09-19
CORE-17185: Change external event retry to be time based rather than count based

JamesHR3:
**Problem Statement**

External events are used by the flow engine to make requests to other parts of the system in order to perform some operation. For example, if a flow needs to write something to the database, an external event request is generated and sent to the DB worker to do this.

Other services may return a transient error, indicating that the flow engine should retry the request. Previously this retry mechanism was handled by the flow scheduler. When the scheduled event was received by the flow engine, it would re-evaluate the checkpoint and discover there was an external event request being held. If this request was last sent outside of a configurable retry window, the request would be resent. However, the flow scheduler has since been removed, and so at present the retry mechanism does not function correctly.

**Solution**

Resend the external event back to the processing service on receipt of a transient error. By doing this, the external event code is no longer dependent on the scheduler for retries.

In order for this change to function correctly, the retry mechanism needs to change. Previously it used a count-based approach, where after a configurable number of attempts the flow would fail. The new approach to retries cannot work like this as the full count is hit very quickly in low load scenarios (similar to the session layer). To address this, external events are now retried for a configurable amount of time (reusing the existing retry window config). If this window has expired and the flow has resent the external event, the flow will fail on receipt of another failure.

**Testing**

Unit and integration test coverage shows that external events are resent on receiving the response with this change.


components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowGlobalPostProcessorImpl.kt
```diff
@@ -150,12 +150,11 @@ class FlowGlobalPostProcessorImpl @Activate constructor(
      * Check to see if any external events needs to be sent or resent due to no response being received within a given time period.
```
LWogan: kdocs will need adjusting

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
```diff
@@ -154,66 +151,59 @@ class ExternalEventManagerImpl(
...
+                // the retry late. This guarantees an external event will be tried at least twice. After that though,
+                // retrying further is unlikely to clear the problem.
+                externalEventState.retries++
+            }
+            else -> {
+                throw FlowFatalException("External event with request ID ${externalEventState.requestId} exceeded " +
+                        "the retry window.")
+            }
         }
     }
```
ben-millar, line 196: I like the use of the `when` block here, but gating the logic with a 'pitfall' first case feels a little unexpected. Might be more clear to just use an early return, e.g.
```suggestion
    private fun checkRetry(externalEventState: ExternalEventState, instant: Instant, retryWindow: Duration) {
        if ((externalEventState.sendTimestamp + retryWindow) >= instant) return
        
        if (0 == externalEventState.retries) {
            // Retrying once outside the window is required in case the flow engine receives the event to trigger
            // the retry late. This guarantees an external event will be tried at least twice. After that though,
            // retrying further is unlikely to clear the problem.
            externalEventState.retries++
        } else {
            throw FlowFatalException("External event with request ID ${externalEventState.requestId} exceeded " +
                        "the retry window.")
        }
    }
```

ben-millar (approved):
LGTM! One small suggestion, but I'll leave that up to you

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
JamesHR3: It's an interesting one. It's actually not an early return at all, but an explicit return vs an implicit one. Personally I find it easier to reason that all branches have an implicit return in, but I agree having a branch with no code in is a little surprising (hence the comment).

I'm going to leave this for now I think, but thanks for the comment, it's an interesting question!

PR #4673, opened 2023-09-20
CORE-16822 - Token Selection - When a token service event fails it should report an error back to the flow

filipesoliveira:
Before this change, flows would hang whenever there was an issue in the token selection service. This fix ensure that the flow status changes to FAILURE if an error is raised by the token selection service.

### DIFF
--- a/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/TokenCacheEventProcessor.kt
+++ b/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/TokenCacheEventProcessor.kt
@@ -1,8 +1,11 @@
 package net.corda.ledger.utxo.token.cache.services
 
+import net.corda.data.KeyValuePairList
+import net.corda.data.flow.event.external.ExternalEventContext
 import net.corda.data.ledger.utxo.token.selection.event.TokenPoolCacheEvent
 import net.corda.data.ledger.utxo.token.selection.key.TokenPoolCacheKey
 import net.corda.data.ledger.utxo.token.selection.state.TokenPoolCacheState
+import net.corda.flow.external.events.responses.factory.ExternalEventResponseFactory
 import net.corda.ledger.utxo.token.cache.converters.EntityConverter
 import net.corda.ledger.utxo.token.cache.converters.EventConverter
 import net.corda.ledger.utxo.token.cache.entities.TokenEvent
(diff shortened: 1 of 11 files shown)


components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/TokenCacheEventProcessor.kt
```diff
@@ -67,13 +76,20 @@ class TokenCacheEventProcessor constructor(
...
-            return StateAndEventProcessor.Response(state, listOf(), markForDLQ = true)
+            val responseMessage = externalEventResponseFactory.platformError(
+                ExternalEventContext(
+                    tokenEvent.externalEventRequestId,
+                    tokenEvent.flowId,
+                    KeyValuePairList(listOf())
+                ),
+                e
+            )
+            StateAndEventProcessor.Response(state, listOf(responseMessage), markForDLQ = true)
```
owenstanford: markForDLQ should be false

components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/TokenCacheEventProcessor.kt
```diff
@@ -34,9 +38,14 @@ class TokenCacheEventProcessor constructor(
         event: Record<TokenPoolCacheKey, TokenPoolCacheEvent>
     ): StateAndEventProcessor.Response<TokenPoolCacheState> {
 
-        try {
-            val tokenEvent = eventConverter.convert(event.value)
+        val tokenEvent = try {
+            eventConverter.convert(event.value)
+        } catch (e: Exception) {
+            log.error("Unexpected error while processing event '${event}'. The event will be sent to the DLQ.", e)
+            return StateAndEventProcessor.Response(state, listOf(), markForDLQ = true)
```
josephzunigadaly, line 45: In the catch block on line 92 of this file, we have added `responseMessage` as content of the list in the second parameter. Do we need a similar `responseMessage` here?

filipesoliveira: I am not sure. My main reasoning is that if the conversion fails then it is not possible to create the `ExternalEventContext` which prevents the code from creating the response message.

filipesoliveira: The `event` variable which is being converted contains the `flowId` plus the `externalEventRequestId` but these two values are only available after the conversion.

josephzunigadaly (approved):

owenstanford (approved):

PR #4674, opened 2023-09-20
CORE-16233 Implement Service APIs - DBWorker

mbrkic-r3:
There is no processing of `ExternalEvent`s in `DBWorker`. Only reference was due to dependency `components:persistence:entity-processor-service`, which is not needed anymore so it was removed.


ronanbrowne (approved):
Lgtm

PR #4675, opened 2023-09-20
CORE-16796 Print intended Kafka topic configuration 

bpaunescu:
This PR introduces a new topic create sub-command called `generate` which uses the Corda 5 worker-to-topic mappings in combination with user provided kafka permissions to generate and output the desired Kafka configuration in YAML format. This command replaces the `script` command. 

If there's no topic config specified, it is expected that Kafka default will be used. 

Example of output:

```
topics:
  - name: avro.schema
    config:
      cleanup.policy: compact
      segment.ms: 600000
      delete.retention.ms: 300000
      min.compaction.lag.ms: 60000
      max.compaction.lag.ms: 604800000
      min.cleanable.dirty.ratio: 0.5
  - name: certificates.rpc.ops
    config: {}
  - name: certificates.rpc.ops.resp
    config: {}
  - name: config.management.request
    config: {}
  - name: config.management.request.resp
    config: {}
  - name: config.topic
    config:
      cleanup.policy: compact
      segment.ms: 600000
      delete.retention.ms: 300000
      min.compaction.lag.ms: 60000
      max.compaction.lag.ms: 604800000
      min.cleanable.dirty.ratio: 0.5
acls:
  - topic: avro.schema
    users:
      - name: Chris
        operations:
          - read
          - write
          - describe
      - name: Dan
        operations:
          - read
          - write
          - describe
      - name: Fiona
        operations:
          - read
          - write
          - describe
      - name: Mo
        operations:
          - read
          - write
          - describe
  - topic: certificates.rpc.ops
    users:
      - name: Dan
        operations:
          - read
          - describe
  - topic: certificates.rpc.ops.resp
    users:
      - name: Dan
        operations:
          - write
          - describe
  - topic: config.management.request
    users:
      - name: Dan
        operations:
          - read
          - describe
  - topic: config.management.request.resp
    users:
      - name: Dan
        operations:
          - write
          - describe
  - topic: config.topic
    users:
      - name: Chris
        operations:
          - read
          - describe
      - name: Dan
        operations:
          - read
          - write
          - describe
      - name: Fiona
        operations:
          - read
          - describe
      - name: Mo
        operations:
          - read
          - describe
```


tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Generate.kt
```diff
@@ -0,0 +1,48 @@
+package net.corda.cli.plugins.topicconfig
+
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import picocli.CommandLine
+
+@CommandLine.Command(name = "generate", description = ["Generates a textual representation of the intended Kafka topic configuration"])
```
davidcurrie: I am picking here as the description makes it clear, but "generate" on its own could equally be taken to mean that it's actually going to generate the topics/ACLs themselves. Is there a name that would better convey that this isn't actually going to change anything? E.g. "yaml" would be a more direct substitution for the previous "script", or maybe something like Terraform's "plan"?

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Generate.kt
```diff
@@ -0,0 +1,48 @@
...
+class Generate : Runnable {
+
+    @CommandLine.ParentCommand
+    var create: Create? = null
```
davidcurrie: It's now unfortunate that the `bootstrap-server` and `kafka-config` options are defined on the top-level `topic` command as they're not really needed for this sub-command. Given that picocli seems picky about the placement of the arguments, I'm not sure whether we actually want to move them now but maybe they could be made optional on the top-level command and only enforced based on the sub-command that is being used?

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Generate.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    companion object {
+        private val logger: Logger = LoggerFactory.getLogger(this::class.java)
+    }
+
+    override fun run() {
+        val generatedConfig = mutableMapOf<String, MutableList<Any>>()
+        generatedConfig["topics"] = mutableListOf()
+        generatedConfig["acls"] = mutableListOf()
```
davidcurrie: Given one of the things we're expecting the user may wish to do is diff the output with a previous version, is it worth ensuring that the topics/acls are ordered e.g. alphabetically?

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Generate.kt
```diff
@@ -0,0 +1,48 @@
...
+
+        create!!.getTopicConfigs().map { topicConfig ->
+            val topicName = create!!.getTopicName(topicConfig)
+
+            logger.info("Generating configuration for topic $topicName")
+            generatedConfig["topics"]!!.add(mapOf("name" to topicName, "config" to topicConfig.config))
```
davidcurrie: It's a matter of style, but I'd prefer to see the lists generated and then added to `generatedConfig` to avoid the `!!`.

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Generate.kt
```diff
@@ -0,0 +1,48 @@
...
+                    operations.add("write")
+                if (it in usersReadAccess)
+                    operations.add("read")
+                mapOf("name" to it, "operations" to operations.reversed())
+            }
+
+            generatedConfig["acls"]!!.add(acl)
+        }
+
+        println(create!!.mapper.writeValueAsString(generatedConfig))
```
davidcurrie: Given logs are also written to `stdout` by default, how does the user separate the generated config from log entries?

davidcurrie (commented):
I am happy with the output and the logic, but have a few questions about using the command in anger.

davidcurrie:
Looks like detekt is not happy though...

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Generate.kt
bpaunescu: How about `preview`? Also like `plan` if preview doesn't work.

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Generate.kt
bpaunescu: Will have a look and tidy things up a bit.

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Generate.kt
bpaunescu: I've refactored this due to the second half of the epic. I'm using data classes instead. The order is determined by the corda-api topic yaml files and, unless those files are changed, the order will always be the same. I've actually had an issue in this PR that my test was failing because someone added a new yaml in the corda-api. Diff-ing the outputs showed perfectly what was added.

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Generate.kt
bpaunescu: I need to double check this, but as far as I could see nothing else is printed when using only the `generate` subcommand. Also, I've refactored a bit so this may not be an issue anymore.

bpaunescu: We could add some text to clearly separate the configuration from the rest of the cli output and a `-f` parameter to dump it in a file instead of relying on redirecting the entire output.

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Generate.kt
bpaunescu: I will address this in the next PR where I also delete the "delete" functionality.

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Preview.kt
```diff
@@ -0,0 +1,16 @@
+package net.corda.cli.plugins.topicconfig
+
+import picocli.CommandLine
+import java.io.FileDescriptor
+import java.io.FileWriter
+
+@CommandLine.Command(name = "generate", description = ["Generates a textual representation of the intended Kafka topic configuration"])
```
davidcurrie: Given this is the bit that the customer actually sees, this is the part I was most interested in seeing change!

bpaunescu: Oh damn, I forgot to change that bit too 🤣

davidcurrie (dismissed):
LGTM

davidcurrie (approved):
👍

PR #4680, opened 2023-09-20
CORE-16823 - Token Selection - Notary x500 name should be stored in the state output table

filipesoliveira:
The Notary x500 name is being stored again in the DB.

image


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/PostgresUtxoQueryProvider.kt
```diff
@@ -43,10 +43,10 @@ class PostgresUtxoQueryProvider @Activate constructor(
     override val persistTransactionOutput: String
```
owenstanford, line 43: We also need to change the token query and balance query to filter on the notary field.

owenstanford (approved):
LGTM

blsemo (approved):
LGTM

PR #4682, opened 2023-09-21
CORE-13119 Rollback ledger custom queries to offset-based paging

relyafi:
This PR temporarily reverts ledger custom queries from using stable query paging (introduced in #4548) back to offset query paging.

This is to unblock QA testing, as this change requires QA test adaptations. Stable-query paging will be re-introduced shortly.


blsemo (approved):
LGTM

kyriathar (approved):

PR #4684, opened 2023-09-21
CORE-17261 - Clarify logs for stale serial number

dimosr:
Adding some more information in the logs shown in the MGM side when a member provides a wrong serial number, e.g. to include the submitted serial and the latest one.


nikinagy (approved):
LGTM

PR #4687, opened 2023-09-22
CORE-17112: Avoid delete user or schema for BYODB scenario

BenYip123:
- Split `isPlatformManagedDb` into two separate flags to use the cluster db and for when the connection strings are specified

### DIFF
--- a/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDb.kt
+++ b/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDb.kt
@@ -13,7 +13,7 @@ import net.corda.db.core.DbPrivilege
  */
 internal interface VirtualNodeDb {
     val isPlatformManagedDb: Boolean
-    val ddlConnectionProvided: Boolean
+    val connectionProvided: Boolean
     val dbConnections: Map<DbPrivilege, DbConnection?>
     val dbType: VirtualNodeDbType
 
--- a/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDb.kt
+++ b/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDb.kt
@@ -13,6 +13,7 @@ import net.corda.db.core.DbPrivilege
  */
 internal interface VirtualNodeDb {
     val isPlatformManagedDb: Boolean
+    val ddlConnectionProvided: Boolean
     val dbConnections: Map<DbPrivilege, DbConnection?>
     val dbType: VirtualNodeDbType
 
--- a/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDb.kt
+++ b/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDb.kt
@@ -8,12 +8,13 @@ import net.corda.db.core.DbPrivilege
  * Represents a Virtual Node Database
  *
  * @property isPlatformManagedDb true if the database objects are managed by the Corda platform
+ * @property ddlConnectionProvided true if a DDL connection string has been provided
  * @property dbConnections Map of [DbPrivilege] type to its associated connection config.
  * @property dbType DB type (usage)
  */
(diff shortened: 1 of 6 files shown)


components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/CreateVirtualNodeOperationHandler.kt
```diff
@@ -68,7 +68,7 @@ internal class CreateVirtualNodeOperationHandler(
             }
 
             // For each of the platform DB's run the creation process
-            for (vNodeDb in vNodeDbs.values.filter { it.isPlatformManagedDb }) {
+            for (vNodeDb in vNodeDbs.values.filter { it.isPlatformManagedDb || it.ddlConnectionProvided }) {
```
owenstanford: I'm not sure this is quite correct, looking at @josephzunigadaly's comment in the ticket, we only want to run this line `vNodeDb.createSchemasAndUsers()` if `isPlatformManagedDb` is true, otherwise we skip it, where as we want to run the other steps e.g. ` vNodeDb.runDbMigration()` when either `it.isPlatformManagedDb` or `it.ddlConnectionProvided` is true. The reason being that if a DDL connection has been provided then the schema/user already exists, so we should never try and create it

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDb.kt
```diff
@@ -8,11 +8,13 @@ import net.corda.db.core.DbPrivilege
  * Represents a Virtual Node Database
  *
  * @property isPlatformManagedDb true if the database objects are managed by the Corda platform
+ * @property ddlConnectionProvided true if a DDL connection string has been provided
  * @property dbConnections Map of [DbPrivilege] type to its associated connection config.
  * @property dbType DB type (usage)
  */
 internal interface VirtualNodeDb {
     val isPlatformManagedDb: Boolean
+    val ddlConnectionProvided: Boolean
```
vkolomeyko, line 17: Apologies, I am very late for the "party".
Is it a valid combination to have `isPlatformManagedDb = true` and  `ddlConnectionProvided=true` ?
If not, do we have anything in our code that guards against it?

BenYip123: I believe it should still default to Platform Managed since if `isPlatformManagedDb` is true then there is a missing DML string. 
I have this line which should cover this case but I think it could be improved
https://github.com/corda/corda-runtime-os/blob/621b627a3044a0c7a05eee63981fcdf92d07b282/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/VirtualNodeDbFactoryImpl.kt#L116

components/virtual-node/virtual-node-write-service-impl/src/test/kotlin/net/corda/virtualnode/write/db/impl/VirtualNodeDbFactoryImplTest.kt
```diff
@@ -252,10 +256,10 @@ class VirtualNodeDbFactoryImplTest {
 
         // Uniqueness is set to false
         assertAll(dbs.filter { (dbType, _) -> dbType == VirtualNodeDbType.UNIQUENESS }
-            .map { (dbType, db) -> { assertFalse(db.isPlatformManagedDb, dbType.name) } })
+            .map { (dbType, db) -> { assertFalse(db.ddlConnectionProvided, dbType.name) } })
         // Other types are set to true
         assertAll(dbs.filter { (dbType, _) -> dbType != VirtualNodeDbType.UNIQUENESS }
-            .map { (dbType, db) -> { assertTrue(db.isPlatformManagedDb, dbType.name) } })
+            .map { (dbType, db) -> { assertTrue(db.ddlConnectionProvided, dbType.name) } })
```
josephzunigadaly, line 263: The other tests in this file have been updated to check both flags. I think that's a good approach and it would be nice to see it continue here in this test.

Please can we keep asserts for isPlatformManagedDb and add ddlConnectionProvided as new asserts?

BenYip123: Sure, that sounds good

josephzunigadaly (approved):

PR #4688, opened 2023-09-22
CORE-13623 Client side idempotency id for persistence requests

kyriathar:
- removes `ExternalEventExecutor.execute` method which takes request id, since the event (message) retrying currently only happens when a message is not fully processed and therefore Kafka not notified for this message (Kafka's inherent retrying).

- elaborate comment on request id and idempotency.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
@@ -20,9 +20,9 @@ class ExternalEventExecutorImpl @Activate constructor(
 
     @Suspendable
     override fun <PARAMETERS : Any, RESPONSE : Any, RESUME> execute(
-        requestId: UUID,
         factoryClass: Class<out ExternalEventFactory<PARAMETERS, RESPONSE, RESUME>>,
-        parameters: PARAMETERS
+        parameters: PARAMETERS,
+        requestId: UUID,
     ): RESUME {
         @Suppress("unchecked_cast")
         return with(flowFiberService.getExecutingFiber()) {
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
@@ -23,9 +23,9 @@ class ExternalEventExecutorImpl @Activate constructor(
         factoryClass: Class<out ExternalEventFactory<PARAMETERS, RESPONSE, RESUME>>,
         parameters: PARAMETERS
     ): RESUME {
-        // `requestId` is a unique id per event. It is used to achieve idempotency by de-duplicating events processing (on processors side).
-        // Consuming duplicate events can happen from retrying an event from Kafka which however did some persistent work previously but
-        // did not fully succeed (Kafka was not notified), therefore we retry/ reprocess it.
+        // `requestId` is a unique id per event. It is used to achieve idempotency by de-duplicating events processing (Kafka consumers side).
+        // Consuming duplicate events can happen from retrying an event from Kafka which however did some persistent work previously
+        // but did not fully succeed (Kafka was not notified), therefore we retry/ reprocess it.
         val requestId = UUID.randomUUID().toString()
         @Suppress("unchecked_cast")
         return with(flowFiberService.getExecutingFiber()) {
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
@@ -23,9 +23,9 @@ class ExternalEventExecutorImpl @Activate constructor(
         factoryClass: Class<out ExternalEventFactory<PARAMETERS, RESPONSE, RESUME>>,
         parameters: PARAMETERS
     ): RESUME {
-        // `requestId` is a unique id per event. It is used to achieve idempotency by de-duplicating events processing, on Kafka consumers side.
-        // Consuming duplicate events can happen from retrying an event from Kafka which however did some persistent work previously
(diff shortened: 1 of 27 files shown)


simon-johnson-r3 (dismissed):

libs/flows/flow-api/src/main/kotlin/net/corda/flow/external/events/executor/ExternalEventExecutor.kt
```diff
@@ -27,30 +30,17 @@ interface ExternalEventExecutor {
...
-     * @param parameters The [PARAMETERS] object.
-     *
-     * @param PARAMETERS The type to pass to the factory when suspending/creating the event.
-     * @param RESPONSE The type received as a response from the external processor.
-     * @param RESUME The type the flow resumes with after calling [execute].
-     * @return The object that the flow will resume with.
+     * Calling the other [execute] overload with a random [UUID] as the request id.
      */
     @Suspendable
     fun <PARAMETERS : Any, RESPONSE : Any, RESUME> execute(
```
mbrkic-r3: It seems we don't need this function, just a default value for the one above.

mbrkic-r3 (dismissed):
LGTM

libs/flows/flow-api/src/main/kotlin/net/corda/flow/external/events/executor/ExternalEventExecutor.kt
kyriathar: Yes indeed, good point

libs/flows/flow-api/src/main/kotlin/net/corda/flow/external/events/executor/ExternalEventExecutor.kt
```diff
@@ -27,30 +30,17 @@ interface ExternalEventExecutor {
...
         factoryClass: Class<out ExternalEventFactory<PARAMETERS, RESPONSE, RESUME>>,
         parameters: PARAMETERS
-    ): RESUME
+    ): RESUME = execute(UUID.randomUUID(), factoryClass, parameters)
```
jujoramos: Since the `requestId` is used to deduplicate events, would it make sense to entirely remove this method and force all API consumers to consciously set the `requestId` manually?

kyriathar: Actually good catch, yes will implement this.

jujoramos (dismissed):
LGTM

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
```diff
@@ -20,15 +20,15 @@ class ExternalEventExecutorImpl @Activate constructor(
...
+        requestId: UUID,
         factoryClass: Class<out ExternalEventFactory<PARAMETERS, RESPONSE, RESUME>>,
         parameters: PARAMETERS
     ): RESUME {
         @Suppress("unchecked_cast")
         return with(flowFiberService.getExecutingFiber()) {
             suspend(
                 FlowIORequest.ExternalEvent(
-                    requestId,
+                    requestId.toString(),
```
relyafi: Is there a reason `FlowIORequest.ExternalEvent` isn't being updated to use a UUID instead of a string if a UUID is now mandatory?

relyafi (dismissed):
LGTM from a ledger / notary perspective, one comment to consider

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
kyriathar: Yeah good point, I just thought we are close to serialization now so left it as string.

libs/flows/flow-api/src/main/kotlin/net/corda/flow/external/events/executor/ExternalEventExecutor.kt
```diff
@@ -27,29 +30,7 @@ interface ExternalEventExecutor {
...
-     * @param parameters The [PARAMETERS] object.
-     *
-     * @param PARAMETERS The type to pass to the factory when suspending/creating the event.
-     * @param RESPONSE The type received as a response from the external processor.
-     * @param RESUME The type the flow resumes with after calling [execute].
-     * @return The object that the flow will resume with.
-     */
-    @Suspendable
-    fun <PARAMETERS : Any, RESPONSE : Any, RESUME> execute(
+        requestId: UUID,
```
vlajos: Is it not enough to create this requestId in the impl of this function?
The callers do nothing with it apart from passing it to this function.
Or do I overlook something?
https://github.com/corda/corda-runtime-os/pull/4688#discussion_r1334262711 looks like saying the opposite of this idea. :)

kyriathar: I think the question is where is re-trying happening. Looking at this API (`ExternalEventExecutor`) but also the code further down the stack (I couldn't find any retrying) I believe it should be the caller of this API.

vlajos: (Without knowing too much about the lower layers.)
In all the 30 cases?
I suspect we'll end up using something centralized more or less as a replacement for this `execute` or on top of this one.

kyriathar: Well spotted @vlajos . Will look into it.

simon-johnson-r3 (approved):

libs/flows/flow-api/src/main/kotlin/net/corda/flow/external/events/executor/ExternalEventExecutor.kt
kyriathar: [Suggestion](https://github.com/corda/corda-runtime-os/pull/4688#discussion_r1334546229) applied

filipesoliveira (approved):

jujoramos (approved):
LGTM

mbrkic-r3 (approved):

vlajos (approved):

PR #4689, opened 2023-09-22
CORE-5556, CORE-15396 - Add Support for custom global post-processing handlers

owenstanford:
Add Support for custom global post-processing handler sin the flow pipeline. These handlers enable non flow components to implement their own bespoke logic at the post-processing stage of the pipeline.
Add support for automated token claim release when a flow completes.


owenstanford:
The new smoke test and flow will be moved to the e2e tests as part of CORE-17299

components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/entities/ClaimRelease.kt
```diff
@@ -7,3 +7,4 @@ data class ClaimRelease(
     val usedTokens: Set<String>,
     override val poolKey: TokenPoolKey
 ) : TokenEvent
+
```
vkolomeyko: We can possibly undo changes to this file.

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/publisher/factory/PublisherFactory.kt
```diff
@@ -41,6 +41,4 @@ interface PublisherFactory {
         rpcConfig: RPCConfig<REQUEST, RESPONSE>,
         messagingConfig: SmartConfig,
     ): RPCSender<REQUEST, RESPONSE>
-
```
vkolomeyko, line 44: Non-material changes, we can probably undo them.

testing/cpbs/ledger-utxo-demo-app/src/main/kotlin/com/r3/corda/demo/utxo/token/selection/TokenSelectionFlow.kt
```diff
@@ -62,4 +62,5 @@ class TokenSelectionFlow : ClientStartableFlow {
             throw e
         }
     }
-}
```
vkolomeyko: Non-material changes, we can probably undo them.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/impl/token/selection/impl/PoolKey.kt
```diff
@@ -1,12 +1,20 @@
 package net.corda.ledger.utxo.impl.token.selection.impl
 
+import com.fasterxml.jackson.annotation.JsonAutoDetect
+import com.fasterxml.jackson.annotation.JsonProperty
 import net.corda.data.ledger.utxo.token.selection.key.TokenPoolCacheKey
 
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
```
vkolomeyko: I am wondering why do we need these additional annotations for JSon marshalling? Why previous version of `data class` was not good enough?

vkolomeyko (commented):
Looks broadly fine, but I am not at all an expert in this piece of functionality.
@filipesoliveira - would you mind having a look as well?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/impl/token/selection/entities/TokenClaimCheckpointState.kt
```diff
@@ -0,0 +1,13 @@
+package net.corda.ledger.utxo.impl.token.selection.entities
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect
+import com.fasterxml.jackson.annotation.JsonProperty
+import net.corda.ledger.utxo.impl.token.selection.impl.PoolKey
+
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+data class TokenClaimCheckpointState (
```
lankydan, line 5: Do we need the annotation? I wouldn't expect Jackson to have issues with this class.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/impl/token/selection/impl/PoolKey.kt
```diff
@@ -1,12 +1,20 @@
...
 data class PoolKey(
+    @JsonProperty("shortHolderId")
```
lankydan: Wouldn't expect jackson to need these annotations.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
```diff
@@ -107,7 +111,24 @@ class FlowEventProcessorImpl(
...
+        for (postProcessingHandler in postProcessingHandlers) {
+            try {
+                cleanupEvents.addAll(postProcessingHandler.postProcess(pipeline.context))
+            } catch (e: Exception) {
+                log.error(
+                    "The flow event post processing handler '${postProcessingHandler::javaClass}' failed and will be ignored.",
+                    e
+                )
+            }
         }
```
lankydan, line 128: Feels like this should go into the pipeline class.

Either as a separate step after `globalPostProcessing` or as part of it.

You could also argue that the existing code in the global post processor could leverage this new post processing interface so that the session, flow mapper, external events and retries can be broken out and housed separately.

That last part can be done separately though.

owenstanford: The problem we have is the exception handling can mark the flow as failed, which we need to know about before we execute the post processing handler as we need to send clean-up events whenever a flow dies. This is not the ideal solution but I don't think we could achieve the same effect without a more comprehensive restructure.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/impl/token/selection/entities/TokenClaimCheckpointState.kt
owenstanford: Yes, without them you get the following error:
```Cannot construct instance of `net.corda.ledger.utxo.impl.token.selection.entities.TokenClaimCheckpointState` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)```
My (limited) understanding is this relates to how data classes are represented in java. Removing any of the annotation causes the deserialization unit test to fail.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/impl/token/selection/impl/PoolKey.kt
owenstanford: see comment above

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/impl/token/selection/impl/PoolKey.kt
owenstanford: before this change we never serialized/deserialized this class.
See comment below for why the annotations are needed

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/publisher/factory/PublisherFactory.kt
owenstanford: I've undone the others but keeping this one as it cleans up the formatting.

vkolomeyko (dismissed):
LGTM

lankydan (dismissed):

mbrkic-r3:
What is the impact on flow pipeline processing performance? Are handlers doing any I/O operations?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
mbrkic-r3: Is it more complex than performing each step in try/catch block, storing throwable in context and returning immediately if throwable is already set?

owenstanford:
> What is the impact on flow pipeline processing performance? Are handlers doing any I/O operations?
@mbrkic-r3 
No they are very lightweight and in all but exceptional cases don't need to do anything.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/impl/token/selection/impl/PoolKey.kt
mbrkic-r3: Using `val objectMapper = ObjectMapper().registerKotlinModule().registerModule(JavaTimeModule())` in `FlowCheckpointImpl` might help.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
owenstanford: If I understand the suggestion, I think that would be more complex and would potentially miss the step, The bottom line is we need to apply this post processing stage after exception handling or we need to redesign the way we handle the various failure conditions, for this change I propose we leave as-is, happy to raise a tech-debt ticket if we want to come back and reconsider this.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/impl/token/selection/impl/PoolKey.kt
owenstanford: Switched to this method as it's much cleaner, thanks

mbrkic-r3 (approved):
LGTM

lankydan (approved):

vkolomeyko (approved):
LGTM

PR #4690, opened 2023-09-22
CORE-17300 Multi-SourceEventMediator - MessageBusConsumer 

mbrkic-r3:
As part of the multi-source event mediator implementation we will have a new type of consumer which should be able to poll messages from different endpoints (from Kafka topics, as of today). This change introduces `MediatorConsumer` interface and `MessageBusConsumer` that implements it.

### DIFF
--- a/libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/MediatorConsumer.kt
+++ b/libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/MediatorConsumer.kt
@@ -1,0 +1,36 @@
+package net.corda.messaging.api.mediator
+
+import kotlinx.coroutines.Deferred
+import net.corda.messagebus.api.CordaTopicPartition
+import net.corda.messagebus.api.consumer.CordaConsumerRecord
+import java.time.Duration
+
+/**
+ * Multi-source event mediator message consumer.
+ */
+interface MediatorConsumer<K : Any, V : Any> : AutoCloseable {
+
+    /**
+     * Subscribes to a message bus.
+     */
+    fun subscribe()
+
+    /**
+     * Poll messages from the consumer with a [timeout].
+     *
+     * @param timeout - The maximum time to block if there are no available messages.
+     */
+    fun poll(timeout: Duration): List<CordaConsumerRecord<K, V>>
+
+    /**
+     * Asynchronously commit the consumer offsets.
+     *
+     * @return [CompletableFuture] with committed offsets.
+     */
+    fun commitAsyncOffsets(): Deferred<Map<CordaTopicPartition, Long>>
+
+    /**
+     * Resets consumer's offsets to the last committed positions.
+     */
+    fun resetEventOffsetPosition()
+}
--- a/libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/MediatorConsumer.kt
+++ b/libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/MediatorConsumer.kt
@@ -23,14 +23,14 @@ interface MediatorConsumer<K : Any, V : Any> : AutoCloseable {
     fun poll(timeout: Duration): List<CordaConsumerRecord<K, V>>
 
     /**
-     * Asynchronously commit the consumer offsets.
+     * Asynchronously commit the consumer offsets. This function should be called only after `poll` was called.
      *
-     * @return [CompletableFuture] with committed offsets.
+     * @return [Deferred] with committed offsets.
      */
-    fun commitAsyncOffsets(): Deferred<Map<CordaTopicPartition, Long>>
+    fun asyncCommitOffsets(): Deferred<Map<CordaTopicPartition, Long>>
 
     /**
-     * Resets consumer's offsets to the last committed positions.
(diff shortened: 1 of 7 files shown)


emilybowe (dismissed):
LGTM (pending 🟢 )

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/MediatorConsumer.kt
```diff
@@ -0,0 +1,36 @@
...
+     * Poll messages from the consumer with a [timeout].
+     *
+     * @param timeout - The maximum time to block if there are no available messages.
+     */
+    fun poll(timeout: Duration): List<CordaConsumerRecord<K, V>>
+
+    /**
+     * Asynchronously commit the consumer offsets.
+     *
+     * @return [CompletableFuture] with committed offsets.
```
simon-johnson-r3: it doesn't return a `CompletableFuture`

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/MediatorConsumer.kt
```diff
@@ -0,0 +1,36 @@
...
+import net.corda.messagebus.api.CordaTopicPartition
+import net.corda.messagebus.api.consumer.CordaConsumerRecord
+import java.time.Duration
+
+/**
+ * Multi-source event mediator message consumer.
+ */
+interface MediatorConsumer<K : Any, V : Any> : AutoCloseable {
+
+    /**
```
simon-johnson-r3, line 13: From the perspective of this interface, subscribing is unnecessary, and in fact it's fairly arbitrary when I call it. The important bit is poll. If you changed the contract such that any subscribing to the underlying bus implementation was done in the first poll, you remove the client code getting it wrong.

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/MediatorConsumer.kt
```diff
@@ -0,0 +1,36 @@
...
+     * @param timeout - The maximum time to block if there are no available messages.
+     */
+    fun poll(timeout: Duration): List<CordaConsumerRecord<K, V>>
+
+    /**
+     * Asynchronously commit the consumer offsets.
+     *
+     * @return [CompletableFuture] with committed offsets.
+     */
+    fun commitAsyncOffsets(): Deferred<Map<CordaTopicPartition, Long>>
```
simon-johnson-r3: Are these last two methods safe to call if I never polled?

simon-johnson-r3: minor perhaps but this is really `asyncCommitOffsets` :-D
It's the commit that's async not the offsets, which have no concept of asynchronicity!

mbrkic-r3: Most functions just call the function with the same name in `CordaConsumer` and it was already [there](https://github.com/corda/corda-runtime-os/blob/00f7303b83ed7c07f1c07efac7e5554b06d2c056/libs/messaging/message-bus/src/main/kotlin/net/corda/messagebus/api/consumer/CordaConsumer.kt#L146) `commitSyncOffsets` so I used the same "naming convention". But you are right, I'll change it at all places (hopefully it won't affect too many files).

mbrkic-r3: Everything is delegated to `CordaConsumer.kt` and it's documentation doesn't say anything about it so I'll check implementation and update documentation.

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/MediatorConsumer.kt
mbrkic-r3: Yes, from interface perspective it would be simpler without it. Having it separated helps to identify issues before component's status is set to "UP" and also "reuse" error handling we have in `StateAndEventSubscription` (which is battle tested), while having it within `poll` might make it more complex. Due to time constraints I'd rather keep it similar to `StateAndEventSubscription` for now and make improvements later.

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/MediatorConsumer.kt
mbrkic-r3: Fixed

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/MediatorConsumer.kt
mbrkic-r3: It's not clear for the first one, but it makes sense to do it only if `poll` was called, and the last one can be called before `poll`. Documentation updated.

mbrkic-r3: Fixed.

simon-johnson-r3 (approved):

thiagoviana (approved):
LGTM

PR #4691, opened 2023-09-22
CORE-16983 / CORE-16967 Atomic Swap with Facade

jmacmahonr3:
API additions in https://github.com/corda/corda-api/pull/1282


components/flow/flow-service/build.gradle
```diff
@@ -48,6 +48,7 @@ dependencies {
     implementation project(':libs:platform-info')
     implementation project(":libs:serialization:serialization-avro")
     implementation project(":libs:tracing")
+    implementation project(':libs:crypto:merkle-impl')
```
vlajos, line 54: Depending on `-impl` looks a bit suspicious.
Dependencies should be injected via OSGi.
Although probably some direct constructors are not available via OSGi.

wzur-r3 (approved):
LGTM

PR #4693, opened 2023-09-22
CORE-17276 Refactor DB bootstrap into two containers

davidcurrie:
Refactor DB bootstrapping from 14 containers down to 2 so that a) they run faster as Kubernetes is not checking for updated images between every stage and b) it is easier to reason about the logic when making updates.

### DIFF
--- a/charts/corda-lib/templates/_bootstrap.tpl
+++ b/charts/corda-lib/templates/_bootstrap.tpl
@@ -139,62 +139,201 @@ spec:
       securityContext:
         {{- . | toYaml | nindent 8 }}
       {{- end }}
-      containers:
-        - name: fin
+      initContainers:
+        - name: generate
           image: {{ include "corda.bootstrapCliImage" . }}
           imagePullPolicy: {{ .Values.imagePullPolicy }}
           {{- include "corda.bootstrapResources" . | nindent 10 }}
           {{- include "corda.containerSecurityContext" . | nindent 10 }}
-          command:
-            - /bin/bash
-            - -e
-            - -c
-          args: ["echo", "'DB Bootstrapped'"]
+          command: [ 'sh', '-c', '-e' ]
+          args:
+            - |
+              #!/bin/sh
+
+              set -ev
+
+              JDBC_URL="jdbc:{{ include "corda.clusterDbType" . }}://{{ required "A db host is required" .Values.db.cluster.host }}:{{ include "corda.clusterDbPort" . }}/{{ include "corda.clusterDbName" . }}"
+
+              echo 'Generating DB specification'
+              mkdir /tmp/db
+              java -Dpf4j.pluginsDir=/opt/override/plugins -Dlog4j2.debug=false -jar /opt/override/cli.jar database spec \
+                -g "config:${DB_CLUSTER_SCHEMA},rbac:${DB_RBAC_SCHEMA},crypto:${DB_CRYPTO_SCHEMA},stateManager:${DB_STATE_MANAGER_SCHEMA}" \
+                -u "${PGUSER}" -p "${PGPASSWORD}" \
+                --jdbc-url "${JDBC_URL}" \
+                -c -l /tmp/db
+
+              echo 'Generating RBAC initial DB configuration'
+              mkdir /tmp/rbac
+              java -Dpf4j.pluginsDir=/opt/override/plugins -Dlog4j2.debug=false -jar /opt/override/cli.jar initial-config create-db-config \
+                -u "${RBAC_DB_USER_USERNAME}" -p "${RBAC_DB_USER_PASSWORD}" \
+                --name "corda-rbac" \
+                --jdbc-url "${JDBC_URL}?currentSchema=${DB_RBAC_SCHEMA}" \
+                --jdbc-pool-max-size {{ .Values.bootstrap.db.rbac.dbConnectionPool.maxSize | quote }} \
+              {{- with .Values.bootstrap.db.rbac.dbConnectionPool.minSize }}
+                --jdbc-pool-min-size {{ . | quote }}
+              {{- end }}
+                ${DB_RBAC_MIN_POOL_SIZE_ARGS} \
+                --idle-timeout {{ .Values.bootstrap.db.rbac.dbConnectionPool.idleTimeoutSeconds | quote }} \
+                --max-lifetime {{ .Values.bootstrap.db.rbac.dbConnectionPool.maxLifetimeSeconds | quote }} \
+                --keepalive-time {{ .Values.bootstrap.db.rbac.dbConnectionPool.keepaliveTimeSeconds | quote }} \
+                --validation-timeout {{ .Values.bootstrap.db.rbac.dbConnectionPool.validationTimeoutSeconds | quote }} \
+              {{- if (((.Values).config).vault).url }}
+                -t "VAULT" --vault-path "dbsecrets" --key "rbac-db-password" \
+              {{- else }}
+                --salt "${SALT}" --passphrase "${PASSPHRASE}" \
+              {{- end }}
+                -l /tmp/rbac
+
+              echo 'Generating virtual nodes initial DB configuration'
+              mkdir /tmp/vnodes
+              java -Dpf4j.pluginsDir=/opt/override/plugins -Dlog4j2.debug=false -jar /opt/override/cli.jar initial-config create-db-config \
+                -a -u "${DB_CLUSTER_USERNAME}" -p "${DB_CLUSTER_PASSWORD}" \
+                --name "corda-virtual-nodes" \
+                --jdbc-url "${JDBC_URL}" \
+                --jdbc-pool-max-size {{ .Values.bootstrap.db.rbac.dbConnectionPool.maxSize | quote }} \
+              {{- with .Values.bootstrap.db.rbac.dbConnectionPool.minSize }}
+                --jdbc-pool-min-size {{ . | quote }}
+              {{- end }}
+                ${DB_RBAC_MIN_POOL_SIZE_ARGS} \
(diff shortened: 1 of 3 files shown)


charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -132,69 +132,191 @@ spec:
...
           {{- include "corda.containerSecurityContext" . | nindent 10 }}
-          command:
-            - /bin/bash
-            - -e
-            - -c
-          args: ["echo", "'DB Bootstrapped'"]
+          command: [ 'sh', '-c', '-e' ]
+          args:
+            - |
+              #!/bin/sh
```
dickon, line 151: I think is quite a lot of shell script to have in helm, and I think I'd rather have this logic in Kotlin so that the development cycle for it is more pleasant. I think I can argue it either way though.

davidcurrie: I did start by trying to have the shell scripts outside of the Helm chart and imported, but sadly, a library Helm chart such as `corda-lib` can only access files from the parent Helm chart it is consumed in, so we'd end up with copies in `corda` and `enterprise-corda`. In the defence of the scripts, they're doing little other than stringing together a sequence of commands. I'd agree that may still be better done in code, although you've then got to work out how to marshal the large number of arguments. As you say, swings and roundabouts...

dickon:
The bootstrap hasn't struck me as epecially slow, but that may be because I'm using local Docker Kubernetes.

This will conflict a bit with what @malachyb and I did recently if that hasn't been merged yet, but it's workable.

I think my preference is to have as little complexity in Helm as possible, and reduce it all to a Kotlin `corda-cli` Corda instlaler module which can either generate 10+ blocks of SQL content and maybe some commands to run (e.g. vault setup) or simply execute them directly. We'd then basically run that with a bunch of environment variables via Helm containing the secrets, and it could be run locally. I'd also like to align that with whatever local developers do.

To put it another way, let's have as much as possible in Kotlin and minimise the logic in Helm/Kubernetes since not everyone uses Kubernetes. Maybe in practice nearly all production deployments do, but that's less true of development of Corda itself, and perhaps development of Cordapps. 

I think this may overlap a little with QA frameworks such as `corda-non-function-test`'s `eks.sh`.

davidcurrie:
> The bootstrap hasn't struck me as epecially slow, but that may be because I'm using local Docker Kubernetes.

If you're using the local developer overrides, then they set the image pull policy to `IfNotPresent` so you don't get the extra call to check the registry between each init container. We will only really see the benefits here when we reduce the number of times that we start a JVM (in line with what you suggest below), but I see this as a stepping stone towards that. The main driver at the moment is reducing the cognitive load on anyone who has to make changes in this area. It also makes it easier to find the relevant logs when something goes wrong: with the generation generally reliable, failures are more likely to be in the final container such that `kubectl logs` just works.

> This will conflict a bit with what @malachyb and I did recently if that hasn't been merged yet, but it's workable.

I don't think it has merged yet, but happy to help out there if needed.

> I think my preference is to have as little complexity in Helm as possible, and reduce it all to a Kotlin `corda-cli` Corda instlaler module which can either generate 10+ blocks of SQL content and maybe some commands to run (e.g. vault setup) or simply execute them directly. We'd then basically run that with a bunch of environment variables via Helm containing the secrets, and it could be run locally. I'd also like to align that with whatever local developers do.
> 
> To put it another way, let's have as much as possible in Kotlin and minimise the logic in Helm/Kubernetes since not everyone uses Kubernetes. Maybe in practice nearly all production deployments do, but that's less true of development of Corda itself, and perhaps development of Cordapps.
> 
> I think this may overlap a little with QA frameworks such as `corda-non-function-test`'s `eks.sh`.

I don't disagree, but that's a bigger piece of work and impacts the product externals. I consider this PR a useful step in that direction, as it clarifies what commands are currently run as part of the bootstrapping.

charts/corda-lib/templates/_bootstrap.tpl
dickon: Okay, happy with this approach as an incremental step forward

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -132,69 +132,191 @@ spec:
...
+                --jdbc-url "${JDBC_URL}" \
+                --jdbc-pool-max-size {{ .Values.bootstrap.db.rbac.dbConnectionPool.maxSize | quote }} \
+              {{- with .Values.bootstrap.db.rbac.dbConnectionPool.minSize }}
+                --jdbc-pool-min-size {{ . | quote }}
+              {{- end }}
+                --idle-timeout {{ .Values.bootstrap.db.rbac.dbConnectionPool.idleTimeoutSeconds | quote }} \
+                --max-lifetime {{ .Values.bootstrap.db.rbac.dbConnectionPool.maxLifetimeSeconds | quote }} \
+                --keepalive-time {{ .Values.bootstrap.db.rbac.dbConnectionPool.keepaliveTimeSeconds | quote }} \
+                --validation-timeout {{ .Values.bootstrap.db.rbac.dbConnectionPool.validationTimeoutSeconds | quote }} \
+              {{- if (((.Values).config).vault).url }}
```
dickon, line 199: Some of this is repeated 3 times, which is small enough I can live with it for now, especially with a view to collapsing this into a single run of corda-cli later.

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -132,69 +132,191 @@ spec:
...
+                -t "VAULT" --vault-path "dbsecrets" --key "rbac-db-password" \
+              {{- else }}
+                --salt "${SALT}" --passphrase "${PASSPHRASE}" \
+              {{- end }}
+                -l /tmp/rbac
+
+              echo 'Generating virtual nodes initial DB configuration'
+              mkdir /tmp/vnodes
+              java -Dpf4j.pluginsDir=/opt/override/plugins -Dlog4j2.debug=false -jar /opt/override/cli.jar initial-config create-db-config \
+                -a -u "${DB_CLUSTER_USERNAME}" -p "${DB_CLUSTER_PASSWORD}" \
```
dickon, line 188: I'm not actually sure why we need the admin flag here but not the other cases. However I think you've correctly ported  what we had before. So that's good, I just would like at some point to understand why `corda-cli` needs different flags in this case.

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -132,69 +132,191 @@ spec:
...
+                -l /tmp/crypto
+
+              echo 'Generating REST API user initial configuration'
+              java -Dpf4j.pluginsDir=/opt/override/plugins -Dlog4j2.debug=false -jar /opt/override/cli.jar initial-config create-user-config \
+                -u "${REST_API_ADMIN_USERNAME}" -p "${REST_API_ADMIN_PASSWORD}" \
+                -l /tmp
+
+              echo 'Generating crypto initial configuration'
+              java -Dpf4j.pluginsDir=/opt/override/plugins -Dlog4j2.debug=false -jar /opt/override/cli.jar initial-config create-crypto-config \
+                --salt "${SALT}" --passphrase "${PASSPHRASE}" \
```
dickon, line 234: There's actually a small change here whics is that this code specifies salt and passpharse even when using Vault. I'm not sure if the corda-cli module would still work in that case, but it's probably not what we want since we're specifying meaningless values for Vault.

charts/corda-lib/templates/_bootstrap.tpl
dickon: actually 4 times.

charts/corda-lib/templates/_bootstrap.tpl
dickon: Disregard, I reread the old code more carefully.

dickon (dismissed):
LGTM, nice work.

While I'm a bit hestiant to see the repetition, I found the abstracted version I did with all the conditionals actually quite hard to maintain, and would take some effort to repeat that here, so let's go with this as it is now.

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -496,108 +618,3 @@ Bootstrap Corda CLI environment variables
...
-
-           {{- /* specify DB password */ -}}
-           {{- " '-p'" -}}, '$({{ .environmentVariablePrefix -}}_PASSWORD)',
-         {{- end -}}
-
-         {{- if and (not (eq .name "rest")) (not (eq .subCommand "create-crypto-config")) -}}
-             {{- " '--name'" -}}, 'corda-{{ .longName | default .name }}',
-             {{- " '--jdbc-url'" -}}, 'jdbc:{{ include "corda.clusterDbType" . }}://{{ required "A db host is required" .Values.db.cluster.host }}:{{ include "corda.clusterDbPort" . }}/{{ include "corda.clusterDbName" . }}{{- if .schema }}?currentSchema={{.schema }}{{- end -}}',
-             {{- " '--jdbc-pool-max-size'" -}}, {{ (index .Values.bootstrap.db (.dbName | default .name)).dbConnectionPool.maxSize | quote }},
-             {{- if not (kindIs "invalid" (index .Values.bootstrap.db (.dbName | default .name)).dbConnectionPool.minSize) -}}
```
kyriathar, line 533: I can see we didn't port this `{{- if not (kindIs "invalid"...` for min pool size. I remember it was needed back then because helm would interpret 0 as null and so the below would ignore 0 as if it was a null value:

```
{{- with (index .Values.bootstrap.db (.dbName | default .name)).dbConnectionPool.minSize -}}
   {{- " '--jdbc-pool-min-size'" -}}, {{ . | quote }},
{{- end -}}
```

davidcurrie: Good catch - I had forgotten about that zero scenario. I will reinstate the previous logic.

kyriathar (approved):
LGTM

PR #4694, opened 2023-09-22
CORE-16200 Multi-Source Event Mediator configurations and factories

mbrkic-r3:
Multi-Source Event Mediator configurations and factories for consumers and messaging clients.

### DIFF
--- a/libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/factory/MediatorConsumerFactoryFactory.kt
+++ b/libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/factory/MediatorConsumerFactoryFactory.kt
@@ -1,0 +1,21 @@
+package net.corda.messaging.api.mediator.factory
+
+import net.corda.libs.configuration.SmartConfig
+
+/**
+ * Factory for creating multi-source event mediator consumer factories.
+ */
+interface MediatorConsumerFactoryFactory {
+    /**
+     * Creates a message bus consumer factory.
+     *
+     * @param topicName Topic name.
+     * @param groupName Consumer group name.
+     * @param messageBusConfig Message bus related configuration.
+     */
+    fun createMessageBusConsumerFactory(
+        topicName: String,
+        groupName: String,
+        messageBusConfig: SmartConfig,
+    ) : MediatorConsumerFactory
+}
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/factory/MediatorConsumerFactoryFactoryImpl.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/factory/MediatorConsumerFactoryFactoryImpl.kt
@@ -1,0 +1,28 @@
+package net.corda.messaging.mediator.factory
+
+import net.corda.libs.configuration.SmartConfig
+import net.corda.messagebus.api.consumer.builder.CordaConsumerBuilder
+import net.corda.messaging.api.mediator.factory.MediatorConsumerFactoryFactory
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.Reference
+
+/**
+ * Factory for creating multi-source event mediator consumers.
+ */
+@Component(service = [MediatorConsumerFactoryFactory::class])
+class MediatorConsumerFactoryFactoryImpl @Activate constructor(
+    @Reference(service = CordaConsumerBuilder::class)
+    private val cordaConsumerBuilder: CordaConsumerBuilder,
+): MediatorConsumerFactoryFactory {
+    override fun createMessageBusConsumerFactory(
+        topicName: String,
+        groupName: String,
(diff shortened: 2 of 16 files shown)


libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/factory/MediatorConsumerFactoryFactory.kt
```diff
@@ -0,0 +1,21 @@
+package net.corda.messaging.api.mediator.factory
+
+import net.corda.libs.configuration.SmartConfig
+
+/**
+ * Factory for creating multi-source event mediator consumer factories.
+ */
+interface MediatorConsumerFactoryFactory {
```
jujoramos, line 8:
```suggestion
interface MediatorConsumerFactory {
```

mbrkic-r3: It's not a typo, there is also a `MediatorConsumerFactory` that this one creates :wink:

mbrkic-r3: Multi-source event mediator needs to work with factories (in order to recreate consumers/clients for specific error types). Those factories will be different instances for different clients (e.g. Flow Engine, Flow Mapper) so they couldn't be created using DI. What could be created using DI (i.e. what will be the same) are factories of those factories, so this is a way to avoid client code to depend on implementation packages.

jujoramos: What about `MediatorConsumerFactoryProvider` or `MediatorConsumerFactoryCreator`?, wouldn't that be more readable? 😅

mbrkic-r3: I don't think having `FactoryFactory` is a bad thing, it's consistent naming where you remove the last `Factory` suffix and know exactly what that factory is creating.

ben-millar: I don't have any strong feelings, but an `Abstract Factory` is an established pattern whereby we'd have an `AbstractConsumerFactory` which is implemented by `MessageBusConsumerFactory`, etc.

We'd then create instances of these concrete factories with a `ConsumerFactoryProducer` (the factory factory, so to speak). That hierarchy could be more clear. Also, in both cases, potentially we could enumerate the types with enums rather than having functions for each (E.g. `createMessageBusConsumerFactory()`).

That might look something like:

```
interface AbstractConsumerFactory {
    fun createConsumer(): AbstractConsumer
}

----

class MessageBusConsumerFactory : AbstractConsumerFactory {
    override fun createConsumer(): AbstractConsumer {
        return MessageBusConsumer()
    }
}

----

class RPCConsumerFactory : AbstractConsumerFactory {
    override fun createConsumer(): AbstractConsumer {
        return RPCConsumer()
    }
}

----

enum class ConsumerType {
    MESSAGE_BUS,
    RPC
}

object ConsumerFactoryProducer {
    fun getFactory(consumerType: ConsumerType): AbstractConsumerFactory {
        return when (consumerType) {
            ConsumerType.MESSAGE_BUS -> MessageBusConsumerFactory()
            ConsumerType.RPC -> RPCConsumerFactory()
        }
    }
}
```

But, again, I don't wanna create more work for the sake of work. Feel free to ignore if you don't think it's clearer.

mbrkic-r3: That's a nice solution when all factories implement the same interface, which I don't think will be the case here (not sure yet how it will look for RPC, but very unlikely it will be same, as discussed below).

Here is example for specifying consumer factories:
```
eventMediatorConfigBuilder
    .consumerFactories(
        mediatorConsumerFactoryFactory.createMessageBusConsumerFactory(
            FLOW_SESSION_EVENT_TOPIC, CONSUMER_GROUP, messagingConfig
        ),
        mediatorConsumerFactoryFactory.createMessageBusConsumerFactory(
            FLOW_START_EVENT_TOPIC, CONSUMER_GROUP, messagingConfig
        ),
    )
```

Factory accepts topic name (`String`), consumer group (`String`) and messaging config (`SmartConfig`). For some other type of consumer, topic name wouldn't be required, while some other parameters (of different types) would.

We could make interface same for different factories by using `Map<String, Any>`s, but I thought it's nicer to have type safety and also more concise to provide parameters in this way instead of key/value pairs.

mbrkic-r3: @jujoramos @ben-millar
Does explanation above resolve your comments?

thiagoviana (approved):

PR #4695, opened 2023-09-25
CORE-13119 Re-instate stable query paging for ledger custom queries

relyafi:
Re-instate changes implemented in #4548 and reverted in #4682


vlajos (approved):
LGTM.

PR #4702, opened 2023-09-26
CORE-17331 Add initial-rbac all-cluster-roles sub-command

davidcurrie:
To reduce the number of containers executed during bootstrap, add a new sub-command to the initial-rbac plugin that creates all three of the cluster-scoped roles. (A single container could have invoked all three sub-commands but the approach taken here avoids starting the JVM three times.)


vkolomeyko (approved):
LGTM

PR #4703, opened 2023-09-26
CORE-17274 remove 5.0 compat flows

vlajos:
SInce 5.0-5.1 flow communication is not possible, we can remove the 5.1 flows' 5.0 compatibility modes.
We keep the related infrastructure intact for future versions.


nkovacsx (approved):
LGTM

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainSenderFlowV1Test.kt
```diff
@@ -1,3 +1,80 @@
 package net.corda.ledger.utxo.flow.impl.flows.backchain.v1
 
-//  Same as V2, so removed.
\ No newline at end of file
+import net.corda.crypto.core.SecureHashImpl
```
lankydan, line 3: Not sure what is going on with the diff here.

lankydan (approved):

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/backchain/v1/TransactionBackchainSenderFlowV1Test.kt
vlajos: Previously, the v1/v2 tests were the same. So, I removed the v1 ones to not have to maintain them.
And this got reverted now.

PR #4619, opened 2023-09-11
NOTICK- Claim debugging51

owenstanford:

PR #4706, opened 2023-09-26
CORE-17353 - Add metrics to the token selection

filipesoliveira:
Metrics have to be added to the token selection service so they can be used as part of the performance tests.

This PR is mirroring https://github.com/corda/corda-runtime-os/pull/4619.

### DIFF
--- a/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/TokenSelectionMetricsImpl.kt
+++ b/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/TokenSelectionMetricsImpl.kt
@@ -1,0 +1,34 @@
+package net.corda.ledger.utxo.token.cache.services
+
+import net.corda.ledger.utxo.token.cache.entities.TokenEvent
+import net.corda.metrics.CordaMetrics
+import net.corda.utilities.time.Clock
+import java.time.Duration
+
+class TokenSelectionMetricsImpl(private val clock: Clock) : TokenSelectionMetrics {
+    override fun <T> recordProcessingTime(tokenEvent: TokenEvent, block: () -> T): T {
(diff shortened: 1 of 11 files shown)


components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/TokenSelectionMetricsImpl.kt
```diff
@@ -0,0 +1,34 @@
...
+            .withTag(CordaMetrics.Tag.TokenSelectionEvent, tokenEvent.javaClass.simpleName)
+            .build().record(Duration.ofNanos(System.nanoTime() - start))
+        return result
+    }
+
+    override fun <T> recordDbOperationTime(dbOperation: String, block: () -> T): T {
+        val start = System.nanoTime()
+        val result = block()
+        CordaMetrics.Metric.TokenSelectionDbExecutionTime.builder()
+            .withTag(CordaMetrics.Tag.TokenSelectionEvent, dbOperation)
```
owenstanford: Sorry this was my fault in the original code I cut/paste the tag from the metric above, can we create a `CordaMetrics.Tag.TokenSelectionDbOperation` or see if there is a generic DB operation that we can use for this tag

libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
```diff
@@ -837,7 +842,17 @@ object CordaMetrics {
         /**
          * Partition of a message bus topic published to or consumed from.
          */
-        Partition("partition")
+        Partition("partition"),
+
+        /**
+         * Partition of a message bus topic published to or consumed from.
```
filipesoliveira: Is this comment correct, @owenstanford ?

vkolomeyko (approved):
LGTM

PR #4708, opened 2023-09-26
CORE-17354 Force initial reconciliation to support updating configuration after upgrade

kyriathar:
**Problem Statement**

If we extend the config schema e.g. add a property X,  and we upgrade a cluster we would expect property X to be available to the components that need it. However, that's not the case, property X is not available and the components trying to access it throw on trying to get it from provided config.

What happens is, reconciliation compares DB records to Kafka records (based on version field) and the ones found to be needing updating on Kafka get published to Kafka. But right before getting published, they go through the defaulting process where properties not present in DB config get added with defaulted values. 

In the above scenario, Kafka is already populated (since we upgrade), meaning Kafka config version match DB config version and therefore no update is needed. Meaning we also miss going through the defaulting process and that's where we miss the newly added property X.

**Solution**

Forcing only the initial reconciliation, i.e. doing a full reconciliation even for DB records that are of same version compared to Kafka ones means we will go through the defaulting process and the updated properties will become available to Kafka.

**Testing**

A config update & cluster upgrade was done locally to reproduce issue and witness fix:
 
1. cluster was deployed
2.  new property was added to config schema and then some code trying to access it 
3. image (db-worker) was rebuilt and new pod started. In the logs there was the exception for the missing property
4. force an initial reconciliation code was added, image rebuilt and new pod started, the property was made available

### DIFF
--- a/components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
+++ b/components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
@@ -131,9 +131,9 @@ internal class ReconcilerEventHandler<K : Any, V : Any>(
                     // On config section schema update, because Kafka is already populated and DB and Kafka records' versions
                     // match for that config section it means that config section would not get reconciled.
                     // This means newly added property(ies) to config schema will not get added to config published on Kafka.
-                    // With forcing reconciliation, all DB configs go through reconciliation (version is overlooked) and therefore
-                    // through the defaulting config process which will add the property(ies) and subsequently will publish them
-                    // to Kafka. We only need to force the first reconciliation.
+                    // With forcing reconciliation, all DB configs go through reconciliation again (version is overlooked) and
+                    // therefore through the defaulting config process which will add the property(ies) and subsequently
+                    // will publish them to Kafka. We only need to force the first reconciliation.
                     if (forceInitialReconciliation && firstRun) {
                         dbRecord.version >= matchedKafkaRecord.version // reconcile all db records
                     } else {
--- a/components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
+++ b/components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
@@ -135,7 +135,7 @@ internal class ReconcilerEventHandler<K : Any, V : Any>(
                     // therefore through the defaulting config process which will add the property(ies) and subsequently
                     // will publish them to Kafka. We only need to force the first reconciliation.
                     if (forceInitialReconciliation && firstRun) {
-                        dbRecord.version >= matchedKafkaRecord.version // reconcile all db records again
+                        dbRecord.version >= matchedKafkaRecord.version // reconcile all db records again (forced reconciliation)
                     } else {
                         dbRecord.version > matchedKafkaRecord.version // reconcile db updated records
                     } || dbRecord.isDeleted // reconcile db soft deleted records
--- a/components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
+++ b/components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
@@ -27,7 +27,7 @@ internal class ReconcilerEventHandler<K : Any, V : Any>(
     keyClass: Class<K>,
     valueClass: Class<V>,
     var reconciliationIntervalMs: Long,
-    private val forceInitialReconciliation: Boolean = false,
+    private val forceInitialReconciliation: Boolean,
 ) : LifecycleEventHandler {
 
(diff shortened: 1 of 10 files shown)


components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerImpl.kt
```diff
@@ -14,7 +14,8 @@ internal class ReconcilerImpl<K : Any, V : Any>(
     keyClass: Class<K>,
     valueClass: Class<V>,
     coordinatorFactory: LifecycleCoordinatorFactory,
-    reconciliationIntervalMs: Long
+    reconciliationIntervalMs: Long,
+    forceInitialReconciliation: Boolean = false,
```
simon-johnson-r3: if you're supposed to instantiate this from the factory, adding a default here wouldn't seem to serve much of a purpose

components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
```diff
@@ -122,8 +127,12 @@ internal class ReconcilerEventHandler<K : Any, V : Any>(
                 val toBeReconciled = if (matchedKafkaRecord == null) {
                     !dbRecord.isDeleted // reconcile db inserts (i.e. db column cpi.is_deleted == false)
                 } else {
-                    dbRecord.version > matchedKafkaRecord.version // reconcile db updates
-                            || dbRecord.isDeleted // reconcile db deletes
+                    // Forced reconciliation is meant to fix CORE-17354
```
simon-johnson-r3: It would be better here to explain what the code does and why instead of relate it back to a JIRA ticket I think (see next comment too)

components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
```diff
@@ -146,6 +155,10 @@ internal class ReconcilerEventHandler<K : Any, V : Any>(
             }
         }
 
+        if (firstRun) {
```
simon-johnson-r3: minor, but you don't need this `if`, always setting firstRun to false at the end of the function is semantically fine and logically equivalent

components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
```diff
@@ -122,8 +127,12 @@ internal class ReconcilerEventHandler<K : Any, V : Any>(
                 val toBeReconciled = if (matchedKafkaRecord == null) {
                     !dbRecord.isDeleted // reconcile db inserts (i.e. db column cpi.is_deleted == false)
                 } else {
-                    dbRecord.version > matchedKafkaRecord.version // reconcile db updates
-                            || dbRecord.isDeleted // reconcile db deletes
+                    // Forced reconciliation is meant to fix CORE-17354
+                    if (forceInitialReconciliation && firstRun) {
+                        dbRecord.version >= matchedKafkaRecord.version // force reconcile
```
simon-johnson-r3: might be worth a comment here to explain the nuances of the if/else, i.e. that one will reconcile the same version again in case of changes, and the other will look for only updated versions

components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
```diff
@@ -26,6 +27,7 @@ internal class ReconcilerEventHandler<K : Any, V : Any>(
     keyClass: Class<K>,
     valueClass: Class<V>,
     var reconciliationIntervalMs: Long,
+    private val forceInitialReconciliation: Boolean = false,
```
simon-johnson-r3: this is only instantiated by the impl which always passes a value, so the default seems unnecessary

components/reconciliation/reconciliation-impl/src/test/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandlerTest.kt
```diff
@@ -49,4 +53,57 @@ internal class ReconcilerEventHandlerTest {
...
+
+        reconcilerEventHandler =
+            ReconcilerEventHandler(
+                dbReader,
+                kafkaReader,
+                writer = mock(),
+                keyClass = String::class.java,
+                valueClass = Int::class.java,
+                10L,
+                forceInitialReconciliation = true
```
simon-johnson-r3: is there a test already where this is set to false and you don't get this behaviour?

components/reconciliation/reconciliation-impl/src/main/kotlin/net/corda/reconciliation/impl/ReconcilerEventHandler.kt
kyriathar: Think I had in mind test code as well, you are right that shouldn't be considered.

simon-johnson-r3 (approved):

PR #4709, opened 2023-09-26
CORE-16181 Implement RPC Producer that implements RPC client

ben-millar:
This PR adds a new implementation of `MessagingClient`, `RPCClient`, handling synchronous calls to the Corda workers via RPC/HTTP as well as a `HTTPRetryConfig` and `HTTPRetryExecutor` to enable simple retrying logic for our HTTP requests on specified failures (typically, retryable 5XX failures).

This PR also extracts the worker endpoint paths (E.g. `/ledger`, `/persistence` etc.) from their individual classes and moves them to a common consts file so they can be used by both clients and servers.


charts/corda/templates/workers.yaml
```diff
@@ -1,25 +1,82 @@
+{{- define "cluster-ip-service" }}
```
davidcurrie: Function names should be "namespaced" to avoid collisions when combining charts. Also, this function should be defined in `corda-lib` so that it can be reused in the Enterprise Helm chart.
```suggestion
{{- define "corda.cluster-ip-service" }}
```

charts/corda/templates/workers.yaml
```diff
@@ -1,25 +1,82 @@
...
+  # Which version of K8s API to use
```
davidcurrie: We don't generally both with comments on YAML that is only going to be seen by k8s!

charts/corda/templates/workers.yaml
```diff
@@ -1,25 +1,82 @@
...
+  apiVersion: v1
+    # The kind of K8s object to create
+  kind: Service
+  metadata:
+    # -- Dynamically generate service name for each worker in values.yaml
+    name: {{ .workerName }}-cluster-ip-service
```
davidcurrie: We generally strive for resource names to be unique to a Helm release (at least in theory, you could deploy the Corda Helm chart twice in one namespace). Take a look at how we get the worker name in `corda.worker`.

charts/corda/templates/workers.yaml
```diff
@@ -1,25 +1,82 @@
+{{- define "cluster-ip-service" }}
+  # Which version of K8s API to use
+  apiVersion: v1
```
davidcurrie: Generally, functions output YAML that is left-aligned, then the function caller indents it if required with `nindent`/`indent`.

charts/corda/templates/workers.yaml
```diff
@@ -1,25 +1,82 @@
...
+    selector:
+      app: {{ .workerName }}-worker
+    # -- List of ports this service exposes and the corresponding ports it routes to
+    ports:
+      # -- The protocol for this port
+        - protocol: TCP
+        # -- The port on which this service listens
+        port: {{ index .Values .workerName "clusterIPService" "port" }}
+        # -- The port on the selected pod to which traffic will be routed
+        targetPort: {{ index .Values .workerName "clusterIPService" "targetPort" }}
```
davidcurrie: I wouldn't let the user configure these. This is internal communication and the user has no need to be able to specify different values here.

charts/corda/templates/workers.yaml
```diff
@@ -1,25 +1,82 @@
...
+{{- end -}}
+
+# -- crypto
+{{- include "cluster-ip-service" (dict "workerName" "crypto") }}
```
davidcurrie: if this is included for all workers, I'd just inline the definition of `cluster-ip-service` into `corda.worker`

charts/corda/templates/workers.yaml
```diff
@@ -1,25 +1,82 @@
...
+    ports:
+      # -- The protocol for this port
+        - protocol: TCP
+        # -- The port on which this service listens
+        port: {{ index .Values .workerName "clusterIPService" "port" }}
+        # -- The port on the selected pod to which traffic will be routed
+        targetPort: {{ index .Values .workerName "clusterIPService" "targetPort" }}
+{{- end -}}
+
+# -- crypto
```
davidcurrie: These comments will end up in the output YAML, which may or may not be what you intended. Use `{{/* ... */}}` for comments that should only appear in the template.

charts/corda/values.yaml
```diff
@@ -593,6 +593,12 @@ workers:
               name: ""
               # -- the crypto worker SASL password secret key for client connection to Kafka
               key: ""
+    # Configuration for the ClusterIP service for load balancing between crypto-worker instances
+    clusterIPService:
+      # -- The port on which the service will be exposed
+      port: 7000
+      # -- The port on the pod to which the service will forward traffic
+      targetPort: 7000
```
davidcurrie: I was going to object to the name, but, as above, I wouldn't expose any of this in external configuration. It's only giving them rope to hang themselves and making your life more complicated when it comes to passing this information to clients.

charts/corda/templates/workers.yaml
```diff
@@ -1,25 +1,57 @@
+{{/* crypto */}}
+{{- include "corda-lib.corda.cluster-ip-service" (dict "worker" "crypto") }}
```
ben-millar: TODO: Inline this into the worker definition

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        return deferred;
+    }
+
+    private fun processMessage(message: MediatorMessage<*>): MediatorMessage<*> {
+        val payload = serializePayload(message)
+
+        val request = HttpRequest.newBuilder()
+            .uri(URI("http://corda-flow-mapper-worker-cluster-ip-service:7000"))
```
driessamyn: The URL will need to be passed in by whatever is consuming this (or the message itself).

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -0,0 +1,121 @@
...
+            .PUT(HttpRequest.BodyPublishers.ofByteArray(payload))
+            .build()
+
+        val response = httpClient.send(request, HttpResponse.BodyHandlers.ofByteArray())
+
+        // TODO: The HTTP client should be abstracted out into its own class which handles retries
```
driessamyn: I'd argue this should _not_ be part of the HTTP client but something else.
(Single responsibility principle).
I.e. a "re-tryer" could be passed into the usage of this.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
ben-millar: This is temp for testing; sorry should've had a comment!

charts/corda-lib/templates/_worker.tpl
```diff
@@ -314,6 +327,9 @@ spec:
           {{- if $.Values.tracing.samplesPerSecond }}
           - "--trace-samples-per-second={{ $.Values.tracing.samplesPerSecond }}"
           {{- end }}
+          {{- range $workerName, $_ := .Values.workers }}
+            - {{ printf "worker.endpoint.%s=%s:$s" $workerName (include "worker-internal-service-name" $workerName {{ include "worker.port" . }}) | quote }}
```
davidcurrie:
```suggestion
            - {{ printf "worker.endpoint.%s=%s:$s" $workerName ( include "worker-internal-service-name" $workerName ) {{ include "worker.port" . }} | quote }}
```

charts/corda-lib/templates/_worker.tpl
```diff
@@ -77,6 +77,19 @@ spec:
...
+metadata:
+  name: {{ include "worker-internal-service-name" $workerName }}
+spec:
+  type: ClusterIP
+  selector:
+    app: {{ $workerName }}
+  ports:
+      - protocol: TCP
+        port: {{ include "worker.port" . }}
+        targetPort: {{ include "worker.port" . }}
```
davidcurrie: For `targetPort` you can actually just use the name of the port (i.e. `worker`), not that it makes much difference when you're doing the include anyway.

charts/corda-lib/templates/_worker.tpl
davidcurrie: Here `$workerName` will be just something like `rest` not the full `corda-rest-worker`.

applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/DefaultWorkerParams.kt
```diff
@@ -72,4 +72,37 @@ class DefaultWorkerParams(healthPortOverride: Int = WORKER_SERVER_PORT) {
         description = ["Configuration for the state manager."]
     )
     var stateManagerParams = emptyMap<String, String>()
+
```
driessamyn: As discussed, these should all be moved to the FlowWorker class rather than the default.

applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/DefaultWorkerParams.kt
```diff
@@ -72,4 +72,37 @@ class DefaultWorkerParams(healthPortOverride: Int = WORKER_SERVER_PORT) {
...
+    @Option(names = ["--worker-endpoint-crypto"], description = ["Internal RPC endpoint for the CryptoWorker"], required = true)
+    var cryptoWorkerEndpoint: String? = null
+
+    @Option(names = ["--worker-endpoint-db"], description = ["Internal RPC endpoint for the DBWorker"], required = true)
+    var dbWorkerEndpoint: String? = null
```
driessamyn: This doesn't exist I believe.

applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/DefaultWorkerParams.kt
```diff
@@ -72,4 +72,37 @@ class DefaultWorkerParams(healthPortOverride: Int = WORKER_SERVER_PORT) {
...
+
+    @Option(names = ["--worker-endpoint-flow"], description = ["Internal RPC endpoint for the FlowWorker"], required = true)
+    var flowWorkerEndpoint: String? = null
+
+    @Option(names = ["--worker-endpoint-flowMapper"], description = ["Internal RPC endpoint for the FlowMapperWorker"], required = true)
+    var flowMapperWorkerEndpoint: String? = null
```
driessamyn: These don't have http endpoints I believe.

applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/DefaultWorkerParams.kt
```diff
@@ -72,4 +72,37 @@ class DefaultWorkerParams(healthPortOverride: Int = WORKER_SERVER_PORT) {
...
+    var membershipWorkerEndpoint: String? = null
+
+    @Option(names = ["--worker-endpoint-rest"], description = ["Internal RPC endpoint for the RestWorker"], required = true)
+    var restWorkerEndpoint: String? = null
+
+    @Option(names = ["--worker-endpoint-p2pGateway"], description = ["Internal RPC endpoint for the P2PGatewayWorker"], required = true)
+    var p2pGatewayWorkerEndpoint: String? = null
+
+    @Option(names = ["--worker-endpoint-p2pLinkManager"], description = ["Internal RPC endpoint for the P2PLinkManagerWorker"], required = true)
+    var p2pLinkManagerWorkerEndpoint: String? = null
```
driessamyn: These don't have http endpoints I believe.

emilybowe (approved):
LGTM

PR #4710, opened 2023-09-26
CORE-16797 Apply Kafka topic configurations

bpaunescu:
- Remove `delete` functionality of the topic configuration plugin
- Add ability to apply topics configuration from a user specified file
- Moved kafka connectivity arguments from `topic` command to `connect` subcommand as they're only needed there

If a topic already exists, it will not be re-created. Instead, the specified configuration for existing topics will be applied. As with topics, existing configuration will not be deleted. This means that if a topic has the following config 

`cleanup.policy: compact
  segment.ms: 600000
  delete.retention.ms: 300000`

and the configuration to be applied is 

`cleanup.policy: compact
  segment.ms: 300000
  min.compaction.lag.ms: 60000
  max.compaction.lag.ms: 604800000`

the end result will be

`cleanup.policy: compact
  segment.ms: 300000
  delete.retention.ms: 300000
  min.compaction.lag.ms: 60000
  max.compaction.lag.ms: 604800000`

All topic ACLs are applied as specified in the file or generated based on user provided arguments. Access cannot be restricted via the plugin.


charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -372,10 +370,10 @@ spec:
             {{- end }}
             '-r', '{{ .Values.bootstrap.kafka.replicas }}',
             '-p', '{{ .Values.bootstrap.kafka.partitions }}',
-            'connect'{{- if .Values.bootstrap.kafka.cleanup }},
-            '-d',
+            'connect',
```
simon-johnson-r3, line 375: We can't really allow this change because we've changed the public api.

Even if we believed customers weren't using this tool, we'd have to change the public docs:
https://docs.r3.com/en/platform/corda/5.0/deploying-operating/deployment/deploying/manual-bootstrapping.html

and have to pick through our own codebases and documentation to reflect this. For example I did a quick search in this repo and found a couple of hits at least, e.g. https://github.com/corda/corda-runtime-os/blob/31bcefb4364ef5bc7f93b6d901856d667f37e289/libs/tracing/README.md?plain=1#L82

We deploy Corda in various pipelines and scripts we might be breaking here. And of course this impacts our customers too.

It would be simpler all round to revert it and live with the way we had it.

simon-johnson-r3 (changes requested):
All looks good except to revert the breaking change.

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/CreateConnect.kt
```diff
@@ -38,56 +38,56 @@ class CreateConnect : Runnable {
...
 
-            val topicConfigsToProcess = if (delete) {
-                if (existingTopicNames.isNotEmpty()) {
-                    println("Deleting existing topics: ${existingTopicNames.joinToString()}")
-                    val configOp = listOf(AlterConfigOp(ConfigEntry("retention.ms", "1"), AlterConfigOp.OpType.SET))
-                    val alterConfigs = existingTopicNames.associate { ConfigResource(ConfigResource.Type.TOPIC, it) to configOp }
-                    client.incrementalAlterConfigs(alterConfigs).all().get(wait, TimeUnit.SECONDS)
-                    client.deleteTopics(existingTopicNames).all().get(wait, TimeUnit.SECONDS)
+                if (existingTopicsToUpdate.isNotEmpty()) {
+                    println("The following topics already exist and will not be included in the configuration update: " +
```
davidcurrie: @simon-johnson-r3 - is this the expected behaviour for existing topics, or are we expecting to reconcile config to match the latest requirements?

davidcurrie (commented):
LGTM but question for Simon on what should happen to existing topics.

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/CreateConnect.kt
simon-johnson-r3: I'd say it's not ideal but we could live with it if we had to.

@bpaunescu is there a technical reason we cannot apply the new configuration?

bpaunescu: I initially thought it's not possible to update topic configuration, but there is a method `incrementalAlterConfigs()` which takes a map of topic to config and updates. Happy to do that. I believe the previous behaviour of the tool was to either delete or ignore existing topics before applying the configuration.

bpaunescu: @davidcurrie @simon-johnson-r3 I've updated the code to no longer ignore existing topics. Any configuration either from the preview or from user provided file will be applied as is to existing topics. Existing values are updated, new ones are added. Nothing is removed.

davidcurrie (approved):
LGTM. Having added the ability to update the config, that does still leave the potential for a mismatch in partition/replica count but that's a harder problem to solve and it sounds like you might be revisiting the partition part anyway.

simon-johnson-r3 (approved):

PR #4711, opened 2023-09-26
CORE-15064: Set `mixinStandardHelpOptions = true`

vkolomeyko:
For all the CLI plugins where `subcommands` are used. This would enable to pass `-h` argument and properly receive help into StdOut rather than to StdErr like it is currently.


driessamyn (approved):
LGTM

ronanbrowne (approved):
LGTM

dimosr (approved):
LGTM - @corda/corda-platform-network-team files

PR #4712, opened 2023-09-26
CORE-17326: Search by Metadata & Last Update Time

jujoramos:
- Make query provider implementation for state manager pluggable.
- Add SQL and API methods to filter states by last updated time and
  single key from within the metadata.


libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -79,17 +79,34 @@ interface StateManager : AutoCloseable {
      * @param finish Time filter upper bound (inclusive).
      * @return States that were last updated between [start] and [finish] times.
      */
-    fun getUpdatedBetween(start: Instant, finish: Instant): Map<String, State>
+    fun updatedBetween(start: Instant, finish: Instant): Map<String, State>
```
JamesHR3: For these APIs, I wouldn't be against making a data class for the filters. It seems that sets of parameters must always appear together, and doing this would probably make the API clearer. That would especially be true if we had to extend the API in the future (e.g. to allow querying by multiple metadata keys at once).

Something like
```kotlin
data class TimeFilter(val start: Instant, val finish: Instant)
data class MetadataFilter(val key: String, val operation: Operation, val value: Any)
```

What do you think?

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/StateRepository.kt
```diff
@@ -55,15 +55,41 @@ interface StateRepository {
...
+     * @param start Lower bound for the time  filter.
+     * @param finish Upper bound for the time filter.
      * @param key The name of the key in the metadata to apply the comparison on.
      * @param operation The comparison operation to perform.
      * @param value The value to compare against .
      * @return Collection of states found.
      */
-    fun filterByMetadata(entityManager: EntityManager, key: String, operation: Operation, value: Any): Collection<StateEntity>
+    @Suppress("LongParameterList")
+    fun filterByUpdatedBetweenAndMetadata(
```
JamesHR3, line 90: If you put the filters into data classes you can get rid of the suppression here.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
jujoramos: Sounds about right, thanks 👍

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/SearchOptions.kt
```diff
@@ -0,0 +1,17 @@
...
+enum class Operation {
+    Equals,
+    NotEquals,
+    LesserThan,
+    GreaterThan,
+}
+
+data class IntervalFilter(val start: Instant, val finish: Instant)
+
+data class SingleKeyFilter(val key: String, val operation: Operation, val value: Any)
```
JamesHR3: Might want some KDocs on these, e.g. what's the `SingleKeyFilter` against, what're allowed for `value`s, is the time interval inclusive/exclusive etc.

jujoramos: Updated 👍 . The inclusivity or exclusivity for the interval depends on the actual implementation of the query, so I left that part out of the KDdocs for the data class (they are included within `StateManager.updatedBetween()` and `StateManager.findUpdatedBetweenWithMetadataFilter()` though).

JamesHR3 (approved):

PR #4714, opened 2023-09-27
CORE-17367 On `Scheduler` error going `DOWN` prevents k8s replacing the pod

kyriathar:
- on `triggerAndScheduleNext` error take coordinator to `ERROR` since it means a scheduled task is no longer working. Which means this is about the core system so `ERROR` feels justified. It also means k8s will replace the pod.
- remove unneeded try catch as expression


jujoramos (commented):
Would be worth adding a small unit test to verify that the changes works.

jujoramos (approved):
LGTM

driessamyn (approved):
LGTM

PR #4718, opened 2023-09-27
CORE-16200 Multi-Source Event Mediator configuration and factory.

mbrkic-r3:
Multi-Source Event Mediator configuration and factory.

### DIFF
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/statemanager/StateManagerImpl.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/statemanager/StateManagerImpl.kt
@@ -1,0 +1,53 @@
+package net.corda.messaging.mediator.statemanager
+
+import net.corda.libs.statemanager.api.Operation
+import net.corda.libs.statemanager.api.State
+import net.corda.libs.statemanager.api.StateManager
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import java.time.Instant
+
+// TODO This is used temporarily until State Manager implementation is finished
+@Component(service = [StateManager::class])
+class StateManagerImpl  @Activate constructor() : StateManager {
+    private val storage = mutableMapOf<String, State>()
+
+    override fun create(states: Collection<State>): Map<String, Exception> {
+        return states.mapNotNull {
+            storage.putIfAbsent(it.key, it)
+        }.associate { it.key to RuntimeException("State already exists [$it]") }
+    }
+
+    override fun get(keys: Collection<String>): Map<String, State> {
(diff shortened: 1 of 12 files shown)


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/statemanager/StateManagerImpl.kt
```diff
@@ -0,0 +1,53 @@
...
+import net.corda.libs.statemanager.api.Operation
+import net.corda.libs.statemanager.api.State
+import net.corda.libs.statemanager.api.StateManager
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import java.time.Instant
+
+// TODO This is used temporarily until State Manager implementation is finished
+@Component(service = [StateManager::class])
+class StateManagerImpl  @Activate constructor() : StateManager {
```
jujoramos, line 12: The real `StateManagerImpl` has been checked in already, any reason to not use that instead of the mock?

mbrkic-r3: When last checked there was still some stuff missing (I think DB driver) to make it functional in FlowWorker, is this finished now?

jujoramos: I thought I saw the driver being added to the flow worker gradle build in a `PR` from last week but I can't seem to find it now so yep, you're right 👍

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/statemanager/StateManagerImpl.kt
```diff
@@ -0,0 +1,53 @@
...
+    private val storage = mutableMapOf<String, State>()
+
+    override fun create(states: Collection<State>): Map<String, Exception> {
+        return states.mapNotNull {
+            storage.putIfAbsent(it.key, it)
+        }.associate { it.key to RuntimeException("State already exists [$it]") }
```
emilybowe, line 18: will "it" print ok as it's a state (it looks like)?

emilybowe (approved):
LGTM

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/statemanager/StateManagerImpl.kt
mbrkic-r3: Maybe not, but it's not really important, this will soon be deleted (replaced by the real State Manager implementation).

PR #4719, opened 2023-09-27
CORE-17360: Update execute cleanup event to match new avro object

JamesHR3:
### Problem statement

The flow mapper currently uses an in-process scheduler to execute cleanups of the mapper states. This mechanism will cease to work when moving to the new state storage mechanism, and so it must be replaced with the new scheduler implementation built into the DB worker. Doing this requires some small topic schema and Avro changes. A corresponding small change is required in the runtime code to keep things compiling.

### Solution

The new Avro object contains a list of ids to clean up, rather than using one message per key. When the new mechanism is fully implemented this will be populated. At present the old mechanism is still in place, so this just needs to be populated with an empty list.


emilybowe (dismissed):
LGTM

jujoramos (dismissed):

conal-smith (dismissed):

jujoramos (approved):
LGTM

PR #4721, opened 2023-09-27
CORE-17378 Fixed break in the trace logging.

owenstanford:
Fixed an issue in the trace logging, causing the trace chain to break.


vkolomeyko:
Look like test `net.corda.ledger.utxo.token.cache.impl.services.TokenCacheEventProcessorTest` is failing in CI due to these changes.

vkolomeyko (approved):
LGTM

PR #4723, opened 2023-09-27
CORE-17357: Create a ScheduledTaskHandler for processing scheduled tasks in the flow mapper

JamesHR3:
### Problem statement

The flow mapper cleanup logic requires integrating with the new scheduler to function once the new state management functionality is introduced. To do this, there needs to be a component in the flow mapper responsible for listening to scheduled tasks and fanning out the corresponding IDs to the set of mapper processors.

### Solution

Create a new class responsible for this functionality.

At present, this class is not integrated into the flow mapper. This cannot be done until the mapper is moved to use the new state management mechanism.

### Testing

Testing entirely carried out through unit tests.


components/flow/flow-mapper-service/src/main/kotlin/net/corda/session/mapper/service/executor/ScheduledTaskHandler.kt
```diff
@@ -0,0 +1,52 @@
...
+        return batches.map {
+            ExecuteCleanup(it)
+        }
+    }
+
+    private fun getExpiredStateIds() : List<String> {
+        val windowExpiry = clock.instant() - Duration.ofMillis(cleanupWindow)
+        val states = stateManager.getUpdatedBetween(Instant.MIN, windowExpiry)
+        // This should use the state manager lookup API combining a metadata lookup and a time window, but that doesn't
+        // exist yet.
```
jujoramos: I've recently merged the changes, in case you want to rebase and use the new API instead of coming back to this later.

jujoramos (dismissed):

conal-smith (dismissed):

components/flow/flow-mapper-service/src/test/kotlin/net/corda/session/mapper/service/executor/ScheduledTaskHandlerTest.kt
```diff
@@ -0,0 +1,128 @@
...
+        val scheduledTaskHandler = ScheduledTaskHandler(
+            StateManagerImpl(),
+            clock,
+            window * 5
+        )
+        val output = scheduledTaskHandler.process()
+        assertThat(output).isEmpty()
+    }
+
+    private inner class StateManagerImpl : StateManager {
```
ben-millar: Maybe this should be named `MockStateManagerImpl` or similar for clarity; ~~could cause confusion if people are in the base class looking for implementations.~~

EDIT: It's a private inner class so it probably won't be seen from elsewhere so this isn't a big deal, I guess.

ben-millar (dismissed):

ben-millar (approved):

jujoramos (approved):

PR #4726, opened 2023-09-28
CORE-17382 Allow e2e test config manager to merge with existing config

charlieR3:
The current implementation of the test config manager requires a test to specify the full config override per test. This means we can’t easily set a global configuration for the full test run e.g. disabling the certificate revocation checks.

Recently the tests were updated so that the revocation check was disabled per test so that other gateway configurations could also be set without removing the revocation checks. This also meant that revocation checks were disabled at the end of every test causing the gateway logs to become littered with certificate revocation check errors. 

In this PR, the config manager has been updated to merge configuration overrides set in a test with any existing overrides, so that we can set global configurations for all tests and still allow individual tests to change configs in the same section.


dimosr (approved):
LGTM

dimosr:
I just realised the comment I left on https://github.com/corda/corda-e2e-tests/pull/244 might have an impact on this one as the utils are here.

ronanbrowne (approved):
LGTM - blt

PR #4731, opened 2023-09-28
CORE-17487 Parallel flow pipeline and Task Manager

lankydan:
Add support for parallel flows executing within the flow pipeline.

This includes:
- Temporary multi-thread implementation of the state and event pattern.
- Using the `SameThreadExecutor` for flow fibers, which causes the fiber to run on the thread that started the fiber.
- Using a Kryo Pool for Kryo instances, as kryo instances are not thread-safe themselves.

Add `TaskManager` and a `TaskManagerFactory` which is a wrapper around a `ScheduledExecutorService` used for gathering metrics on executing tasks. The built-in Micrometer metrics for executor services are used, along with some additional metrics in the `TaskManager` that are more specific to our use-cases.


lankydan:
Metrics exposed by the micrometer executor service metrics (couldn't find documentation on them, so had to look them up in grafana):

| Metric                                                  | Type      | Description                                                                           |
|---------------------------------------------------------|-----------|---------------------------------------------------------------------------------------|
| corda_taskmanager_executor_active_threads	              | gauge	    | The approximate number of threads that are actively executing tasks	                  |
| corda_taskmanager_executor_completed_tasks_total	      | counter	  | The approximate total number of tasks that have completed execution	                  |
| corda_taskmanager_executor_idle_seconds_count	          | counter	  | Count of events that have been observed for the base metric ()	                      |
| corda_taskmanager_executor_idle_seconds_max	            | gauge		  |                                                                                       |
| corda_taskmanager_executor_idle_seconds_sum	            | counter	  | Total sum of all observed values for the base metric ()	                              |
| corda_taskmanager_executor_pool_core_threads	          | gauge	    | The core number of threads for the pool                                               |
| corda_taskmanager_executor_pool_max_threads	            | gauge	    | The maximum allowed number of threads in the pool	                                    |
| corda_taskmanager_executor_pool_size_threads	          | gauge	    | The current number of threads in the pool	                                            |
| corda_taskmanager_executor_queue_remaining_tasks	      | gauge	    | The number of additional elements that this queue can ideally accept without blocking	|
| corda_taskmanager_executor_queued_tasks	                | gauge	    | The approximate number of tasks that are queued for execution	                        |
| corda_taskmanager_executor_scheduled_once_total	        | counter		|                                                                                       |
| corda_taskmanager_executor_scheduled_repetitively_total	| counter		|                                                                                       |
| corda_taskmanager_executor_seconds_count	              | counter	  | Count of events that have been observed for the base metric ()                        |	
| corda_taskmanager_executor_seconds_max	                | gauge		  |                                                                                       |
| corda_taskmanager_executor_seconds_sum	                | counter	  | Total sum of all observed values for the base metric ()                               |

The metric includes a tag which is the name given to a task manager, for example "Flow state and event". This can be used to distinguish between different task managers used in different scenarios.

libs/task-manager/build.gradle
```diff
@@ -0,0 +1,21 @@
+plugins {
+    id 'corda.common-publishing'
+    id 'corda.common-library'
+}
+
+description 'Task manager'
+
+dependencies {
+    compileOnly 'org.osgi:osgi.annotation:8.0.1'
```
ronanbrowne: Minor, but my understanding is this dependency version is provided by the java platform sub-module we have in[ corda-api ](https://github.com/corda/corda-api/blob/release/os/5.1/corda-api/build.gradle#L44-L48) which this module also depends on, so expressing a version here is not necessary, much like you have done already with `org.jetbrains.kotlin:kotlin-osgi-bundle` , it's taken from the API layer

ronanbrowne (changes requested):

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/factory/MultiSourceEventMediatorFactoryImpl.kt
```diff
@@ -37,7 +48,11 @@ class MultiSourceEventMediatorFactoryImpl @Activate constructor(
             serializer,
             stateDeserializer,
             stateManager,
-            taskManager,
+            taskManagerFactory.createThreadPoolTaskManager(
```
mbrkic-r3, line 51: @Omar-awad Do we want to use one `TaskManager` instance per worker or it's fine that event mediator has it's own?

libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
```diff
@@ -656,6 +656,18 @@ object CordaMetrics {
...
         }
+
+        object TaskManager {
+            /**
+             * Time it took to execute a task, includes time waiting to be scheduled.
+             */
+            object TaskCompletionTime : Metric<Timer>("taskmanager.completion.time", CordaMetrics::timer)
+
+            /**
+             * Time it took to execute a task, includes time waiting to be scheduled.
```
mbrkic-r3: This seems to be wrong (copy of above)

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/impl/KryoCheckpointSerializerBuilderImpl.kt
```diff
@@ -82,16 +80,21 @@ class KryoCheckpointSerializerBuilderImpl(
...
-            kryoFactory.apply(classResolver),
-            serializers + publicKeySerializers + otherCustomSerializers,
-            classSerializer
-        )
-
-        return KryoCheckpointSerializer(kryo).also {
-            // Clear the builder state
-            serializers.clear()
-            singletonInstances.clear()
+        val pool = object : Pool<Kryo>(true, false, 8) {
```
mbrkic-r3, line 83: Parameter `maximumCapacity` shouldn't be hardcoded. Number of parallel flow executions should be configurable   (and also used by flow engine instance of event mediator).

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/factory/MultiSourceEventMediatorFactoryImpl.kt
```diff
@@ -37,7 +48,11 @@ class MultiSourceEventMediatorFactoryImpl @Activate constructor(
...
+                name = "MultiSourceEventMediator",
+                threadName = "multi-source-event-mediator",
+                threads = 8
```
mbrkic-r3: Parameter `threads` shouldn't be hardcoded but come from `EventMediatorConfig`.

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/impl/KryoCheckpointSerializerBuilderImpl.kt
```diff
@@ -82,16 +80,21 @@ class KryoCheckpointSerializerBuilderImpl(
...
+            override fun create(): Kryo {
+                this.peak
```
mbrkic-r3: Is this needed?

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/factory/FlowFiberFactoryImpl.kt
```diff
@@ -31,10 +28,12 @@ class FlowFiberFactoryImpl @Activate constructor(
         private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
     }
 
-    private val currentScheduler: FiberScheduler = FiberExecutorScheduler(
-        "Same thread scheduler",
-        ScheduledSingleThreadExecutor()
-    )
+    private val currentThreadFiberExecutor = object : FiberExecutorScheduler("Flow Fiber scheduler", SameThreadExecutor.getExecutor()) {
```
mbrkic-r3, line 31: What happens when current thread is used to run `Fiber`, is it blocked until `Fiber` is finished? I'm wondering whether timeout detection will work while `Fiber` is running (i.e. whether line below is not executed after `Fiber` is already finished).
https://github.com/corda/corda-runtime-os/blob/release/os/5.1/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowExecutionPipelineStage.kt#L94

lankydan: It blocks the current thread.

The same thread executor literally calls the fiber's run method and returns once its finished.

If we want timeout detection then we're probably better off using a normal thread pool instead.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/factory/MultiSourceEventMediatorFactoryImpl.kt
lankydan: Raised https://github.com/corda/corda-api/pull/1287 to support this change.

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/impl/KryoCheckpointSerializerBuilderImpl.kt
lankydan: Not hard coding this is going to be a larger change since this configuration of the kryo serializer is pretty deep down and I can't see (right now) how to pass a config object to it.

Can we merge this PR as is and we can raise a ticket to sort out the configuration?

mbrkic-r3: Yes, please rise a ticket for it.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/factory/FlowFiberFactoryImpl.kt
mbrkic-r3: Could you please create a ticket to take a look how to solve that later.

libs/task-manager/build.gradle
```diff
@@ -0,0 +1,21 @@
...
+
+description 'Task manager'
+
+dependencies {
+    compileOnly 'org.osgi:osgi.annotation'
+
+    implementation platform("net.corda:corda-api:$cordaApiVersion")
+    implementation project(":libs:metrics")
+    implementation project(":libs:utilities")
+    implementation "com.google.guava:guava:$guavaVersion"
```
driessamyn: Is it worth bringing Guava in just for the `ThreadFactoryBuilder`? Isn't that one of the libs we always have trouble patching etc?

driessamyn (dismissed):
LGTM

libs/task-manager/build.gradle
lankydan: I can try take it out, its just the way I commonly setup thread names.

lankydan:
API PR - https://github.com/corda/corda-api/pull/1287

ronanbrowne (dismissed):
LGTM re buid / my inial comments

libs/serialization/serialization-kryo/src/main/kotlin/net/corda/kryoserialization/impl/KryoCheckpointSerializerBuilderImpl.kt
rick-r3: There's no need to set an upper bound is there?  The default is `Integer.MAX_VALUE`.  It only makes instances if none are available, and so it will make the _minimum_ necessary.    Unless you are trying to constrain parallelism using this bound.... but a thread pool would be a more natural way to do that.

libs/serialization/serialization-kryo/src/test/kotlin/net/corda/kryoserialization/KryoCheckpointSerializerTest.kt
```diff
@@ -374,4 +375,12 @@ class KryoCheckpointSerializerTest {
             Class.forName(classCaptor.lastValue)
         }
     }
+
+    private fun Kryo.toPool(): Pool<Kryo> {
+        return object : Pool<Kryo>(true, false, 4) {
```
rick-r3, line 380: Again, presumably no need to define an upper bound.

libs/serialization/serialization-kryo/src/test/kotlin/net/corda/kryoserialization/serializers/SingletonSerializeAsTokenSerializerTest.kt
```diff
@@ -68,4 +69,11 @@ class SingletonSerializeAsTokenSerializerTest {
         }
     }
 
+    private fun Kryo.toPool(): Pool<Kryo> {
+        return object : Pool<Kryo>(true, false, 4) {
```
rick-r3, line 73: No need for upper bound

driessamyn (approved):
LGTM

PR #4732, opened 2023-09-28
CORE-15064: Rename `-v` option in `Verify` CLI plugin

vkolomeyko:
As it is clashing with standard option to display version information of the CLI.

Strictly speaking it is a backward compatibility braking change as far CLI tool is concerned. But I cannot think of any better way of doing it given that we clashed with standard CLI options which are available in the rest of the CLI commands.


kyriathar (approved):
OK (packaging)

PR #4733, opened 2023-09-28
CORE-17397: Turn off tracing logs in kubernetes

josephzunigadaly:
This PR turns off logging trace spans in both the console and helm chart log4j configuration. These trace spans are intended to go into a log file and they should be switched off by default.

It also changes the existing configuration to be very specific about which class we want to send to the `corda-trace-spans` log file. So we don't sweep logs related to the operation of tracing into that file.


davidcurrie (approved):
👍 for @corda/cluster-management

kyriathar (approved):
OK (sandboxing)

PR #4734, opened 2023-09-28
CORE-9031: vNode upgrade - forceUpgrade parameter to override operation in progress

Tom-Fitzpatrick:
This adds a "forceUpgrade" query parameter to the vNode upgrade endpoint which allows the upgrade to bypass validating that there is no OperationInProgress on the vNode.

This is being added to allow fixing a vNode that has failed during an operation and been left in an invalid state with a non null OperationInProgress field. By allowing an upgrade to be forced, this will provide a way to reset the OperationInProgress in this scenario. 

Change to the VirtualNodeUpgradeRequest Avro schema can be found here: https://github.com/corda/corda-api/pull/1263


gradle.properties
```diff
@@ -45,8 +45,8 @@ commonsLangVersion = 3.12.0
 commonsTextVersion = 1.10.0
 bouncycastleVersion=1.73
 # Corda API libs revision (change in 4th digit indicates a breaking change)
-# Change to 5.0.0.xx-SNAPSHOT to pick up maven local published copy
-cordaApiVersion=5.1.0.26-beta+
+# Change to 5.1.0.xx-SNAPSHOT to pick up maven local published copy
+cordaApiVersion=5.1.0.27-alpha-1696263944025
```
Tom-Fitzpatrick: Will update this once the corda-api PR is merged

processors/rest-processor/src/integrationTest/resources/swaggerBaseline-v1.json
```diff
@@ -3860,6 +3860,17 @@
             "nullable" : false,
             "example" : "string"
           }
+        }, {
+          "name" : "forceupgrade",
```
vkolomeyko: Hm... we probably should not modify `v1` i.e. C5.0 OpenApi.
We probably should leave `v1` API as it was as it was and introduce a new parameter in `v5_1` API only.

libs/virtual-node/virtual-node-endpoints/src/main/kotlin/net/corda/libs/virtualnode/endpoints/v1/VirtualNodeRestResource.kt
```diff
@@ -116,6 +117,12 @@ interface VirtualNodeRestResource : RestResource {
...
         virtualNodeShortId: String,
         @RestPathParameter(description = "The file checksum of the CPI to upgrade to.")
-        targetCpiFileChecksum: String
+        targetCpiFileChecksum: String,
+        @RestQueryParameter(
+            description = "Whether this upgrade should be forced regardless of OperationInProgress.",
+            default = "false",
+            required = false
+        )
+        forceUpgrade: Boolean = false
```
vkolomeyko, line 147: I think we should leave this method signature as it was, but change the annotation to constrain it to `C5_0`. Then we can create a new method which will start at `C5_1` version with this additional parameter.

Tom-Fitzpatrick: Thanks, will do. I wasn't sure if it would merit the versioning, but that makes sense. I'll make this change and update the swagger Baseline.

processors/rest-processor/src/integrationTest/resources/swaggerBaseline-v1.json
```diff
@@ -4006,6 +4006,7 @@
             "example" : "string"
           },
           "status" : {
+            "type" : "object",
```
vkolomeyko: These changes are non-essential.
For clarity, I would suggest revert any changes to this file completely.

Tom-Fitzpatrick: Done, thanks

processors/rest-processor/src/integrationTest/resources/swaggerBaseline-v1.json
```diff
@@ -5542,4 +5542,3 @@
     }
   }
 }
-
```
vkolomeyko: Mega-nit: Still minor change at the end :-)

vkolomeyko (dismissed):
One minor comment posted.

processors/rest-processor/src/integrationTest/resources/swaggerBaseline-v1.json
Tom-Fitzpatrick: Oop, got it now

vkolomeyko (dismissed):
LGTM

components/virtual-node/virtual-node-rest-service-impl/src/main/kotlin/net/corda/virtualnode/rest/impl/v1/VirtualNodeRestResourceImpl.kt
```diff
@@ -261,20 +261,40 @@ internal class VirtualNodeRestResourceImpl(
...
+        targetCpiFileChecksum: String,
+        forceUpgrade: Boolean
+    ): AsyncResponse {
         val currentVirtualNode = virtualNodeValidationService.validateAndGetVirtualNode(virtualNodeShortId)
         val currentCpi = requireNotNull(cpiInfoReadService.get(currentVirtualNode.cpiIdentifier)) {
             "Current CPI ${currentVirtualNode.cpiIdentifier} associated with virtual node $virtualNodeShortId was not found."
         }
 
-        if (currentCpi.fileChecksum.toHexString().slice(targetCpiFileChecksum.indices) == targetCpiFileChecksum) {
+        if (!forceUpgrade && currentCpi.fileChecksum.toHexString().slice(targetCpiFileChecksum.indices) == targetCpiFileChecksum) {
```
josephzunigadaly: I don't think we should apply `forceUpgrade` here. The suggestion to add an override was for `operationInProgress` and this is a different validation check.

components/virtual-node/virtual-node-rest-service-impl/src/test/kotlin/net/corda/configuration/rest/impl/v1/VirtualNodeRestResourceImplTest.kt
```diff
@@ -145,6 +146,28 @@ class VirtualNodeRestResourceImplTest {
         }
     }
 
+    @Test
+    fun `upgradeVirtualNode succeeds trying to upgrade to same CPI as current one if forceUpgrade is true`() {
```
josephzunigadaly: If we apply my previous comment, we'd also need to remove this test.

components/virtual-node/virtual-node-rest-service-impl/src/main/kotlin/net/corda/virtualnode/rest/impl/v1/VirtualNodeRestResourceImpl.kt
Tom-Fitzpatrick: My thinking here was that `forceUpgrade` would be used to allow retrying failed upgrades that caused the vNode to get into a bad state, which we block otherwise. But true, this is a different change to the one required. I'll remove this now and it can always be added later

josephzunigadaly (dismissed):

vkolomeyko (approved):
LGTM

PR #4735, opened 2023-09-28
CORE-17009 Manage Multi-Source Event Mediator Consumers and Clients

mbrkic-r3:
Managing consumers, messaging clients and message router for Multi-Source Event Mediator:

- Creating them before starting to process messages
- Recreating them to recover from errors (when needed, based on error type)
- Closing them

### DIFF
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
@@ -5,9 +5,17 @@ import net.corda.avro.serialization.CordaAvroSerializer
 import net.corda.libs.statemanager.api.StateManager
 import net.corda.lifecycle.LifecycleCoordinatorFactory
 import net.corda.lifecycle.LifecycleCoordinatorName
+import net.corda.lifecycle.LifecycleStatus
+import net.corda.messaging.api.exception.CordaMessageAPIIntermittentException
+import net.corda.messaging.api.mediator.MediatorConsumer
+import net.corda.messaging.api.mediator.MessageRouter
+import net.corda.messaging.api.mediator.MessagingClient
 import net.corda.messaging.api.mediator.MultiSourceEventMediator
 import net.corda.messaging.api.mediator.config.EventMediatorConfig
 import net.corda.messaging.api.mediator.taskmanager.TaskManager
+import net.corda.messaging.mediator.factory.MediatorComponentFactory
+import org.slf4j.LoggerFactory
+import java.util.UUID
 
 // TODO This will be implemented with CORE-15754
 @Suppress("LongParameterList", "unused_parameter")
@@ -18,7 +26,23 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
     private val stateManager: StateManager,
     private val taskManager: TaskManager,
     lifecycleCoordinatorFactory: LifecycleCoordinatorFactory,
-): MultiSourceEventMediator<K, S, E> {
+) : MultiSourceEventMediator<K, S, E> {
+
+    private val log = LoggerFactory.getLogger("${this.javaClass.name}-${config.name}")
+
+    private var consumers = listOf<MediatorConsumer<K, E>>()
+    private var clients = listOf<MessagingClient>()
+    private lateinit var messageRouter: MessageRouter
+    private val mediatorComponentFactory = MediatorComponentFactory(
+        config.messageProcessor, config.consumerFactories, config.clientFactories, config.messageRouterFactory
+    )
+    private val uniqueId = UUID.randomUUID().toString()
+    private val lifecycleCoordinatorName = LifecycleCoordinatorName(
+        "MultiSourceEventMediator--${config.name}", uniqueId
+    )
+    private val lifecycleCoordinator =
+        lifecycleCoordinatorFactory.createCoordinator(lifecycleCoordinatorName) { _, _ -> }
+
     override val subscriptionName: LifecycleCoordinatorName
         get() = TODO("Not yet implemented")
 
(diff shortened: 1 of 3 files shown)


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
```diff
@@ -18,15 +26,89 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
     private val stateManager: StateManager,
     private val taskManager: TaskManager,
     lifecycleCoordinatorFactory: LifecycleCoordinatorFactory,
-): MultiSourceEventMediator<K, S, E> {
+) : MultiSourceEventMediator<K, S, E> {
+
+    private val log = LoggerFactory.getLogger("${this.javaClass.name}-${config.name}")
```
emilybowe, line 31: should the logger be in a companion object? 🤔

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
```diff
@@ -18,15 +26,89 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
...
+            try {
+                consumers = mediatorComponentFactory.createConsumers(::onSerializationError)
+                clients = mediatorComponentFactory.createClients(::onSerializationError)
+                messageRouter = mediatorComponentFactory.createRouter(clients)
+
+                consumers.forEach { it.subscribe() }
+                lifecycleCoordinator.updateStatus(LifecycleStatus.UP)
+
+                while (!stopped) {
+                    // Poll and process events
```
emilybowe, line 75: do you need another internal while loop for !stopped here?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/factory/MediatorComponentFactory.kt
```diff
@@ -0,0 +1,80 @@
...
+     * Creates message consumers.
+     *
+     * @param onSerializationError Function for handling serialization errors.
+     * @return List of created [MediatorConsumer]s.
+     */
+    fun createConsumers(
+        onSerializationError: (ByteArray) -> Unit
+    ): List<MediatorConsumer<K, E>> {
+        check(consumerFactories.isNotEmpty()) {
+            "None consumer factory set in configuration"
```
emilybowe: I'd change `none -> no`

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/factory/MediatorComponentFactory.kt
```diff
@@ -0,0 +1,80 @@
...
+     * Creates messaging clients.
+     *
+     * @param onSerializationError Function for handling serialization errors.
+     * @return List of created [MessagingClient]s.
+     */
+    fun createClients(
+        onSerializationError: (ByteArray) -> Unit
+    ): List<MessagingClient> {
+        check(clientFactories.isNotEmpty()) {
+            "None client factory set in configuration"
```
emilybowe: here also `None -> no`

libs/messaging/messaging-impl/src/test/kotlin/net/corda/messaging/mediator/factory/MediatorComponentFactoryTest.kt
```diff
@@ -0,0 +1,166 @@
...
+        )
+    }
+
+    @Test
+    fun `successfully creates consumers`() {
+        val onSerializationError: (ByteArray) -> Unit = {}
+
+        val mediatorConsumers = mediatorComponentFactory.createConsumers(onSerializationError)
+
+        Assertions.assertEquals(consumerFactories.size, mediatorConsumers.size)
```
emilybowe: can you import the Assertions library and then not have to do `Assertions.` prefix ?

emilybowe (dismissed):
LGTM, some nit picks

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
mbrkic-r3: Logger includes mediator's name, so we can distinguish logs for different mediator clients (e.g. flow engine or flow mapper).

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
mbrkic-r3: The first loop is for re-creating consumers/clients in case of errors, the second is for processing polled events. They both have the same check as we want to exit both if mediator is stopped. This might look a bit different in the end, depending on TaskManager implementation, which should be providing/running mediator's main thread and it's API (might use another mechanism to ask mediator to stop).

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/factory/MediatorComponentFactory.kt
mbrkic-r3: fixed

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/factory/MediatorComponentFactory.kt
mbrkic-r3: fixed

libs/messaging/messaging-impl/src/test/kotlin/net/corda/messaging/mediator/factory/MediatorComponentFactoryTest.kt
mbrkic-r3: fixed

emilybowe (approved):
LGTM, assuming 🟢  build

PR #4739, opened 2023-09-29
CORE-16202 Multi-Source Event Mediator - Routing messages to messaging clients

mbrkic-r3:
Created endpoint property that will be set by messaging router and used by multi-source event mediator messaging clients.


emilybowe (approved):
LGTM (pending 🟢 )

PR #4741, opened 2023-09-29
CORE-15749 : Changes to replace hardcoded session keys

SaurabhR3:
1. Replaced hard coded session keys with newly generated one.
2. Added new key category INTEROP_SESSION
3. Some refactoring.

### DIFF
--- a/components/interop/interop-identity-write-service-impl/src/main/kotlin/net/corda/interop/identity/write/impl/SessionKeyGenerator.kt
+++ b/components/interop/interop-identity-write-service-impl/src/main/kotlin/net/corda/interop/identity/write/impl/SessionKeyGenerator.kt
@@ -18,7 +18,7 @@ class SessionKeyGenerator(
     private val hsmRegistrationClient: HSMRegistrationClient
 ) {
     companion object {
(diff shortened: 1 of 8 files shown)


components/interop/interop-identity-write-service-impl/src/main/kotlin/net/corda/interop/identity/write/impl/SessionKeyGenerator.kt
```diff
@@ -0,0 +1,71 @@
...
+        private const val CATEGORY = "INTEROP_SESSION"
+        private const val ALIAS = "alias"
+        private const val SCHEME = "CORDA.ECDSA.SECP256R1"
+    }
+
+    fun getOrCreateSessionKey(
+        identityToPublish: InteropIdentity,
+        vNodeShortHash: String
+    ): String {
+        val alias = "INTEROP-SESSION-${identityToPublish.x500Name}"
```
szymonsztuka, line 30: More an observation: this may be fine however by using the X500 name and not an interop application name we constraint to have a single key for all Interop Groups of an identity.

ronanbrowne (approved):
LGTM (blt files)

PR #4742, opened 2023-09-29
CORE-17010 Multi-Source Event Mediator - Polling Events

mbrkic-r3:
- Implemented polling consumers and committing offsets to Multi-Source Event Mediator.
- Function `poll` of `MediatorConsumer` made asynchronious.

### DIFF
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
@@ -1,11 +1,13 @@
 package net.corda.messaging.mediator
 
+import kotlinx.coroutines.runBlocking
 import net.corda.avro.serialization.CordaAvroDeserializer
 import net.corda.avro.serialization.CordaAvroSerializer
 import net.corda.libs.statemanager.api.StateManager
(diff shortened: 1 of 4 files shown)


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
```diff
@@ -111,4 +113,33 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
         consumers.forEach { it.close() }
         clients.forEach { it.close() }
     }
+
+    private fun processEvents() {
+        log.debug { "Polling and processing events" }
+        val messages = pollConsumers()
+        if (messages.isNotEmpty()) {
+            // TODO Process messages
```
emilybowe, line 121: should we commit this TODO ?

emilybowe (approved):
LGTM pending 🟢

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
mbrkic-r3: This will be implemented with another PR (there are several tickets in Multi-Source Mediator Pattern epic that are related to processing events)

PR #4743, opened 2023-09-29
CORE-17358: Create a class responsible for cleaning up mapper states

JamesHR3:
### Problem statement

The flow mapper cleanup mechanism needs to be replaced, as the current mechanism will no longer function when the mapper is migrated to use the new state manager and event mediator APIs. Completing this requires a class responsible for picking up `ExecuteCleanup` events and removing the corresponding states.

### Solution

Create a handler for these events.

For now, this class is not plugged in anywhere. No assumptions have been made about where these events are published, although it is assumed they will be sent to a new topic to keep the cleanup events off the main flow mapper event topic.

### Testing

Unit tests for now.


components/flow/flow-mapper-service/src/main/kotlin/net/corda/session/mapper/service/executor/CleanupProcessor.kt
```diff
@@ -0,0 +1,31 @@
...
+        val failed = stateManager.delete(states.values)
+        if (failed.isNotEmpty()) {
+            logger.info(
+                "Failed to delete ${failed.size} mapper states when executing a cleanup event. Failed IDs: ${
+                    failed.keys.joinToString(
+                        ","
+                    )
+                }"
+            )
+        }
```
jujoramos, line 29: Should the processor return the failed events back or similar so something else can eventually retry the deletion?.

JamesHR3: My logic here is that the scheduled task trigger will re-discover these the next time it fires, and we'll attempt to re-delete them at that point.

jujoramos: 👍

jujoramos (approved):
LGTM

components/flow/flow-mapper-service/src/main/kotlin/net/corda/session/mapper/service/executor/CleanupProcessor.kt
mbrkic-r3: Should `CleanupProcessor ` have a name that is used for logging so it's easier to see what went wrong, e.g. whether we had issue with cleanup of flow events or flow mapper events (assuming we'll have an instance for each use case and both might be using random UUIDs for state IDs so it's hard to figure out from log where is the issue)?

PR #4744, opened 2023-09-29
CORE-17500 - The same claimed tokens must be returned when a claim is processed more than once

filipesoliveira:


components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/entities/internal/PoolCacheStateImpl.kt
```diff
@@ -33,6 +33,11 @@ class PoolCacheStateImpl(
         return cacheState.tokenClaims.any { it.claimId == claimId }
     }
 
+    override fun claim(claimId: String): TokenClaim? {
+        // There must only exist one claim
+        return cacheState.tokenClaims.filter { it.claimId == claimId }.firstOrNull()
```
vkolomeyko:
```suggestion
        return cacheState.tokenClaims.filter { it.claimId == claimId }.singleOrNull()
```

filipesoliveira: Thanks, @vkolomeyko

components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/handlers/TokenClaimQueryEventHandler.kt
```diff
@@ -19,12 +22,42 @@ class TokenClaimQueryEventHandler(
...
+
+            return recordFactory.getSuccessfulClaimResponse(
+                event.flowId,
+                event.externalEventRequestId,
+                event.poolKey,
+                claim.claimedTokens.map {
+                    DbCachedToken(it.stateRef,
+                        BigDecimal(
+                            BigInteger(
+                                ByteArray(it.amount.unscaledValue.remaining())
```
filipesoliveira: There is a possible issue in this code.

filipesoliveira: There is no issue.

vkolomeyko (commented):
LGTM

But I am vaguely familiar with this code. Possibly, needs an input from @owenstanford 
Also, do we need a unit test to pin the expected behaviour in production code?

components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/handlers/TokenClaimQueryEventHandler.kt
```diff
@@ -19,12 +22,42 @@ class TokenClaimQueryEventHandler(
...
+        val claim = state.claim(claimId)
+        if(claim != null) {
+
+            logger.warn("A token claim is being processed more than once. ClaimId: $claimId")
+
+            return recordFactory.getSuccessfulClaimResponse(
+                event.flowId,
+                event.externalEventRequestId,
+                event.poolKey,
+                claim.claimedTokens.map {
```
owenstanford: We don't need to do this, the RecordFactory converts the CachedToken back to an avro object, so we are converting from avro -> DTO -> avro, we can just add an overload for getSuccessfulClaimResponse that takes a list of avro tokens, this will simplify this code.
We should also add a unit test to cover this change for the claim handler (and the record factory)

owenstanford (approved):
LGTM

PR #4745, opened 2023-09-29
Monitor non-cluster database connections

dickon:
Previousy we only monitored the cluster database connection internal to the DbConnectionManager. If passwords changed on the other connections we will keep trying the old values and not go unhealthy.

This means that the lifecycle will go down in all cases if the database password is rotated, which gives us a chance to reread it, via Kubernetes restarting the whole worker.

A better solution which picks up passwords dynamically seems to require a bigger change.

Add docstrings and improve the clarity of some code.


driessamyn (changes requested):
Can we discuss this?

I don't think we can do this as this would effectively keep all the VNode connections "alive", preventing the pools giving back all their connections.
We can do this for cluster connection, but not for VNode connections, which should get closed after a time being idle.

CC @simon-johnson-r3 - I think this is an issue since we now cache _all_ connection pools (compared to we used to only cache the cluster ones).
I agree with @dickon that we need a way to force those "failed" connections to be replaced (which happened all time previously before the cache was put in), but we can't keep these alive un-necessarily.

PR #4746, opened 2023-09-30
CORE-17387 - flow checkpoint maintenance events

driessamyn:
Create a component in the flow service responsible for processing scheduled task triggers and generating checkpoint maintenance events.
This includes integration with the State Manager component and as this is the first component that uses it, this change includes a number of small changes to that library.

API change: https://github.com/corda/corda-api/pull/1272


components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/FlowMaintenanceImpl.kt
```diff
@@ -0,0 +1,76 @@
...
+    private val stateManagerFactory: StateManagerFactory,
+) : FlowMaintenance {
+    companion object {
+        private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+    }
+
+    private val coordinator = coordinatorFactory.createCoordinator<FlowExecutor>(::eventHandler)
+    override fun onConfigChange(config: Map<String, SmartConfig>) {
+        val messagingConfig = config.getConfig(ConfigKeys.MESSAGING_CONFIG)
+        // TODO - fix config key. The state manager has nothing to do with messaging.
```
driessamyn: @conalsmith-r3 , @jujoramos please advice here.
I think we need a `ConfigKey` for the StateManager.

components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/SessionTimeoutTaskProcessor.kt
```diff
@@ -0,0 +1,46 @@
...
+    override val valueClass: Class<ScheduledTaskTrigger>
+        get() = ScheduledTaskTrigger::class.java
+
+    override fun onNext(events: List<Record<String, ScheduledTaskTrigger>>): List<Record<*, *>> {
+        // If we receive multiple, there's probably an issue somewhere, and we can ignore all but the last one.
+        return events.lastOrNull { it.key == ScheduledTask.SCHEDULED_TASK_NAME_SESSION_TIMEOUT }?.value?.let { trigger ->
+            logger.trace("Processing trigger scheduled at ${trigger.timestamp}")
+            // TODO - temporary query
+            // TODO - we must be able to specify additional filters so we can limit to selecting those sessions that are still open
+            // TODO - we must be able to limit by type of state
```
driessamyn, line 35: @jujoramos , @conalsmith-r3 , please advice how this can be implemented.

components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/SessionTimeoutTaskProcessor.kt
```diff
@@ -0,0 +1,46 @@
...
+            val checkpoints = stateManager.find(
+                SingleKeyFilter("session.expiry", Operation.LesserThan, now())
+            )
+            if (checkpoints.isEmpty()) {
+                logger.trace("No flows to time out")
+                emptyList()
+            } else {
+                // TODO - return an avro message (schema TBC) for each checkpoint
+                // TODO - define topic to publish message on
```
driessamyn: @JamesHR3 , should this be a new topic?
Do we know yet what the schema for the message needs to look like?

applications/examples/sandbox-app/build.gradle
```diff
@@ -65,6 +65,7 @@ dependencies {
     runtimeOnly project(':libs:messaging:db-message-bus-impl')
     runtimeOnly project(':libs:serialization:serialization-checkpoint-api')
     runtimeOnly project(':libs:serialization:serialization-kryo')
+    runtimeOnly project(":libs:state-manager:state-manager-db-impl")
```
jujoramos: Do you need this one here?. If you need it, wouldn't it be better to have the dependency to `state-manager-api` instead?.

components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/FlowMaintenanceImpl.kt
jujoramos: I've created a ticket for this already 👍 .

components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/SessionTimeoutTaskProcessor.kt
jujoramos: We haven't implemented custom filters yet, but there's already a ticket created for it.

components/ledger/ledger-common-flow/build.gradle
```diff
@@ -50,6 +50,7 @@ dependencies {
     integrationTestRuntimeOnly project(':libs:messaging:db-message-bus-impl')
     integrationTestRuntimeOnly project(':libs:messaging:messaging-impl')
     integrationTestRuntimeOnly project(':libs:flows:session-manager-impl')
+    integrationTestRuntimeOnly project(":libs:state-manager:state-manager-db-impl")
```
jujoramos: Any reason not to use `:libs:state-manager:state-manager-api` instead?.

components/ledger/ledger-consensual-flow/build.gradle
```diff
@@ -52,6 +52,7 @@ dependencies {
     integrationTestRuntimeOnly project(':libs:messaging:messaging-impl')
     integrationTestRuntimeOnly project(':libs:lifecycle:lifecycle-impl')
     integrationTestRuntimeOnly project(':libs:flows:session-manager-impl')
+    integrationTestRuntimeOnly project(":libs:state-manager:state-manager-db-impl")
```
jujoramos: Any reason not to use `:libs:state-manager:state-manager-api` instead?.

components/ledger/ledger-utxo-flow/build.gradle
```diff
@@ -73,6 +73,7 @@ dependencies {
     integrationTestRuntimeOnly project(':libs:lifecycle:lifecycle-impl')
     integrationTestRuntimeOnly project(':libs:flows:session-manager-impl')
     integrationTestRuntimeOnly project(':libs:membership:membership-impl')
+    integrationTestRuntimeOnly project(":libs:state-manager:state-manager-db-impl")
```
jujoramos: Any reason not to use `:libs:state-manager:state-manager-api` instead?.

jujoramos (commented):
I know it's still in draft and some changes are blocked by the state manager implementation, just leaving some comments.

applications/examples/sandbox-app/build.gradle
driessamyn: It's a runtime dependency. I.e. there has to be at least 1 implementation of the OSGi "Service" `StateManagerFactory`, otherwise the verify task fails.
It can't be a dependency of the `-api` module as that would suggest that the "only" implementation of this is the DB one.
This is  obv true atm, but I considered whether it would be worth creating a test implementation of the StateManger (e.g. in-memory only). Probably out-of-scope for this, but I can see we may want that at some point.

components/ledger/ledger-common-flow/build.gradle
driessamyn: It's a runtime dependency. These OSGi tests have a dependency on FlowService, which means we must provide an implementation of the State Manager.
This is an example of when a naive in-memory implementation would be useful (and avoid a bunch of additional dependencies).

components/ledger/ledger-consensual-flow/build.gradle
driessamyn: As above

components/ledger/ledger-utxo-flow/build.gradle
driessamyn: As above

applications/examples/sandbox-app/build.gradle
jujoramos: 👍

components/flow/flow-service/build.gradle
```diff
@@ -12,51 +12,52 @@ dependencies {
     compileOnly 'org.osgi:osgi.annotation'
     compileOnly "co.paralleluniverse:quasar-osgi-annotations:$quasarVersion"
 
-    implementation project(":components:configuration:configuration-read-service")
-    implementation project(":components:external-messaging-services")
-    implementation project(":components:membership:membership-group-read")
+    implementation project(':components:configuration:configuration-read-service')
```
jujoramos, line 15: There are not functional changes in this file, and I'm not asking the changes to be reverted, I'm asking just out of curiosity: did you change `'` by `"` because of personal preference?, or are there any internal guidelines we should follow somewhere?.

components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/FlowMaintenanceImpl.kt
```diff
@@ -0,0 +1,81 @@
...
+    private val coordinator = coordinatorFactory.createCoordinator<FlowMaintenance>(::eventHandler)
+    override fun onConfigChange(config: Map<String, SmartConfig>) {
+        // TODO - fix config key (CORE-17437).
+        if(config.containsKey(ConfigKeys.MESSAGING_CONFIG)) {
+            val messagingConfig = config.getConfig(ConfigKeys.MESSAGING_CONFIG)
+            val stateManagerConfig = config.getConfig(ConfigKeys.MESSAGING_CONFIG)
+            coordinator.createManagedResource("FLOW_MAINTENANCE_SUBSCRIPTION") {
+                subscriptionFactory.createDurableSubscription(
+                    SubscriptionConfig(
+                        "flow.maintenance.tasks",
```
jujoramos, line 46: Nitpick: make it a constant?.

libs/state-manager/state-manager-api/src/test/kotlin/net/corda/libs/statemanager/api/MetadataTests.kt
```diff
@@ -0,0 +1,47 @@
...
+                123,
+                true
+            )
+    }
+    @ParameterizedTest
+    @MethodSource("acceptedTypes")
+    fun `accept primitive types`(value: Any) {
+        assertDoesNotThrow {
+            Metadata(mapOf("foo" to value))
+        }
```
jujoramos, line 26: Nitpick: why the `assertDoesNotThrow{}` here?, if the exception is thrown the test will fail anyways (and removing it also removes an unneeded dependency to `jupiter`).

libs/state-manager/state-manager-api/src/test/kotlin/net/corda/libs/statemanager/api/MetadataTests.kt
```diff
@@ -0,0 +1,47 @@
+package net.corda.libs.statemanager.api
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+import org.junit.jupiter.api.assertThrows
```
jujoramos: Nitpick: we are mixing two assertions libraries in the test, can we use only one (`assert4j would be my preferred choice but it's up to you)?

libs/state-manager/state-manager-api/src/test/kotlin/net/corda/libs/statemanager/api/MetadataTests.kt
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    @Test
+    fun `fail all non-primitive types`() {
+        val list = listOf("Na Na Na Na Na Na Na Na", "Batman")
+        val ex = assertThrows<IllegalArgumentException> {
+            Metadata(mapOf("joker" to Superman(1000), "batman" to list))
+        }
+        assertThat(ex).hasMessageContainingAll("joker", "batman", Superman::class.java.name, list.javaClass.name)
```
jujoramos: Maybe use the following instead?:

```
assertThatThrownBy { Metadata(mapOf("joker" to Superman(1000), "batman" to list)) }
  .isExactlyInstanceOf(IllegalArgumentException::class.java)
  .hasMessageContainingAll("joker", "batman", Superman::class.java.name, list.javaClass.name)
```

jujoramos (dismissed):
LGTM, just some minor comments.

components/flow/flow-service/build.gradle
driessamyn: Fair question.
There is a functional change (adding the state manager dependency). I just noticed the file was inconsistent, sometimes using `"` and sometimes `'` so fixed that to the Groovy recommendation: [Single-quoted strings are plain java.lang.String and don’t support interpolation](http://docs.groovy-lang.org/latest/html/documentation/index.html#all-strings)

components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/FlowMaintenanceImpl.kt
driessamyn: It is a constant, just an in-line one ;)

My reasoning always is that if it's just just once, there's not much point in making it a constant as all it does is obscure the real value (i.e. you have to look somewhere else for what it is). When it's used multiple times off course there is a benefit in it being declared only once.

But happy to change if people feel strongly about this.

libs/state-manager/state-manager-api/src/test/kotlin/net/corda/libs/statemanager/api/MetadataTests.kt
driessamyn: It just makes it more explicit I guess. 
The test framework does threat the 2 things different (test failed vs error). 
E.g.

```kotlin
@Test
    fun foo() {
        throw IllegalArgumentException("foo")
    }

    @Test
    fun bar() {
        assertDoesNotThrow {
            throw IllegalArgumentException("bar")
        }
    }
```
 
results in:

image

I agree it technically doesn't make much difference, but semantically maybe it does.

components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/FlowMaintenanceImpl.kt
jujoramos: 👍

components/flow/flow-service/build.gradle
jujoramos: 👍

libs/state-manager/state-manager-api/src/test/kotlin/net/corda/libs/statemanager/api/MetadataTests.kt
driessamyn: I guess the `assertThrows` syntax is a bit more natural than assertJ's `assertThatThrownBy` and is more common across the codebase:

```
❯ grep -r --include "*.kt" assertThat | wc -l
    9129
❯ grep -r --include "*.kt" assertThrows | wc -l
    1792
❯ grep -r --include "*.kt" assertThatThrownBy | wc -l
     369
```

Both libraries seem to be used widely, with a slight preference for the junit assertions according to the imports:

```
❯ grep -r --include "*.kt" "import org.assertj.core.api.Assertions" | wc -l
     852
❯ grep -r --include "*.kt" "import org.junit.jupiter.api.assert" | wc -l
    1169
```

I don't really mind. I don't mind we settle to use one, but using multiple equally doesn't bother me as long as the code is readable.

libs/state-manager/state-manager-api/src/test/kotlin/net/corda/libs/statemanager/api/MetadataTests.kt
driessamyn: Updated 👍

components/flow/flow-service/build.gradle
```diff
@@ -85,13 +86,14 @@ dependencies {
...
     integrationTestRuntimeOnly project(':libs:flows:session-manager-impl')
-    integrationTestRuntimeOnly project(":libs:lifecycle:lifecycle-impl")
+    integrationTestRuntimeOnly project(':libs:lifecycle:lifecycle-impl')
     integrationTestRuntimeOnly project(':libs:messaging:db-message-bus-impl')
     integrationTestRuntimeOnly project(':libs:messaging:messaging-impl')
     integrationTestRuntimeOnly project(':libs:serialization:serialization-checkpoint-api')
     integrationTestRuntimeOnly project(':libs:serialization:serialization-kryo')
-    integrationTestRuntimeOnly project(":components:membership:membership-group-read-impl")
-    integrationTestRuntimeOnly project(":components:virtual-node:cpk-read-service-impl")
+    integrationTestRuntimeOnly project(":libs:state-manager:state-manager-db-impl")
```
JamesHR3: This one is in double quotes.

components/ledger/ledger-common-flow/build.gradle
JamesHR3: Think you've got some double quotes here too (sorry!)

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/Metadata.kt
```diff
@@ -4,28 +4,26 @@ package net.corda.libs.statemanager.api
...
-        java.lang.String::class.java,
-        Number::class.java,
-        java.lang.Number::class.java,
-        Boolean::class.java,
-        java.lang.Boolean::class.java,
-    )
-
-    private fun isPrimitiveOrBoxedValue(value: Any): Boolean {
-        return supportedType.any { it.isAssignableFrom(value.javaClass) }
+    private val map: Map<String, Any> = emptyMap()
```
JamesHR3, line 7: It might be valuable to have some way of generating a new metadata object using the map from an old one? I can see us appending keys to an existing metadata objects when processing flow/flow mapper events.

The comment above the class might need changing too.

JamesHR3 (commented):
Couple of nitpicking things but looks good.

components/ledger/ledger-common-flow/build.gradle
driessamyn: No need to apologise. I should have gotten it "right" first time ;) 
Should be fixed now.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/Metadata.kt
driessamyn: Updated the comment.

Given Metadata is a Map, I think you can just use it like one and I think the Kotlin idiomatic way is to use `plus` in that scenario.
I have added a test to show this works:

```kotlin
@Test
    fun `new meta with additional elements`() {
        val meta1 = Metadata(mapOf("foo" to "bar"))
        assertThat(meta1.plus("batman" to "joker"))
            .containsExactlyInAnyOrderEntriesOf(mapOf("foo" to "bar", "batman" to "joker"))
    }
```

libs/state-manager/state-manager-api/src/test/kotlin/net/corda/libs/statemanager/api/MetadataTests.kt
jujoramos: In that case, and just to keep consistency across this single test (use single assertion library), I would use something like the following:
```
assertThatCode {
   Metadata(mapOf("foo" to value))
}.doesNotThrowAnyException()
```

libs/state-manager/state-manager-api/src/test/kotlin/net/corda/libs/statemanager/api/MetadataTests.kt
jujoramos: Agreed, but I'd argue that we should keep consistency and use a single assertions framework per test class, even though we are not consistent across the codebase.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/Metadata.kt
JamesHR3: Ah, my mistake - I'd missed it implements `Map`.

JamesHR3 (dismissed):

mag1c-48 (dismissed):
LGTM - build files

blsemo (dismissed):
Ledger changes are fine

jujoramos (dismissed):

simon-johnson-r3 (approved):

blsemo (approved):
Re-approved for ledger

JamesHR3 (approved):

mcgovc (approved):
LGTM - build files

PR #4749, opened 2023-10-02
CORE-15064: Rename `-pfv` option in `Verify` CLI plugin

vkolomeyko:
As it can be interpreted as `-p -f -v`.

### DIFF
--- a/tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/Verify.kt
+++ b/tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/Verify.kt
@@ -23,7 +23,7 @@ class Verify : Runnable {
         description = ["Package type (CPK/CPB/CPI)", "Detected from file name extension if not specified"])
     var type: PackageType? = null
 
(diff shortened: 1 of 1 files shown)


anien (dismissed):
Thanks! LGTM.

tools/plugins/package/src/main/kotlin/net/corda/cli/plugins/packaging/Verify.kt
```diff
@@ -23,7 +23,7 @@ class Verify : Runnable {
         description = ["Package type (CPK/CPB/CPI)", "Detected from file name extension if not specified"])
     var type: PackageType? = null
 
-    @CommandLine.Option(names = ["--packageFormatVersion", "-pfv"],
+    @CommandLine.Option(names = ["--packageFormatVersion", "-o"],
```
kyriathar: Could we also change long form option to `--package-format-version`. I can see we don't use camel case in other commands.

kyriathar (approved):

anien (approved):
Nice catch Kyriakos. Agree with this version.

PR #4754, opened 2023-10-02
CORE-16183 Align implementation with ledger data model simplifications (part 1)

nkovacsx:
These are the runtime-os side changes of the data model simplifications from https://github.com/corda/corda-api/pull/1270.


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
```diff
@@ -128,27 +119,14 @@ interface UtxoRepository {
...
     /**
-     * Persists or updates transaction [transactionStatus]. There is only one status per transaction. In case that status already
+     * Updates transaction [transactionStatus]. There is only one status per transaction. In case that status already
      * exists, it will be updated only if old and new statuses are one of the following combinations (and ignored otherwise):
      * - UNVERIFIED -> *
      * - VERIFIED -> VERIFIED
      * - INVALID -> INVALID
      */
-    fun persistTransactionStatus(
+    fun updateTransactionStatus(
```
relyafi, line 129: Is there a reason we changed this? I don't necessarily think it's inaccurate but it does feel inconsistent with the other functions now.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -30,7 +30,7 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
             tc_output.leaf_idx, 
             tc_output_info.data as output_info_data,
             tc_output.data AS output_data 
-            FROM {h-schema}utxo_visible_transaction_state AS rts
+            FROM {h-schema}utxo_visible_transaction_output AS rts
```
relyafi: Wouldn't it make sense to update the aliases in all these queries to reflect the new table names being used?

I know we need to keep some aliases the same for backwards compatibility with ledger custom queries, but I think that only affects the ones in the query provider token classes?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -39,11 +39,11 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
...
                 AND tc_output.leaf_idx = tc_output_info.leaf_idx
                 AND tc_output.group_idx = ${UtxoComponentGroup.OUTPUTS.ordinal}
-            JOIN {h-schema}utxo_transaction_status AS ts
-                ON ts.transaction_id = tc_output.transaction_id
+            JOIN {h-schema}utxo_transaction AS ts
+                ON ts.id = tc_output.transaction_id
             AND rts.consumed IS NULL
             AND ts.status = :verified
-            ORDER BY tc_output.created, tc_output.transaction_id, tc_output.leaf_idx"""
+            ORDER BY ts.updated, tc_output.transaction_id, tc_output.leaf_idx"""
```
relyafi: Using `updated` rather than `created` is a change in behavior, and could be dangerous if a transaction is consumed during query execution. Shouldn't we stick to `created`?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -117,12 +117,19 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
...
-            JOIN {h-schema}utxo_transaction_status AS ts
-                ON ts.transaction_id = tc_output.transaction_id
+            JOIN {h-schema}utxo_transaction AS ts
+                ON ts.id = tc_output.transaction_id
             AND tc_output.transaction_id in (:transactionIds)
             AND (tc_output.transaction_id||':'|| tc_output.leaf_idx) in (:stateRefs)
             AND ts.status = :verified
             AND tc_output_info.group_idx = ${UtxoComponentGroup.OUTPUTS_INFO.ordinal}
-            ORDER BY tc_output.created, tc_output.transaction_id, tc_output.leaf_idx"""
+            ORDER BY ts.updated, tc_output.transaction_id, tc_output.leaf_idx"""
```
relyafi: As earlier - shouldn't we continue to use `created`?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -117,12 +117,19 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
...
+            .trimIndent()
+
+    override val updateTransactionStatus: String
```
relyafi, line 129: I'm not entirely convinced by this logic - it might be fine but it's probably safer to keep the existing `ON CONFLICT` logic we had.

Rather than creating a separate update operation, wouldn't it be safer to adapt `persistTransaction` to behave like `persistTransactionStatus` did?

testing/ledger/ledger-hsqldb/src/integrationTest/kotlin/net/corda/testing/ledger/tests/HsqldbVaultNamedQueryTest.kt
```diff
@@ -107,25 +95,34 @@ class HsqldbVaultNamedQueryTest {
                 id = txId.toString(),
                 privacySalt = byteArrayOf(),
                 accountId = ACCOUNT_ID.toString(),
-                created = timestamp
+                created = timestamp,
+                status = "V",
```
relyafi: Should we use the constant for this instead of a literal string?

components/ledger/ledger-persistence/src/integrationTest/kotlin/net/corda/ledger/persistence/utxo/tests/UtxoPersistenceServiceImplTest.kt
```diff
@@ -256,39 +256,42 @@ class UtxoPersistenceServiceImplTest {
...
         entityManagerFactory.transaction { em ->
 
-            em.createNativeQuery("DELETE FROM {h-schema}utxo_visible_transaction_state").executeUpdate()
+            em.createNativeQuery("DELETE FROM {h-schema}utxo_visible_transaction_output").executeUpdate()
 
             createTransactionEntity(entityFactory, transaction1, status = VERIFIED).also { em.persist(it) }
             createTransactionEntity(entityFactory, transaction2, status = VERIFIED).also { em.persist(it) }
 
-            repository.persistTransactionVisibleStates(
+            repository.persistTransactionOutput(
```
lankydan: Should this be `persistVisibleTransactionOutput` based on the schema?

I'm only calling this out because I saw some instances of `visibleTransactionOutput` and some of `transactionOutput`.

As long as we choose one of them, I'll be happy 👍

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -39,11 +39,11 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
                 ON tc_output.transaction_id = tc_output_info.transaction_id
                 AND tc_output.leaf_idx = tc_output_info.leaf_idx
                 AND tc_output.group_idx = ${UtxoComponentGroup.OUTPUTS.ordinal}
-            JOIN {h-schema}utxo_transaction_status AS ts
-                ON ts.transaction_id = tc_output.transaction_id
+            JOIN {h-schema}utxo_transaction AS ts
```
lankydan: Should `ts` become something else? `t` or something like that?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -30,7 +30,7 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
             tc_output.leaf_idx, 
             tc_output_info.data as output_info_data,
             tc_output.data AS output_data 
-            FROM {h-schema}utxo_visible_transaction_state AS rts
```
lankydan, line 33: Where did the `rts` even come from in these queries? 🤔 

Should `rts` be `vto`? 🤔

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -52,7 +52,7 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
...
+            FROM {h-schema}utxo_visible_transaction_output AS rts
```
lankydan: Similar naming things to previous comment.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
vlajos: Having that logic in persistTransaction query, would let us remove the `updateTransactionStatus` from the end of the "big" `persistTransaction` PersistenceImpl method.
But this is not a new function; it just got moved from the pg specific code to the generic. There is a dedicated API endpoint which uses it.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -117,12 +117,19 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
...
             AND tc_output_info.group_idx = ${UtxoComponentGroup.OUTPUTS_INFO.ordinal}
-            ORDER BY tc_output.created, tc_output.transaction_id, tc_output.leaf_idx"""
+            ORDER BY ts.updated, tc_output.transaction_id, tc_output.leaf_idx"""
+            .trimIndent()
+
+    override val updateTransactionStatus: String
+        get() = """
+            UPDATE {h-schema}utxo_transaction SET status = :newStatus, updated = :updatedAt
+            WHERE id = :transactionId 
+            AND (status = :newStatus OR status = '$UNVERIFIED')"""
```
lankydan, line 133: I'm not 100% sure here myself without looking into it further.

Are we sure that

```sql
UPDATE {h-schema}utxo_transaction SET status = :newStatus, updated = :updatedAt
            WHERE id = :transactionId 
            AND (status = :newStatus OR status = '$UNVERIFIED')"""
```

Does exactly what the original code did?

```sql
ON CONFLICT(transaction_id) DO
                UPDATE SET status = EXCLUDED.status, updated = EXCLUDED.updated
                WHERE utxo_transaction_status.status = EXCLUDED.status OR utxo_transaction_status.status = '$UNVERIFIED'"""
```

I think you've done it right, but I've not really looked into the `ON CONFLICT` syntax before so wanted to double check.

lankydan (commented):
A few comments but generally looks good to me.

testing/ledger/ledger-hsqldb/src/integrationTest/kotlin/net/corda/testing/ledger/tests/HsqldbVaultNamedQueryTest.kt
```diff
@@ -286,7 +276,6 @@ class HsqldbVaultNamedQueryTest {
 
         assertAll(
             { assertEquals(txId.toString(), visibleState.transaction.id) },
-            { assertEquals(10, visibleState.groupIndex) },
```
vlajos: What did happen with the `groupIndex` asserts?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
nkovacsx: Well, it's not a persist anymore, it's just doing an update but I can change it back if that feels better.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
nkovacsx: Yeah okay could do that

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
nkovacsx: Good point, I don't even know why I changed this.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
nkovacsx: true

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
nkovacsx: Yeah, I put a lot of thought into this before raising the PR but the problem is we have an API call for the update status. Not sure what would be the best here. We could do an insertion here and do `ON CONFLICT` but that would always just update anyway.

components/ledger/ledger-persistence/src/integrationTest/kotlin/net/corda/ledger/persistence/utxo/tests/UtxoPersistenceServiceImplTest.kt
nkovacsx: I'll stick with `visibleTransactionOutput`

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
nkovacsx: Yeah changed it to `vto`

testing/ledger/ledger-hsqldb/src/integrationTest/kotlin/net/corda/testing/ledger/tests/HsqldbVaultNamedQueryTest.kt
nkovacsx: I accidentally removed that I think. Added it back now.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
relyafi: If it's genuinely an update only operation then fair enough.

components/ledger/ledger-persistence/src/integrationTest/kotlin/net/corda/ledger/persistence/utxo/tests/UtxoPersistenceServiceImplTest.kt
nkovacsx: Renamed things to visibleTransactionOutput throughout the code.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
relyafi: If you've thought through it and are absolutely sure there's no issues (even if there's a race condition with a persist happening in parallel) then I'm fine with that, I just remember thinking I was going to do similar when I was looking to do the multi database support work and talked myself out of it because I didn't think it was safe. But this might be slightly different.

vlajos: If https://github.com/corda/corda-runtime-os/pull/4754/files/2411f7d5d58e1aaa7f9bd6dc51793738fc630f05#diff-8f1ffec85c1aacd29dc7c134fe22640d17b971ac9e64a34f67fd1bbb4ef03493R20 would have the on conflict logic, then https://github.com/corda/corda-runtime-os/pull/4754/files/2411f7d5d58e1aaa7f9bd6dc51793738fc630f05#diff-715e152e621b5544da986ce77325bcb15d74deb70268677552a5e0eda82f7f42R206
could be removed.
Which is one less query, so beneficial.
But otherwise, I think the PR's current behaviour is already correct.
And we should still keep this updateTransactionStatus for the dedicated API call.

testing/ledger/ledger-hsqldb/src/integrationTest/kotlin/net/corda/testing/ledger/tests/HsqldbVaultNamedQueryTest.kt
```diff
@@ -262,7 +254,6 @@ class HsqldbVaultNamedQueryTest {
         val visibleState = executeQuery(sqlText, txId) {}.single()
         assertAll(
             { assertEquals(txId.toString(), visibleState.transaction.id) },
-            { assertEquals(30, visibleState.groupIndex) },
```
vlajos: These`groupIndex` asserts look like removed from a few more places.

lankydan (dismissed):

testing/ledger/ledger-hsqldb/src/integrationTest/kotlin/net/corda/testing/ledger/tests/HsqldbVaultNamedQueryTest.kt
```diff
@@ -38,6 +38,7 @@ class HsqldbVaultNamedQueryTest {
         private const val UTXO_VISIBLE_TX_STATE = "utxo_visible_transaction_output"
         private const val BASE_QUERY = "SELECT * FROM $UTXO_VISIBLE_TX_STATE AS visible_state WHERE "
         private const val TIMEOUT_MILLIS = 10000L
+        private const val TRANSACTION_VERIFIED_STRING = "V"
```
relyafi: Can we use the existing `TransactionStatus.VERIFIED` enum value?

nkovacsx: We'd need to add a test dependency to the `ledger-common-data` module. Not sure if it's worth it, if you feel like I'm happy to add the dependency.

relyafi: Personally I think it's ok to add a dependency from tests to the common data module - it would be the other way round that would be problematic.

In practice it doesn't matter too much, but having a single constant used everywhere reduces the chance of breakages if we choose to change it in future for whatever reason.

relyafi (dismissed):

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -83,13 +83,13 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
     override val findTransactionStatus: String
```
vlajos, line 83: This is called a few times before `findTransaction`, so maybe those blocks could be optimized.
But I think that could go into a separate change if it is important enough at all.

vlajos (dismissed):

lankydan (dismissed):

vlajos (dismissed):

filipesoliveira (dismissed):

blsemo (dismissed):
LGTM

filipesoliveira (dismissed):

blsemo (dismissed):

owenstanford (approved):

lankydan (approved):

blsemo (approved):

PR #4757, opened 2023-10-02
CORE-17056 Add network plugin smoke tests

YashNabar:
Adds Combined Worker smoke tests for the Network plugin for the following commands: `onboard-mgm`, `onboard-member`, `lookup group-parameters`, `get-registrations`, `export-group-policy`.

Also includes minor command improvements.

Test pipeline: https://ci02.dev.r3.com/job/Corda5/job/corda-runtime-os-combined-worker-plugins-smoke-tests/job/PR-4757/


tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
```diff
@@ -112,19 +112,13 @@ abstract class BaseOnboard : Runnable, RestCommand() {
     lateinit var name: String
 
     @Option(
-        names = ["--ca"],
-        description = ["The CA location (default to ~/.corda/ca)"]
-    )
-    var caHome: File = File(File(File(System.getProperty("user.home")), ".corda"), "ca")
```
dimosr, line 118: What's the reasoning for removing this? Don't we want to allow the user to override it anymore?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/ExportGroupPolicy.kt
```diff
@@ -24,7 +24,7 @@ class ExportGroupPolicy : Runnable, RestCommand() {
     @Option(
         names = ["-h", "--holding-identity-short-hash"],
         arity = "1",
-        description = ["Short hash of the holding identity to be checked."]
+        description = ["The holding identity ID of the MGM."]
```
dimosr:
```suggestion
        description = ["The holding identity short hash of the MGM."]
```

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMember.kt
```diff
@@ -31,19 +29,23 @@ import net.corda.membership.lib.MemberInfoExtension.Companion.LEDGER_KEY_SIGNATU
...
+    }
+
     @Option(
-        names = ["--cpb-file"],
+        names = ["--cpb-file", "-b"],
         description = [
-            "Location of a CPB file (Use either cpb-file or cpi-hash).",
-            "The CPI will be signed with default options."
+            "Location of a CPB file (Use either --cpb-file or --cpi-hash).",
+            "The plugin will generate a CPI signed with default options."
```
dimosr: The generation of the CPI - does it happen only when the user specifies `--cpb-file`?

If that's the case, it would be good to clarify this because it follows a sentence that talks about both options so can be a bit ambiguous.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMember.kt
```diff
@@ -65,19 +67,19 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
-        names = ["--cpi-hash"],
-        description = ["The CPI hash (Use either cpb-file or cpi-hash)."]
+        names = ["--cpi-hash", "-h"],
+        description = ["The CPI hash of a previously uploaded CPI (use either --cpb-file or --cpi-hash)."]
     )
     var cpiHash: String? = null
 
     @Option(
-        names = ["--pre-auth-token"],
+        names = ["--pre-auth-token", "-a"],
```
dimosr, line 77:
```suggestion
        names = ["--pre-auth-token", "-t"],
```
or would this clash with anything else?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMember.kt
```diff
@@ -107,75 +109,27 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
-            }
-            if (currentCpis.any { it.cpiFileChecksum == cpiFileChecksum }) {
-                println("CPI was already uploaded in $cpiFile. CPI hash checksum is $cpiFileChecksum")
-                return cpiFileChecksum
-            } else {
-                println("CPI $cpiFileChecksum no longer exists. Will create the CPI file again")
-                cpiHashesFile.delete()
-            }
+        val groupId = json.readTree(groupPolicyFile.inputStream()).get("groupId").asText()
+        val cpiName = "${cpbFile.name}-$groupId"
```
dimosr: Do we need to calculate the hash of the CPB's/CPI's content and include that in the name? Thinking of scenarios where a user might have changed the contents of the file (e.g. made some CorDapp changes) but retained file name the same. Wouldn't we want to re-upload it in that case, instead of assuming it's the same?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMember.kt
```diff
@@ -107,75 +109,27 @@ class OnBoardMember : Runnable, BaseOnboard() {
...
+        val cpiFile = File(cpisRoot, "$cpiName.cpi")
+        println("Creating and uploading CPI using CPB '${cpbFile.name}'")
+        val cpisFromCluster = createRestClient(CpiUploadRestResource::class).use { client ->
+            client.start().proxy.getAllCpis().cpis
+        }
+        cpisFromCluster.firstOrNull { it.id.cpiName == cpiName && it.id.cpiVersion == CPI_VERSION }?.let {
+            println("CPI already exists, using CPI ${it.id}")
+            return it.cpiFileChecksum
         }
         if (!cpiFile.canRead()) {
```
dimosr: already existing code, but is there a reason we do `canRead()` instead of `exists()` here?

tools/plugins/network/src/pluginSmokeTest/kotlin/net/corda/cli/plugins/network/OnboardMemberTest.kt
```diff
@@ -0,0 +1,237 @@
...
+            assertEquals(mgmName, outputStub.getFirstPartyName())
+        }
+    }
+
+    @BeforeEach
+    fun beforeEach() {
+        outputStub = OutputStub()
+    }
+
+    @Order(0)
```
dimosr: Is this ordering this before all the other tests? What's the reason it's needed?

dimosr: Reading below, I suspect it's to ensure this CPB hasn't been uploaded yet. If so, are there any ways to do that while ensuring there's no ordering needed? 

For instance, what if we have an extra CPB with a different name just for this test?

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/BaseOnboard.kt
YashNabar: This was removed from the design based on this [thread](https://github.com/corda/platform-eng-design/pull/552#discussion_r1253161337)

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMember.kt
YashNabar: That would clash with `RestCommand`'s target option

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMember.kt
YashNabar: Can't see a reason, will switch to `exists()`

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMember.kt
YashNabar: That was an oversight, thanks. Fixed in [dc28d64](https://github.com/corda/corda-runtime-os/pull/4757/commits/dc28d64deb7c6194f44f73271855a82ad61c1109)

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/utils/FileUtils.kt
```diff
@@ -0,0 +1,13 @@
...
+import net.corda.v5.base.util.EncodingUtils.toBase64
+import java.io.File
+import java.security.MessageDigest
+
+fun Collection<File>.hash(): String {
+    val digest = MessageDigest.getInstance("SHA-256")
+    this.forEach { file ->
+        digest.update(file.readBytes())
+    }
+    return digest.digest().let(::toBase64).replace('/', '.').replace('+', '-').replace('=', '_')
```
dimosr, line 13: What is the reason behind the replacements? Is it to avoid causing issues when the hash is used as a filename in a filesystem?

Worth a comment as it's not too clear by just looking at the code.

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/OnboardMember.kt
```diff
@@ -184,8 +139,8 @@ class OnBoardMember : Runnable, BaseOnboard() {
         val creator = CreateCpiV2()
         creator.cpbFileName = cpbFile.absolutePath
         creator.groupPolicyFileName = groupPolicyFile.absolutePath
-        creator.cpiName = cpbFile.name
-        creator.cpiVersion = "1.0"
+        creator.cpiName = "${cpbFile.name}-$combinedHash"
```
dimosr: worth extracting this into a common function now, as it's generated in a few places in the same way?

dimosr (commented):
LGTM - just a couple of minor questions

tools/plugins/network/src/main/kotlin/net/corda/cli/plugins/network/utils/FileUtils.kt
YashNabar: Yes, clarified in [f329f89](https://github.com/corda/corda-runtime-os/pull/4757/commits/f329f898463c927da0936fdf9046cb72e3125673)

dimosr (dismissed):

dimosr (approved):

PR #4758, opened 2023-10-02
CORE-17352 - Update docs to remove ledger keys from notary registration

dimosr:


yift-r3 (commented):
We should add a comment in the `Assign HSM and generate key pairs` section about the ledger key needed only when the member is not a notary.

dimosr:
Good catch! I've added that now.

yift-r3 (approved):

dimosr:
Merged into: https://github.com/corda/corda-runtime-os/wiki/Member-Onboarding-(Dynamic-Networks)

PR #4752, opened 2023-10-02
ES-1414 Update PreviewTest.kt to use toURI() call

ronanbrowne:
Calling `.file` fails in a Windows CI environment, where as `toURI()` seems to be more forgiving regardless of OS / environment 

Passing windows CI test:  https://ci02.dev.r3.com/job/Corda5/view/Full%20Health%205.1/job/Nightlys/job/Corda-Runtime-OS/job/Windows/view/change-requests/job/PR-4752/

PR #4759, opened 2023-10-03
ES-1414 - use Path instead of File to read from file

bpaunescu:
Addition to https://github.com/corda/corda-runtime-os/pull/4752 . We missed a place where failure would occur on Windows Server environments due to usage of `File(string)`.

Tests on Windows now pass https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2FNightlys%2FCorda-Runtime-OS%2FWindows/detail/bogdan%2Fnotick-fix-win-build/8/tests


ronanbrowne (approved):
LGTM

PR #4772, opened 2023-10-04
CORE-17478 - remove helm chart values for kafka topic cleanup

bpaunescu:


davidcurrie:
I suspect there are also some usages of the option that need removing e.g. in `values.yaml` in the root of the repo.

davidcurrie (approved):

ronanbrowne (approved):
LGTM

PR #4774, opened 2023-10-04
CORE-16234: For persistence api add endpoint and change naming of existing endpoint

emilybowe:
For the RPC Epic the __Persistence_ api_ will require two endpoints. One endpoint is already merged which relates to ledger-related persistence. This PR will be making a suggested change to the naming of that endpoint for clarity, and will also add the second endpoint required. 

When this work is merged we will have the endpoints:
/**ledger** - for ledger persistence
/**persistence** - for entity persistence


components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/EntityRequestSubscriptionFactoryImpl.kt
```diff
@@ -26,28 +31,42 @@ class EntityProcessorFactoryImpl @Activate constructor(
...
         )
+    }
 
-        return EntityProcessorImpl(subscription)
+    override fun createRpcSubscription(): RPCSubscription<EntityRequest, FlowEvent> {
+        val processor = EntityRpcRequestProcessor(
+            currentSandboxGroupContext,
+            entitySandboxService,
+            responseFactory,
+            //TODO - what to do about config payload below?
```
mbrkic-r3: It's not needed (this was checking size of Kafka messages)

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityRpcRequestProcessor.kt
```diff
@@ -0,0 +1,229 @@
...
+    override val responseClass: Class<FlowEvent>,
+    private val requestsIdsRepository: RequestsIdsRepository = RequestsIdsRepositoryImpl()
+) : SyncRPCProcessor<EntityRequest, FlowEvent> {
+
+    private companion object {
+        val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+    }
+
+    override fun process(request: EntityRequest): FlowEvent {
+        //TODO - include null filter? See EntityRequestProcessor
```
mbrkic-r3: Is it possible that null response is returned?

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityRpcRequestProcessor.kt
```diff
@@ -0,0 +1,229 @@
...
+        //TODO fix timestamp here
+/*        CordaMetrics.Metric.Db.EntityPersistenceRequestLag.builder()
```
mbrkic-r3: I think we don't need this, not sure why we have it in `EntityRequestProcessor.onNext` (with wrong operation "Crypto Event"), since it's also in `processEvent`

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityRpcRequestProcessor.kt
emilybowe: @ben-millar was just saying about this that he thinks we don't need that since this will be handling one request at a time (until batch support ticket)

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/EntityRequestSubscriptionFactoryImpl.kt
emilybowe: ok, I'll need to make some changes to `PersistenceServiceInternal` in that case

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityRpcRequestProcessor.kt
emilybowe: great, removed 👍

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/EntityRequestSubscriptionFactoryImpl.kt
emilybowe: ok I've pushed something for that, let's see if that will work

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionRpcTests.kt
```diff
@@ -0,0 +1,383 @@
...
+import java.util.UUID
+
+
+/**
+ * Test persistence exceptions.  Most of these tests don't require the database
+ * set up parts, because we're mocking.
+ */
+@ExtendWith(ServiceExtension::class, BundleContextExtension::class, DBSetup::class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class PersistenceExceptionRpcTests {
```
driessamyn: Isn't this pretty much a carbon copy of `PersistenceExceptionRpcTests`?
This is a bit of a smell. Either we've duplicated code which we shouldn't, or we are testing things twice.

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionTests.kt
```diff
@@ -294,7 +296,7 @@ class PersistenceExceptionTests {
 
         val dogDbCount = getDogDbCount(virtualNodeInfo.vaultDmlConnectionId)
         // There shouldn't be a dog duplicate entry in the DB, i.e. dogs count in the DB should still be 1
-        assertEquals(1, dogDbCount)
+        assertEquals(initialDogDbCount + 1, dogDbCount)
```
driessamyn, line 299: Is this because we're running the 2 test against the same DB?

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceServiceRpcInternalTests.kt
```diff
@@ -0,0 +1,887 @@
...
+import javax.persistence.EntityManagerFactory
+
+sealed class RpcQuerySetup {
+    data class RpcNamedQuery(val params: Map<String, String>, val query: String = "Dog.summon") : RpcQuerySetup()
+    data class RpcAll(val className: String) : RpcQuerySetup()
+}
+
+@ExtendWith(ServiceExtension::class, BundleContextExtension::class, DBSetup::class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class PersistenceServiceRpcInternalTests {
```
driessamyn: As above, is this duplicating an existing test?

vlajos (dismissed):
Ledger related changes look good to me.

driessamyn (dismissed):
I don't understand the amount of duplication that has been introduced.

I also think these changes need Smoketests of the HTTP service so that we can validate they actually work, but I understand it has been agreed with @thiagoviana and @mbrkic-r3 that they are not needed.

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionRpcTests.kt
thiagoviana: > I don't understand the amount of duplication that has been introduced.

We shouldn't have duplicated tests.
My understanding is that the older one will be edited/deleted when the integration work is done.


> I also think these changes need Smoketests of the HTTP service so that we can validate they actually work, but I understand it has been agreed with @thiagoviana and @mbrkic-r3 that they are not needed.

@driessamyn we agreed that they must and will be written. I also asked your help with that matter as it's clear we need it.

relyafi (dismissed):
Uniqueness changes look ok

components/persistence/entity-processor-service-impl/src/integrationTest/kotlin/net/corda/entityprocessor/impl/tests/PersistenceExceptionRpcTests.kt
driessamyn: I'm still not sure why we need this test. I.e. which code paths does this test exercise that the previous test doesn't?

driessamyn (approved):
LGTM

relyafi (approved):
No changes to uniqueness since last review

vlajos (approved):
ledger OK

PR #4775, opened 2023-10-04
CORE-17322 Specify certificate type when verification fails

YashNabar:
If TLS or Session certificate verification fails while setting up a locally-hosted identity, the error message now specifies the certificate type which failed verification.

Before:
```
{
    "title": "Could not import certificate: The CertificateType was not signed by the correct certificate authority",
    "status": 500,
    "details": {}
}
```

After this change:

Screenshot 2023-10-05 at 12 03 58

Screenshot 2023-10-05 at 12 08 47

### DIFF
--- a/components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/HostedIdentityEntryFactory.kt
+++ b/components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/HostedIdentityEntryFactory.kt
@@ -259,7 +259,7 @@ internal class HostedIdentityEntryFactory(
         if (certificateType.trustRoots.isEmpty()) {
             throw CordaRuntimeException("The group ${holdingIdentity.groupId} P2P parameters ${certificateType.parameterName} is empty")
         }
-        certificateType.trustRoots
+        val trustRoot = certificateType.trustRoots
             .asSequence()
             .map { rootCertificateStr ->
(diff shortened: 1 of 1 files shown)


components/membership/certificates-client-impl/src/main/kotlin/net/corda/membership/certificate/client/impl/HostedIdentityEntryFactory.kt
```diff
@@ -274,8 +274,11 @@ internal class HostedIdentityEntryFactory(
                     false
                 }
             }.firstOrNull()
-            ?: throw CordaRuntimeException(
-                "The ${CertificateType::class.java.simpleName} was not signed by the correct certificate authority"
-            )
+            if (trustRoot == null) {
+                val type = certificateType as? CertificateType.TlsCertificate ?: certificateType as CertificateType.SessionCertificate
```
williamvigorr3: It might be better to add a field to `CertificateType` instead (with the name)? WDYT? Then we could not use CamelCase, which might be confusing to a user.

YashNabar: Addressed in [3b7e63f](https://github.com/corda/corda-runtime-os/pull/4775/commits/3b7e63f39e3700fbad4bf0c4f088fa1e2b09d604)

williamvigorr3 (approved):
LGTM

PR #4777, opened 2023-10-04
CORE-17476 Modifying helm charts to create ClusterIP services for workers, passing to FlowWorker

ben-millar:
As part of the Corda topology redesign, communication between the flow engine and Corda workers is moving from a Kafka message bus to synchronous RPC calls. In order to support this, we're adding a load balancing layer (Here, a `ClusterIP` service per worker) which will expose a single endpoint which distributes calls across all worker instances. 

The endpoint for this ClusterIP service is now being passed into the additional startup args of the `FlowWorker`; a future PR will use these to define a routing component for external messages.

#### Related PRs:
- https://github.com/corda/corda-api/pull/1276 (Adding endpoint IDs to API)

### DIFF
--- a/applications/workers/release/flow-worker/src/main/kotlin/net/corda/applications/workers/flow/FlowWorker.kt
+++ b/applications/workers/release/flow-worker/src/main/kotlin/net/corda/applications/workers/flow/FlowWorker.kt
@@ -94,6 +94,6 @@ private class FlowWorkerParams {
     @Mixin
     var defaultParams = DefaultWorkerParams()
 
-    @Option(names = ["--endpoint"], description = ["Internal RPC endpoints for Corda workers"], required = true)
+    @Option(names = ["--endpoint"], description = ["Internal REST endpoints for Corda workers"], required = true)
     val workerEndpoints: Map<String, String> = emptyMap()
 }
--- a/applications/workers/release/flow-worker/src/main/kotlin/net/corda/applications/workers/flow/FlowWorker.kt
+++ b/applications/workers/release/flow-worker/src/main/kotlin/net/corda/applications/workers/flow/FlowWorker.kt
@@ -24,6 +24,7 @@ import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
 import org.slf4j.LoggerFactory
 import picocli.CommandLine.Mixin
+import picocli.CommandLine.Option
 
 /** The worker for handling flows. */
 @Suppress("Unused", "LongParameterList")
@@ -92,4 +93,7 @@ class FlowWorker @Activate constructor(
 private class FlowWorkerParams {
     @Mixin
     var defaultParams = DefaultWorkerParams()
+
+    @Option(names = ["--endpoint"], description = ["Internal RPC endpoints for Corda workers"], required = true)
+    val workerEndpoints: Map<String, String> = emptyMap()
 }
--- a/applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/Constants.kt
+++ b/applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/Constants.kt
@@ -5,5 +5,5 @@ internal const val HTTP_SERVICE_UNAVAILABLE_CODE = 503
 internal const val HTTP_HEALTH_ROUTE = "/isHealthy"
 internal const val HTTP_METRICS_ROUTE = "/metrics"
 internal const val HTTP_STATUS_ROUTE = "/status"
-internal const val WORKER_SERVER_PORT = 7000
+internal const val WORKER_SERVER_PORT = 7000 // Note: This variable is defined in charts/corda-lib/templates/_helpers.kt:622
 internal const val NO_CACHE = "no-cache"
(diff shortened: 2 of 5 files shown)


applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/Constants.kt
```diff
@@ -5,5 +5,5 @@ internal const val HTTP_SERVICE_UNAVAILABLE_CODE = 503
 internal const val HTTP_HEALTH_ROUTE = "/isHealthy"
 internal const val HTTP_METRICS_ROUTE = "/metrics"
 internal const val HTTP_STATUS_ROUTE = "/status"
-internal const val WORKER_SERVER_PORT = 7000
+internal const val WORKER_SERVER_PORT = 7000 // Note: This variable is defined in charts/corda-lib/templates/_helpers.kt:622
```
davidcurrie: This is really the definition of the port. The Helm chart variable tells the rest of the world what this value is. I don't know whether this port is configurable on a worker but, if it is, the Helm chart could pass the value in and then be sure that everything ties up nicely (even if the value we pass in is still 7000!).

charts/corda/templates/workers.yaml
```diff
@@ -4,8 +4,16 @@
...
+{{- $args := list }}
+{{- $servicesAccessed := list "crypto" "verification" "uniqueness" "persistence" }}
+{{- range $worker := $servicesAccessed }}
+{{- $endpoint := include "corda.getWorkerEndpoint" (dict "context" $ "worker" $worker) }}
+{{- $args = append $args (printf "--endpoint=%s" $endpoint) }}
+{{- end }}
+
 {{- include "corda.worker" ( list $ .Values.workers.flow "flow"
-  ( dict "stateManagerDbAccess" true )
+  ( dict "stateManagerDbAccess" true "additionalWorkerArgs" $args )
```
davidcurrie: It probably doesn't make much difference while the flow worker is the only worker doing this but I'd still be tempted to pass `servicesAccessed` in the dictionary and then do the processing of that within `corda.worker` rather than here.

charts/corda/templates/workers.yaml
```diff
@@ -4,8 +4,16 @@
 {{- include "corda.worker" ( list $ .Values.workers.db "db"
   ( dict "clusterDbAccess" true )
 ) }}
+
+{{- $args := list }}
+{{- $servicesAccessed := list "crypto" "verification" "uniqueness" "persistence" }}
+{{- range $worker := $servicesAccessed }}
+{{- $endpoint := include "corda.getWorkerEndpoint" (dict "context" $ "worker" $worker) }}
+{{- $args = append $args (printf "--endpoint=%s" $endpoint) }}
```
davidcurrie: nit: just for readability, I'd be tempted to indent these lines to indicate that they're nested within the `range` (the white space being eaten by `{{-` anyway).

davidcurrie (dismissed):
I suspect the port isn't configurable, and if not, making it so would not be part of this PR. The other suggestions are just stylistic.

applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/Constants.kt
driessamyn: It defaults to 7000, but you can override it using the `-p` param: https://github.com/corda/corda-runtime-os/blob/f45367e5763814cae7977816f8e46d9b5700f0f7/applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/DefaultWorkerParams.kt#L37

(just FYI)

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -614,3 +614,28 @@ data:
   {{ $keySecretKey }}: {{ $keySecretValue }}
   {{ $caSecretKey }}: {{ $caSecretValue }}
 {{- end }}
+
+{{/*
+The port which should be used to connect to Corda worker instances
+*/}}
+{{- define "worker.port" -}}
```
driessamyn: Bit nitpicky (and don't propose it needs to be changed), but I guess technically this is the LB port, not the worker port?

applications/workers/release/flow-worker/src/main/kotlin/net/corda/applications/workers/flow/FlowWorker.kt
```diff
@@ -92,4 +93,7 @@ class FlowWorker @Activate constructor(
 private class FlowWorkerParams {
     @Mixin
     var defaultParams = DefaultWorkerParams()
-}
\ No newline at end of file
+
+    @Option(names = ["--endpoint"], description = ["Internal RPC endpoints for Corda workers"], required = true)
```
jujoramos: Nitpick: can we update the description to HTTP endpoints instead, as that what they are?. Just trying to avoid weeks of works in the future trying to refactor from RPC to HTTP again (it already happened for the old `rpc-worker`)?.

applications/workers/worker-common/src/main/kotlin/net/corda/applications/workers/workercommon/Constants.kt
ben-millar: Good to know! 🤔 I'll remove the comment then, as it's not adding any clarity to the code 😅

charts/corda-lib/templates/_helpers.tpl
ben-millar: Technically its both the port for the worker and for the load balancer; the LB forwards port :7000 to port :7000 on the respective worker. I'm on the fence about it, but I figure `worker.port` makes the overall intent more clear.

davidcurrie: oh - but now you've highlighted it again, I realise this breaks the naming convention for functions. They are all `corda.` something. This is because the functions are in a global namespace when charts are combined together.

davidcurrie (dismissed):
👍

driessamyn (dismissed):
LGTM

jujoramos (dismissed):
LGTM

driessamyn (approved):
LGTM

conal-smith (approved):

davidcurrie (approved):

PR #4779, opened 2023-10-05
CORE-12786: Add an Evm Worker

fowlerrr:
This is the shell of the worker.  The processor wil be added after


fowlerrr:
We won't need a worker any time soon and things will probably change so we'll revisit when necessary

PR #4780, opened 2023-10-05
CORE-17437: State Manager Configuration Section

jujoramos:
Promote State Manager configuration to its own section and remove it
from under the Messaging configuration.


jujoramos:
Related api change: https://github.com/corda/corda-api/pull/1277

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/PostgresDbSetup.kt
```diff
@@ -37,7 +37,7 @@ class PostgresDbSetup(
             "net/corda/db/schema/messagebus/db.changelog-master.xml" to "MESSAGEBUS",
             "net/corda/db/schema/rbac/db.changelog-master.xml" to "RBAC",
             "net/corda/db/schema/crypto/db.changelog-master.xml" to "CRYPTO",
-            "net/corda/db/schema/statemanager/db.changelog-master.xml" to "STATE_MANAGER",
+            "net/corda/db/schema/state-manager/db.changelog-master.xml" to "STATE_MANAGER",
```
sprintengine: we also need to change this in _bootstrap.tpl. however, I already know this isn't working correctly, and I've made changes to calling database spec in two other branches, so I think leave it for now to reduce my merge conflicts.

libs/configuration/configuration-merger/src/main/kotlin/net/corda/libs/configuration/merger/ConfigMerger.kt
```diff
@@ -3,25 +3,27 @@ package net.corda.libs.configuration.merger
...
+     *
      * @param bootConfig boot config created on startup
-     * @param messagingConfig messaging config taken from the topic
-     * @return Messaging config with boot config values merged into it.
+     * @param configKey path for the boot configuration section to merge
+     * @param existingConfig existing configuration taken from the topic
+     * @return configuration with boot config values under the specified section merged into it
      */
-    fun getMessagingConfig(bootConfig: SmartConfig, messagingConfig: SmartConfig? = null) : SmartConfig
+    fun getConfig(bootConfig: SmartConfig, configKey: String, existingConfig: SmartConfig?): SmartConfig
```
sprintengine, line 19: Yeah this is a better API

libs/configuration/configuration-validation/src/main/kotlin/net/corda/libs/configuration/validation/impl/ConfigurationValidatorImpl.kt
```diff
@@ -117,6 +117,8 @@ internal class ConfigurationValidatorImpl(private val schemaProvider: SchemaProv
 
     private fun getSchema(schemaInput: InputStream, applyDefaults: Boolean): JsonSchema {
         val schemaValidatorsConfig = SchemaValidatorsConfig().apply {
+            // Try to convert string to declared type when validating the schema.
+            isTypeLoose = true
```
sprintengine, line 121: +1 

This validator does appear to be set up to apply defaults when a property is missing, but perhaps not when an object is missing. Wonder what is the correct approach to get defaults applied, maybe our state-manager config schema is missing a default object. For example, the defaults weren't getting applied to the state manager jdbc directory. Something I might try out at a later time when I get some free time.

applications/workers/worker-common/src/test/kotlin/net/corda/applications/workers/workercommon/internal/BootstrapConfigTest.kt
```diff
@@ -184,31 +182,4 @@ class BootstrapConfigTest {
...
-        assertSoftly { softly ->
-            softly.assertThat(config.getInt(MessagingConfig.StateManager.JDBC_POOL_MAX_SIZE)).isEqualTo(111)
-            softly.assertThat(config.getInt(MessagingConfig.StateManager.JDBC_POOL_MIN_SIZE)).isEqualTo(222)
-            softly.assertThat(config.getInt(MessagingConfig.StateManager.JDBC_POOL_IDLE_TIMEOUT_SECONDS)).isEqualTo(333)
-            softly.assertThat(config.getInt(MessagingConfig.StateManager.JDBC_POOL_MAX_LIFETIME_SECONDS)).isEqualTo(444)
-            softly.assertThat(config.getInt(MessagingConfig.StateManager.JDBC_POOL_KEEP_ALIVE_TIME_SECONDS)).isEqualTo(555)
-            softly.assertThat(config.getInt(MessagingConfig.StateManager.JDBC_POOL_VALIDATION_TIMEOUT_SECONDS)).isEqualTo(666)
-        }
-
-    }
```
sprintengine, line 213: Surely if the "isLooseType" setting works, this test should still pass?

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/PostgresDbSetup.kt
sprintengine: If I merge [PR-4791](https://github.com/corda/corda-runtime-os/pull/4791) first, you'll have to update the bootstrap template https://github.com/corda/corda-runtime-os/pull/4791/files#diff-6488cdf0bdcc5c363da8948f42ead979f4346e4fcff509d46f959ab517fcecd6

EDIT - this now merged

jujoramos: 👍

libs/configuration/configuration-validation/src/main/kotlin/net/corda/libs/configuration/validation/impl/ConfigurationValidatorImpl.kt
jujoramos: Agreed in that we might want to test the defaulting part further in the future but, as you correctly stated, that shouldn't be part of this `PR` so I'll resolve this comment 👍

applications/workers/worker-common/src/test/kotlin/net/corda/applications/workers/workercommon/internal/BootstrapConfigTest.kt
jujoramos: The test passes just fine, yeah, but we are not manually casting the properties ourselves anymore so I don't think the test belong here. I'll add another one under `ConfigurationValidatorImplTest` instead.

JamesHR3 (dismissed):

mbrkic-r3 (dismissed):

JamesHR3 (dismissed):

conal-smith (dismissed):

mbrkic-r3 (dismissed):

components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/FlowMaintenanceImpl.kt
```diff
@@ -40,17 +40,14 @@ class FlowMaintenanceImpl @Activate constructor(
     private var stateManager: StateManager? = null
 
     override fun onConfigChange(config: Map<String, SmartConfig>) {
-        // TODO - fix config key (CORE-17437).
-        if(config.containsKey(ConfigKeys.MESSAGING_CONFIG)) {
+        if (config.containsKey(ConfigKeys.STATE_MANAGER_CONFIG)) {
```
driessamyn: I don't think this is correct as this assumes that the messaging config is always already there when the state manager config arrives. 

I think they need to be treated separate.

simon-johnson-r3 (dismissed):

driessamyn (changes requested):

driessamyn (approved):
LGTM

JamesHR3 (approved):

PR #4786, opened 2023-10-05
CORE-16966: Proof of Action Lock State (Part 2)

julia-filipczak:
Adding the reclaim functionality to the atomic swap cordapp.

### DIFF
--- a/testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/AssetContract.kt
+++ b/testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/AssetContract.kt
@@ -32,10 +32,7 @@ class AssetContract : Contract {
                 "The transaction requires one Asset output." using (transaction.getOutputStates(Asset::class.java).size == 1)
                 "The transaction requires one Asset input." using (transaction.getInputStates(Asset::class.java).size == 1)
 
-                val input = transaction.getInputStates(Asset::class.java).first()
-
                 val output = transaction.getOutputStates(Asset::class.java).first()
-                "Owner must change in this transaction." using (input.owner != output.owner)
 
                 "Both old and new owner must sign the transaction." using (transaction.signatories.containsAll(output.participants))
             }
--- a/testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
+++ b/testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
@@ -56,6 +56,12 @@ class LockContract : Contract {
                         (transaction.getInputStates(LockState::class.java).size == 1)
                 "There should be one output state as an unlocked asset state needs to be created" using
                         (transaction.getOutputStates(OwnableState::class.java).size == 1)
+                val inputLockState = transaction.getInputStates(LockState::class.java).first()
+                "Only lock state creator can do a reclaim" using (transaction.signatories.contains(inputLockState.creator))
+                val timeWindowUntil = transaction.timeWindow.from
+                "The time window must have expired. Reversal of Encumbered assets requires a time window that starts " +
+                        "only after the time-window to claim the encumbered assets" using (timeWindowUntil != null &&
+                        timeWindowUntil.isAfter(inputLockState.timeWindow))
             }
 
             else -> {
--- a/testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
+++ b/testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
@@ -64,8 +64,6 @@ class LockContract : Contract {
                 "The lock state time window must have expired. Reclaim of encumbered assets requires a time window that starts " +
                         "only after the lock state time window" using (reclaimTransactionStartTime != null &&
                         reclaimTransactionStartTime.isAfter(inputLockState.timeWindow))
-                "Time window should not be longer than a day." using (reclaimTransactionStartTime != null &&
-                        transaction.timeWindow.until < reclaimTransactionStartTime.plusSeconds(86400))
             }
 
             else -> {
--- a/testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
+++ b/testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
@@ -52,9 +52,9 @@ class LockContract : Contract {
             }
 
             is LockCommands.Reclaim -> {
-                val inputLockState = transaction.getInputStates(LockState::class.java).first()
                 "There should be one lock input state as the lock state needs to be consumed" using
                         (transaction.getInputStates(LockState::class.java).size == 1)
+                val inputLockState = transaction.getInputStates(LockState::class.java).first()
                 "There should be one output state as an unlocked asset state needs to be created" using
                         (transaction.getOutputStates(OwnableState::class.java).size == 1)
                 "Original owner gets the asset" using
--- a/testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
+++ b/testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
@@ -60,10 +60,12 @@ class LockContract : Contract {
                 "Original owner gets the asset" using
                         (transaction.getOutputStates(OwnableState::class.java).first().owner == inputLockState.creator)
                 "Only lock state creator can do a reclaim" using (transaction.signatories.contains(inputLockState.creator))
-                val timeWindowUntil = transaction.timeWindow.from
-                "The time window must have expired. Reversal of Encumbered assets requires a time window that starts " +
-                        "only after the time-window to claim the encumbered assets" using (timeWindowUntil != null &&
-                        timeWindowUntil.isAfter(inputLockState.timeWindow))
+                val reclaimTransactionStartTime = transaction.timeWindow.from
+                "The lock state time window must have expired. Reclaim of encumbered assets requires a time window that starts " +
+                        "only after the lock state time window" using (reclaimTransactionStartTime != null &&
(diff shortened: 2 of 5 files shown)


testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
```diff
@@ -50,6 +51,19 @@ class LockContract : Contract {
                         (transaction.signatories.containsAll(outputState.participants))
             }
 
+            is LockCommands.Reclaim -> {
+                "There should be one lock input state as the lock state needs to be consumed" using
+                        (transaction.getInputStates(LockState::class.java).size == 1)
+                "There should be one output state as an unlocked asset state needs to be created" using
+                        (transaction.getOutputStates(OwnableState::class.java).size == 1)
+                val inputLockState = transaction.getInputStates(LockState::class.java).first()
+                "Only lock state creator can do a reclaim" using (transaction.signatories.contains(inputLockState.creator))
```
szymonsztuka, line 62: This checks if one of the signatories is the creator, but doesn't check the direction in which the assets was transferred. What if both parties signed a transaction, then this check passes regardless who is receiving the asset (I know that this is highly unlikely scenario).

szymonsztuka (changes requested):

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/AssetContract.kt
```diff
@@ -32,10 +32,7 @@ class AssetContract : Contract {
                 "The transaction requires one Asset output." using (transaction.getOutputStates(Asset::class.java).size == 1)
                 "The transaction requires one Asset input." using (transaction.getInputStates(Asset::class.java).size == 1)
 
-                val input = transaction.getInputStates(Asset::class.java).first()
```
jmacmahonr3, line 35: I'd add a todo or reach out to the ledger team about composite keys being considered the same owner. This is fine for now but it's important to keep in mind

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
```diff
@@ -52,11 +52,13 @@ class LockContract : Contract {
             }
 
             is LockCommands.Reclaim -> {
+                val inputLockState = transaction.getInputStates(LockState::class.java).first()
```
szymonsztuka: Move this line after the line with the check `transaction.getInputStates(LockState::class.java).size == 1`, once we know we have a lock state then we can safely access it via `.first()` without worry of throwing NullPointerException.

szymonsztuka (changes requested):

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
```diff
@@ -50,6 +51,21 @@ class LockContract : Contract {
...
+            is LockCommands.Reclaim -> {
+                "There should be one lock input state as the lock state needs to be consumed" using
+                        (transaction.getInputStates(LockState::class.java).size == 1)
+                val inputLockState = transaction.getInputStates(LockState::class.java).first()
+                "There should be one output state as an unlocked asset state needs to be created" using
+                        (transaction.getOutputStates(OwnableState::class.java).size == 1)
+                "Original owner gets the asset" using
+                        (transaction.getOutputStates(OwnableState::class.java).first().owner == inputLockState.creator)
+                "Only lock state creator can do a reclaim" using (transaction.signatories.contains(inputLockState.creator))
+                val timeWindowUntil = transaction.timeWindow.from
```
szymonsztuka: Maybe this variable should be called `timeWindowFrom` or maybe better `recailTransactionStartTime` or something like that?

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
```diff
@@ -50,6 +51,21 @@ class LockContract : Contract {
...
+                "The time window must have expired. Reversal of Encumbered assets requires a time window that starts " +
+                        "only after the time-window to claim the encumbered assets" using (timeWindowUntil != null &&
```
szymonsztuka: I'm finding this error message a bit confusing : it roughly has two messages:  "_a time window hasn't expired_" , "_a time window needs to start after some other time window_" , + inconsistent wording "time widow" vs "time-window".

testing/cpbs/atomic-swap-workflows/src/main/kotlin/com/r3/corda/atomic/swap/workflows/ReclaimFlow.kt
```diff
@@ -0,0 +1,95 @@
...
+            val myInfo = memberLookup.myInfo()
+            val lockState = encumberedTx.getOutputStateAndRefs(LockState::class.java).singleOrNull()
+                ?: throw IllegalStateException("Transaction with id: ${flowArgs.signedTransactionId} has no lock state")
+            val assetState = encumberedTx.getOutputStateAndRefs(Asset::class.java).singleOrNull()
+                ?: throw IllegalStateException("Transaction with id: ${flowArgs.signedTransactionId} has no asset state")
+            val newAssetState = assetState.state.contractState.withNewOwner(
+                myInfo.ledgerKeys.first(),
+                listOf(myInfo.ledgerKeys.first())
+            )
+            val timeWindowForReclaim = lockState.state.contractState.timeWindow.plusSeconds(31)
```
szymonsztuka: Why 31?

testing/cpbs/atomic-swap-workflows/src/main/kotlin/com/r3/corda/atomic/swap/workflows/ReclaimFlow.kt
```diff
@@ -0,0 +1,95 @@
...
+
+            val txBuilder = ledgerService.createTransactionBuilder()
+
```
szymonsztuka: remove the extra line

testing/cpbs/atomic-swap-workflows/src/main/kotlin/com/r3/corda/atomic/swap/workflows/ReclaimFlow.kt
```diff
@@ -0,0 +1,95 @@
...
+                .setNotary(encumberedTx.notaryName)
+                .setTimeWindowBetween(timeWindowForReclaim, Instant.now())
```
szymonsztuka: I'm quite surprised how this can work, in normal system there will be always i minimal delay between a transaction creation and sending it to notary,  so notary will never have chance to sign this transaction if the time window expiration in set to `Instant.now()`

szymonsztuka: Maybe this transaction is not notarized by a notary, or the notary has disabled time window checking?

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
```diff
@@ -60,10 +60,12 @@ class LockContract : Contract {
...
-                val timeWindowUntil = transaction.timeWindow.from
-                "The time window must have expired. Reversal of Encumbered assets requires a time window that starts " +
-                        "only after the time-window to claim the encumbered assets" using (timeWindowUntil != null &&
-                        timeWindowUntil.isAfter(inputLockState.timeWindow))
+                val reclaimTransactionStartTime = transaction.timeWindow.from
+                "The lock state time window must have expired. Reclaim of encumbered assets requires a time window that starts " +
+                        "only after the lock state time window" using (reclaimTransactionStartTime != null &&
+                        reclaimTransactionStartTime.isAfter(inputLockState.timeWindow))
+                "Time window should not be longer than a day." using (reclaimTransactionStartTime != null &&
+                        transaction.timeWindow.until < reclaimTransactionStartTime.plusSeconds(86400))
```
szymonsztuka: Remove this ad-hoc rule, why one day? Anyway this makes reclaim testing impossible.

szymonsztuka (changes requested):

testing/cpbs/atomic-swap-workflows/src/main/kotlin/com/r3/corda/atomic/swap/workflows/ReclaimFlow.kt
julia-filipczak: Just a random number

testing/cpbs/atomic-swap-contracts/src/main/kotlin/com/r3/corda/atomic/swap/contracts/LockContract.kt
julia-filipczak: Setting an upper limit so that the time window doesn't last for ages. Can change it to a different value. Why does it make testing impossible?

szymonsztuka: Ok, right, true it doesn't impact testing. However we should not decided that you can only 24 hours reclaim window opportunity. The asset may stuck forever in limbo. Rather I would expect the new owner has fixed amount of time to unlock. But anyway if we want to model this we probably need to have optional values in the lock state set at the beginning e.g.: I give you X and you have unlimited time to unlock but after Y time I can reclaim, and then I have limited time to reclaim. Or the other option you have limited time to unlock after i have unlimited time to reclaim. But for sure such values should not be hard-coded in the contract code designed to be a general example.

testing/cpbs/atomic-swap-workflows/src/main/kotlin/com/r3/corda/atomic/swap/workflows/ReclaimFlow.kt
szymonsztuka: Make it 30 seconds but also allow to pass the value to the flow, so user can tune it, make this parameter optional and if user doesn't provide a value set to 30 seconds.

szymonsztuka (approved):

szymonsztuka (approved):

PR #4787, opened 2023-10-06
CORE-17411: Fix MGM info not getting updated after upgrading the same vNode multiple times

BenYip123:
- Serial field getting incorrect serial,  to fix this we query for all APPROVED registration requests and sort to get the latest request

E2E test: [CORE-17411: Add test to upgrade the same vNode multiple times with correct serial and version #260](https://github.com/corda/corda-e2e-tests/pull/260)


vkolomeyko (approved):
LGTM

PR #4793, opened 2023-10-06
CORE-17525 - Some test and tooling dependency upgrades.

driessamyn:
Includes a bunch of fixes for a Detekt rule that comes with the newest version of Detekt:
https://detekt.dev/docs/rules/performance/#foreachonrange


lankydan (approved):

simon-johnson-r3 (approved):

vlajos (approved):

yift-r3 (approved):
Most of the forEach on a range are for small predefined numbers in test code, surely the performance implication in those is not that bad.

Sadly, we can't tell detect to ignore those.

driessamyn:
> Most of the forEach on a range are for small predefined numbers in test code, surely the performance implication in those is not that bad.
> 
> Sadly, we can't tell detect to ignore those.

Agreed, it's not so important for those cases, but the rule is in fact a good one (and detekted a couple of genuine cases where it may matter), it doesn't really make the code less readable and fixing it up doesn't really take more time than surpressing it (and gets into a good habit), so I thought it'd be ok to just fix them.

vkolomeyko (approved):
LGTM

PR #4796, opened 2023-10-06
CORE-17516 Series of backchain code fixes

lankydan:
Fix several issues in the backchain code:

- Missing `@Suspendable` annotations in `TransactionBackchainReceiverFlowV1`.
- Don't store `SmartConfig` as a variable in `TransactionBackchainReceiverFlowV1` to prevent a Kryo serialization error.
- Remove comma from SQL statement that was causing an error.
- Correct `String` to `TransactionStatus` conversion.
- Correctly match types between flow and persistence code.

### DIFF
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/request/handlers/UtxoFindTransactionIdsAndStatusesRequestHandler.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/request/handlers/UtxoFindTransactionIdsAndStatusesRequestHandler.kt
@@ -25,7 +25,7 @@ class UtxoFindTransactionIdsAndStatusesRequestHandler(
             externalEventResponseFactory.success(
                 externalEventContext,
                 EntityResponse(
-                    existingTransactions.map { ByteBuffer.wrap(serializationService.serialize(it).bytes) },
+                    listOf(ByteBuffer.wrap(serializationService.serialize(existingTransactions).bytes)),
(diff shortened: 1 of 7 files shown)


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/request/handlers/UtxoFindTransactionIdsAndStatusesRequestHandler.kt
```diff
@@ -25,7 +25,7 @@ class UtxoFindTransactionIdsAndStatusesRequestHandler(
             externalEventResponseFactory.success(
                 externalEventContext,
                 EntityResponse(
-                    existingTransactions.map { ByteBuffer.wrap(serializationService.serialize(it).bytes) },
+                    listOf(ByteBuffer.wrap(serializationService.serialize(existingTransactions).bytes)),
```
blsemo, line 28: Wouldn't that create a list with a single element that contains all the bytes from the whole `existingTransactions` list?

lankydan: Its a `List` containing a serialized `Map` containing the secure hash to status mappings.

If we wanted to change it, either it would be a list of pairs then we'd recreate the map on the other side, or we use a different data structure entirely.

vlajos (approved):
LGTM

blsemo (approved):

PR #4797, opened 2023-10-06
CORE-17047 Enable duplicate key detection in group policy parsing

YashNabar:
Disallow duplicate key entries in the group policy. Before this change, providing duplicate keys in a static network group policy resulted in the last instance of that key being used (ignoring other instances). 

**Test**:

Upload a CPI with a static network group policy containing the following "groupParameters" block:
```
"groupParameters": {
  "corda.minimum.platform.version": 50000,
  "ext.group.key.1": "value0",
  "ext.group.key.1": "value1"
}
```

Screenshot 2023-10-13 at 11 51 00

Logs:
```
2023-10-06 15:39:56.851 [durable processing thread cpi.chunk.writer-cpi.upload] ERROR net.corda.membership.lib.impl.grouppolicy.GroupPolicyParserImpl {} - GroupPolicy file is incorrectly formatted and parsing failed. Caused by: Duplicate field 'ext.group.key.1'
```


dimosr (commented):
Should we adjust the message of `BadGroupPolicyException` to include the message of the nested exception, similar to what we do when we log, so that the user is informed of the actual reason in the response (e.g. `Duplicate field ...`)? Changes look good otherwise.

dimosr (approved):

PR #4809, opened 2023-10-09
CORE-17542 Remove unused token ledger event messages.

owenstanford:
This change removes unused token selection ledger event messages. These events are redundant now token selection queries the DB directly.


vkolomeyko (approved):
LGTM

lankydan (approved):

PR #4811, opened 2023-10-09
CORE-16236 - Flow Mapper Worker - Remove 'External Event Responses' (unused) Dependency

thiagoviana:
External Events are not being used by Flow Mapper Worker. This PR is to remove it from Gradle.


ronanbrowne (approved):
LGTM

PR #4820, opened 2023-10-10
CORE-16242 - Synchronous RPC Pattern API Implementation for Crypto Worker

thiagoviana:
HTTP RPC server integration into the Crypto processor. Allows Crypto processor to receive HTTP external event requests sent from the flow engine. Client integration will come later.

Obs:

- Full integration with Event Mediator to be implemented;
- New tests will be added/updated then;


components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+
+    private fun getRequestExpireAt(request: FlowOpsRequest): Instant =
+        request.context.requestTimestamp.plusSeconds(getRequestValidityWindowSeconds(request))
+
+    private fun getRequestValidityWindowSeconds(request: FlowOpsRequest): Long {
+        return request.context.other.items.singleOrNull {
+            it.key == CryptoFlowOpsTransformer.REQUEST_TTL_KEY
+        }?.value?.toLong() ?: 300
```
LWogan: should this 300 be a config value?

LWogan: I will create a tech debt ticket. This code is copied from the old processor and has been around a while.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/CryptoRPCSmokeTests.kt
```diff
@@ -0,0 +1,203 @@
...
+        assertEquals(expected.requestingComponent, actual.requestingComponent)
+        assertEquals(expected.requestTimestamp, actual.requestTimestamp)
+        assertThat(actual.responseTimestamp.epochSecond)
+            .isGreaterThanOrEqualTo(expected.requestTimestamp.epochSecond)
+            .isLessThanOrEqualTo(now.epochSecond)
+        assertTrue(
+            actual.other.items.size == expected.other.items.size &&
+                    actual.other.items.containsAll(expected.other.items) &&
+                    expected.other.items.containsAll(actual.other.items)
+        )
```
driessamyn: Very much a nit, but this would be a good candidate for an `assertSoftly`

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
```diff
@@ -0,0 +1,172 @@
...
+            MDC_EXTERNAL_EVENT_ID to request.flowExternalEventContext.requestId
+        ) + translateFlowContextToMDC(request.flowExternalEventContext.contextProperties.toMap())
+
+        val result = withMDC(mdc) {
+            val requestPayload = request.request
+            val startTime = System.nanoTime()
+            logger.info("Handling ${requestPayload::class.java.name} for tenant ${request.context.tenantId}")
+
+            try {
+                if (Instant.now() >= expireAt) {
```
driessamyn: Is TTL still relevant in the RPC pattern?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
```diff
@@ -0,0 +1,172 @@
...
+                    logger.warn(
+                        "Event ${requestPayload::class.java.name} for tenant ${request.context.tenantId} " +
+                                "is no longer valid, expired at $expireAt"
+                    )
+                    externalEventResponseFactory.transientError(
+                        request.flowExternalEventContext,
+                        ExceptionEnvelope("Expired", "Expired at $expireAt")
+                    )
```
driessamyn: How does this work?
Is this sending error messages back over Kafka instead of responding to the request?

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/CryptoRPCSmokeTests.kt
```diff
@@ -0,0 +1,203 @@
...
+            groupId,
+            staticMemberList
+        )
+        val aliceActualHoldingId = getOrCreateVirtualNodeFor(aliceX500, cpiName)
+        assertThat(aliceActualHoldingId).isEqualTo(aliceHoldingId)
+        registerStaticMember(aliceHoldingId)
+    }
+
+    @Test
+    fun `RPC endpoint accepts a request and returns back a response`() {
```
driessamyn, line 118: Is it worth adding non-happy path scenarios?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
```diff
@@ -0,0 +1,172 @@
...
+                } else {
+                    val response = executor.executeWithRetry {
+                        handleRequest(requestPayload, request.context)
+                    }
+
+                    if (Instant.now() >= expireAt) {
+                        logger.warn(
+                            "Event ${requestPayload::class.java.name} for tenant ${request.context.tenantId} " +
+                                    "is no longer valid, expired at $expireAt"
+                        )
```
driessamyn: If TTL is still relevant, presumably we don't need to check twice?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
```diff
@@ -0,0 +1,172 @@
...
+                        externalEventResponseFactory.transientError(
```
driessamyn: As above comment re: responding to Kafka.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
```diff
@@ -0,0 +1,172 @@
...
+                            request.flowExternalEventContext,
+                            ExceptionEnvelope("Expired", "Expired at $expireAt")
+                        )
+                    } else {
+                        externalEventResponseFactory.success(
```
driessamyn: As above comment re: responding to Kafka.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
```diff
@@ -0,0 +1,172 @@
...
+                            FlowOpsResponse(createResponseContext(request), response, null)
+                        )
+                    }
+                }
+            } catch (throwable: Throwable) {
+                logger.error(
+                    "Failed to handle ${requestPayload::class.java.name} for tenant ${request.context.tenantId}",
+                    throwable
+                )
+                externalEventResponseFactory.platformError(request.flowExternalEventContext, throwable)
```
driessamyn, line 86: As above comment re: responding to Kafka.

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessorTests.kt
```diff
@@ -0,0 +1,565 @@
...
+import org.mockito.kotlin.whenever
+import java.nio.ByteBuffer
+import java.security.PublicKey
+import java.time.Instant
+import java.util.UUID
+import kotlin.random.Random
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+ class CryptoFlowOpsRpcProcessorTests {
```
driessamyn, line 63: What's the value of duplicating this test (additional 500+ lines of code)?

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -298,30 +306,35 @@ class CryptoProcessorImpl @Activate constructor(
...
-    
-    private fun startBusProcessors(event: ConfigChangedEvent, coordinator: LifecycleCoordinator) {
-        val cryptoConfig = event.config.getConfig(CRYPTO_CONFIG)
+    }
 
+    private fun startProcessors(event: ConfigChangedEvent, coordinator: LifecycleCoordinator) {
+        val cryptoConfig = event.config.getConfig(CRYPTO_CONFIG)
 
-        // make the processors
+        // crate processors
```
driessamyn:
```suggestion
        // create processors
```

driessamyn (changes requested):
Needs review from the crypto team, but I have a few comments/questions.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
driessamyn: I don't think TTL is still relevant in the HTTP context. It should be handled by client time-out I think.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
LWogan: it looks like the response factories are being reused to generate Records and then only the FlowEvent (record.value) part is being used. Im not sure if we want to refactor this now. or do it as part of the removal of the old bus processor?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
```diff
@@ -0,0 +1,172 @@
...
+import net.corda.utilities.trace
+import net.corda.utilities.translateFlowContextToMDC
+import net.corda.utilities.withMDC
+import org.slf4j.LoggerFactory
+import java.nio.ByteBuffer
+import java.time.Duration
+import java.time.Instant
+
+@Suppress("LongParameterList")
+class CryptoFlowOpsRpcProcessor(
```
dickon, line 39: I see you aren't yet deleting `CryptoFlowOpsBusProcessor`, i.e.  the Kafka variant. That's okay for now but I would like the simplicity  on only having one transport in due course. In the mean time it's quite a bit of code duplication we can look at reducing if we think we'll be on HTTP and Kafka for a while.

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessorTests.kt
LWogan: I'm not 100% sure as ive only just taken over this PR this afternoon. But I believe the plan is to delete the other test when the rpc client is integrated and that this test would be kept.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
dickon: I don't see why TTL is not relevant; on a busy system with lots of service fabric or other middleware in the way it could take long enough to get here that the client may already have timed out.

LWogan: @mbrkic-r3 Do you know if TTL is still relevant? I'd guess not given it is a synchronous call now?

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -298,30 +306,35 @@ class CryptoProcessorImpl @Activate constructor(
...
             )
         )
-    })
-    
-    
-    private fun startBusProcessors(event: ConfigChangedEvent, coordinator: LifecycleCoordinator) {
-        val cryptoConfig = event.config.getConfig(CRYPTO_CONFIG)
+    }
 
+    private fun startProcessors(event: ConfigChangedEvent, coordinator: LifecycleCoordinator) {
```
dickon, line 311: Suggest `startHttpProcessor`, since this function does not start _all_ the processors, only the ones that respond to HTTP. I also think "CryptoOpsRpcProcessor" would be better "CryptoOpsHttpProcessor" in this  PR since  RPC in my mind includes the RPC pattern we have over Kafka. And I'd possibly even rename "CryptoOpsBusProcessor"  to "CryptoOpsKafkaProcessor" since Bus is potentially confusingly vague when you have HTTP as well.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
mbrkic-r3: Was this TTL used just for Kafka or as general TTL of event, where we should not process it after it expires?
In the first case it might not be relevant anymore if we have now timeouts on RPC calls (do we @ben-millar ?)

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
dickon: Actually disregard that, I see the interface `SyncRPCProcessor` is already in the branch, so we would have to do a lot of renaming to consistenly say HTTP rather than RPC and this isn't the time. Better to just define "RPC" in Corda 5 as usually meaning  HTTP.

LWogan: The method starts both the bus processor and http processors, so they're not all related to the http side of things. We will have 2 sets of processors for a short period until the http client is integrated and the bus processor of external events is removed. As least that is my understanding returning from holidays today.

I dont think we should include kafka in any naming as it may not be the bus implementation. The HTTP implementation isn't really a bus at least not in r3/corda terminology. The bus for us refers to Records on "topics" which are either backed by a db or kafka currently.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
LWogan: marking as resolved, lets use the first comment to discuss

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
ben-millar: I've added a 30 second timeout to the RPC client; this can be modified in the new `HttpClientFactory`.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/CryptoRPCSmokeTests.kt
dickon: Agreed we should at some point have non-happy-path scenarios; I'm not particular sure what the code does. Having looked at SyncRPCSubscriptionImpl, I think  we are returning HTTP status 200 with a response which constains a transient error and I think that's confusing. 

But that's irrelevant to this new code, and would need addressing for all uses of SyncRPCSubscriptionImpl.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
dickon: @LWogan is correct, `ExteranlEventResposneFactory` basically just maps from one set of types to another, and I'd love to get rid of it, but this isn't the time and the new code seems to do the right thing.

components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessorTests.kt
driessamyn: I don't think this test actually tests the new class though.

dickon: Huh? CryptoFlowOpsRpcProcessor is constructed and used.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/CryptoRPCSmokeTests.kt
LWogan: Looks like the RPC code here is returning FlowEvents with error payloads when passing through an error flow, which results in 200 status code. The SyncRPCSubscriptionImpl expects an exception to buble in order to return an error status code. Is this correct?

dickon: Also we've got a general lack of non-happy path unit and integration tests. Hopefuflly e2e and QA cover more of that.

LWogan: sorry ignore my above comment. holiday brain. We leave it up to the flow engine to decipher the response. The fact it is contextually a a flow error is nothing to do with the http client side of things

LWogan: added an error test for 404 error, and 200 response with ExternalEventResponse.error set

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
LWogan: I removed the TTL logic from the HTTP server side and left it in place for the bus processor

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
driessamyn: Sorry. My bad.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsRpcProcessor.kt
driessamyn: @thiagoviana , can you make sure we have JIRAs planned for the clean-up? This comment holds true of all the other instances where we added RPC operations.

driessamyn (approved):
Still not particularly keen on the test duplication, but ok for now.

dickon (approved):
Overall LGTM. I'd love to take the time to deduplicate all the duplication this adds, ahead of dropping Kafka entirely, and that should be possible, but it's fiddly to do under time pressure so let's do this for now. I'm keen to help to clean this up later when we have time available.

This does not add an HTTP version of CryptoOpsBusProcessor, so that's an inconsistency, but it's quite possibly only the flow ops that have performance implications for our key performance metrics.

thiagoviana:
> Overall LGTM. I'd love to take the time to deduplicate all the duplication this adds, ahead of dropping Kafka entirely, and that should be possible, but it's fiddly to do under time pressure so let's do this for now. I'm keen to help to clean this up later when we have time available.
> 
> This does not add an HTTP version of CryptoOpsBusProcessor, so that's an inconsistency, but it's quite possibly only the flow ops that have performance implications for our key performance metrics.

Let's address this in the next iteration. @LWogan @mbrkic-r3

PR #4823, opened 2023-10-10
CORE-17596 Adding service endpoints in MessagingConfig with support for CombinedWorker

ben-millar:
#### Description
A new ClusterIP service has been created for each worker, and the service endpoints are being passed into the FlowWorker as startup params. This PR pulls those endpoints from the BootConfig into the MessagingConfig which is used to create the subscription which backs the flow engine (Currently S&E, soon to be multi-source).

We also modify the `CombinedWorker` deployment so that it's passed values of `localhost:7000` for these endpoints.

#### Related PRs:
https://github.com/corda/corda-api/pull/1276


wzur-r3 (approved):
LGTM

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -637,5 +637,5 @@ Get the endpoint argument for a given worker
 {{- $worker := .worker }}
 {{- $workerName := printf "%s-%s-worker" (include "corda.fullname" $context) (include "corda.workerTypeKebabCase" $worker) }}
 {{- $workerServiceName := include "corda.workerInternalServiceName" $workerName }}
-{{- printf "worker.http.endpoint.%s=%s:%s" $worker $workerServiceName (include "corda.workerServicePort" $) }}
+{{- printf "endpoints.%s=%s:%s" $worker $workerServiceName (include "corda.workerServicePort" $) }}
```
mbrkic-r3, line 640: What is this used for? Should it be "serviceEndpoints"?

mbrkic-r3 (approved):
LGTM

charts/corda-lib/templates/_helpers.tpl
ben-millar: `--serviceEndpoint` is the CLI flag for the argument; `endpoints.crypto` etc. forms part of the config path for these values in the `BootConfig` https://github.com/corda/corda-api/pull/1276/files#diff-57ec9e3b2eda16e998a38e40277749be610592a9271ce9248725600d9ee14be1R46

PR #4824, opened 2023-10-10
CORE-17562 Metrics for Multi-Source Event Mediator

mbrkic-r3:
Metrics for Multi-Source Event Mediator


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/TaskManagerHelper.kt
```diff
@@ -98,11 +106,13 @@ internal class TaskManagerHelper<K : Any, S : Any, E : Any>(
     fun executeProcessorTasks(
         processorTasks: Collection<ProcessorTask<K, S, E>>
     ): List<ProcessorTask.Result<K, S, E>> {
-        return processorTasks.map { processorTask ->
-            taskManager.execute(TaskType.SHORT_RUNNING, processorTask::call)
-        }.map {
-            it.join()
-        }
+        return processorMeter.recordCallable {
```
Omar-awad, line 109: We need to measure the time required to process an inbound event and its subsequent calls in addition to separate metric to measure the external synchronous calls being made during the event processing.

mbrkic-r3:
This was replaced by https://github.com/corda/corda-runtime-os/pull/4879

PR #4825, opened 2023-10-10
CORE-10180: Escape string parameters when creating json to prevent possible injection attacks

anien:
- Escape string parameters
- Fix most of the kdoc parameters

### DIFF
--- a/tools/plugins/initial-config/build.gradle
+++ b/tools/plugins/initial-config/build.gradle
@@ -21,7 +21,6 @@ dependencies {
     kapt "info.picocli:picocli:$picocliVersion"
 
     implementation 'net.corda:corda-config-schema'
-    implementation "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"
     implementation "com.fasterxml.jackson.module:jackson-module-kotlin:$jacksonVersion"
 
     implementation project(':libs:permissions:permission-common')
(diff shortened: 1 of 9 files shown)


dickon:
Looks okay, probably needs some unit tests.

tools/plugins/initial-config/build.gradle
```diff
@@ -21,6 +21,8 @@ dependencies {
     kapt "info.picocli:picocli:$picocliVersion"
 
     implementation 'net.corda:corda-config-schema'
+    implementation "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"
```
ronanbrowne: Just double-checking you need

`implementation "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"`

as far as I see you are only using 

`com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
`

ronanbrowne (changes requested):
See dependency query

tools/plugins/initial-config/build.gradle
anien: Ah, good point. I was trying few bits and didn't check which packages I need at the end. I'll double check.

dickon:
LGTM

dickon (dismissed):
LGTM

anien:
Changing base branch shows some outstanding commits that shouldn't be part of this PR. I'll create a new one.

PR #4827, opened 2023-10-10
CORE-16778 Fix exception handling for invalid minimum platform version

YashNabar:
Fixes exception handling for minimum platform version submitted using the MGM API to improve the returned error message.

**Test:**

Screenshot 2023-10-10 at 16 43 47


libs/membership/membership-common/src/main/kotlin/net/corda/membership/lib/verifiers/GroupParametersUpdateVerifier.kt
```diff
@@ -30,7 +30,12 @@ class GroupParametersUpdateVerifier {
             }
         }
         parameters[MPV_KEY]?.let {
-            if (it.toInt() !in 50000..99999) {
+            val mpv = try {
+                it.toInt()
```
yift-r3: You can use `toIntOrNull`

yift-r3 (approved):

PR #4828, opened 2023-10-10
CORE-17429 State metadata support in message processor

mbrkic-r3:
State metadata support in message processor:

- Provides the state manager `Metadata` from the event mediator as an argument to the event processor.
- The event processor now returns a new `Metadata` object alongside the state that it wants to write.
- The event mediator overwrites the old `Metadata` with the new one, and write this back to the state manager.


components/flow/flow-mapper-service/src/integrationTest/kotlin/net/corda/session/mapper/service/integration/TestFlowMessageProcessor.kt
```diff
@@ -19,7 +20,8 @@ class TestFlowMessageProcessor(
 
     override fun onNext(
         state: Checkpoint?,
-        event: Record<String, FlowEvent>
+        event: Record<String, FlowEvent>,
+        metadata: Metadata?,
```
jujoramos: I though the idea was to encapsulate the new `StateAndEventProcessor` within a different interface and use a new data class instead of passing an extra argument?:

Screenshot 2023-10-10 at 17 15 42

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/processor/StateAndEventProcessor.kt
```diff
@@ -38,7 +39,12 @@ interface StateAndEventProcessor<K : Any, S : Any, E : Any> {
         /**
          * Flag to indicate processing failed and the State and Event should be moved to the Dead Letter Queue
          */
-        val markForDLQ: Boolean = false
+        val markForDLQ: Boolean = false,
+
+        /**
+         * The updated state metadata in response to an incoming event from [onNext].
+         */
+        val updatedMetadata: Metadata? = null,
```
JamesHR3: Will this value overwrite the old one, or will it be merged somehow?

If it is an overwrite, I think there's a danger that the processor could forget to pass the old metadata back if there's nothing to update, which could cause some difficult to track bugs. If it's a merge, that's a different behaviour to the state, which might be a bit surprising.

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/processor/StateAndEventProcessor.kt
```diff
@@ -55,7 +62,7 @@ interface StateAndEventProcessor<K : Any, S : Any, E : Any> {
      * NOTE: The returned events will be published and the processed events will be consumed atomically as a
      * single transaction.
      */
-    fun onNext(state: S?, event: Record<K, E>): Response<S>
+    fun onNext(state: S?, event: Record<K, E>, metadata: Metadata? = null): Response<S>
```
JamesHR3: I wouldn't be against merging `state` and `metadata` into a single parameter as some form of data object. It might make it clearer what the relationship is between the inputs for these parameters and the output in the `StateAndEventProcessor.Response`.

components/flow/flow-mapper-service/src/integrationTest/kotlin/net/corda/session/mapper/service/integration/TestFlowMessageProcessor.kt
jujoramos: Comment was meant to be added in [this line](https://github.com/corda/corda-runtime-os/blob/mbrkic-m3/CORE-17429/state-metadata/libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/processor/StateAndEventProcessor.kt#L65) 😓

mbrkic-r3: It was, but then it's not possible to 
`override val stateValueClass = MessageProcessor.State<FlowMapperState>::class.java` 
(due to "Only classes are allowed on the left hand side of a class literal"). Not sure if that could be solved due to type erasure.

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/processor/StateAndEventProcessor.kt
mbrkic-r3: It's overwrite, same as for state. I guess tests should show that processor forgot to return updated state. I could also delete default value and update all processors to return null if you would prefer that?

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/processor/StateAndEventProcessor.kt
mbrkic-r3: ok, let's merge them

components/flow/flow-mapper-service/src/main/kotlin/net/corda/session/mapper/service/executor/FlowMapperMessageProcessor.kt
```diff
@@ -57,9 +58,12 @@ class FlowMapperMessageProcessor(
         return traceStateAndEventExecution(event, "Flow Mapper Event - $eventType") {
             eventProcessingTimer.recordCallable {
                 if (!isExpiredSessionEvent(value)) {
-                    val executor = flowMapperEventExecutorFactory.create(key, value, state, flowConfig)
+                    val executor = flowMapperEventExecutorFactory.create(key, value, state?.value, flowConfig)
                     val result = executor.execute()
-                    StateAndEventProcessor.Response(result.flowMapperState, result.outputEvents)
+                    StateAndEventProcessor.Response(
+                        State(result.flowMapperState, metadata = null),
```
JamesHR3: I think this should return the metadata that was put in, rather than nulling it out.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/converters/impl/FlowEventContextConverterImpl.kt
```diff
@@ -4,14 +4,15 @@ import net.corda.data.flow.state.checkpoint.Checkpoint
...
+import net.corda.messaging.api.processor.StateAndEventProcessor.State
 import org.osgi.service.component.annotations.Component
 
 @Suppress("Unused")
 @Component(service = [FlowEventContextConverter::class])
 class FlowEventContextConverterImpl : FlowEventContextConverter {
     override fun convert(flowContext: FlowEventContext<*>): StateAndEventProcessor.Response<Checkpoint> {
         return StateAndEventProcessor.Response(
-            flowContext.checkpoint.toAvro(),
+            State(flowContext.checkpoint.toAvro(), metadata = null),
```
JamesHR3: Similarly, here we should probably use the input metadata.

components/flow/flow-mapper-service/src/main/kotlin/net/corda/session/mapper/service/executor/FlowMapperMessageProcessor.kt
jujoramos: +1

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/processor/StateAndEventProcessor.kt
```diff
@@ -17,6 +18,13 @@ import net.corda.messaging.api.records.Record
  * [CordaFatalException] and will cause the subscription to close.
  */
 interface StateAndEventProcessor<K : Any, S : Any, E : Any> {
```
jujoramos, line 20: Would it make sense to add a brand new `StateAndEventProcessor` and only update those processors that need updating instead of doing everything in one go?.

JamesHR3: I think we're fully plugged in now, so might as well leave it. It'll make integration easier later.

jujoramos (approved):

filipesoliveira (approved):
The changes made to the `TokenCacheEventProcessor ` look fine.

yift-r3 (approved):
LGTM

PR #4829, opened 2023-10-10
CORE-16626 Replace string UUIDs with UUID objects

ac101m:
# Changes
- Switch UUID strings to UUID objects.
- Refactor interop group policy publishing so that it no longer requires a UUID parameter.
- Fix a bug in group policy deletion.


dickon:
The public API part of this change seems that it might break customer code, so would violate https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4601151665/Evolution+of+the+CorDapp+API

szymonsztuka (approved):

szymonsztuka (approved):

szymonsztuka (approved):

PR #4831, opened 2023-10-11
CORE-16246 Add tags to enable individual properties set on topics during creation

bpaunescu:
Topic schemas can now be edited to contain multiple tags per topic

`topics:
  ConfigManagementRequestTopic:
    name: config.management.request
    tags:
      - tag01
      - tag03`

These tags can then be mapped to custom properties by passing them as arguments to the CLI topic-config tool. Currently only `partitions` is supported to allow users to specify partition numbers per topic.
Users may use the global override (`-p`) in addition to the new option `-t` to set a value for all topics except the tagged ones.

Usage:
`./corda-cli.sh topic -b localhost:9094 -k client.properties create -t tag01=partitions:5 -t tag02=partitions:2 connect`

The above command will set the default partition number 1 for all topics except the ones tagged with `tag01` and `tag02` which will have 5 and 2 respectively.

We currently support tag lists for each topic. If a topic has more than one tag which is used to set partition number, the first tag encountered in the list will be used.

### DIFF
--- a/tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Create.kt
+++ b/tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Create.kt
@@ -40,6 +40,15 @@ class Create(
     var partitionOverride: Int = 1
 
     @CommandLine.Option(
+        names = ["-t", "--tag"],
+        description = ["One or more tags associated with topics and their respective number of partitions " +
+                "e.g. -t t01=partitions:3 -t t02=partitions:5"]
+    )
+    // Ideally, this would use a custom converter and a custom data type to provide some future flexibility
+    // Potentially to be implemented as a last step. For now, we'll manually parse the inputs
+    var tagsToPropertiesMap: Map<String, String> = emptyMap()
+
+    @CommandLine.Option(
         names = ["-u", "--user"],
         description = ["One or more Corda workers and their respective Kafka users e.g. -u crypto=Charlie -u rest=Rob"]
     )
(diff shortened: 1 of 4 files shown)


tools/plugins/topic-config/src/test/kotlin/net/corda/cli/plugins/topicconfig/CreateConnectTest.kt
```diff
@@ -36,6 +40,71 @@ class CreateConnectTest {
...
+                "segment.ms" to "600000",
+                "delete.retention.ms" to "300000",
+                "min.compaction.lag.ms" to "60000",
+                "max.compaction.lag.ms" to "604800000",
+                "min.cleanable.dirty.ratio" to "0.5"
+        )))
+
+        // tag arguments do not exist, default will be set
+        command.create?.partitionOverride = 1
+        command.create?.tagsToPropertiesMap = mapOf("tag02" to "partitions:5")
```
simon-johnson-r3, line 77: Should this not throw? Otherwise end user typos would go unnoticed.

tools/plugins/topic-config/src/main/kotlin/net/corda/cli/plugins/topicconfig/Create.kt
```diff
@@ -47,6 +56,7 @@ class Create(
 
     data class TopicConfig(
         val name: String,
+        val tags: List<String> = emptyList(),
```
simon-johnson-r3, line 59: As per slack conversation, I'm in two minds about this, it's more future proof to allow multiple tags, but as suggested from the inability to test this, the current interface of the tool is a little confusing as it lets the user do things that make no sense.

PR #4832, opened 2023-10-11
CORE-16203 Replace State and Event pattern with Multi-Source Event Mediator in FlowWorker

mbrkic-r3:
- Using Multi-Source Event Mediator instead of State and Event Subscription in FlowWorker.
- Polling consumers and committing offsets performed in mediator thread in a synchronous way (without using coroutines).


ben-millar (approved):
LGTM

PR #4833, opened 2023-10-11
CORE-17622 - Refactor - Clean-up - State and Event Subscription

thiagoviana:
### CONTEXT

Due to dependencies still in development by other teams, some dummy implementations were put in place so we could speed up the development.
This code ended up being merged into the main branch.

### SOLUTION

This PR's removes the unused / unnecessary code from State and Event Subscription Implementation and Tests, and from Subscription Factory.


mbrkic-r3 (approved):
LGTM

PR #4834, opened 2023-10-11
CORE-16203 Replace State and Event pattern with Multi-Source Event Mediator in FlowMapperWorker

mbrkic-r3:
Using Multi-Source Event Mediator instead of State and Event Subscription in FlowMapperWorker.


emilybowe (approved):
LGTM

mbrkic-r3:
This was replaced by https://github.com/corda/corda-runtime-os/pull/4875

PR #4763, opened 2023-10-03
CORE-16419 - template, schema and values yaml changes to support isolated state manager db

conal-smith:
Following from previous helm chart pipeline changes, use the deployed state-manager postgres db in workers that require it and have stateManagerDbAccess enabled.

Allow state manager database to be configured by helm properties coming from values.yaml

PR #4836, opened 2023-10-11
CORE-16320 - Remove state-manager schema bootstrap configuration 

conal-smith:
**Description**
In several places we are using state manager schema from helm chart values.yaml but the worker itself cannot yet be configured to use a different schema.

**Solution**
Hard-code schema to STATE_MANAGER for now, so we can make this configurable in the future if necessary. Remove configurability of the schema in helm bootstrapping.

### DIFF
--- a/charts/corda-lib/templates/_bootstrap.tpl
+++ b/charts/corda-lib/templates/_bootstrap.tpl
@@ -157,7 +157,7 @@ spec:
               mkdir /tmp/db
               java -Dpf4j.pluginsDir=/opt/override/plugins -Dlog4j2.debug=false -jar /opt/override/cli.jar database spec \
                 -s "config,rbac,crypto,statemanager" \
-                -g "config:${DB_CLUSTER_SCHEMA},rbac:${DB_RBAC_SCHEMA},crypto:${DB_CRYPTO_SCHEMA},statemanager:${DB_STATE_MANAGER_SCHEMA}" \
+                -g "config:${DB_CLUSTER_SCHEMA},rbac:${DB_RBAC_SCHEMA},crypto:${DB_CRYPTO_SCHEMA},statemanager:STATE_MANAGER" \
                 -u "${PGUSER}" -p "${PGPASSWORD}" \
                 --jdbc-url "${JDBC_URL}" \
                 -c -l /tmp/db
@@ -249,8 +249,6 @@ spec:
               value: {{ .Values.bootstrap.db.rbac.schema | quote }}
             - name: DB_CRYPTO_SCHEMA
               value: {{ .Values.bootstrap.db.crypto.schema | quote }}
-            - name: DB_STATE_MANAGER_SCHEMA
-              value: {{ .Values.bootstrap.db.stateManager.schema | quote }}
             {{- include "corda.bootstrapClusterDbEnv" . | nindent 12 }}
             {{- include "corda.configSaltAndPassphraseEnv" . | nindent 12 }}
             {{- include "corda.bootstrapCliEnv" . | nindent 12 }}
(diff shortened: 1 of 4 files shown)


jujoramos (approved):
LGTM

davidcurrie:
> so we can make this configurable in the future if necessary

I believe that this is already necessary in that we have environments that share a single database instance and rely on being able to override the schema for multi-tenancy.

@anton-subbotin / @Peter-Mitcham - can you confirm whether my statement above is true?

anton-subbotin:
> > so we can make this configurable in the future if necessary
> 
> I believe that this is already necessary in that we have environments that share a single database instance and rely on being able to override the schema for multi-tenancy.
> 
> @anton-subbotin / @Peter-Mitcham - can you confirm whether my statement above is true?

@davidcurrie 
> we have environments that share a single database instance

that is true

> rely on being able to override the schema for multi-tenancy

not sure whether this might be required in the future, but previously in case of sharing a single Aurora RDS instance we didn't need to override the schema (in other words, we weren't overriding the affected parameter.

anton-subbotin:
@davidcurrie but assuming this:
> but the worker itself cannot yet be configured to use a different schema.

do you think that we still can have a valid scenario for:
> share a single database instance and rely on being able to override the schema for multi-tenancy
?

anton-subbotin:
@conalsmith-r3  not critical (as it shouldn't break anything), but would be nice to make a similar update in https://github.com/corda/corda5-enterprise/tree/release/ent/5.1/charts/corda-enterprise

davidcurrie:
@anton-subbotin - well, the state manager DB is new. Do you know if any of the QA tests currently override the config/crypto/RBAC schemas to allow multiple tests to proceed in parallel using the same DB instance?

anton-subbotin:
@davidcurrie 
>Do you know if any of the QA tests currently override the config/crypto/RBAC schemas to allow multiple tests to proceed in parallel using the same DB instance?

No, as far as I know. Do I understand it right that schema override would allow to share the same named _database entity_ (referred to as `$.Values.db.cluster.database`) ? If yes, then in the flow we've been using/orchestrating, the _database_ is unique per Corda deployment

conal-smith:
> @davidcurrie
> 
> > Do you know if any of the QA tests currently override the config/crypto/RBAC schemas to allow multiple tests to proceed in parallel using the same DB instance?
> 
> No, as far as I know. Do I understand it right that schema override would allow to share the same named _database entity_ (referred to as `$.Values.db.cluster.database`) ? If yes, then in the flow we've been using/orchestrating, the _database_ is unique per Corda deployment

Schema override here would just change the schema inside $.Values.db.cluster.database.

If you deploy a separate cluster on the same RDS instance with a different $.Values.db.cluster.database, you'll have a separate state-manager bootstrapped in that database.

e.g. if two Corda helm installs override `$.Values.db.cluster.database` with "test-cluster-1" and "test-cluster-2" respectively, then the single RDS instance will have two databases:
- "test-cluster-1" 
  - "STATE_MANAGER"
  - "config"
  - "rbac"
- "test-cluster-2"
  - "STATE_MANAGER"
  - "config"
  - "rbac"

Soon, when we complete https://github.com/corda/corda-runtime-os/pull/4763, the state-manager will not be in the cluster db instance and the state-manager host and database will be configurable.

In summary, I think (and correct me if I'm wrong) we can go ahead with this change, schema here won't impact deployment of multiple clusters against the same RDS instance.

davidcurrie:
Agreed - I thought we might be using a single database with different schema but it sounds like that is not the case. We should make sure that the docs indicate that anyone using the CLI directly needs to specify `STATE_MANAGER` as the schema.

davidcurrie (approved):

anton-subbotin (approved):

PR #4837, opened 2023-10-11
CORE-15552 Add `findUnconsumedStatesByExactType` paging API

lankydan:
Add findUnconsumedStatesByExactType to UtxoLedgerService , allowing developers to retrieve states by type from the vault in pages. This dissuades developers from retrieving every state from the vault in one go and returning them to a flow.

Reuse the vault named query logic to create a
`findUnconsumedStatesByExactType` API that supports paging without too much extra work.

`DefaultVaultNamedQueryFactoryImpl` registers a query for finding states by their exact type.

This factory is then manually registered in
`VaultNamedQueryFactoryProvider`.

We could potentially add a better lookup mechanism in the future, but there is unlikely to be additional `VaultNamedQueryFactory`s that we register in the future.


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/impl/DefaultVaultNamedQueryFactory.kt
```diff
@@ -0,0 +1,5 @@
+package net.corda.ledger.persistence.query.impl
+
+import net.corda.v5.ledger.utxo.query.VaultNamedQueryFactory
+
+interface DefaultVaultNamedQueryFactory : VaultNamedQueryFactory
```
relyafi: As an interface, I think this would benefit from a KDoc.

Also, I wonder if `PlatformVaultNamedQueryFactory` would be a more accurate name? Default suggests to me that this is something that would be used if there were no other query factories available.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/impl/DefaultVaultNamedQueryFactory.kt
```diff
@@ -0,0 +1,5 @@
+package net.corda.ledger.persistence.query.impl
```
relyafi: Isn't an `impl` package an inappropriate place for an interface?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/registration/impl/VaultNamedQueryFactoryProvider.kt
```diff
@@ -17,11 +19,13 @@ import org.slf4j.LoggerFactory
...
-    service = [ UsedByPersistence::class ],
+    service = [UsedByPersistence::class],
     property = [SandboxConstants.CORDA_MARKER_ONLY_SERVICE],
     scope = ServiceScope.PROTOTYPE
 )
 class VaultNamedQueryFactoryProvider @Activate constructor(
+    @Reference(service = DefaultVaultNamedQueryFactory::class)
+    private val defaultVaultNamedQueryFactory: DefaultVaultNamedQueryFactory,
     @Reference(service = VaultNamedQueryBuilderFactory::class)
     private val vaultNamedQueryBuilderFactory: VaultNamedQueryBuilderFactory
```
relyafi: Actually, looking at this, I'm unconvinced as to why we need an entirely separate factory. Can't we just always register "platform" queries with the existing `VaultNamedQueryRegistry`, and use the existing factory?

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -61,12 +61,9 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
...
                 AND tc_output.leaf_idx = tc_output_info.leaf_idx
                 AND tc_output.group_idx = ${UtxoComponentGroup.OUTPUTS.ordinal}
-            JOIN {h-schema}utxo_transaction_output AS tx_o
-                ON tx_o.transaction_id = tc_output.transaction_id
-                AND tx_o.leaf_idx = tc_output.leaf_idx
             JOIN {h-schema}utxo_transaction AS tx
-                ON tx.transaction_id = tx_o.transaction_id
-            WHERE tx_o.type = :type    
+                ON tx.id = tc_output.transaction_id
+            WHERE vto.type = :type
```
relyafi: I'm not sure why this is appearing in the diff. Looks like it's part of @nkovacsx recent data model simplification changes, so if we've merged / rebased these then this shouldn't be a diff?

lankydan:
API PR - https://github.com/corda/corda-api/pull/1292

relyafi (commented):
I think the overall approach to re-using the existing custom query logic is a good idea, but I'm unconvinced about the need for a separate factory for this query.

Also, I'd like to see some new E2E test cases proving that the query works with the paging functionality. If we want to do this in the E2E test repo (rather than smoke tests in this PR) that's fine, but we should reference that PR from this one.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
lankydan: It is a correction, there wasn't an E2E test around this so it failed when tested.

lankydan:
E2E tests - https://github.com/corda/corda-e2e-tests/pull/276

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/impl/DefaultVaultNamedQueryFactory.kt
lankydan: I put it inline with `DefaultContractStateVaultJsonFactory` which we already have. I can rename both to `platform` though, I don't have any issues with doing that.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/registration/impl/VaultNamedQueryFactoryProvider.kt
lankydan: That was my original solution; however, leveraging a `VaultNamedQueryFactory` is much easier. This is because the registration step occurs as part of `VaultNamedQueryBuilder.register`, meaning that we have to reimplement or extract much more code to achieve the same outcome.

I could remove the OSGI around the `DefaultVaultNamedQueryFactory` though and construct it directly within `VaultNamedQueryFactoryProvider`'s constructor.

vlajos (dismissed):
It needs a sync, but otherwise, LGTM.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
lankydan: This code is now deleted anyway.

vlajos (dismissed):

vlajos (approved):

blsemo (approved):
LGTM

PR #4839, opened 2023-10-11
CORE-17368 - Revert fix for missing config 'frequencyOfExpirationPoll' in ExpirationProcessor

nikinagy:
We added a new config value to define the polling frequency for the ExpirationProcessor. Previously, the processor failed to reach a healthy lifecycle state when we did a platform upgrade, because there was no mechanism to push the newly introduced config values to Kafka.
We added a temporary fix which would set the value of the frequencyOfExpirationPoll to a default in case of a missing config exception happens on upgrade.

We now have in place a fix to force initial config reconciliation on platform upgrade scenarios, so this temporary fix should be safe to remove.

Testing done:
- Deployed 5.0.
- Registered some membership data.
- Updated the platform to 5.1 image published by this PR.
- Waited for membership worker to come up.
- Check the logs looking for `ConfigException`s, but was all clear.

### DIFF
--- a/components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/mgm/ExpirationProcessorTest.kt
+++ b/components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/mgm/ExpirationProcessorTest.kt
@@ -246,32 +246,32 @@ class ExpirationProcessorTest {
             verify(publisher).start()
         }
 
-        // test required for 5.0 to 5.1 platform upgrade
         @Test
-        fun `use hard-coded defaults until newest version of config is received via reconciliation`() {
-            val dummyTime = 100L
-            whenever(membershipConfig.getLong(MAX_DURATION_BETWEEN_EXPIRED_REGISTRATION_REQUESTS_POLLS))
-                .thenThrow(ConfigException.Missing("")).thenReturn(dummyTime)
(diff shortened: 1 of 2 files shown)


components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/mgm/ExpirationProcessorTest.kt
```diff
@@ -246,33 +245,33 @@ class ExpirationProcessorTest {
...
+                .thenReturn(dummyTime)
+            whenever(membershipConfig.getLong(EXPIRATION_DATE_FOR_REGISTRATION_REQUESTS))
+                .thenReturn(dummyTime)
+
             handler.firstValue.processEvent(configChangedEvent, coordinator)
 
             verify(coordinator, times(2)).setTimer(any(), capturedTimeframes.capture(), capturedEvents.capture())
 
             SoftAssertions.assertSoftly {
                 it.assertThat(capturedTimeframes.firstValue).isLessThanOrEqualTo(TimeUnit.MINUTES.toMillis(180))
```
williamvigorr3: I think this test shouldn't post two config change events anymore. We can probably just change this line to `it.assertThat(capturedTimeframes.secondValue).isLessThanOrEqualTo(TimeUnit.MINUTES.toMillis(dummyTime))` and remove the second assert. Probably the verify needs to change also.

williamvigorr3 (commented):
Implementation looks good.

williamvigorr3 (approved):
LGTM

PR #4841, opened 2023-10-11
CORE-17688: Fix sandbox level cache

jennyang-r3:
## Why this PR?
We have `CacheKey` data class with holidingIdentity in sandbox. When CPI is upgraded with the same bundle name, holding Identity doesn't change, hence it picks up previous version cached with the same holding identity. This caused failure in CPI upgrade so this PR fixes the issue.

### DIFF
--- a/components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/context/FlowServiceTestContext.kt
+++ b/components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/context/FlowServiceTestContext.kt
@@ -426,12 +426,7 @@ class FlowServiceTestContext @Activate constructor(
     override fun resetFlowFiberCache() {
     ALL_TEST_VIRTUAL_NODES.forEach {
         flowFiberCache.remove(
-            VirtualNodeContext(
-                it.toCorda(),
-                setOf(CPK1_CHECKSUM),
-                FLOW,
-                null
-            )
+            VirtualNodeContext(it.toCorda(), setOf(CPK1_CHECKSUM), FLOW, null)
         )
     }
 }
--- a/components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/context/FlowServiceTestContext.kt
+++ b/components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/context/FlowServiceTestContext.kt
@@ -37,6 +37,7 @@ import net.corda.flow.testing.fakes.FakeFlowFiberFactory
 import net.corda.flow.testing.fakes.FakeMembershipGroupReaderProvider
 import net.corda.flow.testing.fakes.FakeSandboxGroupContextComponent
 import net.corda.flow.testing.tests.ALL_TEST_VIRTUAL_NODES
+import net.corda.flow.testing.tests.CPK1_CHECKSUM
 import net.corda.flow.testing.tests.FLOW_NAME
 import net.corda.flow.testing.tests.SESSION_PROPERTIES
 import net.corda.flow.utils.KeyValueStore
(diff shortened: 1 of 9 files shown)


blsemo (dismissed):
LGTM

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/context/FlowServiceTestContext.kt
```diff
@@ -421,8 +424,17 @@ class FlowServiceTestContext @Activate constructor(
...
+        flowFiberCache.remove(
+            VirtualNodeContext(
+                it.toCorda(),
+                setOf(CPK1_CHECKSUM),
+                FLOW,
+                null
+            )
+        )
     }
+}
```
lankydan: What happened to this formatting?

components/flow/flow-service/src/test/kotlin/net/corda/flow/application/crypto/MySigningKeysCacheImplTest.kt
```diff
@@ -70,16 +85,21 @@ class MySigningKeysCacheImplTest {
...
+        whenever(currentSandboxGroupContext.get()).thenReturn(sandbox)
+
+        // alice put cache
+        mySigningKeysCache.putAll(mapOf(KEY_A to KEY_A, KEY_B to null))
+        // bob put cache
+        mySigningKeysCache.putAll(mapOf(KEY_C to KEY_C, KEY_D to null))
+        mySigningKeysCache.remove(aliceVirtualNodeContext)
 
+        // there should bob's cache only
         assertThat(mySigningKeysCache.get(setOf(KEY_A, KEY_B, KEY_C, KEY_D))).containsExactlyInAnyOrderEntriesOf(
```
lankydan, line 110: We should also do a get on alice's vnode context to check that it is empty.

This means we should make the whenever:

```
whenever(sandbox.virtualNodeContext).thenReturn(
            aliceVirtualNodeContext,
            bobVirtualNodeContext
        )
```

Something like:

```
whenever(virtualNodeContext.holdingIdentity).thenReturn(
            ALICE_X500_HOLDING_IDENTITY.toCorda(),
            BOB_X500_HOLDING_IDENTITY.toCorda(),
            ALICE_X500_HOLDING_IDENTITY.toCorda(),
            BOB_X500_HOLDING_IDENTITY.toCorda()
```

Then have the assert `assertThat(mySigningKeysCache.get(setOf(KEY_A, KEY_B, KEY_C, KEY_D)))` return nothing the first time, since its using alice's vnode which was removed.

Then the assert that is already there stays unchanged.

lankydan: We should check if similar logic applies to the other tests as well.

libs/virtual-node/sandbox-group-context/src/main/kotlin/net/corda/sandboxgroupcontext/SandboxedCache.kt
```diff
@@ -20,25 +20,18 @@ interface SandboxedCache {
     /**
      * A cache key holding sandbox information and the _real_ key to the cache.
      *
-     * @property holdingIdentity The [HoldingIdentity] of the sandbox that added the key-value pair.
+     * @property virtualNodeContext The [VirtualNodeContext] of the sandbox that added the key-value pair.
```
simon-johnson-r3, line 21: Header for this class states
```
 * All [SandboxedCache]s should contain an internal cache that contains _at least_ the [HoldingIdentity] of the sandbox that put key-value
 * pairs into the cache.
```
so will need changing too

simon-johnson-r3 (commented):
Just one minor comment about a comment

simon-johnson-r3 (approved):

components/flow/flow-service/src/integrationTest/kotlin/net/corda/flow/testing/context/FlowServiceTestContext.kt
```diff
@@ -421,8 +424,12 @@ class FlowServiceTestContext @Activate constructor(
     }
 
     override fun resetFlowFiberCache() {
-        ALL_TEST_VIRTUAL_NODES.forEach { flowFiberCache.remove(it.toCorda()) }
+    ALL_TEST_VIRTUAL_NODES.forEach {
+        flowFiberCache.remove(
+            VirtualNodeContext(it.toCorda(), setOf(CPK1_CHECKSUM), FLOW, null)
+        )
     }
+}
```
lankydan, line 432: This still looks wrong.

lankydan: Formatting wise

lankydan (approved):

emilybowe (approved):
LGTM

PR #4846, opened 2023-10-12
CORE-16323: Fix State Manager Optimistic Locking

jujoramos:
Fix update and delete SQL statements within the State Manager
implementation so only the states with the specified key and
version are updated. This prevents race conditions while executing
the optimistic locking and also reduces by half the amount of SQL
statements that needs be executed for every update and delete.

### DIFF
--- a/libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
+++ b/libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
@@ -1,6 +1,7 @@
 package net.corda.libs.statemanager.impl.tests
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.convertToMetadata
 import net.corda.db.admin.impl.ClassloaderChangeLog
 import net.corda.db.admin.impl.LiquibaseSchemaMigratorImpl
 import net.corda.db.schema.DbSchema
@@ -13,7 +14,6 @@ import net.corda.libs.statemanager.api.State
 import net.corda.libs.statemanager.api.StateManager
 import net.corda.libs.statemanager.api.metadata
 import net.corda.libs.statemanager.impl.StateManagerImpl
-import net.corda.libs.statemanager.impl.convertToMetadata
 import net.corda.libs.statemanager.impl.model.v1.StateEntity
 import net.corda.libs.statemanager.impl.model.v1.StateManagerEntities
 import net.corda.libs.statemanager.impl.repository.impl.KEY_PARAMETER_NAME
--- a/libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
+++ b/libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
@@ -1,7 +1,7 @@
 package net.corda.libs.statemanager.impl.tests
 
 import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.databind.convertToMetadata
+import net.corda.libs.statemanager.impl.convertToMetadata
 import net.corda.db.admin.impl.ClassloaderChangeLog
 import net.corda.db.admin.impl.LiquibaseSchemaMigratorImpl
 import net.corda.db.schema.DbSchema
--- a/libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
+++ b/libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
@@ -1,7 +1,6 @@
 package net.corda.libs.statemanager.impl.tests
 
 import com.fasterxml.jackson.databind.ObjectMapper
-import net.corda.libs.statemanager.impl.convertToMetadata
 import net.corda.db.admin.impl.ClassloaderChangeLog
 import net.corda.db.admin.impl.LiquibaseSchemaMigratorImpl
 import net.corda.db.schema.DbSchema
@@ -14,6 +13,7 @@ import net.corda.libs.statemanager.api.State
 import net.corda.libs.statemanager.api.StateManager
 import net.corda.libs.statemanager.api.metadata
 import net.corda.libs.statemanager.impl.StateManagerImpl
+import net.corda.libs.statemanager.impl.convertToMetadata
 import net.corda.libs.statemanager.impl.model.v1.StateEntity
 import net.corda.libs.statemanager.impl.model.v1.StateManagerEntities
 import net.corda.libs.statemanager.impl.repository.impl.KEY_PARAMETER_NAME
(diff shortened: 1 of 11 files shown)


libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -1,6 +1,7 @@
 package net.corda.libs.statemanager.impl.tests
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import net.corda.libs.statemanager.impl.convertToMetadata
```
JamesHR3: What happened with this import? Shouldn't it be with the others?

jujoramos: IntelliJ hates me, that's what happened 🤣. Fixed.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/PostgresQueryProvider.kt
```diff
@@ -16,7 +16,7 @@ class PostgresQueryProvider : AbstractQueryProvider() {
         get() = """
             UPDATE $STATE_MANAGER_TABLE SET
             key = :$KEY_PARAMETER_NAME, value = :$VALUE_PARAMETER_NAME, version = version + 1, metadata = CAST(:$METADATA_PARAMETER_NAME as JSONB), modified_time = CURRENT_TIMESTAMP AT TIME ZONE 'UTC'
-            WHERE key = :$KEY_PARAMETER_NAME
+            WHERE key = :$KEY_PARAMETER_NAME AND version = :$VERSION_PARAMETER_NAME
```
sprintengine, line 19: EDIT - we're updating 1 at a time so we get back a 0 or 1 if the row was updated 👍 ignore below

is this where we'd add `RETURNING key`?

I've never done it myself, I guess query would look:

```
UPDATE $STATE_MANAGER_TABLE
SET
    key = :$KEY_PARAMETER_NAME,
    value = :$VALUE_PARAMETER_NAME,
    version = version + 1,
    metadata = CAST(:$METADATA_PARAMETER_NAME as JSONB),
    modified_time = CURRENT_TIMESTAMP AT TIME ZONE 'UTC'
WHERE key = :$KEY_PARAMETER_NAME
AND version = :$VERSION_PARAMETER_NAME
RETURNING key;
```

jujoramos: We are updating/deleting a single row per statement so we don't need the `RETURNING` keyword in this case: if the update operation returns something other than zero we know that the query didn't touch any rows and, as such, we can safely infer that somebody else already overwrote the same state.

sprintengine: I'm not sure using entitymanager.executeUpdate if it's even possible to get results back other than the number of rows updated..

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -86,26 +59,26 @@ class StateManagerImpl(
 
     override fun update(states: Collection<State>): Map<String, State> {
         entityManagerFactory.transaction { em ->
-            val (updatable, mismatchVersions) = checkVersionAndPrepareEntitiesForPersistence(states, em)
-            stateRepository.update(em, updatable)
+            val failedUpdates = stateRepository.update(em, states.map { it.toPersistentEntity() })
```
owenstanford: ~would this not be the successful updates? the returning will return the key of the rows that were updated, the diff of these from the set of submitted keys would be the fails~
Ignore I see you are tracking this in the repo class

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -33,29 +33,50 @@ class StateRepositoryImpl(private val queryProvider: QueryProvider) : StateRepos
...
+                    .executeUpdate().also {
+                        if (it == 0) {
+                            failedKeys.add(state.key)
+                        }
+                    }
             }
+
+            return failedKeys
         } catch (e: Exception) {
             logger.warn("Failed to updated batch of states - ${states.joinToString { it.key }}", e)
```
sprintengine: should this be `failedKeys.joinToString...` ? (and moving failedKeys declaration above `try {`)

sprintengine: Oh we run entire batch in a transaction so they'll all fail

jujoramos: Yep

sprintengine: Feels like this warning should be a level above, i.e. the update function doesn't care if there is a transaction and the caller can report that entire batch failed

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
jujoramos: 👍

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -33,29 +33,50 @@ class StateRepositoryImpl(private val queryProvider: QueryProvider) : StateRepos
             .setParameter(KEYS_PARAMETER_NAME, keys)
             .resultListAsStateEntityCollection()
 
-    override fun update(entityManager: EntityManager, states: Collection<StateEntity>) =
+    override fun update(entityManager: EntityManager, states: Collection<StateEntity>): Collection<String> {
         try {
-            states.forEach {
+            val failedKeys = mutableListOf<String>()
+
+            states.forEach { state ->
```
owenstanford: Probably not for this PR but at some point we should switch to a batch update as it should be quicker that single updates

jujoramos: Agreed.

owenstanford (dismissed):
looks ok to me
If we don't already have it, we should create a ticket to add metrics to the state manager, I think we will want to know throughput and read/write TPS + latency once we start performance testing

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
jujoramos: Done

owenstanford (approved):

conal-smith (approved):
lgtm

PR #4847, opened 2023-10-12
CORE-17371 fix CPI signerSummaryHash and fileChecksum in FlowContextProperties

vlajos:
CORE-17371 fix CPI signerSummaryHash and fileChecksum in FlowContextProperties
Using toString() instead of toHexString() prepends them with the used algorithm name.

Related ledger metadata fields before the change:
image
And after:
image


ben-millar (approved):
LGTM

PR #4852, opened 2023-10-12
CORE-17297: Fix incompatible responder flow failure handling

Tom-Fitzpatrick:
This PR addresses the failing scenario where two vNodes with with incompatible flow protocol versions try to communicate. 

The behaviour we are seeing in the e2e test [here](https://github.com/corda/corda-e2e-tests/blob/5b9558c030982e0a26adbbc8d2b05f094ef3cfe3/e2etest/src/e2eTest/kotlin/net/corda/e2etest/virtualnode/VirtualNodeUpgradeFlowCompatibilityTest.kt#L154) shows the responder flow hits an error when initiating the checkpoint (FlowFatalException), this is caught in the `runCatching`  in `SessionEventHandler.initializeCheckpointAndGetResult` and results in calling `SendErrorMessage` to kill the initiating flow, but this fails with a `FlowSessionStateException`. This goes into the generic error handling in `FlowEventProcessorImpl.getFlowPipelineResponse` and we fail to kill the initiating flow, leaving in in an invalid RUNNING state indefinitely. 

To handle this:
This changes `SessionEventHandler` logic during checkpoint initialization to throw `FlowFatalException` when no valid responder available for flow protocol. This means we don't finish initializing the checkpoint. 

Handling of `FlowFatalException` in `FlowEventExceptionProcessorImpl` is modified to correctly create a `SessionError` record in this case without using the unitialized checkpoint. This kills the flow correctly and it gets `FLOW_FAILED` status. 

The e2e test on https://github.com/corda/corda-e2e-tests/pull/256 passes with this change and will be reenabled once this fix goes in.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
@@ -1,11 +1,15 @@
 package net.corda.flow.pipeline.impl
 
+import net.corda.data.ExceptionEnvelope
 import net.corda.data.flow.FlowKey
+import net.corda.data.flow.event.FlowEvent
+import net.corda.data.flow.event.MessageDirection
 import net.corda.data.flow.event.SessionEvent
 import net.corda.data.flow.event.StartFlow
 import net.corda.data.flow.event.mapper.FlowMapperEvent
 import net.corda.data.flow.event.mapper.ScheduleCleanup
 import net.corda.data.flow.event.session.SessionData
+import net.corda.data.flow.event.session.SessionError
 import net.corda.data.flow.output.FlowStates
 import net.corda.data.flow.output.FlowStatus
 import net.corda.data.flow.state.session.SessionProcessState
@@ -30,6 +34,7 @@ import net.corda.flow.utils.toMap
 import net.corda.libs.configuration.SmartConfig
 import net.corda.libs.configuration.getLongOrDefault
 import net.corda.messaging.api.records.Record
+import net.corda.schema.Schemas
 import net.corda.schema.configuration.FlowConfig
 import net.corda.schema.configuration.FlowConfig.PROCESSING_MAX_RETRY_WINDOW_DURATION
 import net.corda.session.manager.Constants
@@ -146,6 +151,7 @@ class FlowEventExceptionProcessorImpl @Activate constructor(
 
         var errorEvents: List<net.corda.messaging.api.records.Record<*, *>> = emptyList()
         var cleanupEvents: List<net.corda.messaging.api.records.Record<*, *>> = emptyList()
+        var records: List<net.corda.messaging.api.records.Record<*, *>> = emptyList()
 
         if(context.inputEventPayload is SessionEvent) {
             val inputPayload = context.inputEventPayload as SessionEvent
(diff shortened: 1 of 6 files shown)


vkolomeyko:
Looks like there is a problem with Config change test: `cluster configuration changes are picked up and workers continue to operate normally() – net.corda.applications.workers.smoketest.flow.ConfigurationChangeTest`

After config change the flow is stuck in `START_REQUESTED` state and Combined Worker seems to be unhealthy due to:
```
10:03:17.907 [state/event processing thread membership.ops.async.retries-membership.async.request.retries] ERROR net.corda.messaging.subscription.StateAndEventSubscriptionImpl-STATE_AND_EVENT--membership.ops.async.retries--membership.async.request.retries--6385a329-ebff-4adc-a9e2-2d63713731d0 {} - runConsumeLoop Throwable caught, subscription in an unrecoverable bad state
net.corda.messaging.api.exception.CordaMessageAPIFatalException: Consumer not part of consumer group membership.ops.async.retries
	at net.corda.messagebus.db.consumer.ConsumerGroup.getTopicPartitionsFor$db_message_bus_impl(ConsumerGroup.kt:36) ~[?:?]
	at net.corda.messagebus.db.consumer.DBCordaConsumerImpl.updateTopicPartitions(DBCordaConsumerImpl.kt:262) ~[?:?]
	at net.corda.messagebus.db.consumer.DBCordaConsumerImpl.close(DBCordaConsumerImpl.kt:242) ~[?:?]
	at net.corda.messaging.subscription.consumer.StateAndEventConsumerImpl.close(StateAndEventConsumerImpl.kt:204) ~[corda-messaging-impl-5.1.0.0-SNAPSHOT.jar:?]
	at net.corda.messaging.subscription.StateAndEventSubscriptionImpl.closeStateAndEventProducerConsumer(StateAndEventSubscriptionImpl.kt:180) ~[corda-messaging-impl-5.1.0.0-SNAPSHOT.jar:?]
	at net.corda.messaging.subscription.StateAndEventSubscriptionImpl.runConsumeLoop(StateAndEventSubscriptionImpl.kt:171) ~[corda-messaging-impl-5.1.0.0-SNAPSHOT.jar:?]
	at net.corda.messaging.subscription.StateAndEventSubscriptionImpl.access$runConsumeLoop(StateAndEventSubscriptionImpl.kt:38) ~[corda-messaging-impl-5.1.0.0-SNAPSHOT.jar:?]
	at net.corda.messaging.subscription.StateAndEventSubscriptionImpl$threadLooper$1.invoke(StateAndEventSubscriptionImpl.kt:56) ~[corda-messaging-impl-5.1.0.0-SNAPSHOT.jar:?]
	at net.corda.messaging.subscription.StateAndEventSubscriptionImpl$threadLooper$1.invoke(StateAndEventSubscriptionImpl.kt:56) ~[corda-messaging-impl-5.1.0.0-SNAPSHOT.jar:?]
	at net.corda.messaging.subscription.ThreadLooper.runConsumeLoop(ThreadLooper.kt:174) [corda-messaging-impl-5.1.0.0-SNAPSHOT.jar:?]
	at net.corda.messaging.subscription.ThreadLooper.access$runConsumeLoop(ThreadLooper.kt:27) [corda-messaging-impl-5.1.0.0-SNAPSHOT.jar:?]
	at net.corda.messaging.subscription.ThreadLooper$StoppableThread$start$1$1.invoke(ThreadLooper.kt:106) [corda-messaging-impl-5.1.0.0-SNAPSHOT.jar:?]
	at net.corda.messaging.subscription.ThreadLooper$StoppableThread$start$1$1.invoke(ThreadLooper.kt:106) [corda-messaging-impl-5.1.0.0-SNAPSHOT.jar:?]
	at kotlin.concurrent.ThreadsKt$thread$thread$1.run(Thread.kt:30) [kotlin-osgi-bundle-1.8.21.jar:1.8.21]
```
FAO: @driessamyn and @JamesHR3

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
```diff
@@ -138,40 +144,71 @@ class FlowEventExceptionProcessorImpl @Activate constructor(
...
+        var cleanupEvents: List<Record<*, *>> = emptyList()
+        var records: List<Record<*, *>> = emptyList()
 
-        if(activeSessionIds.isNotEmpty()) {
-            checkpoint.putSessionStates(
-                flowSessionManager.sendErrorMessages(
-                    context.checkpoint, activeSessionIds, exception, exceptionHandlingStartTime
-                )
+        val inputPayload = context.inputEventPayload as? SessionEvent
+        if(inputPayload?.payload is SessionData ) {
```
LWogan, line 152: Why is SessionData singled out here as the only SessionEvent payload to response with an error in this way?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
```diff
@@ -138,40 +144,71 @@ class FlowEventExceptionProcessorImpl @Activate constructor(
...
 
-        val errorEvents =
-            flowSessionManager.getSessionErrorEventRecords(checkpoint, context.flowConfig, exceptionHandlingStartTime)
-        val cleanupEvents = createCleanupEventsForSessions(
-            getScheduledCleanupExpiryTime(context, exceptionHandlingStartTime),
-            checkpoint.sessions.filterNot { it.hasScheduledCleanup }
-        )
+            errorEvents = listOf(sessionError)
+        } else {
+            val activeSessionIds = getActiveSessionIds(checkpoint)
```
LWogan, line 161: This code terminates all active sessions as the checkpoint is being deleted . But now it is based on a condition around the input payload not being a SessionData. Doesn't this mean that we  will not terminate other sessions in some cases?

LWogan (commented):
The description of the PR doesn't adequately explain what the problem is and how it is solved. The changes look quite subtle. Could you add more detail please?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
Tom-Fitzpatrick: Thanks Lorcan, I updated the description with more info about the problem we are tackling here, but basically, when we have two vNodes with incompatible flow protocol versions, we are not failing the initiating flow correctly and it is left running indefinitely. I have more detail in the description now. 

To answer your question here, along with the changes in SessionEventHandler, receiving SessionData here captures the scenario which is failing in our e2e test, with two incompatible flow versions on vNodes. 

I chatted with @JamesHR3 as I was looking into this bug, and this was the initial thought for when we want to create a SessionError record in this way. I'm not sure if there are other scenarios where we might hit a FlowFatalException with a SessionData payload, so maybe we could make the logic here more targeted. Very open to suggestions here.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt
Tom-Fitzpatrick: True, that is possible. In the case with two vNodes with incompatible flows, we fail before the checkpoint is initialized, so we cannot call getActiveSessionIds or we hit errors for trying to access the flow checkpoint.

It may be that we need to identify this scenario with better logic than the payload being SessionData. Or perhaps I could throw a new type of exception when this happens and handle it in a separate handler?

Tom-Fitzpatrick:
> The description of the PR doesn't adequately explain what the problem is and how it is solved. The changes look quite subtle. Could you add more detail please?

PR #4854, opened 2023-10-12
CORE-17514 Change the Link Manager to handle bad group policy

williamvigorr3:
This can happen if we can't query the DB worker for the MGM's group policy.

### DIFF
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/common/MessageConverter.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/common/MessageConverter.kt
@@ -34,6 +34,7 @@ import org.apache.avro.AvroRuntimeException
 import org.slf4j.LoggerFactory
 import java.io.IOException
 import java.nio.ByteBuffer
+import net.corda.membership.lib.exceptions.BadGroupPolicyException
 
 /**
  * This class contains code which can be used to convert between [LinkOutMessage]/[LinkInMessage] and
@@ -231,12 +232,16 @@ class MessageConverter {
                         "which is not in the network map. The message was discarded.")
                 return null
             }
-            val p2pParams = groupPolicyProvider.getP2PParameters(source)
+            val p2pParams = try {
+                groupPolicyProvider.getP2PParameters(source)
+            } catch (except: BadGroupPolicyException) {
+                logger.warn("The group policy data is unavailable or cannot be parsed for our identity = $source Error ${except.message}" +
+                    " The message was discarded.")
+                return null
+            }
             if (p2pParams == null) {
-                logger.warn(
-                    "Could not find the group info in the " +
-                        "GroupPolicyProvider for our identity = $source. The message was discarded."
-                )
+                logger.warn("Could not find the group info in the GroupPolicyProvider for our identity = $source." +
+                    " The message was discarded.")
                 return null
             }
 
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/common/MessageConverter.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/common/MessageConverter.kt
@@ -235,7 +235,7 @@ class MessageConverter {
             val p2pParams = try {
                 groupPolicyProvider.getP2PParameters(source)
             } catch (except: BadGroupPolicyException) {
-                logger.warn("The group policy data is unavailable or cannot be parsed for our identity = $source Error ${except.message}" +
+                logger.warn("The group policy data is unavailable or cannot be parsed for our identity = $source. Error ${except.message}" +
                     " The message was discarded.")
                 return null
             }
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/common/MessageConverter.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/common/MessageConverter.kt
@@ -235,8 +235,8 @@ class MessageConverter {
             val p2pParams = try {
                 groupPolicyProvider.getP2PParameters(source)
             } catch (except: BadGroupPolicyException) {
-                logger.warn("The group policy data is unavailable or cannot be parsed for our identity = $source. Error ${except.message}" +
-                    " The message was discarded.")
+                logger.warn("The group policy data is unavailable or cannot be parsed for our identity = $source. Error:" +
+                    " ${except.message}. The message was discarded.")
                 return null
             }
             if (p2pParams == null) {
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
@@ -437,7 +437,7 @@ internal class SessionManagerImpl(
         val p2pParams = try {
             groupPolicyProvider.getP2PParameters(counterparties.ourId)
         } catch (except: BadGroupPolicyException) {
-            logger.warn("The group policy data is unavailable or cannot be parsed for ${counterparties.ourId}. Error ${except.message}. " +
+            logger.warn("The group policy data is unavailable or cannot be parsed for ${counterparties.ourId}. Error: ${except.message}. " +
                 "The sessionInit message was not sent.")
             return emptyList()
         }
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
@@ -95,8 +95,10 @@ import java.util.concurrent.ScheduledExecutorService
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicReference
 import java.util.concurrent.locks.ReentrantReadWriteLock
+import net.corda.membership.lib.exceptions.BadGroupPolicyException
 import net.corda.p2p.crypto.protocol.api.InvalidSelectedModeError
 import net.corda.p2p.crypto.protocol.api.NoCommonModeError
+import net.corda.p2p.linkmanager.sessions.SessionManagerWarnings.badGroupPolicy
 import kotlin.concurrent.read
 import kotlin.concurrent.write
 
@@ -432,7 +434,13 @@ internal class SessionManagerImpl(
         multiplicity: Int
     ): List<Pair<AuthenticationProtocolInitiator, InitiatorHelloMessage>> {
 
-        val p2pParams = groupPolicyProvider.getP2PParameters(counterparties.ourId)
+        val p2pParams = try {
+            groupPolicyProvider.getP2PParameters(counterparties.ourId)
+        } catch (except: BadGroupPolicyException) {
+            logger.warn("The group policy data is unavailable or cannot be parsed for ${counterparties.ourId}. Error ${except.message} " +
+                "The sessionInit message was not sent.")
+            return emptyList()
+        }
         if (p2pParams == null) {
             logger.warn(
                 "Could not find the p2p parameters in the GroupPolicyProvider for ${counterparties.ourId}." +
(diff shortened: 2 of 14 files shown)


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -549,7 +557,13 @@ internal class SessionManagerImpl(
             return null
         }
 
-        val p2pParams = groupPolicyProvider.getP2PParameters(sessionCounterparties.ourId)
+        val p2pParams = try {
+            groupPolicyProvider.getP2PParameters(sessionCounterparties.ourId)
+        } catch (except: BadGroupPolicyException) {
+            logger.warn("The group policy data is unavailable or cannot be parsed for ${sessionCounterparties.ourId}. Error: " +
```
nikinagy, line 563: Can we have a function for these messages as well?

components/membership/group-policy-impl/src/main/kotlin/net/corda/membership/impl/grouppolicy/GroupPolicyProviderImpl.kt
```diff
@@ -333,8 +334,13 @@ class GroupPolicyProviderImpl @Activate constructor(
...
                     val gp = parseGroupPolicy(holdingIdentity)
                     if (gp is MGMGroupPolicy) {
-                        groupPolicies[holdingIdentity] = gp
-                        callBack(holdingIdentity, gp)
+                        if (gp.validate(holdingIdentity)) {
+                            groupPolicies[holdingIdentity] = gp
+                            callBack(holdingIdentity, gp)
+                        } else {
+                            groupPolicies.remove(holdingIdentity)
+                            return
```
nikinagy: I still wonder if this will cause issues in starting some of our membership components, because of the missing group policy.

components/membership/group-policy-impl/src/main/kotlin/net/corda/membership/impl/grouppolicy/GroupPolicyProviderImpl.kt
```diff
@@ -333,8 +334,13 @@ class GroupPolicyProviderImpl @Activate constructor(
                     val holdingIdentity = member.viewOwningMember.toCorda()
                     val gp = parseGroupPolicy(holdingIdentity)
                     if (gp is MGMGroupPolicy) {
-                        groupPolicies[holdingIdentity] = gp
-                        callBack(holdingIdentity, gp)
+                        if (gp.validate(holdingIdentity)) {
+                            groupPolicies[holdingIdentity] = gp
+                            callBack(holdingIdentity, gp)
+                        } else {
+                            groupPolicies.remove(holdingIdentity)
```
nikinagy: Why do we need this else? Wouldn't line 345 be enough?

components/membership/group-policy-impl/src/main/kotlin/net/corda/membership/impl/grouppolicy/GroupPolicyProviderImpl.kt
williamvigorr3: This should just invalidate the cache (or am I wrong). I was trying to do something similar to `startVirtualNodeHandle`.

williamvigorr3: Note that this might means that the callback doesn't get called but we already do something similar if the CpiInfoReadService is unavailable when the callback to the virtualNodeInfoReadService is called. See line `223` and `274`.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
williamvigorr3: I normally only use a function if we use the same error more than once but don't really mind.

components/membership/group-policy-impl/src/main/kotlin/net/corda/membership/impl/grouppolicy/GroupPolicyProviderImpl.kt
williamvigorr3: Line 345 only happens if `gp` is not an `MGMGroupPolicy`. I will try and tidy this up (too much nesting).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
nikinagy: Ah okay, makes sense, they just all seemed very similar early morning.

components/membership/group-policy-impl/src/main/kotlin/net/corda/membership/impl/grouppolicy/GroupPolicyProviderImpl.kt
nikinagy: But in this case we won't have group policy, right? What worries me is that at a lot of places eg sync service, registration proxy we make a call to get the group policy. I assume that call will fail if we don't have the policy cached and we are mgm, because we will try to query. I can see we are catching the BadGroupPolicyException at these places btw so probably not a showstopper for this fix, but something to keep in mind.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/common/MessageConverter.kt
```diff
@@ -235,7 +235,7 @@ class MessageConverter {
             val p2pParams = try {
                 groupPolicyProvider.getP2PParameters(source)
             } catch (except: BadGroupPolicyException) {
-                logger.warn("The group policy data is unavailable or cannot be parsed for our identity = $source Error ${except.message}" +
+                logger.warn("The group policy data is unavailable or cannot be parsed for our identity = $source. Error ${except.message}" +
```
nikinagy: its only nit, but I guess we should have : or use quotes for the exception message to make this nicer.

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/common/MessageConverterTest.kt
```diff
@@ -175,6 +177,27 @@ class MessageConverterTest {
         )
     }
 
+    @Test
+    fun `linkOutMessageFromAuthenticatedMessageAndKey returns null if BadGroupPolicy when look up the group policy provider for our id`() {
```
nikinagy: Can we rephrase the test names a bit better like saying "when BadGroupPolicy exception is being thrown on group policy look up"? Or something like that. Same for the other test cases.

components/membership/group-policy-impl/src/main/kotlin/net/corda/membership/impl/grouppolicy/GroupPolicyProviderImpl.kt
```diff
@@ -344,6 +344,18 @@ class GroupPolicyProviderImpl @Activate constructor(
...
 
+        private fun MGMGroupPolicy.validate(holdingIdentity: HoldingIdentity): Boolean {
+            return try {
+                this.p2pParameters
+                true
+            } catch (e: BadGroupPolicyException) {
+                logger.warn("Something went wrong while parsing the group policy for virtual node with holding identity" +
+                    " [$holdingIdentity]. Error: ${e.message} The group policy will be removed from the cache, to be parsed and cached on" +
+                    " the next read.")
+                false
```
dimosr, line 355: Is this what's going to happen here, e.g. given we use a compacted subscription, aren't we running the risk of discarding that record (offsets committed etc.) and never processing it again, thus ending up with a missing group policy from the in-memory map?

components/membership/group-policy-impl/src/main/kotlin/net/corda/membership/impl/grouppolicy/GroupPolicyProviderImpl.kt
williamvigorr3: I am planning on improving this in a follow up ticket. I need to give this some thought.

components/membership/group-policy-impl/src/main/kotlin/net/corda/membership/impl/grouppolicy/GroupPolicyProviderImpl.kt
dimosr: I just noticed that's already been discussed [here](https://github.com/corda/corda-runtime-os/pull/4854#discussion_r1357911381). Resolving this comment, so that we can have the conversation in a single place.

components/membership/group-policy-impl/src/main/kotlin/net/corda/membership/impl/grouppolicy/GroupPolicyProviderImpl.kt
dimosr: While taking a quick look at it, that's the main thing that stood out to me, i.e. the fact that we might process the record/commit offsets etc. without adding it in the in-memory map (and even worse, removing any that might have existed previously). So, if a record for the same MGM doesn't exist from that point forward in the topic, we'll end up with a missing group policy. 

If that's the case, I think we'll need to deal with this as part of this PR (not in a follow-up ticket) as it could cause problems that might be unrecoverable. 

I agree that if we catch the`BadGroupPolicyException` at the call-site and treat it properly (which seems to be what we are doing), we can avoid triggering the DB queries here since there's no easy way to replay/reprocess at this point.

components/membership/group-policy-impl/src/main/kotlin/net/corda/membership/impl/grouppolicy/GroupPolicyProviderImpl.kt
williamvigorr3: I think if that happens we will re-query the `virtualNodeInfoReadService` when we try and look it up by holding identity again (see lines 270 + 195). This in memory map is in a layer above the `virtualNodeInfoReadService`, this should not cause a record to be discarded (unless I missed something).

This is the area I would like to improve in a follow up.

dimosr: ah ok, it seems you are right. I thought we're taking everything from the compacted processor, hold it in-memory and just return whatever is there from the `getGroupPolicy`. But, looking at it now, it seems that if an entry is missing we don't return a `null`, but instead fall back to reading from the CPI again. So, it should be safe for now (albeit a bit confusing), we could clean it up later.

libs/membership/membership-impl/src/test/kotlin/net/corda/membership/lib/impl/grouppolicy/v1/MGMGroupPolicyImplTest.kt
```diff
@@ -152,26 +152,85 @@ class MGMGroupPolicyImplTest {
...
+                it.assertThat(p2pParameters.tlsTrustRoots.first()).isEqualTo(TEST_CERT)
+                it.assertThat(p2pParameters.tlsPki).isEqualTo(TlsPkiMode.STANDARD_EV3)
+                it.assertThat(p2pParameters.tlsVersion).isEqualTo(VERSION_1_2)
+                it.assertThat(p2pParameters.protocolMode).isEqualTo(AUTH)
+                it.assertThat(p2pParameters.mgmClientCertificateSubject).isEqualTo(null)
+
+                it.assertThat(groupPolicy.mgmInfo).isNull()
+                it.assertThat(groupPolicy.cipherSuite.entries).isEmpty()
+            }
+            verify(propertyQuery).invoke()
```
nikinagy, line 229: can't we have this just once?

williamvigorr3: I wanted to check the query is only called once. i.e. we don't do another query when we access the `protocolParameters` if we already accessed the `p2pParameters`.

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerTest.kt
```diff
@@ -2029,6 +2168,50 @@ class SessionManagerTest {
 
     }
 
+    @Test
+    fun `sessions are removed even if BadGroupPolicyException is thrown on group policy lookup`() {
```
nikinagy, line 2172:
```suggestion
    fun `sessions are removed if BadGroupPolicyException is thrown on group policy lookup`() {
```

nikinagy: discussed offline

nikinagy (approved):
LGTM

dimosr (approved):

PR #4857, opened 2023-10-12
CORE-16246 Allow topic configuration overrides

davidcurrie:
This PR provides the ability to override topic configuration, including replica and partition counts, on the `topic create` CLI commands and via the Helm chart.

The YAML file output by the `topic create preview` now includes `partitions` and `replicas` fields per topic. These default to the global parameters. These are then consumed by the `topic create connect` command so can be manually modified between preview and use.

The `topic create preview` and `topic create connect` commands also both take an `--overides`/`-o` option, which is the optional path to a file of the same format as the preview file. This file's contents are merged to override those generated/passed in.

The contents of this file can be provided as YAML under `bootstrap.kafka.overrides`. For example, the following Helm chart configuration can be used to override the default part count of 10 for the `avro.schema` topic to 5:

```yaml
bootstrap:
  kafka:
    partitions: 10
    overrides:
      topics:
        - name: "avro.schema"
          partitions: 5
```

### DIFF
--- a/charts/corda-lib/templates/_bootstrap.tpl
+++ b/charts/corda-lib/templates/_bootstrap.tpl
@@ -373,6 +373,9 @@ spec:
             {{- end }}
             '-r', '{{ .Values.bootstrap.kafka.replicas }}',
             '-p', '{{ .Values.bootstrap.kafka.partitions }}',
+            {{- if .Values.bootstrap.kafka.overrides }}
+            '-o', '/tmp/overrides.yaml',
+            {{- end }}
             'connect',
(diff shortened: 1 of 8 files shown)


simon-johnson-r3 (approved):

wzur-r3 (approved):
LGTM

vkolomeyko:
This CLI change will also need CLI baseline to be updated in E2E repository. Or else E2E tests will fail like so: https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda5-e2e-tests/detail/release%2F5.1/1054/tests

PR #4861, opened 2023-10-13
CORE-17747 remove attachments

blsemo:
Remove the implementation stubs for the attachments API deleted in this PR: https://github.com/corda/corda-api/pull/1297

### DIFF
--- a/libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/transaction/UtxoComponentGroup.kt
+++ b/libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/transaction/UtxoComponentGroup.kt
@@ -16,7 +16,7 @@ enum class UtxoComponentGroup {
     SIGNATORIES,
     OUTPUTS_INFO,
     COMMANDS_INFO,
-    DATA_ATTACHMENTS,
+    UNUSED, // can't remove to keep the Merkle hash stable
     INPUTS,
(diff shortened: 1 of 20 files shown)


lankydan (dismissed):

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/transaction/UtxoComponentGroup.kt
```diff
@@ -16,7 +16,7 @@ enum class UtxoComponentGroup {
     SIGNATORIES,
     OUTPUTS_INFO,
     COMMANDS_INFO,
-    DATA_ATTACHMENTS,
+    UNUSED, // can't remove to keep the Merkle hash stable
```
vlajos, line 19: Should we need to update utxoComponentGroupStructure below as well?
(That one is used only in transaction metadata to describe the structure.)

blsemo: No, that needs to stay as it is - we don't want to change anything in the binary representation that would upset the Merkle hash.

vlajos (dismissed):

rick-r3 (dismissed):

vlajos (approved):

rick-r3 (approved):

PR #4862, opened 2023-10-13
CORE-17604 Moving worker RPC endpoint paths to consts file

ben-millar:
The 4 workers which will communicate over RPC now have endpoints defined (E.g., `/persistence`, `/ledger`, etc.). Currently these are hardcoded strings, but this PR moves these to a consts file so they can be re-used between client and server implementations.


driessamyn (approved):
LGTM

emilybowe (approved):
LGTM

blsemo (approved):
LGTM

relyafi (approved):

PR #4878, opened 2023-10-15
* DO NOT MERGE * CORE-17768 Topology changes - Thread per consumer

mbrkic-r3:
This is alternative solution to https://github.com/corda/corda-runtime-os/pull/4931

Using new topology Kafka topics (flow.start, flow.session, flow.mapper.start, flow.mapper.session.in, flow.mapper.session.out).

Each consumer uses a long running thread to poll messages and store them in an internal buffer. Messages are polled until `maxPollRecords` are polled, then polling is paused until there is again buffer space available. When event mediator calls `poll` it immediately (no poll timeout) gets back messages from consumer's buffer.

Note that config values are hardcoded in [MessageBusConsumer.kt](https://github.com/corda/corda-runtime-os/pull/4878/files#diff-8ce13cdd8a558def0f533ec3ebf729fa4344c7acc93a857b6ea19331a698f387):
```
 private val DEFAULT_POLL_TIMEOUT = Duration.ofMillis(500)
 private const val DEFAULT_MAX_POLL_RECORDS = 500
 private const val BUFFER_QUEUE_TIMEOUT_MILLIS = 100L
```


mbrkic-r3:
E2E cluster tests are failing due to topic permissions. This should be fine after RPC stuff is integrated (since events should be sent to RPC endpoint instead of those topics).

JamesHR3 (commented):
This PR looks reasonable, but I think it needs testing against a cluster with multiple flow and flow mapper workers running. I have a feeling that the flow fiber cache in particular will cause some problems when we have multiple topics in place for a single flow, and we'll need to think about how that works.

PR #4881, opened 2023-10-16
CORE-17778 Use the query provider instead of raw SQL in findTransactionIdsAndStatuses

nkovacsx:
### Overview
Currently, we use a raw SQL in the findTransactionIdsAndStatuses function in UtxoRepositoryImpl. 
We should change this to use the QueryProvider instead. 

**This is not going to change functionality** just a nice to have thing to avoid confusion in the future.

### Test plan
Make sure that the current integration / e2e tests run.


jennyang-r3 (commented):
LGTM 
qq - do we have any tests for this?

blsemo (approved):
LGTM

blsemo:
> LGTM qq - do we have any tests for this?

Not really - as Nandor mentioned in the description, this doesn't change any functionality, it's just a code tidy up (encapsulating the raw Postgres SQL in a query provider so we could cleanly swap it out should we need to).
Therefore, not breaking any existing tests is sufficient coverage.

PR #4883, opened 2023-10-16
CORE-17483: Fix configuration change smoke test

josephzunigadaly:
Our configuration change test causes the lifecycle components within workers to restart. They receive the updated configuration and restart with their new configuration. We had pushed this test to the end of the run to avoid other tests being affected by the restart.

We have identified two issues with this test:
* The order we had defined to push the test to the end had not taken effect
* The code to test the cluster is in a good state did not handle exceptions that naturally happen when a client tries to connect to a server that is restarting; so the test was not waiting for the cluster to reach a good state

To address this we are making the following changes:
* Extract the configuration change test into its own test class
* Remove the method order annotation from the test, we want class order instead
* Increase the `@Order` value to `Int.MAX_VALUE` which is the maximum allowed
   * The default order value is `1073741823`, we previously used `999`, so we were not passing the default
* Change `startRpcFlow` to use `assertWithRetryIgnoringExceptions` which retries on exception


vkolomeyko (approved):
LGTM

PR #4885, opened 2023-10-16
CORE-16181 Implementing RPC client, routing external events through RPC

ben-millar:
This PR adds a new implementation of MessagingClient, RPCClient, handling synchronous calls to the Corda workers via RPC/HTTP as well as a HTTPRetryConfig and HTTPRetryExecutor to enable simple retrying logic for our HTTP requests on specified failures (typically, retryable 5XX failures).

Certain external events (Crypto, Uniqueness, Persistence, Verification) are now routed through this new RPC pathway rather than the Kafka/DB message bus.

Note: A single `HttpClient` is reused between all `RPCClient` instances, provided by a singleton `HttpClientFactory` as this is what's recommended to get the maximum benefits from shared state + connection pooling. The Java `HttpClient` will handle things like connection pooling, reusing connections to avoid socket limits, etc.


thiagoviana (dismissed):

components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
```diff
@@ -89,28 +98,36 @@ class FlowEventMediatorFactoryImpl @Activate constructor(
...
         .messageProcessor(messageProcessor)
-        .messageRouterFactory(createMessageRouterFactory())
+        .messageRouterFactory(createMessageRouterFactory(messagingConfig))
         .threads(configs.getConfig(ConfigKeys.FLOW_CONFIG).getInt(FlowConfig.PROCESSING_THREAD_POOL_SIZE))
         .threadName("flow-event-mediator")
         .stateManager(stateManager)
         .build()
 
-    private fun createMessageRouterFactory() = MessageRouterFactory { clientFinder ->
+    private fun createMessageRouterFactory(messagingConfig: SmartConfig) = MessageRouterFactory { clientFinder ->
```
JamesHR3, line 112: Does this function need `messagingConfig`?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -0,0 +1,109 @@
...
+        return MediatorMessage(deserializedResponse, mutableMapOf("statusCode" to response.statusCode()))
+    }
+
+
+    private fun deserializePayload(payload: ByteArray): Any {
+        return try {
+            deserializer.deserialize(payload)!!
+        } catch (e: Exception) {
+            val errorMsg = "Failed to deserialize payload of size ${payload.size} bytes due to: ${e.message}"
+            log.error(errorMsg)
```
JamesHR3: Think this should be `log.warn`, unless you can't carry on in the worker at all after this (which seems unlikely).

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -0,0 +1,109 @@
...
+            onSerializationError?.invoke(errorMsg.toByteArray())
+            throw(e)
```
JamesHR3: Nitpick, but do you need the brackets here?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -0,0 +1,109 @@
...
+            .build()
+) : MessagingClient {
+    private val deserializer = cordaAvroSerializerFactory.createAvroDeserializer({}, Any::class.java)
+
+    private companion object {
+        private val log: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+    }
+
+    override fun send(message: MediatorMessage<*>): MediatorMessage<*>? {
+        return try {
```
JamesHR3, line 38: Might not hurt to have some trace level logging around the success path.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private fun checkResponseStatus(statusCode: Int) {
+        when (statusCode) {
+            in 400..499 -> throw CordaHTTPClientErrorException(statusCode, "Server returned status code $statusCode.")
+            in 500..599 -> throw CordaHTTPServerErrorException(statusCode, "Server returned status code $statusCode.")
+        }
+    }
+
+    private fun handleExceptions(e: Exception) {
```
JamesHR3, line 90: These should also all be `log.warn`.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -0,0 +1,109 @@
...
+            is InterruptedException -> log.error("Operation was interrupted in RPCClient: ", e)
+            is IllegalArgumentException -> log.error("Invalid argument provided in RPCClient call: ", e)
+            is SecurityException -> log.error("Security violation detected in RPCClient: ", e)
+            is IllegalStateException -> log.error("Coroutine state error in RPCClient: ", e)
+            is CordaHTTPClientErrorException -> log.error("Client-side HTTP error in RPCClient: ", e)
+            is CordaHTTPServerErrorException -> log.error("Server-side HTTP error in RPCClient: ", e)
+            else -> log.error("Unhandled exception in RPCClient: ", e)
+        }
+
+        throw e;
```
JamesHR3: Don't think you need the semicolon here.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/factory/HttpClientFactory.kt
```diff
@@ -0,0 +1,14 @@
+package net.corda.messaging.mediator.factory
+
+import java.net.http.HttpClient
+import java.time.Duration
+
+object HttpClientFactory {
+    private val httpClient: HttpClient by lazy {
+        HttpClient.newBuilder()
+            .connectTimeout(Duration.ofSeconds(30))
```
JamesHR3: Could make this 30s a constant in here.

Potentially we could make this configurable, but I don't think we should do that now.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/factory/RPCClientFactory.kt
```diff
@@ -0,0 +1,23 @@
...
+import net.corda.messaging.api.mediator.factory.MessagingClientFactory
+import net.corda.messaging.mediator.RPCClient
+
+class RPCClientFactory(
+    private val id: String,
+    private val cordaSerializationFactory: CordaAvroSerializationFactory
+): MessagingClientFactory {
+
+    override fun create(config: MessagingClientConfig): MessagingClient {
+        val httpClient = HttpClientFactory.getClient()
```
JamesHR3: Does the `HttpClientFactory` need separating out here? Is it good enough to hold on to a `HttpClient` instance inside this factory and lazily create that the first time this factory's create method is called?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/HTTPRetryConfig.kt
```diff
@@ -0,0 +1,37 @@
...
+ * @property factor The multiplier used to increase the delay for each subsequent retry. Default is 2.0
+ * @property retryOn A set of exception classes that should trigger a retry when caught.
+ *                   If an exception not in this list is caught, it will be propagated immediately without retrying.
+ *                   Default is the generic [Exception] class, meaning all exceptions will trigger a retry.
+ */
+data class HTTPRetryConfig(
+    val times: Int = 3,
+    val initialDelay: Long = 100,
+    val factor: Double = 2.0,
+    val retryOn: Set<KClass<out Exception>> = setOf(Exception::class)
```
JamesHR3: Do you want `Class` here instead of `KClass`?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
ben-millar: I was working on some C++ projects over the weekend; the side effects are clear 😂

components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
ben-millar: Yeah, the `messagingConfig` contains the RPC worker service endpoints which were passed in through the `FlowWorker`/`CombinedWorker` startup args.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/factory/HttpClientFactory.kt
ben-millar: Good point. I think I'll leave this one until we make it configurable rather than throwing a temp constant in somewhere.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/factory/RPCClientFactory.kt
ben-millar: Good point!

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -0,0 +1,111 @@
...
+) : MessagingClient {
+    private val deserializer = cordaAvroSerializerFactory.createAvroDeserializer({}, Any::class.java)
+
+    private companion object {
+        private val log: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+    }
+
+    override fun send(message: MediatorMessage<*>): MediatorMessage<*>? {
+        return try {
+            log.trace("Received RPC external event send request for endpoint ${message.endpoint()}")
```
JamesHR3: I think you want the other version of `trace` with the `{}`. That saves you evaluating this string every time.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -0,0 +1,111 @@
...
+        return MediatorMessage(deserializedResponse, mutableMapOf("statusCode" to response.statusCode()))
+    }
+
+
+    private fun deserializePayload(payload: ByteArray): Any {
+        return try {
+            deserializer.deserialize(payload)!!
+        } catch (e: Exception) {
+            val errorMsg = "Failed to deserialize payload of size ${payload.size} bytes due to: ${e.message}"
+            log.warn(errorMsg)
```
JamesHR3: Might be worth putting `e` in as the cause here too to give a more detailed stack trace.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/HTTPRetryConfig.kt
```diff
@@ -0,0 +1,35 @@
...
+        private var retryOn: Set<Class<out Exception>> = setOf(Exception::class.java)
+
+        fun times(times: Int) = apply { this.times = times }
+        fun initialDelay(delay: Long) = apply { this.initialDelay = delay }
+        fun factor(factor: Double) = apply { this.factor = factor }
+        fun retryOn(vararg exceptions: Class<out Exception>) = apply { this.retryOn = exceptions.toSet() }
+
+        fun build(): HTTPRetryConfig {
+            return HTTPRetryConfig(times, initialDelay, factor, retryOn)
+
```
JamesHR3: Minor - there's an extra space here :).

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/HTTPRetryExecutor.kt
```diff
@@ -0,0 +1,42 @@
...
+
+class HTTPRetryExecutor {
+    companion object {
+        private val log: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+
+        fun <T> withConfig(config: HTTPRetryConfig, block: () -> T): T {
+            var currentDelay = config.initialDelay
+            for (i in 0 until config.times - 1) {
+                try {
+                    log.trace("HTTPRetryExecutor making attempt #${i + 1}.")
```
JamesHR3: Similarly worth updating to use the other form of `trace` here.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/HTTPRetryExecutor.kt
```diff
@@ -0,0 +1,43 @@
...
+            var currentDelay = config.initialDelay
+            for (i in 0 until config.times - 1) {
+                try {
+                    log.trace { "HTTPRetryExecutor making attempt #${i + 1}." }
+                    val result = block()
+                    log.trace { "Operation successful after #${i + 1} attempt/s." }
+                    return result
+                } catch (e: Exception) {
+                    if (config.retryOn.none { it.isInstance(e) }) {
+                        log.warn("HTTPRetryExecutor caught a non-retryable exception")
```
JamesHR3: Probably want `e` as the cause here as well. You could additionally put `e.message` in the log string.

JamesHR3 (commented):
Nearly there!

JamesHR3 (approved):

thiagoviana (approved):

PR #4886, opened 2023-10-16
CORE-17627: Add metadata to flow mapper states for mapper status

JamesHR3:
### Problem description

The new flow mapper cleanup mechanism involving the scheduled task trigger and the state manager requires that the state metadata is populated with the current flow mapper status. Currently this is not written, meaning no states will be picked up when attempting mapper cleanup.

### Solution

Add the mapper status as a metadata field to the output state metadata.

The value used is the enum value as a string.


emilybowe (dismissed):
LGTM

thiagoviana (approved):

PR #4901, opened 2023-10-17
CORE-17779 add metric for a HTTP RPC Client request

LWogan:
Add a metric to record some information about how long it takes for an RPCClient to receive a response, as well as a distribution of response message size


Omar-awad (dismissed):
LTGM

thiagoviana (dismissed):

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -73,8 +79,43 @@ class RPCClient(
...
+        }
+    }
+
+    private fun buildMetricForResponse(
+        startTime: Long,
+        operationStatus: String,
+        request: HttpRequest,
+        response: HttpResponse<ByteArray>? = null
+    ) {
+        val endTime = Instant.now().toEpochMilli()
```
sprintengine: Think we might have been using `System.nanoTime()` isntead e.g. https://github.com/jujoramos/corda-runtime-os/blob/e86a998e6ca8799066ad10a930b6e440414d99a7/components/uniqueness/uniqueness-checker-client-service-impl/src/main/kotlin/net/corda/uniqueness/client/impl/LedgerUniquenessCheckerClientServiceImpl.kt#L71

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -73,8 +79,43 @@ class RPCClient(
...
+        CordaMetrics.Metric.Messaging.HTTPRPCResponseTime.builder()
+            .withTag(CordaMetrics.Tag.OperationStatus, operationStatus)
```
JamesHR3, line 105: Is it worth having both the operation status and the response code as tags? In theory that would increase the cardinality of our metrics if we ever decided to treat a 2xx code as an error (not sure why we'd do that). Feels like we're better off doing one or the other?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -73,8 +79,43 @@ class RPCClient(
...
+            .withTag(CordaMetrics.Tag.HttpMethod, request.method())
+            .withTag(CordaMetrics.Tag.HttpRequestUri, request.uri().toString())
+            .withTag(CordaMetrics.Tag.HttpResponseCode, response?.statusCode().toString())
+            .build()
+            .record(endTime - startTime, TimeUnit.MILLISECONDS)
+
+        if (response != null) {
+            CordaMetrics.Metric.Messaging.HTTPRPCResponseSize.builder()
```
JamesHR3, line 112: I'd be tempted to leave this one out entirely. If it is kept, I think we only record this in successful cases anyway so it might be worth taking off things like success/failure and the response code.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -73,8 +79,43 @@ class RPCClient(
...
+        startTime: Long,
+        operationStatus: String,
+        request: HttpRequest,
+        response: HttpResponse<ByteArray>? = null
+    ) {
+        val endTime = Instant.now().toEpochMilli()
+        CordaMetrics.Metric.Messaging.HTTPRPCResponseTime.builder()
+            .withTag(CordaMetrics.Tag.OperationStatus, operationStatus)
+            .withTag(CordaMetrics.Tag.HttpMethod, request.method())
+            .withTag(CordaMetrics.Tag.HttpRequestUri, request.uri().toString())
```
JamesHR3: Are any of these URIs templated? We'll get a cardinality spike if we record a separate time series for each possible templated value in the URI.

libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
```diff
@@ -692,6 +703,16 @@ object CordaMetrics {
...
+        /**
+         * The URI that a HTTP request was sent to
+         */
+        HttpRequestUri("http.request.uri"),
+
+        /**
+         * Response code received for a HTTP response
+         */
+        HttpResponseCode("http.response.code"),
+
```
JamesHR3, line 715: Do we not already have similar tag names for use in the REST processor that we could reuse here?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -73,8 +79,43 @@ class RPCClient(
...
+    private fun buildMetricForResponse(
+        startTime: Long,
+        operationStatus: String,
+        request: HttpRequest,
+        response: HttpResponse<ByteArray>? = null
+    ) {
+        val endTime = Instant.now().toEpochMilli()
+        CordaMetrics.Metric.Messaging.HTTPRPCResponseTime.builder()
+            .withTag(CordaMetrics.Tag.OperationStatus, operationStatus)
+            .withTag(CordaMetrics.Tag.HttpMethod, request.method())
```
sprintengine: Wonder will HttpMethod be useful here or does it just increase metric cardinality, maybe the value for HttpRequestUri could combine `request.method() + request.uri().toString()`. Wouldn't hold up the PR for it

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
LWogan: the operation status of failed indicates an exception was thrown so its possible no response was received

libs/metrics/src/main/kotlin/net/corda/metrics/CordaMetrics.kt
LWogan: i dont see any

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
LWogan: for now the paths are very simple with no unique identifiers. In the format of ${server-url}/${WorkerPath}.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
LWogan: ive left it in and taken off the tags suggested for removal

thiagoviana (dismissed):

Omar-awad (approved):
LGTM

PR #4904, opened 2023-10-17
CORE-17863: Close db connection pool after getOwnedKeyRecord

yift-r3:
Close the repo (and hence the connection) when the `getOwnedKeyRecord` is called.

Tested by running the scalability large network test and observing that the maximum number of concurrent open connections per virtual node was reduced from ~20 to 2 (and those two are closed after use).


dickon (approved):
LGTM

dimosr (approved):

PR #4907, opened 2023-10-17
CORE-17626: Add integration test for flow mapper cleanup

JamesHR3:
### Problem description

The mapper cleanup code did not have an integration test verifying that mapper states are removed from state storage after some time has passed.

### Solution

Add an integration test.

A small bug was found in setting the key for the cleanup event, as the UUID here should have been converted to a string.

This has also been verified manually against the combined worker.


emilybowe (approved):
LGTM

PR #4908, opened 2023-10-17
CORE-17934 Reduce log noise

lankydan:
Move a lot of `info` logs down to `debug`.

Reword some of the logs.

Leave comments on other logs that could do with futher looking into.

### DIFF
--- a/components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/HSMRegistrationBusProcessor.kt
+++ b/components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/HSMRegistrationBusProcessor.kt
@@ -36,7 +36,7 @@ class HSMRegistrationBusProcessor(
 
     override fun onNext(request: HSMRegistrationRequest, respFuture: CompletableFuture<HSMRegistrationResponse>) {
         try {
-            logger.info("Handling {} for tenant {}", request.request::class.java.name, request.context.tenantId)
+            logger.debug {"Handling ${request.request::class.java.name} for tenant ${request.context.tenantId}" }
             val response = executor.executeWithRetry {
                 handleRequest(request.request, request.context)
             }
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
@@ -12,15 +12,13 @@ import net.corda.data.flow.state.external.ExternalEventStateType
 import net.corda.flow.external.events.factory.ExternalEventRecord
 import net.corda.flow.pipeline.exceptions.FlowFatalException
 import net.corda.messaging.api.records.Record
-import net.corda.utilities.FLOW_TRACING_MARKER
 import net.corda.utilities.debug
+import net.corda.utilities.trace
 import org.osgi.service.component.annotations.Activate
 import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
-import org.slf4j.Marker
-import org.slf4j.MarkerFactory
 import java.nio.ByteBuffer
 import java.time.Duration
 import java.time.Instant
@@ -35,7 +33,6 @@ class ExternalEventManagerImpl(
 
     private companion object {
         val log: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
-        val flowTraceMarker: Marker = MarkerFactory.getMarker(FLOW_TRACING_MARKER)
     }
 
     @Activate
@@ -81,10 +78,7 @@ class ExternalEventManagerImpl(
         externalEventResponse: ExternalEventResponse
     ): ExternalEventState {
         val requestId = externalEventResponse.requestId
-        // This manual tracing has been superseded by Owen/Joseph's work (talk with Juan to check, CORE-12440).
-        if (log.isTraceEnabled) {
-            log.trace(flowTraceMarker, "Processing response for external event with id '{}'", requestId)
-        }
+        log.trace { "Processing response for external event with id '$requestId'" }
 
         if (requestId == externalEventState.requestId) {
             log.debug { "External event response with id $requestId matched last sent request" }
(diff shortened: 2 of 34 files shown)


dickon (dismissed):
LGTM

needs a JIRA ticket

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
```diff
@@ -198,12 +201,14 @@ class ExternalEventManagerImpl(
...
         val eventToSend = externalEventState.eventToSend
         eventToSend.timestamp = instant
-        log.info(
-            flowTraceMarker,
-            "Dispatching external event with id '{}' to '{}'",
-            externalEventState.requestId,
-            eventToSend.topic
-        )
+        if (log.isDebugEnabled) {
+            log.debug(
```
vkolomeyko: Should we use lambda version, i.e.: `log.debug {...}` ?

libs/rest/rest-server-impl/src/main/kotlin/net/corda/rest/server/impl/context/ContextUtils.kt
```diff
@@ -102,8 +102,7 @@ internal object ContextUtils {
             val ctxMethod = ctx.method()
             withMDC(restContext()?.principal ?: "<anonymous>", ctxMethod, ctx.path()) {
                 val methodLogger = ctxMethod.loggerFor()
-                methodLogger.info("Servicing $ctxMethod request to '${ctx.path()}'")
```
vkolomeyko, line 105: I would rather leave this `info` line in. It is invaluable when we are trying to understand whether we actually hit REST worker with a particular request. And from my experience working with logs, it does not produce too much volume anyway.

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
lankydan: Not on this one because of the `flowTraceMarker` but I need to catchup with Juan to see if we need the marker still anyway.

components/membership/membership-persistence-client-impl/src/main/kotlin/net/corda/membership/impl/persistence/client/MembershipPersistenceClientImpl.kt
```diff
@@ -167,7 +168,11 @@ class MembershipPersistenceClientImpl(
         viewOwningIdentity: HoldingIdentity,
         groupParameters: InternalGroupParameters,
     ): MembershipPersistenceOperation<InternalGroupParameters> {
-        logger.info("Persisting group parameters.")
+        // More context is needed.
+        // We could then consider putting this back to INFO based on the context it contains
+        // Generally I think whatever triggered the group parameters to be persisted should log this at info,
+        // this does seem an important thing to log generally.
+        logger.debug { "Persisting group parameters." }
```
dimosr: Yes, seems like it would be potentially a bit more useful if that was at the call-site. And maybe containing some more information.

Seems like this is invoked from the static registration service and the sync service (for dynamic networks). And we could potentially add the network (or the holding identity) and the epoch. That can then be matched against the DB. 

P.S. I think that can be a bit controversial, as someone could make the argument that it's better logging that at this level, so that it's not potentially forgotten at the higher level. And then correlated with any logs at the higher level. No strong feelings on this, I would be happy with both.

components/membership/membership-persistence-client-impl/src/main/kotlin/net/corda/membership/impl/persistence/client/MembershipPersistenceOperationImpl.kt
```diff
@@ -42,7 +43,11 @@ internal class MembershipPersistenceOperationImpl<T>(
             return Either.Right(MembershipPersistenceResult.Failure(failureReason))
         }
         val requestId = request.context.requestId
-        logger.info("Sending membership persistence RPC request ID: $requestId.")
+        // Should really have logging in the calls that filter down to this that specify what this "persistence request" is from a
+        // customers perspective.
+        // The log lines above this would be info for important requests, like registering a member and debug for querying mgm info.
+        // We could then remove this log line or delegate it to trace.
+        logger.debug { "Sending membership persistence RPC request ID: $requestId." }
```
dimosr, line 46: I think a similar argument could be made as above here on whether it's better to do it at the shared function or the invocation site. You're making a good point that maybe not all requests should be logged, which could be an argument for moving it at the invocation site. 

Someone with more experience on issues we have faced around this area from the @corda/corda-platform-network-team might be able to provide more insight (since I know there have been many investigations through time around the use of DB that this might have contributed to).

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/QueryRegistrationRequestsHandler.kt
```diff
@@ -18,7 +19,8 @@ internal class QueryRegistrationRequestsHandler(persistenceHandlerServices: Pers
         context: MembershipRequestContext,
         request: QueryRegistrationRequests,
     ): RegistrationRequestsQueryResponse {
-        logger.info("Retrieving registration requests.")
+        // Is there more context that would be useful here?
```
dimosr: yeah, worth adding the parameters of the query (e.g. name, statuses, limit) I suppose.

dimosr (commented):
I checked the network-related components, left a few comments but otherwise looks reasonable

components/membership/membership-persistence-client-impl/src/main/kotlin/net/corda/membership/impl/persistence/client/MembershipPersistenceOperationImpl.kt
lankydan: @dimosr @charlieR3 I suggest we raise a ticket to review the logs of the MGM code, hopefully this is something that could be done for 5.2. I don't have the context to make meaningful changes to the MGM logs myself.

dimosr: sure, do you want to raise a ticket and add some details on the areas you want us to put more focus on it (so that the review can be a bit more targeted)? And we'll pick it up.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/HSMRegistrationBusProcessor.kt
```diff
@@ -36,7 +36,7 @@ class HSMRegistrationBusProcessor(
 
     override fun onNext(request: HSMRegistrationRequest, respFuture: CompletableFuture<HSMRegistrationResponse>) {
         try {
-            logger.info("Handling {} for tenant {}", request.request::class.java.name, request.context.tenantId)
+            logger.debug {"Handling ${request.request::class.java.name} for tenant ${request.context.tenantId}" }
```
simon-johnson-r3, line 39: there's a missing space here, but I wouldn't bother fixing unless you're pushing more changes already

lankydan: 😱

driessamyn (approved):
LGTM

PR #4911, opened 2023-10-17
CORE-17013 Adding repartition handling to MessageBusConsumer

ben-millar:
#### Context:
The `MultiSourceEventMediatorImpl` is calling `poll()` on the `MessageBusConsumer`, processing the polled records, and then calling `syncCommitOffsets()`. If a rebalance occurs while we're protesting the polled records, we'll see errors when trying to commit offsets to any partitions revoked in the rebalance.

#### Fix:
Added a `CordaConsumerRebalanceListener` to the `MessageBusConsumer` which tracks any partitions that have been revoked. In `SyncCommitOffsets()` we now only commit offsets for those partitions which have not previously been revoked.

#### To-do:
Update tests

### DIFF
--- a/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImpl.kt
+++ b/libs/messaging/db-message-bus-impl/src/main/kotlin/net/corda/messagebus/db/consumer/DBCordaConsumerImpl.kt
@@ -1,6 +1,11 @@
 package net.corda.messagebus.db.consumer
 
+import java.time.Duration
+import java.time.Instant
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.concurrent.withLock
 import net.corda.avro.serialization.CordaAvroDeserializer
+import net.corda.messagebus.api.CordaOffsetAndMetadata
 import net.corda.messagebus.api.CordaTopicPartition
 import net.corda.messagebus.api.consumer.CordaConsumer
 import net.corda.messagebus.api.consumer.CordaConsumerRebalanceListener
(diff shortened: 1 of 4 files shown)


JamesHR3 (commented):
I'm not sure this is necessary because the rebalance is only triggered when `poll` is called and the callback happens in the same thread as the one that called `poll`. See https://kafka.apache.org/24/javadoc/index.html?org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html. That says that `onPartitionsRevoked` will only be called when `poll` is called and is guaranteed to be called before `onPartitionsAssigned` in the consumer the partition has been assigned to. That means our offsets are already correct when this is invoked and we won't attempt to commit the offset on the next loop anyway (`consumer.assignment` won't give us it). I think the same is true for `onPartitionLost` - by the time this is invoked, we'll have already committed our most up-to-date offset position for the lost partition.

I think this becomes relevant when we're not committing offset positions on every poll loop - then we'll want to ensure that we commit whatever offsets we have in the `onPartitionsRevoked`  callback, and don't do it in the `onPartitionsLost` callback (as those partitions may be assigned to another consumer).

ben-millar:
> I'm not sure this is necessary because the rebalance is only triggered when `poll` is called and the callback happens in the same thread as the one that called `poll`. See https://kafka.apache.org/24/javadoc/index.html?org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html. That says that `onPartitionsRevoked` will only be called when `poll` is called and is guaranteed to be called before `onPartitionsAssigned` in the consumer the partition has been assigned to. That means our offsets are already correct when this is invoked and we won't attempt to commit the offset on the next loop anyway (`consumer.assignment` won't give us it). I think the same is true for `onPartitionLost` - by the time this is invoked, we'll have already committed our most up-to-date offset position for the lost partition.
> 
> I think this becomes relevant when we're not committing offset positions on every poll loop - then we'll want to ensure that we commit whatever offsets we have in the `onPartitionsRevoked` callback, and don't do it in the `onPartitionsLost` callback (as those partitions may be assigned to another consumer).

Chatted some more about this offline. If it is the case that partitions can't be revoked until we next call poll(), then there's no issue here and I can descope.

PR #4915, opened 2023-10-18
CORE-17882: Add a route for flow events

JamesHR3:
### Problem description

The flow worker may post input events back to itself in transient error scenarios. This currently causes errors as there is no route defined for flow events in the flow worker event mediator.

### Solution

Add a route for flow events.

### Additional changes

`NewConfigurationReceived` is an unused lifecycle event, which can be removed.


driessamyn (approved):
LGTM

PR #4920, opened 2023-10-18
CORE-17883 Moving status code check to HTTPRetryExecutor, retrying on 404

ben-millar:
This change modifies the `RPCClient` so that it will retry on 4XX and 5XX errors from HTTP calls; it also segments any potential exceptions into `intermittent` or `fatal`, where `intermittent` exceptions can be retried at the mediator level without bringing it down.

We may need future work to communicate these errors back to the flow level and fail the flow.


ben-millar:
Merging directly into release instead

PR #4924, opened 2023-10-18
CORE-17886 - set bundle at startup

driessamyn:
Javalin threw an exception when restarting and this is believed to be an OSGi classloader issue. 
This change holds on to the bundle at start-up rather than trying to re-load it when the lifecycle coordinator restarts the webserver.


driessamyn:
superseded by https://github.com/corda/corda-runtime-os/pull/5020

PR #4925, opened 2023-10-18
CORE-17883 RPCClient stability improvements

ben-millar:
Currently, if the `RPCClient` triggers an exception, that exception will bring down the `MultiSourceEventMediator`.

This change modifies the `RPCClient` so that it will retry on 4XX and 5XX errors from HTTP calls; it also segments any potential exceptions into `intermittent` or `fatal`, where `intermittent` exceptions can be retried at the mediator level without bringing it down.

We may need future work to communicate these errors back to the flow level and fail the flow.

### DIFF
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/HTTPRetryExecutor.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/HTTPRetryExecutor.kt
@@ -1,6 +1,10 @@
 package net.corda.messaging.utils
 
+import java.net.http.HttpResponse
+import net.corda.messaging.api.exception.CordaHTTPClientErrorException
+import net.corda.messaging.api.exception.CordaHTTPServerErrorException
 import net.corda.utilities.trace
+import net.corda.v5.base.exceptions.CordaRuntimeException
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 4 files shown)


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/HTTPRetryExecutor.kt
```diff
@@ -1,43 +1,66 @@
...
-                    log.trace { "Attempt #${i + 1} failed due to ${e.message}. Retrying in $currentDelay ms..." }
-                    Thread.sleep(currentDelay)
-                    currentDelay = (currentDelay * config.factor).toLong()
-                }
+            for (i in 0 until config.times) {
+                val result = tryAttempt(i, config, block)
+                if (result != null) return result
+
+                log.trace { "Attempt #${i + 1} failed. Retrying in $currentDelay ms..." }
+                Thread.sleep(currentDelay)
```
driessamyn, line 22: Not for now, but we may want to review the design of this later on. 
I think the Thread.sleep here is blocking, which may not be desirable. Ideally, I think, the send would be asynchronously and we can `delay` here rather than block the thread.
But this may need a bit more analysis and discussion.

LWogan (approved):
LGTM

thiagoviana (approved):

PR #4926, opened 2023-10-18
CORE-17881: WIP Individual State Manager Per Worker Type

jujoramos:


charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -356,6 +324,115 @@ spec:
...
+{{- if .Values.bootstrap.db.enabled }}
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "corda.fullname" . }}-setup-{{ $workerName }}-state-manager-db
+  labels:
+    {{- include "corda.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-install
```
davidcurrie, line 345: These pre-install jobs run serially, so it's worth considering whether having these as multiple containers running in parallel on a single job would be preferable.

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -630,165 +630,119 @@ data:
...
-      name: {{ include "corda.bootstrapStateManagerDbDefaultSecretName" . | quote }}
-      key: "username"
-      {{- else if .Values.stateManager.db.username.valueFrom.secretKeyRef.name }}
-      name: {{ .Values.stateManager.db.username.valueFrom.secretKeyRef.name | quote }}
-      key: {{ required "Must specify stateManager.db.username.valueFrom.secretKeyRef.key" .Values.stateManager.db.username.valueFrom.secretKeyRef.key | quote }}
-      {{- else }}
-      name: {{ include "corda.stateManagerDbDefaultSecretName" . | quote }}
-      key: "username"
-      {{- end }}
+      {{- if and ($worker.stateManager.db.host) ($bootConfig.username.valueFrom.secretKeyRef.name) }}
```
davidcurrie, line 732: Should it fall through to use the state manager credentials for the worker if there are none specified on bootstrap?

charts/corda-lib/templates/_helpers.tpl
```diff
@@ -630,165 +630,119 @@ data:
...
-State Manager DB port
+State Manager Database port
 */}}
 {{- define "corda.stateManagerDbPort" -}}
-{{- if $.Values.stateManager.db.host }}
-{{- .Values.stateManager.db.port | default "5432" -}}
+{{- $ := index . 0 }}
+{{- $worker := index . 1 }}
+{{- if $worker.stateManager.db.host -}}
+{{- $worker.stateManager.db.port | default "5432" -}}
```
davidcurrie, line 654: Not introduced here, but my preference is to not specify default values within templates as those default values are not visible to the user. There should be a default in `values.yaml` and the schema should enforce that it is not removed.

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -356,6 +324,115 @@ spec:
...
+
+              echo 'Generating State Manager DB specification for {{ $workerName }}-worker...'
+              STATE_MANAGER_JDBC_URL="{{- include "corda.stateManagerJdbcUrl" ( list . $worker ) -}}"
+              mkdir /tmp/stateManager
+              java -Dpf4j.pluginsDir=/opt/override/plugins -Dlog4j2.debug=false -jar /opt/override/cli.jar database spec \
+                -s "statemanager" -g "statemanager:state_manager" \
+                -u "${STATE_MANAGER_PGUSER}" -p "${STATE_MANAGER_PGPASSWORD}" \
+                --jdbc-url "${STATE_MANAGER_JDBC_URL}" \
+                -c -l /tmp/stateManager
+              echo 'Generating State Manager DB specification for {{ $workerName }}-worker... Done'
```
davidcurrie, line 382: `$workerName` already has `-worker` on the end of it...
```suggestion
              echo 'Generating State Manager DB specification for {{ $workerName }}... Done'
```

charts/corda-lib/templates/_worker.tpl
```diff
@@ -243,7 +243,7 @@ spec:
         {{- include "corda.clusterDbEnv" $ | nindent 10 }}
         {{- end }}
         {{- if $optionalArgs.stateManagerDbAccess }}
```
davidcurrie, line 245: You no longer need `stateManagerDbAccess` as you can just use the presence of `stateManager` in the worker values

charts/corda/values.yaml
```diff
@@ -408,30 +373,56 @@ bootstrap:
...
-          secretKeyRef:
-            # -- the password secret name
-            name: ""
-            # -- the password secret key
-            key: ""
+      - worker: flow
+        # the username configuration
+        username:
+          # -- the username, defaults to this value
+          value: "state_manager_user"
```
davidcurrie, line 380: I'd expect these to have default values in the worker configuration, not under bootstrap. And should they have different user names by default (if they're likely to default to being in the same cluster DB)?

sprintengine: I think it should match whats in
- https://github.com/corda/corda-shared-build-pipeline-steps/pull/1267/files#diff-1e4dc523d753a21fccff4f0f4a60c7630bdd81843761d077ab99f8e597856478
- https://github.com/corda/corda-shared-build-pipeline-steps/pull/1267/files#diff-89c779ab5e5f6d83d2c0dcdbe03d0d5ebb1f91fdb30a605e22906023908b36e0

PR #4927, opened 2023-10-18
CORE-17885: Use explicit conversion of session timeout to long

JamesHR3:
### Problem description

When the metadata associated with the checkpoint goes through a cycle of being written to and read from the state manager, it is converted to and from JSON. As part of this, the fact that the original type of the session timeout metadata is a long is lost. Casting from int to long is not allowed, and so this results in an error when the timeout processor attempts to use this metadata.

### Solution

Cast instead to `Number` and then do an explicit conversion to long.

As `Number` is a supertype of all numeric values, this cast should always succeed. It's then possible to do an explicit conversion via the `toLong` method.

### Additional changes

A previous PR introduced a routing change to allow flow events to be outputted from the flow engine for transient errors. The unit test for the mediator factory has been enhanced to assert that this route (and others) is correct.


thiagoviana (approved):

emilybowe (approved):
LGTM

PR #4935, opened 2023-10-19
CORE-17733 - SM to update via JDBC in batch

conal-smith:
**Problem Description**

StateManager update() executes one individual `entityManager... .executeUpdate` per state.

Instead, we can batch this with a query that executes only one statement to the database engine.

**Solution**

Use JDBC and `PreparedStatement` and `execute()`.
Preserve transactionality of the funciton and roll back all when one fails with an exception.
Return those updates that were missed due to "optimistic locking version mismatch".


libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/StateRepository.kt
```diff
@@ -33,21 +34,21 @@ interface StateRepository {
...
+     * @param connection Used to interact with the state manager persistence context.
      * @param states Collection of states to be updated.
      * @return Collection of keys for states that could not be updated due to optimistic locking check failure.
      */
-    fun update(entityManager: EntityManager, states: Collection<StateEntity>): Collection<String>
+    fun update(connection: Connection, states: Collection<StateEntity>): Collection<String>
 
     /**
      * Delete states with the given keys from the persistence context.
      * Transaction should be controlled by the caller.
```
sprintengine: this may no longer be true for update/delete, mgith move transaction call outside of here

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -230,29 +230,30 @@ class StateManagerIntegrationTest {
         }
     }
 
-    @ValueSource(ints = [1, 10])
+    @ValueSource(ints = [1, 5, 10, 20, 50, 100, 200, 500])
```
jujoramos: Did you intend to commit these changes?, not sure is worth keeping it as we're not profiling the test itself.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/PostgresQueryProvider.kt
```diff
@@ -15,8 +15,13 @@ class PostgresQueryProvider : AbstractQueryProvider() {
...
             UPDATE $STATE_MANAGER_TABLE SET
-            key = :$KEY_PARAMETER_NAME, value = :$VALUE_PARAMETER_NAME, version = version + 1, metadata = CAST(:$METADATA_PARAMETER_NAME as JSONB), modified_time = CURRENT_TIMESTAMP AT TIME ZONE 'UTC'
-            WHERE key = :$KEY_PARAMETER_NAME AND version = :$VERSION_PARAMETER_NAME
+            key = ?, value = ?, version = ? + 1, metadata = CAST(? as JSONB), modified_time = CURRENT_TIMESTAMP AT TIME ZONE 'UTC'
+            WHERE key = ? AND version = ?
+        """.trimIndent()
+
+    override val deleteStatesByKey: String
+        get() = """
+            DELETE FROM $STATE_MANAGER_TABLE s WHERE s.key = ? AND s.version = ?
```
jujoramos: This one doesn't seem to be postgres specific, so it can remain in `AbstractQueryProvider`, right?.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -34,42 +30,53 @@ class StateRepositoryImpl(private val queryProvider: QueryProvider) : StateRepos
...
-                .executeUpdate().also {
-                    if (it == 0) {
-                        failedKeys.add(state.key)
-                    }
-                }
+    private fun getFailedKeysFromResults(results: IntArray?, map: List<String>): List<String> {
+        val failed = mutableListOf<String>()
+        results?.mapIndexed { idx, result ->
+            if (result == 0)
+                failed.add(map[idx])
```
jujoramos: Just for clarity purpose: would it be worth adding the `RETURNING` statement to the SQL queries or something similar instead of doing this?.

owenstanford (commented):
Ideally we would remove the entity manager entirely, maybe not for this PR, but at least have a ticket to do it?

jujoramos:
> Ideally we would remove the entity manager entirely, maybe not for this PR, but at least have a ticket to do it?

Yep, we do: CORE-17025

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
sprintengine: I don't think this is trivial with the `preparedStatement` approach, for one - `executeBatch` doesn't return us the result of the query just how many rows it updated. The approach here is pretty much equivalent to the previous behaviour which determined failed updates by the returned number of updated rows on `executeUpdate`.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
sprintengine: Probably I'll revert, might be beneficial to test a few extra batch sizes though. Sum total time to run all tests in this file is only around 3s anyway.

libs/db/db-core/src/main/kotlin/net/corda/db/core/utils/ConnectionUtils.kt
```diff
@@ -0,0 +1,27 @@
...
+ * Commits transaction if no exceptions were thrown by [block]. Otherwise rolls back the transaction.
+ *
+ * Finally closes the connection after committing or rolling back the changes.
+ *
+ * @param block The code to execute before committing the transaction.
+ * @param R The type returned by [block].
+ *
+ * @return The result of executing [block].
+ */
+inline fun <R> Connection.transaction(block: (Connection) -> R): R {
```
driessamyn, line 17: Can you add a unit test for this? (see `transactionExecutor`)

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -15,6 +17,7 @@ import javax.persistence.EntityManagerFactory
 class StateManagerImpl(
     private val stateRepository: StateRepository,
     private val entityManagerFactory: EntityManagerFactory,
+    private val dataSource: CloseableDataSource,
```
driessamyn, line 20: As this is updated to use plain SQL, can't we remove the JPA stuff completely?
Creating the EMF is going to be quite wasteful.

jujoramos: We certainly should, but I think that must be addressed in a different PR to keep things separated.

jujoramos (dismissed):

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/PostgresQueryProvider.kt
```diff
@@ -15,8 +15,8 @@ class PostgresQueryProvider : AbstractQueryProvider() {
     override val updateState: String
         get() = """
             UPDATE $STATE_MANAGER_TABLE SET
-            key = :$KEY_PARAMETER_NAME, value = :$VALUE_PARAMETER_NAME, version = version + 1, metadata = CAST(:$METADATA_PARAMETER_NAME as JSONB), modified_time = CURRENT_TIMESTAMP AT TIME ZONE 'UTC'
-            WHERE key = :$KEY_PARAMETER_NAME AND version = :$VERSION_PARAMETER_NAME
+            key = ?, value = ?, version = ? + 1, metadata = CAST(? as JSONB), modified_time = CURRENT_TIMESTAMP AT TIME ZONE 'UTC'
```
vkolomeyko:
```suggestion
            key = ?, value = ?, version = version + 1, metadata = CAST(? as JSONB), modified_time = CURRENT_TIMESTAMP AT TIME ZONE 'UTC'
```

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/PreparedStatementHelper.kt
```diff
@@ -0,0 +1,51 @@
...
+     *
+     * If a SQL command includes an optimistic locking version check in the query then 0 will be returned on a
+     * optimistic version lock mismatch.
+     *
+     * @param batchResults results from `PreparedStatement.executeBatch()`
+     * @param commandKeys list of keys for each command, in the order they were added to `PreparedStatement`
+     * @return a list of commands IDs that had failures or affected zero rows
+     * @throws PreparedStatementHelperException if number of results does not match number of commands
+     */
+    fun extractFailedKeysFromBatchResults(batchResults: IntArray, commandKeys: List<String>): List<String> {
```
vkolomeyko:
```suggestion
    fun extractFailedKeysFromBatchResults(batchResults: IntArray, commandKeys: List<String>): Set<String> {
```

Should be possible to use `Set` as keys better be unique.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -59,15 +62,15 @@ class StateManagerImpl(
...
-                }
+            val failedUpdates = dataSource.connection.transaction { conn ->
+                stateRepository.update(conn, states.map { it.toPersistentEntity() })
+            }
+
+            return if (failedUpdates.isEmpty()) {
+                emptyMap()
+            } else {
+                logger.warn("Optimistic locking check failed while updating States ${failedUpdates.joinToString()}")
+                get(failedUpdates)
```
vkolomeyko, line 74: Should we be a bit more careful here? 
Since call to `get()` method executed `IN` query, if we have > n * 100 elements, we might have a problem.
IMO it would be safer to break it into batches.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -59,15 +62,15 @@ class StateManagerImpl(
 
     override fun update(states: Collection<State>): Map<String, State> {
```
vkolomeyko, line 63: Since `net.corda.messaging.mediator.StateManagerHelper#persistStates` call this method its return type can be changed to `Map<String, State>`, i.e. non-nullable.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
sprintengine: I think it's a good concern, I did have chunking at one point but afaik it was removed in the review. I think typical batch sizes could be 500, but this is configurable. It would be a good idea to run some tests and determine if batch sizes of 500 states in the IN clause will be problematic with a populated postgres SM database before we make this change. I don't think I'll address it in this PR.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/PreparedStatementHelper.kt
```diff
@@ -0,0 +1,52 @@
...
+    fun extractFailedKeysFromBatchResults(batchResults: IntArray, commandKeys: List<String>): Set<String> {
+        if (batchResults.size != commandKeys.size) {
+            ("Number of results from batch (size: ${batchResults.size}) " +
+                    "does not match number of commands in the request (size ${commandKeys.size}").let {
+                log.warn(it)
+                throw PreparedStatementHelperException(it)
+            }
+        }
+        return batchResults.zip(commandKeys) { result, key -> CommandResultPair(key, result) }
+            .filter { it.result == NO_RECORD_UPDATED || it.result == Statement.EXECUTE_FAILED }
```
jujoramos: What happens if we get a `SUCCESS_NO_INFO`?. We could end up not updating a row due to optimistic locking check but assuming we did, in which case the repository would return a wrong answer to the user, right?. Wouldn't it be better to return the key in these cases and let the user retry/verify if needed?.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -34,42 +31,31 @@ class StateRepositoryImpl(private val queryProvider: QueryProvider) : StateRepos
...
-                }
+    override fun delete(connection: Connection, states: Collection<StateEntity>): Collection<String> {
+        return connection.prepareStatement(queryProvider.deleteStatesByKey).use { preparedStatement ->
+            for (s in states) {
+                preparedStatement.setString(1, s.key)
+                preparedStatement.setInt(2, s.version)
+                preparedStatement.addBatch()
+            }
+            val results = preparedStatement.executeBatch()
+            extractFailedKeysFromBatchResults(results, states.map { it.key })
```
jujoramos: Do the iterator (`for (s in states)`) and `states.map { it.key }` operations have the same order semantics?. We are relying on the actual position of the keys and results here, so we need to make sure they match (maybe populate a temporal list within the loop while adding statements to the batch).

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -34,42 +31,31 @@ class StateRepositoryImpl(private val queryProvider: QueryProvider) : StateRepos
...
+            for (s in states) {
+                preparedStatement.setString(1, s.key)
+                preparedStatement.setBytes(2, s.value)
+                preparedStatement.setString(3, s.metadata)
+                preparedStatement.setString(4, s.key)
+                preparedStatement.setInt(5, s.version)
+                preparedStatement.addBatch()
+            }
+            val results = preparedStatement.executeBatch()
+            extractFailedKeysFromBatchResults(results, states.map { it.key })
```
jujoramos: Do the iterator (`for (s in states)`) and `states.map { it.key }` operations have the same order semantics?. We are relying on the actual position of the keys and results here, so we need to make sure they match (maybe populate a temporal list within the loop while adding statements to the batch).

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
sprintengine: I think it is a valid point, but I'm hesitant to make that change in this PR incase I introduce a bug, I know the mediator does set the state to `null` when it deletes states, though it may be "State.value" that gets set to null.

driessamyn, line 63: I think the `update` function returning a collection of failed updates is maybe a bit weird/confusing as an API.
Would it not make more sense to throw an exception in update and have a `tryUpdate` function that explicitly allows for failed ones. It may then also be more natural for the return type of tryUpdate to include both the ones that were successful and those who weren't (e.g. return a data type that has 2 collections).

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/PostgresQueryProvider.kt
```diff
@@ -15,8 +15,8 @@ class PostgresQueryProvider : AbstractQueryProvider() {
     override val updateState: String
         get() = """
             UPDATE $STATE_MANAGER_TABLE SET
-            key = :$KEY_PARAMETER_NAME, value = :$VALUE_PARAMETER_NAME, version = version + 1, metadata = CAST(:$METADATA_PARAMETER_NAME as JSONB), modified_time = CURRENT_TIMESTAMP AT TIME ZONE 'UTC'
-            WHERE key = :$KEY_PARAMETER_NAME AND version = :$VERSION_PARAMETER_NAME
+            key = ?, value = ?, version = version + 1, metadata = CAST(? as JSONB), modified_time = CURRENT_TIMESTAMP AT TIME ZONE 'UTC'
```
driessamyn: Why do we need the JSONB cast?

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/PreparedStatementHelper.kt
driessamyn: I agree that this looks a bit risky.

Instead could we do something like.

For table:

```sql
CREATE TABLE Foo(id INT, bar VARCHAR(255));
INSERT INTO Foo(id, bar)
VALUES
    (1, 'a'),
    (2, 'b'),
    (3, 'c'),
    (4, 'd'),
    (5, 'e');
```
bulk update:

```sql
UPDATE Foo AS f
SET bar =  c.name FROM
(
    VALUES
        (1, 'aa'),
        (2, 'bb')
) AS c(id, name)
WHERE c.id = f.id RETURNING f.*;
```

returns:

```
1,aa
2,bb
```

sprintengine: Unfortunately `executeBatch` only returns the number of rows updated, not the results from the queries. 

An alternative batching approach could be to use temp datasets with common table expressions. Something like:

```
WITH updated_states (key, value, version, metadata) AS 
( 
VALUES (:key1, :value1, :version1, CAST(:metadata1 AS JSONB)), (:key2, :value2, :version2, CAST(:metadata2 AS JSONB)), 
… 
) 
UPDATE $STATE_MANAGER_TABLE SET 
… 
FROM updated_states 
WHERE $STATE_MANAGER_TABLE.key = updated_states.key AND $STATE_MANAGER_TABLE.version = updated_states.version 
RETURNING $STATE_MANAGER_TABLE.key;
```
Then, in kotlin:
```
val query = entityManager.createNativeQuery(NativeSQL) states.forEach { state -> query.setParameter("key", state.key) … }

val result = query.resultList
result.forEach { row ->
    updatedKeys.add((row as Array<*>)[0].toString())
}
return updatedKeys
```

driessamyn: But can't we just use `execute()` and call `resultSet`?

driessamyn: This works, for example, returning both updated rows (in the above table structure):

```kotlin
Class.forName("org.postgresql.Driver")
DriverManager
    .getConnection("jdbc:postgresql://localhost:5432/cordacluster", "postgres", "password").
        use { con ->
            val args = listOf(1 to "aa", 2 to "bb")
            con.prepareStatement("""
                UPDATE Foo AS f
                SET bar =  c.name FROM
                (
                    VALUES ${List(args.size) { "(?,?)" }.joinToString(",")}
                ) AS c(id, name)
                WHERE c.id = f.id RETURNING f.*;
            """.trimIndent()).use { stmt ->
                repeat(args.size) {
                    stmt.setInt(it*2 + 1, args[it].first)
                    stmt.setString(it*2 + 2, args[it].second)
                }
                stmt.execute()
                val resuls = stmt.resultSet
                while (resuls.next()) {
                    println("Updated: ${resuls.getInt(1)}|${resuls.getString(2)}")
                }
            }
        }
```

sprintengine: This doesn't work with `executeBatch`, neither does "RETURNING" with `executeBatch`. resultSet on preparedStatement that has executed batch returns `null`.

I could use `execute()` instead of `executeBatch()`, I can measure performance difference, but the original intent of this ticket was to use batching, not just get rid of hibernate. Would be interesting to compare perf of above approach though.

I do agree `RETURNING key` would be a much safer approach.

owenstanford: I suspect @driessamyn's approach would be slightly faster (needs testing) and cleaner for the determining which if any rows were skipped, but given the time pressure. However the current version is probably good enough for 5.1?

sprintengine: EDIT - sorry I misread the above query, I'll rerun these tests below with similar style query.

Performance of using `execute()` without batching is akin to the original code, thus there is no benefit of making that change:

```
override fun update(connection: Connection, states: Collection<StateEntity>): Collection<String> {
        val successfulKeys = mutableListOf<String>()
        states.forEach { s ->
            connection.prepareStatement(queryProvider.updateState).use { preparedStatement ->
                preparedStatement.setString(1, s.key)
                preparedStatement.setBytes(2, s.value)
                preparedStatement.setString(3, s.metadata)
                preparedStatement.setString(4, s.key)
                preparedStatement.setInt(5, s.version)
                preparedStatement.execute()
                val results = preparedStatement.resultSet
                while (results.next()) {
                    successfulKeys.add(results.getString(1))
                }
            }
        }
        return states.map { it.key } - successfulKeys.toSet()
    }
```

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
sprintengine: Sure, I like that idea

sprintengine: Given urgency of this change I might hold off making that here to reduce the footprint of the change, but I think it's a good idea in general

driessamyn (dismissed):
LGTM

driessamyn (dismissed):
LGTM

driessamyn (approved):

driessamyn:
E2E tests seems to have timed out: https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda5-e2e-tests/detail/feature%2FCORE-17733/10/pipeline

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -230,32 +231,40 @@ class StateManagerIntegrationTest {
...
         )
     }
 
+    @Test
+    fun `does not throw when updating states of size 0`() {
+        val failedUpdates = assertDoesNotThrow { stateManager.update(emptyList()) }
+
+        assertThat(failedUpdates).isEmpty()
+    }
+
```
jujoramos, line 267: This likely belongs to unit test, not integration.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -58,16 +61,17 @@ class StateManagerImpl(
     }
 
     override fun update(states: Collection<State>): Map<String, State> {
+        if (states.isEmpty()) return emptyMap()
```
jujoramos, line 64: This is already checked in the repository implementation, isn't it?.

jujoramos (approved):

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
sprintengine: Yep was realising in my sleep last night I'd put it in the integration test by accident! 😢

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
sprintengine: Yeah, but IMO the state-manager shouldn't have to "know" that the repository can't handle an empty list. I.e. from SM perspective, this is just to reduce lines of code executed .

I think repository needs the check as that's where the error occurs, as a safety net. In here, may or may not need it, but I don't see the point in executing right the way through to the repository if the function can exit right here.

jujoramos: That's arguable... why wouldn't a repository class (any class for what's worth) be able to handle an empty collection?. I could understand if a class "fails" to handle a `null` parameter, but empty collections shouldn't be an issue at all. Not a blocking change, anyways, I'm okay with leaving things as they are and make the change later.

sprintengine: I think we're saying similar things, the repository is dealing with the empty collection, this check here is basically not related, right?

jujoramos: Agreed, so why do you need it?

sprintengine: Just to avoid taking a connection from the hikaripool and starting a transaction with an empty list

PR #4937, opened 2023-10-19
CORE-17834: Allow empty responses from HTTP RPC

josephzunigadaly:
Changes to support HTTP RPC in Token Selection:
- Allow `RPCClient.deserializePayload` to return null because we have a "release tokens" request that needs no response
- Send `corda-request-key` HTTP header (read from `MediatorMessage`) because we need to send a consistent identifier for sticky load balancer routing.

### DIFF
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
@@ -76,6 +76,7 @@ class RPCClient(
             .flatMap { (key, value) -> listOf(key, value) }
             .toTypedArray()
 
+        @Suppress("SpreadOperator")
         return HttpRequest.newBuilder()
             .uri(URI(message.endpoint()))
             .headers(*headers)
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
@@ -55,9 +55,9 @@ class RPCClient(
     }
 
 
-    private fun deserializePayload(payload: ByteArray): Any {
+    private fun deserializePayload(payload: ByteArray): Any? {
         return try {
-            deserializer.deserialize(payload)!!
+            deserializer.deserialize(payload)
         } catch (e: Exception) {
             val errorMsg = "Failed to deserialize payload of size ${payload.size} bytes due to: ${e.message}"
             log.warn(errorMsg, e)
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
@@ -18,6 +18,8 @@ import net.corda.utilities.trace
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
+typealias HeadersMap = Map<String, String>
+
 class RPCClient(
     override val id: String,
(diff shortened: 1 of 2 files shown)


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -67,10 +73,19 @@ class RPCClient(
...
     private fun buildHttpRequest(message: MediatorMessage<*>): HttpRequest {
-        return HttpRequest.newBuilder()
+
+        val builder = HttpRequest.newBuilder()
             .uri(URI(message.endpoint()))
             .POST(HttpRequest.BodyPublishers.ofByteArray(message.payload as ByteArray))
-            .build()
+
+        // Add HTTP headers
+        message.getPropertyOrNull<HeadersMap>(HEADERS_PROPERTY)?.let { headers ->
```
owenstanford: I think for this we want to use the existing key property and generate a single custom header.
looking at `TaskManagerHelper.kt` we are adding the record key as a property of the `MediatorMessage`
```
    private fun Record<*, *>.toMessage() =
        MediatorMessage(
            value!!,
            headers.toMessageProperties().also { it[MSG_PROP_KEY] = key },
        )
```
We can use this to create a custom header on the HTTP request, something like
```
// Add HTTP headers
builder.header("corda-request-key", message.getProperty<Any>(MSG_PROP_KEY).toString())
```

josephzunigadaly: Done

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -67,10 +71,17 @@ class RPCClient(
...
+
+        val builder = HttpRequest.newBuilder()
             .uri(URI(message.endpoint()))
             .POST(HttpRequest.BodyPublishers.ofByteArray(message.payload as ByteArray))
-            .build()
+
+        // Add key HTTP header
+        message.getPropertyOrNull(MSG_PROP_KEY)?.let { key ->
+            builder.header("corda-request-key", key.toString())
+        }
```
vkolomeyko:
```suggestion
        message.getPropertyOrNull(MSG_PROP_KEY)?.let { value ->
            builder.header("corda-request-key", value.toString())
        }
```

Also, might be a good idea to introduce a constant for `"corda-request-key"` in case we need it from elsewhere in the code. E.g. in the unit test.

vkolomeyko (approved):
LGTM

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -12,12 +12,15 @@ import net.corda.messaging.api.exception.CordaHTTPServerErrorException
...
 import net.corda.messaging.api.mediator.MessagingClient
 import net.corda.messaging.api.mediator.MessagingClient.Companion.MSG_PROP_ENDPOINT
+import net.corda.messaging.api.mediator.MessagingClient.Companion.MSG_PROP_KEY
 import net.corda.messaging.utils.HTTPRetryConfig
 import net.corda.messaging.utils.HTTPRetryExecutor
 import net.corda.utilities.trace
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
+const val CORDA_REQUEST_KEY_HEADER = "corda-request-key"
```
ben-millar, line 22: Could this be moved to an external consts file?

ben-millar (approved):
Changes LGTM. This change doesn't support empty requests, only empty responses, but it sounds like that's intentional. Could be worth adding a unit tests for the null response scenario.

PR #4939, opened 2023-10-20
ES-1518: switch to KubernetesAgent class for Jenkins agent definition

wzur-r3:
* use KubernetesAgent to handle creation of `label`, `nodeSelector` and `defaultContainer` in more consistent way across many projects


driessamyn (approved):
LGTM

PR #4942, opened 2023-10-20
CORE-17733 - batch the creation of states in state-manager 

conal-smith:
Splitting this out into its own PR as it is potentially more risky

### DIFF
--- a/libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
+++ b/libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
@@ -246,12 +246,7 @@ class StateManagerIntegrationTest {
             )
         }
 
-        val start = System.nanoTime()
         val failedUpdates = stateManager.update(statesToUpdate)
-        val end = System.nanoTime()
-        val total: Long = (end - start)
-        val totalMs = total/1_000_000
-        println("Elapsed time for $stateCount states: $totalMs ms")
 
         assertThat(failedUpdates).isEmpty()
         softlyAssertPersistedStateEntities(
--- a/libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
+++ b/libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
@@ -509,45 +509,29 @@ class StateManagerIntegrationTest {
             { i, _ -> """{ "number": $i, "boolean": ${i % 2 == 0}, "string": "random_$i" }""" }
         )
 
-        assertThat(
-            stateManager.findByMetadataMatchingAll(
-                listOf(
-                    MetadataFilter("number", Operation.GreaterThan, 5),
-                    MetadataFilter("number", Operation.LesserThan, 7),
-                    MetadataFilter("boolean", Operation.Equals, true),
-                    MetadataFilter("string", Operation.Equals, "random_6"),
-                )
-            )
-        ).hasSize(1)
+        assertThat(stateManager.findByMetadataMatchingAll(listOf(
+            MetadataFilter("number", Operation.GreaterThan, 5),
+            MetadataFilter("number", Operation.LesserThan, 7),
+            MetadataFilter("boolean", Operation.Equals, true),
+            MetadataFilter("string", Operation.Equals, "random_6"),
+        ))).hasSize(1)
 
-        assertThat(
-            stateManager.findByMetadataMatchingAll(
-                listOf(
-                    MetadataFilter("number", Operation.GreaterThan, 5),
-                    MetadataFilter("number", Operation.LesserThan, 7),
-                    MetadataFilter("boolean", Operation.Equals, true),
-                    MetadataFilter("string", Operation.Equals, "non_existing_value"),
-                )
-            )
-        ).isEmpty()
+        assertThat(stateManager.findByMetadataMatchingAll(listOf(
+            MetadataFilter("number", Operation.GreaterThan, 5),
+            MetadataFilter("number", Operation.LesserThan, 7),
+            MetadataFilter("boolean", Operation.Equals, true),
+            MetadataFilter("string", Operation.Equals, "non_existing_value"),
+        ))).isEmpty()
 
-        assertThat(
-            stateManager.findByMetadataMatchingAll(
-                listOf(
-                    MetadataFilter("number", Operation.GreaterThan, 0),
-                    MetadataFilter("boolean", Operation.Equals, true),
-                )
-            )
-        ).hasSize(count / 2)
+        assertThat(stateManager.findByMetadataMatchingAll(listOf(
+            MetadataFilter("number", Operation.GreaterThan, 0),
+            MetadataFilter("boolean", Operation.Equals, true),
+        ))).hasSize(count / 2)
 
-        assertThat(
-            stateManager.findByMetadataMatchingAll(
-                listOf(
-                    MetadataFilter("number", Operation.NotEquals, 0),
-                    MetadataFilter("string", Operation.Equals, "non_existing_key"),
-                )
-            )
-        ).isEmpty()
+        assertThat(stateManager.findByMetadataMatchingAll(listOf(
+            MetadataFilter("number", Operation.NotEquals, 0),
+            MetadataFilter("string", Operation.Equals, "non_existing_key"),
+        ))).isEmpty()
     }
 
     @Test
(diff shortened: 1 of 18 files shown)


libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -89,6 +88,11 @@ class StateManagerIntegrationTest {
 
     private fun buildStateKey(index: Int) = "key_$index-$testUniqueId"
 
+    private val createStateQuery = """
+            INSERT INTO ${DbSchema.STATE_MANAGER_TABLE}
+            VALUES (:$KEY_PARAMETER_NAME, :$VALUE_PARAMETER_NAME, :$VERSION_PARAMETER_NAME, CAST(:$METADATA_PARAMETER_NAME as JSONB), CURRENT_TIMESTAMP AT TIME ZONE 'UTC')
+    """.trimIndent()
+
```
jujoramos: Can we use the query from `queryProvider` instead?

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -14,14 +14,24 @@ class StateRepositoryImpl(private val queryProvider: QueryProvider) : StateRepos
...
+                preparedStatement.setInt(3, s.version)
+                preparedStatement.setString(4, s.metadata)
+                preparedStatement.addBatch()
+            }
+            // Execute the batch of prepared statements.
+            // The elements in the 'results' array correspond to the commands in the batch.
+            // The order of elements in 'results' follows the order in which the statements were added to the batch.
+            // - An update count greater than or equal to zero indicates that the command was processed successfully,
+            //   and it represents the number of rows in the database affected by the command.
+            // - If optimistic locking check fails for a statement in the batch, that statement will have a '0' in the 'results' array.
```
jujoramos: There is not optimistic locking for the create statement, the insert can only fail if the unique key is violated, right?

jujoramos (changes requested):

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
sprintengine: Good spot I copy/pasted this, if insert fails the query returns a "0" in the result IntArray

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
sprintengine: Yeah I can change this function to use positional parameters instead of named parameters,

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -38,9 +34,8 @@ import org.junit.jupiter.api.TestInstance
 import org.junit.jupiter.params.ParameterizedTest
 import org.junit.jupiter.params.provider.ValueSource
 import java.time.Instant
-import java.util.UUID
+import java.util.*
```
jujoramos: Weird that `detekt` didn't pick this up?, shouldn't we use specific imports instead of the wildcard?.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/PostgresQueryProvider.kt
```diff
@@ -9,7 +9,8 @@ class PostgresQueryProvider : AbstractQueryProvider() {
     override val createState: String
         get() = """
             INSERT INTO $STATE_MANAGER_TABLE
-            VALUES (:$KEY_PARAMETER_NAME, :$VALUE_PARAMETER_NAME, :$VERSION_PARAMETER_NAME, CAST(:$METADATA_PARAMETER_NAME as JSONB), CURRENT_TIMESTAMP AT TIME ZONE 'UTC')
+            VALUES (?, ?, ?, CAST(? as JSONB), CURRENT_TIMESTAMP AT TIME ZONE 'UTC')
+            ON CONFLICT DO NOTHING
```
sprintengine, line 13: I added a test where it tries to create some states with a primary key too long for the column.

The test does not behave exactly as expected:

**Expected outcome:** 
The failing prepared statement commands will be added to the results `IntArray` with a "EXECUTE_FAILED" value (-2)

**Actual outcome:**
The entire batch is aborted with an exception, and rolled back entire transaction (no states in the batch were created):
```
java.sql.BatchUpdateException: Batch entry 0 INSERT INTO state
VALUES ('key_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx1-fabf5853-8b61-4fcf-a282-822a729e549d', ?, 0, CAST('{}' as JSONB), CURRENT_TIMESTAMP AT TIME ZONE 'UTC')
ON CONFLICT DO NOTHING was aborted: ERROR: value too long for type character varying(255)  Call getNextException to see other errors in the batch.
	at org.postgresql.jdbc.BatchResultHandler.handleError(BatchResultHandler.java:165)
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2402)
	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:574)
	at org.postgresql.jdbc.PgStatement.internalExecuteBatch(PgStatement.java:896)
	at org.postgresql.jdbc.PgStatement.executeBatch(PgStatement.java:919)
	at org.postgresql.jdbc.PgPreparedStatement.executeBatch(PgPreparedStatement.java:1685)
	at com.zaxxer.hikari.pool.ProxyStatement.executeBatch(ProxyStatement.java:127)
	at com.zaxxer.hikari.pool.HikariProxyPreparedStatement.executeBatch(HikariProxyPreparedStatement.java)
	at net.corda.libs.statemanager.impl.repository.impl.StateRepositoryImpl.create(StateRepositoryImpl.kt:26)
	at net.corda.libs.statemanager.impl.StateManagerImpl.create(StateManagerImpl.kt:39)
	at net.corda.libs.statemanager.impl.tests.StateManagerIntegrationTest.primaryKeyLengthFailuresWhenPersistingSomeStatesDoesNotHaltTheEntireBatch(StateManagerIntegrationTest.kt:218)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
Caused by: org.postgresql.util.PSQLException: ERROR: value too long for type character varying(255)
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2713)
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2401)
	... 12 more
```

sprintengine: I made some changes to address this scenario.

It is likely an edge case, but ironically an edge case that actually came up today in [slack discussions](https://r3holdco.slack.com/archives/C03ABNX2RHP/p1698047294341219)

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -31,23 +32,37 @@ class StateManagerImpl(
...
+     * @param stateEntities a list of states to create
+     * @return a collection of keys of states that failed to persist
+     */
+    private fun createStatesWithRetryOnBatchError(stateEntities: List<StateEntity>): Collection<String> {
+        return try {
+            dataSource.connection.transaction { conn ->
+                stateRepository.create(conn, stateEntities)
+            }
+        } catch (e: StateManagerBatchingException) {
+            logger.info("Error during batch state creation. Retrying creation of ${e.failedStates.size} states without batching.", e)
```
sprintengine, line 59: Conflicts, such as a primary key already exists, are handled in the batch create function. They are "ignored" and return a 0 which is translated to a "failed key".

This logic may be overkill, but catches errors in the batch such as `value too long for type character varying(255)`. These aren't treated as conflicts by postgres driver, but abort the entire batch.

Aborting entire batch is unnecessary because some states in a batch could be perfectly fine. Re-attempting to create these states individually will avoid infinite loop of flow engine retries and states in the same batch never completing.

PR #4945, opened 2023-10-20
CORE-17837 Add `UtxoSignedLedgerTransactionKryoSerializer`

lankydan:
Add a Kryo serializer for `UtxoSignedLedgerTransaction`.

This fixes a bug where `UtxoSignedLedgerTransaction` was being serialized with the Kryo serializer for `UtxoSignedTransaction`, due to it implementing the same interface; however, the serializer can only correct serialize `UtxoSignedTransactionImpl`s.

Adding a specific serializer means Kryo picks up the correct one to serialize and deserialize with.


blsemo (approved):
LGTM

PR #4946, opened 2023-10-20
CORE-17411: Workaround when we encounter enriched MGM request

vkolomeyko:
So that we will not perform re-registration at all in this case.
Also adding some additional logging.

### DIFF
--- a/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
+++ b/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
@@ -65,7 +65,7 @@ internal class VirtualNodeUpgradeOperationHandler(
         private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
 
         private val Map<String, Any?>.isEnriched: Boolean
-            get() = containsKey("corda.cpi.name")
+            get() = containsKey(MemberInfoExtension.MEMBER_CPI_NAME)
     }
 
     private val keyValuePairListDeserializer: CordaAvroDeserializer<KeyValuePairList> by lazy {
--- a/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
+++ b/components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
@@ -217,6 +217,7 @@ internal class VirtualNodeUpgradeOperationHandler(
      * after the virtual node has been upgraded, so that the member CPI version is up-to-date.
      * Republishes the MGM's Member Info, if the Group Policy was changed.
      */
+    @Suppress("NestedBlockDepth")
     private fun reRegisterMember(upgradedVNodeInfo: VirtualNodeInfo, cpiMetadata: CpiMetadata) {
         val holdingIdentity = upgradedVNodeInfo.holdingIdentity
         val membershipGroupReader = membershipGroupReaderProvider.getGroupReader(holdingIdentity)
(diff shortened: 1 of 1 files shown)


BenYip123 (dismissed):

vkolomeyko:
E2E test run to validate this change: https://ci02.dev.r3.com/job/Corda5/view/Full%20Health%205.1/job/corda5-e2e-tests/job/release%252F5.1/1111/

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -63,6 +63,9 @@ internal class VirtualNodeUpgradeOperationHandler(
 
     private companion object {
         private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+
+        private val Map<String, Any?>.isEnriched: Boolean
+            get() = containsKey("corda.cpi.name")
```
vkolomeyko: @nikinagy - I will probably need your input here, if this is the best way to tell that request is enriched.

nikinagy: Yes, this a good indicator. I would advise though to use the MemberInfoExtension.MEMBER_CPI_NAME for easier future maintenance.

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -63,6 +63,9 @@ internal class VirtualNodeUpgradeOperationHandler(
 
     private companion object {
         private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+
+        private val Map<String, Any?>.isEnriched: Boolean
+            get() = containsKey(MemberInfoExtension.MEMBER_CPI_NAME)
```
dimosr: nit, but given we don't have an explicit type here and it's just a map I think it'd be better if we did that via a private function receiving the context as an input parameter, rather than an extension function over a map.

components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -245,22 +249,31 @@ internal class VirtualNodeUpgradeOperationHandler(
...
+                        val registrationRequestDetails = payload.sortedBy { it.serial }.last()
+                        
                         val registrationContext = registrationRequestDetails
                             .memberProvidedContext.data.array()
                             .deserializeContext(keyValuePairListDeserializer)
                             .toMutableMap()
 
+                        if (registrationContext.isEnriched) {
+                            logger.warn("Enriched registration request retrieved, which cannot be used for re-registration. " +
+                                    "Please perform MGM re-registration of vNode $holdingIdentity manually.")
```
dimosr: Assuming we intend this log to be surfaced to operators, I think it would be better if we can keep it a bit more user-friendly. For example, an operator would probably not know what "enriched" means in this context. We could keep the log generic and some commentary in a comment for us e.g.:
```
// In some cases, the registration request contains the platform-transformed data, instead of the user-provided context so we skip re-registration. This is applicable only for old registration requests, as it's now fixed.
logger.warn("The platform was not able to automatically re-register the vNode. " +
                                    "Please perform re-registration of vNode $holdingIdentity manually.")
```
I assume that our release notes will also contain something generic about this (e.g. "there are some scenarios, where an automatic re-registration cannot be performed after the CPI upgrade ....), since users don't need to know the exact details.

dimosr (commented):
Just a few minor comments, otherwise looks fine to me.

nikinagy (approved):
LGTM

dimosr (approved):

PR #4948, opened 2023-10-20
CORE-17881: Individual State Manager Per Worker Type

jujoramos:
Update Helm Chart so every worker type can configure its own intance
of the State Manager.


charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -356,6 +324,114 @@ spec:
...
+{{- $ := index . 0 }}
+{{- $workerKey := index . 1 }}
+{{- $authConfig := index . 2 }}
+{{- $worker := (index $.Values.workers $workerKey) }}
+{{- $workerName := printf "%s-worker" ( include "corda.workerTypeKebabCase" $workerKey ) }}
+{{- with index . 0 }}
+{{- if .Values.bootstrap.db.enabled }}
+---
+apiVersion: batch/v1
+kind: Job
```
davidcurrie: My previous comment about running these jobs serially being sub-optimal still stands.

charts/corda-lib/templates/_bootstrap.tpl
```diff
@@ -356,6 +324,114 @@ spec:
...
+      {{- end }}
+      initContainers:
+        - name: generate
+          image: {{ include "corda.bootstrapCliImage" . }}
+          imagePullPolicy: {{ .Values.imagePullPolicy }}
+          {{- include "corda.bootstrapResources" . | nindent 10 }}
+          {{- include "corda.containerSecurityContext" . | nindent 10 }}
+          env:
+            {{- include "corda.bootstrapCliEnv" . | nindent 12 }}
+            {{- include "corda.bootstrapStateManagerDbEnv" ( list $ $worker $workerKey $authConfig ) | nindent 12 }}
```
davidcurrie: I think this should be `corda.stateManagerDbEnv`; otherwise, we're putting the bootstrap credentials into the `db_connections` table. (Looks like this may have been an existing issue.)

charts/corda-lib/templates/_stateManager.tpl
```diff
@@ -0,0 +1,155 @@
...
+
+{{/* State Manager JDBC URL */}}
+{{- define "corda.stateManagerJdbcUrl" -}}
+{{- $ := index . 0 }}
+{{- $workerConfig := index . 1 }}
+{{- if $workerConfig.stateManager.db.host -}}
+jdbc:{{- include "corda.stateManagerDbType" (list $ $workerConfig) -}}://{{- $workerConfig.stateManager.db.host -}}:{{- include "corda.stateManagerDbPort" (list $ $workerConfig) -}}/{{- include "corda.stateManagerDbName" (list $ $workerConfig) -}}
+{{- else -}}
+jdbc:{{- include "corda.clusterDbType" $ -}}://{{- $.Values.db.cluster.host -}}:{{- include "corda.clusterDbPort" $ -}}/{{- include "corda.clusterDbName" $ -}}
+{{- end -}}
```
davidcurrie, line 63: The same comment I made on Conal's original PR: this conditional is redundant. You could just use the first branch and then all of the parts fall through to their respective cluster DB equivalents anyway.

charts/corda-lib/templates/_stateManager.tpl
```diff
@@ -0,0 +1,155 @@
...
+        {{-   if $bootConfig.username.valueFrom.secretKeyRef.name }}
+        name: {{ $bootConfig.username.valueFrom.secretKeyRef.name | quote }}
+        key: {{ required (printf "Must specify bootstrap.db.stateManager.%s.username.valueFrom.secretKeyRef.key" $workerKey) $bootConfig.username.valueFrom.secretKeyRef.key | quote }}
+        {{-   else }}
+        name: {{ include "corda.stateManagerDefaultBootSecretName" ( list $ $workerKey ) | quote }}
+        key: "username"
+        {{-  end }}
+        {{- else }}
+        {{- include "corda.bootstrapClusterPgUser" $ | nindent 6 }}
+        {{- end }}
```
davidcurrie: nit: this section is doubly indented

charts/corda-lib/templates/_stateManager.tpl
```diff
@@ -0,0 +1,155 @@
...
+- name: STATE_MANAGER_PGPASSWORD
+  valueFrom:
+    secretKeyRef:
+        {{- if $worker.stateManager.db.host }}
```
davidcurrie: ditto

charts/corda-lib/templates/_stateManager.tpl
```diff
@@ -0,0 +1,155 @@
...
+
+{{/* State Manager Database Credentials Environment Variables */}}
+{{- define "corda.stateManagerDbEnv" -}}
+{{- $ := index . 0 -}}
+{{- $worker := index . 1 -}}
+{{- $workerKey := index . 2 -}}
+- name: STATE_MANAGER_USERNAME
+  valueFrom:
+    secretKeyRef:
+        {{- if $worker.stateManager.db.host }}
```
davidcurrie: ditto

charts/corda-lib/templates/_stateManager.tpl
```diff
@@ -0,0 +1,155 @@
...
+        name: {{ include "corda.stateManagerDefaultSecretName" ( list $ $workerKey ) | quote }}
+        key: "username"
+        {{-  end }}
+        {{- else }}
+        {{- include "corda.clusterDbUsername" $ | nindent 6 }}
+        {{- end }}
+- name: STATE_MANAGER_PASSWORD
+  valueFrom:
+    secretKeyRef:
+        {{- if $worker.stateManager.db.host }}
```
davidcurrie: ditto

charts/corda/templates/bootstrap.yaml
```diff
@@ -2,3 +2,13 @@
 {{- include "corda.bootstrapDbJob" . }}
 {{- include "corda.bootstrapKafkaJob" . }}
 {{- include "corda.bootstrapRbacJob" . }}
+{{- if .Values.bootstrap.db.enabled }}
+{{-   range $workerName, $authConfig := .Values.bootstrap.db.stateManager }}
+{{-     $workerConfig := (index $.Values.workers $workerName) }}
+{{-     if and (not $workerConfig.stateManager.db.host) (ne $authConfig.username.value "") (not $authConfig.valueFrom) -}}
```
davidcurrie: `""` is falsy and I think you need to look all the way down to the secret name making the logic something like:
```suggestion
{{-     if and ( not $workerConfig.stateManager.db.host ) ( not $authConfig.username.value ) ( not $authConfig.valueFrom.secretKeyRef.name ) -}}
```

charts/corda/templates/bootstrap.yaml
```diff
@@ -2,3 +2,13 @@
...
+{{-       fail ( printf "Must specify workers.%s.stateManager.host when bootstrap.db.stateManager.%s is configured" $workerName $workerName ) }}
```
davidcurrie: Is the error message the wrong way round? When they specify a separate host, they need to provide credentials?

charts/corda/templates/secrets.yaml
```diff
@@ -37,6 +26,19 @@
     ( dict "salt" ( dict "generate" 32 ) "passphrase" ( dict "generate" 32 ) )
   )
 }}
+{{- range $workerKey, $workerConfig := .Values.workers }}
+{{-   if and ($workerConfig.stateManager) ($workerConfig.stateManager.db.host) ($workerConfig.stateManager.db.username) (and (or ($workerConfig.stateManager.db.username.value) ($workerConfig.stateManager.db.username.valueFrom.secretKeyRef.name))) -}}
```
davidcurrie: We deliberately specify a supported version of `helm` that doesn't have this problem but, FYI, older versions of go-template did not have lazy evaluation so you had to nest conditionals rather than using `and`. Also worth noting that you can use parentheses to trap null pointers e.g. `( ( $workerConfig.stateManager ).db.host )` will evaluate to `null` if `$workerConfig.stateManager` is `null` rather than failing.

davidcurrie: All that said, I don't understand why this is checking anything other than that `db.host` has been set?

charts/corda-lib/templates/_bootstrap.tpl
davidcurrie: Ignore that - I remember you telling me before that this isn't populating `db_connections`, it's using the credentials with liquibase.

charts/corda-lib/templates/_bootstrap.tpl
jujoramos: It is sub-optimal, agreed, but I would need some bash magic in order to pass an arbitrary list of credentials and generate them one by one within a single job (different workers can connect to different state manager databases)... I'll create a ticket or something to improve it in the future (this part is not user facing so we can do it after the release).

charts/corda-lib/templates/_stateManager.tpl
jujoramos: I'd need to an extra function for `{{- $workerConfig.stateManager.db.host -}}` in that case 🤔

charts/corda-lib/templates/_stateManager.tpl
jujoramos: What do you mean?, I've tested the chart with different options to verify all branches and the resulting yaml file seemed correct 🤔

charts/corda-lib/templates/_bootstrap.tpl
davidcurrie: It's kind of user-facing in that they're the ones who have to sit there while the pre-install jobs take an age to run through. It's certainly not going to help the setup times for our tests. I don't think there's any bash magic required - just take the same logic you have for iterating to create the jobs and iterate to create containers within a job.

charts/corda-lib/templates/_stateManager.tpl
davidcurrie: You're right - I missed that one.

charts/corda-lib/templates/_stateManager.tpl
davidcurrie: It'll be valid YAML but whereas you have an explicit indent of 6 for the bootstrap user, you have an indent of 8 for everything else.

davidcurrie: But don't check the 6 to 8! The indentation is two spaces so it should be 6 here not 8.

jujoramos: Will make the change, thanks 👍

charts/corda-lib/templates/_stateManager.tpl
jujoramos: Will make the change, thanks 👍

charts/corda-lib/templates/_stateManager.tpl
jujoramos: Will make the change, thanks 👍

charts/corda-lib/templates/_stateManager.tpl
jujoramos: Will make the change, thanks 👍

charts/corda/templates/secrets.yaml
jujoramos: 👍

charts/corda/templates/bootstrap.yaml
jujoramos: Done (used another approach and improved the condition, though)

charts/corda/templates/bootstrap.yaml
jujoramos: 👍

charts/corda-lib/templates/_bootstrap.tpl
jujoramos: I've tried using the same approach as with the regular bootstrap job (one `init-container` to generate the `sql` and one `container` to apply it), but it causes issues when trying to deploy Corda against a single state manager database instance (the developer use case I imagine), basically because the same Liquibase file is applied in parallel and the sql doesn't contain the `IF NOT EXISTS` for the `databasechangeloglock` table (containers executed within a `Job` are all started at once):
```
apply-state-manager-flow-mapper-worker Applying State Manager Specification for flow-mapper-worker...
apply-state-manager-flow-mapper-worker psql:/tmp/stateManager-flowMapper/statemanager.sql:2: NOTICE:  schema "state_manager" already exists, skipping
apply-state-manager-flow-mapper-worker psql:/tmp/stateManager-flowMapper/statemanager.sql:20: ERROR:  relation "databasechangeloglock" already exists

apply-state-manager-flow-worker Applying State Manager Specification for flow-worker...
apply-state-manager-flow-worker Applying State Manager Specification for flow-worker... Done!
```

---

Instead of the regular approach, so, I've included the state manager initialisation into the existing `setup-db` job but purely through `init-containers` to force sequential execution (is slower than before as we have more `init-containers`, but it's still faster than the original approach within the `PR` of having yet another 2 Jobs for the state manager setup).

conal-smith (commented):
LGTM, tested helm template debug with a few state-manager overrides and it works as expected
```
helm template corda charts/corda --debug --values values-prereqs.yaml --values state-manager.yaml >  with-state-manager-yaml.txt
```

Tested without state-manager.yaml and it defaults correctly to cluster.
```
helm template corda charts/corda --debug --values values-prereqs.yaml >  without-sm.txt
```

Happy to approve once endToEndPipeline is merged

conal-smith (approved):
lgtm

ronanbrowne (approved):
LGTM

.ci/e2eTests/corda.yaml
```diff
@@ -8,12 +8,20 @@ bootstrap:
           secretKeyRef:
             key: "postgres-password"
     stateManager:
-      username:
-        value: "state-manager-user"
-      password:
-        valueFrom:
-          secretKeyRef:
-            key: "password"
+      flow:
```
JasonSinclair95, line 11: Do you still need this declared in the corda.yaml when the exact same is declared in the file [state-manager.yaml](https://github.com/corda/corda-runtime-os/pull/4948/files#diff-846a5c1c17ed9e802428faa4a5a513055ac384c16426c47ca3ad1b9b03ccd4f0)

JasonSinclair95 (changes requested):

.ci/e2eTests/corda.yaml
jujoramos: We need this here, yes, not only because the credentials are different, but because the files are used for different use cases (one for the pipeline, the other one for local deployment during development).

JasonSinclair95 (approved):
LGTM

PR #4949, opened 2023-10-20
CORE-17297: Fix incompatible responder flow failure handling

Tom-Fitzpatrick:
This PR addresses the scenario where two vNodes with with incompatible flow protocol versions try to communicate. At the moment, this results in hitting a `FlowSessionStateException` in the responder flow. This gets handled by the generic exception handler in `FlowEventProcessorImpl.getFlowPipelineResponse` and the initiating flow getting left in a RUNNING state indefinitely.

The `FlowSessionStateException` is thrown by the responder flow because we have not created a SessionState when we try to send the Error message.

This PR adds session State creation to `sendErrorMessage `so that it can complete correctly and fail the initiating flow.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
@@ -21,6 +21,7 @@ import net.corda.flow.pipeline.sessions.FlowSessionManager
 import net.corda.flow.pipeline.sessions.protocol.FlowAndProtocolVersion
 import net.corda.flow.state.FlowCheckpoint
 import net.corda.flow.utils.KeyValueStore
+import net.corda.flow.utils.emptyKeyValuePairList
 import net.corda.flow.utils.keyValuePairListOf
 import net.corda.flow.utils.toMap
 import net.corda.session.manager.Constants.Companion.FLOW_PROTOCOL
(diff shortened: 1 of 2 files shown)


components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
```diff
@@ -231,6 +232,14 @@ class SessionEventHandler @Activate constructor(
         sessionId: String,
         exception: Throwable
     ) {
+        context.checkpoint.putSessionState(sessionManager.generateSessionState(
+            sessionId,
+            emptyKeyValuePairList(),
+            (context.inputEventPayload as SessionEvent).initiatingIdentity,
+            Instant.now(),
+            SessionStateType.CONFIRMED
```
LWogan: May aswel set the status as ERROR when its created. The sendErrorMesages() function will set it to ERROR anyway, but it might read more clearly to the next person if we initialize the state as status ERROR

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
```diff
@@ -231,6 +232,14 @@ class SessionEventHandler @Activate constructor(
         sessionId: String,
         exception: Throwable
     ) {
+        context.checkpoint.putSessionState(sessionManager.generateSessionState(
```
LWogan, line 235: This change makes sense. It was something missed when refactoring the flow session protocol. We used to always create the session state before looking up the flow versioning is established.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
Tom-Fitzpatrick: Ah cheers, have made this change now

LWogan (approved):
LGTM

vkolomeyko (approved):
LGTM

PR #4950, opened 2023-10-20
CORE-17899 Add nginx for token selection sharding

davidcurrie:
Setting `workers.tokenSelection.sharding.enabled=true` results in the creation of an nginx ingress controller that is forwarding to the token selection internal service but with affinity to the individual worker instances based on a hash of `CORDA-REQUEST-KEY`.

The nginx configuration is based on the results of running:
```
helm template ingress-nginx ingress-nginx \
  --repo https://kubernetes.github.io/ingress-nginx \
  --set controller.service.type=ClusterIP,controller.replicaCount=2,controller.ingressClassResource.name=corda-internal,controller.scope.enabled=true,controller.admissionWebhooks.enabled=false,rbac.scope=true --skip-crds
  ```
  
Changes made to the `corda` Helm chart will need to be replicated for Corda Enterprise, and the `nginx` image packaged with those delivered to customers. The documentation then needs to cover how to use the packaged image rather than the one from `registry.k8s.io` by overriding the values in `workers.tokenSelection.sharding.image`.

### DIFF
--- a/charts/corda-lib/templates/_helpers.tpl
+++ b/charts/corda-lib/templates/_helpers.tpl
@@ -812,7 +812,48 @@ Get the endpoint argument for a given worker
 {{- define "corda.getWorkerEndpoint" }}
 {{- $context := .context }}
 {{- $worker := .worker }}
-{{- $workerName := printf "%s-%s-worker" (include "corda.fullname" $context) (include "corda.workerTypeKebabCase" $worker) }}
-{{- $workerServiceName := include "corda.workerInternalServiceName" $workerName }}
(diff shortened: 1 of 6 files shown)


josephzunigadaly (approved):

vkolomeyko:
E2E build is started to validate these changes: https://ci02.dev.r3.com/job/Corda5/view/Full%20Health%205.1/job/corda5-e2e-tests/job/release%252F5.1/1131/

vkolomeyko (approved):
LGTM

PR #4951, opened 2023-10-20
CORE-17389: Process flow timeout events and clean up checkpoints

JamesHR3:
### Problem description

The flow engine currently responds to scheduled task triggers to produce flow timeout events when a session has timed out. These need to be processed to remove the timed out checkpoint and send any cleanup events (session errors and mapper cleanup events) to the rest of the system. The flow status also needs to be updated.

### Solution

Add a processor for the flow timeout events.

This PR also includes a checkpoint cleanup handler class, which generates the relevant events for when a flow is terminated with an error. This can be reused in the `FlowEventExceptionProcessor`, although this has not been integrated in this PR. The objective here is to encapsulate this cleanup logic. At present, there is a bug in the flow engine that means not all of this logic is present everywhere it is needed.

### DIFF
--- a/components/flow/flow-mapper-service/src/main/kotlin/net/corda/session/mapper/service/executor/FlowMapperMessageProcessor.kt
+++ b/components/flow/flow-mapper-service/src/main/kotlin/net/corda/session/mapper/service/executor/FlowMapperMessageProcessor.kt
@@ -66,6 +66,8 @@ class FlowMapperMessageProcessor(
                         mapOf(FLOW_MAPPER_STATUS to it.toString())
                     } ?: mapOf()
 
+
+
                     StateAndEventProcessor.Response(
                         State(result.flowMapperState, state?.metadata?.let {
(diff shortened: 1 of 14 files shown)


LWogan (approved):
[redacted comment]

LWogan (changes requested):
redacted my last approval. I think theres an issue with the flow session timeout checkpoint cleanup part

LWogan (approved):
I've made some updates to handle:
- ERROR flow mapper state cleanup
- merging messaging config values into flow config for use by the flow session cleanup logic

I've raised some bugs around:
- flow engine does not cleanup RPC flow mapper states
- UtxoDemoFlow leaves sessions in a hung state (eventually cleaned up by logic in this pr)

Other than that this PR looks good to me. We should make use of the FlowMaintenance class in the flow engine to avoid duplicate code. A ticket has been raised for this.

Omar-awad (approved):
LGTM

PR #4956, opened 2023-10-23
[CORE-16983] Prerequistites for CORE-16983 Serialize and Send Merkle Proof through Façade 

szymonsztuka:
_Code changes copied from https://github.com/corda/corda-runtime-os/pull/4691 to simplify that PR._

- Remove convenience methods for creating a client proxy with a default Json Mappers
`Any.buildDispatcher`and `Facade.getClientProxy`. These methods are used for testing only. In the following code change (the next PR) we add new serializes and deserializes with e.g. Merkle Tree Proof. For unit testing new serializes will be not needed and the core Facade module doesn't create them, however using a default JSON serialize add confusion - should I replaced the existing "default"  JSON Mapper with a new one, or should I leave it as it is. By removing any defaults we delegate this problem when the proxy is actually created, for production code it was always like that, now it's also this way for test code.

- As a cleanup - remove an unused Facade Proxy instance in a test.


ac101m (approved):
LGTM

PR #4957, opened 2023-10-23
CORE-17941 Create basic plugin and make backchain optional

nkovacsx:
### **This is only a SPIKE ticket work on a SPIKE branch**


components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/filtered/factory/FilteredTransactionFactoryImpl.kt
```diff
@@ -38,7 +39,7 @@ class FilteredTransactionFactoryImpl @Activate constructor(
     @Suspendable
     override fun create(
         wireTransaction: WireTransaction,
-        componentGroupFilterParameters: List<ComponentGroupFilterParameters>
+        componentGroupFilterParameters: List<ComponentGroupFilterParameters>,
```
jennyang-r3, line 42: nit: unnecessary comma

PR #4959, opened 2023-10-23
CORE-17867 - Added logging to `VirtualNodeWriteEventHandler`

filipesoliveira:
An e2e test is flaky and extra logging is required in order to understand what might be causing issues.


vkolomeyko (approved):
LGTM

PR #4961, opened 2023-10-23
CORE-17932 ensure the flow service depends on the membership group reader providers lifecycle

LWogan:
Small fix to ensure the flow engine does not try to run without the membership cache populated and ready. This has commonly affected multicluster large network tests

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowService.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowService.kt
@@ -16,6 +16,7 @@ import net.corda.lifecycle.RegistrationStatusChangeEvent
 import net.corda.lifecycle.StartEvent
 import net.corda.lifecycle.StopEvent
 import net.corda.lifecycle.createCoordinator
+import net.corda.membership.read.MembershipGroupReaderProvider
 import net.corda.sandboxgroupcontext.service.SandboxGroupContextComponent
 import net.corda.schema.configuration.ConfigKeys.BOOT_CONFIG
 import net.corda.schema.configuration.ConfigKeys.FLOW_CONFIG
@@ -69,6 +70,7 @@ class FlowService @Activate constructor(
                             LifecycleCoordinatorName.forComponent<CpiInfoReadService>(),
                             LifecycleCoordinatorName.forComponent<FlowExecutor>(),
                             LifecycleCoordinatorName.forComponent<FlowMaintenance>(),
+                            LifecycleCoordinatorName.forComponent<MembershipGroupReaderProvider>(),
                         )
                     )
                 flowMaintenance.start()
--- a/components/flow/flow-service/src/test/kotlin/net/corda/flow/service/FlowServiceTest.kt
+++ b/components/flow/flow-service/src/test/kotlin/net/corda/flow/service/FlowServiceTest.kt
@@ -7,6 +7,7 @@ import net.corda.flow.MINIMUM_SMART_CONFIG
 import net.corda.flow.maintenance.FlowMaintenance
 import net.corda.lifecycle.LifecycleCoordinatorName
 import net.corda.lifecycle.test.impl.LifecycleTest
+import net.corda.membership.read.MembershipGroupReaderProvider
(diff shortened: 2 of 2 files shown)


ben-millar (approved):

components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowService.kt
```diff
@@ -69,6 +70,7 @@ class FlowService @Activate constructor(
                             LifecycleCoordinatorName.forComponent<CpiInfoReadService>(),
                             LifecycleCoordinatorName.forComponent<FlowExecutor>(),
                             LifecycleCoordinatorName.forComponent<FlowMaintenance>(),
+                            LifecycleCoordinatorName.forComponent<MembershipGroupReaderProvider>(),
```
dimosr, line 73: Not very familiar with all the components of the flow engine, but given we do `start` on a couple of them below (e.g. flow executor) right after `followStatusChangesByName`, doesn't that mean it won't necessarily wait for `MembershipGroupReaderProvider` to go UP first? 

I would have thought we have to wait for the `UP` signal (indicating all the dependencies went up) before starting the components that use them.

dimosr (commented):
Have we tested the fix in any way? Given the large network test is a bit non-deterministic around this issue, an easier way to test the issue is fixed would be to introduce an artificial delay to the membership group reader and see if the flow engine will start first or will wait for it.

components/flow/flow-service/src/main/kotlin/net/corda/flow/service/FlowService.kt
LWogan: FlowService and MembershipGroupReaderProvider are  dependent component of FlowProcessorImpls coordinator. This means when FlowProcessor starts its coordinator, FlowService will be started:
```
private val dependentComponents = DependentComponents.of(
        ::configurationReadService,
        ::flowService,
        ::virtualNodeInfoReadService,
        ::cpiInfoReadService,
        ::sandboxGroupContextComponent,
        ::membershipGroupReaderProvider,
        ::groupParametersReaderService,
        ::cpkReadService,
        ::groupPolicyProvider,
        ::membershipQueryClient
    )

    private val lifecycleCoordinator =
        coordinatorFactory.createCoordinator<FlowProcessorImpl>(dependentComponents, ::eventHandler)
```

When FlowService is started it will kick off its own coordinator. The StartEvent for the FlowService coordinator lists some service that it will follow the status for, now including the MembershipGroupReaderProvider. 
```
is StartEvent -> {
                registration?.close()
                registration =
                    coordinator.followStatusChangesByName(
                        setOf(
                           .......
                            LifecycleCoordinatorName.forComponent<MembershipGroupReaderProvider>(),
                        )
                    )
                flowMaintenance.start()
                flowExecutor.start()
            }
```

This means that the FlowService will receive a RegistrationStatusChangeEvent when all those services have been marked as UP. The MembershipGroupReaderProvider appears to only mark itself as UP when its cache is populated, which is good.

```
                if (event.status == LifecycleStatus.UP) {
                    configHandle?.close()
                    configHandle = configurationReadService.registerComponentForUpdates(
                        coordinator,
                        configSections
                    )
                ...
            }
```

When MembershipGroupReaderProvider and all other services are UP, FlowServices coordinator receives a RegistrationStatusChangeEvent with status UP. At this point the FlowService starts listening for config. Then it receives a  ConfigChangedEvent and will then start the Flow Engine via the FlowExecutor service.

```
is ConfigChangedEvent -> {
                ....
                flowExecutor.onConfigChange(config)
                    ....
            }

```

dimosr: hm ok, I think I see now.. `flowExecutor.start()` doesn't actually kick-off processing, it just starts the coordinator. It's the `onConfigChange` that kicks off processing.

dimosr (approved):
LGTM - I'd suggest running a manual test as a sanity check to ensure the order is as expected now at runtime before merging.

LWogan:
Did a manual test inserting a delay to the membership provider caching and confirmed the expected behavior

PR #4963, opened 2023-10-24
CORE-17899: Eliminate use of deprecated annotation for Ingress

vkolomeyko:
Before the change:
```
> helm install corda -n corda charts/corda --values values-prereqs.yaml --wait
W1024 13:39:43.341482   16912 warnings.go:70] annotation "kubernetes.io/ingress.class" is deprecated, please use 'spec.ingressClassName' instead
```

Tested-by: vkolomeyko
By using local deployment of C5 cluster

### DIFF
--- a/charts/corda-lib/templates/_worker.tpl
+++ b/charts/corda-lib/templates/_worker.tpl
@@ -62,9 +62,9 @@ metadata:
   {{- with .annotations }}
     {{- toYaml . | nindent 4 }}
   {{- end }}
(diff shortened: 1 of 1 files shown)


josephzunigadaly (approved):

vkolomeyko:
Looks like more work is required to remove this working due to:
```
The newer ingressClassName field on Ingresses is a replacement for that annotation, but is not a direct equivalent. While the annotation was generally used to reference the name of the Ingress controller that should implement the Ingress, the field is a reference to an IngressClass resource that contains additional Ingress configuration, including the name of the Ingress controller.
```

PR #4965, opened 2023-10-24
CORE-17950 make start flow events idempotent

LWogan:
The multi source mediator currently updates the state manager after processing each event in a batch. If there is an error that occurs after this has happened, the consumer will rollback to its last poll position and replay events. The StartFlow event is the only FlowEvent not idempotent. This PR addresses this and will ignore subsequent StartFlowEvents unless the checkpoint is null or we are in a retry state and flow state is null (indicating the StartFlow is what needs to be retried)


ben-millar (dismissed):

mbrkic-r3 (dismissed):
LGTM

thiagoviana (approved):

Omar-awad (approved):
LGTM

PR #4969, opened 2023-10-25
CORE-17624: Fail "Create CPI" operations when the CLI reports failure

josephzunigadaly:
Our E2E tests showed an error where an attempt to create a CPI had failed, but they continued further into the test instead of exiting at the first sign of failure.

This change turns `CreateCpiV2` into a `Callable<Int>`, where the `Int` return type is the exit code of the command line process. This allows us to communicate failures using the exit code as well as console output. That makes it easy for code calling `CreateCpiV2` to test for failures and act appropriately.

Changes in this PR:
- Change `CreateCpiV2` to return an exit code (zero on success, non-zero on error)
- Update all usages to check the exit code and stop what they are doing when an error happens


kyriathar (approved):
OK (packaging)

nikinagy (approved):
LGTM (network)

PR #4970, opened 2023-10-25
CORE-17930 - Automated re-registrations triggered by CPI upgrade are failing intermittently

nikinagy:
Description:
There were intermittent failures observed on e2e test branch caused by race condition of registration request persistence on membership side. We were persisting the registration request with different member provided context at 2 places. The member context field is a non-updatable field by default, so it was up to the race condition which version for persisted. However, with the 2nd version re-registration attempts were failing.

Fix:
- Persist same context everywhere.
- Added null check of serial on vnode upgrade.
- Let the persistence update the serial of the request.

Testing:
Triggered the disabled vnode upgrade tests multiple times locally and check that we have the correct registration request in the db with the serial numbers in place.


components/virtual-node/virtual-node-write-service-impl/src/main/kotlin/net/corda/virtualnode/write/db/impl/writer/asyncoperation/handlers/VirtualNodeUpgradeOperationHandler.kt
```diff
@@ -270,8 +270,10 @@ internal class VirtualNodeUpgradeOperationHandler(
                             return
                         }
 
-                        val updatedSerial = registrationRequestDetails.serial + 1
-                        registrationContext[MemberInfoExtension.SERIAL] = updatedSerial.toString()
+                        if (registrationRequestDetails.serial != null) {
```
vkolomeyko, line 273: Is it a valid situation to have a `serial = null` after vNode been created and registered with MGM?

nikinagy: Unfortunately yes, it could happen. The `serial` there is a nullable field. Ideally it should not happen, but if for some reason on the membership side we fail to persist the serial we could end up with `null` in theory.

BenYip123: Will the code attempt to re-register the vnode if the serial is null and what would happen in that case?

nikinagy: If serial is `null` we will try to perform a lookup for previous member information and predict the serial based on that. This can be found in `DynamicMemberRegistrationService`.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistRegistrationRequestHandler.kt
```diff
@@ -28,33 +28,41 @@ internal class PersistRegistrationRequestHandler(
                             " is already persisted. Persistence request was discarded.")
                     return@transaction
                 }
-                if (!it.canMoveToStatus(request.status)) {
+                // In case of processing persistence requests in an unordered manner we need to make sure the serial
+                // gets persisted. The existing status of the request won't be modified.
+                if (request.status == RegistrationStatus.SENT_TO_MGM) {
+                    em.merge(createEntity(request, currentStatus.status.toStatus()))
```
williamvigorr3: Should we be more defensive here and make sure only the serial has changed from null to non-null or incremented? Incase an old request gets replayed with status `RegistrationStatus.SENT_TO_MGM`.

vkolomeyko (dismissed):
LGTM

But I do not have a detailed knowledge of this code, therefore review from @dimosr would also be beneficial.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistRegistrationRequestHandler.kt
```diff
@@ -28,33 +28,41 @@ internal class PersistRegistrationRequestHandler(
...
+                    return@transaction
+                } else if (!it.canMoveToStatus(request.status)) {
```
dimosr: Should we do it the other way around to minimise potential side-effects from updating other fields apart from serial? e.g. if it's a backwards move *and* the new status is `SENT_TO_MGM` (and potentially the serial is null) then keep the existing request that's in the DB and only update the serial?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
```diff
@@ -366,13 +374,15 @@ class DynamicMemberRegistrationService @Activate constructor(
                     memberId.value
                 )
 
+                // This call will update the status and serial information, but the context will remain the same
+                // as it was before, because the memberContext column is not updatable.
```
dimosr: If I understand correctly, the reason we are doing this is not because the column is not updatable (which we could change), but the fact that we want to always store the user-provided context in this field going forward, right? 

P.S. IIRC, we also agreed to store the platform-transformed one in a separate column in the future, right? If so, can we raise a ticket so that we don't forget (and maybe we can link it here as part of the comment)?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
```diff
@@ -311,6 +311,14 @@ class DynamicMemberRegistrationService @Activate constructor(
 
                 val signedMemberContext = sign(memberId, publicKey, signatureSpec, memberContext)
                 val signedRegistrationContext = sign(memberId, publicKey, signatureSpec, registrationContext)
+                val signedRegistrationRequestMemberContext = sign(
+                    memberId,
+                    publicKey,
+                    signatureSpec,
+                    context.filterNot {
+                        it.key == SERIAL || REGISTRATION_CONTEXT_FIELDS.contains(it.key)
```
williamvigorr3, line 324: Do we need to also use hashes instead of the full `pem` keys?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
```diff
@@ -311,6 +311,14 @@ class DynamicMemberRegistrationService @Activate constructor(
 
                 val signedMemberContext = sign(memberId, publicKey, signatureSpec, memberContext)
                 val signedRegistrationContext = sign(memberId, publicKey, signatureSpec, registrationContext)
+                val signedRegistrationRequestMemberContext = sign(
```
williamvigorr3: Should we explain why this is different to `signedMemberContext`?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
```diff
@@ -366,13 +374,15 @@ class DynamicMemberRegistrationService @Activate constructor(
                     memberId.value
                 )
 
+                // This call will update the status and serial information, but the context will remain the same
```
williamvigorr3: This is only true if the other persistence has already happened. Might be worth calling out both cases here.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/mgm/MGMRegistrationRequestHandler.kt
```diff
@@ -47,17 +51,14 @@ internal class MGMRegistrationRequestHandler(
                 registrationId = registrationId.toString(),
                 requester = holdingIdentity,
                 memberContext = SignedData(
-                    ByteBuffer.wrap(mgmInfo.memberContextBytes),
-                    mgmInfo.memberSignature,
-                    mgmInfo.memberSignatureSpec
+                    ByteBuffer.wrap(serializedMemberContext),
```
dimosr, line 54: the fact that we don't have valid signatures here is a bit confusing (aware that some of it was in pre-existing code).

What is the reason for that? Is it for convenience given the MGM signs the data pre-data distribution, so these are not used (in which case, a subsequent question I have would be should fix this and populate the signatures properly with the MGM keys at some point)? Or something else?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/mgm/MGMRegistrationRequestHandler.kt
```diff
@@ -47,17 +51,14 @@ internal class MGMRegistrationRequestHandler(
...
+                    signatureWithKey,
```
williamvigorr3, line 55: So we no longer persist the signature/spec? Is this a problem?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/mgm/MGMRegistrationRequestHandler.kt
nikinagy: I don't know. Probably best to ask on our channel if someone remembers the reason.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
nikinagy: Here I meant that this is the second call for persistence, so you will only update the status and serial with this call. I can add something about the context to provide more info.
Plus: We can raise a ticket but I would prefer if we would discuss what's the plan there in long term before wildly opening any tickets. Eg we should think about what would be the best from migration POV, etc. I am not even sure what is the expected output here from user POV: do we want to store both network and rpc type everywhere? Do we need any changes on static side?

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
nikinagy: I don't get this question, because the registration context submitted by members will not have the pem keys. This filtering is also identical to the one used at MemberResourceClientImpl to make sure everything is identical.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
nikinagy: Ok, I will add everything here then

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/mgm/MGMRegistrationRequestHandler.kt
nikinagy: No, because we previously used dummies too. See code at MGMRegistrationMemberInfoHandler line 201 and no, I don't know why we are doing this.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
dimosr: ok, let's have these discussions then. I was under the impression we discussed that last week and we agreed that it would be sensible to store both pieces of data in the database, we wouldn't really have to do any migration for existing registration requests (that would have an empty column). But, if we need a discussion with a wider group, happy to have that. I was just reminded by this and trying to make sure we don't forget about the future plans.

nikinagy: I am still worried about the migration since on member side we have rpc type, on MGM side for MGM we have rpc type but for member we have network type.
I vaguely remember we were talking about adding two additional columns for these? And leave current one as is? Maybe that could work.

dimosr: The memory I have of our previous discussion is we settled on having 2 columns (one for user-provided context / one for platform-transformed one), for both MGM and member registrations. Once we do that, the existing column from MGM will switch from being the platform-transformed one to the user-provided one and we'd accept that as a compromise to avoid having to do a complicated DB migration (assuming it's not used anywhere internally so it shouldn't have any impact on platform logic, it's just that users would start seeing different data from the REST endpoints).

nikinagy: Right, that was my main worry, but if we are happy with that compromise then I am fine with this change. I will raise a ticket then.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistRegistrationRequestHandler.kt
nikinagy: addressed in [a2e6eb9](https://github.com/corda/corda-runtime-os/pull/4970/commits/a2e6eb9f3cfc61632dc1f21b972957a21bada0ca)

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
nikinagy: addressed in [a2e6eb9](https://github.com/corda/corda-runtime-os/pull/4970/commits/a2e6eb9f3cfc61632dc1f21b972957a21bada0ca)

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistRegistrationRequestHandler.kt
nikinagy: addressed in [a2e6eb9](https://github.com/corda/corda-runtime-os/pull/4970/commits/a2e6eb9f3cfc61632dc1f21b972957a21bada0ca)

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
nikinagy: Opened CORE-18001

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistRegistrationRequestHandler.kt
```diff
@@ -33,28 +33,60 @@ internal class PersistRegistrationRequestHandler(
                         "Registration request [$registrationId] has status: ${currentStatus.status}" +
                                 " can not move it to status ${request.status}"
                     )
+                    // In case of processing persistence requests in an unordered manner we need to make sure the serial
+                    // gets persisted. All other existing data of the request will remain the same.
+                    if (request.status == RegistrationStatus.SENT_TO_MGM && currentStatus.serial == null) {
+                        logger.info("Updating request [$registrationId] serial to ${currentStatus.serial}")
+                        em.merge(createEntityBasedOnPreviousEntity(currentStatus, request.registrationRequest.serial))
```
dimosr: I know it's existing code, but I'd suggest now renaming `currentStatus` to `currentRegistrationRequest` or something similar, since the request has a status itself so it can end up being confusing/misleading.

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistRegistrationRequestHandler.kt
```diff
@@ -28,30 +28,54 @@ internal class PersistRegistrationRequestHandler(
...
+                    }
                     return@transaction
                 }
             }
-            em.merge(createEntity(request, request.status))
+            em.merge(createEntityBasedOnRequest(request))
+        }
+    }
+
+    private fun createEntityBasedOnPreviousEntity(entity: RegistrationRequestEntity, newSerial: Long): RegistrationRequestEntity {
```
williamvigorr3: Very much a nit: should we rename `entity` to `previousEntity`?

williamvigorr3 (dismissed):
LGTM - I left a very minor comment.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
```diff
@@ -309,8 +309,13 @@ class DynamicMemberRegistrationService @Activate constructor(
...
                 val signatureSpec = memberContext.getFirst(SESSION_KEYS_SIGNATURE_SPEC)
 
+                // This is the user provided part of registration context with transformations. It will be the member
+                // provided context of the member information after successful registration.
                 val signedMemberContext = sign(memberId, publicKey, signatureSpec, memberContext)
+                // This is the context used during registration process, e.g. pre-auth tokens will be placed here.
                 val signedRegistrationContext = sign(memberId, publicKey, signatureSpec, registrationContext)
+                // This is the user provided part of registration context in its original form without any
+                // transformations.
                 val signedRegistrationRequestMemberContext = sign(
```
dimosr: would it be a bit more clear if we renamed these the following way:
* signedMemberContext -> signedPlatformTransformedMemberContext
* signedRegistrationRequestMemberContext -> signedUserProvidedMemberContext

dimosr (commented):
LGTM - just some minor comments on naming

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistRegistrationRequestHandler.kt
nikinagy: addressed in [d987623](https://github.com/corda/corda-runtime-os/pull/4970/commits/d987623dbddfd30b1a6632941c9913af859c60d9)

components/membership/membership-persistence-service-impl/src/main/kotlin/net/corda/membership/impl/persistence/service/handler/PersistRegistrationRequestHandler.kt
nikinagy: addressed in [d987623](https://github.com/corda/corda-runtime-os/pull/4970/commits/d987623dbddfd30b1a6632941c9913af859c60d9)

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
nikinagy: addressed in [d987623](https://github.com/corda/corda-runtime-os/pull/4970/commits/d987623dbddfd30b1a6632941c9913af859c60d9)

dimosr (approved):

williamvigorr3 (approved):
LGTM

vkolomeyko (approved):
LGTM

PR #4972, opened 2023-10-25
CORE-17976 - increase webserver max request payload

driessamyn:
Increase max request payload Javalin config

Also enable the debug logging (when configured) and integrate with tracing.

### DIFF
--- a/libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
+++ b/libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
@@ -5,6 +5,7 @@ import net.corda.libs.platform.PlatformInfoProvider
 import net.corda.lifecycle.LifecycleCoordinatorFactory
 import net.corda.lifecycle.LifecycleStatus
 import net.corda.lifecycle.createCoordinator
+import net.corda.tracing.configureJavalinForTracing
 import net.corda.utilities.classload.executeWithThreadContextClassLoader
 import net.corda.utilities.executeWithStdErrSuppressed
 import net.corda.web.api.Endpoint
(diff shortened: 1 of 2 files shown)


ben-millar (approved):
LGTM, but I'm not sure off-hand if there's a message size limitation on the client side.

thiagoviana (approved):
LGTM.

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
```diff
@@ -32,7 +33,21 @@ class JavalinServer(
...
         @Reference(service = PlatformInfoProvider::class)
         platformInfoProvider: PlatformInfoProvider,
-    ) : this(coordinatorFactory, { Javalin.create() }, platformInfoProvider)
+    ) : this(
+        coordinatorFactory,
+        {
+            Javalin.create { config ->
+                // hardcode to 100Mb for now
+                // TODO CORE-17986: make configurable
+                config.maxRequestSize = 100_000_000L
```
jujoramos, line 42: I'm not 100% sure this will work as it is, according to the docs [here](https://javalin.io/documentation#httpconfig):

> config.http.maxRequestSize = longValue;       // the max size of request body that can be accessed without using using an InputStream

Is the responsibility of the endpoint implementors to use the body as an `InputStream`, or is it handled automatically by Javalin?

Omar-awad (approved):
LGTM

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
driessamyn: I think what this means is that when you use inputstream the max doesn't apply, but I don't think we are.

I'm creating an integration test to prove this, but I think it's ok to merge this change as there shouldn't be a regression.

jujoramos: 👍

driessamyn: https://github.com/corda/corda-runtime-os/pull/4978

PR #4973, opened 2023-10-25
CORE-17963: add lock around sandbox

dickon:
The flow worker now calls sandbox code from multiple threads. This appears to cause problems in the sandbox code, so for now put a reentrant lock around all the interface methods of `SandboxGroupContextServiceImpl`. 

The only potentially issue is that there are callbacks out of the sandbox code when eviction happens, and they will hold the lock. However, this appears to be harmless.  Without this test, we saw large network test failures in the first few flows, and with this we are running 14,396 flows without issues.

In future we should review all the code used by the flow worker for thread safety.


dickon:
This large network test  https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2FLarge%20Network%20Tests/detail/PR-69/98/pipeline/ has got much further, @dimosr says we are good to proceed and are confident enough this helps.

simon-johnson-r3 (approved):

PR #4977, opened 2023-10-25
CORE-17958 restrict flow mapper state keys from exceeding the 255 character limit in the state storage db

LWogan:
The cllient requestid string is combined with the vnode shorthash to make a key for flow mapper states records for start flow events. The flow mapper state is stored in the state storage which has a limit of 255 characters. vnode shorthash is 12 characters with an additional separator character added this adds to the max 255.

Previously the JSON string representation of the FlowKey object was stored as the key. This easily exceeds 255 characters as seen by the solutions team.

API  PR: https://github.com/corda/corda-api/pull/1316


thiagoviana (approved):

jujoramos (commented):
LGTM

PR #4980, opened 2023-10-26
ES-1463: config update for automated slack messages

ronanbrowne:
Minor non-functional change to remove the automated tagging on slack notifications for the last committer in case of failure. This leads to excess noise due to Jenkins picking up committers from the shared library, which is also checked out, and a test failure may have nothing to do with the last committer.


wzur-r3 (approved):
LGTM

PR #4981, opened 2023-10-26
CORE-18016 rebalance in progress exception should be a transient issue not a fatal one

LWogan:
currently we are not handling this exception so it is treated as fatal. We should allow the consumer to call poll() to complete the rebalance


mbrkic-r3 (approved):
LGTM

ronanbrowne (approved):
LGTM

PR #4983, opened 2023-10-26
CORE-18013 - Guard against race condition in create a scheduled task.

driessamyn:


libs/scheduler/scheduler-datamodel/src/main/kotlin/net/corda/libs/scheduler/datamodel/db/TaskSchedulerLogEntityRepository.kt
```diff
@@ -22,10 +28,25 @@ class TaskSchedulerLogEntityRepository {
         readQuery.setParameter(TASK_SCHEDULER_LOG_QUERY_PARAM_NAME, taskName)
         readQuery.lockMode = LockModeType.PESSIMISTIC_WRITE
         return readQuery.resultList.singleOrNull()?:
-            TaskSchedulerLogEntity(taskName, schedulerId, Instant.MIN, Date.from(Instant.now()))
-                .also {
-                    em.persist(it)
+            // try to persist, but catch a constraint violation caused by a possible race condition
```
simon-johnson-r3, line 31: shouldn't we just wrap this section in
```
entityManager.transaction.begin()
...
entityManager.transaction.commit()
```
?

driessamyn: This is already in a transaction that is committed in `SchedulerLockImpl` (after it has updated the row)

libs/scheduler/scheduler-datamodel/src/integrationTest/kotlin/TaskSchedulerLogEntityTest.kt
```diff
@@ -137,6 +139,59 @@ class TaskSchedulerLogEntityTest {
...
     }
 
+    @Test
+    fun `duplicate insert throws ConstraintViolationException which can be handled in the TX`() {
+        val taskId = "foo${UUID.randomUUID()}"
+        val log = TaskSchedulerLogEntity(taskId, "bar")
+
+        emf.createEntityManager().transaction {
+            it.persist(log)
+            it.flush()
```
kyriathar, line 149: nit: this flush here is redundant as a flush will take place when committing this transaction.

dickon (approved):
LGTM

blsemo (approved):
LGTM

libs/scheduler/scheduler-datamodel/src/integrationTest/kotlin/TaskSchedulerLogEntityTest.kt
```diff
@@ -137,6 +139,59 @@ class TaskSchedulerLogEntityTest {
...
+                    throw e
+                }
+            }
+        }
+
+        assertThat(retry).isEqualTo(log)
+    }
+
+    @Test
+    fun `getOrInitialiseLog lazily creates entity`() {
```
kyriathar, line 176: Not really sure I understand _lazily creates entity_

kyriathar (commented):
As I mentioned offline I think I would be in favor of taking out of this processing the persisting of the scheduler task in the DB, and have it in some separate shorter lived DB transaction when for example creating a scheduler (with a _if not exists insert_). The current DB transaction here now looks a bit overstuffed because when the DB entry does not exist we have a transaction that contains both insert and then update on that entry. The DB transaction seems a bit long lived as well because in between we do things like publishing a record to Kafka as well.

libs/scheduler/scheduler-datamodel/src/integrationTest/kotlin/TaskSchedulerLogEntityTest.kt
driessamyn: I meant it tries to get the entity and creates it if one doesn't exist yet.

PR #4986, opened 2023-10-26
CORE-18017: do query and add under a DB transaction

dickon:
We have seen `TenantInfoService.populate` fail with a constraint violation, which is probably a race. The DB operation involves up to two reads and two writes, across two tables, so do it under a transaction and handle the constraint violation.

### DIFF
--- a/components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/HSMRepositoryImpl.kt
+++ b/components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/HSMRepositoryImpl.kt
@@ -11,6 +11,7 @@ import net.corda.orm.utils.transaction
 import net.corda.orm.utils.use
 import net.corda.v5.base.util.EncodingUtils.toHex
 import org.slf4j.LoggerFactory
+import java.lang.UnsupportedOperationException
 import java.time.Instant
 import java.util.UUID
 import javax.persistence.EntityManager
@@ -138,6 +139,7 @@ class HSMRepositoryImpl(
                 }
             }
         }
+        throw UnsupportedOperationException() // unreachable
     }
 
     private fun findHSMAssociationEntity(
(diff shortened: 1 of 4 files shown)


components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/HSMRepositoryImpl.kt
```diff
@@ -95,6 +97,51 @@ class HSMRepositoryImpl(
...
+                            }
+                        } else {
+                            logger.trace(
+                                "Reusing tenant category association $tenantId $category with wrapping key alias " +
+                                        tenantAssociation.masterKeyAlias
+                            )
+                            tenantAssociation
+                        }
+                    }
+                } catch (_: RollbackException) {
```
driessamyn: I think this needs to be more specific than a rollback exception as some exceptions you don't want to retry.
You don't even need to retry with a key constraint exception as you know that isn't going to fail the second time.
I also don't think the 1s sleep is wanted or necessary. There's nothing that will change in that 1 second.

driessamyn: See https://github.com/corda/corda-runtime-os/pull/4983/files for a similar change.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/HSMRepositoryImpl.kt
```diff
@@ -95,6 +97,51 @@ class HSMRepositoryImpl(
         }
     }
 
+    override fun createOrLookupCategoryAssociation(
+        tenantId: String,
+        category: String,
+        masterKeyPolicy: MasterKeyPolicy
+    ): HSMAssociationInfo {
+        entityManagerFactory.createEntityManager().use {
+            while (true) {
```
driessamyn: I don't think an infinite loop here is ok.
This would lock up/freeze the worker indefinitely in case of a genuine error/bug.

driessamyn (dismissed):
LGTM

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/HSMRepositoryImpl.kt
dickon: Agreed, done

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/HSMRepositoryImpl.kt
dickon: yes, removed

driessamyn (approved):

PR #4987, opened 2023-10-26
CORE-18014 Token selection failover edge condition

owenstanford:
When a token selection process request fails, due to a state concurrency check, we need to clear the available token cache as it could be invalid.


vkolomeyko (approved):
LGTM

driessamyn (approved):
LGTM

josephzunigadaly (approved):

PR #4989, opened 2023-10-26
CORE-11470: Clear operationInProgress when no DDL is provided

josephzunigadaly:
We expect the DBA to deploy schema updates when no DDL is provided. We need to clear our internal `operationInProgress` flag in this situation because it covers internal work, not the external work of the DBA.


josephzunigadaly:
The diff is clearer with the "Hide whitespace" option enabled.

vkolomeyko (approved):
LGTM

PR #4994, opened 2023-10-27
CORE-18025 Tactical fix for uniqueness checker state errors

relyafi:
### Summary

The uniqueness checker stores full error message details in the database to guarantee idempotency so that on replay of a request, the exact original error message can be returned to the client. The underlying database field is capped at 1024 bytes, and will throw an exception if this is exceeded as a failsafe mechanism.

State reference based errors are currently unbounded in size, as the capture each state that is unknown or in conflict. Further, these states are currently encoded inefficiently, which means we can currently only store details of a single conflicting state before hitting this issue.

This PR provides a tactical fix to this issue by limiting these errors to the first conflicting / unknown state that caused the error. The full set of states are written to the uniqueness processor log, so these can be interrogated by a notary operator for further information.

A strategic solution will follow this change in a later release (CORE-17155), looking at better encoding of the error message and subsequently increasing the number of states that can be reported.

### Testing

The common uniqueness assertion test library has been modified to check that only one of the conflicting states is present in any uniqueness checker state based error. This should be reverted when the strategic fix is made. Also added a new test case to prove only a single error is returned when multiple conflicting states exist in a single transaction.


Omar-awad (dismissed):
LGTM

blsemo (dismissed):
LGTM

blsemo (approved):
Re-approved after test fix

PR #4996, opened 2023-10-27
CORE-16149 Add inbound session timeout metric

charlieR3:
Add additional metric to capture the rate of inbound session timeouts.

_**NOTE: This currently targets the 5.1 branch but the change will not be merged to that branch. Waiting for the 5.2 branch but this can be reviewed against 5.1 in the meantime.**_

### DIFF
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManager.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManager.kt
@@ -15,8 +15,8 @@ internal interface SessionManager : LifecycleWithDominoTile {
     fun processSessionMessage(message: LinkInMessage): LinkOutMessage?
     fun inboundSessionEstablished(sessionId: String)
     fun messageAcknowledged(sessionId: String)
-    fun sessionMessageReceived(sessionId: String)
-    fun dataMessageReceived(sessionId: String)
+    fun sessionMessageReceived(sessionId: String, source: HoldingIdentity, destination: HoldingIdentity?)
+    fun dataMessageReceived(sessionId: String, source: HoldingIdentity, destination: HoldingIdentity)
 
     fun recordsForSessionEstablished(
         session: Session,
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManager.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManager.kt
@@ -15,7 +15,6 @@ internal interface SessionManager : LifecycleWithDominoTile {
     fun processSessionMessage(message: LinkInMessage): LinkOutMessage?
     fun inboundSessionEstablished(sessionId: String)
     fun messageAcknowledged(sessionId: String)
-    fun sessionMessageReceived(sessionId: String, source: HoldingIdentity, destination: HoldingIdentity?)
     fun dataMessageReceived(sessionId: String, source: HoldingIdentity, destination: HoldingIdentity)
 
     fun recordsForSessionEstablished(
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
@@ -339,7 +339,7 @@ internal class SessionManagerImpl(
         }
     }
 
-    override fun sessionMessageReceived(sessionId: String, source: HoldingIdentity, destination: HoldingIdentity?) {
+    fun sessionMessageReceived(sessionId: String, source: HoldingIdentity, destination: HoldingIdentity?) {
         dominoTile.withLifecycleLock {
             heartbeatManager.sessionMessageReceived(sessionId, source, destination)
         }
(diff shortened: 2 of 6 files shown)


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -1286,12 +1287,16 @@ internal class SessionManagerImpl(
...
+        private fun recordOutboundSessionTimeoutMetric(source: HoldingIdentity, destination: HoldingIdentity) {
             CordaMetrics.Metric.OutboundSessionTimeoutCount.builder()
                 .withTag(CordaMetrics.Tag.SourceVirtualNode, source.x500Name.toString())
                 .withTag(CordaMetrics.Tag.DestinationVirtualNode, destination.x500Name.toString())
                 .withTag(CordaMetrics.Tag.MembershipGroup, source.groupId)
                 .build().increment()
         }
+
+        private fun recordInboundSessionTimeoutMetric() {
+            CordaMetrics.Metric.InboundSessionTimeoutCount.builder().build().increment()
```
williamvigorr3: Is it worth recording the `source` and `destination` and `MembershipGroup`? In a large multi cluster network, we maybe expect one party to be problematic.  

To do this would have to refactor the code (as we only have this information for data messages). We could also store the time out metric differently for established and unestablished inbound sessions. So maybe it's not worth it.

williamvigorr3 (commented):
Code looks good. I left a comment.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
charlieR3: I tried to keep this simple and just report what's currently available in the metric but I'll have a look now and see how much of a refactor is required to store the additional info in the inbound sessions to support that.

charlieR3: Updated. In one session message case there was no access to the destination node so it is not populated.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManager.kt
```diff
@@ -15,8 +15,8 @@ internal interface SessionManager : LifecycleWithDominoTile {
     fun processSessionMessage(message: LinkInMessage): LinkOutMessage?
     fun inboundSessionEstablished(sessionId: String)
     fun messageAcknowledged(sessionId: String)
-    fun sessionMessageReceived(sessionId: String)
-    fun dataMessageReceived(sessionId: String)
+    fun sessionMessageReceived(sessionId: String, source: HoldingIdentity, destination: HoldingIdentity?)
```
williamvigorr3: I think we can remove this from the interface.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -338,15 +339,15 @@ internal class SessionManagerImpl(
         }
     }
 
-    override fun sessionMessageReceived(sessionId: String) {
+    override fun sessionMessageReceived(sessionId: String, source: HoldingIdentity, destination: HoldingIdentity?) {
```
williamvigorr3: This can be `private` (I think it it only used in this class).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -1286,12 +1290,20 @@ internal class SessionManagerImpl(
...
                 .withTag(CordaMetrics.Tag.SourceVirtualNode, source.x500Name.toString())
                 .withTag(CordaMetrics.Tag.DestinationVirtualNode, destination.x500Name.toString())
                 .withTag(CordaMetrics.Tag.MembershipGroup, source.groupId)
                 .build().increment()
         }
+
+        private fun recordInboundSessionTimeoutMetric(source: HoldingIdentity, destination: HoldingIdentity?) {
+            CordaMetrics.Metric.InboundSessionTimeoutCount.builder()
+                .withTag(CordaMetrics.Tag.SourceVirtualNode, source.x500Name.toString())
+                .withTag(CordaMetrics.Tag.DestinationVirtualNode, destination?.x500Name?.toString() ?: NOT_APPLICABLE_TAG_VALUE)
```
williamvigorr3, line 1304: `NOT_APPLICABLE_TAG_VALUE` I had a debate here about if we should have a specific UNKNOWN field, but concluded this is Okay. Maybe it is better not to add a specific one. So if you agree we can leave this as is.

williamvigorr3 (commented):
I left some of minor comments but looks good.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
charlieR3: I actually had a different value here originally. The problem was that the variable we use for destination virtual node takes possible values from the `corda_p2p_message_outbound_total` and `corda_p2p_message_inbound_total` metrics so for the destination filter to work we need to align the possible values with that so it was better to use the same NOT_APPLICABLE tag.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
charlieR3: The only other place it's used is in unit tests so it depends on whether we want to keep testing this function specifically or not. I could leave it public on the implementation if so since it won't be available through the interface

williamvigorr3: Good point. Let's leave it.

williamvigorr3 (approved):
LGTM

PR #4997, opened 2023-10-27
CORE-18000 - Capture the exception thrown by the state manager and abort the current batch of requests

filipesoliveira:


components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/services/PerformanceClaimStateStoreImpl.kt
```diff
@@ -69,9 +69,20 @@ class PerformanceClaimStateStoreImpl(
...
             )
 
-            val mismatchedState = stateManager.update(listOf(stateManagerState))
-                .map { it.value }
-                .firstOrNull()
+            val mismatchedState = try {
+               stateManager.update(listOf(stateManagerState))
+                    .map { it.value }
+                    .firstOrNull()
+            } finally {
```
vkolomeyko: I think the idea was to "capture the exception"... whereas `finally` block will be called always.
```suggestion
            } catch(ex: Exception) {
                logger.warn("Exception during execution of an update", ex)
```

vkolomeyko (approved):
LGTM

owenstanford (approved):
LGTM

filipesoliveira:
DON'T MERGE THIS PR FOR NOW.

PR #5006, opened 2023-10-30
DA5-8 MerkleTree Proof serializers and deserializers

szymonsztuka:
Add serializers and deserializers for MerkleTreeProof and other related classes. New 'InteropJsonMarshallingServiceImpl' implementing new `InteropJsonMarshallingService` uses those serializes. The serializer needs `MerkleProofImpl` which are getting via another service `MerkleProofProvider` / `MerkleProofProviderImplementation`.
The new serializers are needed to pass signatures  as arguments of an RPC Flow invocation and as a response from RPC Flow. For now the new service is created separately to existing one existing `JsonMarshallingServiceImpl` - the new service has the same code as existing one with extra `MerkleProof` etc, however the new service is only needed inside a flow. We could merge both services in one in the future.


libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/merkle/MerkleTreeProofProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+     * Creates a [MerkleProof]
+     *
+     * @param proofType
+     * @param treeSize
+     * @param leaves The leaves of the tree.
+     * @param hashes
+     *
+     * @return A new [MerkleProof] instance.
+     */
+    @Suspendable
```
vlajos: Are there any calls which suspend inside the implementations?

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/merkle/MerkleTreeProofProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.crypto.SecureHash
+import net.corda.v5.crypto.merkle.MerkleProofType
+import net.corda.v5.crypto.merkle.IndexedMerkleLeaf
+import net.corda.v5.crypto.merkle.MerkleProof
+
+/**
+ * [MerkleTreeProofProvider] creates [MerkleProof]s.
+ */
+@DoNotImplement
```
vlajos: I think this is only relevant for public APIs.

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/merkle/MerkleTreeProofProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import net.corda.v5.base.annotations.DoNotImplement
+import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.crypto.SecureHash
+import net.corda.v5.crypto.merkle.MerkleProofType
+import net.corda.v5.crypto.merkle.IndexedMerkleLeaf
+import net.corda.v5.crypto.merkle.MerkleProof
+
+/**
+ * [MerkleTreeProofProvider] creates [MerkleProof]s.
```
vlajos: So far, we used Merkle proof (without `tree`).
I think the terminology would be more consistent with removing `tree` from this interface and its implementation's name

libs/application/application-impl/src/main/kotlin/net/corda/application/impl/services/json/ProofOfActionSerializers.kt
```diff
@@ -0,0 +1,217 @@
...
+}
+
+class DigitalSignatureWithKeyIdSerializer : JsonSerializer<DigitalSignature.WithKeyId>() {
+    override fun serialize(
+        signature: DigitalSignature.WithKeyId?,
+        generator: JsonGenerator?,
+        provider: SerializerProvider
+    ) {
+        generator!!.writeStartObject()
+        generator.writeObjectField("by", signature!!.by) //TODO NPE
```
vlajos: Is this todo still relevant?

libs/application/application-impl/src/main/kotlin/net/corda/application/impl/services/json/ProofOfActionSerializers.kt
```diff
@@ -0,0 +1,217 @@
...
+    override fun deserialize(
+        parser: JsonParser,
+        ctxt: DeserializationContext
+    ): DigitalSignatureMetadata {
+        val node = parser.codec.readTree(parser) as JsonNode
+        val timestamp = Instant.parse(node.get("timestamp").asText())
+        val signatureSpec = parser.codec.treeToValue(node.get("signatureSpec"), SignatureSpec::class.java)
+        @Suppress("unchecked_cast")
+        val properties = parser.codec.treeToValue(node.get("properties"), Map::class.java)
+            .toMutableMap() as MutableMap<String, String>
```
vlajos: The AMQP serializer uses `SortedMap` in its internal container, and when it recreates the Object.
I do not know if it is needed, but it may be worth following that approach.
I think its purpose is to have a stable serialization output, which is needed to guarantee that the verification will verify the same data that the sign signed.

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/merkle/MerkleTreeProofProvider.kt
szymonsztuka: I've removed `@Suspendable`, we will see in the QA tests with the flows if it was needed.

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/merkle/MerkleTreeProofProvider.kt
szymonsztuka: Removed

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/merkle/MerkleTreeProofProvider.kt
szymonsztuka: Renamed to `MerkleProofProvider`/ `MerkleProofProviderImpl`

libs/application/application-impl/src/main/kotlin/net/corda/application/impl/services/json/ProofOfActionSerializers.kt
szymonsztuka: I change the method signature to have non-nullable parameters and  these checks are now gone.

libs/application/application-impl/src/main/kotlin/net/corda/application/impl/services/json/ProofOfActionSerializers.kt
szymonsztuka: I think it may stay as it is? This serialization is used for object to JSON / JSON to object only.

szymonsztuka: `TransactionSignatureServiceImpl` creates `DigitalSignatureMetadata` via `mapOf`, so nothing sorted.

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/merkle/MerkleTreeProofProvider.kt
vlajos: Thanks.

vlajos (dismissed):

vlajos (approved):

PR #5020, opened 2023-10-31
CORE-18088 Race condition registering RPC endpoints in JavalinServer

mbrkic-r3:
Ensuring that JavalinServer is thread-safe by changing the`Set` implementation for registered endpoints in `JavalinServer` to prevent threading issues when multiple clients register endpoint at the same time as well as synchronising Javalin server starts across the process in case of multiple servers.

### DIFF
--- a/libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
+++ b/libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
@@ -60,14 +60,15 @@ class JavalinServer(
     override val endpoints: MutableSet<Endpoint> = ConcurrentHashMap.newKeySet()
 
     override fun start(port: Int) {
-        check(null == server) { "The Javalin webserver is already initialized" }
-        coordinator.start()
-        startServer(port)
+        serverStartLock.withLock {
+            check(null == server) { "The Javalin webserver is already initialized" }
+            coordinator.start()
+            startServer(port)
+        }
     }
 
     private fun startServer(port: Int) {
         // Ensure the server can only be started once at a time.
-        serverStartLock.withLock {
             log.info("Starting Worker Web Server on port: $port")
             server = javalinFactory()
             endpoints.forEach {
@@ -91,7 +92,6 @@ class JavalinServer(
                 ctx.status(404)
             }
             coordinator.updateStatus(LifecycleStatus.UP)
-        }
     }
 
     private fun stopServer() {
@@ -100,18 +100,22 @@ class JavalinServer(
     }
 
     private fun restartServer() {
-        // restart server without marking the component down.
-        checkNotNull(server) { "Cannot restart a non-existing server" }
-        val port = server?.port()
-        stopServer()
-        checkNotNull(port) { "Required port is null" }
-        startServer(port)
+        serverStartLock.withLock {
+            // restart server without marking the component down.
+            checkNotNull(server) { "Cannot restart a non-existing server" }
+            val port = server?.port()
+            stopServer()
+            checkNotNull(port) { "Required port is null" }
+            startServer(port)
+        }
     }
 
     override fun stop() {
-        coordinator.updateStatus(LifecycleStatus.DOWN)
-        stopServer()
-        coordinator.stop()
+        serverStartLock.withLock {
+            coordinator.updateStatus(LifecycleStatus.DOWN)
+            stopServer()
+            coordinator.stop()
+        }
     }
 
     override fun registerEndpoint(endpoint: Endpoint) {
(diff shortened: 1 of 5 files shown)


driessamyn (dismissed):
LGTM

driessamyn:
I'm ok with this change is it's pretty harmless, but I'm 99% sure this will not fix the issue we saw.
This is because this log message:

```
2023-10-27T21:47:44.742261644Z stdout F {"instant":{"epochSecond":1698443264,"nanoOfSecond":741000000},"thread":"lifecycle-coordinator-0","level":"INFO","loggerName":"net.corda.web.server.JavalinServer","message":"Handler added to webserver: HandlerMetaInfo(httpMethod=POST, path=/api/5.1/ledger, handler=net.corda.web.server.JavalinServer$$Lambda$928/0x0000000801aaa000@3b3f335a, roles=[])","endOfBatch":false,"loggerFqcn":"org.apache.logging.slf4j.Log4jLogger","threadId":23,"threadPriority":5}
```

is generated by the Javalin registration handler, so won't be impacted by the thread safety of the Set that is maintained in `JavalinServer`

LWogan (dismissed):
LGTM

thiagoviana (dismissed):

vkolomeyko (dismissed):
LGTM

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
```diff
@@ -58,48 +63,76 @@ class JavalinServer(
...
-            server?.start(port)
-        } else {
-            // We temporarily switch the context class loader to allow Javalin to find `WebSocketServletFactory`.
-            executeWithThreadContextClassLoader(bundle.adapt(BundleWiring::class.java).classLoader) {
-                // Required because Javalin prints an error directly to stderr if it cannot find a logging
-                // implementation via standard class loading mechanism. This mechanism is not appropriate for OSGi.
-                // The logging implementation is found correctly in practice.
-                executeWithStdErrSuppressed {
-                    server?.start(port)
+//                JavalinStarter.startServer(
```
LWogan: we should remove this commented out code

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
```diff
@@ -58,48 +63,76 @@ class JavalinServer(
...
+//                    "RPC Server",
+//                    server!!,
+//                    port,
+//                )
+
+                val bundle = FrameworkUtil.getBundle(WebSocketServletFactory::class.java)
+                val name = "HTTP RPC"
+                if (bundle != null) {
+//                    val bundleList = threadLocalClassLoaderBundles.plus(bundle)
```
LWogan: remove

driessamyn: I'm just trying something out to drill down into the issue, please ignore the last few commits.

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
```diff
@@ -68,38 +66,32 @@ class JavalinServer(
...
     private fun startServer(port: Int) {
-        log.info("Starting Worker Web Server on port: $port")
-        server = javalinFactory()
-        endpoints.forEach {
-            registerEndpointInternal(it)
-        }
+        // Ensure the server can only be started once at a time.
+        serverStartLock.withLock {
+            log.info("Starting Worker Web Server on port: $port")
+            server = javalinFactory()
```
LWogan, line 73: The startServer method is protected by this lock, but I think the stopServer() could be executed concurrently if two threads call restartServer(). This could mean that `server ` gets set to null after this point. There is a checkNotNull() as part of `registerEndpointInternal(it)` which would throw an IllegalStateException and bubble up unexpectedly, and would probably cause the worker to move to an ERROR state and be bounced by k8s.

This scenario could happen on the persistence worker which has 2 endpoints/SyncRPCSubscriptions. Its a bit of an edge case. If a dependent service causes the parent lifecycle coordinators, LedgerPersistenceService & FlowPersistenceService to receive a RegistrationStatusChangeEvent of DOWN, then they will both restart the http subscriptions and will both call restartServer().

As an aside (and I think its already a known issue), the "green path" of this use case would cause the JavalinServer to bounce 4 times. Twice for closing the 2 subs, and twice more for starting new ones.

driessamyn: Good point.

I move the lock to the public start & stop methods and also the restart method.
We just need to be careful this doesn't create a potential deadlock scenario, but I think this should be ok.

LWogan (approved):
LGTM

thiagoviana (approved):

PR #5021, opened 2023-10-31
CORE-18099 -  Fix Fiber Cache

lankydan:
The suspend count in the current implementation of the Fiber cache does not work correctly meaning that all checkpoints end up being deserialised.
This changes fixes this issue by ensuring the suspension count is always updated before the cache is updated, and that the correct suspension count is added to the cache value.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
@@ -72,7 +72,17 @@ class FlowFiberCacheImpl @Activate constructor(
     }
 
     override fun get(key: FlowKey, suspendCount: Int): FlowFiber? {
-        return cache.getIfPresent(key)?.takeIf { it.suspendCount == suspendCount }?.fiber
+        val fiber = cache.getIfPresent(key)
+        return if(null == fiber) {
+            logger.info("Fiber not found in cache: ${key.id}")
+            null
+        } else if (fiber.suspendCount == suspendCount) {
+            logger.debug { "Fiber found in cache: ${key.id}" }
+            fiber.fiber
+        } else {
+            logger.info("Fiber found in cache but at wrong suspendCount (${fiber.suspendCount} <-> $suspendCount): ${key.id}")
+            null
+        }
     }
 
     override fun remove(key: FlowKey) {
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
@@ -80,7 +80,7 @@ class FlowFiberCacheImpl @Activate constructor(
             logger.debug { "Fiber found in cache: ${key.id}" }
             fiber.fiber
         } else {
-            logger.info("Fiber found in cache but at wrong suspendCount (${fiber.suspendCount} <-> $suspendCount): ${key.id}")
+            logger.warn("Fiber found in cache but at wrong suspendCount (${fiber.suspendCount} <-> $suspendCount): ${key.id}")
             null
         }
     }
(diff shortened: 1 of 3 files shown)


driessamyn:
I don't think this change is sufficient as the cache needs to be updated too.

I changed the `get` method on the cache to this to add some logging:

```kotlin
override fun get(key: FlowKey, suspendCount: Int): FlowFiber? {
        val fibre = cache.getIfPresent(key)
        return if(null == fibre) {
            logger.info("Fibre not found in cache: ${key.id}")
            null
        } else if (fibre.suspendCount == suspendCount) {
            logger.info("Fibre found in cache: ${key.id}")
            fibre.fiber
        } else {
            logger.info("Fibre found in cache but at wrong suspendCount (${fibre.suspendCount} <-> $suspendCount): ${key.id}")
            null
        }

//        return cache.getIfPresent(key)?.takeIf { it.suspendCount == suspendCount }?.fiber
    }
```

Running this in the combined worker confirms that it never hits the cache, only messages like these are logged:

```
Fibre found in cache but at wrong suspendCount (2 <-> 3): 82130e29-14e9-4fbc-92cb-901145b995b4-INITIATED
...
Fibre found in cache but at wrong suspendCount (5 <-> 6): 82130e29-14e9-4fbc-92cb-901145b995b4-INITIATED
```

Added this commit which may fix this order: https://github.com/corda/corda-runtime-os/pull/5021/commits/a8409cd7298b7627bee0725321a769d8898835a4

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
```diff
@@ -72,7 +72,17 @@ class FlowFiberCacheImpl @Activate constructor(
...
-        return cache.getIfPresent(key)?.takeIf { it.suspendCount == suspendCount }?.fiber
+        val fiber = cache.getIfPresent(key)
+        return if(null == fiber) {
+            logger.info("Fiber not found in cache: ${key.id}")
+            null
+        } else if (fiber.suspendCount == suspendCount) {
+            logger.debug { "Fiber found in cache: ${key.id}" }
+            fiber.fiber
+        } else {
+            logger.info("Fiber found in cache but at wrong suspendCount (${fiber.suspendCount} <-> $suspendCount): ${key.id}")
```
jujoramos: Should this be logged under `WARN` level instead?.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
```diff
@@ -72,7 +72,17 @@ class FlowFiberCacheImpl @Activate constructor(
     }
 
     override fun get(key: FlowKey, suspendCount: Int): FlowFiber? {
-        return cache.getIfPresent(key)?.takeIf { it.suspendCount == suspendCount }?.fiber
+        val fiber = cache.getIfPresent(key)
+        return if(null == fiber) {
+            logger.info("Fiber not found in cache: ${key.id}")
```
jujoramos, line 77: Should this be logged under `DEBUG` level instead?.

jujoramos (dismissed):

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
lankydan: I feel the same.

lankydan:
Generally LGTM, but technically this is my PR so I can't approve it

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
driessamyn: I think we should leave it as info for the time being, at least until we've flushed out all the issues around deserialisation. This should still only log when there's a cache miss, which should be very rare considering the size of the cache.

components/flow/flow-service/src/main/kotlin/net/corda/flow/fiber/cache/impl/FlowFiberCacheImpl.kt
driessamyn: Updated

thiagoviana (approved):

jujoramos (approved):

PR #5024, opened 2023-11-01
CORE-17766: config redaction on arrays

dickon:
Arrange to replace secrets in arrays as well as objects, by pulling out the logic.

Supplement the SmartConfig implementation to implement config lists.

Add some tests.


anien (approved):
Looks good to me.

simon-johnson-r3 (approved):

PR #5025, opened 2023-11-01
DA5-8 Unit tests for MerkleProof JSON serialization

szymonsztuka:
Added test for `DigitalSignatureAndMetadataSerialisationModule` (renamed now from `ProofOfActionSerialisationModule`) which was created in https://github.com/corda/corda-runtime-os/commit/aa29e223177de22823146cf3950ec988698403ac. 
The test requires a transaction signature with a `MerkleProof`. This type of signature can be produced by the base test class `UtxoLedgerTest` available for use in `components/ledger/ledger-utxo-flow` module where the test is placed.
The test requires a service implementation of `MerkleProofProvider` which is injected into the SUT, the implementation is in `libs/crypto/merkle-impl` hence with Gradle dependency `testImplementation project(':libs:crypto:merkle-impl')`.
The actual test the class `DigitalSignatureAndMetadataSerialisationModule` is from `libs/application/application-impl`, so the test is not placed in the same module as the class.
There are 2 reasons: `application-impl` module has no test dependencies on either `testing/ledger/ledger-utxo-base-test` or `libs/crypto/merkle-impl`,  while adding it to `components/ledger/ledger-utxo-flow` requires addtion of the second latter dependency only.
The second reason is that if the test is run from withing a module other than `components/ledger` then it throws an exception that Sandbox cannot be mocked.
The other placing of the test could me moved later, however putting it inside `components/ledger/ledger-utxo-flow` looked as reasonable compromise.

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/serialization/ProofOfActionSerializationTests.kt
+++ b/components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/serialization/ProofOfActionSerializationTests.kt
@@ -2,8 +2,15 @@ package net.corda.ledger.utxo.serialization
 
 import com.fasterxml.jackson.core.JsonGenerator
 import com.fasterxml.jackson.core.JsonParser
-import com.fasterxml.jackson.databind.*
+import com.fasterxml.jackson.databind.MapperFeature
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.DeserializationFeature
+import com.fasterxml.jackson.databind.JsonSerializer
+import com.fasterxml.jackson.databind.JsonDeserializer
+import com.fasterxml.jackson.databind.SerializationFeature
+import com.fasterxml.jackson.databind.SerializerProvider
 import com.fasterxml.jackson.databind.exc.InvalidDefinitionException
+import com.fasterxml.jackson.databind.exc.MismatchedInputException
 import com.fasterxml.jackson.databind.json.JsonMapper
 import com.fasterxml.jackson.databind.module.SimpleModule
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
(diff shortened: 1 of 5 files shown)


components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/serialization/ProofOfActionSerializationTests.kt
```diff
@@ -0,0 +1,153 @@
...
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.Assertions.assertNotNull
+import org.junit.jupiter.api.Assertions.assertThrowsExactly
+import org.junit.jupiter.api.Test
+import java.util.*
+import kotlin.test.assertContains
+
+
+class ProofOfActionSerializationTests : UtxoLedgerTest() {
```
szymonsztuka: We could consider name `DigitalSignatureAndMetadataSerializationTest` as well as this is what `ProofOfActionSerialisationModule` does. Maybe the module could be renames as well. From interopability business perspective `ProofOfAction` term makes more sense as this is why `DigitalSignatureAndMetadata` is used in a swap, however for the wider platform context this can be meaningless.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/serialization/ProofOfActionSerializationTests.kt
```diff
@@ -0,0 +1,153 @@
...
+
+        private val mapperWithoutProofOfActionModule =
+            JsonMapper.builder().enable(MapperFeature.BLOCK_UNSAFE_POLYMORPHIC_BASE_TYPES).build().apply {
+                registerModule(KotlinModule.Builder().build())
+                registerModule(JavaTimeModule())
+                enable(DeserializationFeature.FAIL_ON_READING_DUP_TREE_KEY)
+                setTimeZone(TimeZone.getTimeZone("UTC"))
+                disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
+                registerModule(standardTypesModule())
+                // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
vlajos: I think these excalamation marks could be removed.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/serialization/ProofOfActionSerializationTests.kt
vlajos: And the rename looks reasonable to me.

vlajos (approved):

PR #5028, opened 2023-11-01
CORE-18211 - State manager concurrent optimistic locking tests

conal-smith:
**Problem**

Hidden failure due to assertion on different thread from the main test thread.

**Solution**

- Move assertion to the main test thread.
- Add separate test that uses multi-threading to test optimistic locking behaviour across threads


libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -267,39 +268,30 @@ class StateManagerIntegrationTest {
         val conflictingKeys = (1..totalCount).filter { it % 2 == 0 }.map { buildStateKey(it) }
         val persistedStates = stateManager.get(allKeys)
 
-        val latch = CountDownLatch(1)
```
sprintengine, line 270: This test may or may not still be valid, it covers scenario where the same thread tries updating states sequentially, but mismatched version. Obviously the new test repeats this scenario with concurrent threads.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -267,39 +268,30 @@ class StateManagerIntegrationTest {
...
-                statesToUpdateSecondThread.add(
-                    State(state.key, "u2_$it".toByteArray(), state.version, metadata("u2" to it))
-                )
-            }
+        assertThat(stateManager.update(statesToUpdateFirstThread)).isEmpty()
 
-            latch.await()
-            val failedUpdates = stateManager.update(statesToUpdateSecondThread)
-            assertThat(failedUpdates).containsOnlyKeys(conflictingKeys)
-            assertThat(failedUpdates).containsValues(
```
sprintengine, line 296: JUnit test assertions must be in the main thread, so failures here were being missed (although the assertions was wrong to begin with)

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/TestHelper.kt
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Example: a list of [a, b, c, d, e, f, g, h, i] split among 3 threads with 1 shared states per thread results in:
+ * [[a, b, c, d], [d, e, f, g], [g, h, i, a]]
+ *
+ * @param states a list of states to be divided
+ * @param numThreads the number of thread groups to divide the states among
+ * @param sharedStatesPerThread the number of states to be shared between thread groups, only one thread can successfully
+ *  update a state with a given version.
+ */
+fun divideStatesBetweenThreads(
```
sprintengine: `windowed` didn't quite fit here, had to write my own function so the last group had element from the first.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -309,6 +300,48 @@ class StateManagerIntegrationTest {
...
+                updaterJob(groupsOfStates[threadGroup])
+            }
+        })
+
+        val failedKeysPerThread = futures.map { it.get() }
+
+        val expectedFailedKeys = sharedStatesPerThread * numThreads
+        assertThat(failedKeysPerThread.flatten())
+            .hasSize(expectedFailedKeys)
+            .withFailMessage("Expected one failure for every state shared between another thread")
```
jujoramos: Would it be possible to somehow get the list of actual failed keys vs expected failed keys here?, otherwise the assertion itself won't be much useful when diagnosing issues.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -384,6 +409,26 @@ class StateManagerIntegrationTest {
...
+
+        val allKeys = (1..totalCount).map { buildStateKey(it) }
+        val persistedStates = stateManager.get(allKeys)
+
+        val failedKeysFirstDelete = stateManager.delete(persistedStates.values)
+        assertThat(failedKeysFirstDelete).isEmpty()
+
+        val failedKeysSecondDelete = stateManager.delete(persistedStates.values)
+        assertThat(failedKeysSecondDelete).isEmpty()
+    }
```
jujoramos: Do need this test at all?.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -309,6 +300,48 @@ class StateManagerIntegrationTest {
         )
     }
 
+    @Test
+    fun `optimistic locking ensures no double updates across threads`() {
```
jujoramos, line 307: Worth adding the same tests for the delete operation?.

sprintengine: We can't really make an assertion in the delete case, because we don't receive back the entities that failed to delete. 

That's because our delete query is `DELETE FROM $STATE_MANAGER_TABLE s WHERE s.$KEY_COLUMN = ? AND s.$VERSION_COLUMN = ?` - maybe it would be better to return the keys that were deleted.

If we have the multi-threaded optimistic locking test all we can assert is the states were deleted (by someone) and there were no failures (which is probably still worthwhile).

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
sprintengine: I'll convert this to be multi-threaded

jujoramos: I was asking whether we need a test to check that delete is idempotent, actually; is it needed?.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
jujoramos: The delete operation returns the list of states that failed to be deleted due to optimistic locking so yeah, I'd say is worth having a "concurrent" test for it as well.

sprintengine: Ah I was wrong, I forgot we query to get the failed deletes! Yeah definitely worth having this test so

sprintengine: If an update fails because the state was already deleted, we do not return this as a "failed update", which is most likely a bug IMO.

We ought to return something to indicate an update failed because the state was deleted, i.e. the State "nullable" in the response map.

```
fun update(states: Collection<State>): Map<String, State?>
```

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
sprintengine: I've removed this one

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
sprintengine: done

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -58,9 +58,10 @@ interface StateManager : Lifecycle {
      *
      * @param states Collection of states to be updated.
      * @return Map with the most up-to-date version of the states, associated by key for easier access, that failed
-     *      the optimistic locking check.
+     *      the optimistic locking check. If this state failed to be updated because the key was deleted the key is
+     *      associated with null.
      */
-    fun update(states: Collection<State>): Map<String, State>
+    fun update(states: Collection<State>): Map<String, State?>
```
sprintengine: This allows caller to know if an update failed because a state was already deleted.

@jujoramos let me know if it was intentional to hide this from the caller, I suspect there's no reason why we shouldn't let the caller know a state was attempted to be updated but it was deleted.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
```diff
@@ -223,13 +223,18 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
...
+                val failedToUpdateStateDoesNotExist = (failedToUpdate - failedToUpdateOptimisticLockFailure).map { it.key }
+
+                states = failedToCreate + failedToDelete + failedToUpdateOptimisticLockFailure
                 groups = if (states.isNotEmpty()) {
                     allocateGroups(flowEvents.filterKeys { states.containsKey(it) }.values.flatten())
                 } else {
                     listOf()
                 }
                 states.keys.forEach { asynchronousOutputs.remove(it) }
+                failedToUpdateStateDoesNotExist.forEach { asynchronousOutputs.remove(it) }
```
sprintengine: @JamesHR3 let me know what you think here, AFAIK if the state manager tries to update a State but the checkpoint didn't exist, we should be able to ignore it, and remove any async outputs for that state.

jujoramos: I don't think these changes belong to this `PR`, they should be tackled in another ticket.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
jujoramos: I don't think these changes belong to this PR, they should be tackled in another ticket.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -71,26 +71,36 @@ class StateManagerImpl(
...
+
+        logger.warn(
+            "Optimistic locking check failed while updating States ${optimisticLockingFailedUpdates.keys.joinToString()}" +
+                    if (deletedStateKeys.isNotEmpty()) " Failed to update the following States because they were already deleted: ${deletedStateKeys.joinToString()}"
+                    else ""
+        )
+
+        return optimisticLockingFailedUpdates + deletedStateKeys.associateWith { null }
+    }
+
```
jujoramos: I don't think these changes belong to this PR, they should be tackled in another ticket.

libs/state-manager/state-manager-db-impl/src/test/kotlin/net/corda/libs/statemanager/impl/StateManagerImplTest.kt
```diff
@@ -99,6 +99,26 @@ class StateManagerImplTest {
...
+        val result = stateManager.update(listOf(apiStateOne, apiStateTwo, apiStateThree))
+        assertThat(result).containsExactly(
+            entry(persistedStateTwo.key, persistedStateTwo.toState()),
+            entry(persistentStateThree.key, null)
+        )
+        verify(stateRepository).get(connection, listOf(apiStateTwo.key, apiStateThree.key))
+        verify(stateRepository).update(connection, listOf(persistentStateOne, persistentStateTwo, persistentStateThree))
+        verifyNoMoreInteractions(stateRepository)
+    }
+
```
jujoramos: I don't think these changes belong to this PR, they should be tackled in another ticket.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -267,39 +268,30 @@ class StateManagerIntegrationTest {
...
 
-        updater1.join()
-        updater2.join()
+        val failedUpdates = stateManager.update(statesToUpdateSecondThread.values)
+        assertThat(failedUpdates).containsOnlyKeys(conflictingKeys)
+        assertSoftly {
+            failedUpdates.values.map { state ->
+                // assert the real version has bumped by one
+                it.assertThat(state!!.version).isEqualTo(statesToUpdateSecondThread[state.key]!!.version + 1)
+            }
```
jujoramos: Can we assert that the entire state is the expected one and not just the version?.

jujoramos (changes requested):
Tests look good, but the part related to the behavioural changes shouldn't be part of this `PR`; we need a different ticket for it.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -73,12 +73,15 @@ object DbUtils {
      * @param dbPassword DB password. If value is not provided, value of the system property "postgresPassword" is used.
      *                   If system property is not set then value "password" is used
      */
+    @JvmOverloads
```
jujoramos: Do we need this extra annotation here?, are we invoking the `DbUtils` utility methods from Java anywhere?.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/DbUtils.kt
```diff
@@ -73,12 +73,15 @@ object DbUtils {
...
+    @Suppress("LongParameterList")
     fun createPostgresDataSource(
         dbUser: String? = null,
         dbPassword: String? = null,
         schemaName: String? = null,
         createSchema: Boolean = false,
-        rewriteBatchedInserts: Boolean = false
+        rewriteBatchedInserts: Boolean = false,
+        maxJdbcConnections: Int = 5
```
jujoramos: Nitpick: maybe call it maximumPoolSize for consistency?.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/MultiThreadedTestHelper.kt
```diff
@@ -0,0 +1,143 @@
...
+     */
+    data class StateGrouping(
+        /**
+         * A list of States assigned to this thread, not including the overlapping states to test optimistic locking.
+         */
+        val assignedStates: List<State>,
+        /**
+         * The list of States in this group that overlap with the next group, for assertions on failed keys.
+         */
+        val statesOverlappingNextGroup: List<State>
```
jujoramos: Nitpick: `overlappingStates`.

jujoramos (dismissed):

jujoramos (dismissed):

jujoramos (approved):

PR #5032, opened 2023-11-02
CORE-18147: Fix State Manager Querying

jujoramos:
Check that the filters used by the calling API is not empty and return
immediately to prevent unnecessary interactions with the database.


Omar-awad (approved):
LGTM

conal-smith (commented):
LGMT

PR #5038, opened 2023-11-03
CORE-18144 Add LinkedKeySet serializer to Kryo

Omar-awad:
### Problem description

Flows will occasionally fail with a stack trace within Kryo serialization indicating that the `this.this$0` pointer was `null`. This is caused by the flow stack containing a reference to a `LinkedKeySet`, which Kryo cannot serialize (or deserialize) correctly with its default serializers.

### Solution

Add a custom Kryo serializer handling `LinkedKeySet`. Note that this needs to preserve the reference to the map that it was generated from in order to fulfil the API requirements of `LinkedKeySet`. This is done by asking Kryo to serialize the map instead of the key set, and relying on Kryo's internal reference handling to ensure references are correctly restored in the deserialized object graph.

### Testing

Unit tests, plus performance test runs showing the original problem is resolved.

### Other changes

- Removed an unused `FlowIORequest`.

### DIFF
--- a/charts/corda/values.yaml
+++ b/charts/corda/values.yaml
@@ -92,6 +92,10 @@ metrics:
       - "corda_p2p_session_outbound_total"
       - "corda_p2p_session_inbound_total"
       - "corda_membership_actions_handler_time_seconds_(count|sum|max)"
+      - "corda_consumer_poll_time_seconds_(count|sum|max)"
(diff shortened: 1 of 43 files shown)


jujoramos (approved):

lankydan (approved):

vkolomeyko (approved):
LGTM

PR #5042, opened 2023-11-03
CORE-17972: implementation of send & receive tx

jennyang-r3:
## Send and Receive Transaction
This PR implements a feature to allow sending transaction and receiving it.
I added APIs and flows for:
- `sendTransaction()`
- `receiveTransaction()`
The transaction is verified + backchain since transaction sent is expected to be `VERIFIED` transaction.

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
@@ -1,0 +1,67 @@
+package net.corda.ledger.utxo.flow.impl.flows.transactiontransmission
+
+import net.corda.ledger.common.data.transaction.TransactionStatus
+import net.corda.ledger.utxo.flow.impl.flows.finality.getVisibleStateIndexes
+import net.corda.ledger.utxo.flow.impl.persistence.UtxoLedgerPersistenceService
+import net.corda.ledger.utxo.flow.impl.transaction.UtxoSignedTransactionInternal
+import net.corda.ledger.utxo.flow.impl.transaction.verifier.TransactionVerificationException
+import net.corda.ledger.utxo.flow.impl.transaction.verifier.UtxoLedgerTransactionVerificationService
+import net.corda.v5.ledger.utxo.VisibilityChecker
+import net.corda.sandbox.CordaSystemFlow
+import net.corda.v5.application.flows.CordaInject
+import net.corda.v5.application.flows.FlowEngine
+import net.corda.v5.application.flows.SubFlow
+import net.corda.v5.application.messaging.FlowSession
+import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.ledger.utxo.transaction.UtxoSignedTransaction
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+@CordaSystemFlow
+class ReceiveTransactionFlow(
+    private val session: FlowSession
+) : SubFlow<UtxoSignedTransaction> {
+
+    private companion object {
+        private val log: Logger = LoggerFactory.getLogger(ReceiveTransactionFlow::class.java)
+    }
+
+    @CordaInject
+    lateinit var transactionVerificationService: UtxoLedgerTransactionVerificationService
+
+    @CordaInject
+    lateinit var visibilityChecker: VisibilityChecker
+
+    @CordaInject
+    lateinit var persistenceService: UtxoLedgerPersistenceService
+
+    @CordaInject
+    lateinit var flowEngine: FlowEngine
+
+    @Suspendable
+    override fun call(): UtxoSignedTransaction {
+        val transaction = session.receive(UtxoSignedTransaction::class.java)
+        verifyTransaction(transaction)
+        persistTransaction(transaction)
+
+        return transaction
+    }
+
+    @Suspendable
+    private fun persistTransaction(transaction: UtxoSignedTransaction) {
+        val visibleStatesIndexes = (transaction as UtxoSignedTransactionInternal).getVisibleStateIndexes(visibilityChecker)
+        persistenceService.persist(transaction, TransactionStatus.VERIFIED, visibleStatesIndexes)
+        if (log.isDebugEnabled) {
+            log.debug("Recorded transaction = ${transaction.id} since transaction is verified")
+        }
+    }
+
+    @Suspendable
+    private fun verifyTransaction(signedTransaction: UtxoSignedTransaction) {
+        try {
+            transactionVerificationService.verify(signedTransaction.toLedgerTransaction())
+        } catch(e: TransactionVerificationException){
+            throw e
+        }
+    }
+}
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
@@ -1,13 +1,14 @@
 package net.corda.ledger.utxo.flow.impl.flows.transactiontransmission
 
-import net.corda.ledger.common.data.transaction.TransactionStatus
-import net.corda.ledger.utxo.flow.impl.flows.finality.getVisibleStateIndexes
-import net.corda.ledger.utxo.flow.impl.persistence.UtxoLedgerPersistenceService
+import net.corda.ledger.common.flow.flows.Payload
+import net.corda.ledger.utxo.flow.impl.flows.backchain.InvalidBackchainException
+import net.corda.ledger.utxo.flow.impl.flows.backchain.TransactionBackchainResolutionFlow
+import net.corda.ledger.utxo.flow.impl.flows.backchain.dependencies
 import net.corda.ledger.utxo.flow.impl.transaction.UtxoSignedTransactionInternal
-import net.corda.ledger.utxo.flow.impl.transaction.verifier.TransactionVerificationException
 import net.corda.ledger.utxo.flow.impl.transaction.verifier.UtxoLedgerTransactionVerificationService
-import net.corda.v5.ledger.utxo.VisibilityChecker
 import net.corda.sandbox.CordaSystemFlow
+import net.corda.utilities.trace
+import net.corda.v5.application.crypto.DigitalSignatureAndMetadata
 import net.corda.v5.application.flows.CordaInject
 import net.corda.v5.application.flows.FlowEngine
 import net.corda.v5.application.flows.SubFlow
@@ -30,37 +31,41 @@ class ReceiveTransactionFlow(
     lateinit var transactionVerificationService: UtxoLedgerTransactionVerificationService
 
     @CordaInject
-    lateinit var visibilityChecker: VisibilityChecker
-
-    @CordaInject
-    lateinit var persistenceService: UtxoLedgerPersistenceService
-
-    @CordaInject
     lateinit var flowEngine: FlowEngine
 
     @Suspendable
     override fun call(): UtxoSignedTransaction {
-        val transaction = session.receive(UtxoSignedTransaction::class.java)
-        verifyTransaction(transaction)
-        persistTransaction(transaction)
+        val transaction = session.receive(UtxoSignedTransactionInternal::class.java)
+        try {
+            transaction.verifySignatorySignatures()
+            transaction.verifyAttachedNotarySignature()
+            transactionVerificationService.verify(transaction.toLedgerTransaction())
+        } catch (e: Exception) {
+            val message = "Failed to verify transaction and signatures of transaction: ${transaction.id}"
+            log.warn(message, e)
+            session.send(Payload.Failure<List<DigitalSignatureAndMetadata>>(message))
+            throw e
+        }
+
+        val transactionDependencies = transaction.dependencies
+        if (transactionDependencies.isNotEmpty()) {
+            try {
+                flowEngine.subFlow(TransactionBackchainResolutionFlow(transactionDependencies, session))
+            } catch (e: InvalidBackchainException) {
+                val message = "Invalid transaction found during back-chain resolution, marking transaction with ID " +
+                        "${transaction.id} as invalid."
+                log.warn(message, e)
+                session.send(Payload.Failure<List<DigitalSignatureAndMetadata>>(message))
+                throw e
+            }
+        } else {
+            log.trace {
+                "Transaction with id ${transaction.id} has no dependencies so backchain resolution will not be performed."
+            }
+        }
+
+        session.send(Payload.Success("Successfully received transaction."))
 
         return transaction
-    }
-
-    @Suspendable
-    private fun verifyTransaction(signedTransaction: UtxoSignedTransaction) {
-        try {
-            transactionVerificationService.verify(signedTransaction.toLedgerTransaction())
-        } catch(e: TransactionVerificationException) { throw e }
-    }
-
-    @Suspendable
-    private fun persistTransaction(transaction: UtxoSignedTransaction) {
-        val visibleStatesIndexes = (transaction as UtxoSignedTransactionInternal).getVisibleStateIndexes(visibilityChecker)
-        log.info("xxx visibleStatesIndexes = $visibleStatesIndexes")
-        persistenceService.persist(transaction, TransactionStatus.VERIFIED, visibleStatesIndexes)
-        if (log.isDebugEnabled) {
-            log.debug("Recorded transaction = ${transaction.id} since transaction is verified")
-        }
     }
 }
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
@@ -75,8 +75,8 @@ class ReceiveTransactionFlow(
             throw CordaRuntimeException(message, e)
         }
 
-        session.send(Payload.Success("Successfully received transaction."))
         ledgerPersistenceService.persist(transaction, VERIFIED, transaction.getVisibleStateIndexes(visibilityChecker))
+        session.send(Payload.Success("Successfully received transaction."))
 
         return transaction
     }
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
@@ -48,20 +48,19 @@ class ReceiveTransactionFlow(
     }
 
     @Suspendable
-    private fun persistTransaction(transaction: UtxoSignedTransaction) {
-        val visibleStatesIndexes = (transaction as UtxoSignedTransactionInternal).getVisibleStateIndexes(visibilityChecker)
-        persistenceService.persist(transaction, TransactionStatus.VERIFIED, visibleStatesIndexes)
-        if (log.isDebugEnabled) {
-            log.debug("Recorded transaction = ${transaction.id} since transaction is verified")
-        }
-    }
-
-    @Suspendable
     private fun verifyTransaction(signedTransaction: UtxoSignedTransaction) {
         try {
             transactionVerificationService.verify(signedTransaction.toLedgerTransaction())
-        } catch(e: TransactionVerificationException){
-            throw e
+        } catch(e: TransactionVerificationException) { throw e }
+    }
+
+    @Suspendable
+    private fun persistTransaction(transaction: UtxoSignedTransaction) {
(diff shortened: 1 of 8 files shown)


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    @Suspendable
+    private fun verifyTransaction(signedTransaction: UtxoSignedTransaction) {
+        try {
+            transactionVerificationService.verify(signedTransaction.toLedgerTransaction())
+        } catch(e: TransactionVerificationException){
+            throw e
+        }
```
lankydan: Formatting.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
```diff
@@ -0,0 +1,67 @@
...
+    override fun call(): UtxoSignedTransaction {
+        val transaction = session.receive(UtxoSignedTransaction::class.java)
+        verifyTransaction(transaction)
+        persistTransaction(transaction)
+
+        return transaction
+    }
+
+    @Suspendable
+    private fun persistTransaction(transaction: UtxoSignedTransaction) {
```
lankydan: Switch the order of the `persistTransaction` and `verifyTransaction` methods to follow the usage order in `call`, makes it slightly easier to read the code.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
```diff
@@ -0,0 +1,67 @@
...
+    @CordaInject
+    lateinit var persistenceService: UtxoLedgerPersistenceService
+
+    @CordaInject
+    lateinit var flowEngine: FlowEngine
+
+    @Suspendable
+    override fun call(): UtxoSignedTransaction {
+        val transaction = session.receive(UtxoSignedTransaction::class.java)
+        verifyTransaction(transaction)
```
lankydan: We need to receive the transaction's backchain before we can verify the transaction. If you tried to do this with a transaction that had a backchain, it would fail on trying to find the inputs to the received transaction. You should be able to find the code you need in `ReceiveFinalityFlowV1`.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
```diff
@@ -0,0 +1,67 @@
...
+            log.debug("Recorded transaction = ${transaction.id} since transaction is verified")
+        }
+    }
+
+    @Suspendable
+    private fun verifyTransaction(signedTransaction: UtxoSignedTransaction) {
+        try {
+            transactionVerificationService.verify(signedTransaction.toLedgerTransaction())
+        } catch(e: TransactionVerificationException){
+            throw e
```
lankydan: I think its worth storing the transaction as invalid when an error occurs like we do finality flow.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/SendTransactionFlow.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    @CordaInject
+    lateinit var flowEngine: FlowEngine
+
+    @CordaInject
+    lateinit var utxoLedgerPersistenceService: UtxoLedgerPersistenceService
+
+    @Suspendable
+    override fun call() {
+        requireNotNull(utxoLedgerPersistenceService.findSignedTransaction(transaction.id)) {
```
lankydan: I understand why you're doing this, but I wonder if there is a better way?

Maybe we put the logic into the receive transaction flow? Where it determines whether to store the transaction as UNVERIFIED or VERIFIED based on the signatures of the transaction? If all the required signatures have been added, then we can store as VERIFIED, otherwise store it as UNVERIFIED?

I think this sort of logic will be needed anyway, now that I think about it, because we need logic to prevent people sending UNVERIFIED transactions to others where the receivers think its a VERIFIED transaction and store it as such.

This does also mean we can send UNVERIFIED transactions around; however, I'm not sure how useful that is in this scenario.

Doing this, would remove the need to call `findSignedTransaction` here, removing the need to do a database lookup and therefore have a better general performance.

@blsemo @nkovacsx thoughts?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
jennyang-r3: @blsemo If we are not on the same page, drop me a line.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
jennyang-r3: @blsemo wdyt?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/SendTransactionFlow.kt
jennyang-r3: I'm confused about why we need to store UNVERIFIED transactions if we can prevent sending UNVERIFIED ones in the first place.

lankydan: We need to prevent it because there is nothing preventing the sending flow from running a modified version of the code (I think they can actually implement their own send transaction flow from our API) that could send an unverified transaction to the peer calling `ReceiveTransactionFlow`.

By verifying the received transaction's signatures in `ReceiveTransactionFlow`, it ensures that the receiving member can be sure that it is indeed a verified transaction.

If we don't verify the signatures, then we have to trust that the sender is sending a verified and notarised transaction.

jennyang-r3: In that case my follow up question would be:
> _(I think they can actually implement their own send transaction flow from our API)_

How can they implement their own and send tx using our API if we our API don't provide to pass flows in?

> _If we don't verify the signatures, then we have to trust that the sender is sending a verified and notarised transaction._

Are you saying verifying tx on the receiver side is not enough? 
IIUC, this feature was for DC team to distribute their ref states to observers, who are not immediate participants. If they are just observers(receivers in this case) are their signatures included in the transaction they received?

lankydan: No, the verification code called at the moment isn't enough. The logic in the backchain verifier does the checks that I'd expect us to do here as well (afaik):

```kotlin
 log.trace { "Backchain resolution of $initialTransactionIds - Verifying transaction $transactionId" }
                        transaction.verifySignatorySignatures()
                        transaction.verifyAttachedNotarySignature()
                        utxoLedgerTransactionVerificationService.verify(transaction)
                        log.trace { "Backchain resolution of $initialTransactionIds - Verified transaction $transactionId" }
```

After pasting this code here though, it does make me wonder if we can simply call the backchain resolution code on the transaction we want to send to the peer. Then it can handle all the receiving of its dependencies and verifying of the sent transaction. The logging code we have though might make it confusing :/.

In answer to the second question. We wouldn't be verifying that the receiving member is a signer of the transaction, only that all the signatories specified in the transaction, did, in fact, sign the transaction.

jennyang-r3: >  it does make me wonder if we can simply call the backchain resolution code on the transaction we want to send to the peer.

But then if the transaction doesn't have dependencies, it won't verify all these since this will be only run in backchain flow, will they?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
```diff
@@ -0,0 +1,71 @@
...
+        val transaction = session.receive(UtxoSignedTransactionInternal::class.java)
+        try {
+            transaction.verifySignatorySignatures()
+            transaction.verifyAttachedNotarySignature()
+            transactionVerificationService.verify(transaction.toLedgerTransaction())
+        } catch (e: Exception) {
+            val message = "Failed to verify transaction and signatures of transaction: ${transaction.id}"
+            log.warn(message, e)
+            session.send(Payload.Failure<List<DigitalSignatureAndMetadata>>(message))
+            throw e
```
lankydan: Can we do `throw CordaRuntimeException(message, e)` here.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        val transactionDependencies = transaction.dependencies
+        if (transactionDependencies.isNotEmpty()) {
+            try {
+                flowEngine.subFlow(TransactionBackchainResolutionFlow(transactionDependencies, session))
+            } catch (e: InvalidBackchainException) {
+                val message = "Invalid transaction found during back-chain resolution, marking transaction with ID " +
```
lankydan: We should change the message since we're not marking the transaction as invalid.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/SendTransactionFlow.kt
```diff
@@ -0,0 +1,61 @@
...
+    lateinit var flowEngine: FlowEngine
+
+    @CordaInject
+    lateinit var utxoLedgerPersistenceService: UtxoLedgerPersistenceService
+
+    @Suspendable
+    override fun call() {
+        flowMessaging.sendAll(transaction as UtxoSignedTransactionInternal, sessions.toSet())
+
+        sessions.forEach {
```
lankydan, line 37: I think we should put all the code into a single `forEach`/`for`.

In the scenario that an invalid transaction is sent, it will still do backchain resolution on every session, and only then stopping because it gets an error back.

If we combine the `forEach` then each session will fully complete and tell the sender whether the transaction is valid or not, the sender can then fail at that point since it received an error saying the transaction was invalid.

It also means that each session that is receiving a transaction can move on quicker, whereas at the moment, the first session gets the backchain, waits for the other sessions to also get it and then moves on once they have.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/SendTransactionFlow.kt
```diff
@@ -0,0 +1,61 @@
...
+                }
+            }
+        }
+
+        sessions.forEach {
+            val sendingTransactionResult = it.receive(Payload::class.java)
+            when (sendingTransactionResult) {
+                is Payload.Success -> return
+                is Payload.Failure ->
+                    throw UnverifiedTransactionSentException(
```
lankydan: I don't think its worth having a specific exception in this scenario. A `CordaRuntimeException` should be fine.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
```diff
@@ -0,0 +1,71 @@
...
+                session.send(Payload.Failure<List<DigitalSignatureAndMetadata>>(message))
+                throw e
+            }
+        } else {
+            log.trace {
+                "Transaction with id ${transaction.id} has no dependencies so backchain resolution will not be performed."
+            }
+        }
+
+        session.send(Payload.Success("Successfully received transaction."))
```
lankydan: We need to persist the transaction, AFAIK, this API isn't just about sending and receiving a transaction, but receiving a transaction and persisting it (so that it can be used later).

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
```diff
@@ -0,0 +1,71 @@
...
+            transactionVerificationService.verify(transaction.toLedgerTransaction())
+        } catch (e: Exception) {
+            val message = "Failed to verify transaction and signatures of transaction: ${transaction.id}"
+            log.warn(message, e)
+            session.send(Payload.Failure<List<DigitalSignatureAndMetadata>>(message))
+            throw e
+        }
+
+        val transactionDependencies = transaction.dependencies
+        if (transactionDependencies.isNotEmpty()) {
```
lankydan, line 52: The backchain resolution code needs to be before the `transactionVerificationService.verify` call otherwise it won't work as it won't have resolved the dependencies needed by verification.

lankydan (commented):
We need to write some unit tests for these flows.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/SendTransactionFlow.kt
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            val sendingTransactionResult = it.receive(Payload::class.java)
+            when (sendingTransactionResult) {
+                is Payload.Success -> return
+                is Payload.Failure ->
+                    throw CordaRuntimeException(
+                        "Failed to send transaction: ${transaction.id} due to unverified transaction sent."
+                    )
+            }
```
lankydan: I think this would look better with an `if` instead of a `when` so we can remove the `return` since the method returns `Unit` anyway.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlowTest.kt
```diff
@@ -0,0 +1,148 @@
...
+        whenever(sessionAlice.receive(UtxoSignedTransactionInternal::class.java)).thenReturn(transaction)
+        whenever(visibilityChecker.containsMySigningKeys(listOf(publicKeyAlice))).thenReturn(true)
+        whenever(visibilityChecker.containsMySigningKeys(listOf(publicKeyBob))).thenReturn(true)
+
+        callReceiveTransactionFlow(sessionAlice)
+
+        verify(sessionAlice).receive(UtxoSignedTransactionInternal::class.java)
+        verify(transaction).verifySignatorySignatures()
+        verify(transaction).verifyAttachedNotarySignature()
+        verify(transactionVerificationService).verify(ledgerTransaction)
```
lankydan: I think all these verifications are overkill for the test, the main one is the `persist` since that depends on the rest of the verification calls passing.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlowTest.kt
```diff
@@ -0,0 +1,148 @@
...
+        )
+
+        callReceiveTransactionFlow(sessionAlice)
+
+        verify(flowEngine).subFlow(TransactionBackchainResolutionFlow(transaction.dependencies, sessionAlice))
+        verify(sessionAlice).send(Payload.Success("Successfully received transaction."))
+    }
+
+    @Test
+    fun `receiving invalid transaction should throw exception`() {
```
lankydan: For completeness sake, you should also add tests where `verifyAttachedNotarySignature` and `transactionVerificationService. verify` throw and cause the flow to fail (like this test).

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlowTest.kt
```diff
@@ -0,0 +1,148 @@
...
+        whenever(sessionAlice.receive(UtxoSignedTransactionInternal::class.java)).thenReturn(transaction)
+        whenever(transaction.verifySignatorySignatures()).thenThrow(
+            CordaRuntimeException("Failed to verify"))
+
+        assertThatThrownBy { callReceiveTransactionFlow(sessionAlice) }
+            .isInstanceOf(CordaRuntimeException::class.java)
+            .hasMessageContaining("Failed to verify transaction")
+
+        verify(sessionAlice).receive(UtxoSignedTransactionInternal::class.java)
+        verify(transaction).verifySignatorySignatures()
```
lankydan: We don't need to verify `verifySignatorySignatures` because we know its called since we're throwing an exception when its called.

Would be good to check that we sent that failure payload here as well.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/SendTransactionFlowTest.kt
```diff
@@ -0,0 +1,109 @@
...
+            Payload.Success(successMessage)
+        )
+
+        callSendTransactionFlow(transaction, sessions)
+
+        verify(flowEngine, never()).subFlow(TransactionBackchainSenderFlow(TX_ID, sessionAlice))
+    }
+
+    @Test
+    fun `sending unverified transaction with no dependencies should not call backchain flow`() {
```
lankydan: I don't think this test makes sense.

The transaction being unverified is separate from the backchain code being called. We only find out that at the transaction being sent is incorrect after backchain verification occurs.

The content of the test seems fine but the name sounds wrong.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/SendTransactionFlow.kt
```diff
@@ -0,0 +1,54 @@
...
+            } else {
+                log.trace {
+                    "Transaction with id ${transaction.id} has no dependencies so backchain resolution will not be performed."
+                }
+            }
+
+            val sendingTransactionResult = it.receive(Payload::class.java)
+            if (sendingTransactionResult is Payload.Failure) {
+                throw CordaRuntimeException(
+                    "Failed to send transaction: ${transaction.id} due to unverified transaction sent."
```
blsemo: This should rather print the failure message than infering what went wrong.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/transactiontransmission/ReceiveTransactionFlow.kt
```diff
@@ -0,0 +1,83 @@
...
+            transactionVerificationService.verify(transaction.toLedgerTransaction())
+        } catch (e: Exception) {
+            val message = "Failed to verify transaction and signatures of transaction: ${transaction.id}"
+            log.warn(message, e)
+            session.send(Payload.Failure<List<DigitalSignatureAndMetadata>>(message))
+            throw CordaRuntimeException(message, e)
+        }
+
+        session.send(Payload.Success("Successfully received transaction."))
+        ledgerPersistenceService.persist(transaction, VERIFIED, transaction.getVisibleStateIndexes(visibilityChecker))
```
blsemo, line 78: I think lines 78 and 79 are in the wrong order - it's only a success _after_ we stored the new transaction.

blsemo (commented):
A couple of small issues, but otherwise looks good to me.

lankydan (dismissed):

blsemo (dismissed):
LGTM

blsemo (approved):
Re-approved

PR #5047, opened 2023-11-06
CORE-18209 Adjust the mediator to send asynchronous events directly before saving the state to try to avoid a race condition

LWogan:
In the mediator code we poll a batch of events, process each input event, send any output events immediately (HTTP requests & Bus records), and after completing the batch of input events we save the states to the state storage in a batch.
A race condition appears as the output events of an input event can be a SessionEvent output to the bus and subsequently processed by another pod.
This can result in a SessionEvent replied back to our flowId and processed on a different topic by a different pod.
This can all be done while the initial flow worker is still processing its batch of inputs events and before the state has been saved to the state storage.
As a result when the other flow worker pod picks up the SessionEvent it doesn't find the checkpoint and logs the error we saw.

A temporary fix would be to send bus events directly before saving to the state storage. 
If the state storage optimistic locking fails to save the state then duplicate output records would be sent(this is fine) however the race condition will reappear

### DIFF
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
@@ -166,7 +166,7 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
             var groups = allocateGroups(messages.map { it.toRecord() })
             var states = stateManager.get(messages.map { it.key.toString() }.distinct())
             while (groups.isNotEmpty()) {
-                val busEvents = mutableListOf<MediatorMessage<Any>>()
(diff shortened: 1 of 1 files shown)


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
```diff
@@ -280,6 +293,8 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
             val records = buckets[key]!!
             group[key] = records
         }
+
+        log.info("groups keys: ${groups.map { it.keys }}")
```
JamesHR3: Should probably take this log out.

JamesHR3 (commented):
Looks reasonable as a workaround. Just a small log to take out.

JamesHR3 (approved):

driessamyn (approved):

jujoramos (approved):

PR #5048, opened 2023-11-06
CORE-18212 - default the state manager max connections to 8 

conal-smith:
**Problem:**

Default SM connections is 5, default taskManager threads is 8. By default, there could be a bottleneck as threads may be waiting for db connections to be released.

**Solution**

Set the default max number of connections to 8 for state manager.

**Ambiguous**

Token selection does not use the TaskManager nor the MultiSourceEventMediator, however I think keeping the max connections consistent for all workers is a better approach.


conal-smith:
Closing because multi source event mediator does not use more than 5 threads

PR #5050, opened 2023-11-07
CORE-18194 - Revert `fetch.max.wait.ms` default config

driessamyn:
The `fetch.max.wait.ms` config had been reduced aggressively which was causing high CPU load (see ticket CORE-18162) and, in turn, reduced performance in the E2E tests.

### DIFF
--- a/libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
+++ b/libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
@@ -19,7 +19,7 @@ consumer = ${common} {
     enable.auto.commit = false
     # Retrieve 500 records maximum per poll. Note that poll will return immediately if any records are available, so a
     # batch may contain fewer than 500 records.
-    max.poll.records = 200
+    max.poll.records = 100
     # Time to allow between polls on a consumer before a rebalance occurs that removes this consumer's partitions.
     max.poll.interval.ms = 300000
     # Timeout of heartbeats between the consumer and the broker. If no heartbeat is received in this timeframe, a
@@ -29,7 +29,7 @@ consumer = ${common} {
     heartbeat.interval.ms = 20000
     # The maximum amount of time kafka broker will wait before answering a consumer request if there hasn't been
     # an update to the topic
-    fetch.max.wait.ms = 5
+    fetch.max.wait.ms = 20
 }
 
 # Defaults for all producers.
@@ -78,7 +78,7 @@ roles {
         # within this pattern is transactional by default, and the transaction commit also causes a flush.
         producer = ${producer} {
             # Maximum time to wait for additional messages before sending the current batch.
-            linger.ms = 5
+            linger.ms = 50
             # Maximum amount of memory in bytes (not messages) that will be used for each batch. Can not be higher than
             # the value configured for "message.max.bytes" on the broker side (1mb by default).
             batch.size = 204800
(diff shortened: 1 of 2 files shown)


driessamyn:
E2E test run with this change: https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda5-e2e-tests/detail/release%2F5.1/1225/pipeline/

vkolomeyko (dismissed):
LGTM

libs/messaging/kafka-message-bus-impl/src/main/resources/kafka-messaging-defaults.conf
```diff
@@ -29,7 +29,7 @@ consumer = ${common} {
     heartbeat.interval.ms = 20000
     # The maximum amount of time kafka broker will wait before answering a consumer request if there hasn't been
     # an update to the topic
-    fetch.max.wait.ms = 5
+    fetch.max.wait.ms = 500
```
jujoramos, line 32: The original value (before the topology changes) was `1500`, I think @Tom-Fitzpatrick added this property a while ago after some investigation during the early perf tests executed in March/June this year.
Is there any reason to use `500` instead of `1500`? (asking out of curiosity only).

driessamyn: I set it to `500` because that's what the [Kafka default is](https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html).
However, before merging, I'd like to get @Omar-awad 's input into why it was lowered to 5 as I'm not entirely sure what the impact is (except from the observed lower CPU usage and the impact of that).

jujoramos: > I set it to 500 because that's what the [Kafka default is](https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html).

Got it, thanks for the clarification!

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
```diff
@@ -68,7 +68,7 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
     private val running = AtomicBoolean(false)
     // TODO This timeout was set with CORE-17768 (changing configuration value would affect other messaging patterns)
     //  This should be reverted to use configuration value once event mediator polling is refactored (planned for 5.2)
-    private val pollTimeout = Duration.ofMillis(5)
+    private val pollTimeout = Duration.ofMillis(50)
```
driessamyn, line 71: FYI @Omar-awad

kyriathar (approved):
LGTM (`fetch.max.wait.ms` bit)

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
driessamyn: E2E test that includes this change: https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda5-e2e-tests/detail/release%2F5.1/1231/pipeline

vkolomeyko (approved):

Omar-awad (approved):
Approved

PR #5053, opened 2023-11-07
CORE-17495: Add checks to ensure that CPKs do not share a name

malachyb:
A Cordapp CPK has a field used to identify it as the same Cordapp across multiple versions. This PR ensures that CPKs do not share a name to prevent confusion in the upgrade process.

### DIFF
--- a/libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
+++ b/libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
@@ -95,11 +95,11 @@ class CpiLoaderV2(private val clock: Clock = UTCClock()) : CpiLoader {
                         verifySignature = false,
                         cpkFileName = Paths.get(it.entry.name).fileName.toString()
                     ).also {
-                        val cpkName = it.metadata.cpkId.name
-                        if (cordappCpkNames.contains(cpkName)) {
-                            throw PackagingException("Multiple CPKs share the Corda-CPK-Cordapp-Name $cpkName.")
+                        val cordappCpkName = it.metadata.cpkId.name
+                        if (cordappCpkNames.contains(cordappCpkName)) {
+                            throw PackagingException("Multiple CPKs share the Corda-CPK-Cordapp-Name $cordappCpkName.")
                         }
-                        cordappCpkNames.add(cpkName)
+                        cordappCpkNames.add(cordappCpkName)
                     }
                 }
         }
--- a/libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
+++ b/libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
@@ -59,16 +59,6 @@ class CpiLoaderV2(private val clock: Clock = UTCClock()) : CpiLoader {
                 emptyList()
             }
 
-            val cordappCpkNames = HashSet<String>()
-
-            cpks.forEach {
-                val name = it.metadata.cpkId.name
-                if (cordappCpkNames.contains(name)) {
-                    throw PackagingException("Multiple CPKs share a Corda-CPK-Cordapp-Name")
-                }
-                cordappCpkNames.add(name)
-            }
-
             val mainAttributes = jarInputStream.manifest.mainAttributes
             return CpiImpl(
                 CpiMetadata(
@@ -92,6 +82,8 @@ class CpiLoaderV2(private val clock: Clock = UTCClock()) : CpiLoader {
     private fun calculateHash(cpiBytes: ByteArray) = cpiBytes.hash(DigestAlgorithmName.SHA2_256).bytes
 
     private fun readCpksFromCpb(cpb: InputStream, expansionLocation: Path, cpiLocation: String?): List<Cpk> {
+        val cordappCpkNames = HashSet<String>()
+
         return JarInputStream(cpb, false).use { cpbInputStream ->
             readJar(cpbInputStream)
                 .filter { it.entry.name.endsWith(".jar") }
@@ -102,7 +94,14 @@ class CpiLoaderV2(private val clock: Clock = UTCClock()) : CpiLoader {
                         cpkLocation = cpiLocation.plus("/${it.entry.name}"),
                         verifySignature = false,
                         cpkFileName = Paths.get(it.entry.name).fileName.toString()
-                    )
+                    ).apply {
+                        metadata.cpkId.name.let {cpkName ->
+                            if (cordappCpkNames.contains(cpkName)) {
+                                throw PackagingException("Multiple CPKs share the Corda-CPK-Cordapp-Name $cpkName.")
+                            }
+                            cordappCpkNames.add(cpkName)
+                        }
+                    }
                 }
         }
     }
--- a/libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
+++ b/libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
@@ -94,13 +94,12 @@ class CpiLoaderV2(private val clock: Clock = UTCClock()) : CpiLoader {
                         cpkLocation = cpiLocation.plus("/${it.entry.name}"),
                         verifySignature = false,
                         cpkFileName = Paths.get(it.entry.name).fileName.toString()
-                    ).apply {
-                        metadata.cpkId.name.let {cpkName ->
-                            if (cordappCpkNames.contains(cpkName)) {
-                                throw PackagingException("Multiple CPKs share the Corda-CPK-Cordapp-Name $cpkName.")
-                            }
-                            cordappCpkNames.add(cpkName)
+                    ).also {
+                        val cpkName = it.metadata.cpkId.name
+                        if (cordappCpkNames.contains(cpkName)) {
+                            throw PackagingException("Multiple CPKs share the Corda-CPK-Cordapp-Name $cpkName.")
                         }
+                        cordappCpkNames.add(cpkName)
                     }
                 }
         }
--- a/libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
+++ b/libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
@@ -59,6 +59,16 @@ class CpiLoaderV2(private val clock: Clock = UTCClock()) : CpiLoader {
                 emptyList()
             }
 
+            val cordappCpkNames = HashSet<String>()
+
+            cpks.forEach {
+                val name = it.metadata.cpkId.name
+                if (cordappCpkNames.contains(name)) {
+                    throw PackagingException("Multiple CPKs share a Corda-CPK-Cordapp-Name")
+                }
+                cordappCpkNames.add(name)
+            }
+
             val mainAttributes = jarInputStream.manifest.mainAttributes
             return CpiImpl(
                 CpiMetadata(
--- a/libs/packaging/packaging/src/test/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2Test.kt
+++ b/libs/packaging/packaging/src/test/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2Test.kt
@@ -35,15 +35,15 @@ class CpiLoaderV2Test {
 
     @Test
     fun `CPI loading fails with duplicate CPKs`() {
-        val cpkName = "com.test.duplicate"
+        val cordappCpkName = "com.test.duplicate"
 
         val inMemoryCpk1 = TestCpkV2Builder()
             .name("test1.jar")
-            .bundleName(cpkName)
+            .bundleName(cordappCpkName)
 
         val inMemoryCpk2 = TestCpkV2Builder()
             .name("test2.jar")
-            .bundleName(cpkName)
+            .bundleName(cordappCpkName)
 
         val inMemoryCpb = TestCpbV2Builder()
             .cpks(inMemoryCpk1, inMemoryCpk2)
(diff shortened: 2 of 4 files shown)


libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -35,10 +35,21 @@ internal class SandboxGroupImpl(
     private val classTagFactory: ClassTagFactory,
     private val bundleUtils: BundleUtils
 ) : SandboxGroupInternal {
+    private val cpkNames = ConcurrentHashMap.newKeySet<String>()
+    init {
+        cpkSandboxes.forEach {
```
simon-johnson-r3, line 40: Kotlin is usually pretty good at this stuff and it's always worth a search for any functions that might help.

Here for example you could do
```
if (cpkSandboxes.groupBy { it.cpkMetadata.cpkId.name }.values.any { it.size > 1 }) {
    throw SandboxException("CPK $name is declared more than once.")
}
```
Basically turn cpkSandboxes into a map where the key is the name and the value a list of sandboxes, so any values which have more than 1 item in the list must have duplicate names. There might be other ways to do it which are even easier.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -35,10 +35,21 @@ internal class SandboxGroupImpl(
...
+            val name = it.cpkMetadata.cpkId.name
+            if (cpkNames.contains(name)) {
+                throw SandboxException("CPK $name is declared twice.")
```
simon-johnson-r3: Minor, but I would change this error to more than once as you don't actually know how many times it's declared.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -35,10 +35,21 @@ internal class SandboxGroupImpl(
...
+            } else {
+                cpkNames.add(it.cpkMetadata.cpkId.name)
+            }
```
simon-johnson-r3, line 43: We want Corda to continue to run but fail whatever request lead to the sandbox creation sensibly, so the REST api called can make sense of what happened. Can you confirm this is what happens if we fail on construction like this?

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
malachyb: I didn't realise Kotlin had that. I'll change that now.

vkolomeyko:
Should we re-target this PR to 5.2 branch?

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
```diff
@@ -59,6 +59,16 @@ class CpiLoaderV2(private val clock: Clock = UTCClock()) : CpiLoader {
                 emptyList()
             }
 
+            val cordappCpkNames = HashSet<String>()
```
simon-johnson-r3: you could save yourself a loop here and roll this into `readCpksFromCpb`

```
    private fun readCpksFromCpb(cpb: InputStream, expansionLocation: Path, cpiLocation: String?): List<Cpk> {
        val cordappCpkNames = HashSet<String>()
        return JarInputStream(cpb, false).use { cpbInputStream ->
            readJar(cpbInputStream)
                .filter { it.entry.name.endsWith(".jar") }
                .map {
                    CpkReader.readCpk(
                        it.bytes.inputStream(),
                        expansionLocation,
                        cpkLocation = cpiLocation.plus("/${it.entry.name}"),
                        verifySignature = false,
                        cpkFileName = Paths.get(it.entry.name).fileName.toString()
                    ).apply {
                        metadata.cpkId.name.let { cpkName ->
                            if (cordappCpkNames.contains(cpkName)) {
                                throw PackagingException("Multiple CPKs share a Corda-CPK-Cordapp-Name")
                            }
                            cordappCpkNames.add(cpkName)
                        }
                    }
                }
        }
    }
```
We don't upload CPIs so often though, and they don't contain that many CPKs, so only implement this if you think the code is better.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -35,10 +35,15 @@ internal class SandboxGroupImpl(
     private val classTagFactory: ClassTagFactory,
     private val bundleUtils: BundleUtils
 ) : SandboxGroupInternal {
+    init {
+        cpkSandboxes.groupBy { it.cpkMetadata.cpkId.name }.filter { (it.value.size > 1) }.map {
```
simon-johnson-r3: You've used a different way of detecting duplicates here to above which seems a little inconsistent. The advantage of the other method is it doesn't have to process the entire list before matching a duplicate. But as I said up there, these containers have small number of elements typically. Creating sandbox groups happens whenever we need a new sandbox however, so this occurs a lot more than just uploading a CPI.

I'd say if anything this should have the more efficient solution.

But overall I'd say pick a method and stick with it.

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
```diff
@@ -59,6 +59,16 @@ class CpiLoaderV2(private val clock: Clock = UTCClock()) : CpiLoader {
...
+
+            cpks.forEach {
+                val name = it.metadata.cpkId.name
+                if (cordappCpkNames.contains(name)) {
+                    throw PackagingException("Multiple CPKs share a Corda-CPK-Cordapp-Name")
```
simon-johnson-r3: I'd suggest logging the name here

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
```diff
@@ -102,7 +94,14 @@ class CpiLoaderV2(private val clock: Clock = UTCClock()) : CpiLoader {
                         cpkLocation = cpiLocation.plus("/${it.entry.name}"),
                         verifySignature = false,
                         cpkFileName = Paths.get(it.entry.name).fileName.toString()
-                    )
+                    ).apply {
```
kyriathar: We don't seem to apply anything to the object (`Cpk`). This could rather be an `also`

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
```diff
@@ -102,7 +94,14 @@ class CpiLoaderV2(private val clock: Clock = UTCClock()) : CpiLoader {
...
+                        metadata.cpkId.name.let {cpkName ->
```
kyriathar: We could simplify a bit instead of using `let` and rather `val cpkName = cpk.metadata.cpkId.name`

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
```diff
@@ -36,8 +36,12 @@ internal class SandboxGroupImpl(
     private val bundleUtils: BundleUtils
 ) : SandboxGroupInternal {
     init {
-        cpkSandboxes.groupBy { it.cpkMetadata.cpkId.name }.filter { (it.value.size > 1) }.map {
-            throw SandboxException("CPK ${it.key} is declared more than once.")
+        val cordappCpkNames = HashSet<String>()
```
kyriathar, line 39: I think this check is now redundant as we check for cordapp cpk names clashing at an earlier stage (i.e. at parsing the CPI) so it should no longer be possible to now hit this? So we could maybe revert this change?

libs/packaging/packaging/src/test/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2Test.kt
```diff
@@ -28,5 +32,32 @@ class CpiLoaderV2Test {
...
+            .cpb(inMemoryCpb)
+            .build()
+
+        val e = assertThrows<PackagingException> {
+            CpiLoaderV2().loadCpi(inMemoryCpi.toByteArray(), tmp, "in-memory", false)
+        }
+
+        assert(e.message!!.contains("Multiple CPKs share the Corda-CPK-Cordapp-Name")) {
+            "Failure should be caused by multiple CPKs sharing a Corda-CPK-Cordapp-Name."
+        }
```
kyriathar: Let's better use `assertEquals` here I think and so let's try to assert the complete error message.

libs/packaging/packaging/src/test/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2Test.kt
```diff
@@ -28,5 +32,32 @@ class CpiLoaderV2Test {
...
+
+    @Test
+    fun `CPI loading fails with duplicate CPKs`() {
+        val inMemoryCpk1 = TestCpkV2Builder()
+            .name("test1.jar")
+            .bundleName("duplicate.cpk")
+
+        val inMemoryCpk2 = TestCpkV2Builder()
+            .name("test2.jar")
+            .bundleName("duplicate.cpk")
```
kyriathar: nit: this looks more like a cpk file name

kyriathar (commented):
A few minor comments

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
kyriathar: I only just saw this comment. I have posted [below](https://github.com/corda/corda-runtime-os/pull/5053#discussion_r1412458724) regarding this. Do we still need this check if we say we now check earlier at CPI parsing?

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
simon-johnson-r3: I suggested this originally, it's because `metadata` applies to `this`, but you are correct in pointing out semantically it's not a great use of `apply`. `also` is for "additional effects" so I agree is a better fit.

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
simon-johnson-r3: I'm not sure I agree that the use of let here is controversial, but I'm not opposed to it being changed either. The advantage of using let over temp vals is you limit the scope of the val, but of course here it's a moot point as there is nothing to limit it from.

libs/packaging/packaging/src/test/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2Test.kt
simon-johnson-r3: Personally I'd prefer not to assert on error messages at all. This becomes a cut and past maintenance effort. If you made e.g. a typo in the message at the point of definition, you'd most likely end up pasting the same typo into the test, so it's not really checking much, given this is a sub type of exception anyway.

libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxGroupImpl.kt
simon-johnson-r3: I'd say we might want to keep it. Sandbox groups are created all the time during Corda's operation and having two cpks with the same name will cause undefined behaviour. Uploading CPIs is something that was done once in the past. It's a fairly reasonable piece of defensive coding I think.

Also I realise now I might have proposed using `groupBy` originally and then I suggested changing it! Given we are in a period where performance is key, we might as well stick to lengthier more efficient solutions.

libs/packaging/packaging/src/test/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2Test.kt
malachyb: My thoughts behind asserting on it was that you could run into a situation where you think you're causing the right error but you're actually getting a different one with the same type. I ran into a situation a while ago where there were a few tests that did that so I try to avoid it with my tests.

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
```diff
@@ -94,13 +94,12 @@ class CpiLoaderV2(private val clock: Clock = UTCClock()) : CpiLoader {
...
                         verifySignature = false,
                         cpkFileName = Paths.get(it.entry.name).fileName.toString()
-                    ).apply {
-                        metadata.cpkId.name.let {cpkName ->
-                            if (cordappCpkNames.contains(cpkName)) {
-                                throw PackagingException("Multiple CPKs share the Corda-CPK-Cordapp-Name $cpkName.")
-                            }
-                            cordappCpkNames.add(cpkName)
+                    ).also {
+                        val cpkName = it.metadata.cpkId.name
```
kyriathar: This was my bad proposing `cpkName`, I think this should rather be named `cordappCpkName` shouldn't it?

libs/packaging/packaging/src/test/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2Test.kt
```diff
@@ -35,13 +35,15 @@ class CpiLoaderV2Test {
 
     @Test
     fun `CPI loading fails with duplicate CPKs`() {
+        val cpkName = "com.test.duplicate"
+
         val inMemoryCpk1 = TestCpkV2Builder()
             .name("test1.jar")
-            .bundleName("duplicate.cpk")
+            .bundleName(cpkName)
```
kyriathar: Let's update as above

libs/packaging/packaging/src/test/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2Test.kt
simon-johnson-r3: Yes, I understand. I would suggest we should try and handle that with exception sub classes though, that is what they are for. We don't need to expose those types externally to the component.

kyriathar (approved):
LGTM

simon-johnson-r3 (approved):

PR #5055, opened 2023-11-07
CORE-17194 Adding prefix to consumer group names

ben-millar:
This PR adds the topic prefix to Kafka consumer group names in order to namespace consumer groups per-deployment.

We've been seeing `TopicAuthorizationExceptions` in the `5.0`/`5.1` [compatibility tests](https://ci02.dev.r3.com/job/Corda5/job/corda-e2e-tests-cross-compatibility-tests/) wherein one deployment appears to be trying to access a topic owned by the other deployment. This access is being blocked, as we use a `topicPrefix` to namespace topics within the same Kafka cluster.

It appears that the issue lies in the consumer groups being the same across both deployments. We were able to reproduce the issue and then, by adding a per-deployment UUID to the group names, resolve the observed symptoms.

A cross-version test was triggered with these changes, results can be found here: https://ci02.dev.r3.com/job/Corda5/job/corda-e2e-tests-cross-compatibility-tests/job/PR-17194%252Fbm/6/

_Note: We are currently engaging with confluent support to try to isolate a single root cause or mechanism of action for this._

### DIFF
--- a/libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/config/MessageBusConfigResolver.kt
+++ b/libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/config/MessageBusConfigResolver.kt
@@ -33,8 +33,6 @@ internal class MessageBusConfigResolver(private val smartConfigFactory: SmartCon
         private const val GROUP_PATH = "group"
         private const val CLIENT_ID_PATH = "clientId"
         private const val TRANSACTIONAL_ID_PATH = "transactionalId"
-
-//        private val GROUP_PREFIX: String = UUID.randomUUID().toString()
     }
 
     private val defaults = getResourceConfig(DEFAULT_CONFIG_FILE)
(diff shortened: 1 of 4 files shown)


libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/config/MessageBusConfigResolver.kt
```diff
@@ -158,7 +164,7 @@ internal class MessageBusConfigResolver(private val smartConfigFactory: SmartCon
         return smartConfigFactory.create(
             ConfigFactory.parseMap(
                 mapOf(
-                    GROUP_PATH to group,
+                    GROUP_PATH to GROUP_PREFIX + group,
```
LWogan: I think this will not work when there is multiple workers trying to share load across topics. For example: I think if you spun up 2 flow workers with this change and a start flow event was placed on the start flow topic, i think both flow engines would pick up and process that record. The 2 consumers would be in different consumer groups. 

Very interesting though that we think a consumer group is shared across topics in a rebalance. We should verify this to be true carefully. The consumer group name is user specified (user of the messaging library) in most if not all cases, so it could easily be accidentally duplicated.

ben-millar: Fair point; this isn't a permanent solution. I think we just need to scope our group names per-deployment in the same way as we do for topic names.

LWogan: using topic prefix as the prefix might be better . I think UUID breaks our corda clusters

simon-johnson-r3 (approved):
This is a good change in its own right as it decouples consumer groups from different clusters using the same Kafka deployment - the are clearly not conceptually part of a joined group of consumers. It will change the dynamics of rebalancing (in a good way) on that basis, and I wonder if this is what's causing the bug to go away. However I see you note Confluent are being consulted on this, so that's good enough for me as a follow up for now.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/config/MessageBusConfigResolver.kt
simon-johnson-r3: @LWogan Regarding `Very interesting though that we think a consumer group is shared across topics in a rebalance. We should verify this to be true carefully.` there's a slack thread about this and I share your concerns, because all Kafka documentation would suggest the opposite IMHO.

What is true though is sharing a consumer group like this is considered bad practice, because any client joining or leaving that group causes a rebalance. In this instance clients in a different cluster are in the same consumer group for no reason at all. Potentially a worker in one cluster coming or going could cause a rebalance in a totally different cluster.

I do think it's reasonable we use consumer groups for groups of consumers of the same thing, that is after all what it means, whether or not we can explain the other behaviour.

LWogan (approved):
LGTM

PR #5063, opened 2023-11-08
CORE-18239: Delete Correct State Within Mediator

jujoramos:
Update event mediator so it uses the State returned by the processor,
instead of the original one, when making the decision on whether to
delete it from the persistent storage or not. Failing to do so leaves
orphaned States in the persistent storage, which in turn ends up
causing other issues in the system.

 Co-authored-by: James Higgs <james.higgs@r3.com>


JamesHR3 (dismissed):

driessamyn (dismissed):

driessamyn (approved):
LGTM

PR #5069, opened 2023-11-09
CORE-18228: Swap the send to after the state storage in the event mediator

JamesHR3:
### Problem description

If the event mediator hits a failure after sending messages to Kafka but before storing state, it is possible for the network to receive information that is out of date, as the flow that originally generated it will be replayed. This could particularly be an issue around transaction finalization.

### Solution

Swap the order of these operations.

By storing state first, the system will always have any information stored in the checkpoint before messages are sent elsewhere in the network.

Note that this is really only a workaround while a more thorough solution is developed.

### DIFF
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
@@ -167,7 +167,7 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
             var states = stateManager.get(messages.map { it.key.toString() }.distinct())
 
             while (groups.isNotEmpty()) {
-                val asynchronousOutputs = mutableListOf<MediatorMessage<Any>>()
+                val asynchronousOutputs = ConcurrentHashMap<String, MutableList<MediatorMessage<Any>>>()
                 val statesToCreate = ConcurrentHashMap<String, State?>()
                 val statesToUpdate = ConcurrentHashMap<String, State?>()
                 val statesToDelete = ConcurrentHashMap<String, State?>()
@@ -179,7 +179,7 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
                 }.map { group ->
                     taskManager.executeShortRunningTask {
                         // Process all same flow events in one go
-                        group.map { it ->
+                        group.map {
                             // Keep track of all records belonging to one flow
                             flowEvents.compute(it.key.toString()) { _, v ->
                                 if (v == null) {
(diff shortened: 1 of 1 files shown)


LWogan (dismissed):
LGTM as another tactical fix which has some known reliability impacts, but these should be rarer than the issues we see currently.

jujoramos (dismissed):

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
```diff
@@ -230,6 +229,7 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
                 } else {
                     listOf()
                 }
+                sendAsynchronousEvents(asynchronousOutputs)
```
mbrkic-r3: We shouldn't send out events related to states that failed to persist.

jujoramos (approved):

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
```diff
@@ -279,16 +280,21 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
...
+        busEvents: MutableMap<String, MutableList<MediatorMessage<Any>>>,
         queue: ArrayDeque<Record<K, E>>,
         event: Record<K, E>
     ) {
         val output = response.responseEvents.map { taskManagerHelper.convertToMessage(it) }
         output.forEach { message ->
             val destination = messageRouter.getDestination(message)
             if (destination.type == RoutingDestination.Type.ASYNCHRONOUS) {
-                busEvents.add(message)
+                busEvents.compute(key) { _, value ->
```
mbrkic-r3, line 293: Minor, might be more readable if replaced with `computeIfAbsent(key) { mutableListOf() }.add(message)`

mbrkic-r3 (approved):
LGTM

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
yift-r3: That won't be thread-safe

mbrkic-r3: It shouldn't be an issue assuming only one thread is processing events related to specific key.

JamesHR3: I'll leave it because the rebuild is expensive, but yes, good point!

thiagoviana (approved):

PR #5071, opened 2023-11-09
CORE-18072 CORE-18263 fix resolveStateRef logic and call

nkovacsx:
### Overview

This PR fixes two issues:
- `findSignedLedgerTransaction` explicitly called the `resolveStateRefs` function which resulted in opening two DB connections at the same time. This has been modified to a direct call to the repository instead.
- The `resolveStateRefs` function in the state query service uses a cache but we always fetch all of the state refs instead of just the non-cached ones.


lankydan (dismissed):

lankydan (approved):

blsemo (approved):
LGTM

PR #5076, opened 2023-11-10
CORE-18088 Race condition registering RPC endpoints in JavalinServer

mbrkic-r3:
Synchronized registration/removal of endpoints in `JavalinServer`.

### DIFF
--- a/libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
+++ b/libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
@@ -15,7 +15,6 @@ import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
-import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.locks.ReentrantLock
 import kotlin.concurrent.withLock
 
@@ -57,7 +56,7 @@ class JavalinServer(
     private val coordinator = coordinatorFactory.createCoordinator<WebServer> { _, _ -> }
     private val serverLock = ReentrantLock()
 
-    override val endpoints: MutableSet<Endpoint> = ConcurrentHashMap.newKeySet()
+    override val endpoints: MutableSet<Endpoint> = mutableSetOf()
 
     override fun start(port: Int) {
         serverLock.withLock {
--- a/libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
+++ b/libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
@@ -55,12 +55,12 @@ class JavalinServer(
     private val apiPathPrefix: String = "/api/${platformInfoProvider.localWorkerSoftwareShortVersion}"
     private var server: Javalin? = null
     private val coordinator = coordinatorFactory.createCoordinator<WebServer> { _, _ -> }
-    private val serverStartLock = ReentrantLock()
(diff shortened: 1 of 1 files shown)


vkolomeyko (dismissed):
LGTM

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
```diff
@@ -111,31 +111,35 @@ class JavalinServer(
...
-        if(null != server) registerEndpointInternal(endpoint)
-        // record the path in case we need to register when it's already started
-        endpoints.add(endpoint)
+        serverLock.withLock {
+            if (endpoints.any { it.path == endpoint.path && it.methodType == endpoint.methodType })
+                throw IllegalArgumentException("Endpoint with path ${endpoint.path} and method ${endpoint.methodType} already exists.")
+            // register immediately when the server has been started
+            if (null != server) registerEndpointInternal(endpoint)
+            // record the path in case we need to register when it's already started
+            endpoints.add(endpoint)
```
driessamyn, line 127: If we add the lock around this, then I don't think `endpoints` need to be a concurrent hashmap anymore?

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
```diff
@@ -111,31 +111,35 @@ class JavalinServer(
...
     override fun removeEndpoint(endpoint: Endpoint) {
-        endpoints.remove(endpoint)
-        // NOTE:
-        //  The server needs to be restarted to un-register the path. However, this means everything dependent on
-        //  this is impacted by a restart, which doesn't feel quite right.
-        //  This also means we can't really DOWN/UP the lifecycle status of this because this would end up in a
-        //  relentless yoyo-ing of this component as dependent components keep calling this function.
-        // TODO - review if it is really needed to de-register a path when a Subscription goes down, for example.
-        if(null != server) restartServer()
+        serverLock.withLock {
```
driessamyn, line 132: Does this not risk a deadlock given there is a lock around restartServer?

driessamyn (commented):
I think instead of this change, we just need an additional in `registerEndpointInternal` as `endpoints` is already thread safe, so the only thing missed out in the original PR, I think, is the lock for registering endpoints in a running server.

I think the current change risks a deadlock.

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
mbrkic-r3: yes, I've changed it

libs/web/web-impl/src/main/kotlin/net/corda/web/server/JavalinServer.kt
mbrkic-r3: Not sure how we could have a deadlock?

driessamyn: There is a lock around `restartServer` which is called inside here.
I thought it was called in a second place, which could have resulted in a deadlock, but it's not so it _should_ be ok.
However, I still think it's risky as somebody may make a chance that calls it from another place, and there isn't a reason for the double lock.
I think maybe the lock in restart can be removed.

mbrkic-r3: There is only one lock and it's re-entrant.

driessamyn (approved):
LGTM

PR #5078, opened 2023-11-10
CORE-18268 db-worker restart causes loss of REST API High Availability

kyriathar:
### Problem description

On forced Config reconciliation we would reconcile all config sections and therefore propagate them across the cluster no matter if their config value actually changed or not as part of the defaulting process. 

This caused a lot of components to restart, one of which is the http server within the rest-worker which causes its unavailability on db-worker restarts (Ideally components consuming configuration should check if the next configuration is changed compared to the previous one and only then restart).

### Solution

Contain the above Config propagation on the db-worker side. On forced reconciliation check if after applying the defaulting process to a config value, it changes compared to its Kafka corresponding one and only then reconcile it.

### Testing

Tested locally to an already deployed corda-cluster by: 
1. adding config properties to sections of the config schema 
2. rebuilding the db-worker against the updated `corda-api/data:config-schema`  and re-deploying the db-worker
3. assert the only config sections getting reconciled are only the extended ones (via the debugger and logs).

### DIFF
--- a/components/configuration/configuration-write-service-impl/src/main/kotlin/net/corda/configuration/write/impl/publish/ConfigPublishServiceImpl.kt
+++ b/components/configuration/configuration-write-service-impl/src/main/kotlin/net/corda/configuration/write/impl/publish/ConfigPublishServiceImpl.kt
@@ -107,7 +107,7 @@ class ConfigPublishServiceImpl @Activate constructor(
         TODO("Not yet implemented")
     }
 
-    override fun recordValueMisalignmentAfterDefaulted(
+    override fun recordValueMisalignedAfterDefaulted(
         recordKey: String,
         dbRecordValue: Configuration,
         kafkaRecordValue: Configuration
--- a/components/configuration/configuration-write-service-impl/src/main/kotlin/net/corda/configuration/write/impl/publish/ConfigPublishServiceImpl.kt
+++ b/components/configuration/configuration-write-service-impl/src/main/kotlin/net/corda/configuration/write/impl/publish/ConfigPublishServiceImpl.kt
@@ -139,8 +139,8 @@ class ConfigPublishServiceImpl @Activate constructor(
             }
 
 
-        val configsAreEqual = dbConfigValueDefaulted == kafkaConfigValue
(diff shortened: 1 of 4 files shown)


components/reconciliation/reconciliation/src/main/kotlin/net/corda/reconciliation/ReconcilerWriter.kt
```diff
@@ -18,4 +18,15 @@ interface ReconcilerWriter<K : Any, V : Any> {
     fun remove(recordKey: K)
 
     val lifecycleCoordinatorName: LifecycleCoordinatorName
+
```
simon-johnson-r3, line 21: Some writers allow the checking of records to see if the value content has changed between the db and Kafka before pushing out a new value.

components/configuration/configuration-write-service-impl/src/main/kotlin/net/corda/configuration/write/impl/publish/ConfigPublishServiceImpl.kt
```diff
@@ -102,6 +107,56 @@ class ConfigPublishServiceImpl @Activate constructor(
         TODO("Not yet implemented")
     }
 
+    override fun recordValueMisalignedAfterDefaults(
```
simon-johnson-r3: valuesMisalignedAfterDefaults

components/reconciliation/reconciliation/src/main/kotlin/net/corda/reconciliation/ReconcilerWriter.kt
kyriathar: Applied

components/configuration/configuration-write-service-impl/src/main/kotlin/net/corda/configuration/write/impl/publish/ConfigPublishServiceImpl.kt
kyriathar: Applied

simon-johnson-r3 (approved):

PR #5079, opened 2023-11-12
ES-1608: fix various version and branch refrences to point to 5.2

ronanbrowne:
- fix incorrect helm values in Jenkinsfile variables should reference 5.2
- Update various other missed references in supporting files which should reference 5.2 rather than 5.1


wzur-r3 (approved):
LGTM

dimosr (approved):

PR #5085, opened 2023-11-13
CORE-18311: Eliminate flakiness during CPI upload

vkolomeyko:
Since initial timeout has been reduced to 10ms, it may not be enough time for CPI to filter through all the caches and be found. Hence retrying a bit longer to allow CPI metadata to propagate.


Tom-Fitzpatrick (approved):
LGTM

PR #5090, opened 2023-11-13
CORE-18282: Fix Flaky Tests

jujoramos:
Sort the returned states before building the filters to use so that the
lower bound is always older in time than the upper bound.


jujoramos:
I've executed the test 283 times in a row with the changes, no failures whatsoever.

vkolomeyko (approved):
LGTM

PR #5094, opened 2023-11-14
CORE-10180: Escape string parameters when creating json to prevent possible injection attacks

anien:
Changes:
* Escape string parameters
* Fix most of the kdoc parameters
* Add tests

### DIFF
--- a/tools/plugins/initial-config/src/test/kotlin/net/corda/cli/plugin/initialconfig/TestInitialConfigPluginDb.kt
+++ b/tools/plugins/initial-config/src/test/kotlin/net/corda/cli/plugin/initialconfig/TestInitialConfigPluginDb.kt
@@ -152,7 +152,6 @@ class TestInitialConfigPluginDb {
         val outText = createConfigDbConfig(jdbcUrl, username, password, vaultKey, jdbcPoolMaxSize, jdbcPoolMinSize,
             idleTimeout, maxLifetime, keepaliveTime, validationTimeout, secretsService)
 
-        println(outText)
         assertThat(outText).contains("\"user\":\"test\\\"user\"")
     }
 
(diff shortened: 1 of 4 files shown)


tools/plugins/initial-config/src/test/kotlin/net/corda/cli/plugin/initialconfig/TestInitialConfigPluginDb.kt
```diff
@@ -130,4 +132,54 @@ class TestInitialConfigPluginDb {
...
+        val password = ""
+        val salt = "123"
+        val passphrase = "123"
+        val vaultKey = "corda-config-database-password"
+        val secretsService: SecretsCreateService = EncryptionSecretsServiceImpl(passphrase, salt)
+
+        val outText = createConfigDbConfig(jdbcUrl, username, password, vaultKey, jdbcPoolMaxSize, jdbcPoolMinSize,
+            idleTimeout, maxLifetime, keepaliveTime, validationTimeout, secretsService)
+
+        println(outText)
```
dickon: debug code to delete?

dickon (dismissed):
There's a println that should get removed, and it feels like a bit of extra test coverage would be worth it.

Also I wish we were using a function in a library we already depend on for this, though I don't know of one.

ronanbrowne (dismissed):
LGTM (build file only reviewed)

tools/plugins/initial-config/src/test/kotlin/net/corda/cli/plugin/initialconfig/TestInitialConfigPluginDb.kt
anien: Done

anien:
> There's a println that should get removed, and it feels like a bit of extra test coverage would be worth it.
> 
> Also I wish we were using a function in a library we already depend on for this, though I don't know of one.

Println removed. I am happy to talk about the test coverage. As there is a difference of the checks/escaping depending on how the code is called, I covered both scenarios in the new tests. Not sure what else to add.

dickon (approved):
LGTM, same comments as before

ronanbrowne (approved):
LGTM

PR #5096, opened 2023-11-14
CORE-18254: Publish batches of records asynchronously

vkolomeyko:
Previously when starting flows, there was a lock in place and process of sending records has been performed in the calling thread. This caused:
- Block the current (REST) thread execution.
- Blocked other (REST) thread whilst waiting on a lock.

A dedicated executor service should improve the situation. Also minor changes to KDocs and unit tests.

### Checks made to prove the change is good

- E2E test run from this PR: https://ci02.dev.r3.com/job/Corda5/view/Full%20Health%205.2/job/corda5-e2e-tests/job/release%252F5.2/26/
- Performance test: https://ci02.dev.r3.com/job/Corda5/job/corda-performance-tests/job/release%2F5.2/3/
- Simulations test: https://ci02.dev.r3.com/job/Corda5/job/corda-simulations/job/release%252F5.2/6/


vkolomeyko:
Started an E2E run with this changes: https://ci02.dev.r3.com/job/Corda5/view/Full%20Health%205.2/job/corda5-e2e-tests/job/release%252F5.2/18/

vkolomeyko:
@JamesHR3 @jujoramos - would you know if there is a performance test Jenkins configuration for me run with binaries from this PR to confirm that it is not having any adverse effect on the performance?
I just need a scenario where we start many flows in parallel, the type of CorDapp executed is probably not that important.

components/flow/flow-rest-resource-service-impl/src/main/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImpl.kt
```diff
@@ -208,28 +205,33 @@ class FlowRestResourceImpl @Activate constructor(
...
-                        FlowRestExceptionConstants.NON_FATAL_ERROR,
-                        mapOf("cause" to ex::class.java.simpleName, "reason" to msg)
-                    )
+            
+            // Do not block REST thread execution till future completes, instead add a hook to log an error if batch
+            // publication fails for whatever reason and return to the REST caller that flow start been accepted.
+            // Should they wish to check the actual execution progress, they can always check the status using
+            // client request id provided.
+            batchFuture.exceptionally { 
+                log.error("Failed to publish start flow batch for flowClass: $flowClassName, " +
```
LWogan: Should this be warn? I think there's a case to say that a lot of the error logging in this file are actually warns.
https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4169859082/C5+Logging+goals

LWogan (commented):
These changes look good to me

components/flow/flow-rest-resource-service-impl/src/main/kotlin/net/corda/flow/rest/impl/v1/FlowRestResourceImpl.kt
vkolomeyko: Sure, makes sense. I will change to `WARN`.

LWogan (approved):
LGTM

PR #5097, opened 2023-11-14
CORE-16215: Enhance db utils sql server support

blsemo:
Modify the integration testing framework so it can work with other databases, and is able to run the db/osgi-integration-tests against MS SqlServer.


testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/PostgresHelper.kt
```diff
@@ -0,0 +1,92 @@
...
+            jdbcUrlCopy,
+            user,
+            password,
+            maximumPoolSize = maximumPoolSize
+        )
+    }
+
+
+    override fun createConfig(
+        inMemoryDbName: String,
```
sprintengine, line 73: any way to get rid of `inMemoryDbName` from the interface so the implementations don't all have this?

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/DbUtilsHelper.kt
```diff
@@ -0,0 +1,44 @@
...
+interface DbUtilsHelper {
+    fun getDatabase(): String
+
+    fun getAdminUser(): String
+
+    fun getAdminPassword(): String
+
+    @Suppress("LongParameterList")
+    fun getEntityManagerConfiguration(
+        inMemoryDbName : String = "",
```
sprintengine, line 19: Probably similar here, would be nice if all implementations didn't have to give a inMemoryDbName parameter

build.gradle
```diff
@@ -245,23 +245,34 @@ allprojects {
...
-        inputs.property('postgresPassword', postgresPassword)
+        // Conversion of legacy arguments to keep old Jenkins pipeline happy
+        if (!postgresPort.isBlank()){
+            databasePort = postgresPort
+            databaseType = 'POSTGRES'
+            databaseName = postgresDb
+            databaseHost = postgresHost
+            databaseAdminUser = postgresUser
+            databaseAdminPassword = postgresPassword
+        }
```
jujoramos, line 256: Any chance we can entirely remove the legacy approach and update the old pipelines instead?.

gradle.properties
```diff
@@ -149,12 +150,25 @@ systemLambdaVersion=1.2.1
...
+
+# Legacy properties to stay compatible with the Jenkins pipeline. When setting the postgresPort, it will
+# switch to using a postgres database. Either this mechanism or the generic one above can be used, if setting
+# both postgres... and database... properties, they might interact badly.
 postgresPort=
-postgresDb=
+postgresHost=
 postgresUser=
 postgresPassword=
+postgresDatabase=
```
jujoramos, line 166: Any chance we can entirely remove the legacy approach and update the old pipelines instead?.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -43,7 +43,7 @@ import org.mockito.kotlin.mock
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class StateManagerIntegrationTest {
     private val maxConcurrentThreadJdbcConnections = 10
-    private val dataSource = DbUtils.createPostgresDataSource(maximumPoolSize = maxConcurrentThreadJdbcConnections)
+    private val dataSource = DbUtils.createDataSource(maximumPoolSize = maxConcurrentThreadJdbcConnections)
```
sprintengine, line 46: State Manager has a query provider, currently using `PostgresQueryProvider`, line 65.

So this test likely needs a bit more work to a) implement queries for other databases and b) select appropriate query provider based on similar selection on DBTYPE_PROPERTY.

I think we have a separate ticket for that - it can be handled as part of https://r3-cev.atlassian.net/browse/CORE-16663

vkolomeyko:
Is https://r3-cev.atlassian.net/browse/CORE-16125 the right Jira for this work?
(For me this Jira cannot be found at all)

build.gradle
driessamyn: I agree it would be nice to clean this up, but. wonder if a 2 phase approach would make sense. I.e.:
- Get this change in
- Update the Jenkins pipeline (5.2 version only so old pipelines still work)
- Remove the legacy support from 5.2

driessamyn (dismissed):
LGTM
As mentioned in the comment, maybe we need a follow-up task to update the pipelines and remove the legacy way of setting the DB details.

vkolomeyko (dismissed):

libs/db/readme.md
```diff
@@ -42,27 +42,64 @@ When used in an OSGi context, the following bundles are required
...
-* If integration test has to be executed repeatedly `clean` or `cleanTestOSGi` target has to be executed before `integrationTest`
-or else Gradle may skip execution.
+#### Running against Sqlserver
+
+Start the MS Sqlserver container:
+
+```bash
+docker run -e "ACCEPT_EULA=Y" -e "MSSQL_SA_PASSWORD=yourStrong(!)Password" -p 1433:1433 -d mcr.microsoft.com/mssql/server:2022-latest
+```
+It is recommended to create a test database in the database server using the database tool of your choice - this example
```
simon-johnson-r3, line 95: Why is it recommended but not mandatory? Without knowing the consequences of not doing it I'd not know how to make a choice, so I'd probably not bother.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/AbstractDBHelper.kt
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * The reason for this abstract class is to avoid a lot of code duplication that would've happened in the
+ * PostgresHelper and SQLServerHelper classes
+ *
+ * @property port The port on which the database server is running.
+ * @property host The hostname of the database server.
+ * @property jdbcUrl The JDBC URL used to connect to the database.
+ * @property driverClass The name of the JDBC driver to be used.
+ */
+abstract class AbstractDBHelper : DbUtilsHelper {
```
simon-johnson-r3: Having an interface and an abstract base class seems a little overkill.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/DbUtilsHelperFunctions.kt
```diff
@@ -0,0 +1,17 @@
+package net.corda.db.testkit.dbutilsimpl
+
+
+object DbUtilsHelperFunctions {
+    /**
+     * This function is very similar to System.getProperty(key, defaultValue), but it also
+     * applies the default value to blank properties.
+     */
```
simon-johnson-r3: this can/should be in the abstract base class I think?

LWogan (dismissed):
Looks ok from flow perspective

conal-smith (dismissed):
LGTM state-manager

libs/db/readme.md
ifrankrui: When you don't specify a test database for MSSQL, the test connects to a system database like `master` or `model` by default. I would prefer to make the testing database mandatory. The same approach should be applied to PostgreSQL.

ronanbrowne (dismissed):
LGTM (Blt)

build.gradle
blsemo: What Dries said - I tried to go for the "let's change the pipeline approach" but moved to this after talking to Ronan.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
blsemo: Most tests have not been upgraded to work with anything but Postgres and HSQL - only the db OSGI integration tests were used as a guinea pig to run against SQL server. This PR is just about getting the infrastructure into place, the actual change to support other databases is a separate issue.

libs/db/readme.md
blsemo: Absolutely, but not in this change.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/AbstractDBHelper.kt
blsemo: I can explain that. The structure is like this:

```
interface DbUtilsHelper
^                    ^
|                    |
InMemoryDbHelper    AbstractDBHelper
                     ^             ^
                     |             |
          PostgresHelper      SqlServerHelper
```

simon-johnson-r3: Ok fair enough. Perhaps `AbstractDBHelper` could be better named something like `ExternalDbHelper` but it's not exactly urgent.

I suggest you make the case of the letter 'b' consistent though? I just noticed it's uppercase here and lower in the interface.

testing/db-testkit/src/main/kotlin/net/corda/db/testkit/dbutilsimpl/PostgresHelper.kt
blsemo: Not easy - I'd rather not solve that problem here.

vkolomeyko (approved):
LGTM

ifrankrui (approved):

lankydan (approved):

yift-r3 (approved):
The network changes LGTM.

LWogan (approved):

mcgovc (approved):
LGTM - build file changes

ronanbrowne (approved):

simon-johnson-r3 (approved):

PR #5105, opened 2023-11-15
CORE-18352 Fix app-simulator optional database parameters 

YashNabar:
The app-simulator’s database parameters are optional when running in SENDER mode. This change fixes an issue where the Sender fails to start if these are not provided.


dimosr (approved):
LGTM

PR #5107, opened 2023-11-15
CORE-17917 Adding handling to MessageBusClient for async kafka errors

ben-millar:
A bug was observed whereby Kafka producer errors were not being correctly reported back to the `MultiSourceEventMediator`. This is because our Kafka messages are essentially 'fire-and-forget'; when something goes wrong, Kafka will exceptionally close a future in the background, but we never attempted to capture or track this future.

This PR adds logging around this type of issue to make it clear to the user that something has gone wrong, and passes a lambda to the `CordaProducer` allowing reporting back of any Kafka exceptions that occur. This `future` is currently not utilized in the mediator; that will be covered in an upcoming mediator refactor.

Currently, all Kafka exceptions are wrapped in a `CordaMessageAPIFatalException` as Kafka is already configured to retry intermittent/retryable errors internally. In the future, we may choose to wrap some of these as `CordaMessageAPIIntermittentException`s regardless.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -46,7 +46,7 @@ commonsTextVersion = 1.10.0
 bouncycastleVersion=1.76
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.1.0.xx-SNAPSHOT to pick up maven local published copy
-cordaApiVersion=5.1.0.37-beta+
+cordaApiVersion=5.1.0.37-alpha-1700043137966
 
 disruptorVersion=3.4.4
 felixConfigAdminVersion=1.9.26
--- a/gradle.properties
+++ b/gradle.properties
@@ -46,7 +46,7 @@ commonsTextVersion = 1.10.0
 bouncycastleVersion=1.76
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.1.0.xx-SNAPSHOT to pick up maven local published copy
-cordaApiVersion=5.1.0.37-alpha-1700137352553
+cordaApiVersion=5.1.0.37-alpha-1700151714786
 
 disruptorVersion=3.4.4
 felixConfigAdminVersion=1.9.26
--- a/gradle.properties
+++ b/gradle.properties
@@ -46,7 +46,7 @@ commonsTextVersion = 1.10.0
 bouncycastleVersion=1.76
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.1.0.xx-SNAPSHOT to pick up maven local published copy
-cordaApiVersion=5.1.0.37-alpha-1700058741660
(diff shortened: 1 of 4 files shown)


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MessageBusClient.kt
```diff
@@ -18,9 +21,44 @@ class MessageBusClient(
...
+        exception: Exception?,
+        future: CompletableFuture<Unit>,
+        topic: String
+    ) {
+        when (exception) {
+            null -> future.complete(Unit)
+            else -> {
+                val baseMessage = "Producer clientId $id for topic $topic failed to send."
+                val errorMessage = when (exception) {
+                    is CordaMessageAPIIntermittentException -> baseMessage
```
LWogan: I think the exception here is always going to be a kafka client one and never  a CordaMessageAPIIntermittentException or CordaMessageAPIFatalException?
If thats the case I think you'd need to rework this method a little. 

But we're not doing anything with the future. Its not even being assigned/returned from the MessageBusClient.kt, As you already mentioned we need to follow up later to decide what action if any we want to take with future failures. So i'd be ok with just logging it for now and figuring out how we'll set the futures at a later point.

gradle.properties
```diff
@@ -46,7 +46,7 @@ commonsTextVersion = 1.10.0
 bouncycastleVersion=1.76
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.1.0.xx-SNAPSHOT to pick up maven local published copy
-cordaApiVersion=5.1.0.37-beta+
+cordaApiVersion=5.1.0.37-alpha-1700151714786
```
LWogan: why do you need this?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MessageBusClient.kt
```diff
@@ -18,9 +20,31 @@ class MessageBusClient(
...
+        exception: Exception?,
+        future: CompletableFuture<Unit>,
+        topic: String
+    ) {
+        if (exception == null) {
+            future.complete(Unit)
+        } else {
+            val message = "Producer clientId $id for topic $topic failed to send."
+            log.warn(message, exception)
+            future.completeExceptionally(CordaMessageAPIFatalException(message, exception))
```
LWogan, line 46: This is an acceptable starting position given, like you explained, that the Kafka client retries internally before setting the future with an exception for any exceptions which are retriable.

LWogan (approved):
LGTM. Logging the error is the primary thing we're after for 5.1 and this PR achieves that. We can look at what we'll do with these exceptions in 5.2

thiagoviana (approved):

PR #5111, opened 2023-11-16
CORE-18370 Fix `findUnconsumedStatesByExactType`

lankydan:
The query for `findUnconsumedStatesByExactType` did not actually filter by `consumed IS NULL` so it returned all states regardless of if they were consumed or not.

Added `consumed IS NULL` to fix this.


blsemo (approved):
LGTM

nkovacsx (approved):

PR #5112, opened 2023-11-16
add rewrap a private signing key method

dickon:


components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/SigningRepository.kt
```diff
@@ -17,9 +17,15 @@ interface SigningRepository : Closeable {
...
+     * @param context data about the key store (the DTO)
+     * @param replace allow keys to be replaced in the database normally this is used
+     *                to add keys. If this is set it is permitted to replace the
+     *                records for a key pair. There's no problem setting replace to true
+     *                when adding a fresh key.
+     *
+     * @throws [IllegalStateException] if the key already exists and replace is false.
      */
-    fun savePrivateKey(context: SigningWrappedKeySaveContext): SigningKeyInfo
+    fun savePrivateKey(context: SigningWrappedKeySaveContext, replace: Boolean = false): SigningKeyInfo
```
anien, line 29: Does this mean that if the private key is saved with replace = false, we won't be able to rewrap it? If yes, how are we going to handle upgrades if the keys from previous versions are non-replaceable?

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt
```diff
@@ -595,6 +595,29 @@ open class SoftCryptoService(
             return newGeneration
         }
     }
+
+    override fun rewrapSigningKey(tenantId: String, publicKey: PublicKey, newWrappingKeyAlias: String): SigningKeyInfo {
```
anien: Is there a reason why we use `publicKey` and not `alias`? `signingRepo.findKey(alias)` is supported too.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/CryptoService.kt
```diff
@@ -314,4 +314,23 @@ interface CryptoService {
      *         in the configuration unmanaged keys map or targetAlias is not found
      */
     fun rewrapWrappingKey(tenantId: String, targetAlias: String, newParentKeyAlias: String): Int
+
+    /**
+     * Rotate the encryption of a private key withing a signing key itself, without changing the private key itself,
```
anien, line 320: typo withing -> within

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/SigningRepository.kt
dickon: Rewrapping a key is not the same thing as replacing it. Rewrapping changes the key material to use a different wrapping key, the crypto_wrapping_key row would not change.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt
dickon: The database constaints and column setup do not guarnatee that alias uniquely selects keypairs. Public Key does.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/CryptoService.kt
dickon: indeed, but I guess we'll drop this branch now so I won't take the time to fix the typo

PR #5113, opened 2023-11-16
CORE-18375 Removing unused wrapped kotlinx.coroutines library

ben-millar:
The `kotlinx.coroutines` library was wrapped as an OSGi module within Corda as it was planned to be used within the newly developed multi-source event mediator, but this was later changed. As a result, we should remove the unused code.


mbrkic-r3:
I think this should be removed as well (it was added with https://github.com/corda/corda-runtime-os/commit/0470a7e9e55ec3b504ffdf9995538f5fa759d41a):
https://github.com/corda/corda-runtime-os/blob/release/os/5.1/settings.gradle#L541
https://github.com/corda/corda-runtime-os/blob/release/os/5.1/buildSrc/src/main/groovy/corda.common-library.gradle#L35

ben-millar:
> I think this should be removed as well (it was added with [0470a7e](https://github.com/corda/corda-runtime-os/commit/0470a7e9e55ec3b504ffdf9995538f5fa759d41a)): https://github.com/corda/corda-runtime-os/blob/release/os/5.1/settings.gradle#L541 https://github.com/corda/corda-runtime-os/blob/release/os/5.1/buildSrc/src/main/groovy/corda.common-library.gradle#L35

Good catch, will remove the first. The latter is already removed in this revision.

ronanbrowne (approved):
LGTM

PR #5117, opened 2023-11-16
CORE-12165 prep feature branch for merge

dickon:
- Remove state manager that will be reviewed in a separate PR.
- Fix tests.
- Use data class for key derivation parameters.


components/crypto/crypto-service-impl/src/test/kotlin/net/corda/crypto/service/impl/infra/TestServicesFactory.kt
```diff
@@ -38,7 +38,7 @@ import kotlin.test.assertEquals
  * Provide instances of high level crypto services, with no database underneath, for
  * use for integration test cases that don't involve in-memory databases.
  */
-class TestServicesFactory {
+class     TestServicesFactory {
```
anien: We should revert this white space change.

dickon: done

anien (dismissed):
LGTM!

anien (approved):
LGTM.

PR #5125, opened 2023-11-16
CORE-13442: added logs for custom query

jennyang-r3:
Added more information to logs for custom queries in order to give more information when debugging.

### DIFF
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
@@ -206,11 +206,17 @@ class VaultNamedQueryExecutorImpl(
             log.trace { "Executing try: $currentRetry, fetched ${filteredRawData.size} number of results so far." }
 
             // Fetch the state and refs for the given transaction IDs
-            val rawResults = fetchStateAndRefs(
-                request,
-                vaultNamedQuery.query.query,
-                currentResumePoint
-            )
+            val rawResults = try {
+                fetchStateAndRefs(
+                    request,
+                    vaultNamedQuery.query.query,
+                    currentResumePoint
+                )
+            } catch (e: Exception) {
+                log.warn("Failed to query \"${request.queryName}\" " +
+                        "with parameters \"${deserializedParams}\" limit \"${request.limit}\".", e)
+                throw e
+            }
 
             // If we have no filter, there's no need to continue the loop
             if (vaultNamedQuery.filter == null) {
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
@@ -278,7 +278,7 @@ class UtxoPersistenceServiceImpl(
                     .ifBlank { "{}" } // Default to "{}" if the provided factory returns empty string to avoid exception
             } catch (e: Exception) {
                 // We can't log the JSON string here because the failed before we have a JSON
-                log.warn("Error while processing factory for class: ${it.stateType.name}. Defaulting to empty JSON.")
+                log.warn("Error while processing factory for class: ${it.stateType.name}. Defaulting to empty JSON.", e)
                 "{}"
             }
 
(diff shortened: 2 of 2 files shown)


components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/UtxoPersistenceServiceImpl.kt
```diff
@@ -278,7 +278,7 @@ class UtxoPersistenceServiceImpl(
                     .ifBlank { "{}" } // Default to "{}" if the provided factory returns empty string to avoid exception
             } catch (e: Exception) {
                 // We can't log the JSON string here because the failed before we have a JSON
-                log.warn("Error while processing factory for class: ${it.stateType.name}. Defaulting to empty JSON.")
+                log.warn("Error while processing factory for class: ${it.stateType.name}. Defaulting to empty JSON.", e)
```
lankydan, line 281: I'd had another look through what we log here, and I think it might make sense for the other try catches below this to also log the exception. Errors are unlikely to occur in them; however, that means its even more important to know what happen when it does fail.

The try catches im talking about are the ones below.

```kotlin
it.stateType.name to try {
                jsonMarshallingService.parse(jsonToParse, Any::class.java)
            } catch (e: Exception) {
                log.warn("Error while processing factory for class: ${it.stateType.name}. " +
                        "JSON that could not be processed: $jsonToParse. Defaulting to empty JSON.")
                jsonMarshallingService.parse("{}", Any::class.java)
            }
        }.toMutableMap()
        try {
            jsonMap[ContractState::class.java.name] = jsonMarshallingService.parse(
                defaultContractStateVaultJsonFactory.create(stateAndRef, jsonMarshallingService),
                Any::class.java
            )
        } catch (e: Exception) {
            log.warn("Error while processing factory for class: ${ContractState::class.java.name}. Defaulting to empty JSON.")
            jsonMarshallingService.parse("{}", Any::class.java)
        }
        return try {
            jsonMarshallingService.format(jsonMap)
        } catch (e: JsonProcessingException) {
            // Since we validate the factory outputs one-by-one this should not happen.
            log.warn("Error while formatting combined JSON, defaulting to empty JSON.")
            "{}"
        }
```

@nkovacsx what do you think? I know I probably made you not include the exceptions the first time we wrote this but I think its worth including them now i think more about it.

nkovacsx: Yeah I agree. Even more, I had cases when this failed and it was so hard to trace it back because the exception will contain which part of the JSON was malformed etc.

blsemo: Yes, please log the exceptions! That will make finding problems in this area so much easier, and should not add any overhead on the happy path.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
```diff
@@ -206,11 +206,17 @@ class VaultNamedQueryExecutorImpl(
...
-            )
+            val rawResults = try {
+                fetchStateAndRefs(
+                    request,
+                    vaultNamedQuery.query.query,
+                    currentResumePoint
+                )
+            } catch (e: Exception) {
+                log.warn("Failed to query \"${request.queryName}\" " +
+                        "with parameters \"${deserializedParams}\" limit \"${request.limit}\".", e)
```
nkovacsx, line 217: Did you verify that `${deserializedParams}` actually looks readable in the logs and not just an ugly reference to the object? I don't know off the top of my head how maps are logged by default.

jennyang-r3: That's a good question. Let me check.

lankydan: Maps normally output key values when tostring is called iirc

lankydan (approved):

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/query/execution/impl/VaultNamedQueryExecutorImpl.kt
jennyang-r3: The deserializedParams in log looks like:
```kotlin
"{type=com.r3.corda.demo.utxo.contract.TestUtxoQueryState2, Corda_TimestampLimit=2023-11-20T14:59:42.214694Z}"
```

nkovacsx: Yeah that's perfect thanks for double checking

PR #5126, opened 2023-11-17
CORE-12165: merge key rotation feature work

dickon:
Take new API update with additional Avro, kafka and config for crypto key rotation.

Add new RestResource called KeyRotationRestResource which is automatically injected into the REST procesosr and allows crypto unmanaged wrapping key rotation to be started.

Implement unmanaged key rotation, for now in a relatively simple way where we take work requests on Kafka then process them to completion. This won't necessarily be the best way for the database at scale, and we have a design for spreading the work out over a nested Kafka topic, but let's get something in that is simple and works first.

Add alias to WrappingKeyInfo. It's not clear why this wasn't here before. WrappingKeyInfo is a data transfer object (DTO) and it seems reasonable for it to hide DB implementation detals such as the synthetic primary ID field. Alias is also part of a compound key with generation, and both alias and generation are user-visible features of wrapping keys; for instance a user could specify wrapping key alias when creating  a signing key pair.

Making WrappingKeyEntity key_material updatable since that's what happens during key rotation.

Add state manager arguments to the crypto worker, so that later the crypto worker can use the state manager for status notifications.


components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
```diff
@@ -0,0 +1,96 @@
...
+import net.corda.rest.response.ResponseEntity
+
+/**
+ * Key Rotation API consists of endpoints to request a key rotation operation and to check the status of such operation,
+ */
+@HttpRestResource(
+    name = "Key Rotation API",
+    description = "Contains operations related to rotation of the wrapping keys.",
+    path = "wrappingkey",
+    minVersion = RestApiVersion.C5_1
```
driessamyn: This should be `C5_2` now

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
```diff
@@ -0,0 +1,96 @@
...
+        responseDescription = "",
+    )
+    fun getKeyRotationStatus(): List<Pair<String, List<String>>>
+
+    /**
+     * Initiates the key rotation process. 
+     *
+     * @param oldKeyAlias Alias of the key to be rotated.
+     * @param newKeyAlias Alias of the new key the [oldKeyAlias] key will be rotated with.
+     * @param simulate If true, only work out the number of keys to rotate.
```
driessamyn: I thought simulation was out of scope?

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
```diff
@@ -0,0 +1,96 @@
...
+     * @param timeToLive The maximum number of milliseconds which should pass, after which the key rotation loop will exit.
+     * @param limit The maximum number of key rotations to do, after which the key rotation loop will exit.
```
driessamyn: Do we need these given the design is for each rotation to be handled separately?

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -0,0 +1,157 @@
...
+    override fun getKeyRotationStatus(): List<Pair<String, List<String>>> {
+        TODO("Not yet implemented")
+    }
+
+    override fun startKeyRotation(
+        oldKeyAlias: String,
+        newKeyAlias: String,
+        simulate: Boolean,
+        timeToLive: Int,
+        limit: Int
```
driessamyn: These are said to be optional in the docs, but don't have a default value.
(But see comment above about whether we actually need them).

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -0,0 +1,157 @@
...
+            null,
+            simulate,
+            timeToLive,
+            limit
+        )
+
+        publisher?.publish(listOf(Record(uploadTopic, requestId, keyRotationRequest)))
+            ?: throw ServiceUnavailableException("Key rotation resource has not been initialised.")
+
+        return ResponseEntity.accepted(KeyRotationResponse(requestId, oldKeyAlias, newKeyAlias))
```
driessamyn, line 95: Doesn't this need to wait for the future to complete (and handle a potential exception)?

components/crypto/crypto-rest/src/test/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceTest.kt
```diff
@@ -0,0 +1,177 @@
...
+
+//    @Test
+//    fun `get key rotation status`() {
+//        TODO("Not yet implemented")
+//    }
+//
+//    @Test
+//    fun `get key rotation status for unknown requestID throws`() {
+//        TODO("Not yet implemented")
+//    }
```
driessamyn: I am assuming these will be filled in as part of the status work?

Instead it may be better to add them in and `@Disable` them with a reference to the JIRA.
Disabled tests are a bit more visible than ones commented out.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -0,0 +1,65 @@
...
+    companion object {
+        private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+    }
+
+    override val keyClass: Class<String> = String::class.java
+    override val valueClass = KeyRotationRequest::class.java
+
+    @Suppress("NestedBlockDepth")
+    override fun onNext(events: List<Record<String, KeyRotationRequest>>): List<Record<*, *>> {
+        logger.info("received ${events.size} key rotation requests")
```
driessamyn: This should be debug level logging, not `info`.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -0,0 +1,65 @@
...
+        events.map { it.value}.filterNotNull().forEach { request ->
+            logger.info("processing $request")
```
driessamyn: As above, change to `debug` level logging.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -0,0 +1,65 @@
...
+            // root (unmanaged) keys can be used in clusterDB and vNodeDB
+            // then for each key we will send a record on Kafka
+            // We do not have a code that deals with rewrapping the root key in a cluster DB
+
+            // tenantId in the request is useful ONLY for re-wrapping managed keys!
+            // For unmanaged (root) key we need to go through all vNodes and check if the oldKeyAlias is used there
+            // if yes, then we need to issue a new record for this key to be re-wrapped
+            // we get the tenantId from the vNode info (I hope).
```
driessamyn: Not necessarily because tenants are not always VNodes (they can be "P2P" for example, IIRC).

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -0,0 +1,65 @@
...
+
+            val virtualNodeInfo = virtualNodeInfoReadService.getAll() // Get all the virtual nodes
+            val virtualNodeTenantIds = virtualNodeInfo.map { it.holdingIdentity.shortHash.toString() }
+
+            val allTenantIds = virtualNodeTenantIds+listOf(CryptoTenants.CRYPTO, CryptoTenants.P2P, CryptoTenants.REST)
```
driessamyn: Should we have a central list of "all" non-vnode tenants?
Or use the `allClusterTenants` function?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -0,0 +1,65 @@
...
+            logger.info("Found ${allTenantIds.size} tenants; first few are: ${allTenantIds.take(10)}")
```
driessamyn: Remove or change to debug level.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -0,0 +1,65 @@
...
+            val targetWrappingKeys = allTenantIds.asSequence().map { tenantId ->
+                wrappingRepositoryFactory.create(tenantId).use { wrappingRepo ->
+                    wrappingRepo.findKeysWrappedByAlias (request.oldKeyAlias).map { wki -> tenantId to wki}
+                }
+            }.flatten()
+
+            val truncatedWrappingKeys = targetWrappingKeys.take(request.limit ?: Int.MAX_VALUE)
```
driessamyn: Why is the limit useful here?
I feel this is over-engineered and will cause end-user confusion, and call YAGNI on it.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRewrapBusProcessor.kt
```diff
@@ -0,0 +1,52 @@
...
+            val value = createRewrapResponse(
+                    request!!.requestId,
+                    request.tenantId,
+                    request.oldKeyAlias,
+                    request.newKeyAlias,
+                    endTimestamp
+                )
+            records.add(Record(REWRAP_MESSAGE_RESPONSE_TOPIC, request.requestId, value))
+        }
+        return records
```
driessamyn: Is this to be replaced with the State Manager?

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/WrappingRepository.kt
```diff
@@ -19,7 +19,7 @@ interface WrappingRepository : Closeable {
      * @param key The key material and metadata about version and algorithm.
      * @return The wrapping key that was persisted.
      */
-    fun saveKey(alias: String, key: WrappingKeyInfo): WrappingKeyInfo
+    fun saveKey(key: WrappingKeyInfo): WrappingKeyInfo
```
driessamyn, line 21: Remove alias from the KDoc

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/WrappingRepository.kt
```diff
@@ -30,7 +30,7 @@ interface WrappingRepository : Closeable {
      *           otherwise it creates a new entry with random id.
      * @return The wrapping key that was persisted.
      */
-    fun saveKeyWithId(alias: String, key: WrappingKeyInfo, id: UUID?): WrappingKeyInfo
+    fun saveKeyWithId(key: WrappingKeyInfo, id: UUID?): WrappingKeyInfo
```
driessamyn, line 31: Remove alias from the KDoc

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoService.kt
```diff
@@ -579,15 +580,19 @@ open class SoftCryptoService(
             val oldParentKey = checkNotNull(unmanagedWrappingKeys.get(wrappingKeyInfo.parentKeyAlias)) {
                 "Unable to find parent key ${wrappingKeyInfo.parentKeyAlias} in the configured unmanaged wrapping keys"
             }
+            val newGeneration = wrappingKeyInfo.generation + 1
```
driessamyn: Is a race condition possible here in case `wrappingKeyInfo` is out of date?
Do we need optimistic locking?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -0,0 +1,65 @@
...
+class CryptoRekeyBusProcessor(
+    val cryptoService: CryptoService,
+    private val virtualNodeInfoReadService: VirtualNodeInfoReadService,
+    private val wrappingRepositoryFactory: WrappingRepositoryFactory,
+) : DurableProcessor<String, KeyRotationRequest> {
+    companion object {
+        private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+    }
+
+    override val keyClass: Class<String> = String::class.java
```
driessamyn, line 31: If I read it correctly, I think the key here is tenantId.
Does this make sense?
The implication is that all requests for a particular tenant ID will be going to the same Kafka partition (and hence the same DB processor).
An advantage could be batching (i.e. less DB connections), but I don't think we do that on the consumer side, and if we wanted to do batching we could do this here.
The disadvantage is that I don't think the tenantID is "random" enough , so if you have a single VNode with 1000s of keys, only 1 processor can process them.

I think it may make more sense to use a different field that is more random.

driessamyn (commented):
WIP review - given pair programming was used and all code already seen by 2 people, I only focused on high level review.
Mostly high level comments and questions.
Couple of questions about scope (requirement to support limit and "simulation" mode).

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
dickon: Still an open question on the design PR - https://github.com/corda/platform-eng-design/pull/549#discussion_r1272470205 

I don't know whether it's best to continue that discussion there or in the code, but this morning I replied on the design PR thread.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
dickon: I feel we need some way to control the amount of work done during key rotation, and limit is the easy way to do it from our code, since the hassle is with the operator.  If cancellation gets implemented before this code ships I'm open to taken this out again, though both a limit feature and a way to cancel could be useful to customers.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
dickon: Yes, I'd missed that.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/WrappingRepositoryImpl.kt
```diff
@@ -39,14 +39,14 @@ class WrappingRepositoryImpl(
                         isParentKeyManaged = false,
                         parentKeyReference = key.parentKeyAlias,
                     )
-                )
+                ).also { wrappingKeyEntity -> check(wrappingKeyEntity.generation == key.generation) }
```
driessamyn: Isn't this something that should be done in the DB rather than in code?
There is a standard pattern for optimistic locking which we use elsewhere and which is thread safe (this isn't).

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/WrappingRepositoryImpl.kt
```diff
@@ -39,14 +39,14 @@ class WrappingRepositoryImpl(
...
             }.toDto().also {
-                logger.info("Storing wrapping key with alias $alias in tenant $tenantId")
+                logger.info("Storing wrapping key with alias ${key.alias} in tenant $tenantId")
```
driessamyn: If we are logging, is it worth adding the generation id as well?

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/WrappingRepositoryImpl.kt
```diff
@@ -58,6 +58,22 @@ class WrappingRepositoryImpl(
                 Pair(dao.id, dao.toDto())
             }
         }
+
+    // TODO add test coverage
```
driessamyn: Is there a JIRA for this?
Any reason this wasn't included in the development from the start?

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/WrappingRepositoryImpl.kt
```diff
@@ -58,6 +58,22 @@ class WrappingRepositoryImpl(
...
+    override fun findKeysWrappedByAlias(alias: String): Sequence<WrappingKeyInfo> =
+        entityManagerFactory.createEntityManager().use { it ->
+            it.createQuery(
+                "FROM ${WrappingKeyEntity::class.simpleName} AS k WHERE k.parentKeyReference = :alias",
+                WrappingKeyEntity::class.java
+            ).setParameter("alias", alias).resultStream.map {dao -> dao.toDto() }.toList().asSequence()
```
driessamyn: Why do we go from stream -> list -> stream again?

components/crypto/crypto-softhsm-impl/src/test/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoServiceCachingTests.kt
```diff
@@ -208,6 +208,14 @@ class SoftCryptoServiceCachingTests {
...
             override fun findKeyAndId(alias: String): Pair<UUID, WrappingKeyInfo>? = TODO("Not needed")
 
+            override fun findKeysWrappedByAlias(alias: String): Sequence<WrappingKeyInfo> {
+                TODO("Not yet implemented")
+            }
+
+            override fun getKeyById(id: UUID): WrappingKeyInfo? {
+                TODO("Not yet implemented")
+            }
+
```
driessamyn: Is this intentionally not yet implemented?

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/SigningServiceImpl.kt
```diff
@@ -120,6 +120,22 @@ class SigningServiceImpl @Activate constructor(
...
+    }
+
+    override fun signWithMyCategoryKeys(
+        bytes: ByteArray,
+        category: String,
+        signatureSpec: SignatureSpec
+    ): MutableSet<DigitalSignature.WithKeyId> {
+        TODO("Not yet implemented")
+    }
+
```
driessamyn: Is this related to key rotation?

gradle.properties
```diff
@@ -45,8 +45,8 @@ commonsLangVersion = 3.12.0
 commonsTextVersion = 1.10.0
 bouncycastleVersion=1.76
 # Corda API libs revision (change in 4th digit indicates a breaking change)
-# Change to 5.2.0.xx-SNAPSHOT to pick up maven local published copy
-cordaApiVersion=5.2.0.2-beta+
+# Change to 5.1.0.xx-SNAPSHOT to pick up maven local published copy
+cordaApiVersion=5.1.0.38-alpha-1699268962803
```
driessamyn: FYI we should replace this with a 5.2 reference.

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/ConfigEntityFactory.kt
```diff
@@ -42,8 +43,16 @@ class ConfigEntityFactory(
...
+            listOf(
+                KeyDerivationParameters(
+                    smartConfigFactory.makeSecret(random.randomString(), "corda-master-wrapping-key-passphrase").root(),
+                    smartConfigFactory.makeSecret(random.randomString(), "corda-master-wrapping-key-salt").root()
+                ),
+                KeyDerivationParameters(
+                    smartConfigFactory.makeSecret(random.randomString(), "corda-master-wrapping-key-2-passphrase").root(),
+                    smartConfigFactory.makeSecret(random.randomString(), "corda-master-wrapping-key-2-salt").root()
+                )
+            ),
```
driessamyn, line 55: Why do we generate 2 when bootstrapping?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
dickon: That's correct, and I agree the comment should change.

driessamyn (commented):
Some more minor comments.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
driessamyn: Regardless of the outcome, I think we should remove it from the API until it's implemented.
If I understand it correctly, the main benefit would be the generate a "report" of which keys are impacted, and I feel there are better ways to do this than a "simulation" option, which still would generate a bunch of msg bus traffic, for example.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
dickon: I don't think I understand the REST worker design well enough to say so I'd like to talk to @anien about this when she's in.  I think the idea was to crib what the other resources were doing.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRewrapBusProcessor.kt
dickon: It's a response that acks the rewrap request, not the status of the request. I was thinking we'd make lots of incremental updates to state manager as the work proceeded.

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/ConfigEntityFactory.kt
dickon: So we've got enough to do a key rotation in the e2e tests. Bit of a hack, which might justify a comment since it isn't obvious enough.

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/SigningServiceImpl.kt
dickon: No, as I was saying on Slack, I think it's a git artifact of a revert since we last rebased, and I'm working on a personal branch to fix that. Hopefully github will get the merge right when that goes in.

components/crypto/crypto-softhsm-impl/src/test/kotlin/net/corda/crypto/softhsm/impl/SoftCryptoServiceCachingTests.kt
dickon: yes, this is test fixture and there's no test cases which use hit these endpoints, but the Kotlin compiler requires all overrides to be implemented.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
dickon: Okay, doing that in my branch dickon/CORE-12165-key-rotation-no-limit

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: This was implemented similarly to the start flow request. The main reason for having it like this is that we don't know how long the key rotation will take. So the idea was that we tell the user we have accepted the request and the key rotation status is the way to tell user how we are doing/what's the progress of the key rotation.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
dickon: I'm taking out limit in a PR to the feature branch https://github.com/corda/corda-runtime-os/pull/5129/files

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
dickon: I'm removing these arguments in https://github.com/corda/corda-runtime-os/pull/5129/files

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
dickon: Okay I think this is fine and what I expected for now. The alternative is to make the HTTP request synchronous, so it get a response until the work is done, but given the crypto worker may not be available and we are dispatching to Kafka I don't think that's wanted.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/WrappingRepository.kt
dickon: done in my PR

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/WrappingRepository.kt
dickon: done in my pR

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
dickon: done in my PR

components/crypto/crypto-rest/src/test/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceTest.kt
dickon: will use @Disabled in my PR.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
dickon: No. The code is a bit confusing and I've improved it on my second PR https://github.com/corda/corda-runtime-os/pull/5134

On that branch, things are how I think they should be. The operator calls the REST worker, that creates one `KeyRotationRequest` which may be for the whole cluster or a single vnode. That's then picked up by a crypto worker, and that published a number of `IndvidiaulKeyRotationRequest` messages, potentially a large number. If lots of wrappings are found in the same tenant there will be many messages. As far as I know the work is then picked up by all the separate workers.

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/WrappingRepositoryImpl.kt
dickon: Yeah, I hope I've now dealt with the race in SoftCryptoService.createWrappingKey, and the other path that reaches here, in https://github.com/corda/corda-runtime-os/pull/5134

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/WrappingRepositoryImpl.kt
dickon: done in https://github.com/corda/corda-runtime-os/pull/5134

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
dickon: done on my https://github.com/corda/corda-runtime-os/pull/5134 (and probably the first one too)_

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
dickon: done on https://github.com/corda/corda-runtime-os/pull/5134

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
dickon: Comment fixed on https://github.com/corda/corda-runtime-os/pull/5134

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
dickon: It turns out this was wrong, since the non-vnode tenants share a cluster database, so we only need to look at the table once, otherwise we end up rewrapping keys 3 times. So this code is now obsolete and the new version in https://github.com/corda/corda-runtime-os/pull/5134 has a comment on this.,

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
dickon: done in https://github.com/corda/corda-runtime-os/pull/5134

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
dickon: removed, even though I think it was useful, it's an architectural call.

dickon: removed in https://github.com/corda/corda-runtime-os/pull/5129

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/WrappingRepositoryImpl.kt
dickon: test coverage added in https://github.com/corda/corda-runtime-os/pull/5134

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/WrappingRepositoryImpl.kt
dickon: no good reason, in https://github.com/corda/corda-runtime-os/pull/5134 it is stream -> sequence, since we usually pass around Kotlin types

gradle.properties
dickon: done in https://github.com/corda/corda-api/pull/1349

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/ConfigEntityFactory.kt
dickon: reworked in https://github.com/corda/corda-runtime-os/pull/5126/files so this comment is obsolete.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRewrapBusProcessor.kt
anien: Yes, this was originally planned it might be used for the status, but without a clear idea. It's not needed at the moment and once status is implemented, state manager will be used.

gradle.properties
dickon: done

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/ConfigEntityFactory.kt
dickon: done

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/crypto/SigningServiceImpl.kt
dickon: Removed by fixing what I think was a mistake in the merge Dries did from 5.2

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/WrappingRepository.kt
dickon: in feature branch

components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/WrappingRepository.kt
dickon: in feature branch

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
dickon: done in feature branch

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
dickon: done in feature branch

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/ConfigEntityFactory.kt
anien: We still have two keys here. Do we want to keep that?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
dickon: 5134 is merged, callng this resolved.

dickon:
Matching PR for API change is https://github.com/corda/corda-api/pull/1348

mag1c-48 (dismissed):
LGTM - Build files only

charlieR3 (dismissed):
LGTM - network team files

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Key Rotation API consists of endpoints to request a key rotation operation and to check the status of such operation.
+ */
+@HttpRestResource(
+    name = "Key Rotation API",
+    description = "Contains operations related to rotation of the wrapping keys.",
+    path = "wrappingkey",
+    minVersion = RestApiVersion.C5_2
+)
+interface KeyRotationRestResource : RestResource {
```
vkolomeyko, line 23: Could you please add this new resource to `OpenApiCompatibilityTest` ?

It will also require Open API Baseline to be updated.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/AssertWithRetryBuilder.kt
```diff
@@ -10,9 +10,13 @@ class AssertWithRetryArgs {
     var interval: Duration = Duration.ofMillis(500)
     var startDelay: Duration = Duration.ofMillis(10)
     var command: (() -> SimpleResponse)? = null
-    var condition: ((SimpleResponse) -> Boolean)? = null
```
vkolomeyko, line 13: This seems like unrelated change. Would it not be better to do it as a separate PR?

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Initialises the API implementation. This method may be called multiple times throughout the life
+     * of the API.
+     * @param config A config
+     */
+    fun initialise(config: Map<String, SmartConfig>)
```
vkolomeyko: Do we need this non-REST annotated method on the interface?

IMO better way of onboarding config is by subscribing to configuration change, please see: use of `configurationReadService.registerComponentForUpdates` in `VirtualNodeRestResourceImpl`.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/AssertWithRetryBuilder.kt
dickon: This merges a feature branch that has been open for months, and while I would normally like to keep PRs focussed on a single change I feel in this circumstance it's not worth the time to unpick changes.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
```diff
@@ -0,0 +1,76 @@
+package net.corda.crypto.rest
```
vkolomeyko, line 1: I think this new package requires OSGi export provisions or else this pluggable REST resource is not going to be discovered by REST Worker.

Also, it would probably make sense to do some Dev testing of the new REST endpoint using Combined Worker locally.

anien: The new REST endpoints have already been tested locally in K8s cluster. The endpoints were available in swagger as well as called via curl command.
What do you mean that it might requires OSGi export please?

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
anien: The approach was taken from the `FlowRestResource`. I will take a look if we can simply use private function.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
driessamyn: But doesn't the future just wait for the publish to complete?
I could be wrong, but I think the way it currently is means that if the publish would fail, you would still get an ok response.

anien: If the publisher is not initialised, it throws the exception. There is also a test case for it. I am not that amazingly familiar with the rest resources in Corda, so happy to discuss this if you think it should be done differently.

libs/application/application-db-setup/src/main/kotlin/net/corda/application/dbsetup/ConfigEntityFactory.kt
driessamyn: I think it might be quite confusing.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRewrapBusProcessor.kt
```diff
@@ -0,0 +1,38 @@
...
+                REWRAP_MESSAGE_RESPONSE_TOPIC,
+                request.requestId, IndividualKeyRotationResponse(
+                    request.requestId,
+                    request.tenantId,
+                    request.oldParentKeyAlias,
+                    request.newParentKeyAlias,
+                    request.targetKeyAlias,
+                    Instant.now()
+                )
+            )
```
driessamyn: Why do we need this response AVRO message and topic?
I'm guessing this will be replaced with the SM implementation, so wouldn't it make sense to take these away from the API change?

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
anien: Fixed in [#5143](https://github.com/corda/corda-runtime-os/pull/5143)

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
anien: Fixed in #5143

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRewrapBusProcessor.kt
anien: We agreed over slack discussion that after status using state manager is implemented, we will review all topics and avro schemas introduces in key rotation. Jira ticket is created for this.

dickon: Removed for now.

vkolomeyko (approved):

emilybowe (approved):
LGTM (flow changes only)

driessamyn (approved):
LGTM

lankydan (approved):
Approved on the basis that no ledger code seems to have changed.

mag1c-48 (approved):
LGTM - build

wzur-r3 (approved):
LGTM for settings.gradle owned by BLT

charlieR3 (approved):
LGTM - network team files

PR #5127, opened 2023-11-17
CORE-15267 Make MemberInfo serializable

YashNabar:
Makes `MemberInfo` serializable by registering custom serializers for `MemberContext` and `MGMContext`. The corresponding corda-api change marks these interfaces as `@CordaSerializable` - https://github.com/corda/corda-api/pull/1341

This change was tested manually using a modified version of the combined worker smoke tests, by including `MemberInfo` in one of the flow states used by the tests. [[build](https://ci02.dev.r3.com/job/Corda5/job/corda-runtime-os-combined-worker-e2e-tests/job/PR-yash%2Fserialization-fix%2Ftest/)]
It was also manually confirmed that this change does not break ABI compatibility, by running a modified version of the end-to-end flow tests against a combined worker compiled with the newer corda-api version.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -46,7 +46,7 @@ commonsTextVersion = 1.10.0
 bouncycastleVersion=1.76
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.1.0.xx-SNAPSHOT to pick up maven local published copy
-cordaApiVersion=5.1.0.39-alpha-1700162971190
+cordaApiVersion=5.1.0.39-beta+
 
 disruptorVersion=3.4.4
 felixConfigAdminVersion=1.9.26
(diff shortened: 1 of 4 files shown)


JamesHR3 (commented):
Is it possible to add some unit tests for these serializers? You might need to declare a test dependency on the AMQP serialization library to do so. There may be some issues around sandboxing that make this difficult/impossible, but worth investigating.

YashNabar:
> Is it possible to add some unit tests for these serializers? You might need to declare a test dependency on the AMQP serialization library to do so. There may be some issues around sandboxing that make this difficult/impossible, but worth investigating.

Discussed offline, end-to-end tests and any possible unit tests to be added in a follow-up PR (cc @dimosr @JamesHR3)

charlieR3 (dismissed):

dimosr (approved):

PR #5129, opened 2023-11-18
CORE-12165: key rotation no limit

dickon:
* bump API reference to 5.2
* reduce trace level
* change REST path to 5.2
* drop simulate, timeToLive and limit parameters (due to YAGNI)

### DIFF
--- a/components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/WrappingRepository.kt
+++ b/components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/WrappingRepository.kt
@@ -15,7 +15,6 @@ interface WrappingRepository : Closeable {
     /**
      * Save a wrapping key to the database
      *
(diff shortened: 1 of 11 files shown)


components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/WrappingRepository.kt
```diff
@@ -41,7 +39,7 @@ interface WrappingRepository : Closeable {
     fun findKey(alias: String): WrappingKeyInfo?
 
     /**
-     * Find a wrapping key in the database
+     * Find a wrapping key in the database`
```
anien: Revert the change

dickon: done

anien (dismissed):
LGTM.

anien (approved):
LGTM.

PR #5131, opened 2023-11-20
CORE-12312 - Updated WWW-Authenticate response header

filipesoliveira:
REST framework now returns a descriptive realm value instead of an internal class name.
 Added tests.

Proof:
image


libs/rest/rest-server-impl/src/main/kotlin/net/corda/rest/server/impl/security/provider/basic/UsernamePasswordAuthenticationProvider.kt
```diff
@@ -31,6 +31,6 @@ internal class UsernamePasswordAuthenticationProvider(private val restSecurityMa
     }
 
     override fun provideParameters(): Map<String, String> {
-        return mapOf(REALM_KEY to restSecurityManagerSupplier.get().id.value)
+        return mapOf(REALM_KEY to "Corda REST Worker")
```
vkolomeyko: Nit: We can introduce a constant with value `"Corda REST Worker"` and then can use it in the tests.

vkolomeyko (dismissed):
LGTM

vkolomeyko (approved):
LGTM

josephzunigadaly (approved):

PR #5133, opened 2023-11-20
CORE-18414 Tests for MemberInfo serialization

YashNabar:
Adds unit testing around custom AMQP serializers added for `MemberContext` and `MGMContext` as part of https://github.com/corda/corda-runtime-os/pull/5127.
Also includes some refactoring around `MemberInfo` test utilities.


libs/membership/membership-impl/src/test/kotlin/net/corda/membership/lib/impl/utils/MemberInfoTestUtils.kt
```diff
@@ -0,0 +1,103 @@
...
+)
+internal val ledgerKeys = listOf(key, key)
+
+private const val NULL_KEY = "nullKey"
+private const val DUMMY_KEY = "dummyKey"
+
+@Suppress("SpreadOperator")
+fun createDummyMemberInfo(
+    converters: List<CustomPropertyConverter<out Any>>,
+    additionalMemberInfoProperties: List<Pair<String, String>> = emptyList(),
```
charlieR3: This might be better as `additionalMemberContext` so we could also have `additionalMgmContext` if necessary. I think it's a bit clearer also since I can tell at a glance where those additional properties will be added in the member info.

libs/membership/membership-impl/src/test/kotlin/net/corda/membership/lib/impl/utils/MemberInfoTestUtils.kt
```diff
@@ -0,0 +1,103 @@
...
+) = LayeredPropertyMapMocks.create<MemberContextImpl>(
+    sortedMapOf(
+        MemberInfoExtension.PARTY_NAME to "O=Alice,L=London,C=GB",
+        String.format(MemberInfoExtension.PARTY_SESSION_KEYS, 0) to KEY,
+        MemberInfoExtension.GROUP_ID to "DEFAULT_MEMBER_GROUP_ID",
+        *convertPublicKeys().toTypedArray(),
+        *convertEndpoints().toTypedArray(),
+        *additionalMemberInfoProperties.toTypedArray(),
```
charlieR3: It could be good to add these properties last in case the caller intends to override properties in the default map. e.g. software version

libs/membership/membership-impl/src/test/kotlin/net/corda/membership/lib/impl/utils/MemberInfoTestUtils.kt
```diff
@@ -0,0 +1,103 @@
...
+
+@Suppress("SpreadOperator")
+fun createDummyMemberContext(
+    converters: List<CustomPropertyConverter<out Any>>,
+    additionalMemberInfoProperties: List<Pair<String, String>> = emptyList(),
+) = LayeredPropertyMapMocks.create<MemberContextImpl>(
+    sortedMapOf(
+        MemberInfoExtension.PARTY_NAME to "O=Alice,L=London,C=GB",
+        String.format(MemberInfoExtension.PARTY_SESSION_KEYS, 0) to KEY,
+        MemberInfoExtension.GROUP_ID to "DEFAULT_MEMBER_GROUP_ID",
```
charlieR3: Minor comment but you could use a valid UUID here as a sample group ID in case a test case using this tries to use the group ID.

libs/membership/membership-impl/src/test/kotlin/net/corda/membership/lib/impl/utils/MemberInfoTestUtils.kt
```diff
@@ -46,28 +47,28 @@ private const val DUMMY_KEY = "dummyKey"
...
 fun createDummyMemberContext(
     converters: List<CustomPropertyConverter<out Any>>,
-    additionalMemberInfoProperties: List<Pair<String, String>> = emptyList(),
+    additionalMemberContext: List<Pair<String, String>> = emptyList(),
 ) = LayeredPropertyMapMocks.create<MemberContextImpl>(
     sortedMapOf(
         MemberInfoExtension.PARTY_NAME to "O=Alice,L=London,C=GB",
         String.format(MemberInfoExtension.PARTY_SESSION_KEYS, 0) to KEY,
-        MemberInfoExtension.GROUP_ID to "DEFAULT_MEMBER_GROUP_ID",
+        MemberInfoExtension.GROUP_ID to UUID.randomUUID().toString(),
```
charlieR3: Better to have something deterministic here e.g. `UUID(0,1).toString()`

charlieR3 (dismissed):

charlieR3 (approved):

PR #5135, opened 2023-11-20
CORE-17821 Removed Kafka implementation from CryptoProcessorImpl

ben-millar:
During the transition of worker communication from asynchronous Kafka to synchronous RPC, some code duplication was introduced in order to allow green/blue testing of both communication protocols. Now that synchronous RPC communication is fully implemented, any duplicated or obsolete code should be removed.


components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsProcessor.kt
```diff
@@ -78,12 +80,11 @@ class CryptoFlowOpsRpcProcessor(
                     request.flowExternalEventContext,
                     FlowOpsResponse(createResponseContext(request), response, null)
                 )
-            } catch (throwable: Throwable) {
+            } catch (ex: Exception) {
```
LWogan: Why are we loosening this from Throwable to Exception?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsProcessor.kt
```diff
@@ -78,12 +80,11 @@ class CryptoFlowOpsRpcProcessor(
...
                 logger.error(
```
LWogan: this should be warn, not error

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -371,35 +361,14 @@ class CryptoProcessorImpl @Activate constructor(
                 responderProcessor = hsmRegistrationProcessor,
                 messagingConfig = messagingConfig
             )
+        }.also {
```
LWogan, line 386: nice use of also!

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsProcessor.kt
```diff
@@ -54,7 +55,7 @@ class CryptoFlowOpsRpcProcessor(
     )
 
     override fun process(request: FlowOpsRequest): FlowEvent {
-        logger.trace { "process just started processing ${request::class.java.name}" }
+        logger.trace { "CryptoFlowOpsProcessor began processing request: ${request::class.java.name}" }
```
kyriathar: Think we can omit the name of the class. That will be included by logging library.

Message could be "Processing request ..."

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsProcessor.kt
kyriathar:
```suggestion
            } catch (e: Exception) {
```

(nit)

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -121,11 +118,10 @@ class CryptoProcessorImpl @Activate constructor(
             MESSAGING_CONFIG,
             CRYPTO_CONFIG
         )
-        const val FLOW_OPS_SUBSCRIPTION = "FLOW_OPS_SUBSCRIPTION"
+        const val RPC_SUBSCRIPTION = "RPC_SUBSCRIPTION"
```
kyriathar: I think i would still use "FLOW" somewhere in there to indicate it's the flow-crypto subscription.

LWogan (commented):
A few initial comments. Looks ok to me but I wasn't here when this work was being done so it would be good to get a review from someone more familiar with the changes and crypto

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -309,25 +304,20 @@ class CryptoProcessorImpl @Activate constructor(
     }
 
     private fun startProcessors(event: ConfigChangedEvent, coordinator: LifecycleCoordinator) {
-        val cryptoConfig = event.config.getConfig(CRYPTO_CONFIG)
+        val cryptoConfig = event.config.getConfig(CRYPTO_CONFIG).retrying()
```
kyriathar: So this is not crypto config any more in reality but a subsection of it which is about retrying things. Could we please revert?

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -309,25 +304,20 @@ class CryptoProcessorImpl @Activate constructor(
     }
 
     private fun startProcessors(event: ConfigChangedEvent, coordinator: LifecycleCoordinator) {
```
kyriathar, line 307: So this function now contains both http and Kafka subscriptions. Could we add a comment somewhere to make that a bit clearer? (which is what)

kyriathar (commented):
A few comments

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/CryptoFlowOpsProcessor.kt
ben-millar: It's very rare that we should catch all `Throwable`, which comprises all of `Exception` _and_ all of `Error`. This was probably a mistake. See: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Error.html
>An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
ben-millar: The `cryptoConfig` in the original code was never used anywhere other than to create the `retryingConfig`; I've just inlined that to avoid storing intermediary vals. I can rename this, if you prefer?

kyriathar: Yeah that's fine. So long as the name aligns with what it stores.

kyriathar (approved):
LGTM

PR #5141, opened 2023-11-21
CORE-12165: Minor fixes, typos, removing comments

anien:


dickon (approved):
LGTM. Since we need to do an API dance we have to wait for at least one more feature branch CI build.

PR #5142, opened 2023-11-21
CORE-18365 Extract config in support bundle

davidcurrie:
Also noted that status was not being reliably retrieved on macOS via the bash script. Switched to using `kubectl proxy`.


ronanbrowne (approved):
LGTM  (although I can't comment too much on ps1 side as not my forte)

davidcurrie:
> LGTM (although I can't comment too much on ps1 side as not my forte)

Not mine either. It runs on my machine and, tbh, I don't know whether anyone else is actually using it.

PR #5143, opened 2023-11-21
CORE-12165: Update swagger baseline test, remove function from rest resource

anien:


vkolomeyko (approved):
LGTM

components/crypto/crypto-rest/src/main/java/net/corda/crypto/rest/response/package-info.java
```diff
@@ -0,0 +1,4 @@
+@Export
+package net.corda.crypto.rest.response;
```
dickon, line 2: I'd normally expect to export a single package from a gradle module, so I'm surprised to see 2.

dickon (approved):
LGTM (the comment about exporting 2 packages seems minor).

components/crypto/crypto-rest/src/main/java/net/corda/crypto/rest/response/package-info.java
anien: This was needed for the new data class that has to be visible. I've seen two export packages per module elsewhere, hence it shouldn't be a problem.

vkolomeyko: I would say it is perfectly normal to have more than one package exported from a module. We do have a number of other modules which do the same thing.

PR #5148, opened 2023-11-21
CORE-12165: Return a list from findKeysWrappedByAlias to avoid using the Stream after it is closed

dickon:


components/crypto/crypto-softhsm-impl/src/main/kotlin/net/corda/crypto/softhsm/impl/WrappingRepositoryImpl.kt
```diff
@@ -60,12 +60,12 @@ class WrappingRepositoryImpl(
...
         }
 
-    override fun findKeysWrappedByAlias(alias: String): Sequence<WrappingKeyInfo> =
+    override fun findKeysWrappedByAlias(alias: String): List<WrappingKeyInfo> =
         entityManagerFactory.createEntityManager().use {
             it.createQuery(
                 "FROM ${WrappingKeyEntity::class.simpleName} AS k WHERE k.parentKeyReference = :alias",
                 WrappingKeyEntity::class.java
-            ).setParameter("alias", alias).resultStream.asSequence().map {dao -> dao.toDto() }
+            ).setParameter("alias", alias).resultList.map { dao -> dao.toDto() }
```
anien, line 67: Shouldn't there be also a change in the tests? If I remember correctly we mock `createQuery`. Let me check.

anien: Few more changes were required.

dickon:
This is content is in #5151

PR #5151, opened 2023-11-21
CORE-12165: merge 5.2 changes and use the expected new version of the feature branch on the API

dickon:


anien (approved):
This is a merge from 5.2, reverting UUID changes: 3490b4048d6c950938575b73b1f07bb774cdf441 and changing stream as list in WrappingRepository. If we want to bring UUID changes back, we will do it separately as a direct PR into 5.2. This is to speed up merging this feature branch.

PR #5159, opened 2023-11-22
CORE-18463: Reduce logging in assertWithRetry

josephzunigadaly:
- Remove the retry log on successful runs because it will greatly reduce log output
- Avoid saying "Failed" in the context of retry attempts because "Failed" is a useful keyword when looking for assertion failures
- But do output the URL once, and a full stop for each retry because it is useful to see some activity when watching a test run
- Keep the response for each retry in a list and output the list on failure so we still have the logs on failure
- Output the log of attempts on assertion failure or any unexpected exception

### DIFF
--- a/testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/AssertWithRetryBuilder.kt
+++ b/testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/AssertWithRetryBuilder.kt
@@ -49,14 +49,14 @@ class AssertWithRetryBuilder(private val args: AssertWithRetryArgs) {
 private data class Attempt(val attemptNumber: Int, val timeTried: Duration, val response: String)
 
 private fun Iterable<Attempt>.prettyPrint(): String =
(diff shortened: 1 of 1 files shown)


testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/AssertWithRetryBuilder.kt
```diff
@@ -46,6 +46,21 @@ class AssertWithRetryBuilder(private val args: AssertWithRetryArgs) {
     }
 }
 
+private data class Attempt(val attemptNumber: Int, val timeTried: Duration, val response: String)
+
+private fun formatAttempts(attempts: Iterable<Attempt>): String =
```
vkolomeyko: Minor: I would change to an extension function, like so:
```suggestion
private fun Iterable<Attempt>.prettyPrint(): String =
```

vkolomeyko (dismissed):
Minor comment posted.

vkolomeyko (approved):
LGTM

PR #5160, opened 2023-11-22
CORE-18469: Fix DB Worker becoming unhealthy after running test `ExternalVirtualNodeDB`

BenYip123:
Removes the not null assertion for uniquenessDMLConnectionId as it was causing a null pointer exception


filipesoliveira:
I have tested the change on my machine and the `combined worker` remains healthy after running the test `ExternalVirtualNodeDB`. Thanks for the prompt fix @BenYip123.

filipesoliveira (approved):
LGTM

PR #5162, opened 2023-11-23
CORE-17820 VerificationWorker - Remove Kafka Subscription

mbrkic-r3:
Removed unused verification worker classes related to Kafka subscription (that was replaced with RPC subscription).


components/ledger/ledger-verification/src/integrationTest/kotlin/net/corda/ledger/verification/tests/VerificationRequestProcessorTest.kt
```diff
@@ -156,20 +154,18 @@ class VerificationRequestProcessorTest {
         val request = createRequest(sandbox, holdingIdentity, transaction, cpkSummaries)
 
         // Create request processor
-        val processor = VerificationRequestProcessor(
+        val processor = VerificationRpcRequestProcessor(
```
ben-millar: Nit: I would rename this to `VerificationRequestProcessor`; I think we don't need to qualify it with `Rpc` now that it's the only implementation. (Also because I've already followed this pattern in my changes 😅)

blsemo (dismissed):
Looks fine from a ledger perspective.

components/ledger/ledger-verification/src/main/kotlin/net/corda/ledger/verification/processor/impl/VerificationSubscriptionFactoryImpl.kt
```diff
@@ -30,28 +26,9 @@ class VerificationSubscriptionFactoryImpl @Activate constructor(
...
-
-        return subscriptionFactory.createDurableSubscription(
-            subscriptionConfig,
-            processor,
-            config,
-            null
-        )
-    }
-
     override fun createRpcSubscription(): RPCSubscription<TransactionVerificationRequest, FlowEvent> {
```
ben-millar: I would also rename this to remove the `Rpc` qualifier as I think it's unnecessary, but will leave up to you

ben-millar (dismissed):
Approved with comments

components/ledger/ledger-verification/src/integrationTest/kotlin/net/corda/ledger/verification/tests/VerificationRequestProcessorTest.kt
mbrkic-r3: renamed

components/ledger/ledger-verification/src/main/kotlin/net/corda/ledger/verification/processor/impl/VerificationSubscriptionFactoryImpl.kt
mbrkic-r3: renamed

blsemo (approved):
👍

PR #5166, opened 2023-11-23
CORE-17506 Removing unused flow topics, making Record.topic field optional

ben-millar:
This PR removes the flow-related topics which are no longer in use, and makes the `topic` field in the `Record` object nullable. This nullability is handled for all instances where the `Record.topic` field is read by either filtering out null topics, or by asserting that they're non-null depending on the context.

### DIFF
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/RecordExtensions.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/RecordExtensions.kt
@@ -38,7 +38,9 @@ fun <K: Any, V: Any> EventLogRecord<K, V>.toRecord(): Record<K, V> {
 }
 
 fun Record<*, *>.toCordaProducerRecord(): CordaProducerRecord<*, *> {
-    return CordaProducerRecord(this.topic!!, this.key, this.value, this.headers)
+    val topic = this.topic
+    requireNotNull(topic) { "Topic is not allowed to be null for CordaProducerRecords" }
+    return CordaProducerRecord(topic, this.key, this.value, this.headers)
 }
 
 fun List<Record<*, *>>.toCordaProducerRecords(): List<CordaProducerRecord<*, *>> {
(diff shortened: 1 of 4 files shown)


thiagoviana (dismissed):

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/RecordExtensions.kt
```diff
@@ -38,7 +38,7 @@ fun <K: Any, V: Any> EventLogRecord<K, V>.toRecord(): Record<K, V> {
 }
 
 fun Record<*, *>.toCordaProducerRecord(): CordaProducerRecord<*, *> {
-    return CordaProducerRecord(this.topic, this.key, this.value, this.headers)
+    return CordaProducerRecord(this.topic!!, this.key, this.value, this.headers)
```
LWogan: !! is never a good sign. We want to avoid it at all costs despite javas recent advancements in null pointer exception messages.

Something like this might be better:

```
val topic = this.topic
    requireNotNull(topic) { "Topic is not allowed to be null for CordaProducerRecords" }
    return CordaProducerRecord(topic, this.key, this.value, this.headers)
```

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/records/Record.kt
```diff
@@ -9,7 +9,7 @@ package net.corda.messaging.api.records
  * @property headers Optional list of headers to added to the message.
  */
 data class Record<K : Any, V : Any>(
-    val topic: String,
+    val topic: String?,
```
LWogan, line 13: might be worth adjusting the kdoc slightly for topic. It only needs to be set for records destined for the message bus

testing/p2p/inmemory-messaging-impl/src/main/kotlin/net/corda/messaging/emulation/topic/model/Topics.kt
```diff
@@ -11,14 +11,14 @@ class Topics(
 
     internal fun getWriteLock(records: Collection<Record<*, *>>): PartitionsWriteLock {
         val partitions = records.map {
-            val topic = getTopic(it.topic)
+            val topic = getTopic(it.topic!!)
```
LWogan: same re requires in this class

LWogan (commented):
Looks good, few small comments

LWogan (approved):

PR #5168, opened 2023-11-23
CORE-18499 - Validate cluster readiness before E2E tests

driessamyn:
- Extract cluster readiness test
- Add check in BeforeAll of tests

### DIFF
--- a/applications/workers/workers-smoketest/build.gradle
+++ b/applications/workers/workers-smoketest/build.gradle
@@ -54,7 +54,6 @@ kotlin {
 dependencies {
     // NO CORDA DEPENDENCIES!!
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
-    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:$kotlinCoroutinesVersion"
 
     smokeTestImplementation "net.corda:corda-avro-schema:$cordaApiVersion"
     // Avoid having the schema names and keys scattered across projects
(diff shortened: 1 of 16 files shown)


mbrkic-r3:
Couldn't we still have situation (similar to https://r3holdco.slack.com/archives/C05GFSAS76E/p1700746900629469?thread_ts=1699869617.100379&cid=C05GFSAS76E) where `close` function of `SingleClusterTestConfigManager` (used in `ConfigurationChangeTest`) is called after another test has started (and passed `assertIsReady`)?

driessamyn:
> Couldn't we still have situation (similar to https://r3holdco.slack.com/archives/C05GFSAS76E/p1700746900629469?thread_ts=1699869617.100379&cid=C05GFSAS76E) where `close` function of `SingleClusterTestConfigManager` (used in `ConfigurationChangeTest`) is called after another test has started (and passed `assertIsReady`)?

Yes and no. 
I think any test that is "disruptive" like that needs to be run in isolation. `ConfigurationChangeTest` is, for example, configure like this, so I believe it sufficient to test before the next tests start that the cluster is back "up".
Ideally, I'd like to see all those tests isolated (i.e. not mixed in a class with other tests that can be run in parallel) so we minimize the amount of tests like that.

Note that, for example, in `ConfigTests`, `assertIsReady` is executed before _each_ test.

mbrkic-r3:
Looks to me that `managedConfig` could cause issues anyway, and that it should be modified so it closes `TestConfigManager` (so "revert" config change is triggered before the next tests starts)?

driessamyn:
> Looks to me that `managedConfig` could cause issues anyway, and that it should be modified so it closes `TestConfigManager` (so "revert" config change is triggered before the next tests starts)?

Probably.

In practice, I don't think it matters as this test is configured to run as last, but it's not great to rely on that.
I'm not convinced the abstraction is very helpful as it hides some important things.

ronanbrowne (approved):
LGTM (build files only reviewed)

PR #5170, opened 2023-11-24
CORE-18511 map flows to test

blsemo:
As it can be very tricky to map flows in logfiles to failed tests, this PR uses the test name as request id where possible.


LWogan (dismissed):
LGTM

LWogan (approved):
This a good improvement on what we have now.

PR #5173, opened 2023-11-27
CORE-15180 - The class `net.corda.ledger.utxo.token.cache.entities.TokenBalance` has been refactored

filipesoliveira:
The class `net.corda.ledger.utxo.token.cache.entities.TokenBalance` has been refactored.

### DIFF
--- a/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/entities/internal/TokenBalanceImpl.kt
+++ b/components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/entities/internal/TokenBalanceImpl.kt
@@ -1,0 +1,14 @@
+package net.corda.ledger.utxo.token.cache.entities.internal
+
+import net.corda.v5.ledger.utxo.token.selection.TokenBalance
+import java.math.BigDecimal
+
+internal class TokenBalanceImpl(private val availableBalance: BigDecimal, private val totalBalance: BigDecimal) : TokenBalance {
+    override fun getAvailableBalance(): BigDecimal {
+        return  availableBalance
+    }
+
+    override fun getTotalBalance(): BigDecimal {
+        return totalBalance
+    }
+}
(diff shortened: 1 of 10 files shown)


components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/entities/internal/TokenBalanceImpl.kt
```diff
@@ -0,0 +1,14 @@
+package net.corda.ledger.utxo.token.cache.entities.internal
+
+import net.corda.v5.ledger.utxo.token.selection.TokenBalance
+import java.math.BigDecimal
+
+internal class TokenBalanceImpl(private val availableBalance: BigDecimal, private val totalBalance: BigDecimal) : TokenBalance {
```
vkolomeyko: I would suggest re-naming:
```suggestion
internal class TokenBalanceCacheImpl(private val availableBalance: BigDecimal, private val totalBalance: BigDecimal) : TokenBalance {
```
just to avoid name clash with existing `TokenBalanceImpl` from a different module.

components/ledger/ledger-utxo-token-cache/src/main/kotlin/net/corda/ledger/utxo/token/cache/factories/RecordFactory.kt
```diff
@@ -3,7 +3,7 @@ package net.corda.ledger.utxo.token.cache.factories
 import net.corda.data.flow.event.FlowEvent
 import net.corda.data.ledger.utxo.token.selection.data.Token
 import net.corda.ledger.utxo.token.cache.entities.CachedToken
-import net.corda.ledger.utxo.token.cache.entities.TokenBalance
+import net.corda.v5.ledger.utxo.token.selection.TokenBalance
```
vkolomeyko, line 6: I would also add:
```
api 'net.corda:corda-ledger-utxo'
```
to the Gradle file of this module to enforce transitive dependency on said module for all those places where this module is used.

vkolomeyko (commented):
Looks fine to me.

Couple of comments added.

vkolomeyko (approved):
LGTM

PR #5174, opened 2023-11-27
CORE-16150 - Add message rate metrics for session messages

dimosr:
## Changes
Adding inbound/outbound message rate metrics for session messages.

## Testing
Performed a small multi-cluster test run: https://ci02.dev.r3.com/job/Corda5/job/Large%20Network%20Tests/job/PR-69/196/
The dashboards seem to be picking up the data without any further changes needed.

session_dashboards_2

### DIFF
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
@@ -32,6 +32,7 @@ import net.corda.p2p.linkmanager.sessions.SessionManager
 import net.corda.data.p2p.markers.AppMessageMarker
 import net.corda.data.p2p.markers.LinkManagerReceivedMarker
 import net.corda.p2p.linkmanager.metrics.recordInboundMessagesMetric
+import net.corda.p2p.linkmanager.metrics.recordInboundSessionMessagesMetric
 import net.corda.schema.Schemas
 import net.corda.tracing.traceEventProcessing
 import net.corda.utilities.debug
@@ -104,6 +105,7 @@ internal class InboundMessageProcessor(
     private fun processSessionMessage(message: LinkInMessage): List<Record<String, *>> {
         val response = sessionManager.processSessionMessage(message)
         return if (response != null) {
+            recordInboundSessionMessagesMetric(response.header.sourceIdentity, response.header.destinationIdentity)
             when (val payload = message.payload) {
                 is InitiatorHelloMessage -> {
                     val partitionsAssigned =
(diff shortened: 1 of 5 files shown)


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
```diff
@@ -76,6 +77,9 @@ internal class OutboundMessageProcessor(
                 )
                 emptyList()
             } else {
+                state.messages.forEach {
+                    recordOutboundSessionMessagesMetric(state.sessionCounterparties.ourId, state.sessionCounterparties.counterpartyId)
+                }
```
charlieR3, line 82: Would this result in only the initiatorhello message being logged? I might be wrong, but I thought only the initiatorhello goes through this path and the initiatorhandshake does not.

dimosr: good spot, I think you're right. 

In the `InboundMessageProcessor.processSessionMessage()`, we need to record the equivalent outbound metrics along with the inbound metrics. This should capture the remaining session messages of the handshake.

dimosr: Added this change now, I can see more data on the outbound. So, I did a re-run and updated the screenshots/links in the description.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -133,6 +135,11 @@ internal class InboundMessageProcessor(
                     listOf(Record(Schemas.P2P.LINK_OUT_TOPIC, LinkManager.generateKey(), response))
                 }
             }
+            if (records.isNotEmpty()) {
+                recordInboundSessionMessagesMetric(response.header.destinationIdentity, response.header.sourceIdentity)
+                recordOutboundSessionMessagesMetric(response.header.sourceIdentity, response.header.destinationIdentity)
```
charlieR3: This looks better, but I'm not sure it will accurately capture inbound/outbound session metrics. I think if an error occurs during processing an inbound message, or if the inbound message was `ResponderHandshake` then after `val response = sessionManager.processSessionMessage(message)` `response` will be null and the inbound session metric will never be captured despite being received. Would it be better to record `recordInboundSessionMessagesMetric` at the start of `processSessionMessage` and leave `recordOutboundSessionMessagesMetric` as it currently is?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
charlieR3: Added updated feedback related to this on the change in the `InboundMessageProcessor`

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
dimosr: yes, that's a good point. Let me move the inbound metrics up there.

charlieR3 (approved):

PR #5175, opened 2023-11-27
CORE-18029: Batch delete states in the state manager

JamesHR3:
### Problem description

When the state manager is asked to delete states, it does this one at a time. This means many more database interactions than is required.

### Solution

Delete the full batch of requested states in a single statement executed as a batch.

If an individual state is deleted, the index of that state in the returned array will have a value of 1. A value of 0 or less indicates that either the state was never there, the optimistic lock check failed, or a driver error occurred. The repository treats all these the same and reports the keys as failed. The state manager then does a lookup of any failed keys to see if they exist. The client is only informed of those keys that are still present within the state manager.

### DIFF
--- a/libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
+++ b/libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
@@ -8,7 +8,7 @@ import net.corda.libs.statemanager.impl.repository.StateRepository
 import java.sql.Connection
 import java.sql.Timestamp
 
-// TODO-[CORE-18029 / CORE-18030]: batch delete and create.
+// TODO-[CORE-18030]: batch create.
 class StateRepositoryImpl(private val queryProvider: QueryProvider) : StateRepository {
 
     override fun create(connection: Connection, state: StateEntity) {
--- a/libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
+++ b/libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
@@ -55,21 +55,23 @@ class StateRepositoryImpl(private val queryProvider: QueryProvider) : StateRepos
     }
 
     override fun delete(connection: Connection, states: Collection<StateEntity>): Collection<String> {
-        val failedKeys = mutableListOf<String>()
-
-        states.forEach { state ->
-            connection.prepareStatement(queryProvider.deleteStatesByKey).use {
-                it.setString(1, state.key)
-                it.setInt(2, state.version)
-                it.executeUpdate().also { count ->
-                    if (count == 0) {
-                        failedKeys.add(state.key)
-                    }
-                }
+        return connection.prepareStatement(queryProvider.deleteStatesByKey).use { statement ->
+            // The actual state order doesn't matter, but we must ensure that the states are iterated over in the same
+            // order when examining the result as when the statements were generated.
+            val statesOrdered = states.toList()
+            statesOrdered.forEach { state ->
+                statement.setString(1, state.key)
+                statement.setInt(2, state.version)
+                statement.addBatch()
             }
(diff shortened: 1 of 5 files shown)


libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -55,21 +55,23 @@ class StateRepositoryImpl(private val queryProvider: QueryProvider) : StateRepos
...
+        return connection.prepareStatement(queryProvider.deleteStatesByKey).use { statement ->
+            // The actual state order doesn't matter, but we must ensure that the states are iterated over in the same
+            // order when examining the result as when the statements were generated.
+            val statesOrdered = states.toList()
+            statesOrdered.forEach { state ->
+                statement.setString(1, state.key)
+                statement.setInt(2, state.version)
+                statement.addBatch()
             }
+            statement.executeBatch().zip(statesOrdered).mapNotNull { (count, state) ->
```
sprintengine, line 70: Yeah this is the same way I implemented it, but then after some discussions we decided to avoid using `executeBatch()` because we have to determine which keys failed by zipping with the returned int array.

```
Submits a batch of commands to the database for execution and if all commands execute successfully, returns an array of update counts. The int elements of the array that is returned are ordered to correspond to the commands in the batch, which are ordered according to the order in which they were added to the batch. The elements in the array returned by the method executeBatch may be one of the following:

1 - A number greater than or equal to zero -- indicates that the command was processed successfully and is an update count giving the number of rows in the database that were affected by the command's execution

2 - A value of SUCCESS_NO_INFO -- indicates that the command was processed successfully but that the number of rows affected is unknown
If one of the commands in a batch update fails to execute properly, this method throws a BatchUpdateException, and a JDBC driver may or may not continue to process the remaining commands in the batch. However, the driver's behavior must be consistent with a particular DBMS, either always continuing to process commands or never continuing to process commands. If the driver continues processing after a failure, the array returned by the method BatchUpdateException.getUpdateCounts will contain as many elements as there are commands in the batch, and at least one of the elements will be the following:

3 - A value of EXECUTE_FAILED -- indicates that the command failed to execute successfully and occurs only if a driver continues to process commands after a command fails
```

I think we said a better approach would be to have the query return the keys that were deleted, and use that to drive which keys failed. 

Problem is that `executeBatch()` doesn't allow us to return anything from queries.

Is there a way to incorporate multiple DELETEs into one query, `RETURNING` the keys that were updated? May need some input from @ifrankrui

JamesHR3: I guess if you return the keys that succeeded you also need to do some comparison operations to figure out what things failed (in this case removing those keys from the input set), so you might not gain so much there. It would need a bit of benchmarking I think.

My (admittedly pretty unscientific) testing suggests this is faster despite the extra iteration in the code, as the database operations are significantly faster.

One problem we need to avoid for deletes is returning a key as failed if it never existed in the first place. This is handled at the moment by doing a `get` from the state manager on any key that the state repository returns. Conveniently this also handles any state where the value of `SUCCESS_NO_INFO` or `EXECUTE_FAILED` is returned. For deletes therefore I don't think these return codes are a problem, although I think the picture is more complicated for creates.

jujoramos: Maybe we should add the explanation about why we believe this is okay for the `DELETE` case to the code itself as a comment?, so people looking at this in the future don't need to reason about it from scratch?.

jujoramos (dismissed):
LGTM (I'd just add extra comments to the code about the reasoning for the `DELETE`).

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
JamesHR3: I'll put a comment here. I also updated the kdocs for the API to say that the delete call will return anything it's not sure about as if it had failed, and the state manager must double check anything that is returned.

jujoramos (approved):
LGTM

PR #5176, opened 2023-11-27
CORE-18457: Remove entity persistence duplication

emilybowe:
Ticket _CORE-17819: PersistenceWorker - Remove Kafka Implementation, Review Missing Tests and Duplications_ includes removal of ledger persistence and entity persistence duplication.

This PR addresses subtask **CORE-18457: Remove entity persistence duplication**

This will include an api change to remove unused entity persistence topic: https://github.com/corda/corda-api/pull/1369


LWogan (commented):
I don't see anything untoward here. The JIRA title has "review missing tests" in it. Are you happy there are no missing tests?

emilybowe:
> I don't see anything untoward here. The JIRA title has "review missing tests" in it. Are you happy there are no missing tests?

Yes I think so. I think we've discussed it in brief ITO and Miljenko dropped a note in our team slack on the review for missing tests on his opinion that I think makes sense. Also most tests that were there before were switched over to use the new processor and just ones that are no longer used (specific kafka ones) were deleted. 👍

LWogan (dismissed):
Looks good to me.

thiagoviana (dismissed):

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/external/events/AbstractPersistenceExternalEventFactory.kt
```diff
@@ -23,7 +22,6 @@ abstract class AbstractPersistenceExternalEventFactory<PARAMETERS : Any> :
         parameters: PARAMETERS
     ): ExternalEventRecord {
         return ExternalEventRecord(
-            topic = Schemas.Persistence.PERSISTENCE_ENTITY_PROCESSOR_TOPIC,
```
simon-johnson-r3, line 26: The latest version of `ExternalEventRecord` allows topic to be null but the docs in that file have not been update to reflect what the consequence of this is (they have been updated for the `key` field null). Is it possible to add that as part of this PR?

I assume topic can be null if you aren't intending to use Kafka? But as the user of this class needs to know whether that's true or not, I'm not sure reusing the same class just to transfer payload (just 1 out of 3 things) is the best idea.

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityRequestProcessor.kt
```diff
@@ -26,54 +20,29 @@ import java.time.Instant
...
- * [payloadCheck] is called against each AMQP payload in the result (not the entire Avro array of results)
  */
+
+@Suppress("LongParameterList")
 class EntityRequestProcessor(
     private val currentSandboxGroupContext: CurrentSandboxGroupContext,
     private val entitySandboxService: EntitySandboxService,
     private val responseFactory: ResponseFactory,
-    private val payloadCheck: (bytes: ByteBuffer) -> ByteBuffer,
+    override val requestClass: Class<EntityRequest>,
```
simon-johnson-r3: Would it be more usual to provide types as class parameters?

components/persistence/entity-processor-service/src/main/kotlin/net/corda/entityprocessor/EntityRequestSubscriptionFactory.kt
```diff
@@ -18,8 +16,6 @@ interface EntityRequestSubscriptionFactory {
      * @return A new subscription for [EntityRequest] messages
      */
 
-    fun create(config: SmartConfig): Subscription<String, EntityRequest>
```
simon-johnson-r3, line 21: You need to remove the comment above this. Also I think the header comment for the interface is now out of date?

simon-johnson-r3 (changes requested):

components/persistence/entity-processor-service-impl/src/main/kotlin/net/corda/entityprocessor/impl/internal/EntityRequestProcessor.kt
emilybowe: I checked with @JamesHR3 and you're right on this. I'll change this one in this PR and then raise a separate PR to fix the other workers.

emilybowe: I've raised CORE-18579

components/flow/flow-service/src/main/kotlin/net/corda/flow/application/persistence/external/events/AbstractPersistenceExternalEventFactory.kt
emilybowe: I've updated the kdocs here, and I've raised CORE-18580

components/persistence/entity-processor-service/src/main/kotlin/net/corda/entityprocessor/EntityRequestSubscriptionFactory.kt
emilybowe: I've updated these kdocs

simon-johnson-r3 (dismissed):

LWogan (approved):
LGTM

simon-johnson-r3 (approved):

PR #5177, opened 2023-11-27
CORE-18470: Change key of kafka message to random uuid

anien:
This ensures the work is evenly spread between Kafka partitions during key rotation.


dickon (approved):
LGTM

PR #5178, opened 2023-11-27
CORE-18168: Add new RBAC config sections to support bundle scripts

Tom-Fitzpatrick:


davidcurrie (dismissed):
nit: the sections were in alphabetical order but there's no particular requirement for them to be... you could always argue that "role" comes after "representational"!

vkolomeyko (dismissed):
LGTM

davidcurrie (approved):
🙇‍♂️

Tom-Fitzpatrick:
> nit: the sections were in alphabetical order but there's no particular requirement for them to be... you could always argue that "role" comes after "representational"!

Ah, I hadn't spotted that, restored the alphabetical order now, might as well keep it, thanks :)

vkolomeyko (approved):
LGTM

PR #5180, opened 2023-11-28
CORE-18030: Batch up state creation in the state manager

JamesHR3:
### Problem description

On creation of new states, the state manager iterates over each state and creates a new database transaction for each state. This means for each state the state manager must make a separate call to the database, which is slow.

### Solution

Batch up create calls to the state manager and submit them in a single database request.

This implementation uses a single parameterized query in a similar style to the update query. A temporary table of data is constructed from the input states, and then only those with keys not currently in the database are inserted. Keys that were successfully inserted are returned.

Some extra changes are required to internal state manager APIs to allow for a batched query to be run.


libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -39,22 +39,11 @@ class StateManagerImpl(
...
-                logger.warn("Failed to create state with id ${state.key}", e)
-                failures[state.key] = e
-            }
+        val successfulKeys = dataSource.connection.transaction { connection ->
+            stateRepository.create(connection, states.map { it.toPersistentEntity() })
         }
-
-        return failures
+        val failedKeys = states.map { it.key }.toSet() - successfulKeys.toSet()
+        return failedKeys.associateWith { Exception("Failed to create state") }
```
jujoramos: There's no point in returning this exception now, is it?. Users of the API can't do anything with it and it doesn't provide any extra info as to why the creation failed, so we should just change the API to return the failed keys.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/PostgresQueryProvider.kt
```diff
@@ -11,11 +11,18 @@ import net.corda.libs.statemanager.impl.model.v1.StateEntity.Companion.MODIFIED_
 
 class PostgresQueryProvider : AbstractQueryProvider() {
 
-    override val createState: String
-        get() = """
-            INSERT INTO $STATE_MANAGER_TABLE
-            VALUES (?, ?, ?, CAST(? as JSONB), CURRENT_TIMESTAMP AT TIME ZONE 'UTC')
-        """.trimIndent()
+    override fun createStates(size: Int): String = """
+        WITH data (key, value, version, metadata, modified_time) as (
```
jujoramos: Nitpick: let's use the existing constants for the column names.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
JamesHR3: Agreed!

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -28,10 +28,9 @@ interface StateManager : Lifecycle {
      * persisted and replicas of the underlying persistent storage, if any, are synced.
      *
      * @param states Collection of states to be persisted.
-     * @return Collection of keys for all those states that could not be persisted on the underlying persistent storage,
-     *          along with the actual reason for the failures.
+     * @return Collection of keys for all those states that could not be persisted on the underlying persistent storage.
      */
-    fun create(states: Collection<State>): Map<String, Exception>
+    fun create(states: Collection<State>): List<String>
```
jujoramos: Nitpick: a `Set` would be better here.

jujoramos (dismissed):

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -8,16 +8,28 @@ import net.corda.libs.statemanager.impl.repository.StateRepository
...
 
-    override fun create(connection: Connection, state: StateEntity) {
-        connection.prepareStatement(queryProvider.createState).use { statement ->
-            statement.setString(1, state.key)
-            statement.setBytes(2, state.value)
-            statement.setInt(3, state.version)
-            statement.setString(4, state.metadata)
-            statement.executeUpdate()
+    private companion object {
+        private const val CREATE_NUM_PARAMETERS = 4
```
sprintengine: I guess technically these constants should be alongside the query / determined by the provider

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -8,16 +8,28 @@ import net.corda.libs.statemanager.impl.repository.StateRepository
...
+        private const val CREATE_RESULT_COLUMN_INDEX = 1
+    }
+
+    override fun create(connection: Connection, states: Collection<StateEntity>): Collection<String> {
+        return connection.prepareStatement(queryProvider.createStates(states.size)).use { statement ->
+            states.forEachIndexed { idx, state ->
+                statement.setString(1 + (idx * CREATE_NUM_PARAMETERS), state.key)
```
sprintengine: wonder can we share / combine some of this logic between update's `getParameterIndex`

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
JamesHR3: I've managed to just remove this entirely.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
JamesHR3: I've opted for a slightly different approach to both of these cases - let me know what you think.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -8,16 +8,34 @@ import net.corda.libs.statemanager.impl.repository.StateRepository
...
-            statement.setInt(3, state.version)
-            statement.setString(4, state.metadata)
-            statement.executeUpdate()
+    private companion object {
+        private const val CREATE_RESULT_COLUMN_INDEX = 1
+    }
+
+    override fun create(connection: Connection, states: Collection<StateEntity>): Collection<String> {
+        return connection.prepareStatement(queryProvider.createStates(states.size)).use { statement ->
+            val indices = iterator {
```
sprintengine: Yeah this probably is a better aaproach, can you simplify it and just use generateSequence?

```
val indices = generateSequence(1) { it + 1 }.iterator()
```

sprintengine: Though the repository is probably still coupled to the query.

What if we extracted the creation of the prepared statement to the query provider, and we pass it all the states rather than the size?

Though, https://r3-cev.atlassian.net/browse/CORE-18607 or https://r3-cev.atlassian.net/browse/CORE-16663 could be responsible for that sort of refactoring

JamesHR3: Good point, although I think this needs to be converted to an iterator afterwards.

vkolomeyko (dismissed):

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -38,23 +38,11 @@ class StateManagerImpl(
...
-            try {
-                dataSource.connection.transaction {
-                    stateRepository.create(it, state)
-                }
-            } catch (e: Exception) {
-                logger.warn("Failed to create state with id ${state.key}", e)
-                failures[state.key] = e
-            }
+    override fun create(states: Collection<State>): Set<String> {
+        val successfulKeys = dataSource.connection.transaction { connection ->
```
jujoramos, line 43: This will create an unnecessary db transaction if the list is empty, won't it?, maybe use the same approach as the other batch operations? -? `if (states.isEmpty()) return emptyMap()`.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -187,6 +186,12 @@ class StateManagerIntegrationTest {
         )
     }
 
+    @Test
+    fun `create called with no states succeeds`() {
+        val states = setOf<State>()
+        assertThat(stateManager.create(states)).isEmpty()
+    }
```
jujoramos: Related to my comment about empty state list, this one likely goes in `StateManagerImplTest` instead.

jujoramos (dismissed):
Minor comments, otherwise LGTM.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
JamesHR3: This actually flushed out a bug that I wouldn't have hit in an integration test I think, in that the SQL statement is invalid in that case.

However, I think with the fix in the unit test should instead assert that no DB transaction is made as you said. I'll update both.

jujoramos (approved):
LGTM

LWogan (approved):
LGTM

conal-smith (approved):

vkolomeyko (approved):
LGTM

PR #5181, opened 2023-11-28
CORE-18581 - SM.update() returns null for state keys that don't exist

conal-smith:
Problem:
It's possible when state manager updates states, another thread could have deleted the state. Currently this failure is not recorded, the caller simply doesn't know that it even failed.

Solution:
- `SM.update()` should return null for these keys indicating the State no longer exists.
- Mediator should not try to re-attempt or process any output events for this event.
- Added integration test that utilizes null returned from the API to test update() in multi-threaded test. Deletes are now included in the list of failures.

### DIFF
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
@@ -223,13 +223,18 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
                 val failedToCreate = stateManager.get(failedToCreateKeys.keys)
                 val failedToDelete = stateManager.delete(statesToDelete.values.mapNotNull { it })
                 val failedToUpdate = stateManager.update(statesToUpdate.values.mapNotNull { it })
-                states = failedToCreate + failedToDelete + failedToUpdate
+                val failedToUpdateOptimisticLockFailure = failedToUpdate.mapNotNull { (key, value) -> value?.let { key to it } }.toMap()
+                val failedToUpdateStateDoesNotExist = (failedToUpdate - failedToUpdateOptimisticLockFailure).map { it.key }
+
+                states = failedToCreate + failedToDelete + failedToUpdateOptimisticLockFailure
+
                 groups = if (states.isNotEmpty()) {
                     allocateGroups(flowEvents.filterKeys { states.containsKey(it) }.values.flatten())
                 } else {
                     listOf()
                 }
                 states.keys.forEach { asynchronousOutputs.remove(it) }
+                failedToUpdateStateDoesNotExist.forEach { asynchronousOutputs.remove(it) }
                 sendAsynchronousEvents(asynchronousOutputs.values.flatten())
             }
             metrics.commitTimer.recordCallable {
--- a/libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
+++ b/libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
@@ -306,8 +306,9 @@ class StateManagerIntegrationTest {
         assertThat(failedUpdates).containsOnlyKeys(conflictingKeys)
         assertSoftly {
             failedUpdates.values.map { state ->
+                it.assertThat(state).isNotNull
                 // update A has already bumped the version by 1, causing B's state update to fail
-                it.assertThat(state.version).isEqualTo(statesToUpdateB[state.key]!!.version + 1)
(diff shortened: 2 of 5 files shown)


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MultiSourceEventMediatorImpl.kt
```diff
@@ -223,13 +223,18 @@ class MultiSourceEventMediatorImpl<K : Any, S : Any, E : Any>(
...
+
+                states = failedToCreate + failedToDelete + failedToUpdateOptimisticLockFailure
+
                 groups = if (states.isNotEmpty()) {
                     allocateGroups(flowEvents.filterKeys { states.containsKey(it) }.values.flatten())
                 } else {
                     listOf()
                 }
                 states.keys.forEach { asynchronousOutputs.remove(it) }
+                failedToUpdateStateDoesNotExist.forEach { asynchronousOutputs.remove(it) }
```
sprintengine, line 237: I'm actually not sure we want to remove the async outputs when the state doesn't exist... i.e. I guess there could be cleanup events, errors for counterparties, etc..

We remove the outputs for other failed states because we're going to retry them, but if the states were deleted, this is probably normal for states to be deleted at the end and now if we just discard all the outputs, it'll probably mess up some session close logic / counterparty errors / cleanup.. @LWogan do you want to opine?

LWogan: What is the scenario where we try to update but the state is deleted? 

If we take the flow engine as an example. If some process deleted the state from under us we'd assume that it did the cleanup etc as part of that operation. Our async outputs from the update could in theory conflict with that. e.g async output here to update the flow to RUNNING when some other process has deleted the checkpoint and set flow status to be KILLED. We could end up resetting back to RUNNING which would be incorrect.

I think if our state isn't saved then we should abandon our async outputs and trust that the other process did whatever needs to be done when deleting.

sprintengine: Yeah sound right, if our process attempts to update but it was deleted, we can safely assume the other process has output the necessary events and statuses, and we can abandon our changes, so I think I was right in the first place to do that.

libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -358,6 +359,54 @@ class StateManagerIntegrationTest {
...
+            val failedDeletes = stateManager.delete(statesToDelete).map { it.key }
+            MultiThreadedTestHelper.FailedKeysSummary(failedUpdates, failedDeletes)
+        }
+
+        val allFailedUpdates = threadResults.map { it.failedKeysSummary.failedUpdates }.flatten()
+        val allFailedDeletes = threadResults.map { it.failedKeysSummary.failedDeletes }.flatten()
+        val expectedFailedKeys = threadResults.map { it.assignedStateGrouping.overlappingStates }.flatten().map { it.key }
+
+        // if a thread tries to update a state that was already deleted, it gets that key back as a "failed key", associated with null.
+        // if a thread tries to delete a thread that was already updated, it gets that key back as a "failed key".
```
jujoramos:
```suggestion
        // if a thread tries to delete a state that was already updated, it gets that key back as a "failed key".
```

jujoramos (dismissed):

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -71,26 +71,41 @@ class StateManagerImpl(
...
 
-            return if (failedUpdates.isEmpty()) {
-                emptyMap()
-            } else {
-                logger.warn("Optimistic locking check failed while updating States ${failedUpdates.joinToString()}")
-                get(failedUpdates)
-            }
+            if (failedUpdates.isEmpty()) return emptyMap()
+
+            return getFailedUpdates(failedUpdates)
```
JamesHR3: I think usually you'd write this as

```kotlin
return if (failedUpdates.isEmpty()) {
    emptyMap()
} else {
    getFailedUpdates(failedUpdates)
}
```

JamesHR3 (dismissed):

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
sprintengine: Yeah probably is a bit weird reading it, it basically reads like 
```
if success case - return empty

return failure case
```

LWogan (approved):
LGTM

PR #5185, opened 2023-11-30
CORE-18504: set up state manager for crypto, and populate from REST worker

dickon:
- Adds support for state manager in rest and crypto workers
- Adds JDBC driver for rest worker to allow access to state manager


vkolomeyko (dismissed):
LGTM (on behalf @corda/rest)

processors/member-processor/src/integrationTest/kotlin/net/corda/processor/member/MemberProcessorTestUtils.kt
```diff
@@ -290,6 +291,9 @@ class MemberProcessorTestUtils {
         fun Publisher.publishDefaultCryptoConf(cryptoConfig: SmartConfig) =
             publishConf(ConfigKeys.CRYPTO_CONFIG, cryptoConfig.root().render())
 
+        fun Publisher.publishEmptyStateManagerConf() =
+            publishConf(ConfigKeys.STATE_MANAGER_CONFIG, SmartConfigImpl.empty().root().render())
+
```
anien, line 296: Is this test related to our changes or was it missed in previous PRs?

charts/corda/values.schema.json
```diff
@@ -2045,7 +2062,8 @@
                         },
                         {
                             "required": [
-                                "service"
+                                "service",
+                                "stateManager"
```
anien, line 2067: I think this most of the yaml changes need to be reviewed by @jujoramos.
Specifically here we add `stateManager` in required field for the rest worker, but we don't do the same for crypto worker and I don't understand all the implications of it.

charts/corda/values.yaml
```diff
@@ -656,7 +706,57 @@ workers:
...
+        # -- the State Manager database type
+        type: "postgresql"
+        # -- the State Manager database port
+        port: 5432
+        # -- the State Manager database name
+        name: crypto_state_manager
+        # the State Manager database user configuration
+        username:
+          # -- the State Manager database user
+          value: "rest_state_manager_user"
```
anien, line 725: I would create a ticket for now to make sure we restrict the usage of this particular state manager db to only rest and crypto worker. Currently, it seems the default configuration is that it's shared between all workers that can use state manager.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/response/KeyRotationStatusResponse.kt
```diff
@@ -0,0 +1,18 @@
+package net.corda.crypto.rest.response
+
+/**
+ * The result of a key rotation request.
+ *
+ * @param requestId The unique ID for the key rotation start request.
+ * @param oldKeyAlias Alias of the key to be rotated.
+ * @param newKeyAlias Alias of the new key the [oldKeyAlias] key will be rotated with.
+ */
```
anien: We might need to update Kdocs. But I think this will change further, so maybe a ticket for now that we need to review kdocs and parameters for this file?

components/crypto/crypto-rest/src/test/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceTest.kt
```diff
@@ -2,46 +2,70 @@ package net.corda.crypto.rest.impl
...
+import org.mockito.kotlin.mock
 import org.mockito.kotlin.whenever
+import java.lang.UnsupportedOperationException
+import java.time.Instant
 
 class KeyRotationRestResourceTest {
 
     private lateinit var publisherFactory: PublisherFactory
-    private lateinit var publisher: Publisher
+    private lateinit var publishToKafka: Publisher
```
anien, line 43: I would rename this to `kafkaPublisher`.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -45,78 +56,62 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
 
-    private val uploadTopic = REKEY_MESSAGE_TOPIC
-    private var publisher: Publisher? = null
+    private val configKeys = setOf(
+        ConfigKeys.MESSAGING_CONFIG,
+        ConfigKeys.CRYPTO_CONFIG,
+        ConfigKeys.STATE_MANAGER_CONFIG,
+    )
+
+    private var publishToKafka: Publisher? = null
```
anien, line 72: rename to `kafkaPublisher`?

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -45,78 +56,62 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
-    }
+    override val isRunning: Boolean
+        get() = lifecycleCoordinator.isRunning
 
+    private var subscriptionRegistrationHandle: RegistrationHandle? = null
 
-    override fun getKeyRotationStatus(): List<Pair<String, List<String>>> {
-        TODO("Not yet implemented")
+    override fun start() {
+        logger.info("KeyRotationRestResource starting.")
```
anien: I would either have this as a debug or removed completely.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -45,78 +56,62 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
-            null,
-            null
-        )
-
-        publisher?.publish(listOf(Record(uploadTopic, requestId, keyRotationRequest)))
-            ?: throw ServiceUnavailableException("Key rotation resource has not been initialised.")
-
-        return ResponseEntity.accepted(KeyRotationResponse(requestId, oldKeyAlias, newKeyAlias))
+    override fun stop() {
+        logger.info("KeyRotationRestResource stopping.")
```
anien: Same as above.

anien (commented):
First pass of the PR, I will need to do another one.

processors/member-processor/src/integrationTest/kotlin/net/corda/processor/member/MemberProcessorTestUtils.kt
dickon: This is now required since the crypto procesor waits for the STATE_MANAGER_CONFIG to arrive before it starts, so never starts in this test case, so the member processor integration test never gets past the lifecycle startup sequence and timeout.

Basically we have to touch everywhere which bootstraps crypto processor, and that includes the OSGi tests of things that consume the crypto procesosr, which include the memeber processor.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
dickon: Not keen. `kafkaPublisher` suggests we are publishign the whole of Kafka.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
dickon: done

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
dickon: done

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/response/KeyRotationStatusResponse.kt
dickon: I pulled this out in to another branch

yift-r3 (dismissed):
LGTM (network only)

charts/corda/values.schema.json
sprintengine: What does Crypto worker do if it does not receive a stateManager configuration? If SM is now a required piece of config for the crypto worker, we should reflect it in the schema

charts/corda/values.yaml
sprintengine: I have a design for a closely related ticket in progress: https://r3-cev.atlassian.net/browse/CORE-16416

charts/corda/values.yaml
```diff
@@ -656,7 +706,57 @@ workers:
...
+      # State Manager database configuration
+      db:
+        # -- the State Manager database host
+        host: null
+        # -- the State Manager database type
+        type: "postgresql"
+        # -- the State Manager database port
+        port: 5432
+        # -- the State Manager database name
+        name: crypto_state_manager
```
sprintengine: The intent is to share the one state manager between crypto and rest, right?

This causes the default behavior to point at a different one, this should match the one configured in rest worker.

(And you could probably name it something related to the type of state it holds.)

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -125,25 +118,118 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
+        val newPublisher =
+            publisherFactory.createPublisher(PublisherConfig("KeyRotationRestResource", false), messagingConfig)
+        newPublisher.start()
+        publishToKafka = newPublisher
     }
 
-    override fun start() {
-        lifecycleCoordinator.start()
+    override fun getKeyRotationStatus(requestId: String): List<Pair<String, String>> {
+        val entries = stateManager!!.get(listOf(requestId))
```
sprintengine, line 162: should you check this isn't null first?

charts/corda/values.yaml
sprintengine: Probably the easiest way to restrict the usage is via permissions and credentials. 

Our auto-bootstrapping process uses the root/admin Postgres credentials from the `bootstrap.stateManager.crypto` config, and grants permissions to the user configured at `workers.crypto.stateManager.db.username`.

So as long as the boot config for crypto and rest both have the same pg instance's admin creds, if the credentials at `workers.crypto.stateManager.db.username` and `workers.rest.stateManager.db.username` are different, the boot process should configure both these users to have access.

charts/corda/values.yaml
dickon: State manager is used to store the progress on key rotation, both at request level and individual rewrap work item. We've got more code under development which actually does that, but for now I wanted to get it in.

So, without state manager, the crypto processor can work and even do key rotation, but the operator will get no information about key rotation progress.

sprintengine: I thought the intent was to share one StateManager instance between REST and crypto workers, no? e.g

image

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: Now it sounds more like a function name, not a property name.

davidcurrie (commented):
I've only looked at the charts side of things but I concur with the other comments. If this is the REST worker accessing the crypto worker's DB then the DB details should only be defined on the crypto worker side with the REST worker just having different credentials (and it should be clear in the way those credentials are supplied that they are for accessing the crypto worker's DB, not any potential future state manager that the REST worker might have of its own).

anien:
> I've only looked at the charts side of things but I concur with the other comments. If this is the REST worker accessing the crypto worker's DB then the DB details should only be defined on the crypto worker side with the REST worker just having different credentials (and it should be clear in the way those credentials are supplied that they are for accessing the crypto worker's DB, not any potential future state manager that the REST worker might have of its own).

I have created a ticket for this https://r3-cev.atlassian.net/browse/CORE-18642

charts/corda/values.yaml
anien: Correct, we want to share one StateManager instance between the REST and CRYPTO worker. But we want both rest and crypto worker to be able to read and write.

charts/corda/values.yaml
anien: I have created a ticket for this, so we can work on it separately: https://r3-cev.atlassian.net/browse/CORE-18642

charts/corda/values.schema.json
anien: Changed. State manager is now required for the crypto worker as well. We do need to use it in the key rotation, so I believe it's the best approach to mark it as required.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: I might leave it to another PR for now.

charts/corda/values.yaml
jujoramos: @anien : I believe @conalsmith-r3 's concern is still in place, even if https://r3-cev.atlassian.net/browse/CORE-16416 is implemented. As things stand right now, the `crypto-worker` will use `crypto_state_manager` database and `rest-worker` will use `rest_state_manager`, so they won't be sharing anything.

charts/corda/values.yaml
```diff
@@ -1197,6 +1297,57 @@ workers:
...
+          value: ""
+          # the State Manager database password secret configuration; used in preference to value if name is set
+          valueFrom:
+            # the State Manager database password secret key reference
+            secretKeyRef:
+              # -- the State Manager database password secret name
+              name: ""
+              # -- the State Manager database password secret key
+              key: ""
+        # flow mapper worker JDBC connection pool configuration for State Manager DB
```
jujoramos:
```suggestion
        # rest worker JDBC connection pool configuration for State Manager DB
```

charts/corda/values.yaml
```diff
@@ -1197,6 +1297,57 @@ workers:
...
+        connectionPool:
+          # -- flow mapper worker maximum JDBC connection pool size for state manager DB
```
jujoramos:
```suggestion
          # -- rest worker maximum JDBC connection pool size for state manager DB
```

charts/corda/values.yaml
```diff
@@ -1197,6 +1297,57 @@ workers:
...
+          maxSize: 5
+          # -- flow mapper worker minimum JDBC connection pool size for state manager DB; null value means pool's min size will default to pool's max size value
```
jujoramos:
```suggestion
          # -- rest worker minimum JDBC connection pool size for state manager DB; null value means pool's min size will default to pool's max size value
```

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -112,22 +108,35 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
                     lifecycleCoordinator.updateStatus(LifecycleStatus.DOWN)
                 }
             }
+
             is ConfigChangedEvent -> {
                 initialise(event.config)
+
+                val stateManagerConfig = event.config.getConfig(ConfigKeys.STATE_MANAGER_CONFIG)
+
+                logger.info("State manager config is $stateManagerConfig")
```
jujoramos: Won't this leak credentials into the log?.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -112,22 +108,35 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
+                stateManager?.stop()
+                stateManager = stateManagerFactory.create(stateManagerConfig).also { it.start() }
+                logger.info("State manager created and started $stateManager")
```
jujoramos: Nitpick: the state manager doesn't have a `toString()` so this will likely show rubbish in the logs. you might want to log the `name` instead (or nothing at all).

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -33,8 +40,12 @@ import org.osgi.service.component.annotations.Component
 import org.osgi.service.component.annotations.Reference
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
+import java.time.Instant
 import java.util.UUID
 
+private const val UPLOAD_TOPIC = REKEY_MESSAGE_TOPIC
```
jujoramos: Nitpick: do you need this const?, why don't just use `REKEY_MESSAGE_TOPIC`?.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -138,12 +147,94 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
+    publishStates: ((List<State>) -> Unit),
+    publishRequests: ((List<Record<String, KeyRotationRequest>>) -> Unit)
+): ResponseEntity<KeyRotationResponse> {
+    // We cannot validate oldKeyAlias or newKeyAlias early here on the client side of the RPC since
+    // those values are considered sensitive.
+
+    // We need to create a Record that tells Crypto processor to do key rotation
+    // Do we need to start the publisher? FlowRestResource is not starting its publisher for some reason
+
+    val requestId = UUID.randomUUID().toString()
```
jujoramos, line 201: Can we inject the `clientRequestId` or something set by the user here?, when the time comes, that will make troubleshooting easier 😅.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -207,13 +219,26 @@ class CryptoProcessorImpl @Activate constructor(
             is ConfigChangedEvent -> {
                 tenantInfoService = startTenantInfoService()
                 cryptoService = startCryptoService(event.config.getConfig(CRYPTO_CONFIG), tenantInfoService)
+                val bootConfig = bootConfigRecord?: throw IllegalStateException("boot config not available")
+                val stateManagerConfig: SmartConfig? = if (bootConfig.hasPath(StateManagerConfig.STATE_MANAGER))
+                            bootConfig.getConfig(StateManagerConfig.STATE_MANAGER)
+                        else
+                            null
+                logger.info("making statement manager $stateManagerConfig")
+                logger.info("state manager config is $stateManagerConfig")
```
jujoramos: Similar as my other comments: I don't think these log messages under `INFO` level will be useful as they stand right now, and they might leak connection credentials as well.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: No real need for the const. Using topic name directly now.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: Possibly we can. Let's discuss with others.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
anien: Removed.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: I change the log message to debug and use `.name` even though the name is mainly just random, but in case of some debugging it might be useful.

charts/corda/values.yaml
anien: Ah, got your point now. In this PR, we are only bringing support for the state manager, and start populating it. We will have a follow up PR where we use it properly to share the state between the workers. But very good point! I am going to submit the change here after testing, so we can have it already resolved.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: It doesn't show any details as it's currently implemented. But I also removed it now.

charts/corda/values.yaml
anien: Done.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: From the offline conversation it was decided we keep the current work as is, but will add feature ticket to our backlog.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: Hm, `.name` is breaking few tests. I have removed it and keeping for now just the string.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: Done. PTAL.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/KeyRotationRestResource.kt
```diff
@@ -29,11 +29,14 @@ interface KeyRotationRestResource : RestResource {
...
-        path = "unmanaged/rotation",
+        path = "unmanaged/rotation/{requestId}",
         description = "This method gets the status of the current rotation.",
         responseDescription = "",
     )
-    fun getKeyRotationStatus(): List<Pair<String, List<String>>>
+    fun getKeyRotationStatus(
+        @RestPathParameter(description = "The requestId obtained when starting key rotation request.")
+        requestId: String
+    ): List<Pair<String, String>>
```
vkolomeyko, line 39: Hm... status of rotation as a list of pairs... IMO this is way to generic/open to interpretation.
Can we not create a better `data class` which will represent the status of rotation better?

anien: I already have this changed to a data class in the follow up PR.

applications/workers/release/rest-worker/build.gradle
```diff
@@ -83,6 +83,13 @@ dependencies {
     e2eTestImplementation "net.corda:corda-avro-schema"
     e2eTestImplementation "net.corda:corda-topic-schema"
     e2eTestRuntimeOnly 'org.osgi:osgi.core'
+
+    // This puts the jdbc driver into the docker image in the /opt/jdbc-driver folder
+    // this folder can contain many jdbc drivers (and DataSourceFactory provider bundles).
+    // Postgres doesn't need a DataSourceFactory provider bundle (e.g. pax-jdbc), because
+    // the postgres devs have written their own and it's in this jar (PGDataSourceFactory).
+    dockerImageJdbc "org.postgresql:postgresql:$postgresDriverVersion"
```
josephzunigadaly, line 91: Is this needed in the rest worker?

anien: Because we need db access from the rest worker to state manager db. It was now decided that we would only need read access, so I am going to check if in this case we still need this.

anien: @josephzunigadaly yes, we still need this.

josephzunigadaly: Thank you Anna. I wasn't expecting that to be the answer. I think I'm not the right person to review this. I'd need more information on the design being implemented.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -138,12 +145,99 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
+        val result = mutableListOf<Pair<String, String>>()
+        entries.forEach { entry ->
+            val keyRotationStatus = deserializer.deserialize(entry.value.value)!!
+            result.add(keyRotationStatus.oldParentKeyAlias to keyRotationStatus.requestId.toString())
+        }
+        return result
+    }
+
+    override fun startKeyRotation(oldKeyAlias: String, newKeyAlias: String): ResponseEntity<KeyRotationResponse> {
+        val pTK = tryWithExceptionHandling(logger, "start key rotation") {
```
simon-johnson-r3: I'm not sure what's going on here. The purpose of `checkNotNull` is to throw `IllegalStateException` where you don't expect that to be the case but want to write defensive code. Here it's wrapped with `tryWithExceptionHandling` which implies you are trying to catch some exceptions, but that method doesn't really deal with `IllegalStateException` and even if it did, why would you explicitly throw that just to immediately suppress it?

I would also try to remove the use of these short variable names.

applications/workers/release/rest-worker/build.gradle
sprintengine: Yes State Manager needs a JDBC driver to interact with the state storage for now as the back-end is a postgres db.

vkolomeyko (dismissed):
LGTM

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -138,12 +145,99 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
+        0,
+        0,
+        Instant.now(),
+        Instant.now()
+    )
+
+    publishRequests(listOf(Record(REKEY_MESSAGE_TOPIC, requestId, keyRotationRequest)))
+
+    val serialized = serializeStatus(status)
+    checkNotNull(serialized)
```
simon-johnson-r3: checkNotNull is designed to be used like a wrapper:
```
val serialized = checkNotNull(serializeStatus(status))
```

The fact you have to do this at all is actually a problem with the `CordaAvroSerializer` interface IMHO, it actually throws on error and shouldn't then return a nullable. But that's not for this PR...

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -76,6 +90,30 @@ class CryptoRekeyBusProcessor(
...
+                    i+1,
+                    targetWrappingKeys.count(),
+                    Instant.ofEpochMilli(timestamp), // TODO get creation timestamp rather than this hack
+                    now
+                )
+                logger.info("Rewrapped ${wrappingKeyInfo.alias} in tenant ${tenantId} from "+
+                        "${wrappingKeyInfo.parentKeyAlias} to ${request.newParentKeyAlias}; "+
+                        "generation number now ${generationNumber}: status $status")
+                val flattend = serializer.serialize(status)
+                require(flattend!=null)
```
simon-johnson-r3: `require` is for arguments to this function only (it throws `IllegalArgumentException`) here you should wrap this in `checkNotNull`, see comment above.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -76,6 +90,30 @@ class CryptoRekeyBusProcessor(
...
+                val now = Instant.now()
+                val status = KeyRotationStatus(
+                    request.requestId,
+                    request.managedKey,
+                    request.oldParentKeyAlias,
+                    request.newParentKeyAlias,
+                    request.oldGeneration,
+                    request.tenantId,
+                    generationNumber,
+                    i+1,
```
simon-johnson-r3: `i + 1`

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -76,6 +90,30 @@ class CryptoRekeyBusProcessor(
...
+                    targetWrappingKeys.count(),
+                    Instant.ofEpochMilli(timestamp), // TODO get creation timestamp rather than this hack
```
simon-johnson-r3: we shouldn't add comments like this, either the code should work or it shouldn't be submitted, I would remove it and add a JIRA ticket referencing where it is in the code and why it needs to change at a later date

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -76,6 +90,30 @@ class CryptoRekeyBusProcessor(
                     )
                 }.toList()
             )
+
+            targetWrappingKeys.forEachIndexed { i, (tenantId, wrappingKeyInfo) ->
+                val generationNumber = 0 // TODO find out the correct generation number
```
simon-johnson-r3: we shouldn't add comments like this, either the code should work or it shouldn't be submitted, I would remove it and add a JIRA ticket referencing where it is in the code and why it needs to change at a later date

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -202,21 +218,34 @@ class CryptoProcessorImpl @Activate constructor(
                     setStatus(event.status, coordinator)
                 }
             }
+
             is ConfigChangedEvent -> {
                 tenantInfoService = startTenantInfoService()
                 cryptoService = startCryptoService(event.config.getConfig(CRYPTO_CONFIG), tenantInfoService)
+                val bootConfig = bootConfigRecord ?: throw IllegalStateException("Boot config not available.")
```
simon-johnson-r3: val bootConfig = checkNotNull(bootConfigRecord) { "Boot config not available." }

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -202,21 +218,34 @@ class CryptoProcessorImpl @Activate constructor(
...
+                val stateManagerConfig: SmartConfig? = if (bootConfig.hasPath(StateManagerConfig.STATE_MANAGER))
```
simon-johnson-r3: This is a long-winded way of writing:
```
if (bootConfig.hasPath(StateManagerConfig.STATE_MANAGER)) {
    val stateManagerConfig = bootConfig.getConfig(StateManagerConfig.STATE_MANAGER)
    stateManager = stateManagerFactory.create(stateManagerConfig).also {
        it.start()
    }
}
```

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -202,21 +218,34 @@ class CryptoProcessorImpl @Activate constructor(
...
+                    bootConfig.getConfig(StateManagerConfig.STATE_MANAGER)
+                else
+                    null
+
+                if (stateManagerConfig != null) {
+                    stateManager = stateManagerFactory.create(stateManagerConfig).also {
+                        it.start()
+                    }
+                }
+                logger.debug("State manager created and started.")
```
simon-johnson-r3: This isn't true in all cases. Should we throw here if it wasn't started? I assume the crypto worker isn't operable at this point. It would also be due to bad parameters so the whole cluster is unlikely to ever recover from this state.

We should definitely not log this here though, it could be misleading.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -138,12 +145,99 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
 
-    override fun start() {
-        lifecycleCoordinator.start()
+    @VisibleForTesting
+    fun initialise(config: Map<String, SmartConfig>) {
+        val messagingConfig = config.getConfig(ConfigKeys.MESSAGING_CONFIG)
+
+        // Initialise publisher with messaging config
+        publishToKafka?.close()
+        val newPublisher =
```
simon-johnson-r3: The use of `newPublisher` is unnecessary with this implementation. If you are perhaps expecting something to throw before setting `publishToKafka` to the new value, then you should set `publishToKafka` to `null` after the close.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -138,12 +145,99 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
+            result.add(keyRotationStatus.oldParentKeyAlias to keyRotationStatus.requestId.toString())
+        }
+        return result
+    }
+
+    override fun startKeyRotation(oldKeyAlias: String, newKeyAlias: String): ResponseEntity<KeyRotationResponse> {
+        val pTK = tryWithExceptionHandling(logger, "start key rotation") {
+            checkNotNull(publishToKafka)
+        }
+        val sM = tryWithExceptionHandling(logger, "retrieve key rotation status") {
```
sprintengine: this check on stateManager has the wrong operation name.

Probably better off just wrapping both checks in one go, you can use !! after this block

```
tryWithExceptionHandling(logger, "start key rotation") {
    checkNotNull(publishToKafka)
    checkNotNull(stateManager)
}
```

If you're coming back to this code and can address it as a follow up, I'm happy to approve this PR.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -138,12 +145,99 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
+            publisherFactory.createPublisher(PublisherConfig("KeyRotationRestResource", false), messagingConfig)
+        newPublisher.start()
+        publishToKafka = newPublisher
     }
 
-    override fun stop() {
-        publisher?.close()
-        lifecycleCoordinator.stop()
+    override fun getKeyRotationStatus(requestId: String): List<Pair<String, String>> {
+        val sM = tryWithExceptionHandling(logger, "retrieve key rotation status") {
```
simon-johnson-r3: See comment in next function.

simon-johnson-r3 (changes requested):

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
sprintengine: +1 I agree there's a couple things not quite right here.

It does deal with `IllegalStateException` implicitly in that it'll convert it to a `InternalServerException`, which is also arguably not correct, `ServiceUnavailableException` could be a better exception to throw there. But yes, a better approach might be to just check if these required fields are null, and if so, throw the exception, rather than wrapping a check with another wrapper. 

```
fun validateDependencies() {
  if (state manager == null || publishToKafka == null) throw ServiceUnavailableException
}
```

anien: We want to return `HttpApiException` so it can return HTTP code as this is on the rest worker. `IllegalStateException` is not `HttpApiException`. `tryWithExceptionHandling` throws `InternalServerException` if it doesn't match any other previous exceptions, which in our case it doesn't. `InternalServerException` was recommended by @vkolomeyko.

sprintengine: Just thinking aloud, throwing IntenralServerException is probably OK. `ServiceUnavailableException` makes sense when something calls the crypto worker REST endpoints when they aren't "up" (or lifecycle coordinator isn't up) - I'm not sure we do that check I'd need to look at the javalin handler. So for this PR, when the component is UP but a dependency is null, something has went wrong, InternalServerException is probably appropriate.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: Ah yes, that was copy paste error. I've changed it according your recommendation.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: Changed.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
anien: Done. And thanks for clarification.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
anien: This has now changed. We don't know how many keys have we rotated at this staged. I added a comment.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
anien: The comment is actually wrong. It's the right thing to do as now we provide the timestamp when we create the record.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
anien: This is a leftover from the previous change. We won't know the generation number at this time. So I have updated the comment.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
anien: Updated.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
anien: Updated.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
anien: I added a check and removed the log.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: I have removed the use of the `newPublisher`

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: Hope I have addressed this.

components/crypto/crypto-rest/build.gradle
```diff
@@ -9,11 +9,15 @@ dependencies {
...
     implementation project(':libs:rest:rest')
+    implementation project(":libs:rest:rest-common")
     implementation project(':libs:messaging:messaging')
     implementation project(':libs:configuration:configuration-core')
     implementation project(':components:configuration:configuration-read-service')
     implementation project(':libs:crypto:crypto-core')
     implementation project(':libs:crypto:crypto-config-impl')
+    implementation project(":libs:state-manager:state-manager-api")
+    implementation project(':libs:serialization:serialization-avro')
+    runtimeOnly project(':libs:state-manager:state-manager-db-impl')
```
sprintengine: state manager implementation runtime dependency should just be at the processor level, e.g. [crypto processor](https://github.com/corda/corda-runtime-os/pull/5185/files#diff-0acc5ebb3e3fa40455c3dc052674ff60d272653a540b8dac8c3d8fcea4f27413), so I think you should be able to remove this.

edit - this module is imported by the rest processor, so you should be able to move this dependency to the rest-processor's build.gradle.

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
anien: I had to removed the check because of failing OSGi tests. Log is removed.

components/crypto/crypto-rest/build.gradle
anien: runtimeOnly dependency is now moved to rest processor module.

simon-johnson-r3 (dismissed):

conal-smith (dismissed):

ronanbrowne (dismissed):
LGTM

ronanbrowne (approved):

conal-smith (approved):

conal-smith (approved):
lgtm

vkolomeyko (approved):
LGTM

yift-r3 (approved):
Network changes LGTM

simon-johnson-r3 (approved):

PR #5186, opened 2023-11-30
CORE-17816: contract verifying notary implementation

jennyang-r3:
## Contract verifying notary
This PR implements a _contract-verifying notary_, according to [this design](https://github.com/corda/platform-eng-design/blob/main/core/corda-5/corda-5.2/notary/contract-verifying-notary/index.md#proposed-solution)

**Test**
I manually tested by packaging the `contract-verifying-notary` in the smoke test and set `isBackchainRequired` to `false` and checked the log line:
```
Successfully notarized transaction: SHA-256D:510A13043C5336B897A1B1E22F16402C542C519D6307ED916A43298D416984A8 with notary: CN=Notary-b96c3c03-b746-4be3-9643-46e853685452, OU=Application, O=R3, L=London, C=GB
```


components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/filtered/factory/FilteredTransactionFactoryImpl.kt
```diff
@@ -85,19 +85,32 @@ class FilteredTransactionFactoryImpl @Activate constructor(
             }
 
         val merkleProof = when (parameters) {
-            is ComponentGroupFilterParameters.AuditProof<*> -> {
+            is AuditProof<*> -> {
 
                 val skipFiltering = componentGroupIndex == 0
 
+                // change filtering logic here
```
blsemo: Why is this comment here? Does this require further change, or is it just a left-over that should be deleted?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/filtered/UtxoFilteredTransactionBuilderImpl.kt
```diff
@@ -115,7 +116,18 @@ data class UtxoFilteredTransactionBuilderImpl(
...
                 UtxoComponentGroup.OUTPUTS.ordinal,
                 ContractState::class.java,
-                predicate
+                AuditProofPredicate.Content(predicate)
+            )
+        )
+    }
+
+    @Suspendable
+    override fun withOutputStates(indexes: MutableList<Int>): UtxoFilteredTransactionBuilder {
```
blsemo: Why is that a mutable list?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-client/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/client/ContractVerifyingNotaryClientFlowImpl.kt
```diff
@@ -0,0 +1,88 @@
...
+
+    @CordaInject
+    private lateinit var digestService: DigestService
+
+    @CordaInject
+    private lateinit var utxoLedgerService: UtxoLedgerService
+
+    @Suspendable
+    override fun call(): List<DigitalSignatureAndMetadata> {
+        logger.info("Calling Contract Verifying Notary client...")
```
blsemo: I am not sure that should be a log message at `INFO` level.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-client/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/client/ContractVerifyingNotaryClientFlowImpl.kt
```diff
@@ -0,0 +1,88 @@
...
+            null -> {
+                if (logger.isTraceEnabled) {
+                    logger.trace("Received notarization response from notary service {} for transaction {}",
+                        signedTransaction.notaryName, signedTransaction.id)
+                }
+                notarizationResponse.signatures
+            }
+            else -> {
+                if (logger.isTraceEnabled) {
+                    logger.trace("Received notarization error from notary service {}. Error: {}",
```
blsemo: Should this be a higher log level so that we see errors?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,183 @@
...
+    @CordaInject
+    private lateinit var transactionSignatureService: TransactionSignatureService
+
+    @CordaInject
+    private lateinit var utxoLedgerService: UtxoLedgerService
+
+    @Suspendable
+    override fun call(session: FlowSession) {
+        try {
+            logger.info("Calling Contract Verifying Notary server...")
```
blsemo: That definitely should not be this message at this level - if we want to log at `INFO` for each notary invocation, the message should be more usefule (e.g. tx id, or counterparty?)

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,183 @@
...
+    @Suspendable
+    private fun verifyContract(
+        initialTransaction: UtxoSignedTransaction,
+        filteredTransactionsAndSignatures: List<FilteredTransactionAndSignatures>
+    ) {
+        val dependentStateAndRefs = filteredTransactionsAndSignatures.flatMap { (filteredTransaction, _) ->
+            (filteredTransaction.outputStateAndRefs as UtxoFilteredData.Audit<StateAndRef<*>>).values.values
+        }.associateBy { it.ref }
+
+        println(dependentStateAndRefs)
```
blsemo: Please delete this before merging!

blsemo (commented):
This looks mostly good, a few minor fixes are still required.
One issue I can see that will require a separte story is to make this compatible with any kind of key rotation or evolution - at the moment there is a strong assumption that notary keys are fixed forever, which is fine for a first cut, but needs a follow up.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/filtered/UtxoFilteredTransactionBuilderImpl.kt
jennyang-r3: It was a mutable list in the POC. @lankydan does it have to be immutable?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-client/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/client/ContractVerifyingNotaryClientFlowImpl.kt
jennyang-r3: What should it be? `DEBUG`?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-client/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/client/ContractVerifyingNotaryClientFlowImpl.kt
jennyang-r3: would `warn` be enough?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/filtered/UtxoFilteredTransactionBuilderImpl.kt
lankydan: heh, not sure why i would have done that. Doubt it needs that.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-client/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/client/ContractVerifyingNotaryClientFlowImpl.kt
lankydan: The existing notary client flow does:

```kotlin
if (log.isTraceEnabled) {
            log.trace("Notarizing transaction {} with notary {}", stx.id, notaryRepresentative)
        }
```

So I'd go with that same log line tbh.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-client/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/client/ContractVerifyingNotaryClientFlowImpl.kt
lankydan: In the existing notary client flow it is:

```
if (log.isTraceEnabled) {
                    log.trace("Received notarization error from notary service {}. Error: {}",
                        stx.notaryName, error)
                }
```

So either we change both or leave them both. Receiving a notary error from a notary isn't abnormal behaviour, at the same time, for a flow, it is not really expecting to receive a notary error. Based on that i'd say they should be upped to warn. Worth waiting for @nkovacsx to get back from PTO for his opinion?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
lankydan: Existing notary server flow does:

```kotlin
if (logger.isTraceEnabled) {
                logger.trace("Received notarization request for transaction {}", txDetails.id)
            }
```

I think trace is correct.

components/ledger/ledger-common-flow-api/src/main/kotlin/net/corda/ledger/common/flow/transaction/filtered/factory/ComponentGroupFilterParameters.kt
```diff
@@ -34,9 +34,33 @@ sealed interface ComponentGroupFilterParameters {
...
         override val componentGroupIndex: Int,
         val deserializedClass: Class<T>,
-        val predicate: Predicate<T>
+        val predicate: AuditProofPredicate<T>
     ) : ComponentGroupFilterParameters {
         override val merkleProofType = MerkleProofType.AUDIT
+
+        sealed interface AuditProofPredicate<T> {
+            /**
+             *  [Content] include components to [FilteredTransaction] where the components meet predicate.
```
lankydan:
```suggestion
             *  [Content] include components in a [FilteredTransaction] where the components meet predicate.
```

components/ledger/ledger-common-flow-api/src/main/kotlin/net/corda/ledger/common/flow/transaction/filtered/factory/ComponentGroupFilterParameters.kt
```diff
@@ -34,9 +34,33 @@ sealed interface ComponentGroupFilterParameters {
...
+             *  @property predicate Filtering function that is applied to each deserialized component with the group
+             */
+            class Content<T>(private val predicate: Predicate<T>) : AuditProofPredicate<T>, Predicate<T> {
+                override fun test(t: T): Boolean {
+                    return predicate.test(t)
+                }
+            }
+
+            /**
+             *  [Index] include components of indexes if the indexes exist.
```
lankydan:
```suggestion
             *  [Index] indexes of components to include in a [FilteredTransaction].
```

components/ledger/ledger-common-flow/src/test/kotlin/net/corda/ledger/common/flow/impl/transaction/filtered/FilteredTransactionImplIntegrationTest.kt
```diff
@@ -92,13 +93,28 @@ class FilteredTransactionImplIntegrationTest {
         filteredTransaction = filteredTransactionFactory.create(
             wireTransaction,
             componentGroupFilterParameters = listOf(
-                ComponentGroupFilterParameters.AuditProof(0, TransactionMetadataImpl::class.java) { true },
+                ComponentGroupFilterParameters.AuditProof(
```
lankydan, line 95: It should be possible to tidy up this formatting:

```kotlin
ComponentGroupFilterParameters.AuditProof(
    0,
    TransactionMetadataImpl::class.java,
    AuditProofPredicate.Content { true }
),
```

components/ledger/ledger-common-flow/src/test/kotlin/net/corda/ledger/common/flow/impl/transaction/filtered/factory/FilteredTransactionFactoryImplTest.kt
```diff
@@ -133,8 +134,14 @@ class FilteredTransactionFactoryImplTest {
         filteredTransaction = filteredTransactionFactory.create(
             wireTransaction,
             componentGroupFilterParameters = listOf(
-                ComponentGroupFilterParameters.AuditProof(0, TransactionMetadata::class.java) { true },
-                ComponentGroupFilterParameters.AuditProof(1, Any::class.java) { it is MyClassA || it is MyClassB },
+                ComponentGroupFilterParameters.AuditProof(
+                    0,
+                    TransactionMetadata::class.java,
+                    AuditProofPredicate.Content { true }),
+                ComponentGroupFilterParameters.AuditProof(
```
lankydan, line 142: formatting

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -109,6 +109,9 @@ class UtxoFinalityFlowV1(
                 }
             }
         }
+
+        // backchain switch will be applied in a follow up PR
```
lankydan: Either remove this or leave a TODO referencing the relevant ticket.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-api/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/api/ContractVerifyingNotarizationPayload.kt
```diff
@@ -0,0 +1,18 @@
...
+import net.corda.v5.ledger.utxo.transaction.filtered.UtxoFilteredTransaction
+
+@CordaSerializable
+data class ContractVerifyingNotarizationPayload(
+    val initialTransaction: UtxoSignedTransaction,
+    val filteredTransactionsAndSignatures: List<FilteredTransactionAndSignatures>,
+)
+
+@CordaSerializable
+data class FilteredTransactionAndSignatures(
```
lankydan: Can we put this in a separate class (this structure will probably also be needed in finality flow later on, so we might need to move it later).

Wondering if there is a better name as well (I know i named it).

lankydan (commented):
Need to continue reviewing but left some comments.

We need to test the new filtering by index that was added to the filtered transaction code.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-api/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/api/ContractVerifyingNotarizationPayload.kt
jennyang-r3: How about `NotaryFilteredData`? @blsemo do you have any suggestions?

jennyang-r3: Also, we have two data classes with this name `FilteredTransactionAndSignatures`. One is with 2 fields, `filteredTransaction` and `signatures`, the other one is with 3 fields, those two + `privacySalt`. Could we differently name them to avoid confusion?

lankydan: I think @nkovacsx determined that we didn't need the privacy salt in either of them irrc.

lankydan: Mmmm maybe `FilteredTransactionWithSignatures` is the best name, i can't think of anything off the top of my head. I just wasn't keen on the `and` in the name.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-client/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/client/ContractVerifyingNotaryClientFlowImpl.kt
```diff
@@ -0,0 +1,92 @@
...
+        val notarizationResponse = session.sendAndReceive(
+            NotarizationResponse::class.java,
+            payload
+        )
+
+        return when (val error = notarizationResponse.error) {
+            null -> {
+                if (logger.isTraceEnabled) {
+                    logger.trace("Received notarization response from notary service {} for transaction {}",
+                        signedTransaction.notaryName, signedTransaction.id)
```
lankydan: Formatting:

```kotlin
logger.trace(
    "Received notarization response from notary service {} for transaction {}",
    signedTransaction.notaryName, signedTransaction.id
)
```

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-client/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/client/ContractVerifyingNotaryClientFlowImpl.kt
```diff
@@ -0,0 +1,92 @@
...
+                    )
+                }
+                throw error
+            }
+        }
+    }
+
+    @Suspendable
+    internal fun createPayload(): ContractVerifyingNotarizationPayload {
+        val hashedNotaryKey = digestService.hash(signedTransaction.notaryKey.encoded, DigestAlgorithmName.SHA2_256)
```
lankydan, line 91: Can we use this method instead as it seems to be our "official" way of doing this (note, it does exactly what is done here under the hood).

From `PublicKeyUtils`:

```kotlin
fun PublicKey.fullIdHash(keyEncodingService: KeyEncodingService, digestService: PlatformDigestService): SecureHash =
    digestService.hash(keyEncodingService.encodeAsByteArray(this), DigestAlgorithmName.SHA2_256)
```

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,179 @@
...
+
+            // Extract the data from the signed transaction
+            val initialTransactionDetails = getInitialTransactionDetail(initialTransaction)
+
+            // Verify the signatures
+            if (logger.isTraceEnabled) {
+                logger.trace(
+                    "Verifying signatures for the following dependencies: {}",
+                    filteredTransactionsAndSignatures.map { it.filteredTransaction.id })
+            }
```
lankydan: I don't think its worth a log line here, if something goes wrong then we should see a warning from the catch block.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,179 @@
...
+            require(signatures.isNotEmpty()) { "No notary signatures were received" }
+            filteredTransaction.verify()
+            for (signature in signatures) {
+                transactionSignatureService.verifySignature(
+                    filteredTransaction.id,
+                    signature,
+                    notaryKey
+                )
+            }
+            logger.info("SUCCESSFULLY VERIFIED $filteredTransaction | $signatures")
```
lankydan: Remove this log line.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,179 @@
...
+     * A function that will verify the signatures of the given [filteredTransactionsAndSignatures].
+     */
+    @Suspendable
+    private fun verifySignatures(
+        notaryKey: PublicKey,
+        filteredTransactionsAndSignatures: List<FilteredTransactionAndSignatures>
+    ) {
+        filteredTransactionsAndSignatures.forEach { (filteredTransaction, signatures) ->
+            require(signatures.isNotEmpty()) { "No notary signatures were received" }
+            filteredTransaction.verify()
```
lankydan, line 172: Maybe we should change the name of the method since this method isn't really just verifying the signatures of the filtered transactions. Its calling `filteredTransaction.verify` as well. Probably just `verifyFilteredTransactions` is a good enough name (`verifyDependencies` might also work, but I'll leave that up to you).

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,179 @@
...
+                    signature,
+                    notaryKey
+                )
+            }
+            logger.info("SUCCESSFULLY VERIFIED $filteredTransaction | $signatures")
+        }
+    }
+
+    @Suspendable
+    private fun verifyContract(
```
lankydan: I'd name this `verifyTransaction` or `verify` instead. I'd also rename `utxoLedgerService.verifyContract(ledgerTransaction)` to `verify` since it does a bit more than just verify a transaction's contract.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,179 @@
...
+            requireNotNull(dependentStateAndRefs[stateRef]) {
+                "Missing reference state and ref from the filtered transaction"
+            }
+        }
+
+        val ledgerTransaction = if (inputStateAndRefs.isNotEmpty() || referenceStateAndRefs.isNotEmpty()) {
+            initialTransaction.toLedgerTransaction(inputStateAndRefs, referenceStateAndRefs)
+        } else {
+            initialTransaction.toLedgerTransaction()
+        }
```
lankydan: Is there a downside to just passing in empty lists into `initialTransaction.toLedgerTransaction(inputStateAndRefs, referenceStateAndRefs)`?

lankydan:
We need to remember to write unit tests for the client and server flows.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-client/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/client/ContractVerifyingNotaryClientFlowImpl.kt
jennyang-r3: It looks like this method is not accessible from notary. should I add dependency?

blsemo: Notaries are supposed to be CorDapps, so pulling in other impl projects is frowned upon.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+    cordaProvided platform("net.corda:corda-api:$cordaApiVersion")
+
+    cordaProvided 'net.corda:corda-application'
+    cordaProvided 'net.corda:corda-notary-plugin'
+    cordaProvided 'org.jetbrains.kotlin:kotlin-osgi-bundle'
+    cordaProvided 'org.slf4j:slf4j-api'
+
+    // Common package pulled in as transitive dependency through API
+    cordapp project(":notary-plugins:notary-plugin-contract-verifying:notary-plugin-contract-verifying-api")
+    cordapp project(":testing:cpbs:ledger-utxo-demo-contract")
```
blsemo: That should not be here on the final verifying notary - the contract needs to be built into the CPB for each network, and it won't be the demo contract.

blsemo (commented):
Getting there - but I think we should not merge to the release branch with the UTXO demo contract hardcoded to be includedin the contract verifying noatary.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-client/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/client/ContractVerifyingNotaryClientFlowImplTest.kt
```diff
@@ -0,0 +1,146 @@
...
+import java.security.PublicKey
+import java.time.Instant
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class ContractVerifyingNotaryClientFlowImplTest {
+
+    private companion object {
+
+        /* Signature */
+        val dummyUniquenessSignature = DigitalSignatureAndMetadata(
```
lankydan: It is better to put mocked objects outside of companion objects / variables of the class.

This is because the mocks will be reset between each test, which can be important to some tests.

For parts which are purely static data, like `val txId = SecureHashUtils.randomSecureHash()` then putting them in the companion object is ok.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/build.gradle
```diff
@@ -0,0 +1,40 @@
+plugins {
+    id 'org.jetbrains.kotlin.jvm'
+    id 'corda.common-publishing'
+    id 'net.corda.plugins.cordapp-cpb2'
```
blsemo: Shouldn't we apply `net.corda.plugins.cordapp-cpk2` here rather than `net.corda.plugins.cordapp-cpb2`? This project can never form a full valid CPB as it will always require some contract to be bundled.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,205 @@
...
+        filteredTransactionsAndSignatures.forEach { (filteredTransaction, signatures) ->
+            require(signatures.isNotEmpty()) { "No notary signatures were received" }
+            filteredTransaction.verify()
+            for (signature in signatures) {
+                transactionSignatureService.verifySignature(
+                    filteredTransaction.id,
+                    signature,
+                    notaryKey
+                )
+            }
```
blsemo: Shouldn't that be 
```suggestion
          signatures.any {
                transactionSignatureService.verifySignature(
                    filteredTransaction.id,
                    signature,
                    notaryKey
                )
            }
```
The reason being that we need to verify that the notary has signed it - however, if we receive a non-notary signature, is that really a reason to fail signature verification?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
```diff
@@ -0,0 +1,426 @@
...
+            val responseError = responseFromServer.first().error
+            assertThat(responseError).isNotNull
+            assertThat(responseError).isInstanceOf(NotaryExceptionGeneral::class.java)
+            assertThat((responseError as NotaryExceptionGeneral).errorText)
+                .contains("Error while processing request from client")
+        }
+    }
+
+    @Suppress("LongParameterList")
+    private fun createAndCallServer(
```
blsemo, line 375: This function is a bit too "magic" for my liking - it would be good if there were some explanations what the parameters do and how it works. That might also make some of the tests above more understandable...

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
```diff
@@ -0,0 +1,426 @@
...
+    @Test
+    fun `Contract verifying notary plugin server should respond with error if time window not present on filtered tx`() {
+        createAndCallServer(mockSuccessfulUniquenessClientService(), filteredTxContents = mapOf("timeWindow" to null)) {
+            assertThat(responseFromServer).hasSize(1)
+
+            val responseError = responseFromServer.first().error
+            assertThat(responseError).isNotNull
+            assertThat(responseError).isInstanceOf(NotaryExceptionGeneral::class.java)
+            assertThat((responseError as NotaryExceptionGeneral).errorText).contains(
+                "Error while processing request from client"
```
blsemo: The error returned/checked is not very specific, so how do we know that the error that has happened is what we were thinking would go wrong?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
```diff
@@ -0,0 +1,426 @@
...
+
+            val response = responseFromServer.first()
+            assertThat(response.error).isNull()
+            assertThat(response.signatures).hasSize(1)
+            assertThat(response.signatures.first().by).isEqualTo(notaryVNodeAliceKey.fullIdHash())
+        }
+    }
+
+    @Test
+    fun `Contract verifying notary plugin server should respond with error if time window not present on filtered tx`() {
```
blsemo, line 236: The test names are sufficiently clear, but I find it a bit hard to follow how this is actually tested.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
jennyang-r3: Since `transactionSignatureService.verifySignature()` doesn't return boolean and `.any {}` want us to return boolean, I tweaked the code like this. Is this acceptable?
```kotlin
signatures.any {
                try {
                    transactionSignatureService.verifySignature(
                        filteredTransaction.id,
                        it,
                        notaryKey
                    )
                    true
                } catch (e: Exception) {
                    false
                }
            }
  ```

File: notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/build.gradle [hunk 27]
```diff
@@ -0,0 +1,40 @@
...
+
+    testImplementation "org.assertj:assertj-core:$assertjVersion"
+    testImplementation "org.mockito:mockito-core:$mockitoVersion"
+    testImplementation "org.mockito.kotlin:mockito-kotlin:$mockitoKotlinVersion"
+
+    testImplementation project(":libs:crypto:crypto-core")
+    testImplementation project(':libs:serialization:serialization-amqp')
+    testImplementation project(":libs:uniqueness:common")
+    testImplementation project(":testing:crypto-testkit")
+    testImplementation project(':testing:ledger:ledger-common-testkit')
```
lankydan: Do we need all these test dependencies?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,215 @@
...
+            val (initialTransaction, filteredTransactionsAndSignatures) = session.receive(ContractVerifyingNotarizationPayload::class.java)
+            if (logger.isTraceEnabled) {
+                logger.trace("Received notarization request for transaction {}", initialTransaction.id)
+            }
+
+            // Extract the data from the signed transaction
+            val initialTransactionDetails = getInitialTransactionDetail(initialTransaction)
+
+            // validate whether notary in transaction matches the notary in network
+            validateTransactionNotaryAgainstCurrentNotary(initialTransactionDetails)
```
lankydan, line 82: I don't think we need the code comments since the method names you've given explain whats happening well enough.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,215 @@
...
+
+            // Verify the signatures
+            verifySignatures(initialTransaction.notaryKey, filteredTransactionsAndSignatures)
```
lankydan, line 84: @nkovacsx is it worth having the signature and transaction verification failures return a more specific error than `NotaryExceptionGeneral`?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,215 @@
...
+                            filteredTransaction.id,
+                            it,
+                            notaryKey
+                        )
+                        true
+                    } catch (e: Exception) {
+                        false
+                    }
+                }) {
+                throw CordaRuntimeException("Notary signature is not found with transaction: ${filteredTransaction.id}.")
```
lankydan:
```suggestion
                throw CordaRuntimeException("A valid notary signature is not found with transaction: ${filteredTransaction.id}.")
```

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,215 @@
...
+    private fun verifySignatures(
+        notaryKey: PublicKey,
+        filteredTransactionsAndSignatures: List<FilteredTransactionAndSignatures>
+    ) {
+        filteredTransactionsAndSignatures.forEach { (filteredTransaction, signatures) ->
+            require(signatures.isNotEmpty()) { "No notary signatures were received with transaction: ${filteredTransaction.id}." }
+            filteredTransaction.verify()
+
+            // throw exception if any of the signatures received is not valid notary signature
+            if (!signatures.any {
```
lankydan: `!any` -> `none`

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,215 @@
...
+                    try {
```
lankydan: I think splitting out the call of the if statement to a variable will make the code clearer:

```kotlin
val hasValidSignature = signatures.any {
    try {
        transactionSignatureService.verifySignature(
            filteredTransaction.id,
            it,
            notaryKey
        )
        true
    } catch (e: Exception) {
        false
    }
}

require(!hasValidSignature) {
    "Notary signature is not found with transaction: ${filteredTransaction.id}."
}
```

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
```diff
@@ -0,0 +1,426 @@
...
+            on { leafKeys } doReturn setOf(notaryVNodeAliceKey, notaryVNodeBobKey)
+        }
+
+        val notaryServiceName = MemberX500Name.parse("O=MyNotaryService, L=London, C=GB")
+
+        // The client that initiated the session with the notary server
+        val memberCharlieName = MemberX500Name.parse("O=MemberCharlie, L=London, C=GB")
+
+        // mock services
+        val mockTransactionSignatureService = mock<TransactionSignatureService>()
```
lankydan: don't include mocks inside the companion objects.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
```diff
@@ -0,0 +1,426 @@
...
+            assertThat(responseError).isNotNull
+            assertThat(responseFromServer.first().signatures).isEmpty()
+            assertThat(responseError).isInstanceOf(NotaryExceptionGeneral::class.java)
+            assertThat((responseError as NotaryExceptionGeneral).errorText)
+                .contains("Error while processing request from client")
+        }
+    }
+
+    @Test
+    fun `Contract verifying notary should respond with success if the notary key is simple`() {
```
lankydan: what do you mean "is simple", maybe "is valid" is better here?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
```diff
@@ -0,0 +1,426 @@
...
+            assertThat(responseError).isInstanceOf(NotaryExceptionGeneral::class.java)
+            assertThat((responseError as NotaryExceptionGeneral).errorText).contains(
+                "Error while processing request from client"
+            )
+        }
+    }
+
+    @Test
+    fun `Contract verifying notary plugin server should respond with error if input states are not audit type in the filtered tx`() {
+        val mockInputStateProof = mock<UtxoFilteredData.SizeOnly<StateRef>>()
```
lankydan: Should this be output states? There shouldn't be any input states in the filtered transactions sent to the verifying notary.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
jennyang-r3: This error is coming from this lines of code in the end of `call()` of notary server. It is copied over from nonvalidating notary. Do we need to make this more specific? @nkovacsx @lankydan 
```kotlin
       logger.warn("Error while processing request from client. Cause: $e ${e.stackTraceToString()}")

            session.send(
                NotarizationResponse(
                    emptyList(),
                    NotaryExceptionGeneral(
                        "Error while processing request from client. " +
                        "Please contract notary operator for further details."
                    )
                )
            )
```

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
lankydan: Yeah, I'd generally put the not interesting mocks in the `@BeforeEach` method.

Then in the tests mock the methods that are relevant to the current test case.

However, I'd argue that we leave this as is since @jennyang-r3 based it on the existing `NonValidatingNotaryServerFlowImplTest` which does the same thing.

Ideally we change both, but unless this can be tidied up easy I'd argue to leave it.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterBuilder.kt
```diff
@@ -298,13 +298,14 @@ class ClusterBuilder {
     private fun registerNotaryBody(
         notaryServiceName: String,
         customMetadata: Map<String, String>,
-        isBackchainRequiredNotary: Boolean = true
+        isBackchainRequiredNotary: Boolean = false,
```
lankydan: I thought we wanted the default to stay as `true` here.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
lankydan: I just left a similar question somewhere else, I'll leave this one up to @nkovacsx since he knows this area better than me.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/build.gradle
jennyang-r3: These are needed for unit tests not sure about `testImplementation project(':libs:serialization:serialization-amqp')` it might be used in e2e test but I will remove it for now and add it later if that turned out to be needed.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
blsemo: @vlajos had a very good suggestion to do something like we do in the UtxoSignedTransaction:
https://github.com/corda/corda-runtime-os/blob/b6f31f8d6f5a6f5dab797dc504e1825fc9d17b96/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt#L194

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
nkovacsx: Yeah I agree, this was mostly just a placeholder in the POC.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,205 @@
...
+    @Suspendable
+    private fun verifySignatures(
+        notaryKey: PublicKey,
+        filteredTransactionsAndSignatures: List<FilteredTransactionAndSignatures>
+    ) {
+        filteredTransactionsAndSignatures.forEach { (filteredTransaction, signatures) ->
+            require(signatures.isNotEmpty()) { "No notary signatures were received with transaction: ${filteredTransaction.id}." }
+            filteredTransaction.verify()
+            val hasValidSignature = signatures.any {
+                try {
```
blsemo: @vlajos had a very good suggestion to do something like we do in the UtxoSignedTransaction (I made this comment earlier in a response to another comment, but not sure if that got lost because the original was outdated...)

https://github.com/corda/corda-runtime-os/blob/b6f31f8d6f5a6f5dab797dc504e1825fc9d17b96/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt#L194

lankydan: Yeah, I think that is a good idea. I tried to figure out how to improve the error handling of the code but couldn't put my finger on it. That code looks like what we want.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,217 @@
...
+     */
+    @Suspendable
+    fun verifySignatures(
+        notaryKey: PublicKey,
+        filteredTransactionsAndSignatures: List<FilteredTransactionAndSignatures>
+    ) {
+        val keyIdToNotaryKeys: MutableMap<String, MutableMap<SecureHash, PublicKey>> = mutableMapOf()
+        filteredTransactionsAndSignatures.forEach {(filteredTransaction, signatures) ->
+            try {
+                filteredTransaction.verify()
```
blsemo: Coming to think of it - this should be outside of this `try...catch` block, because is this throws, because the exception thrown in the catch clause is about notary signatures - but this check is about data integrity of the filtered transaction.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -0,0 +1,217 @@
...
+                    " by this notary virtual node (${currentNotaryServiceName})"
+        }
+    }
+
+    /**
+     * A function that will verify the signatures of the given [filteredTransactionsAndSignatures].
+     */
+    @Suspendable
+    fun verifySignatures(
+        notaryKey: PublicKey,
```
blsemo, line 166: Can you please raise a ticket for the backlog to improve this logic so it also works with evolving notary keys?
If the notary key has changed between the transaction that created the inputs and the current transaction, this won't work. We need to look up the key for this notary service at the time the input was created for this check.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
```diff
@@ -0,0 +1,504 @@
...
+                    filteredTxContents["timeWindow"] as TimeWindow?
+                } else {
+                    mockTimeWindow
+                }
+            }
+
+            on { verify() } doAnswer {
+                txVerificationLogic()
+            }
+            on { id } doReturn txId
```
blsemo, line 436: I think it's not a good idea that the filtered (input) transaction has the same id as the transaction we are trying to verify.
This also highlights the fact that we do not check whether we get the right inputs to actually verify the transaction to be notarised (because that would have exploded at this point as the inputs can't have the same id as the current transaction). Can you raise a JIRA to add this check, please?

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/Utils.kt
```diff
@@ -26,6 +26,7 @@ const val CODE_SIGNER_CERT_ALIAS = "cordadev"
 // The CPB and CPI used in smoke tests
 const val TEST_NOTARY_CPI_NAME = "test-notary-server-cordapp"
 const val TEST_NOTARY_CPB_LOCATION = "/META-INF/notary-plugin-non-validating-server.cpb"
+const val TEST_VERIFYING_NOTARY_CPB_LOCATION = "/META-INF/notary-plugin-contract-verifying-server.cpb"
```
blsemo: What is this required for? I thought the verifying notary only publishes a CPK, as the CPB needs to be built with the contract?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
jennyang-r3: I put this ticket under remove requirement for backchain epic for now - https://r3-cev.atlassian.net/browse/CORE-18874

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
jennyang-r3: https://r3-cev.atlassian.net/browse/CORE-18879

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
blsemo: No, that was kind the opposite of what I meant!
Having a separate try catch for every transaction is fine, but the `filtereredTransaction.verify()` failing should be _outside_ the try/catch that throws a `NotaryExceptionInvalidSignature` exception - as that verification failing has nothing to do with signatures!

jennyang-r3: Oh, I misread your comment

blsemo (dismissed):
LGTM

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/build.gradle
```diff
@@ -0,0 +1,39 @@
+plugins {
+    id 'org.jetbrains.kotlin.jvm'
+    id 'corda.common-publishing'
+    id 'net.corda.plugins.cordapp-cpk2'
+}
+
+description 'Corda Contract Verifying Notary Plugin Server'
+
+group 'com.r3.corda.notary.plugin.contractverifying'
+
```
ronanbrowne, line 10: Just to confirm you have purposely left out 

releasable = true

Here as this is not a maven artefact and the CPB will be available on the GitHUb release page as we do with the existing non-validating server CPB today ?

ronanbrowne: To be clear my main question is, from a release perspective, who the consumers of this particular CPB are and if we are expecting a non-R3er to resolve this somehow, i think you are correct not including `releasable = true` here but I just want to have the full picture to be sure there is no outstanding tasks missed when we eventually come to release time for 5.2

ronanbrowne (dismissed):
LGTM - but see my comments

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/build.gradle
jennyang-r3: @nkovacsx do we need 
```groovy
ext {
    releasable = true
}
```
in verifying notary server?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/NotarySignatureVerificationServiceImpl.kt
```diff
@@ -0,0 +1,86 @@
...
+@Component(
+    service = [NotarySignatureVerificationService::class, NotarySignatureVerificationServiceInternal::class, UsedByFlow::class],
+    scope = PROTOTYPE
+)
+class NotarySignatureVerificationServiceImpl @Activate constructor(
+    @Reference(service = TransactionSignatureService::class)
+    private val transactionSignatureService: TransactionSignatureService
+) : NotarySignatureVerificationService, NotarySignatureVerificationServiceInternal, UsedByFlow, SingletonSerializeAsToken {
+    override fun verifyNotarySignatures(
+        transactionId: SecureHash,
```
vlajos: If only the transaction ID is passed here, the transaction metadata won't be compared to the digital signature's for batch signatures (including the Notary ones.)
There is another verifySignature overload in the TransactionSignatureVerification service, which accepts TransactionWithMetadata instead of SecureHash, which checks that part.

jennyang-r3: This API is used to verify dependencies (input + references) in filtered transactions against a new transaction's (signed transaction) notary key. That means we can't pass filtered transaction in here, since it's not `TransactionWithMetadata` and each filtered tx doesn't have signature in it. 

Having said that, it sounds like this logic won't do what we intend to do.. Do you have any other suggestions to verify this properly? @blsemo @nkovacsx wdyt?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/build.gradle
jennyang-r3: This CPK would be packed with users' contract CPK to create a CPB. In this case, will we need releasable?

ronanbrowne: Re looking at code - i see your following the same pattern as notary-plugin-non-validating so that's fine in terms of releaseabe artifacts - this doesn't need to block this PR from my PoV can have follow up conversation if needed on this

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/verifier/NotarySignatureVerificationServiceImpl.kt
vlajos: Would it cause any trouble to make `FilteredTransaction` inherit from `TransactionWithMetadata`?
(Although this may solve only a part of the problem.)

nkovacsx: `UtxoFilteredTransaction` does inherit from `TransactionWithMetadata` now (as per the API PR). This is the function we used to verify in the PoC:
```
transactionSignatureService.verifySignature(
    filteredTransaction,
    signature,
    notaryKey
)
```
I think this is the one @vlajos was talking about. So if we could pass in the `UtxoFilteredTransaction` in here instead of just the TX ID then I think it should be okay. At least that's what I did in the PoC.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
```diff
@@ -263,6 +227,18 @@ data class UtxoSignedTransactionImpl(
         return utxoLedgerTransactionFactory.create(wireTransaction)
     }
 
+    @Suspendable
```
vlajos, line 230: Does this need to be Suspendable? (Probably, it does not make harm to keep it as it, though.)

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/transaction/UtxoSignedTransactionImpl.kt
```diff
@@ -179,51 +182,12 @@ data class UtxoSignedTransactionImpl(
...
-        // transaction was signed with. This means it was signed with one of the notary VNodes (worker).
-        if (!KeyUtils.isKeyFulfilledBy(notaryKey, notaryPublicKeysWithValidSignatures)) {
-            throw TransactionSignatureException(
-                id,
-                "Notary signing keys $notaryPublicKeysWithValidSignatures did not fulfil " +
-                    "requirements of notary service key $notaryKey",
-                null
-            )
-        }
+        notarySignatureVerificationService.verifyNotarySignatures(wireTransaction, notaryKey, signatures, keyIdToNotaryKeys)
```
vlajos, line 186: The original version passed `this` instead of only the `wireTransaction` part.
TBH This should not make any significant difference.
But it would make it more consistent with the other similar functions in this class.

blsemo (dismissed):
LGTM

blsemo (approved):
Re-approved for version change.

ronanbrowne (approved):

PR #5211, opened 2023-12-05
CORE-18645 - Make Corda build system auto format capable

simon-johnson-r3:
To turn on auto formatting in any module in Corda now all you need to do is set the following property in the build.gradle file for that module:
```
ext.cordaEnableFormatting = true
```


detekt-formatting.yaml
```diff
@@ -0,0 +1,268 @@
+formatting:
```
simon-johnson-r3, line 1: This is the stock file from https://github.com/detekt/detekt/blob/v1.23.3/detekt-formatting/src/main/resources/config/config.yml with the exception of line lengths of 140 chars. If we don't set this to 140 here it clashes with the other Detekt config. We can revert this to 120 across the board more easily when auto formatting is available everywhere than we can now. If we make this 120 now, we will have to submit possibly hundreds of code changes along with this PR which is not aligned with our easy rollout per module strategy.

settings.gradle
```diff
@@ -275,7 +275,6 @@ include 'libs:lifecycle:lifecycle-impl'
 include 'libs:lifecycle:lifecycle-test-impl'
 include 'libs:lifecycle:registry'
 include 'libs:membership:certificates-datamodel'
```
simon-johnson-r3, line 276: This sub project doesn't exist which doesn't appear to matter until you try and use Detekt's formatting module on it.

detekt-formatting.yaml
driessamyn: Looking at the [documentation](https://detekt.dev/docs/introduction/configurations/):

> Note: When using a custom config file, the default values are ignored unless you also set the --build-upon-default-config flag.

This implies that if we set the `build-upon-default-config` flag it would be enough to just specify the exception.
Wouldn't that make more sense?

simon-johnson-r3: I can think of four reasons why we might not want that.

1) It suggests these defaults are from https://github.com/detekt/detekt/blob/main/detekt-core/src/main/resources/default-detekt-config.yml which is a different file to this one, these settings are not in that file. They used to be but were pulled out in this PR: https://github.com/detekt/detekt/commit/c536107febe085aba24f9ca27eb5f7a9f40937b8
So if we take the documentation literally it wouldn't work.

2) It will potentially turn things on in Detekt we excluded from the original config. That default file has four times as many lines as our own existing one. If we did want to go down this route, it would be consistent to go through that file too and ensure the same approach is applied.

3) It is more effort to maintain a file with just the "140" changes (excluding other lines) than to do a find and replace on the complete format file.

4) Locking the settings explicitly like this means we can update Detekt without worrying we get extra rules. Perhaps this is why we did this originally?

driessamyn: Understood.

But in that case, we may have to think about how we maintain this consistent across all the repos?
Maybe it could be in a different repo and included in a git submodule or something?

simon-johnson-r3: That's a fair point, but it's a problem we have today already and I don't regard the solutions as trivial - at least it will spark more discussion and I don't expect auto formatting to raise zero concerns as it is. I wonder if we want a gradle plugin which manages all this for us for example, which we simply apply to every other project?

I'd rather we raise another ticket about it anyway. But I take the point we'll only get this in corda OS right now. If you want it everywhere I think we'd have to abandon this and it will take a fair bit longer to get that?

driessamyn: Good idea about the gradle plugin. But I agree, let's merge and once this is adopted for this repo we can figure out how to roll it out to multiple.

driessamyn (approved):
LGTM

seanbrereton (approved):
LGTM (BLT)

PR #5212, opened 2023-12-06
CORE-18691: fix a bad test and clean up code a little, adding kdocs

dickon:
While studying the Merkle proof code by trying to polish it I spotted that one part of the test cases are not doing what's expected.

### DIFF
--- a/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
+++ b/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
@@ -9,6 +9,15 @@ import net.corda.v5.crypto.merkle.MerkleProofRebuildFailureException
 import net.corda.v5.crypto.merkle.MerkleProofType
 import net.corda.v5.crypto.merkle.MerkleTreeHashDigest
 
+/**
+ * Represent a merkle proof, which shows that some leaf data is in a Merkle tree.
+ *
+ * @param proofType
+ * @param treeSize - total number of leaves in the Merkle tree
+ * @param leaves - list of leaves for which we have data
+ * @param hashes - hashes of each node. Number of elements should match number of elements in leaves.
+ */
+
 class MerkleProofImpl(
     private val proofType: MerkleProofType,
     private val treeSize: Int,
--- a/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
+++ b/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
@@ -15,7 +15,11 @@ import net.corda.v5.crypto.merkle.MerkleTreeHashDigest
  * @param proofType
  * @param treeSize - total number of leaves in the Merkle tree
  * @param leaves - list of leaves for which we have data
- * @param hashes - hashes of each node. Number of elements should match number of elements in leaves.
+ * @param hashes - the hashes needed to rebuild the parts of the tree where data is not given
+ *
+ * The number of elements in hashes will depend on the tree size and where in the tree the unknown
+ * data is. There will need to be at least one, to cover the gap left by missing data. There never
+ * needs to be more than the tree size minus the number of leaves specified.
  */
 
 class MerkleProofImpl(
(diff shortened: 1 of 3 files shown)


lankydan (dismissed):

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -9,6 +9,15 @@ import net.corda.v5.crypto.merkle.MerkleProofRebuildFailureException
 import net.corda.v5.crypto.merkle.MerkleProofType
 import net.corda.v5.crypto.merkle.MerkleTreeHashDigest
 
+/**
+ * Represent a merkle proof, which shows that some leaf data is in a Merkle tree.
+ *
+ * @param proofType
+ * @param treeSize - total number of leaves in the Merkle tree
+ * @param leaves - list of leaves for which we have data
+ * @param hashes - hashes of each node. Number of elements should match number of elements in leaves.
```
vlajos: I think this is more accurate:
```suggestion
 * @param hashes - hashes of nodes required to rebuild the Merkle tree.
```
And their number can be much lower than the leaves or nodes.
For example, when the left half of a full tree is in the proof, we need only a single hash, the root of the right subtree.

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
```diff
@@ -316,107 +319,133 @@ class MerkleTreeTest {
...
+                extraIndex,
+                trivialHashDigestProvider.leafNonce(extraIndex),
+                merkleTree.leaves[extraIndex]
+            )
 
-            // We duplicate one leaf which breaks the proof.
-            val badProof6: MerkleProof =
-                MerkleProofImpl(MerkleProofType.AUDIT, proof.treeSize, proof.leaves + proof.leaves.last(), proof.hashes)
-            assertFalse(badProof6.verify(root, hashProvider))
+            // We add one leaf which breaks the proof since we will not have enough hashes now.
```
vlajos: I think we may have enough hashes, but not the ones required. Or we may have too many hashes as well.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: Reworked

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -43,6 +56,17 @@ class MerkleProofImpl(
...
 
+        // we do support and test with leaves.size == treeSize, which may not
+        // be very useful but needed not be a special case
+        if (leaves.size > treeSize) {
+            throw MerkleProofRebuildFailureException("MerkleProof has too many specified keys ${leaves.size} tree size ${treeSize}")
+        }
+        if (leaves.size < treeSize && hashes.isEmpty()) {
+            throw MerkleProofRebuildFailureException("No fill-in hashes specified to MerkleProof")
+        }
+        if (hashes.size > treeSize + leaves.size) {
```
vlajos, line 67: Would this be correct without the `+ leaves.size` ? 
```suggestion
        if (hashes.size > treeSize) {
```
I suspect the exact upper bound is somewhere around the half of the tree size.
(if you create a proof for the left(even) items on the lower level, you need the hashes of the right(odd) ones. If you add more elements to the proof, you can remove hashes. If you remove more elements from the proof, you can remove the removed element's right pair's hash since you need to add their compound hash, but essentially the number of hashes stay the same.)

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
```diff
@@ -316,107 +319,132 @@ class MerkleTreeTest {
...
+        // We remove one leaf which breaks the proof.
+        if (proof.leaves.size > 1) {
+            val badProof3: MerkleProof =
+                MerkleProofImpl(
+                    MerkleProofType.AUDIT,
+                    proof.treeSize,
+                    proof.leaves.take(proof.leaves.size - 1),
+                    proof.hashes
                 )
-                // We add one leaf which breaks the proof.
```
vlajos, line 397: I would keep this comment, just to have a comment for all case. Although it does not give too many extra details.

dickon: I put it back.

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
dickon: fixed

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: Yes, but I don't want to take the time to prove that now and this is safe.

vlajos (approved):

anien (approved):
LGTM.

PR #5213, opened 2023-12-06
CORE-18620: Support persistence of authentication protocol data

yift-r3:
This change supports persisting and restoring the link manager authentication protocol sessions.

API pull request in https://github.com/corda/corda-api/pull/1385

Multi-cluster tests in [here](https://ci02.dev.r3.com/job/Corda5/job/corda-e2e-tests-multi-cluster-tests/job/release%2F5.2/57/)

### DIFF
--- a/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
+++ b/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
@@ -39,17 +39,16 @@ import net.corda.metrics.CordaMetrics
 import net.corda.metrics.CordaMetrics.Metric.InboundSessionCount
 import net.corda.metrics.CordaMetrics.Metric.OutboundSessionCount
 import net.corda.data.p2p.app.MembershipStatusFilter
+import net.corda.data.p2p.crypto.protocol.HandshakeIdentityData
 import net.corda.membership.lib.MemberInfoExtension.Companion.isMgm
 import net.corda.membership.lib.MemberInfoExtension.Companion.sessionInitiationKeys
 import net.corda.p2p.crypto.protocol.api.AuthenticationProtocolInitiator
 import net.corda.p2p.crypto.protocol.api.AuthenticationProtocolResponder
 import net.corda.p2p.crypto.protocol.api.CertificateCheckMode
-import net.corda.p2p.crypto.protocol.api.HandshakeIdentityData
 import net.corda.p2p.crypto.protocol.api.InvalidHandshakeMessageException
 import net.corda.p2p.crypto.protocol.api.InvalidHandshakeResponderKeyHash
 import net.corda.p2p.crypto.protocol.api.InvalidPeerCertificate
-import net.corda.p2p.crypto.protocol.api.RevocationCheckMode
-import net.corda.p2p.crypto.protocol.api.Session
+import net.corda.p2p.crypto.protocol.api.SessionWrapper
 import net.corda.p2p.crypto.protocol.api.WrongPublicKeyHashException
 import net.corda.p2p.linkmanager.LinkManager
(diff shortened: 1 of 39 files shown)


libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/SessionWrapper.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import net.corda.data.p2p.crypto.protocol.AuthenticatedEncryptionSessionDetails
+import net.corda.data.p2p.crypto.protocol.AuthenticatedSessionDetails
+import net.corda.data.p2p.crypto.protocol.Session
+import net.corda.v5.base.exceptions.CordaRuntimeException
+
+/**
+ * A marker interface supposed to be implemented by the different types of sessions supported by the authentication protocol.
+ */
+sealed interface SessionWrapper {
```
dimosr, line 11: I feel it would be slightly cleaner if we retained the existing structure for all the session logic and just have an extra component (e.g. `SessionSerialiser`) with some methods that convert between the existing session types and the newly introduced Avro types.

And I guess similar with the protocol handlers, e.g. retain the handlers as they were and just introduce an extra serialiser that converts them back and forth to avro types.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -886,12 +887,12 @@ internal class SessionManagerImpl(
...
         val ourIdentityData = session.validatePeerHandshakeMessageHandleError(message, peer) ?: return null
         // Find the correct Holding Identity to use (using the public key hash).
-        val ourIdentityInfo = linkManagerHostingMap.getInfo(ourIdentityData.responderPublicKeyHash, ourIdentityData.groupId)
+        val ourIdentityInfo = linkManagerHostingMap.getInfo(ourIdentityData.responderPublicKeyHash.array(), ourIdentityData.groupId)
         if (ourIdentityInfo == null) {
             logger.ourHashNotInMembersMapWarning(
                 message::class.java.simpleName,
                 message.header.sessionId,
-                ourIdentityData.responderPublicKeyHash.toBase64()
+                ourIdentityData.responderPublicKeyHash.array().toBase64()
```
dimosr, line 895: Is there a reason we're using Avro types here? I would have thought Avro types are used only at the point of persistence / retrieval from DB, and then all the logic is executed based on the internal types.

dimosr:
I started reviewing and left a few comments, but at some point I started questioning whether there would be a much simpler and less intrusive way of doing this for now, so I didn't end up the review as it's quite a big PR.

I would have thought that this change could be introduced into the e2e protocol library in a very contained way without touching any other session / link manager code yet. Effectively, just some `toAvro` / `fromAvro` methods that convert the sessions and the protocol handlers into the Avro types (or alternatively a serialiser class that does the same although the former might be simpler as it will probably have to access internal methods). And then retain everything else the same. For the sake of testing, we can also have an extra unit test at the library level, which at every step of the protocol serialises the handler and then deserialises back into the in-memory type before continuing.

That separates concerns a bit better as well, since the user of the library can decide if they want to serialise at all (and if not, they don't have to deal with Avro types at all).

libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/SessionWrapper.kt
dimosr: Or just some `toAvro` / `fromAvro` methods, if that makes it easier to access internal fields.

yift-r3:
A less intrusive version can be found at https://github.com/corda/corda-runtime-os/pull/5227

PR #5214, opened 2023-12-06
CORE-13557 - Docs for MGM re-registration

dimosr:
Documentation for MGM re-registration

### DIFF
--- a/wiki/MGM-Re-registration.md
+++ b/wiki/MGM-Re-registration.md
@@ -1,0 +1,219 @@
+In certain cases an MGM may need to update its own member-provided context, for example after key rotation or after changes to its endpoint information. On the other hand, an MGM that previously tried to register but failed may wish to try again. MGM re-registration can be used in both these scenarios.
+
+> Note: Key rotation is not currently supported.
+
+The instructions on this page assume you have completed the [MGM Onboarding](https://github.com/corda/corda-runtime-os/wiki/MGM-Onboarding) steps.
+
+# Updating Member-provided Context
+
+Currently, updates to the member-provided context are limited to endpoint information only. Changes to other Corda platform properties are not supported at the moment.
+
+An MGM may inspect its current member-provided context either by performing a member lookup, or by looking up its latest registration request.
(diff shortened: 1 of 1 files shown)


wiki/MGM-Re-registration.md
```diff
@@ -0,0 +1,219 @@
...
+$RESGISTER_RESPONSE.registrationStatus
+```
+</details>
+
+This should return a successful response with status SUBMITTED. It will then be processed locally in the cluster, where the MGM is hosted. You can check if the request was approved by checking the status of the registration request.
+
+## Step 2 - Distribute MGM's new Member Info to members
+
+The MGM's `MemberInfo` is currently distributed via the CPI to the members of the network. 
+As a result, after an MGM is re-registered their `MemberInfo` is not automatically distributed to the various members of the network.
```
nikinagy, line 211: Can we maybe mention why is this a problem or what issues this could imply? To raise attention that you should ideally update your cpi asap and maybe prevent the "why I can't re-register as a member etc" type of questions.

dimosr: Good point. I think there are 2 aspects:
1. the fact that if the change performed is not backwards compatible, members will not be able to communicate with the MGM (e.g. for re-registration purposes) without upgrading their CPI. I will add a note for that.
2. the second aspect is only specific to the 5.0 -> 5.1 upgrade, where CPI upgrade is required regardless due to the fact member re-registration is not supported in 5.0. I had a chat about this with Nadine when she was preparing documentation for platform upgrade for the 5.1 release and she thought it would be better to avoid exposing all that complexity to customers and just tell them upgrade the CPI in the docs, which is what she ended up doing. So, given this is a re-registration document (and not platform upgrade), I think it might be better to keep this out of it.

nikinagy: Ok, let's go with the note for number 1 then.

dimosr: Added that now

nikinagy (approved):
LGTM

PR #5219, opened 2023-12-06
CORE-18725 Allow registration to move status from RECEIVED_BY_MGM to STARTED_PROCESSING_BY_MGM

charlieR3:
Resolves error: 
`Could not update status of registration 6de0f6ca-46cb-4f61-a797-755d8dcf3900 from RECEIVED_BY_MGM to STARTED_PROCESSING_BY_MGM, will ignore the update` from `net.corda.membership.impl.persistence.service.handler.BasePersistenceHandler`

`RECEIVED_BY_MGM` and `STARTED_PROCESSING_BY_MGM` had the same order meaning persistence operations moving from one to the other were failing. This didn't seem to impact registrations since the next status will eventually overwrite `RECEIVED_BY_MGM` but if a registration was unsuccessful, the last status is misleading i.e. tests would fail due to status not progressing beyond `RECEIVED_BY_MGM` but in fact they had progressed beyond than `STARTED_PROCESSING_BY_MGM`.


YashNabar (approved):
LGTM

PR #5221, opened 2023-12-07
CORE-15813: REST connection string update

josephzunigadaly:
Save request status under its own unique request ID

API PR: https://github.com/corda/corda-api/pull/1396

### DIFF
--- a/components/virtual-node/virtual-node-rest-service-impl/src/main/kotlin/net/corda/virtualnode/rest/impl/v1/VirtualNodeRestResourceImpl.kt
+++ b/components/virtual-node/virtual-node-rest-service-impl/src/main/kotlin/net/corda/virtualnode/rest/impl/v1/VirtualNodeRestResourceImpl.kt
@@ -15,6 +15,7 @@ import net.corda.data.virtualnode.VirtualNodeOperationStatusResponse
 import net.corda.data.virtualnode.VirtualNodeOperationalState
 import net.corda.data.virtualnode.VirtualNodeStateChangeRequest
 import net.corda.data.virtualnode.VirtualNodeStateChangeResponse
+import net.corda.data.virtualnode.VirtualNodeUpdateDbStatusResponse
 import net.corda.data.virtualnode.VirtualNodeUpgradeRequest
 import net.corda.libs.configuration.helper.getConfig
 import net.corda.libs.external.messaging.serialization.ExternalMessagingRouteConfigSerializerImpl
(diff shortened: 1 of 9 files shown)


libs/virtual-node/virtual-node-datamodel/src/main/kotlin/net/corda/libs/virtualnode/datamodel/repository/VirtualNodeRepository.kt
```diff
@@ -34,6 +34,12 @@ interface VirtualNodeRepository {
      */
     fun findVirtualNodeOperationByRequestId(entityManager: EntityManager, requestId: String): List<VirtualNodeOperationDto>
 
+    /**
+     * Find a virtual node operation by the given operation requestId
+     * @throws VirtualNodeOperationNotFoundException
```
vkolomeyko:
```suggestion
     * Stores virtual node operation into the DB
```

components/virtual-node/virtual-node-rest-service-impl/src/main/kotlin/net/corda/virtualnode/rest/impl/v1/VirtualNodeRestResourceImpl.kt
```diff
@@ -341,12 +342,23 @@ internal class VirtualNodeRestResourceImpl(
...
                     val x = this.operationHistory.first()
-
-                    messageConverter.convert(
-                        x,
-                        OperationTypes.UPGRADE_VIRTUAL_NODE.toString(),
-                        null
-                    )
+                    val originalRequest = Schema.parseJsonToObject(x.requestData)
+
+                    if (originalRequest is Map<*, *> && originalRequest.containsKey("vaultDmlConnection")) {
```
vkolomeyko: IMO feels a bit "weak" to check against generic type of `Map` and use `containsKey`.
Possibly, we can change `requestData` such that it will explicitly contain the `operationType` performed on which we can make a decision.

gradle.properties
```diff
@@ -46,7 +46,7 @@ commonsTextVersion = 1.10.0
 bouncycastleVersion=1.76
 # Corda API libs revision (change in 4th digit indicates a breaking change)
 # Change to 5.2.0.xx-SNAPSHOT to pick up maven local published copy
-cordaApiVersion=5.2.0.12-beta+
+cordaApiVersion=5.2.0.12-alpha-1702402832947
```
josephzunigadaly: Revert version number change
```suggestion
cordaApiVersion=5.2.0.12-beta+
```

vkolomeyko (approved):
LGTM

PR #5222, opened 2023-12-07
CORE-18709: move HTTP client to libs messaging

yift-r3:
Move the HTTP client from the Gateway to the messaging library to allow it to be used by other components.


libs/messaging/messaging-impl/src/integrationTest/kotlin/net/corda/messaging/publisher/HttpRpcClientImplTest.kt
```diff
@@ -0,0 +1,126 @@
...
+import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.mock
+import java.net.ServerSocket
+import java.net.URI
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicInteger
+
+class HttpRpcClientImplTest {
+    @Test
+    fun `client sends the correct reqeust to the server`() {
```
nikinagy:
```suggestion
    fun `client sends the correct request to the server`() {
```
typo

testing/p2p/inmemory-messaging-impl/src/main/kotlin/net/corda/messaging/emulation/publisher/HttpRpcClientIml.kt
```diff
@@ -0,0 +1,10 @@
+package net.corda.messaging.emulation.publisher
+
+import net.corda.messaging.api.publisher.HttpRpcClient
+import java.net.URI
+
+class HttpRpcClientIml : HttpRpcClient {
```
nikinagy:
```suggestion
class HttpRpcClientImpl : HttpRpcClient {
```
typo

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/publisher/factory/PublisherFactory.kt
```diff
@@ -41,4 +42,18 @@ interface PublisherFactory {
...
+     * Create an instance of the [HttpRpcClient]
+     *
+     * The HTTP RPC client will send an RPC requests over HTTP.
+     *
+     * The client is responsible for retries
+     *
+     * The responder side can be found in [SubscriptionFactory] under [createHttpRPCSubscription]
+     *
+     * @param rpcConfig configuration object used to initialize the subscription
+     * @param messagingConfig other configuration settings if needed
```
nikinagy: where are these params? are they the ones specified on the factory itself? if yes, then do we need to document this here?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/HttpRpcClientImpl.kt
```diff
@@ -12,7 +13,7 @@ import java.net.http.HttpResponse
 import java.nio.ByteBuffer
 import java.time.Duration
 
-internal class HttpRpcClient(
+class HttpRpcClientImpl(
```
LWogan, line 16: We will have a HttpRpcClient and also an RpcClient which also uses Http.

I'd expect this might get confusing. I wonder is there a way we could structure the library in a way we could reuse this. Perhaps not as the RpcClient is designed for use with the Mediator. Maybe @ben-millar or @mbrkic-r3 would have some thoughts here

yift-r3: Most of the code was copied from the `RPCClient`. The plan is to make that client use this one at some point... For now, we need a client that is more generic that that one.

LWogan: Could we raise a JIRA to track this, just so its not forgotten about

LWogan (approved):
LGTM

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/publisher/HttpRpcClientImpl.kt
yift-r3: https://r3-cev.atlassian.net/browse/CORE-18802 created

nikinagy (approved):
LGTM

PR #5223, opened 2023-12-07
CORE-18100 reduce flow cache size for combined worker tests to ensure all cache logic is exercised

LWogan:
As part of our e2e tests it is important to exercise cache eviction and deserialization logic on a range of different flows. By setting the cache size to 1 for the combined worker tests and using the default 10000 for the corda cluster tests we can test both scenarios.


mag1c-48 (approved):
LGTM

ben-millar (approved):

PR #5224, opened 2023-12-07
CORE-18106: Minor change to `MultiSourceEventMediatorImpl`

vkolomeyko:


mbrkic-r3 (approved):
LGTM

This change might be soon reverted with CORE-18373 that should restore proper error handling and split long functions into multiple functions (which were the reasons for initial scope of the variable).

PR #5226, opened 2023-12-07
CORE-18691: More comments, render Merkle trees for test cases, add calculateLeveledHash code

dickon:
Increase the detail on the comments in `MerkleProofImpl` and `MerkleTreeImpl`, to make the code easier to maintain.

Improve test coverage on Merkle trees by showing the content and root hashes, including draw out the tree structure. 
Next step is to add rendering of Merkle proofs so that I can assert the whole structure of the proof by matching one string, at which point this unicode rendering will be more useful. 

For example, we now have an assertion that our test input data, in the form of sequential 32 bit ints, hashes like this:

```
              a868a19c7 ┳┳┳ 00:00:00:00
                        ┃┃┗ 00:00:00:01
                        ┃┗┳ 00:00:00:02
                        ┃ ┗ 00:00:00:03
                        ┗┳┳ 00:00:00:04
                         ┃┗ 00:00:00:05
                         ┗┳ 00:00:00:06
                          ┗ 00:00:00:07
```

That's actually all the infromation we reveal in the Merkle Tree implementation right now. The hashes of non-root nodes is not accessible, nor can you tell the tree structure.  The unicode illustration of the structure within is actually a detail of our Merkle tree implementation - it's always a left  binary balanced. Since we don't expose the tree structure, when it comes to talking about Merkle proofs it's hard to know where the proof hashes go, unless you reimplement the proof code.

Previously we don't actually test the root hash of trees, so we could for instance accidentally change the tree balancing which would change the root hash. Since we now include the root hash and check the pretty-printed Merkle tree hashes we are improving test coverage. There's scope for more coverage and we can add that later. I've also added some  assertions I did about specific values within the trees/proofs.

Later, we could extend the text representation with the unicode to proofs, where we can show intermediate hashes filled in by the proofs.

Since `MerkleTreeImpl` and `MerkleProofImpl` do not reveal the structure of the tree, we introduce `calculateedLeveledHashes`, which takes an audit proof and works out the levels and indices within that level of each hash required to fill in gaps when verifying a proof (i.e. the proof hashes). This code has a small amount of test coverage for now, including some negative cases. We run the code over the Merkle proofs we come up with during other tests, and check a little bit of the level information. For now this duplicates code, and the motivation to improve test coverage it to make it safe to move ahead and reduce duplication once we have solid test coverage, in future work. The hope is that this extra information somehow helps serialise proofs and re-use proof elements.

Co-authored by Dan Newton, i.e. based on his POC.


libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/LeveledMerkleProof.kt
```diff
@@ -0,0 +1,12 @@
+package net.corda.crypto.merkle.impl
+
+import net.corda.v5.crypto.extensions.merkle.MerkleTreeHashDigestProvider
+import net.corda.v5.crypto.merkle.IndexedMerkleLeaf
+
+@Suppress("Unused")
+class LeveledMerkleProof(
+     private val loadedLeaves: List<IndexedMerkleLeaf>,
```
vlajos: Can we converge this class a bit to the original MerkleProof one?
For example if this field is the same as in the original (`leaves`, then I would use the same name for it.

vlajos: Also, won't this support Size Proofs?
Or what happened with the type?

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/LeveledMerkleProof.kt
```diff
@@ -0,0 +1,12 @@
...
+     private val loadedHashes: List<LeveledHash>,
+     private val treeSize: Int,
+     private val digest: MerkleTreeHashDigestProvider
```
vlajos: This is a bit suspicious as well.
Generally, it would be good to avoid service pointers. But I do not have the whole context, so this may be okay.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/CalculateLeveledHashes.kt
```diff
@@ -0,0 +1,128 @@
+package net.corda.crypto.merkle.impl
+
+import net.corda.v5.crypto.SecureHash
+import net.corda.v5.crypto.extensions.merkle.MerkleTreeHashDigestProvider
+import net.corda.v5.crypto.merkle.MerkleProof
+import net.corda.v5.crypto.merkle.MerkleProofRebuildFailureException
+
+
+/**
+ * Work out the hashes required to fulfill a proof, and returned them
```
vlajos:
```suggestion
 * Work out the hashes required to fulfill a proof, and return them
```

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/CalculateLeveledHashes.kt
```diff
@@ -0,0 +1,128 @@
...
+ * with their levels and tree indices.
+ *
+ * @param proof A valid Merkle proof
+ * @param digest An object that computes hashes for merkle tree elements
+ * @return
```
vlajos: This is missing.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/CalculateLeveledHashes.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    // ... so that's 4 variables that get updated as we work:
+    // - $hashIndex is the position we are at in the supplied proof hashes
+    // - $currentSize is the number of leafs+nodes at this level of the tree
+    // - $treeDepth is the level of the tree, counting from the root of the tree where $currentSize==1
+    // - $nodeHashes has a list of pairs of the index and hash of the node. We checked we have some content.
+
+    while (currentSize > 1) {
+        // Now walk over the hashes at this tree level, striding over 1 or 2 at a time
+        // We are at level $treeDepth from the top of the tree (counting from 1),
```
vlajos: Is not this initialized with the depth of the tree in L33?

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -80,47 +85,97 @@ class MerkleProofImpl(
         }
         var hashIndex = 0
         val sortedLeaves = leaves.sortedBy { it.index }
-        // work out nodeHashes, which is a map from leaf index to digest
-        var nodeHashes = sortedLeaves.map { Pair(it.index, digest.leafHash(it.index, it.nonce, it.leafData)) }
-        var treeDepth = MerkleTreeImpl.treeDepth(treeSize)
-        var currentSize = treeSize
+        // work out nodeHashes, which is a list of node information for the current level we operate at
```
vlajos, line 88:
```suggestion
        // work out nodeHashes, which is a map of index to hash for the current level we operate at
```
Or something like this?

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -80,47 +85,97 @@ class MerkleProofImpl(
...
             if (nodeHashes.isEmpty()) {
                 throw MerkleProofRebuildFailureException(
                     "MerkleProof does not have enough nodeHashes to calculate root hash."
                 )
             }
             --treeDepth
-            val newItems = mutableListOf<Pair<Int, SecureHash>>()
+            // We could check here that size of nodeHashes is as expected for treeDepth; there should be a closed form.
+
+            // ... so that's 3 variables that get updated as we work:
```
```suggestion
            // ... so these variables get updated as we work:
```

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -80,47 +85,97 @@ class MerkleProofImpl(
...
+            // $nodeHashes has a list of pairs of the index and hash of the node. We checked we have some content.
+
+            val newItems = mutableListOf<Pair<Int, SecureHash>>()   // this will become nodeHashes at the end of this
+            // out iteration
+
+            // Now walk over the hashes at this tree level, striding over 1 or 2 at a time
             var index = 0
             while (index < nodeHashes.size) {
                 val item = nodeHashes[index]
+                // Now walk over the hashes at this tree level, striding over 1 or 2 at a time
```
vlajos: This is the same as L119

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -80,47 +85,97 @@ class MerkleProofImpl(
...
             }
             --treeDepth
-            val newItems = mutableListOf<Pair<Int, SecureHash>>()
+            // We could check here that size of nodeHashes is as expected for treeDepth; there should be a closed form.
+
+            // ... so that's 3 variables that get updated as we work:
+            // - $hashIndex is the position we are at in the supplied proof hashes
+            // $currentSize is the number of leafs+nodes at this level of the tree
+            // $treeDepth is the level of the tree, counting from the root of the tree where $currentSize==1
+            // $nodeHashes has a list of pairs of the index and hash of the node. We checked we have some content.
```
vlajos: This section looks quite similar to https://github.com/corda/corda-runtime-os/pull/5226/files#diff-0258ade85806065faceca39e37aee8072d6efd33f5ea224aad466736fc175f35R36-R40
But not exactly the same.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -80,47 +85,97 @@ class MerkleProofImpl(
...
+
+            val newItems = mutableListOf<Pair<Int, SecureHash>>()   // this will become nodeHashes at the end of this
+            // out iteration
+
+            // Now walk over the hashes at this tree level, striding over 1 or 2 at a time
             var index = 0
             while (index < nodeHashes.size) {
                 val item = nodeHashes[index]
+                // Now walk over the hashes at this tree level, striding over 1 or 2 at a time
+                // We are at level $treeDepth from the top of the tree (counting from 1),
```
vlajos: I think this counts from the other way towards 1. (Or I misunderstand something...)

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -80,47 +85,97 @@ class MerkleProofImpl(
...
                             )
-                            index += 2
-                            continue
+                            // and record that we consumed two values from our working set, and skip on to the
+                            // start of the next loop
+                            index += 2                              // we've consumed two
+                            continue                                // continue the inner level scanning loop
                         }
                     }
+                    // We skip the rest of this section if we chose to make a new node by combining two known hashes.
```
vlajos: I think I would move this comment next to the continue, which triggers it.

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
```diff
@@ -503,4 +574,90 @@ class MerkleTreeTest {
...
+                    grid[x to range.first] = when (curtop) {
+                        '━' -> '┳'
+                        else -> '┃'
+                    }
+                    for (y in range.first + 1 until extent) {
+                        grid[x to y] = '┃'
+                    }
+                    val curbot = grid.getOrDefault(x to range.second, ' ')
+                    grid[x to extent] = when (curbot) {
+                        '━' -> '┻'
```
vlajos: Where is this branch used?

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/LeveledMerkleProof.kt
dickon: I think we can't change the args to `MerkleProofImpl` since it get kyro serialised. So we have to add new classes for new proof types.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/CalculateLeveledHashes.kt
dickon: Ah, I see how that text is confusing now.

dickon: reworded.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: rephrased

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: That's intentional, we would like to eliminate duplication later, since there's actually at least 3 places with the same pattern. Thats in fact why I'm extending the test coverage.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: rephrased

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: clarified

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
dickon: It renders the bottom of the vertical lines.

vlajos: I had the impression that all the tree's roots are on the top left corner and only `T` is used to extend between levels.
And I have not seen any examples of it.
Anyway, just curiosity.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/LeveledMerkleProof.kt
dickon: Deleted; it makes sense to do this in a later PR.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: Actually it is a list. Amusingly, I made the same mistake and thought it was a map when first reding the code, since the initialiser for it includes the word `map` prominently which is why I added the type annotation on L89.

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
dickon: Actually you are right. I wrote it to handle a case that can't happen, and that's confusing (and would be easy to add back) so I simplified.

Nice spot.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: removed

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
vlajos: You are right. Sorry for the noise... :)

vlajos (dismissed):
LGTM

vlajos (approved):

nkovacsx (approved):
LGTM

anien (approved):
Approving on a fact that @vlajos and @nkovacsx have approved as this is their area of expertise and not really crypto, even though it resides in crypto modules.

PR #5227, opened 2023-12-07
CORE-18620: Add to/from Avro methods to p2p sessions

yift-r3:
This is a less intrusive version of https://github.com/corda/corda-runtime-os/pull/5213. It adds a `toAvro` and `toCorda` methods to the protocol sessions and objects to allow us to persist and restore them from persistence.

The API pull request is in https://github.com/corda/corda-api/pull/1385


dimosr:
I'll try and take a deeper look at some point, I only did a quick scan of the high-level approach. Can you add an extra test as described [here](https://github.com/corda/corda-runtime-os/pull/5213#issuecomment-1845581799), similar to the happy path tests of `AuthenticatedEncryptionSessionTest` and `AuthenticatedSessionTest` where at each step of the session negotiation and after the session negotiation, the handlers and the sessions are converted into Avro types and then from Avro types back to the handler/session types before proceeding with the next step? That would help confirm that serialisation/deserialisation occurs with all the necessary data to hydrate them properly, thus not breaking the overall functionality of the protocol.

libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocol.kt
```diff
@@ -46,16 +46,20 @@ import net.corda.data.p2p.gateway.certificates.RevocationCheckRequest
...
+typealias CheckRevocation = (RevocationCheckRequest) -> RevocationCheckResponse
 /**
  * A base, abstract class containing the core utilities for the session authentication protocol.
  * [AuthenticationProtocolInitiator] implements the APIs for the initiator side.
  * [AuthenticationProtocolResponder] implements the APIs for the responder side.
  *
  * For the detailed spec of the authentication protocol, refer to the corresponding design document.
  */
-abstract class AuthenticationProtocol(private val certificateValidatorFactory: (revocationCheckMode: RevocationCheckMode,
+sealed class AuthenticationProtocol(private val certificateValidatorFactory: (revocationCheckMode: RevocationCheckMode,
```
dimosr: worth introducing an extra typealias here? to make the parameter a bit more readable e.g.
```
typealias CertificateValidatorFactory = (revocationCheckMode: RevocationCheckMode,
                                         pemTrustStore: List<PemCertificate>,
                                         checkRevocation: CheckRevocation) -> CertificateValidator
```

libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocolInitiator.kt
```diff
@@ -60,31 +65,38 @@ class AuthenticationProtocolInitiator(
     private val ourPublicKey: PublicKey,
     private val groupId: String,
     private val certificateCheckMode: CertificateCheckMode,
+    private var step: Step = Step.INIT,
+    private var initiatorHandshakeMessage: InitiatorHandshakeMessage? = null,
+    private var session: Session? = null,
```
dimosr: were these brought in here purely for the sake of calling this constructor from the `fromAvro` method, or were there other reasons?

Because if not, I am wondering whether we can do it in a way that doesn't expand the default constructor of the API (to avoid confusing regular users of the API that wouldn't need to know about these). Either by keeping them as private fields that are then manually set from within `fromAvro` or by having an extra constructor that's accessible only internally.

libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocolResponder.kt
```diff
@@ -56,35 +62,40 @@ import kotlin.math.min
  *
  * This class is not thread-safe, which means clients that want to use it from different threads need to perform external synchronisation.
  */
+@Suppress("LongParameterList")
 class AuthenticationProtocolResponder(
     val sessionId: String,
     private val ourMaxMessageSize: Int,
+    private var step: Step = Step.INIT,
```
dimosr: a similar comment/question to the initiator here on whether we can keep internal stuff from leaking through the default constructor.

libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/CertificateCheckMode.kt
```diff
@@ -1,17 +1,37 @@
...
-import net.corda.data.p2p.gateway.certificates.RevocationCheckResponse
+import net.corda.data.p2p.gateway.certificates.RevocationMode
+import net.corda.p2p.crypto.protocol.api.RevocationCheckMode.Companion.fromAvro
+import net.corda.data.p2p.crypto.protocol.CheckCertificate as AvroCheckCertificate
 
 /**
  * How should the authentication protocol check the certificates sent as a part of authentication protocol.
  */
 sealed class CertificateCheckMode {
+    abstract fun toAvro() : AvroCheckCertificate?
```
dimosr, line 11: Is this Avro type the one that's already present and used for sending certificate requests down to the gateway? Wondering whether we should introduce a new Avro type for this to avoid coupling those together, e.g. in case we want to evolve them separately. For instance, the idea for the gateway endpoint was that other processors could use it to perform certificate checks, so it might end up having usages that are not relevant to the p2p protocol.

libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/Session.kt
```diff
@@ -1,8 +1,56 @@
...
+            SecretKeySpec(
+                this.key.array(),
+                this.algorithm,
+            )
+        internal fun SecretKey.toAvro() =
+            AvroSecretKeySpec(
+                this.algorithm,
+                ByteBuffer.wrap(this.encoded),
+            )
+        fun AvroSession.fromAvro(): Session {
```
dimosr: quite minor, but the conventional way I've seen with this is instead of an extension function, you'd have smth like:
```
fun fromAvro(avroSession: AvroSession): Session
```
it's a bit confusing having an extension method on a type `T` called `fromT`.

dimosr (commented):
A few more comments after a more detailed look.

libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/Session.kt
yift-r3: I like the ability to use `t?.fromAvro()` instead of `t?.let { fromAvro(it) }` for nullables.

Maybe instead of `fromAvro` we should use `toCorda`?

dimosr: yes, I think doing `toCorda` would not be as confusing.

libs/p2p-crypto/src/integrationTest/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticatedEncryptionPersistenceTest.kt
dimosr, line 1: The 2 newly added tests might be better placed under the `test` folder, rather than under `integrationTest`?

If I understand correctly, they are not testing integration with anything else - but purely the functionality of the library on its own.

libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocol.kt
```diff
@@ -232,6 +266,58 @@ sealed class AuthenticationProtocol(private val certificateValidatorFactory: (re
...
+        responderHandshakePayloadBytes = header.responderHandshakePayloadBytes?.array()
+        agreedMaxMessageSize = header.agreedMaxMessageSize
+    }
+
+    protected fun toAvro(
+        sessionId: String,
+        ourMaxMessageSize: Int,
+        session: Session?,
+    ) =
+        AuthenticationProtocolHeader(
```
dimosr: how did you end up with the `header` suffix for this? We typically use it for message headers, so it could be a bit confusing.

Would something like `AuthenticationProtocolBaseDetails` or smth be a bit easier to understand?

libs/utilities/src/test/kotlin/net/corda/utilities/crypto/PublicKeyFactoryTest.kt
```diff
@@ -23,6 +23,37 @@ HQ8EBAMCAa4wCgYIKoZIzj0EAwIDRwAwRAIgaxHWXC7NXtd4XHPRgLqV+UcNS7NW
 ORC/w+12HlGG968CICBvpZAN2HHIlo2Vmgak+avL2zdIK6LQo0nXuY+4e0KT
 -----END CERTIFICATE-----
         """
+
+        const val PRIVATE_KEY_PEM = """
+-----BEGIN PRIVATE KEY-----
```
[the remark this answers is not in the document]
dimosr: @yift-r3 can you check with someone from the infra team whether that would be an issue for a release and whether there's something we need to do to add this on an allow list?

libs/p2p-crypto/src/integrationTest/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticatedEncryptionPersistenceTest.kt
yift-r3: They are not really unit tests; they have multiple points of failure, use random keys, etc.  I think it's closer to integration tests than unit tests (I think of it is integration with Java security and BC).

libs/p2p-crypto/src/main/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocol.kt
yift-r3: Renamed

libs/p2p-crypto/src/integrationTest/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticatedEncryptionPersistenceTest.kt
dimosr: ok

dimosr (dismissed):

dimosr (approved):

PR #5231, opened 2023-12-08
CORE-18798 Only pass in backchain property if >5.1 and add upgrade logic

nkovacsx:
- We should only pass in `corda.notary.service.backchain.required` when registering a notary if the version is >= 5.2.
- We should be able to update this flag during re-registration, added that logic too.
- I also added a logic when upgrading from 5.0/5.1 to 5.2+. In the "old" context we won't have the `corda.notary.service.backchain.required` property and we should only upgrade a transition of `null`->`true` because pre-5.2 notaries will always be backchain required.

Added tests to verify.


blsemo (dismissed):
LGTM

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MembershipUtils.kt
```diff
@@ -170,12 +171,16 @@ fun ClusterInfo.onboardNotaryMember(
...
-            "corda.notary.service.backchain.required" to "$isBackchainRequired",
             "corda.notary.service.flow.protocol.name" to "com.r3.corda.notary.plugin.nonvalidating",
             "corda.notary.service.flow.protocol.version.0" to "1",
             "corda.notary.keys.0.id" to notaryKeyId,
             "corda.notary.keys.0.signature.spec" to DEFAULT_SIGNATURE_SPEC
-        ) + (getAdditionalContext?.let { it(holdingId) } ?: emptyMap())
+        ) + (getAdditionalContext?.let { it(holdingId) } ?: emptyMap()) + (
+                // Add the optional backchain property if version is >= 5.2
+                if (restApiVersion != RestApiVersion.C5_0 && restApiVersion != RestApiVersion.C5_1)
+                    mapOf("corda.notary.service.backchain.required" to "$isBackchainRequired")
```
lankydan, line 182: @dimosr @charlieR3 can you guys weigh in here? Seems like errors like the one @nkovacsx has just fixed are bound to keep happening to every property added when we run the upgrade tests.

dimosr: That sounds expected to me, given we are trying to preserve backwards compatibility on our REST endpoint while also having a utility method that is supposed to be used globally across our tests. Seems like tests did their job well :) 

If there's any suggestion on what we can do to catch these things earlier, happy to hear. I don't have anything on the top of my head.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
```diff
@@ -471,11 +472,21 @@ class DynamicMemberRegistrationService @Activate constructor(
                     tlsSubject
 
             previousRegistrationContext?.let { previous ->
+                // This property can only be null when upgrading from 5.0/5.1, and we should move it to `true`
```
nikinagy: maybe we could move this block to a separate function?

components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationServiceTest.kt
```diff
@@ -1457,6 +1472,93 @@ class DynamicMemberRegistrationServiceTest {
...
+            val registrationException = assertThrows<InvalidMembershipRegistrationException> {
+                registrationService.register(registrationResultId, member, newContext.toMap())
+            }
+
+            assertThat(registrationException)
+                .hasStackTraceContaining("Optional back-chain flag can only move to 'true' during platform upgrade.")
+        }
+
+        @Test
+        fun `re-registration allows optional backchain flag to be set to false from true`() {
```
nikinagy: I am bit confused as to how we can move from try to false etc. You mentioned that "because pre-5.2 notaries do not support optional backchain" -> but with the current changes: I can move to true on cpb upgrade and then move to false?

nkovacsx: That's a good point, I just raised this with the group to see what they think.

nkovacsx: Talked to @blsemo to verify, we shouldn't move from anything other than `none` -> `true` during re-registration.

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
nkovacsx: Done

components/membership/registration-impl/src/main/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationService.kt
```diff
@@ -812,4 +814,15 @@ class DynamicMemberRegistrationService @Activate constructor(
             )
         }
     }
+
+    private fun verifyBackchainFlagMovement(previousContext: Map<String, String>, newContext: Map<String, String>) {
+        // This property can only be null when upgrading from 5.0/5.1, and we should move it to `true`
+        // because pre-5.2 notaries do not support optional backchain
```
nikinagy, line 820: just one more nit: can we mention that we also don't allow true to false and false to true updates during re-registrations?

nikinagy (dismissed):
LGTM

blsemo (dismissed):

blsemo (approved):

yift-r3 (approved):

PR #5237, opened 2023-12-11
CORE-17596 Updating combined worker docs to include new startup params

ben-millar:
Now that worker communication is done via synchronous HTTP/RPC, service endpoints must be provided to the Corda cluster on startup. In a standard deployment, these are generated programatically through the helm chart, but for combined worker deployments they must be specified manually as the internal loopback address (localhost) as we're running on a single process.

This PR modifies `CombinedWorker.kt` so that it defaults to the correct values, meaning they no longer have to be supplied as command line arguments when starting it.


mbrkic-r3:
Wold it make sense for `CombinedWorker` to have a default value for this (I can't think of scenario when user would like to provide a different value), so it's not mandatory to set it?

ronanbrowne (approved):
LGTM

ronanbrowne:
can you update [this ](https://github.com/corda/corda-shared-build-pipeline-steps/blob/5.2/vars/combinedWorkerPipeline.groovy#L137)also please

mbrkic-r3 (approved):
LGTM

PR #5238, opened 2023-12-11
CORE-18634 Add HSM category for symmetric encryption

YashNabar:
Adds a new HSM category `ENCRYPTION_SECRET` whose associated `ManagedKey` will be used for encryption operations.

This category is assigned to cluster-level tenant 'p2p' when `CryptoProcessor` starts:

Screenshot 2023-12-11 at 15 10 08

### DIFF
--- a/processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
+++ b/processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
@@ -22,6 +22,7 @@ import net.corda.crypto.config.impl.SALT
 import net.corda.crypto.config.impl.WRAPPING_KEYS
 import net.corda.crypto.config.impl.retrying
 import net.corda.crypto.core.CryptoConsts
+import net.corda.crypto.core.CryptoConsts.Categories.ENCRYPTION_SECRET
 import net.corda.crypto.core.CryptoService
 import net.corda.crypto.core.CryptoTenants
 import net.corda.crypto.core.SigningKeyInfo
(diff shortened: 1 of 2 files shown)


dickon (approved):
LGTM

processors/crypto-processor/src/main/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImpl.kt
```diff
@@ -205,12 +206,14 @@ class CryptoProcessorImpl @Activate constructor(
...
                 tenantInfoService = startTenantInfoService()
                 cryptoService = startCryptoService(event.config.getConfig(CRYPTO_CONFIG), tenantInfoService)
-                CryptoConsts.Categories.all.forEach { category ->
+                (CryptoConsts.Categories.all - ENCRYPTION_SECRET).forEach { category ->
                     CryptoTenants.allClusterTenants.forEach { tenantId ->
                         tenantInfoService.populate(tenantId, category, cryptoService)
                         logger.trace("Assigned SOFT HSM for $tenantId:$category")
                     }
                 }
+                tenantInfoService.populate(CryptoTenants.P2P, ENCRYPTION_SECRET, cryptoService)
```
yift-r3, line 215: What is this line doing?

YashNabar: That is doing the tenant-category assignment. It's separate from the above block because we want this to be assigned to 'p2p' only

yift-r3 (approved):
LGTM (Seems reasonable, assuming that the crypto is OK).

PR #5239, opened 2023-12-11
CORE-16983 Interoperability / Merkle Proof Serializers

szymonsztuka:
Interoperability involves sending a notary signature between two Cordapps running under various Application Networks. 
The predominantly application protocol is JSON. Hence some  crypto primitives like a merkle proof require to have JSON serializers and deserializers.  This PR add several of them to the existing `JsonMarshallingServiceImpl` service. A cordapp can now retrieve these object from arguments of RPC started flow, and as well using the service to create a JSON response with these types.

A preliminary work to enable unit testing the new serializers is in these commits  debe9d55fe88e8bd61c73b8a8d01d16bffbdff6e, f4385393e693b6753f5f8e71b1e1cd7c5a01dee9 - UTXO ledger based test can now use signing service instead of a mock, so the Transaction Signature has a non null MerkeTree proof.

`DigitalSignatureAndMetadataSerialisationModule` adds JSON serializes and JSON deserializers for `MerkleProof` and other related classes. The new module is added to `JsonMarshallingServiceImpl`. The complication was for two new deserializers -  a need to use constructors of the types hidded in internal crypto package, e.g. `MerkleProofImpl`. This was addressed by providing another new service `MerkleProofProvider` (see desing diagram DA5 - Design Doc covering all new and modified classes).

Unit testing - the test requires a transaction signature with Merkle Proof. Such signature can be produced by the base class `UtxoLedgerTest`. The base is available for use in `components/ledger/ledger-utxo-flow` module where the test is placed. The test requires a service implementation of `MerkleProofProvider` which is injected into the SUT, the implementation is inside `libs/crypto/merkle-impl` hence with Gradle dependency `testImplementation project(':libs:crypto:merkle-impl')`. The actual SUT (System Under Test) class `DigitalSignatureAndMetadataSerialisationModule` is from `libs/application/application-impl`, so the test is not placed in the same module as the class. There are two reasons: The first  one: `application-impl` module has no test dependencies on either `testing/ledger/ledger-utxo-base-test` nor `libs/crypto/merkle-impl`, while adding it to `components/ledger/ledger-utxo-flow` requires addition of the latter dependency only.
The second reason is that if the test is run from withing a module other than `components/ledger`, then it throws an exception that _Sandbox cannot be mocked_. Placing the test inside `components/ledger/ledger-utxo-flow` seemed like a reasonable compromise.

Design document https://github.com/corda/platform-eng-design/pull/610#pullrequestreview-1775641210 - see _Solution Design: Serialize/Deserialize Signed Transaction Data_


libs/application/application-impl/src/main/kotlin/net/corda/application/impl/services/json/DigitalSignatureAndMetadataSerializers.kt
```diff
@@ -0,0 +1,218 @@
...
+    ): IndexedMerkleLeaf {
+        val node: JsonNode = parser.codec.readTree(parser)
+        val index = node.get("index").asInt()
+        val nonce = if (node.has("nonce")) node.get("nonce").binaryValue() else null
+        val leafData = node.get("leafData").binaryValue()
+        return merkleProofProvider.createIndexedMerkleLeaf(index, nonce, leafData)
+    }
+}
+
+class MerkleProofSerializer : JsonSerializer<MerkleProof>() {
```
blsemo, line 69: Could we break out the actual Merkle proof serializiation to a separate module (might be intersesting to e.g. JSON serialize filtered transactions later)?
Also, I think all these serializers/deserializers should have their own round trip tests (i.e. round trip a merkle proof and check that it still validates).

blsemo (commented):
Looks fine to me in general but:
- I would love to see the Merkle proof serializers as first class citizens
- More test coverage that the actual serializers/deserializers round trip properly on all levels, with different merkle proof types and edge cases.

libs/application/application-impl/src/main/kotlin/net/corda/application/impl/services/json/DigitalSignatureAndMetadataSerializers.kt
jmacmahonr3: This feel like nice to haves. Im not sure if we'll have the capacity on the team to bring this further than it is right now other than addressing major issues

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/serialization/DigitalSignatureAndMetadataSerializationTests.kt
```diff
@@ -0,0 +1,151 @@
...
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.Assertions.assertNotNull
+import org.junit.jupiter.api.Assertions.assertThrowsExactly
+import org.junit.jupiter.api.Test
+import java.util.*
+import kotlin.test.assertContains
+
+
+class DigitalSignatureAndMetadataSerializationTests : UtxoLedgerTest() {
```
blsemo: It would be good if this had the JSON serialization somewhere in the name or the pacakge name.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/serialization/DigitalSignatureAndMetadataSerializationTests.kt
blsemo, line 1: Shouldn't this test be along the other serialization tests in `libs/ledger`? Or is there a reason why it has to be in components? The serializer under test is defined under libs after all...

blsemo (commented):
Generally fine by me - can you please looking into moving the test into the same area as the implementation, if possible (see comments).
Also, probably someone from @vkolomeyko 's team should double check that the changes to the JsonMarshallingService don't cause unexpected issues.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/serialization/DigitalSignatureAndMetadataSerializationTests.kt
szymonsztuka: This is  explained in the PR description.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/serialization/DigitalSignatureAndMetadataSerializationTests.kt
szymonsztuka: fixed in https://github.com/corda/corda-runtime-os/pull/5239/commits/509b7db3ace1adf21c8b5f36a9a4c1585bae5e5b

vkolomeyko:
> Generally fine by me - can you please looking into moving the test into the same area as the implementation, if possible (see comments). Also, probably someone from @vkolomeyko 's team should double check that the changes to the JsonMarshallingService don't cause unexpected issues.

I do not think my team "owns" `JsonMarshallingService` these days, but I had a quick look and if we consider `MerkleProof` as a general purpose data construct, I guess it will not harm to have JSON marshalling for it.

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/serialization/DigitalSignatureAndMetadataSerializationTests.kt
blsemo: Hmm, that sounds like a significant code smell - if you can't have the test next to the system under test because it requires all those dependencies, then it is _not_ a  unit test, but an integration test, and placing something called a unit test in this rather surprising location without explanation in the code is setting us up for a lot of headscratching later.
Can we at least have:
- a comment on the test explaining where the tested code is and why the test is in a different place (and ideally have the test as an integration test rather than a unit test)?
- Raise a JIRA to rectify or at least log this issue?

szymonsztuka: Added the explanation https://github.com/corda/corda-runtime-os/pull/5239/commits/aa35845441bcf1c7d7bbbf598445019c75fbb81e and raised the ticket https://r3-cev.atlassian.net/browse/CORE-19023.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofProviderImpl.kt
```diff
@@ -0,0 +1,23 @@
...
+import net.corda.v5.crypto.SecureHash
+import net.corda.v5.crypto.merkle.IndexedMerkleLeaf
+import net.corda.v5.crypto.merkle.MerkleProof
+import net.corda.v5.crypto.merkle.MerkleProofType
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.ServiceScope
+
+@Component(
+    service = [ MerkleProofProvider::class],
+    scope = ServiceScope.PROTOTYPE
```
vlajos, line 13: Hmm. I have just noticed. Does this need to be prototype-scoped?

libs/application/application-impl/src/main/kotlin/net/corda/application/impl/services/json/DigitalSignatureAndMetadataSerializers.kt
```diff
@@ -0,0 +1,218 @@
...
+    override fun serialize(
+        merkleProof: MerkleProof,
+        generator: JsonGenerator,
+        provider: SerializerProvider
+    ) {
+        generator.writeStartObject()
```
lankydan, line 75: Would it be easier to instead create a POJO of `MerkleProof` that is internal to this file and serialize and deserialize that instead? That _should_ require less lines of code and saves the more complicated job of creating the json yourself.

You can then read the fields off the pojo when deserializing and pass them into the constructor of the correct class or into a factory class.

lankydan: After looking through this class, I think the same can be said for most of the other serializers and deserializers, that having a Pojo class here would be the simplest option.

The easiest thing would to have crytpo have a jackson dependency but im not sure we'd want to go down that route.

components/ledger/ledger-common-flow/src/test/kotlin/net/corda/ledger/common/flow/impl/transaction/filtered/FilteredTransactionImplIntegrationTest.kt
```diff
@@ -45,7 +46,7 @@ class FilteredTransactionImplIntegrationTest {
 
     private val digestService =
         DigestServiceImpl(PlatformDigestServiceImpl(CipherSchemeMetadataImpl()), null)
-    private val jsonMarshallingService = JsonMarshallingServiceImpl()
+    private val jsonMarshallingService = JsonMarshallingServiceImpl(mock<MerkleProofProvider> {})
```
lankydan: why does this need to be `mock<MerkleProofProvider> {}` and not just `mock()`?

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofProviderImpl.kt
```diff
@@ -0,0 +1,23 @@
...
+)
+class MerkleProofProviderImpl : MerkleProofProvider {
+    override fun createMerkleProof(proofType: MerkleProofType, treeSize: Int, leaves: List<IndexedMerkleLeaf>,
```
lankydan: Can we reformat this method.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofProviderImpl.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import net.corda.crypto.cipher.suite.merkle.MerkleProofProvider
+import net.corda.v5.crypto.SecureHash
+import net.corda.v5.crypto.merkle.IndexedMerkleLeaf
+import net.corda.v5.crypto.merkle.MerkleProof
+import net.corda.v5.crypto.merkle.MerkleProofType
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.ServiceScope
+
+@Component(
```
lankydan, line 11: Where does this service need to be used and what sandbox types should it exist in?

I don't think this needs to be a prototype service as it doesn't depend on any other prototype services itself. If it should only exist in some sandbox types then prototype is fine but we need to define the sandbox type it should be available in.

lankydan: Has this been tested e2e as this will highlight any issues of incorrect sandbox usage.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofProviderImpl.kt
szymonsztuka: I followed the scope of  `JsonMarshallingServiceImpl`  where the service is referenced.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofProviderImpl.kt
szymonsztuka: This service is only used by `JsonMarshallingServiceImpl`. Could you advise if some values should be there?

libs/application/application-impl/src/main/kotlin/net/corda/application/impl/services/json/DigitalSignatureAndMetadataSerializers.kt
szymonsztuka: I haven't fought about this indeed, I followed the existing convention.

components/ledger/ledger-common-flow/src/test/kotlin/net/corda/ledger/common/flow/impl/transaction/filtered/FilteredTransactionImplIntegrationTest.kt
szymonsztuka: Fixed (removed in https://github.com/corda/corda-runtime-os/pull/5239/commits/793e4c9375ff0868bd30f58d786ba5783e705d03).

szymonsztuka: Fixed (removed) in https://github.com/corda/corda-runtime-os/pull/5239/commits/793e4c9375ff0868bd30f58d786ba5783e705d03.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofProviderImpl.kt
szymonsztuka: Fixed (reformatted the whole file) in https://github.com/corda/corda-runtime-os/pull/5239/commits/793e4c9375ff0868bd30f58d786ba5783e705d03.

lankydan (dismissed):

vlajos (dismissed):

ronanbrowne (dismissed):
LGTM (gradle file change)

components/flow/flow-service/build.gradle
```diff
@@ -72,6 +72,7 @@ dependencies {
     testImplementation project(':libs:lifecycle:lifecycle-test-impl')
     testImplementation project(':libs:lifecycle:lifecycle-impl')
     testImplementation project(':libs:lifecycle:registry')
+    testImplementation project(':libs:crypto:merkle-impl')
```
LWogan: why do you need this dependency if there's no other changes to this module?

szymonsztuka: Some of flow-service tests require `integrationTestRuntimeOnly project(':libs:application:application-impl')`, I guess the do use `JsonMarshallingServiceImpl`. With this PR `JsonMarshallingServiceImpl` depends on a new service `MerkleProofProvider` which is provided by `:libs:crypto:merkle-impl`.  I will change the type of the new dependency from `testImplementation` to `integrationTestRuntimeOnly`.

LWogan (approved):
LGTM

vlajos (approved):

lankydan (approved):

ronanbrowne (approved):

PR #5241, opened 2023-12-11
CORE-18501: Check flowStartOperationalStatus after flow checkpoint is initialized 

Tom-Fitzpatrick:
This is intended to fix the underlying issue causing the e2e test "maintenance mode stops flows" to be flakey. https://r3-cev.atlassian.net/browse/CORE-18501

When this fails, it does so with the error: `Caused by: org.apache.avro.AvroTypeException: value null is not a FlowInitiatorType at FlowStatus.initiatorType`. This prevents the FlowStatus from being published to Kafka and so the flow never gets a KILLED status and is instead stuck in START_REQUESTED.

To fix this, this moves checking the flowStartOperationalStatus in StartFlowEventHandler to instead be checked in virtualNodeFlowOperationalChecks, after the flow checkpoint has been made. 

This means we don't have to create a FlowStatus without a checkpoint, which should remove the case where one was being made without a FlowInitiatorType.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/factory/impl/FlowMessageFactoryImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/factory/impl/FlowMessageFactoryImpl.kt
@@ -56,9 +56,6 @@ class FlowMessageFactoryImpl(private val currentTimeProvider: () -> Instant) : F
 
     private fun getCommonFlowStatus(checkpoint: FlowCheckpoint): FlowStatus {
         val startContext = checkpoint.flowStartContext
-        if (startContext.initiatorType == null) {
-            log.warn("Creating FlowStatus with null initiatorType")
-        }
         return FlowStatus().apply {
             key = startContext.statusKey
             initiatorType = startContext.initiatorType
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/factory/impl/FlowMessageFactoryImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/factory/impl/FlowMessageFactoryImpl.kt
@@ -7,6 +7,7 @@ import net.corda.flow.pipeline.factory.FlowMessageFactory
 import net.corda.flow.state.FlowCheckpoint
 import org.osgi.service.component.annotations.Activate
 import org.osgi.service.component.annotations.Component
+import org.slf4j.LoggerFactory
 import java.time.Instant
 
 @Component(service = [FlowMessageFactory::class])
@@ -15,6 +16,10 @@ class FlowMessageFactoryImpl(private val currentTimeProvider: () -> Instant) : F
 
     @Activate
     constructor() : this(Instant::now)
+
+    private companion object {
+        val log = LoggerFactory.getLogger(this::class.java.enclosingClass)
+    }
 
     override fun createFlowCompleteStatusMessage(checkpoint: FlowCheckpoint, flowResult: String?): FlowStatus {
         return getCommonFlowStatus(checkpoint).apply {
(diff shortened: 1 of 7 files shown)


components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/factory/impl/FlowMessageFactoryImpl.kt
```diff
@@ -51,6 +56,9 @@ class FlowMessageFactoryImpl(private val currentTimeProvider: () -> Instant) : F
 
     private fun getCommonFlowStatus(checkpoint: FlowCheckpoint): FlowStatus {
         val startContext = checkpoint.flowStartContext
+        if (startContext.initiatorType == null) {
+            log.warn("Creating FlowStatus with null initiatorType")
```
vkolomeyko: Do we want to add more context to this warning logged from `FlowCheckpoint` ? So that we will be able to trace back how come that `initiatorType` is `null`.

LWogan (changes requested):
Feedback given on slack. I think we can avoid a lot of the extra handling for no checkpoints by allowing the start flow handler to initialize the checkpoint and then we do the operational checks as part of the operational checks step in the flow pipeline

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/factory/impl/FlowMessageFactoryImpl.kt
Tom-Fitzpatrick: More context would be better alright, but for now I've actually removed this log as we're hopefully fixing this issue with a different approach, detailed in the PR description.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/StartFlowEventHandler.kt
```diff
@@ -31,16 +29,6 @@ class StartFlowEventHandler @Activate constructor(
     override fun preProcess(context: FlowEventContext<StartFlow>): FlowEventContext<StartFlow> {
         log.info("Flow [${context.checkpoint.flowId}] started")
 
-        val holdingIdentity = context.inputEventPayload.startContext.identity.toCorda()
```
vkolomeyko, line 34: Looks like we now have unused imports:
```
components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/StartFlowEventHandler.kt:7:1: The import 'net.corda.flow.pipeline.exceptions.FlowMarkedForKillException' is unused. [UnusedImports]	
components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/StartFlowEventHandler.kt:9:1: The import 'net.corda.virtualnode.OperationalStatus' is unused. [UnusedImports]	
components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/StartFlowEventHandler.kt:10:1: The import 'net.corda.virtualnode.read.VirtualNodeInfoReadService' is unused. [UnusedImports]	
components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt:3:1: The import 'net.corda.data.flow.FlowInitiatorType' is unused. [UnusedImports]	
components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt:4:1: The import 'net.corda.data.flow.event.StartFlow' is unused. [UnusedImports]	
components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventExceptionProcessorImpl.kt:5:1: The import 'net.corda.data.flow.output.FlowStates' is unused. [UnusedImports]	
components/flow/flow-service/src/test/kotlin/net/corda/flow/pipeline/handlers/events/StartFlowEventHandlerTest.kt:13:1: The import 'net.corda.virtualnode.read.VirtualNodeInfoReadService' is unused. [UnusedImports]
```
and Detekt is not happy.

Tom-Fitzpatrick: Got it, thanks :)

LWogan (dismissed):
LGTM. Be sure to run the e2e tests against the helm chart built for this pr. I think you already said you did on slack but just double check.

Tom-Fitzpatrick:
I ran the e2e tests locally as well as [here](https://ci02.dev.r3.com/job/Corda5/view/Full%20Health%205.2/job/corda5-e2e-tests/job/release%252F5.2/205/testReport/) and the ones relevant to this PR were all green.

vkolomeyko (approved):
LGTM

LWogan (approved):
LGTM

PR #5247, opened 2023-12-12
CORE-18320 Support for a flow specific timeout on Flow Message API

mbrkic-r3:
Support for a flow specific timeout on Flow Message API. Client can specify a flow session specific timeout on Flow Message API. If set, it will be passed to the initiated party to also use it. If the client does not provide a timeout value then the value from Corda Configuration is used.

Session timeout is in `SessionState` stored in field `sessionProperties`. With this PR `requireClose` is also stored in `sessionProperties` instead of in separate field.


simon-johnson-r3 (dismissed):

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
```diff
@@ -120,10 +121,14 @@ class SessionEventHandler @Activate constructor(
         val counterpartySessionPropertiesMap = counterpartySessionProperties.toMap()
         val requireClose = counterpartySessionPropertiesMap[FLOW_SESSION_REQUIRE_CLOSE] ?: throw FlowFatalException("RequireClose was not" +
                 " set in the session properties")
+        val sessionTimeoutMs = counterpartySessionPropertiesMap[FLOW_SESSION_TIMEOUT_MS]
         val sessionContext = KeyValueStore().apply {
             put(FLOW_PROTOCOL, protocolVersion.protocol)
             put(FLOW_PROTOCOL_VERSION_USED, protocolVersion.protocolVersion.toString())
             put(FLOW_SESSION_REQUIRE_CLOSE, requireClose)
+            if (sessionTimeoutMs != null) {
+                put(FLOW_SESSION_TIMEOUT_MS, sessionTimeoutMs)
```
LWogan, line 130: Is this to handle when the initiating flow doesn't provide a timeout?

What do you think about always setting the timeout by the initiating flow. So it would use their config timeout  if its not set so that both parties are in sync. Otherwise the other party could be on a cluster with a different value and this might result in unexpected behavior?

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/SessionManagerImpl.kt
```diff
@@ -176,7 +176,11 @@ class SessionManagerImpl @Activate constructor(
     ): List<SessionEvent> {
         val lastReceivedMessageTime = sessionState.lastReceivedMessageTime
 
-        val sessionTimeoutTimestamp = lastReceivedMessageTime.plusMillis(config.getLong(FlowConfig.SESSION_TIMEOUT_WINDOW))
+        val sessionTimeout = sessionState.sessionProperties?.let {
```
LWogan: I think we could probably just delete this whole method and handle the session timeouts in one one place, the flow session timeout scheduled job.

LWogan (commented):
Logic is missing to set the Checkpoint flow session expiry MetaData in FlowGlobalPostProcessorImpl.getStateMetadata.

Couple other comments open for debate.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/handlers/events/SessionEventHandler.kt
mbrkic-r3: This will propagate flow specific timeout to initiated flow, but if there was no flow specific timeout set then flows will use values set in cluster configurations. This might mean different values for initiating and initiated flows, but that's how it was working so far, so I kept it the same as it was. Should we change it?

libs/flows/session-manager-impl/src/main/kotlin/net/corda/session/manager/impl/SessionManagerImpl.kt
mbrkic-r3: Changed

LWogan (dismissed):

LWogan (approved):

PR #5256, opened 2023-12-13
CORE-18628 Add StatefulSessionManager to LinkManager

williamvigorr3:
Add Stateful Session Manager as a skeleton class (in a follow up we will change the code to add the state manager to the `StatefulSessionManager`). This can be turned on by a feature flag.

Before this can be merged I need to update the unit tests broken by this change (but it will be quickest to get the implementation reviewed first).


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -55,150 +55,196 @@ internal class InboundMessageProcessor(
     EventLogProcessor<String, LinkInMessage> {
 
     private var logger = LoggerFactory.getLogger(this::class.java.name)
+    private companion object {
```
yift-r3, line 59: (We should probably move the logger to the companion object and make it val)

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -55,150 +55,196 @@ internal class InboundMessageProcessor(
...
-                    val partitionsAssigned =
-                        inboundAssignmentListener.getCurrentlyAssignedPartitions()
-                    if (partitionsAssigned.isNotEmpty()) {
-                        listOf(
-                            Record(Schemas.P2P.LINK_OUT_TOPIC, LinkManager.generateKey(), response),
-                            Record(
-                                Schemas.P2P.SESSION_OUT_PARTITIONS,
-                                payload.header.sessionId,
-                                SessionPartitions(partitionsAssigned.toList())
+        val responses = sessionManager.processSessionMessages(messages.map { it.message })
```
yift-r3: I don't like this pattern. It assumes that the reply list will be aligned with the arguments.

Maybe if the processSessionMessages would accept a callback it would be nicer.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -55,150 +55,196 @@ internal class InboundMessageProcessor(
...
+                        traceEventProcessing(messages[i].originalRecord, tracingEventName) { records }
+                        recordOutboundSessionMessagesMetric(response.header.sourceIdentity, response.header.destinationIdentity)
+                        records
                     }
                 }
-                is InitiatorHandshakeMessage -> {
-                    listOf(Record(Schemas.P2P.LINK_OUT_TOPIC, LinkManager.generateKey(), response))
+            } else {
+                val records = emptyList<Record<String, *>>()
+                traceEventProcessing(messages[i].originalRecord, tracingEventName) { records }
```
yift-r3: Is there a way to move the `traceEventProcessing` to a single place?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -55,150 +55,196 @@ internal class InboundMessageProcessor(
...
+                records
+            }
+        }.flatten()
+    }
+
+    private fun processDataMessages(
+        sessionIdAndMessages: List<Pair<String, TraceableMessage<AvroSealedClasses.DataMessage>>>
```
yift-r3: I'm not sure it will help, but maybe instead of the pair, we could have a data class?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -55,150 +55,196 @@ internal class InboundMessageProcessor(
...
+    ): List<Record<*, *>> {
+        val messages = mutableListOf<Record<*, *>>()
+        val sessionDirectionForMessage = sessionManager.getSessionsById(sessionIdAndMessages.map { it.first })
```
yift-r3: Again, I would prefer to do the get sessions one by one.

Also, would it be reasonable to assume that we will have more than one message for the same session? If so, maybe a map from the session ID to a collection of messages would be better?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -55,150 +55,196 @@ internal class InboundMessageProcessor(
...
+        for ((i, sessionDirection) in sessionDirectionForMessage.withIndex()) {
+            when (sessionDirection) {
+                is SessionManager.SessionDirection.Inbound ->
+                    messages.addAll(processInboundDataMessages(sessionIdAndMessages[i], sessionDirection))
+                is SessionManager.SessionDirection.Outbound -> processOutboundDataMessage(sessionIdAndMessages[i], sessionDirection)?.let {
+                    messages.add(it)
```
yift-r3: Using a `.map` would have allowed us to filter out nulls instead of having an immutable list.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManager.kt
```diff
@@ -10,19 +9,13 @@ import net.corda.p2p.crypto.protocol.api.Session
 import net.corda.virtualnode.HoldingIdentity
 
 internal interface SessionManager : LifecycleWithDominoTile {
-    fun processOutboundMessage(message: AuthenticatedMessageAndKey): SessionState
-    fun getSessionById(uuid: String): SessionDirection
-    fun processSessionMessage(message: LinkInMessage): LinkOutMessage?
+    fun processOutboundMessages(messages: List<AuthenticatedMessageAndKey>): List<SessionState>
+    fun getSessionsById(uuids: List<String>): List<SessionDirection>
+    fun processSessionMessages(messages: List<LinkInMessage>): List<LinkOutMessage?>
```
dimosr: Now that these are converted to batch APIs, would something like the following be a better interface?
```
fun processOutboundMessages(messages: List<AuthenticatedMessageAndKey>): List<Pair<AuthenticatedMessageAndKey, SessionState>>
fun getSessionsById(uuids: List<String>): Map<String, SessionDirection>
fun processSessionMessages(messages: List<LinkInMessage>): List<Pair<LinkInMessage, LinkOutMessage?>>
```
That way we won't have to do a lot of mangling at the client side to match input to output. It's also a relatively standard pattern for batch APIs.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
```diff
@@ -88,14 +90,35 @@ internal class OutboundMessageProcessor(
                 }
             }
         }
-    }
 
-    override fun onNext(events: List<EventLogRecord<String, AppMessage>>): List<Record<*, *>> {
-        val records = mutableListOf<Record<*, *>>()
-        for (event in events) {
-            records += traceEventProcessing(event,"P2P Link Manager Outbound Event") { processEvent(event)}
+        fun recordsForSessionEstablished(
```
yift-r3: (Probably worth making it internal)

Is that the correct class for this function?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
```diff
@@ -247,19 +283,58 @@ internal class OutboundMessageProcessor(
...
-        processAuthenticatedMessage(messageAndKey, true)
+        processAuthenticatedMessages(listOf(TraceableMessage(messageAndKey, null)), true)
+
+    private fun processAuthenticatedMessages(
+        messagesWithKeys: List<TraceableMessage<AuthenticatedMessageAndKey>>,
+        isReplay: Boolean = false
+    ): List<Record<String, *>> {
+        val markerMessagesWithSession = mutableListOf<Record<String, *>>()
+        val messageWithNoSession = mutableListOf<Record<String, *>>()
+        val remoteAuthenticatedMessages = mutableListOf<TraceableMessage<AuthenticatedMessageAndKey>>()
```
yift-r3: Wouldn't a map followed by a flatMap be more readable?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
```diff
@@ -247,19 +283,58 @@ internal class OutboundMessageProcessor(
...
      *
      * [isReplay] - If the message is being replayed we don't persist a [LinkManagerSentMarker] as there is already
      * a marker for this message. If the process is restarted we reread the original marker.
+     *
+     * @return
+     * [ValidateAuthenticatedMessageResult.SessionNeeded] if validation succeeds and the destination identity is hosted in a
+     * different corda cluster. This contains the authenticated message to be authenticated (and optionally encrypted by the session, along
+     * with some marker records.
+     * [ValidateAuthenticatedMessageResult.NoSessionNeeded] in all other cases. This contains some marker records and the message to be
+     * looped back (if validation succeeded).
```
yift-r3, line 300: It's a private function; I'm not sure the docs are helpful.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/PendingSessionMessageQueuesImpl.kt
```diff
@@ -67,7 +75,9 @@ internal class PendingSessionMessageQueuesImpl(
                     "Sending queued message ${message.message.header.messageId} " +
                         "to newly established session ${session.sessionId} with ${counterparties.counterpartyId}"
                 }
-                records.addAll(sessionManager.recordsForSessionEstablished(session, message, counterparties.serial))
+                records.addAll(OutboundMessageProcessor.recordsForSessionEstablished(
```
yift-r3: Maybe we should have a component that holds the session manager, the group policy providers and the clock and use it for the `recordsForSessionEstablished` instead of using them as arguments.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -55,150 +55,196 @@ internal class InboundMessageProcessor(
...
 
-                    else -> {
-                        logger.error("Received unknown payload type ${message.payload::class.java.simpleName}. The message was discarded.")
-                        emptyList()
+                is InboundUnauthenticatedMessage -> {
+                    logger.debug {
+                        "Processing unauthenticated message ${payload.header.messageId}"
                     }
+                    recordInboundMessagesMetric(payload)
+                    listOf(
```
dimosr: unless I am missing something, this list is not included in the messages returned from the event loop?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManager.kt
```diff
@@ -10,19 +9,13 @@ import net.corda.p2p.crypto.protocol.api.Session
 import net.corda.virtualnode.HoldingIdentity
 
 internal interface SessionManager : LifecycleWithDominoTile {
-    fun processOutboundMessage(message: AuthenticatedMessageAndKey): SessionState
-    fun getSessionById(uuid: String): SessionDirection
-    fun processSessionMessage(message: LinkInMessage): LinkOutMessage?
+    fun processOutboundMessages(messages: List<AuthenticatedMessageAndKey>): List<SessionState>
```
yift-r3: I prefer collections instead of lists.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -55,150 +55,196 @@ internal class InboundMessageProcessor(
...
+                        records
                     }
                 }
-                is InitiatorHandshakeMessage -> {
-                    listOf(Record(Schemas.P2P.LINK_OUT_TOPIC, LinkManager.generateKey(), response))
+            } else {
+                val records = emptyList<Record<String, *>>()
+                traceEventProcessing(messages[i].originalRecord, tracingEventName) { records }
+                records
+            }
```
dimosr: minor: could we skip tracing, since there's nothing to trace (looking at the tracing library, it doesn't do much for an empty list which would make sense).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -55,150 +55,196 @@ internal class InboundMessageProcessor(
...
+                sessionIdAndMessage.first,
+                sessionIdAndMessage.second.message
+            ).also { records ->
+                traceEventProcessing(sessionIdAndMessage.second.originalRecord, tracingEventName) { records }
             }
-            records
         } else {
+            traceEventProcessing(sessionIdAndMessage.second.originalRecord, tracingEventName) { emptyList() }
             emptyList()
         }
```
dimosr: similar comment here on whether we could skip tracing to make the code simpler, since there's nothing to trace.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
williamvigorr3: > Again, I would prefer to do the get sessions one by one.

The purpose of this change is so we can eventually batch the query to the State Manager, so we can't get the sessions one by one.
 
> Also, would it be reasonable to assume that we will have more than one message for the same session? If so, maybe a map from the session ID to a collection of messages would be better?

That is possible although making that work with the event tracing code (which assumes a one to one mapping between event and records more complex).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -55,150 +55,196 @@ internal class InboundMessageProcessor(
...
-        val records = mutableListOf<Record<*, *>>()
+        val dataMessages = mutableListOf<Pair<String, TraceableMessage<AvroSealedClasses.DataMessage>>>()
+        val sessionMessages = mutableListOf<TraceableMessage<LinkInMessage>>()
         for (event in events) {
             val message = event.value
             if (message == null) {
                 logger.error("Received null message. The message was discarded.")
                 continue
             }
-            records += traceEventProcessing(event, "P2P Link Manager Inbound Event") {
```
dimosr, line 67: would it be possible to retain that at the top level and execute it for all the accumulated records?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
dimosr: update: I added a separate comment about the possibility of retaining the tracing calls at the top level. If that's possible and we do that, this would be irrelevant.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -0,0 +1,59 @@
...
+        return emptyList()
+    }
+
+    override fun getSessionsById(uuids: List<String>): List<SessionManager.SessionDirection> {
+        return emptyList()
+    }
+
+    override fun processSessionMessages(messages: List<LinkInMessage>): List<LinkOutMessage?> {
+        return emptyList()
+    }
```
dimosr: it would be good to reference the ticket that is supposed to fill in the implementation of those, similar to below.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    override fun inboundSessionEstablished(sessionId: String) {
```
dimosr: Worth pulling all the functions that are not needed by this implementation at the bottom to avoid distraction.

And maybe have a consistent comment for all of them, e.g. `// Not needed by the Stateful Session Manager`. Originally, I liked your comment explaining what's the reason it's not needed (e.g. `/No heartbeat manager in the stateful session manager`), but then I realised that's referring to another implementation, so it can end up being confusing. So we could keep it generic for now.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
dimosr: Also, given this executes at the top level, I would assume it emits traces for everything (markers as well as messages)? When I was looking at the outbound process, I got the impression that we are not emitting traces for some messages, such as markers in some cases.

dimosr (commented):
My main high-level comment is on adjusting the interface of the SessionManager to allow the APIs that are now batch-based to return input along with output, so that it's more convenient for clients to do any mapping they want instead of relying on ordering.

The rest are mostly lower-level comments.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManager.kt
williamvigorr3: I went for something similar to this in 4e11a55. I used a lambda as it lets us pass in a Traceable message, so we can do all the tracing at once. Let me know what you think? I still need to fix some of the other more minor comments.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
williamvigorr3: This has changed significantly now.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManager.kt
dimosr: Unfortunately, the traceability requirements make the interface a bit harder to grasp at first due to the lambdas that feel a bit out of place if you look only at the interface. But, I can't see a better way of achieving this at the moment, so that we can trace in a centralised way. So, fine by me unless someone else has a better idea.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
```diff
@@ -130,22 +130,22 @@ internal class OutboundMessageProcessor(
     /**
      * Class which wraps a message, so we can do event tracing by passing the original event log record to a lower layer.
      */
-    private data class TraceableMessage<T>(
-        val message: T,
+    private data class TraceableItem<T>(
```
dimosr: I noticed that we have 2 types for this: `TraceableItem` and `TraceableMessage`, with the main difference being the type of `originalRecord`. Does this type matter or could we have a single type with a generic type that's used across both inbound and outbound processor?

`traceEventProcessing` seems to be happy enough with just a `EventLogRecord<*, *>`.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
williamvigorr3: We now do all the tracing at the top level (will do some more tidy up of the code structure).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
williamvigorr3: This is done at the top level now.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
williamvigorr3: Tracing is now done at the top level (so this doesn't matter anymore).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
williamvigorr3: I've restructured the SessionManger interface to let us do this. Some discussion here: https://github.com/corda/corda-runtime-os/pull/5256#discussion_r1431177946

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
williamvigorr3: The second point has been resolved.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
williamvigorr3: I made it internal. 

We already have `recordsForSessionEstablished` in this class (as well as functions to make records for markers). As the logic will be the same for `StatefulSessionManagerImpl` and `SessionManagerImpl` it needs to be pulled out of `SessionManager`.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
williamvigorr3: I don't have a strong opinion either way.  I will delete this if you want? I think these were added because what happens here is quite complicated.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/PendingSessionMessageQueuesImpl.kt
williamvigorr3: I think that just leads to more indirection.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
williamvigorr3: I refactored this but I am not entirely sure what you had in mind.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
williamvigorr3: I'll do this. It helps a bit.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -54,151 +55,199 @@ internal class InboundMessageProcessor(
...
+    )
 
     override fun onNext(events: List<EventLogRecord<String, LinkInMessage>>): List<Record<*, *>> {
-        val records = mutableListOf<Record<*, *>>()
-        for (event in events) {
-            val message = event.value
-            if (message == null) {
-                logger.error("Received null message. The message was discarded.")
-                continue
+        val dataMessages = events.mapNotNull { event ->
```
yift-r3: Should we do a `groupBy` to pass over the list once?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
```diff
@@ -54,151 +55,199 @@ internal class InboundMessageProcessor(
...
+                    )),
+                    event
+                )
+            } else {
+                null
+            }
+        }
+
+        return (processSessionMessages(sessionMessages) + processDataMessages(dataMessages) + recordsForUnauthenticatedMessage)
+            .map { traceable ->
```
yift-r3: is `map...flatten` different from `flatMap`?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
```diff
@@ -247,19 +288,57 @@ internal class OutboundMessageProcessor(
...
+            message to validateAndCheckIfSessionNeeded(message.item, isReplay)
+        }
+        val messagesWithSession = validatedMessages.mapNotNull { (message, result) ->
+            if (result is ValidateAuthenticatedMessageResult.SessionNeeded) {
+                TraceableItem(result, message.originalRecord)
+            } else {
+                null
+            }
+        }
+        val messageWithNoSession = validatedMessages.mapNotNull { (message, result) ->
```
yift-r3, line 271: group by?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManager.kt
```diff
@@ -1,28 +1,24 @@
...
 import net.corda.data.p2p.LinkOutMessage
 import net.corda.data.p2p.app.MembershipStatusFilter
 import net.corda.p2p.crypto.protocol.api.Session
 import net.corda.virtualnode.HoldingIdentity
 
 internal interface SessionManager : LifecycleWithDominoTile {
-    fun processOutboundMessage(message: AuthenticatedMessageAndKey): SessionState
-    fun getSessionById(uuid: String): SessionDirection
-    fun processSessionMessage(message: LinkInMessage): LinkOutMessage?
+    fun <T>processOutboundMessages(
```
yift-r3, line 12: Should we limit T to one of the sealed classes?

yift-r3 (commented):
A few nitpick comments, look OK to me.

dimosr (commented):
Reviewed diff since latest review. Just one comment, although minor so happy for it to be addressed in a follow-up in the interest of expediting the merge of this to unblock the subsequent work.

dimosr:
Also, if it's only unit tests remaining and the implementation stabilised worth kicking of a run of the e2e tests (single and multi-cluster) in the background. So, that we can merge with some extra confidence there's no regression on the existing path.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManager.kt
williamvigorr3: Unfortunately we use `TraceableItem<ValidateAuthenticatedMessageResult.SessionNeeded>` as the template argument in the `OutboundMessageProcessor` and not the sealed classes from this interface (not sure if that is what you are suggesting). 

I think it's best to keep them uncoupled and not make `ValidateAuthenticatedMessageResult.SessionNeeded` public. The lambda means that the user of this interface is restricted in what they can do (as the lambda must return an `AuthenticatedMessageAndKey`).

williamvigorr3:
> Also, if it's only unit tests remaining and the implementation stabilised worth kicking of a run of the e2e tests (single and multi-cluster) in the background. So, that we can merge with some extra confidence there's no regression on the existing path.

Multi cluster E2E tests are running here:
https://ci02.dev.r3.com/job/Corda5/job/corda-e2e-tests-multi-cluster-tests/view/change-requests/job/PR-379/2/console

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
williamvigorr3: I don't think we can do that without making unsafe casts (If this worked I would replace the ::Class with an enum type):

I tired this:
```
val sorted = events.groupBy ({ event ->
            when (val payload = event.value?.payload) {
                is AuthenticatedDataMessage, is AuthenticatedEncryptedDataMessage  -> {
                    AuthenticatedDataMessage::class.java
                }
                is ResponderHelloMessage, is ResponderHandshakeMessage, is InitiatorHandshakeMessage, is InitiatorHelloMessage -> {
                    ResponderHelloMessage::class.java
                }
                else -> {null}
            }

        }) { event ->
           when (val payload = event.value?.payload) {
               is AuthenticatedDataMessage -> {
                   payload.header.sessionId.let { sessionId ->
                       sessionId  to TraceableMessage(AvroSealedClasses.DataMessage.Authenticated(payload), event)
                   }
               }
               is AuthenticatedEncryptedDataMessage -> {
                   payload.header.sessionId.let { sessionId ->
                       sessionId to TraceableMessage(AvroSealedClasses.DataMessage.AuthenticatedAndEncrypted(payload), event)
                   }
               }
               is ResponderHelloMessage, is ResponderHandshakeMessage, is InitiatorHandshakeMessage, is InitiatorHelloMessage -> {
                   TraceableMessage(event.value, event)
               }
               is InboundUnauthenticatedMessage {
                   logger.debug {
                       "Processing unauthenticated message ${payload.header.messageId}"
                   }
                   recordInboundMessagesMetric(payload)
                   TraceableMessage(listOf(
                       Record(
                           Schemas.P2P.P2P_IN_TOPIC,
                           LinkManager.generateKey(),
                           AppMessage(payload)
                       )),
                       event
                   )
               }
               else -> {
                   null
               }
           }
       }

```
All the lists we are grouping into have different types.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
williamvigorr3: I changed this to use a single generic type.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessor.kt
williamvigorr3: I fixed this.

williamvigorr3:
Single cluster E2E test running here: https://ci02.dev.r3.com/job/Corda5/job/corda5-e2e-tests/job/PR-379/2/

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessor.kt
williamvigorr3: Again the types of `messagesWithSession` and `messageWithNoSession` are different. So I don't think groupBy helps.

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessorTest.kt
```diff
@@ -119,9 +125,6 @@ class InboundMessageProcessorTest {
         )
 
         assertThat(records).isEmpty()
-        assertThat(loggingInterceptor.errors)
-            .hasSize(2)
-            .contains("Received null message. The message was discarded.")
```
dimosr: should we keep this error logging around?

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/inbound/InboundMessageProcessorTest.kt
```diff
@@ -134,9 +137,6 @@ class InboundMessageProcessorTest {
         )
 
         assertThat(records).isEmpty()
-        assertThat(loggingInterceptor.errors)
```
dimosr: should we keep this error logging around?

dimosr (commented):
A couple of comments after reviewing the latest diff around logging that seems to have been removed.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/EstablishedSessionRecorder.kt
```diff
@@ -0,0 +1,44 @@
...
+import net.corda.membership.grouppolicy.GroupPolicyProvider
+import net.corda.membership.read.MembershipGroupReaderProvider
+import net.corda.p2p.crypto.protocol.api.Session
+import net.corda.utilities.time.Clock
+import net.corda.messaging.api.records.Record
+import net.corda.p2p.linkmanager.LinkManager
+import net.corda.p2p.linkmanager.common.MessageConverter
+import net.corda.schema.Schemas
+
+internal class EstablishedSessionRecorder(
```
dimosr: I think this name can be a bit misleading, since that component is not responsible for recording established sessions if I understand correctly.

Was this method extracted out to facilitate unit testing or were there other reasons? Would `MessageConverter` be a good home for it considering what the method does, instead of creating a new class?

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidatorTest.kt
```diff
@@ -127,32 +119,16 @@ class NetworkMessagingValidatorTest {
     @ParameterizedTest
     @MethodSource("allowedMessagingGroups")
     fun `invokeIfValidInbound performs as expected`(testConfig: TestConfig) {
```
dimosr:
```suggestion
    fun `isValidInbound performs as expected`(testConfig: TestConfig) {
```

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/membership/NetworkMessagingValidatorTest.kt
```diff
@@ -127,32 +119,16 @@ class NetworkMessagingValidatorTest {
...
-        } else {
-            verify(testLambda, never()).invoke()
-            assertThat(output).isNull()
-        }
+        assertThat(valid).isEqualTo(testConfig.canMessage)
     }
 
     @ParameterizedTest
     @MethodSource("allowedMessagingGroups")
     fun `invokeIfValidOutbound performs as expected`(testConfig: TestConfig) {
```
dimosr: isn't this now exactly the same as the previous test? i.e. should we remove it?

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessorTest.kt
```diff
@@ -1066,55 +1076,82 @@ class OutboundMessageProcessorTest {
...
+            val marker = ((it.value as? AppMessageMarker)?.marker) as? LinkManagerProcessedMarker
+            val messageAndKey = marker?.message
+            assertThat(messages).contains(messageAndKey?.message)
         }
         verify(messagesPendingSession, never()).queueMessage(any(), any())
     }
-
     @Test
-    fun `processReplayedAuthenticatedMessage call to recordsForSessionEstablished`() {
+    fun `onNext returns session recorded messages`() {
```
dimosr:
```suggestion
    fun `onNext returns messages returned from recordsForSessionEstablished`() {
```

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessorTest.kt
```diff
@@ -1039,7 +1049,7 @@ class OutboundMessageProcessorTest {
     @Test
     fun `onNext produces a LinkManagerProcessedMarker per message if SessionEstablished`() {
```
dimosr, line 1050:
```suggestion
    fun `onNext produces a LinkManagerProcessedMarker per message if SessionEstablished, but no messages if recordsForSessionEstablished returns no records`() {
```
to make the scenario exercised a bit more explicit?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/EstablishedSessionRecorder.kt
yift-r3: The message converter is a stateless class where all the functions are in the companion object. I want to have a mockable class that holds the classes and can be mocked (so we can test this function regardless of the implementation).

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/outbound/OutboundMessageProcessorTest.kt
yift-r3: It's too long for detekt (but this is not really the use case, I'm testing that it produced the `LinkManagerProcessedMarker` regardless of the `recordsForSessionEstablished` and in the next test I'm testing the `recordsForSessionEstablished` regardless of the `LinkManagerProcessedMarker`, I'll change the test).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/EstablishedSessionRecorder.kt
dimosr: Assuming you agree that the responsibilities of that class fall very close to what we are doing here, would there be an issue with making it a stateful class now (i.e. introducing the constructor parameters needed) and also introduce this method as an instance method, instead of static? In theory, we could consider converting all the existing methods into instance methods as well, but we can skip that for this PR to save some time.

dimosr: Or alternatively as a workaround for now, we could introduce it as a companion method as well. And for any mocking needed, we can use static mocking functionality from mockito.

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/common/MessageConverterTest.kt
```diff
@@ -220,6 +230,146 @@ class MessageConverterTest {
...
+            )
+            val linkOut = records.firstOrNull {
+                it.topic == Schemas.P2P.LINK_OUT_TOPIC
+            }?.value as? LinkOutMessage
+
+            assertThat(linkOut).isNotNull()
+        }
+
+        @Test
+        fun `it will create marker topic`() {
```
dimosr:
```suggestion
        fun `it will create marker message`() {
```

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/common/MessageConverterTest.kt
```diff
@@ -220,6 +230,146 @@ class MessageConverterTest {
...
+                session,
+                1L,
+                messageAndKey,
+            )
+
+            verify(sessionManager).dataMessageSent(session)
+        }
+
+        @Test
+        fun `it with unknown member will not create any record`() {
```
dimosr:
```suggestion
        fun `when called with unknown member, it will not create any record`() {
```

components/link-manager/src/test/kotlin/net/corda/p2p/linkmanager/common/MessageConverterTest.kt
```diff
@@ -220,6 +230,146 @@ class MessageConverterTest {
...
+                session,
+                1L,
+                messageAndKey,
+            )
+
+            assertThat(records).isEmpty()
+        }
+
+        @Test
+        fun `it with unknown member will not notify the manager`() {
```
dimosr:
```suggestion
        fun `when called with unknown member, it will not notify the manager`() {
```

dimosr (commented):
LGTM - just a few minor comments on naming/tests

dimosr (approved):
LGTM

P.S.: If you think we should revisit `MessageConverter` and convert the static methods, feel free to raise a ticket under our tech debt epic so that we can look into it.

yift-r3:
https://r3-cev.atlassian.net/browse/CORE-19121 created

PR #5258, opened 2023-12-13
CORE-18635 Crypto API for session encryption

YashNabar:
Implementation of the encryption/decryption APIs.

API pull request in https://github.com/corda/corda-api/pull/1400


libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/ApiNames.kt
```diff
@@ -0,0 +1,6 @@
+package net.corda.crypto.core
+
+object ApiNames {
+    const val ENCRYPT_PATH = "/crypto-api/session/encrypt"
+    const val DECRYPT_PATH = "/crypto-api/session/decrypt"
```
dimosr, line 5: I know it was in the design like that, but I am now wondering should we make this something generic (as we tried in general to keep the API as generic as possible)? e.g. `crypto-api/encrypt` and `crypto-api/decrypt`.

I guess that would apply generally (as the naming of the classes also includes session for example). 

@dickon  / @driessamyn / @owenstanford can provide another opinion, since they were involved in the original design discussions.

components/crypto/crypto-client/src/main/kotlin/net/corda/crypto/client/SessionEncryptionOpsClient.kt
```diff
@@ -0,0 +1,42 @@
+package net.corda.crypto.client
+
+import net.corda.lifecycle.Lifecycle
+
+/**
+ * Encryption operations client to encrypt and decrypt session data for Link Manager.
+ */
+interface SessionEncryptionOpsClient : Lifecycle {
```
dimosr, line 8: Related to the other comment, I feel we should keep these generic for encryption/decryption. Not tied to sessions.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/SessionDecryptionProcessor.kt
```diff
@@ -0,0 +1,53 @@
...
+    override val requestClass = DecryptRpcCommand::class.java
+    override val responseClass = DecryptionOpsResponse::class.java
+
+    override fun process(request: DecryptRpcCommand): DecryptionOpsResponse {
+        logger.trace { "Processing request: ${request::class.java.name}" }
+
+        return try {
+            executor.executeWithRetry {
+                val plainBytes = cryptoService.decrypt(
+                    CryptoTenants.P2P,
```
dimosr, line 40: Looking at the design, I can see the tenant being provided externally (at the API level, instead of being hardcoded on the server side). I think that would be in line with keeping the API generic, even if we have only one client for the time being.

dimosr (commented):
Code looks reasonable in general, but I think there's an open question for me whether the API should be structured in a more generic way and not tied to sessions / p2p. I've left a few comments on various places, where that's relevant.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/ApiNames.kt
dickon: I think the API paths will probably be important for RBAC. The only other point to do IAM would be the port number and as far as I know there's no intention to make this exposed on a separate port, it's all one Javalin service.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/SessionDecryptionProcessor.kt
dickon: So the trouble is we can't verify the tenantID, so if we include the tenantID in the HTTP request payload the RBAC has to get right into the payload or query string parameter to check it.

libs/crypto/crypto-core/src/main/kotlin/net/corda/crypto/core/ApiNames.kt
driessamyn: I agree with @dickon, we don't want to, unnecessarily, widen the attack surface.
So I agree with the service endpoint including the hardcoded "tenant".
This doesn't prevent us of making it dual purpose in the future when the need arrises and we can then make a decision to add more hardcoded endpoints or make it a parameter with validation.

dimosr: ok, makes sense

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/rpc/SessionDecryptionProcessor.kt
dimosr: yes, that makes sense

components/crypto/crypto-client/src/main/kotlin/net/corda/crypto/client/SessionEncryptionOpsClient.kt
dimosr: Given the discussion on the other [thread](https://github.com/corda/corda-runtime-os/pull/5258#discussion_r1428175551) around the URL path containing `session`, given the conclusion to keep it there I guess it shouldn't be a big deal here either since that's not even part of an API and it's changeable even more easily.

dimosr (dismissed):

components/crypto/crypto-client-impl/src/main/kotlin/net/corda/crypto/client/impl/SessionEncryptionOpsClientComponent.kt
```diff
@@ -0,0 +1,61 @@
...
+class SessionEncryptionOpsClientComponent @Activate constructor(
+    @Reference(service = LifecycleCoordinatorFactory::class)
+    coordinatorFactory: LifecycleCoordinatorFactory,
+    @Reference(service = PublisherFactory::class)
+    private val publisherFactory: PublisherFactory,
+    @Reference(service = ConfigurationReadService::class)
+    configurationReadService: ConfigurationReadService,
+    @Reference(service = PlatformInfoProvider::class)
+    val platformInfoProvider: PlatformInfoProvider,
+) : AbstractConfigurableComponent<SessionEncryptionOpsClientComponent.Impl>(
```
kyriathar: We need to completely remove `AbstractConfigurableComponent` at some point. Could we please not use it as it means we 'd be one step further from that goal?

components/crypto/crypto-client-impl/src/main/kotlin/net/corda/crypto/client/impl/SessionEncryptionOpsClientComponent.kt
```diff
@@ -0,0 +1,61 @@
...
+import net.corda.lifecycle.LifecycleCoordinatorFactory
+import net.corda.lifecycle.LifecycleCoordinatorName
+import net.corda.messaging.api.publisher.factory.PublisherFactory
+import net.corda.schema.configuration.ConfigKeys.MESSAGING_CONFIG
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.Reference
+
+@Component(service = [SessionEncryptionOpsClient::class])
+class SessionEncryptionOpsClientComponent @Activate constructor(
```
kyriathar:
```suggestion
class SessionEncryptionOpsClientImpl @Activate constructor(
```

kyriathar (commented):
Quick pass. Is the reason for a new client so it can be over http while `CryptoOpsClient` is over Kafka?

processors/crypto-processor/src/test/kotlin/net/corda/processors/crypto/internal/CryptoProcessorImplTest.kt
```diff
@@ -44,7 +50,7 @@ import org.mockito.kotlin.verify
 import org.mockito.kotlin.whenever
 import java.util.UUID
 
-class CryptoProcessorImplTest {
+class   CryptoProcessorImplTest {
```
anien, line 53: unnecessary new whitespace

```suggestion
class  CryptoProcessorImplTest {
```

anien: unnecessary new whitespace
```suggestion
class CryptoProcessorImplTest {
```

kyriathar (dismissed):
Had a quick look. Looks OK (for crypto). But let's wait for @anien approval as well crypto wise.

anien:
Was there an agreement if we are keeping `AbstractConfigurableComponent` or removing it? @kyriathar @yift-r3

yift-r3:
> Was there an agreement if we are keeping `AbstractConfigurableComponent` or removing it? @kyriathar @yift-r3

Keep it for this pull request.

anien (dismissed):
LGTM.

anien (approved):
LGTM (Crypto).

seanbrereton (approved):
LGTM (BLT)

PR #5268, opened 2023-12-14
CORE-18638 Add permission for ingress controller to read IngressClass

davidcurrie:
Without permission to read cluster-scoped `IngressClass` resources, the ingress-nginx controller processes updates to all `Ingress` resources in the target namespace, regardless of whether or not they are associated with the ingress class it is configured with. (Reported in https://github.com/kubernetes/ingress-nginx/issues/9662.) This results in the controller updating any `Ingress` resources defined for the REST API or P2P Gateway and consequent loss of connectivity until the rightful controller updates the resources again. The status flip-flops between the two controllers.

This change gives the controller permission to view `IngressClass` resources, which it does not really need (it uses the annotation-based approach to defining the ingress class) but causes it to check the ingress class name correctly.

### DIFF
--- a/charts/corda-lib/templates/_nginx.tpl
+++ b/charts/corda-lib/templates/_nginx.tpl
@@ -3,7 +3,10 @@
 {{- end }}
 
 {{- define "corda.nginxClusterUniqueName" -}}
-{{- printf "%s-%s-nginx" .Release.Namespace . }}
+{{- $workerName := index . 1 }}
+{{- with ( index . 0 ) }}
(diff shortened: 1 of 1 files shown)


davidcurrie:
Verified in https://ci02.dev.r3.com/job/Corda5/job/corda-simulations/job/release%2F5.2/50/ that the `corda-rest-worker` ingress is now untouched by the token selection worker ingress controller.

jujoramos (approved):
LGTM

PR #5270, opened 2023-12-14
CORE-18935 handle completion exceptions thrown within the mediator properly

LWogan:
Can be thrown by CompletableFuture join calls. this will wrap the thrown CordaMessageAPIIntermittentExceptions in CompletionExceptions. This can erroneously treat them as fatal


mbrkic-r3 (approved):
LGTM

PR #5274, opened 2023-12-14
CORE-13504: Add CryptoRewrapBusProcessor metrics to Grafana

malachyb:


malachyb:
An image of the Grafana graph
image

anien (approved):
I think we would need to rename few bits to differentiate between managed and unmanaged key rotation, but that will probably come as part of the managed key rotation epic.

PR #5278, opened 2023-12-14
CORE-17813: Implement skip-able backchain with verifying notary

jennyang-r3:
This PR implements skip-able backchain with a verifying notary.
The notary has to be able to: skip when backchain required flag is false with contract verifying notary or not skip when the flag is true.

The e2e tests for the implementation are [here](https://github.com/corda/corda-e2e-tests/pull/366)

### DIFF
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
@@ -10,7 +10,6 @@ data class FinalityPayload @ConstructorForDeserialization constructor(val map: M
     private companion object {
         const val INITIAL_TRANSACTION = "INITIAL_TRANSACTION"
         const val TRANSFER_ADDITIONAL_SIGNATURES = "TRANSFER_ADDITIONAL_SIGNATURES"
-        const val FILTERED_TRANSACTIONS_AND_SIGNATURES = "FILTERED_TRANSACTIONS_AND_SIGNATURES"
     }
     constructor(initialTransaction: UtxoSignedTransaction, transferAdditionalSignatures: Boolean) : this(
         mapOf(
@@ -19,13 +18,6 @@ data class FinalityPayload @ConstructorForDeserialization constructor(val map: M
         )
     )
 
-    constructor(filteredTransactionsAndSignatures: List<FilteredTransactionAndSignatures>) : this(
-        mapOf(FILTERED_TRANSACTIONS_AND_SIGNATURES to filteredTransactionsAndSignatures)
-    )
-
     val initialTransaction get() = map[INITIAL_TRANSACTION] as UtxoSignedTransactionInternal
     val transferAdditionalSignatures get() = map[TRANSFER_ADDITIONAL_SIGNATURES] as Boolean
-
-    @Suppress("UNCHECKED_CAST")
-    val filteredTransactionsAndSignatures get() = map[FILTERED_TRANSACTIONS_AND_SIGNATURES] as List<FilteredTransactionAndSignatures>
 }
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
@@ -10,6 +10,7 @@ data class FinalityPayload @ConstructorForDeserialization constructor(val map: M
     private companion object {
         const val INITIAL_TRANSACTION = "INITIAL_TRANSACTION"
         const val TRANSFER_ADDITIONAL_SIGNATURES = "TRANSFER_ADDITIONAL_SIGNATURES"
+        const val FILTERED_TRANSACTIONS_AND_SIGNATURES = "FILTERED_TRANSACTIONS_AND_SIGNATURES"
     }
     constructor(initialTransaction: UtxoSignedTransaction, transferAdditionalSignatures: Boolean) : this(
         mapOf(
@@ -17,6 +18,14 @@ data class FinalityPayload @ConstructorForDeserialization constructor(val map: M
             TRANSFER_ADDITIONAL_SIGNATURES to transferAdditionalSignatures
         )
     )
+
+    constructor(filteredTransactionsAndSignatures: List<FilteredTransactionAndSignatures>) : this(
+        mapOf(FILTERED_TRANSACTIONS_AND_SIGNATURES to filteredTransactionsAndSignatures)
+    )
+
     val initialTransaction get() = map[INITIAL_TRANSACTION] as UtxoSignedTransactionInternal
     val transferAdditionalSignatures get() = map[TRANSFER_ADDITIONAL_SIGNATURES] as Boolean
+
+    @Suppress("UNCHECKED_CAST")
+    val filteredTransactionsAndSignatures get() = map[FILTERED_TRANSACTIONS_AND_SIGNATURES] as List<FilteredTransactionAndSignatures>
 }
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
@@ -10,6 +10,7 @@ data class FinalityPayload @ConstructorForDeserialization constructor(val map: M
     private companion object {
         const val INITIAL_TRANSACTION = "INITIAL_TRANSACTION"
         const val TRANSFER_ADDITIONAL_SIGNATURES = "TRANSFER_ADDITIONAL_SIGNATURES"
+        const val FILTERED_TRANSACTIONS_AND_SIGNATURES = "FILTERED_TRANSACTIONS_AND_SIGNATURES"
     }
     constructor(initialTransaction: UtxoSignedTransaction, transferAdditionalSignatures: Boolean) : this(
         mapOf(
@@ -18,6 +19,13 @@ data class FinalityPayload @ConstructorForDeserialization constructor(val map: M
         )
     )
 
+    constructor(filteredTransactionsAndSignatures: List<FilteredTransactionAndSignatures>) : this(
+        mapOf(FILTERED_TRANSACTIONS_AND_SIGNATURES to filteredTransactionsAndSignatures)
+    )
+
     val initialTransaction get() = map[INITIAL_TRANSACTION] as UtxoSignedTransactionInternal
     val transferAdditionalSignatures get() = map[TRANSFER_ADDITIONAL_SIGNATURES] as Boolean
+
+    @Suppress("UNCHECKED_CAST")
+    val filteredTransactionsAndSignatures get() = map[FILTERED_TRANSACTIONS_AND_SIGNATURES] as List<FilteredTransactionAndSignatures>
 }
--- a/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
+++ b/components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
@@ -1,6 +1,7 @@
 package net.corda.ledger.utxo.flow.impl.flows.finality.v1
 
 import net.corda.crypto.core.fullId
+import net.corda.crypto.core.fullIdHash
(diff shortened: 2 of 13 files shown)


components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -100,14 +110,41 @@ class UtxoFinalityFlowV1(
...
+                        "Dependent transaction $transactionId does not exist"
+                    }
+
+                    FilteredTransactionAndSignatures(
+                        utxoLedgerService.filterSignedTransaction(dependency)
+                            .withOutputStates(indexes)
+                            .withNotary()
+                            .withTimeWindow()
+                            .build(),
+                        dependency.signatures.filter { initialTransaction.notaryKey.fullIdHash() == it.by }
```
vlajos: Is this working if the transaction's notary key is composite?

jennyang-r3: should it not?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -100,14 +112,54 @@ class UtxoFinalityFlowV1(
...
+                if (initialTransaction.dependencies.isNotEmpty()) {
+                    flowEngine.subFlow(TransactionBackchainSenderFlow(initialTransaction.id, it))
+                } else {
+                    log.trace {
+                        "Transaction with id ${initialTransaction.id} has no dependencies so backchain resolution will not be performed."
+                    }
                 }
             }
+        } else {
+            // TODO - Optimise this logic.
```
nkovacsx: I think we should raise a ticket for this because this will get lost this way. 
If I remember correctly from the PoC there's a code duplication in the finality/notary flows as they are both using this same logic.
I remember I managed to modify the notary payload so we don't have to re-do this logic in the notary client.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -100,14 +112,54 @@ class UtxoFinalityFlowV1(
...
+                        utxoLedgerService.filterSignedTransaction(dependency)
+                            .withOutputStates(indexes)
+                            .withNotary()
+                            .withTimeWindow()
+                            .build(),
+                        dependency.signatures.filter { newTxNotaryKeyIds.contains(it.by) }
+                    )
+                }
+            flowMessaging.sendAll(FinalityPayload(filteredTransactionsAndSignatures), sessions.toSet())
+            log.info(("SENT FILTERED TRANSACTION!!"))
```
nkovacsx: This feels like a dummy log line here. Caps lock logs on info level is not something we should have in final code.
We should either remove this entirely or create a proper trace log line instead.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
```diff
@@ -115,26 +134,65 @@ class UtxoReceiveFinalityFlowV1(
...
+            val dependentStateAndRefs =
+                filteredTransactionsAndSignatures.flatMap { (filteredTransaction, signatures) ->
+                    require(signatures.isNotEmpty()) { "No notary signatures were received" }
+                    filteredTransaction.verify()
+
+                    for (signature in signatures) {
+                        transactionSignatureVerificationService.verifySignature(
+                            filteredTransaction.id,
+                            signature,
+                            filteredTransaction.notaryKey
```
blsemo: What I don't see here is a check that the filtered transactions have been signed by the same notary as the proposed transaction (i.e. the verifying notary that allows us to skip the backchain).

components/ledger/ledger-utxo-flow/src/test/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1Test.kt
```diff
@@ -515,8 +533,50 @@ class UtxoReceiveFinalityFlowV1Test {
         verify(persistenceService, times(2)).persist(signedTransactionWithOwnKeys, TransactionStatus.UNVERIFIED)
     }
```
blsemo, line 557: We probably also want at test where receive finatlity fails to sign the proposed transaction notary does not match the sigantures on the filtered transaction.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
jennyang-r3: Would this check be enough?
```kotlin
  for (signature in signatures) {
       require(notaryInfo.publicKey.fullIdHash() == signature.by) {
           "Signature received \"${signature.by}\" is not signed by current notary \"${notaryInfo.publicKey.fullIdHash()}\""
       }
       transactionSignatureVerificationService.verifySignature(
          filteredTransaction.id,
          signature,
          filteredTransaction.notaryKey
       )
  }
```

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
jennyang-r3: raised thicket here - https://r3-cev.atlassian.net/browse/CORE-19022
Since we have a ticket I'm going to remove TODO comment.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoReceiveFinalityFlowV1.kt
blsemo: No, you have to do the same song and dance for composite keys you have in the `UtxoFinalityFlowV1` line 140. Possibly also check that the notary service names match?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
```diff
@@ -10,13 +10,22 @@ data class FinalityPayload @ConstructorForDeserialization constructor(val map: M
...
+
+    constructor(filteredTransactionsAndSignatures: List<FilteredTransactionAndSignatures>) : this(
+        mapOf(FILTERED_TRANSACTIONS_AND_SIGNATURES to filteredTransactionsAndSignatures)
+    )
+
     val initialTransaction get() = map[INITIAL_TRANSACTION] as UtxoSignedTransactionInternal
     val transferAdditionalSignatures get() = map[TRANSFER_ADDITIONAL_SIGNATURES] as Boolean
+
+    @Suppress("UNCHECKED_CAST")
+    val filteredTransactionsAndSignatures get() = map[FILTERED_TRANSACTIONS_AND_SIGNATURES] as List<FilteredTransactionAndSignatures>
```
vlajos: Why this new data is sent with this PayLoad class?
I thought only the first message needed this evolvable wrapper to let the initiator add extra flow control flags/information later while keeping the first message backwards compatible.
(like `TRANSFER_ADDITIONAL_SIGNATURES`)

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -100,14 +116,62 @@ class UtxoFinalityFlowV1(
...
+                .mapValues { (_, stateRefs) -> stateRefs.map { stateRef -> stateRef.index } }
+                .map { (transactionId, indexes) ->
+                    val dependency = requireNotNull(utxoLedgerService.findSignedTransaction(transactionId)) {
+                        "Dependent transaction $transactionId does not exist"
+                    }
+                    val newTxNotaryKey = initialTransaction.notaryKey
+                    val dependencyNotaryKey = dependency.notaryKey
+                    require(initialTransaction.notaryName == dependency.notaryName) {
+                        "Notary name of filtered transaction \"${dependency.notaryName}\" doesn't match with " +
+                            "any names of initial transaction \"${initialTransaction.notaryName}\""
```
blsemo: That error message does not make sense anymore if you compare the notary service name rather than the representatives' names.
```suggestion
                        "Notary service \"${dependency.notaryName}\"  of input transaction ${filteredTransaction.id} does not match " +
                            "notary service of current transaction \"${initialTransaction.notaryName}\""
```

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
```diff
@@ -100,14 +116,62 @@ class UtxoFinalityFlowV1(
...
+                    }
+                    notarySignatureVerificationService.verifyNotarySignatures(
+                        dependency,
+                        initialTransaction.notaryKey,
+                        dependency.signatures,
+                        mutableMapOf()
+                    )
+
+                    val newTxNotaryKeyIds = if (newTxNotaryKey is CompositeKey) {
+                        require(KeyUtils.isKeyFulfilledBy(newTxNotaryKey, dependencyNotaryKey)) {
```
blsemo: That doesn't work if `dependencyNotaryKey` is a composite key, I believe. What is it you are checking here?

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
jennyang-r3: I don't fully get your point. Would you mind elaborating? 

wouldn't having a separate constructor for `FILTERED_TRANSACTIONS_AND_SIGNATURES` be enough to make it backwards compatible? Are you suggesting to send this new data in some other container than this payload?

blsemo (commented):
II think we're getting there, but I still got 2 questions in the FinalityFlow.

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/v1/UtxoFinalityFlowV1.kt
jennyang-r3: Yeah I was assuming the `dependencyNotaryKey` is public key, not a composite one here. I will need to remove this check

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/flows/finality/FinalityPayload.kt
blsemo: I think it makes a lot of sense to send everything in one message/package in this case.
- There is a non-trivial overhead to introducing another send/receive, especially when vNodes are in different clusters.
- There is no backwards compatibility issue here: this behaviour is switched on a network-wide flag that cannot be changed once set, so both sides are guaranteed to be on 5.2 or later, or won't work at all (not even get to this point).

blsemo: What Lajos is saying: You don't need to wrap `FilteredTransactionsAndSignatures` in a `FinalityPayload` - at this point the version negotiation is done, we know we are at least on 5.2 and everyone has `backchainRequired = false` set (as it's a network wide flag), so you could just send/receive the collection of transactions and signatures without further wrapping.

blsemo (dismissed):
LGTM

blsemo (approved):
Re-approved

PR #5286, opened 2023-12-15
CORE-18973: Windows formatting fix in `AuthenticationProtocolInitiatorTest`

BenYip123:
Build to validate that `toAvro returns the correct object()` is fixed on a Windows `runtime-os` run: https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2FNightlys%2FCorda-Runtime-OS%2FWindows/detail/benyip%2FCORE-18973%2Fwindows-test-fix/4/tests
Build to validate that `toPem return a valid public key PEM()` is fixed on a Windows `runtime-os` run: https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2FNightlys%2FCorda-Runtime-OS%2FWindows/detail/benyip%2FCORE-18973%2Fwindows-test-fix/5/tests/

### DIFF
--- a/libs/p2p-crypto/src/test/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocolInitiatorTest.kt
+++ b/libs/p2p-crypto/src/test/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocolInitiatorTest.kt
@@ -40,7 +40,7 @@ cZyj8BXGy1y5rL75SE01/3qwAvSAatoTtVtPlPuHn5U1nGDe3AoaS1LvrqIypJjo
 ODGVZOuwAyZ6676J40yue06DiPpysNULHoZu3PEd+DEWKsCyPq3yYtB7E+HvLUAQ
 AQIDAQAB
 -----END PUBLIC KEY-----
-    """.trimIndent()
+    """
 
     private val publicKey by lazy {
         publicKeyPem.toPublicKey()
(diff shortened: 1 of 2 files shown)


libs/p2p-crypto/src/test/kotlin/net/corda/p2p/crypto/protocol/api/AuthenticationProtocolInitiatorTest.kt
```diff
@@ -67,7 +67,7 @@ AQIDAQAB
             assertThat(avro.protocolCommonDetails.sessionId).isEqualTo("sessionId")
             assertThat(avro.supportedModes).containsOnly(ProtocolMode.AUTHENTICATION_ONLY)
             assertThat(avro.groupId).isEqualTo("group")
-            assertThat(avro.ourPublicKey.trim()).isEqualTo(publicKeyPem.trim())
+            assertThat(avro.ourPublicKey.replace("\n", System.lineSeparator()).trim()).isEqualTo(publicKeyPem.trim())
```
ben-millar: Another option here is to convert the returned value to a `PublicKey` object so we can compare them cryptographically rather than doing a string comparison.
```suggestion
            assertThat(avro.ourPublicKey.toPublicKey()).isEqualTo(publicKey)
```
To do that, we can add a helper function to the bottom of this class:
```
private fun String.toPublicKey() : PublicKey =
        publicKeyFactory(this.reader())!!
```
The function above on line 44 should then become:
```
private val publicKey by lazy {
        publicKeyPem.toPublicKey()
    }
```

BenYip123: Added in commit #136720b

yift-r3 (approved):

PR #5288, opened 2023-12-15
CORE-18691: fix comment about what the item.first values are

dickon:


libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/CalculateLeveledHashes.kt
```diff
@@ -56,6 +52,14 @@ fun calculateLeveledHashes(proof: MerkleProof, digest: MerkleTreeHashDigestProvi
...
         while (index < nodeHashes.size) {
             val item = nodeHashes[index]
+            // We are at level $treeDepth from the top of the tree (where 1 is the root of the tree),
+            //     and at $index nodes from the left (counting from 0)
+            // $item is a pair of the left-most leaf index, shifted right by the number of levels we've gone, within that
+            // subtree and the hash at that leaf index.
+            //
+            // So if you take a tree slice at the level we are looking at ($currentSize from the top),
+            // the $item.first will be merkle tree position of the node. That will often be different from
+            // the index, when don't have the full tree structure at this level.
```
anien, line 62: `as we don't have` maybe ?

PR #5297, opened 2023-12-18
CORE-18524 - Add the state manager to the link manager

nikinagy:
Update to support passing state manager configuration to the link manager worker.

Verified with manual dummy test that we are able to write to state manager db:
Started combined worker and added state creation to later verify via checking the content of the DB
Screenshot 2023-12-15 at 13 06 02
Screenshot 2023-12-15 at 13 05 46


davidcurrie (dismissed):
Approved for @corda/cluster-management for changes to `/charts`

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/LinkManager.kt
```diff
@@ -47,11 +49,19 @@ class LinkManager(
 ) : LifecycleWithDominoTile {
 
     companion object {
+        private lateinit var _stateManager: StateManager
```
yift-r3: Why do we need this hack?

nikinagy: I wasn't sure how we will want to use the SM exactly, so I wanted to have an option where you could retrieve the instance from the LM, similarly to how generateKey() is used.

yift-r3: If you put it as a val in the common components, it will be available to everything in the link manager.

yift-r3 (dismissed):

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/LinkManager.kt
nikinagy: Good to know! I reverted this and added the sm into the common components as you suggested.

yift-r3 (approved):

yift-r3 (approved):

seanbrereton (approved):
LGTM (BLT)

PR #5299, opened 2023-12-19
ES-1659/disable notifications

seanbrereton:
disable slack notifications


ronanbrowne (approved):
LGTM - removing notifications from C5 channel as this is essentially a feature branch

PR #5303, opened 2023-12-19
CORE-16348: change to e2e utils

emilybowe:
Small change to test utils for name consistency


ben-millar (approved):
LGTM — I don't think this change has anything to do with the failing tests, but the naming makes the purpose of the field more clear.

PR #5304, opened 2023-12-19
CORE-18141 - Fix for tracing between HTTP RPC and Kafka

filipesoliveira:
Screenshots of the latency tree:

**Query Balance**
image

**Coin creation**
image


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/TracingUtils.kt
```diff
@@ -0,0 +1,38 @@
+package net.corda.messaging.utils
+
+import net.corda.messaging.api.mediator.MediatorMessage
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+internal object TracingUtils {
+
+    private val log: Logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+    private const val TRACING_HEADER_NAME_TRACING_ID = "X-B3-TraceId"
```
owenstanford: I think this code should live in the tracing lib, given the properties of a mediator are just a map we can probably add a generic utility function in the trace lib that can extract the headers from a Map<String, Any> to avoid leaking the tracing impl into the messaging lib.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -63,16 +65,38 @@ class RPCClient(
...
+            tracingHeaders.forEach { (k,v) -> headers[k] = v }
+            MediatorMessage(deserializedResponse, headers)
         }
     }
 
+    private inline fun<T> traceHttpSend(tracingHeaders: List<Pair<String, String>>, send: ()-> T): T {
+        val traceContext = traceSend(tracingHeaders, "http - send - clientId - $id")
+        return traceContext.markInScope().use {
+            try {
+                send()
```
owenstanford: You need a traceContext.finish() to close the span, I think this is why we are seeing this recoding as zero time in the trace.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -63,16 +65,38 @@ class RPCClient(
...
         return deserializedResponse?.let {
-            MediatorMessage(deserializedResponse, mutableMapOf("statusCode" to response.statusCode()))
+            val headers: MutableMap<String,Any> = mutableMapOf("statusCode" to response.statusCode())
+            tracingHeaders.forEach { (k,v) -> headers[k] = v }
+            MediatorMessage(deserializedResponse, headers)
         }
     }
 
+    private inline fun<T> traceHttpSend(tracingHeaders: List<Pair<String, String>>, send: ()-> T): T {
+        val traceContext = traceSend(tracingHeaders, "http - send - clientId - $id")
```
owenstanford: Rather than the client id, lets use the URL (can be read from the `request`) we can then easily see in the trace data what service is being called. If it makes the span name too long we can use the path only from the URL and drop the host as the paths identify the service e.g. `/token-selection`
It would also be good to add it as a tag so it's easily searchable in Grafana

vkolomeyko:
Looks fine to me in principle. I have created another PR into this branch with some minor changes: https://github.com/corda/corda-runtime-os/pull/5310

libs/tracing/src/main/kotlin/net/corda/tracing/TraceUtils.kt
```diff
@@ -0,0 +1,30 @@
+package net.corda.tracing
+
+object TraceUtils {
+
+    private const val TRACING_HEADER_NAME_TRACING_ID = "X-B3-TraceId"
+    private const val TRACING_HEADER_NAME_SPAN_ID = "X-B3-SpanId"
+    private const val TRACING_HEADER_NAME_SAMPLED = "X-B3-Sampled"
+    private const val TRACING_HEADER_NAME_PARENT_SPAN_ID = "X-B3-ParentSpanId"
```
josephzunigadaly: I think we should avoid building this ourselves. Instead we should call/add an overload to one of the many `TracingService.nextSpan` methods which uses an implementation of this from the brave library. 

The advantage of that is that the header names are maintained within the brave library and not here in our own code.

vkolomeyko: I think that would make sense if we expand `nextSpan` onto `MediatorMessage`.
That way we will have a new span every time we convert from Kafka `Record` to `MediatorMessage` or vise versa.

And we will not have to tinker with internal header constants.

filipesoliveira:
As requested, I will look into breaking this PR into smaller changes.

filipesoliveira:
I have reverted the refactoring that seemed unnecessary, @vkolomeyko and @josephzunigadaly.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -105,6 +131,9 @@ class RPCClient(
             builder.header(CORDA_REQUEST_KEY_HEADER, keyValue)
         }
 
+        // Once the HTTP request is created, it cannot be changed. So the builder as to be passed instead
```
vkolomeyko:
```suggestion
        // Once the HTTP request is created, it cannot be changed. So the builder has to be passed instead
```

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -105,6 +131,9 @@ class RPCClient(
...
+        addTraceContextToHttpRequest(builder)
```
vkolomeyko:
```suggestion
        builder.addTraceContext()
```

libs/tracing-impl/src/main/kotlin/net/corda/tracing/brave/BraveRecordExtractor.kt
```diff
@@ -0,0 +1,33 @@
...
+import brave.propagation.TraceContextOrSamplingFlags
+
+// This class is responsible for extracting a trace context based on a set of trace headers.
+class BraveRecordExtractor(tracing: Tracing) {
+
+    // Headers can be stored in any format. This instance contains the logic to extract the trace headers from a
+    // list
+    private val recordExtractorListFormat = tracing.propagation().extractor(
+            Propagation.Getter<List<Pair<String, String>>, String> { headers, key ->
+                headers.reversed().firstOrNull { it.first == key }?.second
```
vkolomeyko: Any particular reason why we are using `.reversed()`?
Ideally, there should be no duplicating `first`s in the list of pairs.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -63,13 +66,19 @@ class RPCClient(
...
+            val request = buildHttpRequest(message)
+            sendWithRetry(request)
+        }
 
         val deserializedResponse = deserializePayload(response.body())
 
-        return deserializedResponse?.let {
-            MediatorMessage(deserializedResponse, mutableMapOf("statusCode" to response.statusCode()))
+        // Convert the response to an instance of the MediatorMessage class and enrich the instance with a trace context
+        return deserializedResponse?.let{
```
vkolomeyko:
```suggestion
        return deserializedResponse?.let {
```

libs/tracing-impl/src/main/kotlin/net/corda/tracing/brave/BraveRecordExtractor.kt
filipesoliveira: This code block are existed and it was moved into this file. I don't know why it calls `reversed`.

libs/tracing/src/main/kotlin/net/corda/tracing/TracingService.kt
```diff
@@ -6,7 +6,32 @@ import java.util.concurrent.ExecutorService
 
 interface TracingService : AutoCloseable {
 
-    fun addTraceHeaders(headers: List<Pair<String, String>>): List<Pair<String, String>>
+    // The method takes the current headers and joins them with the trace headers
+    // A trace context provides the trace headers. Normally, the activate context is used, but it can be overridden
```
vkolomeyko:
```suggestion
    // A trace context provides the trace headers. Normally, the active context is used, but the headers can be overridden
```

Similar sort of change should be applied to all the methods below. Also, it would  probably makes sense to change the format to a KDoc rather than a comment.

libs/tracing-impl/src/main/kotlin/net/corda/tracing/brave/BraveRecordExtractor.kt
filipesoliveira: I have updated the code so an exception is raised if there are duplicated headers.

libs/tracing-impl/src/main/kotlin/net/corda/tracing/brave/BraveRecordExtractor.kt
```diff
@@ -0,0 +1,33 @@
...
+    )
+
+    // Headers can be stored in any format. This instance contains the logic to extract the trace headers from a
+    // map
+    private val recordExtractorMapFormat = tracing.propagation().extractor(
+        Propagation.Getter<Map<String, Any>, String> { headers, key -> headers[key] as String? }
+    )
+
+    // Extracts the trace context based on the headers provided
+    fun extract(headers: List<Pair<String, String>>): TraceContextOrSamplingFlags? {
```
vkolomeyko:
```suggestion
    fun extract(headers: List<Pair<String, String>>): TraceContextOrSamplingFlags {
```
Looking at the implementation of `B3Propagation` it can never return an `null` value.

libs/tracing-impl/src/main/kotlin/net/corda/tracing/brave/BraveRecordExtractor.kt
```diff
@@ -0,0 +1,33 @@
...
+        return recordExtractorListFormat.extract(headers)
+    }
+
+    // Extracts the trace context based on the headers provided
+    fun extract(headers: Map<String, Any>): TraceContextOrSamplingFlags? {
```
vkolomeyko:
```suggestion
    fun extract(headers: Map<String, Any>): TraceContextOrSamplingFlags {
```

libs/tracing-impl/src/main/kotlin/net/corda/tracing/brave/BraveRecordTracing.kt
```diff
@@ -2,17 +2,27 @@ package net.corda.tracing.brave
 
 import brave.Span
 import brave.Tracing
-import brave.propagation.Propagation
+import brave.propagation.TraceContextOrSamplingFlags
 import net.corda.messaging.api.records.EventLogRecord
 import net.corda.messaging.api.records.Record
 
-class BraveRecordTracing(tracing: Tracing) {
+class BraveRecordTracing(val tracing: Tracing) {
```
vkolomeyko:
```suggestion
class BraveRecordTracing(private val tracing: Tracing) {
```

vkolomeyko (dismissed):
LGTM

@josephzunigadaly - I would appreciate if you could have another look as well.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
```diff
@@ -88,6 +98,23 @@ class RPCClient(
         }
     }
 
+    private inline fun <T> traceHttpSend(traceHeaders: Map<String, Any>, uri: URI, send: () -> T): T {
+        val traceContext = traceSend(traceHeaders, "http client - send request - path - ${uri.path}")
```
josephzunigadaly: I think it is better to avoid including the path in the span name. The Grafana UI attempts to list these in a drop down. I don't know how many unique values can exist before it stops working. 

In our recent runs the list is not showing values so maybe it is already a problem? I see that we included client id in the Kafka version of this.

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/CordaKafkaProducerImpl.kt
```diff
@@ -158,7 +158,7 @@ class CordaKafkaProducerImpl(
         partition: Int?,
         callback: CordaProducer.Callback?
     ) {
-        val traceContext = traceSend(record.headers, "send $clientId")
+        val traceContext = traceSend(record.headers, "kafka - send - clientId: $clientId")
```
josephzunigadaly: We should probably remove clientId from this and include it as a tag if it is useful.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
filipesoliveira: This code was added to improve readability otherwise it is hard to reason about the the latency tree without having to click on each span for more details

**Example with the code change:**
image

**Example without the code change:**
image

libs/messaging/kafka-message-bus-impl/src/main/kotlin/net/corda/messagebus/kafka/producer/CordaKafkaProducerImpl.kt
filipesoliveira: I will remove the `clientId`. For now, I don't see the value of adding the `clientId ` as tag so I won't create one.

filipesoliveira:
I have updated the code once more so the span generated by the webserver also contains the url path. Before this change only the http method would be added to the span name.

Example after the code change:
image

filipesoliveira:
I would like to conduct further testing.

filipesoliveira:
After applying the most recent changes:
image

filipesoliveira:
More details are provided when a Kafka record is sent to a topic:
image

image

filipesoliveira:
More details are provided when a Kafka record is sent to a topic:
image

image

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
josephzunigadaly: I can see how that looks better. Do we ever include parameters in the internal requests? Maybe if they are like the ones in the screenshot it's a small enough set of internal paths that we could include them?

The Kafka client ID doesn't seem very readable, I think it was a good move to remove that one.

vkolomeyko (dismissed):
LGTM

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/RPCClient.kt
filipesoliveira: I think we use the body of the http request to send data instead of http parameters. The few http parameters that are used are already displayed in the name of the span.

libs/tracing/src/main/kotlin/net/corda/tracing/TraceThis.kt
```diff
@@ -44,18 +46,72 @@ fun getOrCreateBatchPublishTracing(clientId: String): BatchPublishTracing {
...
+            record.headers,
+            emptyList() // Don't override the current trace context
+        )
+    )
+}
+
+fun addTraceContextToRecord(
+    record: Record<*, *>,
+    traceHeadersToOverrideContext: Map<String, Any> = emptyMap() // By default, don't override the current trace context
+): Record<out Any, out Any> {
```
josephzunigadaly: If we make this change we can remove the cast at the call site:

```suggestion
fun <K : Any, E : Any> addTraceContextToRecord(
    record: Record<K, E>,
    traceHeadersToOverrideContext: Map<String, Any> = emptyMap() // By default, don't override the current trace context
): Record<K, E> {
```

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessor.kt
```diff
@@ -71,24 +72,32 @@ class EventProcessor<K : Any, S : Any, E : Any>(
...
+                // Http - Send the request immediately. Once the response arrives convert it to a kafka record and
+                // add it to the queue, so it can be processed in due course
                 @Suppress("UNCHECKED_CAST")
                 val reply = with(destination) {
                     message.addProperty(MessagingClient.MSG_PROP_ENDPOINT, endpoint)
                     client.send(message) as MediatorMessage<E>?
                 }
                 if (reply != null) {
+                    // Convert response to a record and add it to the queue
+                    @Suppress("UNCHECKED_CAST")
```
josephzunigadaly: With the suggested change to addTraceContextToRecord we can remove the cast:

```suggestion
                    // Convert response to a record and add it to the queue
                    
```

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessor.kt
```diff
@@ -71,24 +72,32 @@ class EventProcessor<K : Any, S : Any, E : Any>(
...
-                            reply.payload,
-                        )
+                        addTraceContextToRecord(
+                            Record(
+                                "",
+                                event.key,
+                                reply.payload
+                            ),
+                            message.properties
+                        ) as Record<K, E>
```
josephzunigadaly: With the suggested change to addTraceContextToRecord we can remove the cast:

```suggestion
                        )
```

libs/tracing-impl/src/main/kotlin/net/corda/tracing/brave/BraveTracingService.kt
```diff
@@ -112,7 +112,17 @@ internal class BraveTracingService(serviceName: String, zipkinHost: String?, sam
         .putRule(and(methodEquals("POST"), pathStartsWith("/api/v1/flow")), sampler(samplesPerSecond))
         .putRule(pathStartsWith("/"), sampler(samplesPerSecond)).build()
 
-    private val httpTracing by lazy { HttpTracing.newBuilder(tracing).serverSampler(serverSampler).build() }
+    private val httpTracing by lazy {
+        HttpTracing.newBuilder(tracing)
+            .serverRequestParser(
+                object : HttpRequestParser.Default() {
+                    override fun spanName(req: HttpRequest?, context: brave.propagation.TraceContext?): String? {
+                        return "http server - ${req?.method()} - ${req?.path()}"
```
josephzunigadaly, line 120: This is another instance of including path in our span names. These paths include resource IDs that expand the number of unique names in the data. I'm still concerned that Grafana tries to list all of these in a drop down and that drop down has stopped showing any options.

It would be better to include the template path as seen in the REST annotations. I'm not sure how we would get that, but if it were possible, that would give us the benefit this provides while keeping the list of unique names small.

josephzunigadaly (approved):

vkolomeyko (approved):
LGTM

mbrkic-r3 (approved):
LGTM

libs/tracing-impl/src/main/kotlin/net/corda/tracing/brave/BraveTracingService.kt
filipesoliveira: As discussed in the stand-up, the possible issue highlighted can be addressed in a different PR since the tracing feature will evolve over time.

PR #5311, opened 2023-12-20
CORE-19064: Create Endpoints to Return schema creation/migration SQL

BenYip123:
Create endpoints to return schema creation/migration SQL and update swagger baseline

### DIFF
--- a/libs/virtual-node/virtual-node-endpoints/src/main/kotlin/net/corda/libs/virtualnode/endpoints/v1/VirtualNodeRestResource.kt
+++ b/libs/virtual-node/virtual-node-endpoints/src/main/kotlin/net/corda/libs/virtualnode/endpoints/v1/VirtualNodeRestResource.kt
@@ -127,29 +127,47 @@ interface VirtualNodeRestResource : RestResource {
     ): AsyncOperationStatus
 
     @HttpGET(
-        path = "db/{dbTypeOrCpiChecksum}",
-        title = "Return creation schema SQL",
-        description = "Returns the SQL needed for intention to create a virtual node or latest uploaded CPI.",
-        responseDescription = "SQL needed to bring schema up to date.",
+        path = "create/db/vault/crypto",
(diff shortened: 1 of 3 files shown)


libs/virtual-node/virtual-node-endpoints/src/main/kotlin/net/corda/libs/virtualnode/endpoints/v1/VirtualNodeRestResource.kt
```diff
@@ -126,6 +126,32 @@ interface VirtualNodeRestResource : RestResource {
...
+    @HttpGET(
+        path = "db/{dbTypeOrCpiChecksum}",
+        title = "Return creation schema SQL",
+        description = "Returns the SQL needed for intention to create a virtual node or latest uploaded CPI.",
+        responseDescription = "SQL needed to bring schema up to date.",
+        minVersion = RestApiVersion.C5_2
+    )
+    fun getCreateSchemaOrCpiSQL(
+        @RestPathParameter(description = "Virtual node database type or CPI checksum")
+        dbTypeOrCpiChecksum: String,
```
vkolomeyko: Hm... I am not sure about `dbTypeOrCpiChecksum`...

When we are talking about DB Type, I am assuming we are have `VAULT`, `CRYPTO` and `UNIQUENESS` in mind?

in which case, would two separate parameters make more sense like:
```
dbType: RestVirtualNodeDbTypeEnum,
cpi: CpiIdentifier
```
?

BenYip123: Yes, VAULT, CRYPTO and UNIQUENESS is what I had in mind. I think the problem is that if we want to specify a CPI to get the SQL of, we can't have any dbType in that case

vkolomeyko (approved):
LGTM

PR #5312, opened 2023-12-20
CORE-19015: Close in memory topic after usage

yift-r3:
With this change, the in-memory message bus tests can stop after the test has finished. This will prevent messages from being repeatedly sent to closed consumers.

This is more stable than before on Windows. See https://ci02.dev.r3.com/job/Corda5/job/Nightlys/job/Corda-Runtime-OS/job/Windows/job/yift%2Fcore-19015%2Fclose-in-mem/


nikinagy (approved):
LGTM

PR #5322, opened 2023-12-21
CORE-18553: Send messages to the link manager

yift-r3:
Forward messages from the Gateway to the link manager via HTTP instead of Kafka.

API pull request is in https://github.com/corda/corda-api/pull/1424


components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/internal/InboundMessageHandler.kt
```diff
@@ -136,6 +137,40 @@ internal class InboundMessageHandler(
...
+    private fun forwardMessage(
+        httpWriter: HttpWriter,
+        requestSource: SocketAddress,
+        messageId: String,
+        p2pMessage: LinkInMessage,
+    ): HttpResponseStatus {
+        val payload = try {
+            linkManagerClient.send(p2pMessage)
+        } catch (e: Exception) {
+            logger.warn("could not forward message to the link managed.", e)
```
nikinagy: would it be beneficial/useful to log out what message (an ID or something) we were unable to forward?

dimosr:
```suggestion
            logger.warn("could not forward message to the link manager.", e)
```

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/internal/InboundMessageHandler.kt
```diff
@@ -136,6 +137,40 @@ internal class InboundMessageHandler(
         }
     }
 
+    private fun forwardMessage(
```
dimosr: I think it's better to move this function further below (after `handleRequest`, where it's invoked), so that we follow the call hierarchy of the various functions. It makes it easier to read that way.

components/gateway/src/main/kotlin/net/corda/p2p/gateway/messaging/internal/LinkManagerRpcClient.kt
```diff
@@ -27,7 +27,7 @@ internal class LinkManagerRpcClient(
         URI.create("http://$endpoint/api/$platformVersion$P2P_LINK_MANAGER_PATH")
     }
 
-    fun send(message: LinkInMessage): AuthenticatedDataMessage? {
-        return client.send<AuthenticatedDataMessage>(url, message)
+    fun send(message: LinkInMessage): GatewayResponse? {
```
dimosr: that seems a bit strange, this is a type supposed to be returned by the gateway on requests, not the link manager. Shouldn't the link manager return the appropriate type (either re-use an existing one if appropriate, or introduce a new one) and then the gateway convert it into a `GatewayResponse`?

yift-r3: The link manager can return either an `AuthenticatedEncryptedDataMessage` or a `AuthenticatedDataMessage` and the `GatewayResponse` has a union of those classes...  We can create a new class just for this...

dimosr: hm, what about the other session messages (e.g. `InitiatorHelloMessage` etc.). Apart from session negotiation, these also play the role of ACKs for session message, so would we want to include those in this mechanism or not?

yift-r3: The current plan is that only data acks will be returned directly.

dimosr: what is the reasoning for not doing the same for session messages too? Seems like a positive change, no?

yift-r3: As far as I remember:
1. The data messages acks are part of the critical path, the session messages are only when we set up the session.
2. We want to handle the ack messages in batches.
3. The link manager handling of the acks should be straightforward, so ideally the HTTP queue will not wait for any computation.

dimosr: I think points 1 and 2 are about why we do it for acks, rather than why we don't it for session messages. For instance, doing it for session messages doesn't prevent us from batching only the ACKs, and not batching the session messages, if we wanted. 

But, point 3 is a valid one. Given the session might not be cached during session establishment, it might have some negative side-effects. We could skip it for the first version of it then. But, later on might be worth doing a a PoC and comparing because skipping the hops over Kafka might mean it would actually be better overall to return session messages as well (since the slow down from loading the session data from the DB would be paid even when going over Kafka).

dimosr: To respond to the original question, we should probably introduce another type yes.

dimosr (dismissed):
LGTM - just a minor comment on the API PR

dimosr (approved):

PR #5323, opened 2023-12-21
CORE-18921: Apply auto format on network worker code

yift-r3:
Apply `ktlint` auto format on the network worker code.


nikinagy (approved):
LGTM

PR #5325, opened 2023-12-21
CORE-18961 Thread safe SandboxServiceImpl

simon-johnson-r3:
When the sandbox cache is stressed it tries to load and unload OSGi bundles concurrently. This is because the `close()` block is executed in whatever thread happens to trigger a purge, whilst another thread can be creating a new sandbox at the same time. Both of these operations interact with `SandboxServiceImpl` which is our abstraction for OSGi bundle interaction. This service contains a number of maps and sets which would be manipulated concurrently at this point, and this is not thread safe.

I could reproduce this consistently by reducing the sandbox cache size to 1. With the introduction of a small amount of extra logging I could see the concurrent creation/closing of sandboxes from multiple threads followed immediately by a "bad wiring" exception from OSGi as it then tried to process the flow message which resulted in a new sandbox in a bundle which was incorrectly configured.

Making the `SandboxServiceImpl` thread safe resulted in the "bad wiring" errors going away, whilst many instances of unloading of bundles could still be observed in the log (this was whilst sandbox cache size was still forced to 1).

### DIFF
--- a/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxServiceImpl.kt
+++ b/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxServiceImpl.kt
@@ -1,4 +1,5 @@
 @file:JvmName("SandboxServiceUtils")
+
 package net.corda.sandbox.internal
 
 import net.corda.crypto.core.SecureHashImpl
@@ -30,6 +31,7 @@ import java.security.DigestInputStream
 import java.security.MessageDigest
 import java.security.PrivilegedAction
 import java.util.UUID
+import java.util.concurrent.ConcurrentHashMap
 import kotlin.streams.asSequence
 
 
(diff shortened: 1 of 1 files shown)


vkolomeyko (approved):
LGTM

ben-millar (approved):
LGTM. It might be worse investigating the performance impact of a straight swap to concurrent hash maps in the future—Maybe we can come up with a more bespoke synchronisation mechanism for our use case/access patterns, if the sandboxes are being hit harder these days.

simon-johnson-r3:
> LGTM. It might be worse investigating the performance impact of a straight swap to concurrent hash maps in the future—Maybe we can come up with a more bespoke synchronisation mechanism for our use case/access patterns, if the sandboxes are being hit harder these days.

This is why I chose ConcurrentHashMaps as 'get's are generally lock free. The get operations here are used by OSGi hooks to ensure code doesn't escape the sandbox, so they are more time critical than the puts/removes, which are only done on sandbox creation or removal but not code execution within the sandbox. Anecdotally (e.g. https://stackoverflow.com/questions/1378310/performance-concurrenthashmap-vs-hashmap) gets are as good as free, and adds are more expensive, but we're talking about adding OSGi bundles here, not huge data sets, so this is possibly a handful of adds per sandbox creation only.

I'd be surprised if we'd be able to beat the Java implementation with our own. You never know, but it would be more important to find out where the biggest bottleneck lies across the entire creation of sandboxes first. I could for instance just as easily ask whether the original maps returned by `mutableMapOf` were slower than some implementation of our own. The issue of threading is rather incidental to focus on, because we always need it, we can never return to non thread safe containers.

We do have ongoing work to profile sandbox creation.

PR #5326, opened 2023-12-21
CORE-19064: Fix endpoint create crypto SQL path

BenYip123:
The path was incorrectly created as `"create/db/vault/crypto"`. It has been changed to `"create/db/crypto"`


vkolomeyko (approved):
LGTM

PR #5329, opened 2023-12-21
CORE-19096: Changes to `ConsumerGroup`

vkolomeyko:
Only affects DB flavour of the message bus which is used for Combined Worker


LWogan (approved):
Makes sense to me

charlieR3 (approved):

PR #5330, opened 2023-12-21
CORE-19067: Check that uploaded CPIs don't have a minimum version higher than the running version of Corda

malachyb:


josephzunigadaly (dismissed):

components/chunking/chunk-db-write-impl/src/main/kotlin/net/corda/chunking/db/impl/validation/ValidationFunctions.kt
```diff
@@ -68,9 +68,9 @@ fun assembleFileFromChunks(
  *
  * @throws ValidationException
  */
-fun FileInfo.validateAndGetCpi(cpiPartsDir: Path, requestId: String): Cpi {
+fun FileInfo.validateAndGetCpi(cpiPartsDir: Path, requestId: String, activeVersion: Int): Cpi {
```
Sakpal: Major: The `activeVersion` does not gives a clear indication of which version are referencing it to? Can you please update the variable name which indicates the version is related to running Corda Platform

Sakpal (dismissed):
LGTM

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/CpiReader.kt
```diff
@@ -26,7 +26,8 @@ object CpiReader {
         inputStream: InputStream,
         expansionLocation: Path,
         cpiLocation: String? = null,
-        verifySignature: Boolean = jarSignatureVerificationEnabledByDefault()
+        verifySignature: Boolean = jarSignatureVerificationEnabledByDefault(),
+        activeCordaPlatformVersion: Int
```
kyriathar: Would it make sense to make this a property to `CpiReader` if it's to remain same and shared across `readCpi` calls? Then perhaps we can add a `CpiReader` instance at `CpiValidatorImpl` constructor instead and cache it there.

And perhaps we could do the same for `CpiLoaderV2` and have it as property to `CpiReader`? Instead of just re-creating on every `readCpi` call now.

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
```diff
@@ -59,6 +60,14 @@ class CpiLoaderV2(private val clock: Clock = UTCClock()) : CpiLoader {
...
             }
 
+            cpks.forEach {
+                val minPlatformVersion = it.metadata.cordappManifest.minPlatformVersion
+                if (activeCordaPlatformVersion < minPlatformVersion) {
+                    throw PackagingException("Platform version of Corda is lower than minimum platform version of CPK" +
+                            " ${it.metadata.cpkId.name} ($activeCordaPlatformVersion < $minPlatformVersion)")
+                }
+            }
+
```
kyriathar, line 69: This is much cleaner than what we did for uniqueness check for cpk cordapp name where we baked that in `readCpksFromCpb`. We should consider doing the same for that as there is not overhead as well. But not under this PR.

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/CpiReader.kt
malachyb: Good point, I'll do that

Sakpal (dismissed):
LGTM

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
```diff
@@ -22,7 +22,7 @@ import java.nio.file.Path
 import java.nio.file.Paths
 import java.util.jar.JarInputStream
 
-class CpiLoaderV2(private val clock: Clock = UTCClock()) : CpiLoader {
+class CpiLoaderV2(private val clock: Clock = UTCClock(), private val activeCordaPlatformVersion: Int? = null) : CpiLoader {
```
kyriathar: Why is this needed to be nullable?

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/CpiReader.kt
```diff
@@ -11,6 +11,11 @@ import java.util.jar.JarInputStream
 
 object CpiReader {
     private val version2 = CpkFormatVersion(2, 0)
+    private var activeCordaPlatformVersion = 0
+
+    fun setActiveCordaPlatformVersion(value: Int) {
```
kyriathar: I think this should rather be initialized through a constructor.

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/internal/v2/CpiLoaderV2.kt
malachyb: I had it nullable in case there were other uses of the class that I hadn't spotted that didn't set the value so that it would still work in that situation.

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/CpiReader.kt
malachyb: Since it's an `object` I don't think it has a constructor

kyriathar: Yes we need to make it a class.

libs/packaging/packaging/src/main/kotlin/net/corda/libs/packaging/CpiReader.kt
```diff
@@ -9,8 +9,9 @@ import java.io.InputStream
 import java.nio.file.Path
 import java.util.jar.JarInputStream
 
-object CpiReader {
+class CpiReader(activeCordaPlatformVersion: Int) {
     private val version2 = CpkFormatVersion(2, 0)
+    private val cpiLoader = CpiLoaderV2(activeCordaPlatformVersion = activeCordaPlatformVersion)
```
kyriathar: nit: specifying the named parameter here doesn't add much since it matches the argument name.

kyriathar (dismissed):

kyriathar (commented):
We should also be having `CpiReader` as property to `CpiValidatorImpl` I think to save re-creating it per request.

kyriathar (approved):

PR #5333, opened 2023-12-22
CORE-18969 add a new flow to the test cordapp which will trigger a flow session timeout

LWogan:
We should take an action into next year to discuss if this repo is the best place for manual test cordapps going forward


LWogan:
build e2e

thiagoviana (approved):

vikram-r3 (approved):
LGTM !!

PR #5334, opened 2023-12-22
CORE-18908: Remvoe notary key ID from member context

yift-r3:


components/membership/registration-impl/src/test/kotlin/net/corda/membership/impl/registration/dynamic/member/DynamicMemberRegistrationServiceTest.kt
```diff
@@ -1499,6 +1497,37 @@ class DynamicMemberRegistrationServiceTest {
             }
         }
 
+        @Test
+        fun `re-registration with previous that contains the notary key ID won't fail`() {
```
nikinagy:
```suggestion
        fun `re-registration with previous context that contains the notary key ID won't fail`() {
```

nikinagy (dismissed):
LGTM just added a nit that can be ignored

nikinagy (approved):
LGTM

PR #5335, opened 2023-12-22
CORE-18961 explicit OSGi locking

simon-johnson-r3:
Added explicit locking to creation and removal methods to ensure maps/sets are updated atomically, but also OSGi bundle loading and unloading is not done concurrently. Methods which only read the already thread safe containers are not touched to avoid adding overhead.

### DIFF
--- a/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxServiceImpl.kt
+++ b/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxServiceImpl.kt
@@ -32,6 +32,8 @@ import java.security.MessageDigest
 import java.security.PrivilegedAction
 import java.util.UUID
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.concurrent.withLock
 import kotlin.streams.asSequence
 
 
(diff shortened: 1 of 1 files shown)


libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxServiceImpl.kt
```diff
@@ -50,8 +52,11 @@ internal class SandboxServiceImpl @Activate constructor(
         )
 
     /**
-     * We use concurrent hash maps everywhere (including sets) to allow concurrent access whilst keeping gets lock free
+     * We use concurrent hash maps everywhere (including sets) to allow concurrent access whilst keeping gets lock free.
+     * We also employ a lock to ensure writes are done to every map/set group atomically and we don't try to install
+     * and uninstall bundles from OSGi concurrently.
      */
+    val bundleLock = ReentrantLock()
```
vkolomeyko, line 59: Would it not better to use:
```suggestion
    private val bundleLock = ReentrantReadWriteLock()
```
?

`areInSameSandbox` is the read-only operation and read aspect of the lock can be used to allow operations to progress in parallel.

vkolomeyko:
```suggestion
    private val bundleLock = ReentrantLock()
```

vkolomeyko (dismissed):
LGTM

Only minor change to make lock `private`.

vkolomeyko (approved):
LGTM

PR #5337, opened 2023-12-28
CORE-18532: Retry Transient Errors in Flow Engine

jujoramos:
Instead of automatically publishing the original event back to the
"flow.event" topic whenever a transient exception occurs when executing
the flow pipeline, automatically retry the exceptions using an
exponential backoff retry mechanism and permanently fail the flow if
the configured time or attemps is reached.

- Create internal retry utility to manage retries in an automated
  fashion with plugable backoff strategy (constant, linear and
  exponential provided out of the box).
- Replace existing utilities in the crypto components and tests with
  the new one and delete unused classes.
- Update FlowEventProcessor to automatically retry transient exceptions
  through the new retry utility using an exponential backoff with a
  growth factor of 250ms.
- Remove unnecessary code and update tests to accommodate for the new
  internal retry.


libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
```diff
@@ -0,0 +1,70 @@
...
+            if ((delayMillis < 0) || (attempt >= maxRetries) || ((elapsedMillis + delayMillis) >= maxTimeMillis)) {
+                ("Execution failed with \"${originalException.message}\" after retrying $attempt times for " +
+                    "$elapsedMillis milliseconds.").also {
+                    logger.warn(it, originalException)
+                    throw exceptionProvider(it, originalException)
+                }
+            }
+
+            attempt++
+            logger.warn(
```
dickon: So we've gone from info to warn in the event of retries on crypto code, is that intentional? I'm not sure we should be warning for all kinds of transient error, could be quite noisy in the log.

jujoramos: > So we've gone from info to warn in the event of retries on crypto code, is that intentional?

What do you mean 🤔 ?. The actual `CryptoRetryingExecutor` implementation [logs a message under `INFO`](https://github.com/corda/corda-runtime-os/blob/release/os/5.2/libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt#L36) for every retry and [another one under `WARN`](https://github.com/corda/corda-runtime-os/blob/release/os/5.2/libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt#L62) for every failed attempt, the refactor keeps one logged under `WARN` but removes the one logged under `INFO`, effectively reducing the noise.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/BackoffStrategy.kt
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Strategy to provide backoff delays when handling transient failures.
+ */
+interface BackoffStrategy {
+    /**
+     * Calculate next wait period in milliseconds for the given attempt.
+     *
+     * @param attempt current failed attempt, starts at 1.
+     * @return wait time, in milliseconds, for the given attempt number. A negative value forces the retry to stop.
```
dickon, line 13: I think that implicit struture that negative values means don't retry is an implicit structure hack and  `ULong?` is preferable.

dickon: ... which would also meaning using `ULong` throughout the change rather than `Long`, which I think is clearer.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
dickon: Okay so maybe INFO+WARN on failure was too much. There's no way here to say that the retries aren't newsworthy, and to let the code using retries suppress the warning. Maybe best to have the callers specify a callback which is called on warnings with a default implementation that warns, so that it can be more useful? 

Probably best if the calling code does nothing that we do produce some noise on retry, making it silent by default feels like it could cause us problems.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/BackoffStrategy.kt
```diff
@@ -0,0 +1,44 @@
...
+     */
+    fun delay(attempt: Int): Long
+}
+
+/**
+ * Constant backoff strategy, delay between retries remains constant between attempts.
+ *
+ * @param delay the constant delay to be applied on each attempt.
```
dickon: What is the unit of time?

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/BackoffStrategy.kt
dickon, line 1: I'm happy to see this go, it never felt like something that should be crypto specific.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoBackoffStrategy.kt
```diff
@@ -0,0 +1,33 @@
+package net.corda.crypto.impl.retrying
+
+import net.corda.utilities.retry.BackoffStrategy
+
+class CryptoBackoffStrategy(
```
dickon: There's nothing specially crypto-relevant about this.  I think it'd be better to have a generic `BackoffStaretgy` which allows a list of times to be specified.  

It's interesting and perhaps a questionable choice that we allow so much configuration of retry timings in the crypto config, but we need to keep compatibility with that.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt
```diff
@@ -29,50 +20,15 @@ open class CryptoRetryingExecutor(
...
-                return result
-            } catch (e: Throwable) {
-                if (!e.isRecoverable()) {
-                    // the exception is not recoverable, no point in retrying
-                    logCompleteFailure(attempt, op)
-                    // throws the original exception
-                    throw e
-                }
-                val backoff = strategy.getBackoff(attempt)
-                if (backoff < 0 || attempt > CRYPTO_MAX_ATTEMPT_GUARD) {
```
dickon, line 52: We've lost this behaviour, which is we previously ignore a  config request to have more than 10 retries and stick at 10. That could matter and would be worth release notes.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/BackoffStrategy.kt
```diff
@@ -0,0 +1,44 @@
...
+    override fun delay(attempt: Int) = growthFactor * attempt
+}
+
+/**
+ * Exponential backoff strategy, the delay between retries increases exponentially with each attempt.
+ *
+ * @param base the base value for exponential growth.
+ * @param growthFactor the multiplier to scale the exponential growth.
+ */
+class Exponential(private val base: Double = 2.0, private val growthFactor: Long = 1000L) : BackoffStrategy {
```
dickon, line 57: This is probably the only one we actually need. Linear retries are usually problematic due to certain failure modes and unless we have a clear need for.

But it should be a random starting point not a fixed number, to prevent lining up failures between different instances.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
dickon: actually no, the original code was no logging on the first attempt, then warn, sleep, then an info before trying again, which is probably good (with the proviso that user code should be able to disable both bits of logging, e.g. with callbacks).

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt
```diff
@@ -29,50 +20,15 @@ open class CryptoRetryingExecutor(
...
+        ) {
+            block()
         }
     }
-
-    private fun logCompleteFailure(attempt: Int, opId: String) {
-        logger.warn("Failed to execute (opId={},attempt={})", opId, attempt)
-    }
-
-    protected open fun <R> execute(block: () -> R): R = block()
```
kyriathar, line 77: Good thing to get rid of this open `execute` function I think. Can't really see what else it would do.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt
```diff
@@ -29,50 +20,15 @@ open class CryptoRetryingExecutor(
...
-                }
-                val result = execute(block)
-                if (attempt > 1) {
-                    logger.info("Retrying was successful (op={},attempt={})", op, attempt)
-                }
-                return result
-            } catch (e: Throwable) {
-                if (!e.isRecoverable()) {
-                    // the exception is not recoverable, no point in retrying
-                    logCompleteFailure(attempt, op)
```
dickon, line 47: This call calls `logger.warn` in the case of non-recoverable errors, which you've lost, and that might have been a useful clue and should probably be on by default. But I would ideally like it suppressable.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
```diff
@@ -0,0 +1,70 @@
...
+    var attempt = 1
+    var originalException: Throwable?
+    val startTime = System.currentTimeMillis()
+
+    while (true) {
+        try {
+            return operation()
+        } catch (throwable: Throwable) {
+            if (!recoverable(throwable) || (maxRetries <= 0) || (maxTimeMillis <= 0)) {
+                throw throwable
```
dickon, line 68: I think not logging anything here by default is probably bad. In the case of the crypto code we would log a warning at this point, which saves from having to do it further up the stack. If we do that it should be suppresssable though, so the caller can choose to handle without noise in the log.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
```diff
@@ -0,0 +1,70 @@
...
+ *  @throws RetryException, or the exception type created through [exceptionProvider], if the original exception was
+ *   marked as retryable but all retry attempts failed.
+ */
+@Suppress("LongParameterList")
+fun <T> tryWithBackoff(
+    logger: Logger,
+    maxRetries: Long,
+    maxTimeMillis: Long,
+    backoffStrategy: BackoffStrategy,
+    recoverable: (Throwable) -> Boolean = { _ -> false },
```
dickon: The callback could usefully take the number of repeats and the time taken so far as argumenets.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt
```diff
@@ -29,50 +20,15 @@ open class CryptoRetryingExecutor(
...
-                    // sleep for a little while and then retry
-                    Thread.sleep(backoff)
-                }
-            }
+        return tryWithBackoff(
+            logger = logger,
+            maxRetries = strategy.maxRetries,
+            maxTimeMillis = Long.MAX_VALUE,
+            backoffStrategy = strategy,
+            recoverable = Throwable::isRecoverable,
```
dickon: I think you've correctly ported the current behaviour with isRecoverable, which is the best choice right now. Not that I think a binary classification of exceptions as recoverable or not is a great long-term choice, I think it should be down to particular caling sites, but what you've done allows us to do that later.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
```diff
@@ -0,0 +1,70 @@
...
+                throw throwable
+            }
+
+            originalException = throwable
+            val delayMillis = backoffStrategy.delay(attempt)
+            val elapsedMillis = System.currentTimeMillis() - startTime
+
+            if ((delayMillis < 0) || (attempt >= maxRetries) || ((elapsedMillis + delayMillis) >= maxTimeMillis)) {
+                ("Execution failed with \"${originalException.message}\" after retrying $attempt times for " +
+                    "$elapsedMillis milliseconds.").also {
```
kyriathar: nit: could we please have this without `also`. 

Also do we need the exception message in that text since we accompany the error text with the exception?

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoBackoffStrategy.kt
```diff
@@ -0,0 +1,33 @@
...
+    maxAttempts: Int,
+    inputDelays: List<Long>,
+) : BackoffStrategy {
+    val maxRetries = maxAttempts.toLong()
+    private val delays: Array<Long> = createBackoffArray(maxAttempts, inputDelays)
+
+    private fun createBackoffArray(maxAttempts: Int, backoff: List<Long>): Array<Long> = when {
+        maxAttempts <= 1 -> emptyArray()
```
dickon: I know you're just porting the old logic, but I don't think we need it, and if we can get a generic solution that wouldn't need this complexity of implementation.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoBackoffStrategy.kt
kyriathar: Should we use a different name for this one. Like what it actually is like the rest subclasses below. It is only used in crypto code but.

dickon (commented):
I added a bunch of comments, which need not blocking merging so if you want to disregard them I can approve anyway.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoBackoffStrategy.kt
```diff
@@ -0,0 +1,33 @@
+package net.corda.crypto.impl.retrying
+
+import net.corda.utilities.retry.BackoffStrategy
+
+class CryptoBackoffStrategy(
+    maxAttempts: Int,
+    inputDelays: List<Long>,
```
kyriathar: nit: should we align this var name with `backoff` below.

kyriathar (commented):
Some minor comments. The crypto bits look good to me in principle.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt
jujoramos: Agreed, but should we even bother to keep the old behaviour in place?. What's the point of allowing users to set something higher than `10` if we are going to silently ignore it underneath?. Instead of that, I've just opened a `PR` to explicitly set the limits within the configuration directly: https://github.com/corda/corda-api/pull/1430.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/BackoffStrategy.kt
jujoramos: 👍

libs/utilities/src/main/kotlin/net/corda/utilities/retry/BackoffStrategy.kt
jujoramos: Will change the KDocs and update the `PR` shortly, thanks.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoBackoffStrategy.kt
kyriathar: (this is related to https://github.com/corda/corda-runtime-os/pull/5337#discussion_r1439445237)

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
```diff
@@ -0,0 +1,70 @@
...
+    backoffStrategy: BackoffStrategy,
+    recoverable: (Throwable) -> Boolean = { _ -> false },
+    exceptionProvider: (String, Throwable) -> CordaRuntimeException = { m, t -> RetryException(m, t) },
+    operation: () -> T,
+): T {
+    var attempt = 1
+    var originalException: Throwable?
+    val startTime = System.currentTimeMillis()
+
+    while (true) {
```
sprintengine, line 62: I wonder if we'd be better making this work using kotlin `repeat(maxRetries)` API?

jujoramos: It'll likely make the implementation harder to understand as the block should break out of the `repeat` loop if the call succeeds, which represents the general or expected usage pattern? 🤔.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
sprintengine: we probably should log at a lower log level, debug or trace maybe?

libs/utilities/src/main/kotlin/net/corda/utilities/retry/BackoffStrategy.kt
```diff
@@ -0,0 +1,44 @@
+package net.corda.utilities.retry
```
sprintengine, line 1: I could see this utility being used in a lot of places.

What's the footprint of someone adding `utilities` as a dependency? I see api dependencies on corda-application, corda-serialization, base-internal, seems a bit heavy for someone to import serialization APIs etc to use only retry, I wonder if we should move this into its own module?

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
dickon: There's nothing wrong with `also` here, I think it's good style.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/BackoffStrategy.kt
jujoramos: I'll keep the implementation as it is for the time being, using `ULong` and `UInt` is a good improvement for the future.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/BackoffStrategy.kt
jujoramos: > This is probably the only one we actually need. Linear retries are usually problematic due to certain failure modes and unless we have a clear need for.

👍 

> But it should be a random starting point not a fixed number, to prevent lining up failures between different instances.

Agreed, but it's up to the user of the utility to set the randomness of the starting point.

driessamyn:
Did you consider using an off-the-shelf library like [Resilience4J](https://resilience4j.readme.io/docs) rather than rolling our own? (not saying we should, but we should evaluate pros/cons).

Also, I thought we'd look at moving away from crypto doing its own bespoke thing and consolidate how we do retries etc with the RPC pattern?

jujoramos:
> Did you consider using an off-the-shelf library like [Resilience4J](https://resilience4j.readme.io/docs) rather than rolling our own? (not saying we should, but we should evaluate pros/cons).

I did at the beginning but our use case is quite simplistic at the moment, adding `resilience4j` felt like killing mosquitos with a cannon... not the most technically valid reason in the world 🤷‍♂️ .

> Also, I thought we'd look at moving away from crypto doing its own bespoke thing and consolidate how we do retries etc with the RPC pattern?

Not sure what the background context is here, I've simply created a retry utility to be shared across the code base and refactored the existing crypto retry mechanism to use the new one. Any other work to replace by RPC pattern should likely be its own epic/task, not included within this `PR`.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/BackoffStrategy.kt
jujoramos: Done

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoBackoffStrategy.kt
jujoramos: Done.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
jujoramos: Done.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt
jujoramos: Done.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoBackoffStrategy.kt
jujoramos: Done.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt
jujoramos: 👍

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoBackoffStrategy.kt
jujoramos: Done.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoBackoffStrategy.kt
jujoramos: Removed.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/BackoffStrategy.kt
jujoramos: The retry mechanism is being used in places where the `utilities` was already imported, so this `PR` doesn't have any negative effect in that regard. Even though it's a valid concern, I'd like to move on as things stand right now and maybe re-asses the need for a new library in the future instead.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
jujoramos: If `tryWithBackoff()` throws the original exception here is simply because the user chose not to use the retry mechanism at all, they shouldn't even be using `tryWithBackoff()` and simply execute the block within `operation()` themselves; that's why I don't log the original exception here.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt
jujoramos: See [this comment](https://github.com/corda/corda-runtime-os/pull/5337#discussion_r1440368313) (and let's keep the discussion there to avoid multiple threads).

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
jujoramos: Done!.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
jujoramos: Done!.

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
jujoramos: Done! (it's up to the user what to do here, the current behaviour for the flow engine is the same as before: log the retry under `INFO` level).

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt
```diff
@@ -29,50 +42,21 @@ open class CryptoRetryingExecutor(
...
-                }
-            }
+        return tryWithBackoff(
+            logger = logger,
+            maxRetries = maxAttempts,
+            maxTimeMillis = Long.MAX_VALUE,
+            backoffStrategy = backoffStrategy,
+            shouldRetry = { _, _, throwable -> throwable.isRecoverable() },
+            onRetryAttempt = { attempt, delay, _ ->
+                logger.warn("Failed to execute, will retry after $delay milliseconds (attempt=$attempt)")
```
kyriathar, line 52: Not sure if this is clashing with other people's review but not really sure if such a message, decoupled from the actual error message will add much.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt
```diff
@@ -29,50 +42,21 @@ open class CryptoRetryingExecutor(
...
+            },
+            onRetryExhaustion = { attempt, _, throwable ->
+                logger.warn("Failed to execute (attempt={})", attempt)
+                CryptoRetryException("Failed to execute on attempt=$attempt", throwable)
```
kyriathar, line 56: Same as above, is the failed attempt number so important to promote it to be an exception's message?

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt
jujoramos: This is what the current `CryptoRetryingExecutor` does (see [here](https://github.com/corda/corda-runtime-os/blob/release/os/5.2/libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt#L62)), that's the only reason why I kept it.

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt
jujoramos: Same as above, this is what the current `CryptoRetryingExecutor` does (see [here](https://github.com/corda/corda-runtime-os/blob/release/os/5.2/libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt#L54)), that's the only reason why I kept it.

kyriathar: Just noticed the exception... `CryptoRetryException`, I can see we already use it as well like so maybe disregard this.

kyriathar (dismissed):

conal-smith:
> Also, I thought we'd look at moving away from crypto doing its own bespoke thing and consolidate how we do retries etc with the RPC pattern?

Yeah I don't remember us discussing or agreeing anything. But as I'm working on [handling transient errors for external processors](https://r3-cev.atlassian.net/browse/CORE-18534) the crypto retry logic breaks the mould. If @dickon agrees we could add a story to remove crypto retry handling and fall back on the RPC client's retry logic, the epic is [here](https://r3-cev.atlassian.net/browse/CORE-18531).

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
```diff
@@ -0,0 +1,82 @@
...
+ *  @throws RetryException, or the exception type created through [onRetryExhaustion], if the original exception was
+ *      marked as retryable but all retry attempts failed.
+ */
+@Suppress("LongParameterList")
+fun <T> tryWithBackoff(
+    logger: Logger,
+    maxRetries: Long,
+    maxTimeMillis: Long,
+    backoffStrategy: BackoffStrategy,
+    shouldRetry: (Int, Long, Throwable) -> Boolean = { _, _, _ -> false },
```
sprintengine, line 46: Should we make the default return `true` here? I think a caller would expect the default behaviour to retry?

jujoramos: I don't think so... the caller is responsible for specifying when an error should be retried so we can't infer that out of the box.

kyriathar (dismissed):
OK (for crypto)

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
```diff
@@ -110,14 +115,39 @@ class FlowEventProcessorImpl(
...
-            pipeline
-                .eventPreProcessing()
-                .virtualNodeFlowOperationalChecks()
-                .executeFlow(flowTimeout)
-                .globalPostProcessing()
-                .context
-        } catch (e: FlowTransientException) {
-            flowEventExceptionProcessor.process(e, pipeline.context)
+            tryWithBackoff(
+                logger = log,
```
LWogan, line 120: This tryWithBackoff invocation looks very busy. Do you think it would look nicer/more readable if the lambda params were extracted out into private functions at the end of this class? Then you could move the comments to the kdocs of the functions?

LWogan (dismissed):
These changes look fine to me.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
jujoramos: 👍

kyriathar (approved):

libs/crypto/crypto-impl/src/main/kotlin/net/corda/crypto/impl/retrying/CryptoRetryingExecutor.kt
dickon: Agreed

dickon (approved):
LGTM

libs/utilities/src/main/kotlin/net/corda/utilities/retry/RetryUtils.kt
sprintengine: but it means someone calling "tryWithBackoff won't actually get backoff unless they also provide a `shouldRetry` lambda, right? If I was using this I think I'd expect the default to retry all exceptions. But it's minor, feel free to merge I don't want to hold up the PR 👍

conal-smith (approved):

LWogan (approved):
LGTM

PR #5338, opened 2023-12-28
CORE-18554: Link manager start to listen to HTTP requests from the Gateway

yift-r3:
Enables an HTTP server in the link manager that listens to requests from the Gateway.

Some changes had been made to the link manager to support handling a single request.


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManager.kt
```diff
@@ -13,8 +13,8 @@ internal interface SessionManager : LifecycleWithDominoTile {
         wrappedMessages: Collection<T>,
         getMessage: (T) -> AuthenticatedMessageAndKey
     ): Collection<Pair<T, SessionState>>
-    fun <T>getSessionsById(uuids: Collection<T>, getSessionId: (T) -> String): Collection<Pair<T, SessionDirection>>
-    fun <T>processSessionMessages(wrappedMessages: Collection<T>, getMessage: (T) -> LinkInMessage): Collection<Pair<T, LinkOutMessage?>>
+    fun getSessionById(uuid: String): SessionDirection
+    fun processSessionMessage(message: LinkInMessage): LinkOutMessage?
```
dimosr, line 17: I believe the intention was to have these APIs be batch APIs to allow for optimisation, e.g. by querying all required sessions with a single operation via the state manager, instead of multiple ones. 

Conceptually, I think it might be possible to retain them as batch by having the common method (e.g. `handleMessage`) be a batch operation, which is called by the event log subscription with multiple input messages and by the http rpc subscription with a single input message.

williamvigorr3: I think we should keep the `SessionManager` API as is (as Dimos is saying). During the design meetings, we concluded we would need to batch lookups of session in the State Manager, even though we are going to add a cache. This cache can only be used once sessions are established (because the StatefulSession Manager uses the optimistic locking in the StateManager to prevent concurrent updates). We also need to query by `SessionId` which is not the primary key, so by batching we will reduce the expense some of these queries.

For now we could split the `InboundMessageProcessor` into two and leave the batching as is for the message bus implementation and make the batch size 1 for the `SyncRPCProcessor`. as Dimos suggested.  Maybe we can completely remove the `EventLogProcessor` from the Link Manager in the future and remove the batching.

@owenstanford WDYT?

yift-r3: The idea behind CORE-18513 is to remove the EventLogProcessor. If we enable it, the backing will always be a batch of one.

dimosr: The intention was for the 2 work streams to be completely independent (shared session & gateway -> link manager HTTP). So, we shouldn't be enforcing non-batch APIs to one work stream because the second work stream would need them, because there's always a chance for the second workstream to be delayed / deprioritised / or potentially not providing enough performance benefits (without batching or in general) to merit merging.

 I think it's ok if we have a batch of one temporarily for now when the feature flag for the HTTP is on, as long as we don't preclude other features to operate with batched operations.

dimosr: @owenstanford can you also share your view on this, please?

yift-r3:
I will re-open a new pull request later

PR #5339, opened 2023-12-28
CORE-18498: Add locks to CA and cluster keys operation

yift-r3:
Description:
Allow running e2e tests in parallel. Changes:
* Add lock around the CA saving and creation.
* Add a lock around changing the configurations based on the section.
* Add a named lock around package upload based on the package name.
* Add lock around checking for cluster key existence.
* Change the TLS type to be an environment variable.
* Change the membership onboarding to accept the MGM details instead of the group policy.

e2e code changes are in https://github.com/corda/corda-e2e-tests/pull/383. This pull request must be merged before that one.

Run in:
* [Upgrade](https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda5-upgrade-tests/detail/PR-383/9/pipeline) (The failure is the same as the [release branch](https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda5-upgrade-tests/detail/release%2F5.2/63/tests))
* [Multi cluster](https://ci02.dev.r3.com/job/Corda5/job/corda-e2e-tests-multi-cluster-tests/job/PR-383/16/)
* [Single Cluster](https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda5-e2e-tests/detail/PR-383/25/pipeline)


testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/config/ConfigTestUtils.kt
```diff
@@ -17,13 +17,12 @@ import java.time.Duration
 /**
  * Runs a function in the scope of a managed configuration which is automatically reverted upon completion.
  */
-fun managedConfig(vararg clusters: ClusterInfo, func: (TestConfigManager) -> Unit) {
-    val clusterDedup = clusters.toSet()
-    if(clusterDedup.size > 1) {
-        MultiClusterTestConfigManager(clusterDedup)
+fun managedConfig(clusters: Collection<ClusterInfo> = emptyList()): TestConfigManager {
```
williamvigorr3, line 20: Can you update the KDOC? There is no longer a function.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/config/TestConfigManager.kt
```diff
@@ -31,12 +31,9 @@ interface TestConfigManager: AutoCloseable {
 
     /**
      * Using the cluster REST API, push all currently loaded configurations and wait for the new config to become
-     * visible.
+     * visible. Then run the block and return the block returned value
```
williamvigorr3: `Then run block, revert the configuration change and return the result of block.`?

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/types/NetworkOnboardingMetadata.kt
```diff
@@ -16,5 +17,21 @@ data class NetworkOnboardingMetadata(
...
     val registrationContext: Map<String, String>,
-    val clusterInfo: ClusterInfo
-)
\ No newline at end of file
+    val clusterInfo: ClusterInfo,
+) {
+    fun asMgm() =
+        MgmMetadata(
+           holdingId = holdingId,
+            clusterInfo = clusterInfo
```
williamvigorr3: Nit: This should be aligned with `holdingId`.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/types/NetworkOnboardingMetadata.kt
```diff
@@ -16,5 +17,21 @@ data class NetworkOnboardingMetadata(
...
+        )
+}
+
+
+class MgmMetadata(
```
williamvigorr3: Can you add a KDoc? It's a bit confusing what this does as `holdingId` and `clusterInfo` aren't exclusive to an MGM.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/config/SingleClusterTestConfigManager.kt
```diff
@@ -11,14 +11,22 @@ import org.assertj.core.api.Assertions.assertThat
...
 
+    private data class LockKey(
+        val clusterKey: String,
+        val configurationSection: String,
+    )
+
     private companion object {
         private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
         private const val GET_CONFIG_TIMEOUT_SECONDS = 30L
+        private val lock = ConcurrentHashMap<LockKey, Lock>()
```
williamvigorr3, line 30: Is this definitely thread safe? 

If one thread adds a new lock to the map: can it cause the iteration order to change causing a deadlock? I think a deadlock can happen if two threads simulatanously try and take the locks in different orders.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/CryptoRPCSmokeTests.kt
```diff
@@ -234,10 +234,10 @@ class CryptoRPCSmokeTests : ClusterReadiness by ClusterReadinessChecker() {
         assertEquals(expected.requestingComponent, actual.requestingComponent)
         assertEquals(expected.requestTimestamp, actual.requestTimestamp)
         assertThat(actual.responseTimestamp.toEpochMilli())
-            .isGreaterThanOrEqualTo(expected.requestTimestamp.toEpochMilli())
-            .isLessThanOrEqualTo(now.toEpochMilli())
+            .isGreaterThanOrEqualTo(expected.requestTimestamp.minusSeconds(10).toEpochMilli())
```
williamvigorr3: Why have you changed this? Also I think it would be clearer if added `10` to the test as a constant.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/CertificateUtils.kt
```diff
@@ -15,15 +15,20 @@ import org.assertj.core.api.Assertions.assertThat
...
 
 /**
  * Get the default CA for testing. This is written to file so it can be shared across tests.
  */
-fun getCa(): FileSystemCertificatesAuthority = CertificateAuthorityFactory
-    .createFileSystemLocalAuthority(
-        KeysFactoryDefinitions("RSA".toAlgorithm(), 3072, null,),
-        File("build${File.separator}tmp${File.separator}ca")
-    ).also { it.save() }
+private val caLocker = ReentrantLock()
```
williamvigorr3: Nit: `caLocker` -> `caLock`

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MGMUtils.kt
```diff
@@ -294,12 +294,28 @@ fun ClusterInfo.declineRegistration(
     }
 }
 
+internal enum class TlsType(
```
williamvigorr3: Can this go into `ClusterInfo` instead? That way we contain all of the `System.getenv` into one file.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MembershipUtils.kt
```diff
@@ -102,7 +102,9 @@ fun ClusterInfo.onboardMember(
             it.writeBytes(tlsCert.toByteArray())
         }
         importCertificate(tlsCertFile, CERT_USAGE_P2P, CERT_ALIAS_P2P)
-        tlsCertificateUploadedCallback(tlsCert)
+        if (TlsType.type == TlsType.MUTUAL) {
```
williamvigorr3, line 103: `TlsType` is essentially a global variable. I know mutual TLS must be set on a cluster level once, so from a functional level it does matter that much. It still makes the code confusing though (as it isn't passed explicitly into this function).

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
```diff
@@ -194,19 +198,25 @@ fun ClusterInfo.createKeyFor(
     keyId["id"].textValue()
 }
 
-fun ClusterInfo.keyExists(
+private val keyExistsLock = ReentrantLock()
+fun ClusterInfo.whenNoKeyExists(
```
williamvigorr3, line 199: Why can't getKey be ran concurrently? Should this be fixed in the platform?

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
```diff
@@ -70,26 +71,29 @@ fun ClusterInfo.conditionallyUploadCordaPackage(
...
     cpiUpload: ClusterBuilder.() -> SimpleResponse
 ) = packageUploadSemaphore.runWith {
-    cluster {
-        if (getExistingCpi(name) == null) {
-            val responseStatusId = cpiUpload().run {
-                assertThat(code).isEqualTo(ResponseCode.OK.statusCode)
-                assertThat(toJson()["id"].textValue()).isNotEmpty
-                toJson()["id"].textValue()
-            }
+    uploading.compute(Pair(this.id, name)) {_, _ ->
```
williamvigorr3: Do we still need the semaphore if we are doing a compute here?

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/config/MultiClusterTestConfigManager.kt
```diff
@@ -19,13 +17,33 @@ class MultiClusterTestConfigManager(
...
-        return this
+
+    override fun <T> applyWithoutRevert(block: () -> T): T {
+        return configManagers.iterator().applyWithoutRevert(block)
     }
-    override fun close() = configManagers.forEach { it.close() }
+
+    private fun <T> Iterator<TestConfigManager>.apply(block: () -> T): T {
+        return if(!hasNext()) {
+            block()
```
williamvigorr3, line 31: Do we need to call this lambda again at the end?

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/config/SingleClusterTestConfigManager.kt
yift-r3: I'll make it sorted; it should reduce the chances of a deadlock.

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/CryptoRPCSmokeTests.kt
yift-r3: The test was failing (the clock of one of the pods was out of synk). See https://github.com/corda/corda-runtime-os/pull/5361

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MembershipUtils.kt
yift-r3: The problem is not just that it has to be set per cluster but that we can't change it after we set it and start a group, and we can't have a group on two types of clusters.

I can't see a reason to pass it to a function (especially if we want to enable parallel testing). I think that after this task, we can add it as a configurable option to the pipeline (and then we can simply retest mTLS every time).

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
yift-r3: The problem is that if the key doesn't exist, we will create a key with that alias. So, without the lock,  two threads might try to create a key simultaneously (or, event worse, one might assume that because the TLS key exists, the certificate was uploaded)...

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/config/MultiClusterTestConfigManager.kt
yift-r3: I don't think so

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
yift-r3: Let me try...

applications/workers/workers-smoketest/src/smokeTest/kotlin/net/corda/applications/workers/smoketest/services/CryptoRPCSmokeTests.kt
williamvigorr3: Probably worth making a const val and call it `POD_CLOCK_DIFF_TOLLERENCE` or something (otherwise it is not clear what this is doing from just reading the code).

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
williamvigorr3: I am a bit confused about why the semaphore was added. Why only allow two threads to run this section.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/config/MultiClusterTestConfigManager.kt
williamvigorr3: I meant why call it if `!hasNext()`? Isn't calling block on all the individual `SingleClusterTestConfigManager` enough?

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/MembershipUtils.kt
williamvigorr3: We could deploy a cluster specifically for mutual TLS (and throw it away after the mutual TLS tests). Probably we don't want to do that. 

I still think it makes the code confusing as the TLS type isn't passed into the function (so it can make it harder to read the code and work out how it's being set).

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/config/MultiClusterTestConfigManager.kt
williamvigorr3: I understand this now. At the end of the recursion we are calling block, this will lock all of the clusters at once..

williamvigorr3 (dismissed):
LGTM

williamvigorr3 (approved):
LGTM

PR #5341, opened 2023-12-29
CORE-17532: Upgrade Liquibase to latest version

malachyb:
The version of liquibase used in corda is currently out of date (4.19, released in January 2023). This PR upgrades to a more modern 4.24, which includes many security updates and performance upgrades.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -62,7 +62,7 @@ kryoVersion = 5.5.0
 kryoSerializersVersion = 0.45
 kotlinCoroutinesVersion=1.6.4
 # Liquibase upgrade to 4.20 and 4.21 failed - see CORE-12612 for more details
-liquibaseVersion = 4.19.0
+liquibaseVersion = 4.24.0
 # Needed by Liquibase:
 beanutilsVersion=1.9.4
 nettyVersion = 4.1.94.Final
--- a/gradle.properties
+++ b/gradle.properties
@@ -61,8 +61,6 @@ kafkaClientVersion=3.4.0_1
 kryoVersion = 5.5.0
 kryoSerializersVersion = 0.45
 kotlinCoroutinesVersion=1.6.4
-# Liquibase upgrade to 4.20 and 4.21 failed - see CORE-12612 for more details
-liquibaseVersion = 4.24.0
 # Needed by Liquibase:
 beanutilsVersion=1.9.4
 nettyVersion = 4.1.94.Final
--- a/libs/db/db-admin-impl/src/integrationTest/kotlin/net/corda/db/admin/impl/LiquibaseSchemaMigratorImplTest.kt
+++ b/libs/db/db-admin-impl/src/integrationTest/kotlin/net/corda/db/admin/impl/LiquibaseSchemaMigratorImplTest.kt
@@ -76,11 +76,23 @@ class LiquibaseSchemaMigratorImplTest {
         println("SQL Script:")
         println(sql1)
         assertThat(sql1.lowercase())
-            .containsPattern("run: +3")
-            .containsPattern("previously run: +0")
-            .containsPattern("filtered out: +1")
-            .containsPattern("total change sets: +4")
-            .containsPattern("dbms mismatch: +1")
+            .contains("create table public.test_table")
+            .contains("create table public.another_table")
+            .contains("create table public.generic_table")
+            .contains("create table public.databasechangelog")
+            .contains("create table public.databasechangeloglock")
+            .doesNotContain("create table public.postgres_table")
+
+        // run it
+        ds.connection.use { db ->
+            sql1.lines()
+                .filter {
+                    it.isNotBlank() && !it.startsWith("--")
+                }
+                .forEach {
+                    db.createStatement().execute(it)
+                }
+        }
 
         // Create another script
         val writer2 = StringWriter()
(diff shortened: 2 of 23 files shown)


gradle.properties
```diff
@@ -67,7 +67,7 @@ kryoVersion = 5.5.0
 kryoSerializersVersion = 0.45
 kotlinCoroutinesVersion=1.6.4
 # Liquibase upgrade to 4.20 and 4.21 failed - see CORE-12612 for more details
```
driessamyn: Please remove this comment.

gradle.properties
```diff
@@ -67,7 +67,7 @@ kryoVersion = 5.5.0
...
-liquibaseVersion = 4.19.0
+liquibaseVersion = 4.24.0
```
driessamyn: Can this be moved to the Version Catalog as part of this PR?

libs/db/db-admin-impl/src/integrationTest/kotlin/net/corda/db/admin/impl/LiquibaseSchemaMigratorImplTest.kt
```diff
@@ -76,23 +76,11 @@ class LiquibaseSchemaMigratorImplTest {
...
-        ds.connection.use { db ->
-            sql1.lines()
-                .filter {
-                    it.isNotBlank() && !it.startsWith("--")
-                }
-                .forEach {
-                    db.createStatement().execute(it)
-                }
-        }
+            .containsPattern("run: +3")
```
driessamyn: Why is this changed, and why is the test that executes the output SQL removed?
This diff suggests we no longer generate valid SQL

libs/db/db-admin-impl/src/integrationTest/kotlin/net/corda/db/admin/impl/LiquibaseSchemaMigratorImplTest.kt
```diff
@@ -101,21 +89,10 @@ class LiquibaseSchemaMigratorImplTest {
...
-        ds.connection.use { db ->
-            sql2.lines()
-                .filter {
-                    it.isNotBlank() && !it.startsWith("--")
-                }
-                .forEach {
-                    db.createStatement().execute(it)
-                }
-        }
+            .containsPattern("run: +1")
```
driessamyn: As above

libs/virtual-node/cpi-datamodel/build.gradle
```diff
@@ -7,6 +7,7 @@ description 'CPI Data Model'
 
 dependencies {
     compileOnly 'org.osgi:osgi.annotation'
+    implementation "org.liquibase:liquibase-core:$liquibaseVersion"
```
driessamyn: Why is this needed?

vkolomeyko (dismissed):
LGTM

libs/db/db-admin-impl/src/main/kotlin/net/corda/db/admin/impl/LiquibaseSchemaMigratorImpl.kt
```diff
@@ -29,7 +42,8 @@ class LiquibaseSchemaMigratorImpl(
             DatabaseFactory
                 .getInstance()
                 .findCorrectDatabaseImplementation(JdbcConnection(connection))
-        }
+        },
+    private val runChanges: Boolean = true
```
driessamyn: Why do we need this new argument?

libs/db/db-admin-impl/src/main/kotlin/net/corda/db/admin/impl/LiquibaseSchemaMigratorImpl.kt
```diff
@@ -110,12 +120,52 @@ class LiquibaseSchemaMigratorImpl(
...
+            )
+
+            try {
+                val scopeObjects = mapOf(
+                    Scope.Attr.database.name to lb.database,
+                    Scope.Attr.resourceAccessor.name to lb.resourceAccessor
+                )
+                Scope.child(scopeObjects) {
+                    val command: CommandScope
+                    if (sql != null) {
```
driessamyn, line 137: There are now 2 ways of identifying whether the SQL should be executed or not.

It may be simpler to do something along the lines of

```kotlin
val runChanges = null != sql

val command = if(runChanges) {
  CommandScope(UpdateCommandStep.COMMAND_NAME[0])
} else {
  CommandScope(UpdateSqlCommandStep.COMMAND_NAME[0]).also {
                        it.setOutput(
                            WriterOutputStream(
                                sql,
                                GlobalConfiguration.OUTPUT_FILE_ENCODING.currentValue
                            )
                        )
  }
}
```

libs/db/db-admin-impl/src/main/kotlin/net/corda/db/admin/impl/LiquibaseSchemaMigratorImpl.kt
malachyb: This is for testing purposes (specifically libs/db/db-admin-impl/src/test/kotlin[...]/LiquibaseSchemaMigratorTest.kt) because mocking CommandScope caused it to need us to be able to import liquibase in a lot of places, so I instead added a toggle to prevent it from actually running the changes.

driessamyn: As discussed on Slack yesterday, I don't understand this comment.
See https://github.com/corda/corda-runtime-os/pull/5421 and in particular https://github.com/corda/corda-runtime-os/pull/5421/commits/a9b7d42d5e37cc6f9ecc15a0f465181c82dbf424 that shows the new argument is completely redundant.

libs/db/db-admin-impl/src/main/kotlin/net/corda/db/admin/impl/LiquibaseSchemaMigratorImpl.kt
```diff
@@ -110,12 +123,39 @@ class LiquibaseSchemaMigratorImpl(
...
+                Scope.child(scopeObjects) {
+                    if (sql != null) {
+                        commandScopeFactory(UpdateSqlCommandStep.COMMAND_NAME).configure(lb, tag).also {
+                            it.setOutput(
+                                WriterOutputStream(
+                                    sql,
+                                    GlobalConfiguration.OUTPUT_FILE_ENCODING.currentValue
+                                )
+                            )
+                            it.execute()
```
driessamyn, line 145: I don't think you need this execute

malachyb: I do, the execute makes it actually generate the SQL. I tested without the execute and the SQL writer was empty.

malachyb:
build e2e

malachyb:
build e2e

malachyb:
build e2e

PR #5344, opened 2023-12-29
CORE-18518 - Baseline performance testing for P2P

nikinagy:
For performance testing purposes we needed to modify the scripts to be able to generate two-way load scenarios. We would also like to test scenarios where multiple members are present on the network, so the onboarding script was modified accordingly. These members should be able to communicate with everyone from the other cluster, the Sender code needed some slight adjustments to achieve this.


applications/tools/p2p-test/app-simulator/scripts/horizontal-scaling-testing/runScenario.sh
```diff
@@ -76,26 +122,55 @@ run_sender() {
     | jq '.appSimulators.sender.enabled=true' \
     > $senderDetailsFile
   echo "Running sender..."
-  SENDER_DETAILS_FILE=$senderDetailsFile "$SCRIPT_DIR"/runSender.sh
+  if [ "$ONE_WAY_LOAD" == "true" ]
+  then
+    SENDER_DETAILS_FILE=$senderDetailsFile "$SCRIPT_DIR"/runSender.sh ONE_WAY
```
yift-r3: We can keep the `ONE_WAY` or `TWO_WAY` in a local variable and remove the if statement in some places.

applications/tools/p2p-test/app-simulator/scripts/horizontal-scaling-testing/runScenario.sh
```diff
@@ -104,35 +179,28 @@ batchSize=50
...
-        count(rm.delivery_latency_ms) / 30 as throughput,
-        max(rm.delivery_latency_ms) / 1000.0 as max_latency,
-        min(rm.delivery_latency_ms)/ 1000.0 as min_latency,
-        avg(rm.delivery_latency_ms)/ 1000.0 as average_latency,
-        (percentile_disc(0.99) within group (order by rm.delivery_latency_ms))/ 1000.0 as p99_latency
-       from received_messages rm
-       group by time_window
-       order by time_window asc
-;" >> "$reportFile"
+#totalNumberOfMessages=60000
```
yift-r3: Do we still need this?

applications/tools/p2p-test/app-simulator/scripts/runReceiver.sh
```diff
@@ -4,38 +4,62 @@ set -e
...
+}
+
+deploy_receiver() {
+  helm upgrade --install \
+      app-simulator $APP_SIMULATOR_CHART_DIR \
+       -f "$SCRIPT_DIR"/receiver.yaml -n $1 \
+        $metrics_args \
+       --set db.appSimulator.password=$2 \
+       --set "imagePullSecrets={docker-registry-cred}" \
+       --set image.tag=5.2.0.0-alpha-1704810192792 \
```
yift-r3: We need to revert the explicit tag change

applications/tools/p2p-test/app-simulator/scripts/runSender.sh
```diff
@@ -4,13 +4,46 @@ set -e
...
+
+deploy_sender() {
+  helm upgrade --install \
+    app-simulator-sender $APP_SIMULATOR_CHART_DIR \
+    -f "$SENDER_DETAILS_FILE" \
+    $metrics_args \
+    -n $1 \
+    --set db.appSimulator.password=$2 \
+    --set "imagePullSecrets={docker-registry-cred}" \
+    --set image.tag=5.2.0.0-alpha-1704810192792 \
```
yift-r3: Same comment about the tag

applications/tools/p2p-test/app-simulator/scripts/settings.sh
```diff
@@ -16,6 +16,8 @@ fi
 # Uncomment to enable mutual TLS
 # MTLS="Y"
 
+RUN_MODE="TWO_WAY"
```
yift-r3: We should probably add something about the new variables to the readme.

applications/tools/p2p-test/app-simulator/src/main/kotlin/net/corda/p2p/app/simulator/AppSimulator.kt
```diff
@@ -319,8 +318,8 @@ data class DBParams(val username: String, val password: String, val host: String
 }
 
 data class LoadGenerationParams(
-    val peer: HoldingIdentity,
-    val ourIdentity: HoldingIdentity,
+    val peers: List<HoldingIdentity>,
```
yift-r3: Collection?

applications/tools/p2p-test/app-simulator/src/main/kotlin/net/corda/p2p/app/simulator/Sender.kt
```diff
@@ -79,17 +81,27 @@ class Sender(
                 val publisher = publisherFactory.createPublisher(PublisherConfig("app-simulator", false), messagingConfig)
                 publisher.use {
                     while (moreMessagesToSend(messagesSent, loadGenParams)) {
-                        val messageWithIds = (1..loadGenParams.batchSize).map {
-                            "$senderId:$client:${++messagesSent}"
-                        }.map {
-                            createMessage(it, senderId, loadGenParams.peer, loadGenParams.ourIdentity, loadGenParams.messageSizeBytes)
+                        val messagesWithIds = mutableListOf<Pair<MessageMetaData, AppMessage>>()
```
yift-r3, line 93: Can you use flatmap instead of a mutable list?

applications/tools/p2p-test/app-simulator/src/main/kotlin/net/corda/p2p/app/simulator/Sender.kt
```diff
@@ -178,7 +190,8 @@ class Sender(
         }
         val payload = MessagePayload(senderId, randomData, Instant.now())
         val message = AuthenticatedMessage(messageHeader, ByteBuffer.wrap(objectMapper.writeValueAsBytes(payload)))
-        return messageId to AppMessage(message)
+        logger.info("Created message from ${srcIdentity.x500Name} to ${destinationIdentity.x500Name}")
```
dimosr: was this left intentionally to INFO or did you just use it for testing while developing and you intended to switch it to DEBUG in the end? Asking because we are already publishing an info-level log at the end of publishing a batch, which is a slightly less frequent event than this one.

applications/tools/p2p-test/app-simulator/scripts/README.md
```diff
@@ -20,11 +20,27 @@ In this mode, deploy.sh deploys 1 worker cluster (a) into K8s.
...
+runReceiver.sh deploys receiving and dbsink app simulators into one or two clusters, based on the run mode.  
+runSender.sh deploys sending app simulator into one or two clusters, based on the run mode.  
 In single-cluster mode, all app simulators are deployed onto cluster-a.
 
 settings.sh contains some settings which can be updated to change the deployment.
 sender.yaml and receiver.yaml contain some settings for the receiving, and dbsink app simulators and the sending app simulator, respectively.
 
 tearDown.sh tears down the worker cluster and kills the port forwarding processes.
+
+### Run modes for app simulator
```
dimosr, line 32: Should we add some docs for configuring the number of vnodes?

yift-r3 (dismissed):

applications/tools/p2p-test/app-simulator/src/main/kotlin/net/corda/p2p/app/simulator/Sender.kt
```diff
@@ -79,17 +81,32 @@ class Sender(
...
-                        val messageWithIds = (1..loadGenParams.batchSize).map {
-                            "$senderId:$client:${++messagesSent}"
-                        }.map {
-                            createMessage(it, senderId, loadGenParams.peer, loadGenParams.ourIdentity, loadGenParams.messageSizeBytes)
+                        val messagesWithIds = mutableListOf<Pair<MessageMetaData, AppMessage>>()
+                        while (messagesWithIds.size < loadGenParams.batchSize) {
+                            senderIds.forEach outer@{ (senderId, senderHoldingId) ->
+                                if (messagesWithIds.size >= loadGenParams.batchSize) return@outer
+                                loadGenParams.peers.forEach { destination ->
+                                    if (messagesWithIds.size >= loadGenParams.batchSize) return@forEach
```
dimosr: This will be a bit problematic in cases where `batchSize` is smaller than the number of combinations (i.e. senders x destinations), since some combinations will always be excluded for every batch.

A way to do the same without being subject to this problem would be to create a list of all the possible combinations and then have an index within that list that persists across each iteration, so that you continue on each batch from the combination you stopped on the previous one.

Note that the proposed option could still suffer from memory issues when the number of combinations is too large (which could be solved by keeping track of the indexes in the sublists, instead of generating the full list which would end up being a bit more complex). But, I suspect this won't be a practical issue for the time being, so we could go for the simpler approach for now.

applications/tools/p2p-test/app-simulator/src/main/kotlin/net/corda/p2p/app/simulator/Sender.kt
```diff
@@ -71,6 +72,24 @@ class Sender(
                 } else {
                     null
                 }
+                val dbConnectionToInsertIndex = if (dbParams != null) {
```
dimosr: Is there a strong reason to store the index on the DB?

If not, I would avoid it for performance reasons.

nikinagy: I am not sure where else we could store it, since in each iteration we are deploying a completely new sender.

dimosr: Not sure I understand.. what's the problem with keeping the index in memory?

applications/tools/p2p-test/app-simulator/src/main/kotlin/net/corda/p2p/app/simulator/Sender.kt
```diff
@@ -78,18 +99,58 @@ class Sender(
...
+                    logger.info("Continue to send messages starting from $currentIndex index")
+
                     while (moreMessagesToSend(messagesSent, loadGenParams)) {
-                        val messageWithIds = (1..loadGenParams.batchSize).map {
-                            "$senderId:$client:${++messagesSent}"
-                        }.map {
-                            createMessage(it, senderId, loadGenParams.peer, loadGenParams.ourIdentity, loadGenParams.messageSizeBytes)
+                        val messagesWithIds = mutableListOf<Pair<MessageMetaData, AppMessage>>()
+                        while (messagesWithIds.size < loadGenParams.batchSize) {
+                            for (i in currentIndex..allPossibleCombinations.size) {
```
dimosr: If you remove the for loop (along with the break) and replace it with a modulo calculation, it could make it a bit cleaner, e.g.
```
while (messagesWithIds.size < loadGenParams.batchSize) {
    .... // current logic
    currentIndex = (currentIndex + 1) % allPossibleCombinations.size
}
```

applications/tools/p2p-test/app-simulator/src/main/kotlin/net/corda/p2p/app/simulator/Sender.kt
nikinagy: Because the way this whole app simulator works is:
1) We deploying a sender that sends x messages
2) Sender task gets completed once it finishes sending all messages
3) For new iteration we trigger a completely new sender instance. -> You won't have anything in memory.
The only alternative I can think of is writing the index to a file.

dimosr: I think not having anything in memory is fine. You'll just start the index from zero on every run.

applications/tools/p2p-test/app-simulator/scripts/README.md
nikinagy: addressed in [a832874](https://github.com/corda/corda-runtime-os/pull/5344/commits/a8328742ec219d6ff94c7cd6fa50dc2f8b50944b)

applications/tools/p2p-test/app-simulator/scripts/horizontal-scaling-testing/runScenario.sh
```diff
@@ -75,67 +118,92 @@ run_sender() {
...
+  echo "---cluster A---" >> "$reportFile"
+  write_report_file $APP_SIMULATOR_DB_NAMESPACE_A $dbPasswordA
+  echo "---cluster B---" >> "$reportFile"
+  write_report_file $APP_SIMULATOR_DB_NAMESPACE_B $dbPasswordB
+fi
 
 
 echo "Tearing down previous clusters"
-"$SCRIPT_DIR"/tearDown.sh
+#"$SCRIPT_DIR"/tearDown.sh
```
dimosr: Did you intend to revert this comment in the end?

applications/tools/p2p-test/app-simulator/scripts/runSender.sh
```diff
@@ -4,13 +4,46 @@ set -e
...
+    --set "imagePullSecrets={docker-registry-cred}" \
+    --set image.tag=$DOCKER_IMAGE_VERSION \
+    --set "appSimulators.sender.senderX500Names=$3" \
+    --set "appSimulators.sender.peerX500Names=$4" \
+    --set "appSimulators.sender.senderGroupId=$GROUP_ID" \
+    --set "appSimulators.sender.peerGroupId=$GROUP_ID" \
+    --set "db.appSimulator.namespace=$5"  \
+    --set appSimulators.sender.replicaCount=$WORKER_REPLICAS \
+    --timeout 30m00s \
+    --wait &
```
dimosr, line 39: what's the logic behind `&`? 

Don't we want to wait until the deployment of the chart is complete, in case there have been errors?

dimosr: Going a bit further I can see you are doing a `wait`, so seems like we are waiting. But, not clear to me why we're doing it that way.

applications/tools/p2p-test/app-simulator/scripts/corda-eks.metrics.yaml
```diff
@@ -7,4 +7,8 @@ metrics:
   podMonitor:
     enabled: true
     labels:
-      release: observability
\ No newline at end of file
+      release: observability
+    keepNames: [ ]
```
yift-r3, line 11: Do we need this change?

dimosr (commented):
just a few minor things, otherwise looks good to me

applications/tools/p2p-test/app-simulator/scripts/horizontal-scaling-testing/runScenario.sh
yift-r3: ?

applications/tools/p2p-test/app-simulator/scripts/runSender.sh
dimosr: suspecting that it might be in order to parallelise the deployment of multiple senders (in the two-way load scenario)? If that's the case, should we specify the specific PIDs to `wait` to be a bit more explicit?

applications/tools/p2p-test/app-simulator/scripts/horizontal-scaling-testing/runScenario.sh
nikinagy: Yes, I left this behind by mistake.

applications/tools/p2p-test/app-simulator/scripts/corda-eks.metrics.yaml
nikinagy: Yes, some of our charts (eg Outbound sessions total) show "no data" without this

applications/tools/p2p-test/app-simulator/scripts/horizontal-scaling-testing/runScenario.sh
nikinagy: addressed in [d691247](https://github.com/corda/corda-runtime-os/pull/5344/commits/d691247d5888fb6ea464e7b5d814b890f377614f)

applications/tools/p2p-test/app-simulator/scripts/runSender.sh
nikinagy: addressed in [d691247](https://github.com/corda/corda-runtime-os/pull/5344/commits/d691247d5888fb6ea464e7b5d814b890f377614f)

applications/tools/p2p-test/app-simulator/scripts/horizontal-scaling-testing/runScenario.sh
nikinagy: addressed in [d691247](https://github.com/corda/corda-runtime-os/pull/5344/commits/d691247d5888fb6ea464e7b5d814b890f377614f)

dimosr (approved):

yift-r3 (approved):

PR #5347, opened 2024-01-02
CORE-19124: allow some slack on timings

dickon:
It's not necessarily the case that the system clocks where the e2e test is precsiely aligned or behind the clock where Corda is running, nor should we assume that. HTTP RPCs can be so fast as to show this up in test failures where we notice the response timestamp is before the request timestamp. So, provided they are within 1 minute I think we don't care which clock is ahead of the others.

Also set a  test clock to the real time since otherwise the 1 minute check doesn't work.


driessamyn (approved):
LGTM

PR #5348, opened 2024-01-02
CORE-18706 - Fixed external messaging after the mediator pattern was introduced

filipesoliveira:
The mediator pattern introduced a translation from `Record< * , *>` to `MediatorMessage<*>`. There are two key points to keep in mind. The field `topic` in the `Record<*,*>` is dropped and the `payload` field in the `MediatorMessage<*>` contains the message to be sent when the `external messaging` feature is used. In order to fix the issue, the value in the `topic` field is now added to the `properties` field in the `MediatorMessage<*>` which is later used by `FlowEventMediatorFactoryImpl`.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
@@ -148,7 +148,7 @@ class FlowEventMediatorFactoryImpl @Activate constructor(
                     rpcEndpoint(UNIQUENESS_WORKER_REST_ENDPOINT, UNIQUENESS_PATH), SYNCHRONOUS)
                 is FlowEvent -> routeTo(messageBusClient,
                     FLOW_EVENT_TOPIC, ASYNCHRONOUS)
-                is String -> routeTo(messageBusClient,
+                is String -> routeTo(messageBusClient, // Handling external messaging
                     message.properties[MSG_PROP_TOPIC] as String, ASYNCHRONOUS)
                 else -> {
                     val eventType = event?.let { it::class.java }
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
@@ -23,6 +23,7 @@ import net.corda.messaging.api.constants.WorkerRPCPaths.UNIQUENESS_PATH
 import net.corda.messaging.api.constants.WorkerRPCPaths.VERIFICATION_PATH
 import net.corda.messaging.api.mediator.MediatorMessage
 import net.corda.messaging.api.mediator.MessageRouter
+import net.corda.messaging.api.mediator.MessagingClient.Companion.MSG_PROP_TOPIC
 import net.corda.messaging.api.mediator.RoutingDestination.Companion.routeTo
 import net.corda.messaging.api.mediator.RoutingDestination.Type.ASYNCHRONOUS
 import net.corda.messaging.api.mediator.RoutingDestination.Type.SYNCHRONOUS
@@ -147,6 +148,8 @@ class FlowEventMediatorFactoryImpl @Activate constructor(
                     rpcEndpoint(UNIQUENESS_WORKER_REST_ENDPOINT, UNIQUENESS_PATH), SYNCHRONOUS)
                 is FlowEvent -> routeTo(messageBusClient,
                     FLOW_EVENT_TOPIC, ASYNCHRONOUS)
+                is String -> routeTo(messageBusClient,
+                    message.properties[MSG_PROP_TOPIC] as String, ASYNCHRONOUS)
                 else -> {
                     val eventType = event?.let { it::class.java }
                     throw IllegalStateException("No route defined for event type [$eventType]")
--- a/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/TaskManagerHelper.kt
+++ b/libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/TaskManagerHelper.kt
@@ -173,9 +173,11 @@ internal class TaskManagerHelper<K : Any, S : Any, E : Any>(
             value!!,
             headers
                 .toMessageProperties()
-                .also {
-                    it[MSG_PROP_KEY] = key;
-                    it[MSG_PROP_TOPIC] = topic?: ""
+                .also { properties ->
+                    properties[MSG_PROP_KEY] = key;
+                    if(topic!=null) {
+                        properties[MSG_PROP_TOPIC] = topic!!
+                    }
                 },
         )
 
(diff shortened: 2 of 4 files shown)


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/TaskManagerHelper.kt
```diff
@@ -164,13 +165,20 @@ internal class TaskManagerHelper<K : Any, S : Any, E : Any>(
             replyMessage!!.payload,
         )
 
+    // This conversion does not preserve all the data contained on the record!
```
owenstanford: Not sure I understand the addition of these comments, is the loss of timestamp (or other record attributes) an issue elsewhere?

components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
```diff
@@ -147,6 +148,8 @@ class FlowEventMediatorFactoryImpl @Activate constructor(
                     rpcEndpoint(UNIQUENESS_WORKER_REST_ENDPOINT, UNIQUENESS_PATH), SYNCHRONOUS)
                 is FlowEvent -> routeTo(messageBusClient,
                     FLOW_EVENT_TOPIC, ASYNCHRONOUS)
+                is String -> routeTo(messageBusClient,
```
mbrkic-r3: Is there a reason to use `String` instead of new type of event?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/TaskManagerHelper.kt
filipesoliveira: I will remove this comment. The only important attribute being lost was the topic name. But the data should be preserved when converting a message from one format to another. Losing data silently leads to bugs as it was the case. The comment was to alert people that a field was being dropped.

components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
filipesoliveira: As I understand, the event is the payload of the message that is sent by the `messageBusClient`. For this scenario, the payload that needs to be sent is a simple string.

mbrkic-r3: What is this 'external messaging' about/related to? Where do those messages actually go to?

filipesoliveira: The `external messaging` is a feature that allows a flow to send messages to third party applications using Kafka. The message is sent to special Kafka topics which are manually created. Third party applications listen to those topics for messages.

mbrkic-r3: I would prefer a new type for this so it's more readable, that would encapsulate all external messaging data (e.g. field for payload String, another for destination topic). However, that would probably require some additional changes so that only payload String is sent to Kafka (maybe just a custom Avro serializer).
For now, could you please put some comments here so it's clear that this route is for handling external messaging?

components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
```diff
@@ -147,6 +148,8 @@ class FlowEventMediatorFactoryImpl @Activate constructor(
                     rpcEndpoint(UNIQUENESS_WORKER_REST_ENDPOINT, UNIQUENESS_PATH), SYNCHRONOUS)
                 is FlowEvent -> routeTo(messageBusClient,
                     FLOW_EVENT_TOPIC, ASYNCHRONOUS)
+                is String -> routeTo(messageBusClient, // Handling external messaging
```
mbrkic-r3, line 151: Please update related unit test as well.

vkolomeyko: `String` is fairly generic type and IMO reserving it for External Messaging is not quite right. After all what makes External Messaging entitled for this generic type considering all the other event types that we have.

Would it not be a better idea to introduce a dedicated `ExternalMessagingEvent` similar to `FlowEvent` ?

components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
filipesoliveira: I agree with what you are saying but the code could be simplified and we could avoid the extra class to encapsulate the message. Mind you that I don't have the full picture and I understand the work done with the message mediator was done on top of something that already exists. My suggestion would be to remove the class `Record`. I cannot see why we cannot create `MediatorMessage` instead of a `Record` and avoid the existing translation apart from the effort required to make the code change. The `MessageRouter` does not seem necessary. The route type should be picked based on whether the message is being sent `sync` or `async` instead of the event type. The `MediatorMessage` contains a field called `properties` which could be used to store the way the message should be sent, i.e.  `sync` or `async`.

components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
filipesoliveira: The `event` is the payload that is sent over the wire. The type is generic because it was agreed that the external messaging should not impose any rules on the message format. Encapsulating the message in a class breaks the pattern that is followed and an awkward ad-hoc check only for the new class will have to be added in. I think there is room to simplify the code and remove the need of an extra class and the required check. I wrote my view here https://github.com/corda/corda-runtime-os/pull/5348#discussion_r1440279474.

components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
mbrkic-r3: Yes, `Mediator` is reusing some existing classes that work with `Record`s (reasons being time constraints and fact that there are other patterns that work with `Record`s, which were not refactored).
`MessageRouter` was introduced to fix leaking of lower level details into higher levels (e.g. topic name is low level detail related to messaging layer, and message processors, which are higher levels, should not be aware of it).
`Mediator` was implemented to support any type of messaging client and they can support any type of message pattern (e.g. send and forget, request-reply). So `sync` or `async` was a property of messaging client, and mediator can be configured to work with any number of messaging clients (i.e. not just 2).
Having "proper" types would make code more organized and easier to handle (compared to using just String for all events and putting all related data in properties). As a quick fix for important bug, I'm fine with this, but we should have a tech debt ticket opened to think more about it and improve implementation.

owenstanford: I agree it makes sense to add a comment here to explain the background to this, but also agree with @filipesoliveira that this is a bug fix for an oversight when the code was changed to mediator so we don't want to do any retrospective design stages as part of this PR.

owenstanford (approved):
LGTM

mbrkic-r3 (approved):

PR #5349, opened 2024-01-02
CORE-18691 Sandbox Logging

simon-johnson-r3:
The bug described in CORE-18691 intermittently pops up despite thread safety improvements (although less regularly). This PR adds more logging to sandbox creation and also bundle installation and uninstallation so we can ensure the correct sandbox is being used at the point of failure.

We already have logging for sandbox eviction and closing and so this extra logging adds some symmetry to that.

On top of this I removed a function we only ever called in tests. This reduces some complexity in the `SandboxCreationService` which has no need for "non started" bundles.

### DIFF
--- a/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxServiceImpl.kt
+++ b/libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxServiceImpl.kt
@@ -98,13 +98,8 @@ internal class SandboxServiceImpl @Activate constructor(
         }
     }
 
-    override fun createSandboxGroup(cpks: Iterable<Cpk>, securityDomain: String) =
-        createSandboxes(cpks, securityDomain, startBundles = true)
-
-    override fun createSandboxGroupWithoutStarting(cpks: Iterable<Cpk>, securityDomain: String) =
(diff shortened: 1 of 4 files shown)


libs/sandbox-internal/src/main/kotlin/net/corda/sandbox/internal/SandboxServiceImpl.kt
```diff
@@ -184,16 +179,11 @@ internal class SandboxServiceImpl @Activate constructor(
     }
 
     /**
-     * Creates a [SandboxGroup] in the [securityDomain], containing a [Sandbox] for each of the [cpks]. On the first
-     * run, also initialises a public sandbox. [startBundles] controls whether the CPK bundles are also started.
+     * Creates a [SandboxGroup] in the [securityDomain], containing a [Sandbox] for each of the [cpks].
      *
      * Grants each sandbox visibility of the public sandboxes and of the other sandboxes in the group.
      */
-    private fun createSandboxes(
```
simon-johnson-r3, line 192: `startBundles` was only ever set to `false` in a test, so it was pointless functionality, which means the public interface which called it with this setting could be removed too

kyriathar (approved):
LGTM

PR #5360, opened 2024-01-03
CORE-18984 use config value to set the poll timeout used in the event mediator

LWogan:
https://github.com/corda/corda-api/pull/1432

use config value to set the poll timeout used in the event mediator. Previously this was hardcoded to set separately to other subscription types. Adding a mediator config section alleviates this issue. Now the mediator consumers poll timeout can be configured via the corda-api messaging config section


mbrkic-r3 (dismissed):
LGTM

mbrkic-r3 (approved):
Re-approved

PR #5362, opened 2024-01-03
CORE-19168, CORE-19176: show intermediate hashes in merkle tree and proof representations

dickon:
Extend the text representation of Merkle trees to include intermediate hashes. This makes the way the trees work more transparent in the unit tests.

SInce the intermediate hashes are private this involves moving the tree rendering code to be available from a `toString()` method. I also extended the tree rendering code to allow labels on all nodes as well as just leaves, and I took the time to document the rendering routine better.

Use ints rather than longs for the trivial hashes so that the test cases look nicer. The way the trivial hashes are done the least significant bits are the distinctive ones so we can't just truncate in renders.

Rendering a Merkle proof requires using the root calculation algorithm with additional bookkeeping, so we add a callback to that code.


libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
```diff
@@ -399,9 +402,11 @@ class MerkleTreeTest {
         leafIndicesCombination: List<Int>,
         root: SecureHash,
         treeSize: Int
-    ): MerkleProof {
-        val proof = merkleTree.createAuditProof(leafIndicesCombination)
+    ): MerkleProofImpl {
+        val proofGeneric = merkleTree.createAuditProof(leafIndicesCombination)
+        val proof = proofGeneric as MerkleProofImpl
 
+        println("Proof ${proof.toString()}")
```
vlajos: This does not look intentional.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleTreeImpl.kt
```diff
@@ -236,4 +236,149 @@ class MerkleTreeImpl(
...
+     *   a9d5543c┳bab170b1┳7901af93━00:00:00:00
+     *           ┃        ┗471864d3━00:00:00:01
+     *           ┗66973b1a━66973b1a━00:00:00:02
+     *
+     * In this case a9d5543c is the hash of the root of the tree.
+     *
+     * @return a textual representation of the MerkleTree
+     */
+
+    override fun toString(): String {
```
vlajos: Can overriding `toString` allow any hash/private information to leak into logs/other unwanted places?

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
```diff
@@ -303,6 +303,7 @@ class MerkleTreeTest {
         val leaves1 = "abcdef".map { it.toString().toByteArray() }
         val leaves2 = "ghijkl".map { it.toString().toByteArray() }
         val tree1 = MerkleTreeImpl.createMerkleTree(leaves1, defaultHashDigestProvider)
+        println(tree1)
```
vlajos: This looks like debugging.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleTreeImpl.kt
```diff
@@ -236,4 +236,149 @@ class MerkleTreeImpl(
...
+            // x is the level of the tree, so x=0 is the top node
+            // work out how many nodes across at the current level of the tree we are at, or -1 if we aren't at a node
+            val nodeIndex = nodeYCoordinates.get(x to y)?:-1
+            "${nodeLabels[x to nodeIndex]?.padEnd(longestLabels[x], ' ')?:(" ".repeat(longestLabels[x]))}${grid.getOrDefault(x to y, ' ')}"
+        }
+        val label: String = leafLabels.getOrNull(y) ?: ""
+        "${line.joinToString("")}$label"
+    }
+
+    return lines.joinToString("\n")     // Return the whole tree as a single string.
```
vlajos: I suspect a big chunk of this will also be needed in proofs.
(Though I am not too sure what is the best place for it.)

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleTreeImpl.kt
dickon: I spoke to Matthew and he recommended not overloading toString because of the overhead, so I'll change it.

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
dickon: deleted

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
dickon: deleted

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleTreeImpl.kt
dickon: Yes, and I've added that now.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleTreeImpl.kt
dickon: Now using `render` methods.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -52,8 +52,22 @@ class MerkleProofImpl(
      * proof or tree creation. [MerkleTreeImpl.createAuditProof]
      * It recreates the routes towards the root element from the items in the leaves to be proven with using
      * the proof hashes when they are needed.
+     *
+     * @param digest the digest service to use, which must generate hashes compatibile with the [hashes] constructor parameter.
```
```suggestion
     * @param digest the digest service to use, which must generate hashes compatible with the [hashes] constructor parameter.
```

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -52,8 +52,22 @@ class MerkleProofImpl(
...
+     * @return the secure hash of the root of the Merkle proof
+     */
+    override fun calculateRoot(digest: MerkleTreeHashDigest): SecureHash = calculateRootInstrumented(digest)
+
+    /**
+     * Walk the Merkle proof, making a callback on each node that is derived along the way, and return
+     * the root hash if the proof is valid.
+     *
+     * @param digest the digest service to use, which must generate hashes compatibile with the [hashes] constructor parameter.
```
```suggestion
     * @param digest the digest service to use, which must generate hashes compatible with the [hashes] constructor parameter.
```

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -158,31 +173,35 @@ class MerkleProofImpl(
...
                         //
-                        // Also remember we used hashIndex by bumping the counter
-                        Pair(
-                            item.first / 2,
-                            digest.nodeHash(treeDepth, item.second, hashes[hashIndex++])
-                        )
+                        // Also remember we used $hashIndex by bumping the counter
+                        foundHashCallback(hashes[hashIndex], treeDepth+1, item.first + 1, hashIndex)
+                        val newHash = digest.nodeHash(treeDepth, item.second, hashes[hashIndex++])
+                        foundHashCallback( newHash, treeDepth, newIndex, null)
```
vlajos, line 193: `foundHashCallback` looks like it is called both when we use one hash from the incoming ones and also when we calculate one from existing data.
Its doc block
> that is calculated during the proof

mentions only the second one.
But its name(`found`) refers more to the first one.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleTreeImpl.kt
```diff
@@ -236,4 +236,157 @@ class MerkleTreeImpl(
...
+        return renderTree(leaves.map { it.joinToString(separator = ":") { b -> "%02x".format(b) } }, hashes)
+    }
+}
+
+/**
+ * Render a left-balanced Merkle tree of a certain size.
+ *
+ * This is a top level function not a method of `MerkleTreeImpl` since it is a pure function which
+ * intentionally has no association with a specific instance of a `MerkleTreeImpl`. It is used by the
+ * `MerkleTreeImpl.render` method.
```
vlajos: and MerkleProofImpl...
Maybe this function could be in a dedicated file to keep the current one a bit simpler.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: applied

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: applied

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleTreeImpl.kt
dickon: Good idea

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: yes good point, doc block rewritten

vlajos (dismissed):

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -52,8 +47,33 @@ class MerkleProofImpl(
...
+    override fun calculateRoot(digest: MerkleTreeHashDigest): SecureHash = calculateRootInstrumented(digest)
+
+    /**
+     * Walk the Merkle proof, making a callback on each node that is derived along the way, and return
+     * the root hash if the proof is valid.
+     *
+     * @param digest the digest service to use, which must generate hashes compatible with the [hashes] constructor parameter.
+     * @param foundHashCallback called on each node that has a node taken from incoming hashes or calculated during the proof.
+     *                          The arguments are the hash of the node, level (with top of the tree at 0),
+     *                          index (position across from the left had side of the tree, starting at 0),
```
kyriathar, line 63: had -> hand

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -52,8 +47,33 @@ class MerkleProofImpl(
...
+     *                          The order that this is called is left to right then bottom to top, i.e. the same
+     *                          order as input hashes are consumed.
+     * @return the secure hash of the root of the Merkle proof
      */
-    override fun calculateRoot(digest: MerkleTreeHashDigest): SecureHash {
+    @Suppress("NestedBlockDepth", "ThrowsCount")
+
+    fun calculateRootInstrumented(
+        digest: MerkleTreeHashDigest,
+        foundHashCallback: (hash: SecureHash, level: Int, nodeIndex: Int, consumed: Int?) -> Unit =
```
kyriathar, line 74: nit: would `onNewHash` be a better name for this callback? Something with `on...` anyway.

kyriathar (commented):
Two minor comments. Haven't looked into the render Merkle tree stuff. I believe @vlajos already reviewed those.

vlajos (approved):

dickon:
Thanks @vlajos and @kyriathar - I have  addressed @kyriathar's typo comment and the rename, in an upcoming change

PR #5363, opened 2024-01-03
CORE-18926: Format the p2p crypto library

yift-r3:
Apply ktlint on the p2p crypto library


YashNabar (commented):
Can we add the commit to the git-blame-ignore file as described [here](https://refactored-adventure-z4oe4g4.pages.github.io/ways-we-work/kotlin/kotlin-auto-code-formatting.html#first-enablingauto-format-run)?

YashNabar (approved):
LGTM (we would need to merge without squashing to preserve the commit hash for 42e98e9d3c9dceb0a6aaeec07ed94c633b42b656)

PR #5364, opened 2024-01-03
CORE-19175: Fix new issue

yift-r3:
We had two issues:
1. The `keyManager.getServerAliases` doesn't always return the correct aliases when actual issuers are not null. Since we don't need the issuers for this method, I reverted it to pass null.
2. The `getCA` created a new issuer at any time instead of saving it and loading it from the disk.
With those fixes, the multi-cluster build looks more stable - see https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda-e2e-tests-multi-cluster-tests/detail/yift%2Fcore-19128%2Ftest-what-went-wrong/4/pipeline


charlieR3 (approved):
LGTM

dickon (approved):
LGTM

PR #5366, opened 2024-01-03
CORE-18534 External Processors to return transient retry codes

conal-smith:
Note - This is the first PR in a series of PRs to update sync-RPC processors on the server side. It includes:
- messaging layer changes, `CordaHTTPTransientServerException` and handling to return HTTP response with 503 status code 
- `EntityRequestProcessor` changes

**Problem description:**
All flow engine external events go through the RPCClient in the messaging library. This has a built-in retry mechanism to retry events for some period of time. If an external event hits a transient error in the external event handler, the flow engine should rely on this retry handling rather than implementing its own.

To do this, the external event handlers must always return a retry code if a transient error is encountered. It may be possible to do this easily using the flow external event library that these handlers are currently using, or it may require some adjustments to the messaging library code to allow for this case.

Completed:
- [x] `EntityRequestProcessor` 

Remaining (separate PRs):
- [ ]  `CryptoFlowOpsProcessor` 
- [ ] `LedgerPersistenceRequestProcessor` 
- [ ] `TokenSelectionSyncRPCProcessor` 
- [ ] `UniquenessCheckMessageProcessor` 
- [ ] `VerificationRequestProcessor`


libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/exception/CordaRPCAPIExceptions.kt
```diff
@@ -19,15 +19,3 @@ class CordaRPCAPIPartitionException(message: String?, exception: Throwable? = nu
  */
 class CordaRPCAPIResponderException(val errorType: String, message: String?, exception: Throwable? = null) :
     CordaRuntimeException(message, exception)
-
```
sprintengine, line 22: This file has exceptions used by the RPC send/receive message pattern. I did not want to mix the new exception types used in sync rpc/http message pattern so I renamed this back to "CordaRPCAPIExceptions" and created a new CordaHTTPExceptions containing client- and server-side exceptions for the sync rpc/http pattern.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/SyncRPCSubscriptionImpl.kt
```diff
@@ -90,11 +91,7 @@ internal class SyncRPCSubscriptionImpl<REQUEST : Any, RESPONSE : Any>(
                 val response = try {
                     processor.process(payload)
                 } catch (ex: Exception) {
-                    val errorMsg = "Failed to process RPC request for $rpcEndpoint"
-                    log.warn(errorMsg, ex)
-                    context.result(errorMsg)
-                    context.status(ResponseCode.INTERNAL_SERVER_ERROR)
-                    return@trace context
+                    return@trace handleProcessorException(log, endpoint, ex, context)
```
sprintengine: messaging code - handling transient exceptions here. I think this is more suitable than an exception handler registered into the Javalin webserver.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/SyncRPCUtils.kt
```diff
@@ -0,0 +1,33 @@
...
+    ex: Exception,
+    context: WebContext
+): WebContext {
+    when (ex) {
+        is CordaHTTPServerTransientException -> {
+            "Transient error processing RPC request for $endpoint: ${ex.message}".also { msg ->
+                log.warn(msg, ex)
+                context.result(msg)
+            }
+            context.status(ResponseCode.SERVICE_UNAVAILABLE)
```
sprintengine: How do you feel about a 503 for temporary transient errors on external processors?

components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/ResponseFactoryImpl.kt
```diff
@@ -44,7 +45,7 @@ class ResponseFactoryImpl @VisibleForTesting internal constructor(
 
     override fun errorResponse(externalEventContext : ExternalEventContext, exception: Exception) = when (exception) {
         is CpkNotAvailableException, is VirtualNodeException -> {
-            transientErrorResponse(externalEventContext, exception)
+            throw CordaHTTPServerTransientException(externalEventContext.requestId, exception)
```
jujoramos, line 48: Should we keep the warning message (previously logged within `transientErrorResponse() -> errorLogMessage()`)?.

components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/ResponseFactoryImpl.kt
```diff
@@ -56,22 +57,14 @@ class ResponseFactoryImpl @VisibleForTesting internal constructor(
             when (persistenceExceptionCategorizer.categorize(exception)) {
                 PersistenceExceptionType.FATAL -> fatalErrorResponse(externalEventContext, exception)
                 PersistenceExceptionType.PLATFORM -> platformErrorResponse(externalEventContext, exception)
-                PersistenceExceptionType.TRANSIENT -> transientErrorResponse(externalEventContext, exception)
+                PersistenceExceptionType.TRANSIENT -> throw CordaHTTPServerTransientException(externalEventContext.requestId, exception)
```
jujoramos, line 60: Should we keep the warning message (previously logged within `transientErrorResponse() -> errorLogMessage()`)?.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/SyncRPCUtils.kt
```diff
@@ -0,0 +1,33 @@
+package net.corda.messaging.utils
```
jujoramos: Can we use `HTTP` instead of `RPC` for names, exception messages, logging, etc. (across the whole `PR`, not just this file)?.

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/processor/SyncRPCProcessor.kt
```diff
@@ -13,6 +13,7 @@ interface SyncRPCProcessor<REQUEST, RESPONSE> {
      * @param request
      *
      * @return the result of the processing of the type RESPONSE
+     * @throws [CordaTransientServerException] if a transient exception occurred that can be retried by the client.
```
JamesHR3: I don't know that the kdocs in this interface do a great job of describing what a user of the messaging library must do to handle particular cases. I think here for example we should state clearly that if the processing throws a `CordaTransientServerException`, then the messaging library will go into retry loop in which the `process` method is retried with backoff. Clients therefore need to make sure that exception is thrown if an encountered error is transient.

As an aside, we should also state that implementors need to be idempotent.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/SyncRPCUtils.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import net.corda.messaging.api.exception.CordaHTTPServerTransientException
+import net.corda.rest.ResponseCode
+import net.corda.web.api.Endpoint
+import net.corda.web.api.WebContext
+import org.slf4j.Logger
+
+fun handleProcessorException(
```
JamesHR3: Shouldn't this be part of the sync RPC subscription class? I don't think we reuse it anywhere?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/SyncRPCUtils.kt
JamesHR3: I think that's probably the right option here.

components/persistence/persistence-service-common/src/main/kotlin/net/corda/persistence/common/ResponseFactoryImpl.kt
sprintengine: The original message read:
```
Exception occurred (type=$errorType) for flow-worker request ${flowExternalEventContext.requestId}
```

The messaging code in SyncRPCSubscriptionImpl logs a warning that reads:
```
Transient error processing RPC request for $endpoint: Transient server exception while processing request '$requestId'. Cause: ${cause?.message}
```

Not sure if we need both?

jujoramos: I didn't notice we were logging the warning in other place 👍

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/SyncRPCUtils.kt
sprintengine: My main concern was really just splitting the exception handling so I could test it in isolation but it's probably not necessary, I'll chuck it back into the messaging class

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/SyncRPCUtils.kt
sprintengine: Are you suggesting rename `SyncRPCProcessor` to `SyncHTTPProcessor`? I wouldn't be against that, maybe in a follow up PR however. Following James' comment below I removed this utils file anyway.

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/processor/SyncRPCProcessor.kt
sprintengine: > if the processing throws a CordaTransientServerException, then the messaging library will go into retry loop in which the process method is retried with backoff

The messaging library won't perform the retry, it translates the transient exception to a 503 status and lets the HTTP client perform retry with backoff. 

I've made a few updates to the kdocs, let me know what you think?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/utils/SyncRPCUtils.kt
jujoramos: 👍

JamesHR3 (approved):

jujoramos (approved):
LGTM

kyriathar (approved):
OK (db changes)

LWogan (approved):
LGTM - flow

PR #5368, opened 2024-01-03
CORE-18978 Remove Kafka schema validation on reconciliation

simon-johnson-r3:
It is unnecessary to validate the schema from Kafka (this is old Config which is about to be reconciled with the new schema if it is incorrect anyway), and doing so (unexpectedly) adds defaults even though we ask it not to. This breaks the reconciliation check on upgrades as the db worker believes new fields are already published to Kafka when they are not.

A separate JIRA ticket is raised to cover the problem with validate https://r3-cev.atlassian.net/browse/CORE-19186, it's not fixed here because we don't need it and it's much lower priority.


kyriathar (approved):
LGTM

PR #5377, opened 2024-01-04
CORE-18770 Add additional RPC metrics to the Flow Engine

ben-millar:
#### Overview:
Adding or modifying metrics to allow more granular analysis of the synchronous HTTP/RPC communication carried out by the flow engine.

#### Changes:
- Move RPC client-side metrics from the `RPCClient` to the `HTTPRetryExecutor` so all HTTP requests are logged, not just the final one executed by the retry handler;
- Add server-side metrics for external events to track processing time on the external worker.
- Narrowed generic type parameter of `HTTPRetryExecutor` from `HttpResponse<T>` to `HttpResponse<ByteArray>`


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/subscription/SyncRPCSubscriptionImpl.kt
```diff
@@ -103,6 +109,7 @@ internal class SyncRPCSubscriptionImpl<REQUEST : Any, RESPONSE : Any>(
                     context.status(ResponseCode.INTERNAL_SERVER_ERROR)
                 }
             }
+            recordMetric(rpcEndpoint, FAILED, startTime)
```
mbrkic-r3: Should it be `SUCCESS` here (might need more logic to handle serialization error above)?

ben-millar: Yeah, good point! Missed this in the rebase

mbrkic-r3 (approved):
LGTM

PR #5378, opened 2024-01-05
CORE-19130: Avoid sending to the null endpoint

yift-r3:
Trying to post a link-out message with null in the address will crash the producer and make the link manager unstable.

This case will be avoided by only sending something if we can find the correct address.

Tests run on https://ci02.dev.r3.com/blue/organizations/jenkins/Corda5%2Fcorda-e2e-tests-multi-cluster-tests/detail/release%2F5.2/139/tests (The test that failed is the one trying to set the protocol version to 2)


charlieR3 (approved):
LGTM

PR #5379, opened 2024-01-05
CORE-19203 Added MockRPCClient and associated factories

ben-millar:
### Description:
This PR adds a `MockRPCClient` based on the `MessagingClient` interface which simulates network delays through a configurable delay time and provides pre-defined responses, along with all of the associated factories required to create it and stub it into the `MultiSourceEventMediator` for isolated testing.


thiagoviana (dismissed):

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/factory/MessagingClientFactoryFactory.kt
```diff
@@ -25,4 +25,15 @@ interface MessagingClientFactoryFactory {
...
         id: String
     ) : MessagingClientFactory
+
+    /**
+     * Creates a mock RPC messaging client factory
+     *
+     * @param id The identifier of this client.
+     * @param delayTime The simulated delay time (in milliseconds) for the clients send() method.
+     */
+    fun createMockRPCClientFactory(
```
LWogan, line 35: I think it might be better to move this code out of the production library./

I'd suggest creating a new project, maybe under the :testing folder.

In this project we could have implementation services for the various components that need to be stubbed out. e,g this one, consumers, producer, vnode, cpiInfo, membershipGroupReader etc.

These services would implement the same APIs, in this case, specifically, MessagingClientFactoryFactory.createRPCClientFactory. 

Then in the final test/app the new library could be made a runtimeOnly dependency. This would make these services available on the classpath. However it wouldn't necessarily mean the correct impl of the APIs would be chosen. To ensure yours gets chose you could set this annotation on the new service.
`@ServiceRanking(Int.MAX_VALUE)`

LWogan (commented):
Stub looks good to me, but i think we need to figure out how this will fit together with the final solution. Perhaps we need a discussion with @mbrkic-r3 about what the plan is here.

PR #5382, opened 2024-01-05
ES-1755: expand use of version catalog file for managing plugin versions

ronanbrowne:
This PR adds several plugins to the version catalog and removes logic applying them via the `setting.gradle` file, and associated gradle.properties versions, as these are now redundant. Also remove direct application of artifactory plugin as this is also redundant as it's applied by our R3 publishing plugin. 

This aligns with our approach in corda-api, and allows for these plugins to then take advantage of dependabot updates, which scan the `toml` file.

Note, due to the `settings.grade` file order of evaluation in the Build life cycle and how the Gradle Enterprise plugins need to be declared, we need to parse the toml file in settings.gradle to retrieve the version.


vkolomeyko (approved):
LGTM

driessamyn (approved):
LGTM

mcgovc (approved):
LGTM

PR #5387, opened 2024-01-08
CORE-19024 Add a new state manager api that allows for a state to be created or updated if it already exists

LWogan:
Add a new state manager api that allows for a state to be created or updated if it already exists

This new API will act similar to a "put" type function.

This function will be used later in the mediator to write async outputs to the DB. 

This means we can overwrite bus bound outputs from a previous run of the consumer processors input when the state (checkpoint) fails to write due to optimistic locking failures.


libs/state-manager/state-manager-db-impl/src/integrationTest/kotlin/net/corda/libs/statemanager/impl/tests/StateManagerIntegrationTest.kt
```diff
@@ -271,6 +271,39 @@ class StateManagerIntegrationTest {
         )
     }
 
+    @ValueSource(ints = [1, 5, 10, 20, 50])
+    @ParameterizedTest(name = "can createOrUpdate existing states (batch size: {0})")
+    fun canCreateOrUpdateExistingStates(stateCount: Int) {
```
sprintengine, line 276: Let's add some multi-threaded optimistic locking tests too, for cases where threads are trying to createOrUpdate the same states, and where threads are trying to createOrUpdate some states with overlapping states being deleted (similar to `optimistic locking ensures no exceptions when double deletes across threads` and `optimistic locking ensures no double updates across threads`

sprintengine: I'd quite like to see how the query handles conflicts when the state was already deleted, will it treat it as a create?

LWogan: there's no optimistic locking in this function though. We either create or overwrite.

LWogan: > I'd quite like to see how the query handles conflicts when the state was already deleted, will it treat it as a create?

If its deleted then the data doesn't exist anymore and it will just execute a create.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -32,6 +30,20 @@ interface StateManager : Lifecycle {
...
+     * A single transactional context is used when interacting with the underlying persistent storage,
+     * so all these states will be persisted/updated or none will.
+     *
+     * Control is only returned to the caller once all [states] that were successfully created/updated have been fully
+     * persisted and replicas of the underlying persistent storage, if any, are synced.
+     *
+     * @param states Collection of states to be persisted.
+     * @return Collection of keys for all those states that could not be persisted on the underlying persistent storage.
+     */
+    fun createOrUpdate(states: Collection<State>): Set<String>
```
jujoramos: I agree with @conalsmith-r3 's previous comment, we shouldn't add methods to the State Manager where optimistic locking is ignored... it breaks the contract of the approved design API and will cause hard to troubleshoot issues if multiple threads or processes execute the method at the same time.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -19,10 +19,8 @@ interface StateManager : Lifecycle {
     /**
      * Persist new [states].
      *
-     * Every operation uses its own transactional context when interacting with the underlying persistent storage and,
-     * as a result, some states might have been successfully persisted and some might have not.
-     * It's the responsibility of calling API to decide whether the operation can be retried or not, based on the
-     * [Exception] returned for the relevant key.
+     * A single transactional context is used when interacting with the underlying persistent storage,
+     * so all these states will be persisted or none will.
```
jujoramos, line 23: This is not entirely true, is it?. We use a single transactional context, sure, but the current query returns those keys for which the creation failed so we can have executions on which some states were successfully created and some were not (`StateManagerIntegrationTest.failuresWhenPersistingSomeStatesDoesNotHaltTheEntireBatch()` covers this scenario).

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/impl/StateRepositoryImpl.kt
```diff
@@ -34,6 +34,26 @@ class StateRepositoryImpl(private val queryProvider: QueryProvider) : StateRepos
...
+                statement.setInt(indices.next(), state.version)
+                statement.setString(indices.next(), state.metadata)
+            }
+            statement.execute()
+            val results = statement.resultSet
+            sequence<String> {
+                while (results.next()) {
+                    yield(results.getString(CREATE_RESULT_COLUMN_INDEX))
+                }
+            }.toList()
```
sprintengine: Looks like the sonar check wants less code duplication in this class

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
JamesHR3: One of the challenges with this API is figuring out what to return. I wonder if this API might be more consistent if it returned `Map<String, State?>`? That's because the only way this can go wrong is if there is an old state and it failed the optimistic locking check, or there's no state and the initial version is wrong. That brings this API in line with the `update` one, with the same semantics on the output.

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -43,9 +43,20 @@ class StateManagerImpl(
...
+    override fun createOrUpdate(states: Collection<State>): Set<String> {
+        if (states.isEmpty()) return emptySet()
+        val successfulKeys = dataSource.connection.transaction { connection ->
+            stateRepository.createOrUpdate(connection, states.map { it.toPersistentEntity() })
+        }
+        return getFailedCreates(states, successfulKeys.toSet())
+    }
+
+    private fun getFailedCreates(states: Collection<State>, successfulKeys: Set<String>) =
+        states.map { it.key }.toSet() - successfulKeys
```
JamesHR3: I don't think you can treat failures the same as the create case. I think you also need to figure out what states failed the optimistic locking check where the original state exists, as in the update case.

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
LWogan: > That's because the only way this can go wrong is if there is an old state and it failed the optimistic locking check,

We can't fail optimistic locking here. We actually want to overwrite the state in that case. But you're right the return type doesn't make sense here. I'll need to think about it. For my use case i don't care if it was a create or update.

LWogan: The closest thing this is comparable to is maps put() functionality. this returns null if its a new value and returns the old value if its an update. Perhaps it would be best copy that idea and return Map<String, State>. where the map contains the old values for updates and no entries for creates.

LWogan: For my use case, there is a possibility where a value already exists. if it does we want to ignore it.

LWogan:
I'll hold off making further changes until this is discussed offline

LWogan:
Closing PR for now. I still think this API has merit, but I don't need it.

PR #5389, opened 2024-01-08
ES-1764/Custom log4j layout

mag1c-48:
The timestamp in the JSON logs of our current layout is given in seconds since epoch, which is not human readable, and hard for QA to analyse by looking at. Making a change to this would include adding more 'overrides' to log4j2.xml file, instead we have opted for creating our own template which now gives us the freedom to modift whatever we might need in the json logs. this can be found in `cordaLayout.json`


charts/corda-lib/templates/_helpers.tpl
```diff
@@ -481,6 +481,20 @@ data:
...
 
+{{/*
+Config map for Log4J json template
+*/}}
+{{- define "corda.log4jJsonTemplateConfigMap" -}}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cordaLayout.json
```
davidcurrie: The resource name can't be fixed as we permit the Helm chart to be installed multiple times in the same namespace. I'd be tempted to just add it to the existing log4jConfigMap though rather than creating another one.

charts/corda/cordaLayout,json
```diff
@@ -0,0 +1,82 @@
+{
```
davidcurrie: filename appears to have a comma as the separator

davidcurrie (changes requested):
PR appears to be missing anything to mount the file from the config map into the container so I'm surprised it works for multiple reasons!

PR #5390, opened 2024-01-08
CORE-18535: Update event mediator to remove shared mutable state and remove dead code

JamesHR3:
### Problem description

As part of the flow retry and timeout enhancements, any timed out synchronous event should cause processing of the corresponding events to fail and the state to be marked as failed. At a later point, it will additionally be required to time out the event processing in the short running task if this goes on too long and also mark the state as failed. These two operations require modification of the same state which is currently being processed in different threads. Making these changes in the existing code will leave a situation where the change is difficult to test and could fail in some edge cases.

### Solution

Refactor the event mediator to do the following:
- In the event processor, return an object containing the new state, any asynchronous events, and whether the state operation is a create/update/delete
- In the consumer processor, merge all the event processor results together and handle accordingly.
- Remove the `ConsumerProcessorState`
- Ensure `Metadata` is merged in `StateManagerHelper` rather than overwritten.
- Eliminate dead code.

This change also adds the proposed shared metadata key, called `FAILED_STATE`. This is introduced in the API module so that clients using the state manager to inspect states it owns can use it to identify anything that has failed.

Note that this change does not include the work to fail processing if the client times out - this will come in a follow up PR.


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessingOutput.kt
```diff
@@ -0,0 +1,20 @@
...
+
+sealed interface StateUpdate {
+    val outputState: State?
+
+    class Create(override val outputState: State) : StateUpdate
+    class Update(override val outputState: State) : StateUpdate
+    class Delete(override val outputState: State) : StateUpdate
+    object Noop : StateUpdate {
+        override val outputState: State? = null
+    }
```
jujoramos: Is this `Noop` needed?, is there ever a case where none of the actual conditions are met?:
```
            val stateUpdate = when {
                state == null && processed != null -> StateUpdate.Create(processed)
                state != null && processed != null -> StateUpdate.Update(processed)
                state != null && processed == null -> StateUpdate.Delete(state)
            }
```

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/ConsumerProcessor.kt
```diff
@@ -137,21 +136,30 @@ class ConsumerProcessor<K : Any, S : Any, E : Any>(
...
+    private fun processOutputs(outputs: Map<String, EventProcessingOutput>): Map<String, State> {
+        val statesToCreate = mutableListOf<State>()
+        val statesToUpdate = mutableListOf<State>()
+        val statesToDelete = mutableListOf<State>()
+        outputs.values.forEach {
+            when (it.stateUpdate) {
+                is StateUpdate.Create -> statesToCreate.add(it.stateUpdate.outputState)
+                is StateUpdate.Update -> statesToUpdate.add(it.stateUpdate.outputState)
+                is StateUpdate.Delete -> statesToDelete.add(it.stateUpdate.outputState)
+                is StateUpdate.Noop -> {} // Do nothing.
```
jujoramos: Similar to my other comment, is this needed?. If it is, shouldn't we log a warning  signalling that something went wrong?.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessingOutput.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import net.corda.libs.statemanager.api.State
+import net.corda.messaging.api.mediator.MediatorMessage
+
+data class EventProcessingOutput(
+    val asyncOutputs: List<MediatorMessage<Any>>,
+    val stateUpdate: StateUpdate
+)
+
+sealed interface StateUpdate {
```
jujoramos: Nitpick: maybe rename this one to something more meaningful?, `StateOperationResult`?, something else?.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessor.kt
```diff
@@ -45,79 +42,54 @@ class EventProcessor<K : Any, S : Any, E : Any>(
...
+                asyncOutputs.addAll(asyncEvents)
+                val returnedMessages = processSyncEvents(groupEntry.key, syncEvents)
+                queue.addAll(returnedMessages)
+            }
+            val processed = stateManagerHelper.createOrUpdateState(groupKey, state, processorState)
+            val stateUpdate = when {
+                state == null && processed != null -> StateUpdate.Create(processed)
+                state != null && processed != null -> StateUpdate.Update(processed)
+                state != null && processed == null -> StateUpdate.Delete(state)
+                else -> StateUpdate.Noop
```
jujoramos: Similar to my other comment, does this ever happen?.

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/constants/MetadataKeys.kt
```diff
@@ -0,0 +1,14 @@
+package net.corda.messaging.api.constants
```
jujoramos: Nitpick: I've added something similar as part of https://r3-cev.atlassian.net/browse/CORE-18539, both `MetadataKeys` and `FAILED_STATE` seem to "general" (we might want to differentiate between timeouts and failed processing, as an example). Maybe the following?:
```
package net.corda.messaging.api.constants

import net.corda.libs.statemanager.api.State

/**
 * These are the keys, stored within the [State.metadata], that are set by the messaging layer.
 * These keys are meant to be managed by messaging layer itself, and can also be used as a mean of communicating
 * certain conditions to the message processors.
 *
 *
 */
object StateMetadataKeys {
    /**
     * Set by the messaging layer whenever a task takes too long to complete and must be cancelled to prevent the
     * processor from hanging or running indefinitely (consumer kicked out of the consumer group when the message bus
     * implementation is Kafka, as an example). It might also imply, for cases on which the processor is leveraging
     * external provided code, that the user code is poorly written,  doesn't handle error scenarios gracefully or
     * simply takes too long to execute.
     */
    const val TASK_TIME_OUT = "task.time.out"

/**
 * Indicates that message processing failed for this state.
 *
 * Clients may wish to perform cleanup on states that have failed message processing. For those that wish to do so, a
 * lookup of states with this key set should be performed. The existence of this key in the metadata indicates that
 * processing of the state failed.
 */
const val FAILED_STATE = "messaging-state-failure"
}
```

jujoramos (commented):
LGTM, just some minor comments/questions.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessingOutput.kt
JamesHR3: Yeah this can happen if both input and output state are `null`. The calculator app would hit this on every execution for example, as it goes through the full flow in a single execution.

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/constants/MetadataKeys.kt
JamesHR3: They're both timeouts really. Mine might be a worse name!

I think from a client perspective it doesn't matter why it failed, just that it did. There's no recovery from either case, it's just a matter of cleaning up. (Flows for example will want to be marked as failed.) My opinion then is that we should use the same key. If yours is in first then let's go with that :).

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessingOutput.kt
JamesHR3: I guess this represents both what the new state is and what operation should be carried out. `StateChangeAndOperation`?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessingOutput.kt
jujoramos: Worth adding a comment maybe?

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/constants/MetadataKeys.kt
jujoramos: Sounds good!. What about the class name, though?, considering that we might have other `Metadata` stuff in the future, worth renaming it to `StateMetadataKeys`?.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessingOutput.kt
jujoramos: Agreed, `StateChangeAndOperation` sounds better 👍

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/constants/MetadataKeys.kt
JamesHR3: `MessagingMetadataKeys`? (I've put it in the messaging API)

jujoramos: Agreed! 👍

jujoramos (dismissed):
LGTM

JamesHR3:
build e2e

JamesHR3:
build e2e

jujoramos (approved):

LWogan (approved):
LGTM

PR #5393, opened 2024-01-08
CORE-19194: implement subset operation on Merkle proofs

dickon:
Includes test coverage. It's not viable to test too many permutations considering all subsets of all proofs up to 16 leaf trees so we use a predictable shuffle order to get some coverage.

Fix bug with MerkleProof.calculateRootInstrumented which showed up only on larger trees.

FIx bugs in Merkle Tree rendering which occur on larger trees, and also sometimes drew the trees out in strange ways that made it look like they had more nodes than they actually do.

Introduce MerkleNode dataclass rather than Pair, for clarity, suggested by Matthew.

### DIFF
--- a/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNode.kt
+++ b/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNode.kt
@@ -3,10 +3,11 @@ package net.corda.crypto.merkle.impl
 import net.corda.v5.crypto.SecureHash
 
 /**
- * A MerkleNode represents a node on a tree
+ * A MerkleNode represents a node on a tree, and is used as the parameter for callbacks when we walk
+ * a proof of Merkle tree - see MerkleProofImpl.calculateRootInstrumented.
  *
- * @param index the position of the node at its level, counting from the left (which is zero)
+ * @param indexWithinLevel the position of the node at its level, counting from the left (which is zero)
  * @param hash the hash of the node, which is a hash of its children
  */
 
-data class MerkleNode(val index: Int, val hash: SecureHash)
+internal data class MerkleNode(val indexWithinLevel: Int, val hash: SecureHash)
--- a/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNode.kt
+++ b/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNode.kt
@@ -6,8 +6,8 @@ import net.corda.v5.crypto.SecureHash
  * A MerkleNode represents a node on a tree, and is used as the parameter for callbacks when we walk
  * a proof of Merkle tree - see MerkleProofImpl.calculateRootInstrumented.
  *
- * @param indexWithinLevel the position of the node at its level, counting from the left (which is zero)
- * @param hash the hash of the node, which is a hash of its children
+ * @param indexWithinLevel The position of the node at its level, counting from the left (which is zero)
+ * @param hash The hash of the node, which is a hash of its children
  */
 
 internal data class MerkleNode(val indexWithinLevel: Int, val hash: SecureHash)
--- a/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNode.kt
+++ b/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNode.kt
@@ -1,0 +1,12 @@
+package net.corda.crypto.merkle.impl
+
+import net.corda.v5.crypto.SecureHash
+
+/**
+ * A MerkleNode represents a node on a tree
+ *
+ * @param index the position of the node at its level, counting from the left (which is zero)
+ * @param hash the hash of the node, which is a hash of its children
+ */
+
+data class MerkleNode(val index: Int, val hash: SecureHash)
--- a/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNodeInfo.kt
+++ b/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNodeInfo.kt
@@ -4,7 +4,7 @@ package net.corda.crypto.merkle.impl
  * Represents additional  information about a node
  *
  * @param node The node itself
- * @param node The level within the Merkle tree, counting from the top with 0 being the top
+ * @param level The level within the Merkle tree, counting from the top with 0 being the top
  * @param consumed If set, the index in the proof hashes used for this node. If unset, the node is calculated.
  */
 internal data class MerkleNodeInfo(val node: MerkleNode, val level: Int, val consumed: Int?)
--- a/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNodeInfo.kt
+++ b/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNodeInfo.kt
@@ -1,0 +1,10 @@
+package net.corda.crypto.merkle.impl
+
+/**
+ * Represents additional  information about a node
+ *
+ * @param node The node itself
+ * @param node The level within the Merkle tree, counting from the top with 0 being the top
+ * @param consumed If set, the index in the proof hashes used for this node. If unset, the node is calculated.
+ */
+internal data class MerkleNodeInfo(val node: MerkleNode, val level: Int, val consumed: Int?)
--- a/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNodeInfo.kt
+++ b/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNodeInfo.kt
@@ -1,7 +1,7 @@
 package net.corda.crypto.merkle.impl
 
 /**
- * Represents additional  information about a node
+ * Represents additional information about a node
  *
  * @param node The node itself
  * @param level The level within the Merkle tree, counting from the top with 0 being the top
--- a/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
+++ b/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
@@ -62,9 +62,8 @@ class MerkleProofImpl(
      *                          The arguments are the hash of the node, level (with top of the tree at 0),
      *                          index (position across from the left hand side of the tree, starting at 0),
      *                          plus the index of that hash within the incoming hashes (starting at 0),
-     *                          or null for calculated nodes. This will be called left to right then bottom to top.
-     *                          The order that this is called is left to right then bottom to top, i.e. the same
-     *                          order as input hashes are consumed.
+     *                          or null for calculated nodes. This will be called left to right then bottom to top,
+     *                          i.e. the same order as input hashes are consumed.
      * @return the secure hash of the root of the Merkle proof
      */
     @Suppress("NestedBlockDepth", "ThrowsCount")
--- a/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
+++ b/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
@@ -230,18 +230,24 @@ class MerkleProofImpl(
 
     /**
      *
-     * Work out a Merkle proof with fewer leaves.
+     * Work out a Merkle proof with specified leaves.
      *
-     * NOTE: currently this is not complete
+     * Throws IllegalArgumentException if some requeted leaf indices are not available in the source proof,
+     * or if no leaf indices are requested.
      *
      * @param leafIndices indices  of the known leaves to include in the output proof
-     * @return A new Merkle proof
+     * @return A new Merkle proof covering the specified leaves.
      */
     fun subset(digest: MerkleTreeHashDigest, leafIndices: List<Int>): MerkleProofImpl {
+        val outLeaves = leaves.filter { it.index in leafIndices }
+        require(outLeaves.size == leafIndices.size) { "some leaves are not available in input proof"}
+        require(outLeaves.size != 0) { "output proof must have at least one known leaf"}
+        // We work out the hashes for the new subset proof by considering, for the original proof, each hash that
+        // is calculated when we verify the proof by calculate the root.
         val outHashes: MutableList<SecureHash> = mutableListOf()
         val treeDepth = MerkleTreeImpl.treeDepth(treeSize)
         calculateRootInstrumented(digest) { hash, level, index, _ ->
-            val adjacentIndex = (index xor 1) // the adjacent node for this level
+            val adjacentIndex = index xor 1 // the adjacent node for this level
             val height = treeDepth - level // how many levels above the leaves, 0 for being at the leaf
             // The subset proof will need this hash if and only if the adjacent sub-tree is a known hash.
             // The adjacent sub-tree (be it leaf or node) will be known if and only if there is known data within it.
@@ -253,7 +259,6 @@ class MerkleProofImpl(
             val adjLHS = adjacentIndex shl height
             if (leafIndices.any { it in adjLHS until adjLHS + (1 shl height) }) outHashes.add(hash)
         }
-        val outLeaves = leaves.filter { it.index in leafIndices }
         return MerkleProofImpl(proofType, treeSize, outLeaves, outHashes)
     }
 
--- a/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
+++ b/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
@@ -105,10 +105,10 @@ class MerkleProofImpl(
         var hashIndex = 0
         val sortedLeaves = leaves.sortedBy { it.index }
         // work out nodeHashes, which is a list of node information for the current level we operate at
-        var nodeHashes: List<Pair<Int, SecureHash>> =
-            sortedLeaves.map { Pair(it.index, digest.leafHash(it.index, it.nonce, it.leafData)) }
+        var nodeHashes: List<MerkleNode> =
+            sortedLeaves.map { MerkleNode(it.index, digest.leafHash(it.index, it.nonce, it.leafData)) }
         var treeDepth = MerkleTreeImpl.treeDepth(treeSize)         // initialised to the depth of tree we should
-        nodeHashes.forEach { item -> onNewHash(item.second, treeDepth, item.first, null)  }
+        nodeHashes.forEach { item -> onNewHash(item.hash, treeDepth, item.index, null)  }
         // need for the number of elements
         var currentSize = treeSize                                 // outer loop variable; the number of
         // leaves left as we roll up the tree
@@ -133,7 +133,7 @@ class MerkleProofImpl(
             // - $treeDepth is the level of the tree, counting from the root of the tree where $currentSize==1
             // - $nodeHashes has a list of pairs of the index and hash of the node. We checked we have some content.
 
-            val newItems = mutableListOf<Pair<Int, SecureHash>>()   // this will become nodeHashes at the end of this
+            val newItems = mutableListOf<MerkleNode>()   // this will become nodeHashes at the end of this
             // out iteration
 
             // Now walk over the hashes at this tree level, striding over 1 or 2 at a time
@@ -146,22 +146,22 @@ class MerkleProofImpl(
                 //
                 // Since index == item.first we don't really need to use item.first
 
-                if (item.first < currentSize and 0x7FFFFFFE) {      // If the level has odd elements, we'll process
+                if (item.index < currentSize and 0x7FFFFFFE) {      // If the level has odd elements, we'll process
                     // the last element later.
                     if (index < nodeHashes.size - 1) {              // If there is a next element...
                         val next = nodeHashes[index + 1]
                         // Decide if we can consume the next two elements since they are adjacent in the Merkle tree
-                        if (item.first xor next.first == 1) {       // ... and they are a pair with the current
+                        if (item.index xor next.index == 1) {       // ... and they are a pair with the current
                             // We now know that the indices ${item.first} and ${next.first} only differ on the bottom bit,
                             // i.e. they are adjacent. Therefore, we can combine them.
 
                             // So, make a single new item, computing a new hash
                             // (Pair is the Kotlin type, nothing to do with pairing nodes)
                             // in the original tree, we create their parent.
-                            val newHash = digest.nodeHash(treeDepth, item.second, next.second)
-                            val newIndex = item.first / 2
+                            val newHash = digest.nodeHash(treeDepth, item.hash, next.hash)
+                            val newIndex = item.index / 2
                             onNewHash( newHash, treeDepth, newIndex, null)
-                            newItems += Pair(newIndex, newHash)
+                            newItems.add(MerkleNode(newIndex, newHash))
                             // and record that we consumed two values from our working set, and skip on to the
                             // start of the next loop
                             index += 2                              // we've consumed two
@@ -179,35 +179,35 @@ class MerkleProofImpl(
                             "MerkleProof root calculation requires more hashes than the proof has."
                         )
                     }
-                    val newIndex = item.first /2
+                    val newIndex = item.index /2
 
                     // We pair the current element with a hash from the proof
-                    newItems += if ((item.first and 1) == 0) {      // Even index means, that the item is on the left
+                    newItems.add(if ((item.index and 1) == 0) {      // Even index means, that the item is on the left
                         // Make new node with
-                        //   - left being current element, index $item.first, hash $item.second
+                        //   - left being current element
                         //   - right being a consumed incoming hash from $hashes[$hashIndex]
                         //
                         // Also remember we used $hashIndex by bumping the counter
-                        onNewHash(hashes[hashIndex], treeDepth+1, item.first + 1, hashIndex)
-                        val newHash = digest.nodeHash(treeDepth, item.second, hashes[hashIndex++])
-                        onNewHash( newHash, treeDepth, newIndex, null)
-                        Pair(newIndex, newHash)
+                        onNewHash(hashes[hashIndex], treeDepth+1, item.index + 1, hashIndex)
+                        val newHash = digest.nodeHash(treeDepth, item.hash, hashes[hashIndex++])
+                        onNewHash(newHash, treeDepth, newIndex, null)
+                        MerkleNode(newIndex, newHash)
                     } else {
                         // Make new node with:
                         //   - left being proof of hash at $hashIndex
                         //   - right being current element, index $item.first, hash $item.second
                         //
                         // Also remember we used hashIndex by bumping the counter.
-                        onNewHash(hashes[hashIndex], treeDepth+1, item.first-1, hashIndex)
-                        val newHash = digest.nodeHash(treeDepth, hashes[hashIndex++], item.second)
-                        onNewHash( newHash, treeDepth, newIndex, null)
-                        Pair(newIndex, newHash)
-                    }
+                        onNewHash(hashes[hashIndex], treeDepth+1, item.index-1, hashIndex)
+                        val newHash = digest.nodeHash(treeDepth, hashes[hashIndex++], item.hash)
+                        onNewHash(newHash, treeDepth, newIndex, null)
+                        MerkleNode(newIndex, newHash)
+                    })
                 } else {
-                    val newIndex = (item.first + 1) / 2
-                    onNewHash( item.second, treeDepth, newIndex, null)
+                    val newIndex = (item.index + 1) / 2
+                    onNewHash( item.hash, treeDepth, newIndex, null)
                     // The last odd element, just gets lifted.
-                    newItems += Pair(newIndex, item.second)
+                    newItems += MerkleNode(newIndex, item.hash)
                 }
                 ++index // whatever of the last 3 cases we took, we consumed one element
             }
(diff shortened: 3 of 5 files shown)


libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -229,6 +229,50 @@ class MerkleProofImpl(
...
+     * Throws IllegalArgumentException if some requeted leaf indices are not available in the source proof,
+     * or if no leaf indices are requested.
+     *
+     * @param leafIndices indices  of the known leaves to include in the output proof
+     * @return A new Merkle proof covering the specified leaves.
+     */
+    fun subset(digest: MerkleTreeHashDigest, leafIndices: List<Int>): MerkleProofImpl {
+        val outLeaves = leaves.filter { it.index in leafIndices }
+        require(outLeaves.size == leafIndices.size) { "some leaves are not available in input proof"}
+        require(outLeaves.size != 0) { "output proof must have at least one known leaf"}
```
mnesbit: Use .isNotEmpty generally for this

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -229,6 +229,50 @@ class MerkleProofImpl(
...
+        // we will build up a set of avaialble leaves as we track. This will start out with the known leaves
```
mnesbit: available

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
```diff
@@ -377,23 +453,27 @@ class MerkleTreeTest {
             }
         }
 
+        val rng = Random(0)
+
         // Test all the possible combinations of leaves for the proof.
```
mnesbit: Comment now incorrect

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
```diff
@@ -377,23 +453,27 @@ class MerkleTreeTest {
             }
         }
 
+        val rng = Random(0)
```
mnesbit, line 464: Be warned that Kotlin's Random algorithm can change, though not often

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: fixed

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: done

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
dickon: yeah, and hopefully as part of a controlled update so we'd notice if we happen to hit new bugs that way.

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
dickon: clarified

blsemo (dismissed):
LGTM

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -58,9 +58,9 @@ class MerkleProofImpl(
...
      *
      * @param digest the digest service to use, which must generate hashes compatible with the [hashes] constructor parameter.
-     * @param foundHashCallback called on each node that has a node taken from incoming hashes or calculated during the proof.
+     * @param onNewHash called on each node that has a node taken from incoming hashes or calculated during the proof.
      *                          The arguments are the hash of the node, level (with top of the tree at 0),
-     *                          index (position across from the left had side of the tree, starting at 0),
+     *                          index (position across from the left hand side of the tree, starting at 0),
      *                          plus the index of that hash within the incoming hashes (starting at 0),
      *                          or null for calculated nodes. This will be called left to right then bottom to top.
      *                          The order that this is called is left to right then bottom to top, i.e. the same
```
vlajos: These two lines look like duplicated partially.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -229,6 +229,50 @@ class MerkleProofImpl(
...
+            val height = treeDepth - level // how many levels above the leaves, 0 for being at the leaf
+            val leftmostLeaf = index shl height
+            val afterLeaf = min(leftmostLeaf + (1 shl height), treeSize)
+            val unknowns = (leftmostLeaf until afterLeaf).any { it !in availableLeaves }
+            val allUnknown = (leftmostLeaf until afterLeaf).all { it !in availableLeaves }
+            val siblingIndex = index xor 1 // work out the sibling node i.e. the node which shares a direct parent node with this one
+            val siblingLeftmostLeaf = siblingIndex shl height
+            val afterSiblingLeaf = min((siblingIndex + 1) shl height, treeSize)
+            val siblingWillBeKnown = (siblingLeftmostLeaf until afterSiblingLeaf).any { it in availableLeaves }
+            // we need this hash in the subset proof iff it covers unknown content ($unknowns is true),
```
```suggestion
            // we need this hash in the subset proof if it covers unknown content ($unknowns is true),
```

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -154,14 +154,15 @@ class MerkleProofImpl(
...
 
-                            val newHash = digest.nodeHash(treeDepth, item.second, next.second)
-                            foundHashCallback( newHash, treeDepth, newItems.size, null)
                             // So, make a single new item, computing a new hash
                             // (Pair is the Kotlin type, nothing to do with pairing nodes)
                             // in the original tree, we create their parent.
-                            newItems += Pair(item.first / 2, newHash)
+                            val newHash = digest.nodeHash(treeDepth, item.second, next.second)
+                            val newIndex = item.first / 2
+                            onNewHash( newHash, treeDepth, newIndex, null)
```
anien: Now we have 3rd parameter of `onNewHash` being `item.first / 2`, before it was `newItems = mutableListOf<Pair<Int, SecureHash>>()`. I don't understand this well, so just checking the change is required.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: "Iff" is actually a word used sometimes in mathematics to mean "if and only if" - see https://en.wikipedia.org/wiki/If_and_only_if 

however it is more readable to write "if and only if" so I'll do that.

Fixed.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: Yes it was a bug before, which actually showed up only  at tree size 15, after more than 30 minutes of CPU time.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: Fixed

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
vlajos: Oh. Thanks. :)

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: I agree it was confusing and there was no explanation of why we divided by 2 before which is partly why I accidentally introduced a bug here. I've restructured the code following a suggestions from @mnesbit to use a dataclass for the Merkle nodes, and made that an argument to `onNewHash`, and I think that makes the code easier to read.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNode.kt
```diff
@@ -0,0 +1,12 @@
...
+import net.corda.v5.crypto.SecureHash
+
+/**
+ * A MerkleNode represents a node on a tree
+ *
+ * @param index the position of the node at its level, counting from the left (which is zero)
+ * @param hash the hash of the node, which is a hash of its children
+ */
+
+data class MerkleNode(val index: Int, val hash: SecureHash)
```
mnesbit: I am not totally sure about the name, but OK. However, this definitely needs to be an internal class. I think index is a bit to vague and perhaps we call it levelIndex?

dickon: done (as indexWithLevel, to avoid confusion with level)

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -106,10 +100,10 @@ class MerkleProofImpl(
         var hashIndex = 0
         val sortedLeaves = leaves.sortedBy { it.index }
         // work out nodeHashes, which is a list of node information for the current level we operate at
-        var nodeHashes: List<Pair<Int, SecureHash>> =
-            sortedLeaves.map { Pair(it.index, digest.leafHash(it.index, it.nonce, it.leafData)) }
+        var nodeHashes: List<MerkleNode> =
+            sortedLeaves.map { MerkleNode(it.index, digest.leafHash(it.index, it.nonce, it.leafData)) }
         var treeDepth = MerkleTreeImpl.treeDepth(treeSize)         // initialised to the depth of tree we should
-        nodeHashes.forEach { item -> foundHashCallback(item.second, treeDepth, item.first, null)  }
+        nodeHashes.forEach { item -> onNewHash(MerkleNodeInfo(item, treeDepth, null))  }
```
vlajos: Would this be the same?
```
        var treeDepth = MerkleTreeImpl.treeDepth(treeSize)         // initialised to the depth of tree we should
        var nodeHashes: List<MerkleNode> =
            sortedLeaves.map {
                MerkleNode(it.index, digest.leafHash(it.index, it.nonce, it.leafData))
                      .also(onNewHash(MerkleNodeInfo(it, treeDepth, null)))
            }
```
?

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -179,37 +174,39 @@ class MerkleProofImpl(
...
                     }
-                    val newIndex = item.first /2
+
+                    // Make up a new node, which will be a level up so will have a node index shifted right one
+                    val newIndex = item.indexWithinLevel / 2
 
                     // We pair the current element with a hash from the proof
-                    newItems += if ((item.first and 1) == 0) {      // Even index means, that the item is on the left
+                    val newNode = (if ((item.indexWithinLevel and 1) == 0) {      // Even index means, that the item is on the left
+                        // Remember we consumed a proof hash for the left hand child
```
vlajos: Are left and right not mixed in these comments? (*)

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -179,37 +174,39 @@ class MerkleProofImpl(
...
-                        //
-                        // Also remember we used $hashIndex by bumping the counter
-                        foundHashCallback(hashes[hashIndex], treeDepth+1, item.first + 1, hashIndex)
-                        val newHash = digest.nodeHash(treeDepth, item.second, hashes[hashIndex++])
-                        foundHashCallback( newHash, treeDepth, newIndex, null)
-                        Pair(newIndex, newHash)
+                        val newHash = digest.nodeHash(treeDepth, item.hash, hashes[hashIndex])
+                        MerkleNode(newIndex, newHash)
                     } else {
+                        // Remember we consumed a proof hash for the right hand child
```
vlajos: And this one is not again. (*)

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -179,37 +174,39 @@ class MerkleProofImpl(
...
 
                     // We pair the current element with a hash from the proof
-                    newItems += if ((item.first and 1) == 0) {      // Even index means, that the item is on the left
+                    val newNode = (if ((item.indexWithinLevel and 1) == 0) {      // Even index means, that the item is on the left
+                        // Remember we consumed a proof hash for the left hand child
+                        onNewHash(MerkleNodeInfo(MerkleNode(item.indexWithinLevel+1, hashes[hashIndex]), treeDepth+1, hashIndex))
                         // Make new node with
-                        //   - left being current element, index $item.first, hash $item.second
+                        //   - left being current element
                         //   - right being a consumed incoming hash from $hashes[$hashIndex]
```
vlajos, line 188: This looks correct.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNodeInfo.kt
```diff
@@ -0,0 +1,10 @@
+package net.corda.crypto.merkle.impl
+
+/**
+ * Represents additional  information about a node
+ *
+ * @param node The node itself
+ * @param node The level within the Merkle tree, counting from the top with 0 being the top
```
vlajos:
```suggestion
 * @param level The level within the Merkle tree, counting from the top with 0 being the top
```

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -179,37 +174,39 @@ class MerkleProofImpl(
...
-                    foundHashCallback( item.second, treeDepth, newIndex, null)
                     // The last odd element, just gets lifted.
-                    newItems += Pair(newIndex, item.second)
+                    val newIndex = (item.indexWithinLevel + 1) / 2
+                    val newNode = MerkleNode(newIndex, item.hash)
+                    onNewHash(MerkleNodeInfo(newNode, treeDepth, null))
+                    newItems += MerkleNode(newIndex, item.hash)
                 }
-                ++index // whatever of the last 3 cases we took, we consumed one element
+                index++ // whatever of the last 3 cases we took, we consumed one element
```
vlajos: I wonder if this would be more understandable or not:
```suggestion
                index++ // whatever of the last 3 cases we took, we consumed one node
```
Just to distinguish it from the input hashes.
I am not too sure tbh.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -226,7 +223,52 @@ class MerkleProofImpl(
...
+        // We work out the hashes for the new subset proof by considering, for the original proof, each hash that
+        // is calculated when we verify the proof by calculating the root.
+        val outHashes: MutableList<SecureHash> = mutableListOf()
+        val treeDepth = MerkleTreeImpl.treeDepth(treeSize)
+        calculateRootInstrumented(digest) { info ->
+            val height = treeDepth - info.level // how many levels above the leaves, 0 for being at the leaf
+            val leftmostLeaf = info.node.indexWithinLevel shl height
+            val afterLeaf = min(leftmostLeaf + (1 shl height), treeSize)
+            val unknowns = (leftmostLeaf until afterLeaf).any { it !in availableLeaves }
+            val allUnknown = (leftmostLeaf until afterLeaf).all { it !in availableLeaves }
```
```suggestion
            val allUnknown = unknows.size == (afterLeaf-leftMostLeaf) 
```
?

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -226,7 +223,52 @@ class MerkleProofImpl(
...
+            // work out the sibling node i.e. the node which shares a direct parent node with this one
+            val siblingIndex = info.node.indexWithinLevel xor 1
+            val siblingLeftmostLeaf = siblingIndex shl height
+            val afterSiblingLeaf = min((siblingIndex + 1) shl height, treeSize)
```
vlajos, line 261: This calculation is not fully symmetric with https://github.com/corda/corda-runtime-os/pull/5393/files#diff-097e799c831ce16be1cfa92a78a730a5eb62fb2e03183e3e093a714c817ab358R252-R253
Is this intentional?
```suggestion
            val siblingLeftmostLeaf = siblingIndex shl height
            val afterSiblingLeaf = min(siblingLeftmostLeaf + (1 shl height), treeSize)
```
Both look correct for me and doing the same, just maybe it would be nicer to do that in the same way.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNode.kt
```diff
@@ -0,0 +1,13 @@
+package net.corda.crypto.merkle.impl
+
+import net.corda.v5.crypto.SecureHash
+
+/**
+ * A MerkleNode represents a node on a tree, and is used as the parameter for callbacks when we walk
+ * a proof of Merkle tree - see MerkleProofImpl.calculateRootInstrumented.
+ *
+ * @param indexWithinLevel the position of the node at its level, counting from the left (which is zero)
+ * @param hash the hash of the node, which is a hash of its children
```
anien:
```suggestion
 * @param indexWithinLevel The position of the node at its level, counting from the left (which is zero)
 * @param hash The hash of the node, which is a hash of its children
```

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNodeInfo.kt
```diff
@@ -0,0 +1,10 @@
+package net.corda.crypto.merkle.impl
+
+/**
+ * Represents additional  information about a node
```
anien:
```suggestion
 * Represents additional information about a node
```

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
```diff
@@ -29,12 +30,21 @@ import org.junit.jupiter.api.Test
...
+import kotlin.test.assertFailsWith
 
 class MerkleTreeTest {
     companion object {
         val digestAlgorithm = DigestAlgorithmName.SHA2_256D
+        private val logger = LoggerFactory.getLogger(this::class.java.enclosingClass)
+
+        // Since there are 2^(2*n) permutations of source leafs we don't want to test too many,
+        // so we do a few at random. In local testing I've taken this up to 1000 which takes 30 minutes
+        // on my laptop.
```
anien:
```suggestion
        // Since there are 2^(2*n) permutations of source leafs we don't want to test too many,
        // so we do a few at random. In local testing it's been taken this up to 1000 which takes 30 minutes
        // on the laptop.
```

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
```diff
@@ -332,17 +340,88 @@ class MerkleTreeTest {
...
+        val proof3 = proof.subset(trivialHashDigestProvider, listOf(2))
+        assertThat(proof3.render(trivialHashDigestProvider)).isEqualToIgnoringWhitespace("""
+                00000612 (calc)┳0000069F (calc)   ┳00000630 (input 1)┳unknown            filtered
+                               ┃                  ┃                  ┗unknown            filtered
+                               ┃                  ┗00000634 (calc)   ┳00000002 (calc)    known leaf
+                               ┃                                     ┗00000003 (input 0) filtered
+                               ┗00000638 (input 2)━unknown           ┳unknown            filtered
+                                                                     ┗unknown            filtered
+        """)
+    }
```
vlajos, line 394: I think the subset tests should compare the derived proof's roots with the originals'.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -179,37 +174,39 @@ class MerkleProofImpl(
...
-                    val newIndex = item.first /2
+
+                    // Make up a new node, which will be a level up so will have a node index shifted right one
+                    val newIndex = item.indexWithinLevel / 2
 
                     // We pair the current element with a hash from the proof
-                    newItems += if ((item.first and 1) == 0) {      // Even index means, that the item is on the left
+                    val newNode = (if ((item.indexWithinLevel and 1) == 0) {      // Even index means, that the item is on the left
+                        // Remember we consumed a proof hash for the left hand child
+                        onNewHash(MerkleNodeInfo(MerkleNode(item.indexWithinLevel+1, hashes[hashIndex]), treeDepth+1, hashIndex))
```
anien:
```suggestion
                        onNewHash(MerkleNodeInfo(MerkleNode(item.indexWithinLevel + 1, hashes[hashIndex]), treeDepth + 1, hashIndex))
```

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -179,37 +174,39 @@ class MerkleProofImpl(
...
-                        // Also remember we used $hashIndex by bumping the counter
-                        foundHashCallback(hashes[hashIndex], treeDepth+1, item.first + 1, hashIndex)
-                        val newHash = digest.nodeHash(treeDepth, item.second, hashes[hashIndex++])
-                        foundHashCallback( newHash, treeDepth, newIndex, null)
-                        Pair(newIndex, newHash)
+                        val newHash = digest.nodeHash(treeDepth, item.hash, hashes[hashIndex])
+                        MerkleNode(newIndex, newHash)
                     } else {
+                        // Remember we consumed a proof hash for the right hand child
+                        onNewHash(MerkleNodeInfo(MerkleNode(item.indexWithinLevel-1, hashes[hashIndex]), treeDepth+1, hashIndex))
```
anien:
```suggestion
                        onNewHash(MerkleNodeInfo(MerkleNode(item.indexWithinLevel - 1, hashes[hashIndex]), treeDepth + 1, hashIndex))
```

anien (commented):
Just a few minor comments, which I am happy to be done later.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: Unknowns is a boolean so we can't get size on it.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: done

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: fixed

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: I did that and took it a bit further

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: yes, fixed

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleNodeInfo.kt
dickon: fixed

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: I think that's a bit more confusing.

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
dickon: I'd like to add this in a later PR if you don't mind

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
vlajos: Eh. Sorry. Anyway, the idea would have been to iterate through the range only once.
Maybe something like this?
```
val numberOfUnknowns = (leftmostLeaf until afterLeaf).count { it !in availableLeaves }
val allUnknown = numberOfUnknowns == (afterLeaf-leftMostLeaf) 
val unknowns = numberOfUnknonws > 0
```

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
vlajos: Then, this approach in the other place too?

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: Possibly in a later PR.

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
dickon: done

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: actually, yes that's better, done.

I don't think it is necessarily the case that we can't do any better but clarity and correctness first.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: Oh you mean
>             val afterLeaf = min(leftmostLeaf + (1 shl height), treeSize)

and
>             val afterSiblingLeaf = min((siblingIndex + 1) shl height, treeSize)

they don't actually have the same value, I did check, what's here is correct. But maybe could use more comments.

libs/crypto/merkle-impl/src/test/kotlin/net/corda/crypto/merkle/impl/MerkleTreeTest.kt
dickon: I did it for this one test, which is better than never checking.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: So for instance
```
>>> print( (5+1) << 3)
48
>>> print ( 5 << (1+3))
80
```

i.e. left bitshift does not have the associativity you think it does.

vlajos (approved):
The code looks correct.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
vlajos: I meant that `afterSiblingLeaf` and `afterLeaf` are derived differently from their inputs.
Their expression structures and operators could be the same. (with different arguments)

PR #5395, opened 2024-01-09
CORE-18305: Enable concurrent membership smoke tests

yift-r3:
I tried to run it a lot of times, in the pipeline and locally, and the tests passed each time


dimosr (approved):
LGTM

PR #5396, opened 2024-01-09
CORE-18535: Fail processing if the retry limit is exceeded and mark state with failed metadata key

JamesHR3:
### Problem description

If event mediator processing needs to send an event to another part of the system, it does so via a synchronous HTTP request. If this fails with a transient problem, the HTTP client will retry for a period of time. If the failure does not clear in the required period, then an intermittent error is thrown. This leads to an issue where an error can cause a message to be continuously replayed as the subscription is reset and consumption restarts at the event with the problem.

### Solution

If the RPC client fails, instead fail processing of the event(s) on this key and mark the corresponding state as failed.

This is implemented entirely within the event processor. The state update that is outputted from the event processor is modified in the event that RPC processing fails.

This allows the messaging client to identify states that have failed to process.


jujoramos (approved):
LGTM

ben-millar (approved):
LGTM

PR #5398, opened 2024-01-09
CORE-18929 Adding deterministic request ID to external events

ben-millar:
### Context:
To support failure recovery, the flow engine now caches output events so that if something goes wrong during the pipeline execution, we can rollback and transmit our outputs without having to re-generate them. However, if there is a failure during the transmission of events, there's a possibility that some events will be retransmitted to external workers after we recover.

External workers rely on the `requestId` of incoming events to perform de-duplication and ensure idempotence. Currently, this `requestId` is a random UUID which is generated upon creation of the external event. We should modify this `requestId` to be deterministic, so that replayed events will not be re-processed unnecessarily by external workers.

### Changes:
Rather than a random UUID, our `requestId` is now generated as a hash of:
```
"$flowId-$hashOfInputParameters-$suspensionCount"
```
We can see below that this key is reproducible for a replayed request:

image

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
@@ -64,8 +64,11 @@ class ExternalEventExecutorImpl @Activate constructor(
     }
 
     private fun generateRequestId(uuid: UUID): String {
-        val executionContext = flowFiberService.getExecutingFiber().getExecutionContext()
-        val flowCheckpoint = executionContext.flowCheckpoint
+        val flowCheckpoint = flowFiberService
+            .getExecutingFiber()
+            .getExecutionContext()
+            .flowCheckpoint
+
         val flowId = flowCheckpoint.flowId
         val suspendCount = flowCheckpoint.suspendCount
 
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
@@ -33,7 +33,7 @@ class ExternalEventExecutorImpl @Activate constructor(
         return with(flowFiberService.getExecutingFiber()) {
             suspend(
                 FlowIORequest.ExternalEvent(
-                    requestId.toString(),
+                    requestId,
                     factoryClass,
                     parameters,
                     externalContext(this)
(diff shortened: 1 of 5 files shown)


components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
```diff
@@ -47,4 +49,26 @@ class ExternalEventExecutorImpl @Activate constructor(
...
+        val flowCheckpoint = flowFiberService
+            .getExecutingFiber()
+            .getExecutionContext()
+            .flowCheckpoint
+
+        val flowId = flowCheckpoint.flowId
+        val suspendCount = flowCheckpoint.suspendCount
+
+        return UUID
+            .nameUUIDFromBytes("$flowId-$uuid-$suspendCount".toByteArray())
```
LWogan, line 69: I think i prefer the ID being something a human can understand from the logs and gleam some info when debugging. i.e $flowId-$hash-$suspendCount

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
```diff
@@ -23,10 +24,11 @@ class ExternalEventExecutorImpl @Activate constructor(
         factoryClass: Class<out ExternalEventFactory<PARAMETERS, RESPONSE, RESUME>>,
         parameters: PARAMETERS
     ): RESUME {
-        // `requestId` is a unique id per event. It is used to achieve idempotency by de-duplicating events processing,
-        // on Kafka consumers side. Consuming duplicate events can happen from retrying an event from Kafka which however
-        // did some persistent work previously but did not fully succeed (Kafka was not notified), therefore we retry/ reprocess it.
-        val requestId = UUID.randomUUID().toString()
+        // `requestId` is a deterministic ID per event which allows us to achieve idempotency by de-duplicating events processing;
```
LWogan, line 27: A unit test to verify the requestId is deterministic would be nice.

Youd need to mock the flowFiberService.getExecutingFiber() and then verify that the context of the requestid in the  ExternalEvent argument passed to mockedFiber.suspend() is what we expect. Have a look at some other unit tests that use argument captors for inspiration. Hopefully it is possible

LWogan (commented):
LGTM lets try get a unit test

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/executor/ExternalEventExecutorImpl.kt
ben-millar: Discussed offline: IDs of this length caused issues with downstream components in testing. The core need for this change is to make IDs deterministic; if we need additional trace functionality we should do that elsewhere.

LWogan (approved):
LGTM

PR #5399, opened 2024-01-09
CORE-18539: Improve Flow Clean Up

jujoramos:
- Automatically clean up and mark as "FAILED" flows for which the
  state hasn't been updated in a configurable amount of time.
- Automatically clean up and mark as "FAILED" those flows for which a
  problem was detected while processing the underlying event and the
  messaging layer marked the relevant state as a failure.


components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/FlowTimeoutTaskProcessor.kt
```diff
@@ -0,0 +1,78 @@
...
+    private val maxIdleTimeMilliseconds = config.getLong(FlowConfig.PROCESSING_MAX_IDLE_TIME)
+
+    override fun onNext(events: List<Record<String, ScheduledTaskTrigger>>): List<Record<*, *>> {
+        // If we receive multiple, there's probably an issue somewhere, and we can ignore all but the last one.
+        return events.lastOrNull {
+            it.key == ScheduledTask.SCHEDULED_TASK_NAME_SESSION_TIMEOUT
+        }?.value?.let { trigger ->
+            logger.trace("Processing trigger scheduled at {}", trigger.timestamp)
+            // TODO - temporary query
+            // TODO - we must be able to limit by type of state
```
JamesHR3, line 48: Are these still valid? (You're unlucky here in renaming the file!)

JamesHR3 (approved):
One question but otherwise looks good.

components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/FlowTimeoutTaskProcessor.kt
jujoramos: They are, yeah (I didn't update the comments, but the "temporary query" is because we are executing two calls to the state manager instead of one; and "limit by type of state" is related to the separation of state managers).

conal-smith (approved):

ben-millar (approved):

PR #5400, opened 2024-01-09
CORE-18534 remaining external event processors

conal-smith:
Note - This is the 2nd PR ([1st](https://github.com/corda/corda-runtime-os/pull/5366)) to update external event processors to throw CordaHTTPServerTransientExceptions  for transient errors. It includes:
- [x] `LedgerPersistenceRequestProcessor` 
- [x] `TokenSelectionSyncRPCProcessor` 
- [x] `VerificationRequestProcessor` 

**Problem description:**
All flow engine external events go through the RPCClient in the messaging library. This has a built-in retry mechanism to retry events for some period of time. If an external event hits a transient error in the external event handler, the flow engine should rely on this retry handling rather than implementing its own.

To do this, the external event handlers must always return a 503 retry code if a transient error is encountered. The messaging library handles this, the processor throws the exception mentioned above.

Remaining (separate PRs):
- [ ]  `CryptoFlowOpsProcessor` 
- [ ] `UniquenessCheckMessageProcessor`

### DIFF
--- a/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/processor/LedgerPersistenceRequestProcessor.kt
+++ b/components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/processor/LedgerPersistenceRequestProcessor.kt
@@ -3,10 +3,13 @@ package net.corda.ledger.persistence.processor
 import net.corda.crypto.core.parseSecureHash
 import net.corda.data.flow.event.FlowEvent
 import net.corda.data.ledger.persistence.LedgerPersistenceRequest
+import net.corda.flow.external.events.responses.exceptions.CpkNotAvailableException
+import net.corda.flow.external.events.responses.exceptions.VirtualNodeException
 import net.corda.flow.utils.toMap
 import net.corda.ledger.persistence.common.InconsistentLedgerStateException
 import net.corda.ledger.persistence.common.UnsupportedLedgerTypeException
 import net.corda.ledger.persistence.common.UnsupportedRequestTypeException
+import net.corda.messaging.api.exception.CordaHTTPServerTransientException
 import net.corda.messaging.api.processor.SyncRPCProcessor
 import net.corda.metrics.CordaMetrics
 import net.corda.persistence.common.EntitySandboxService
@@ -18,8 +21,6 @@ import net.corda.utilities.translateFlowContextToMDC
 import net.corda.utilities.withMDC
 import net.corda.v5.application.flows.FlowContextPropertyKeys.CPK_FILE_CHECKSUM
 import net.corda.virtualnode.toCorda
-import org.slf4j.Logger
-import org.slf4j.LoggerFactory
 import java.time.Duration
 
 /**
@@ -36,21 +37,23 @@ class LedgerPersistenceRequestProcessor(
     override val requestClass = LedgerPersistenceRequest::class.java
     override val responseClass = FlowEvent::class.java
 
-    private companion object {
-        val log: Logger = LoggerFactory.getLogger(LedgerPersistenceRequestProcessor::class.java)
-    }
+    private val transientExceptions = setOf(
+        CpkNotAvailableException::class.java,
+        VirtualNodeException::class.java
+    )
 
     override fun process(request: LedgerPersistenceRequest): FlowEvent {
         val startTime = System.nanoTime()
         val clientRequestId =
             request.flowExternalEventContext.contextProperties.toMap()[MDC_CLIENT_ID] ?: ""
         val holdingIdentity = request.holdingIdentity.toCorda()
+        val requestId = request.flowExternalEventContext.requestId
 
         val result =
             withMDC(
                 mapOf(
                     MDC_CLIENT_ID to clientRequestId,
-                    MDC_EXTERNAL_EVENT_ID to request.flowExternalEventContext.requestId
+                    MDC_EXTERNAL_EVENT_ID to requestId
                 ) + translateFlowContextToMDC(request.flowExternalEventContext.contextProperties.toMap())
             ) {
                 try {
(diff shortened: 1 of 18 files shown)


JamesHR3 (dismissed):
Probably wants the relevant teams to double check the processors aren't adversely impacted, but it looks reasonable to me.

filipesoliveira (dismissed):

components/ledger/ledger-verification/src/main/kotlin/net/corda/ledger/verification/processor/impl/VerificationRequestProcessor.kt
```diff
@@ -55,6 +61,9 @@ class VerificationRequestProcessor(
                     currentSandboxGroupContext.set(sandbox)
                     requestHandler.handleRequest(sandbox, request)
                 } catch (e: Exception) {
+                    if (transientExceptions.contains(e::class.java)) {
+                        throw CordaHTTPServerTransientException(request.flowExternalEventContext.requestId, e)
+                    }
                     errorResponse(request.flowExternalEventContext, e)
```
sprintengine: Just noticed this is still returning transient error response, will address that

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/processor/LedgerPersistenceRequestProcessor.kt
```diff
@@ -36,21 +37,23 @@ class LedgerPersistenceRequestProcessor(
     override val requestClass = LedgerPersistenceRequest::class.java
     override val responseClass = FlowEvent::class.java
 
-    private companion object {
-        val log: Logger = LoggerFactory.getLogger(LedgerPersistenceRequestProcessor::class.java)
-    }
+    private val transientExceptions = setOf(
+        CpkNotAvailableException::class.java,
+        VirtualNodeException::class.java
```
lankydan: The ledger code already handles these errors, see `ResponseFactoryImpl`

lankydan: So either that needs to change or we have some duplicated code.

sprintengine: Yeah nice spot!

components/ledger/ledger-persistence/src/test/kotlin/net/corda/ledger/persistence/processor/impl/LedgerPersistenceRequestProcessorTest.kt
```diff
@@ -58,8 +63,7 @@ class LedgerPersistenceRequestProcessorTest {
         }
     }
 
-    @BeforeEach
-    fun setup() {
+    private fun defaultSetup() {
```
jujoramos: Why not keep the `@BeforeEach` and override the required mocks to throw the needed exception instead of using `defaultSetup()`?.

components/ledger/ledger-verification/src/test/kotlin/net/corda/ledger/verification/processor/impl/VerificationRequestProcessorTest.kt
```diff
@@ -59,8 +67,7 @@ class VerificationRequestProcessorTest {
             )
         }
 
-    @BeforeEach
-    fun setup() {
+    private fun defaultSetup() {
```
jujoramos: I think we should keep the `@BeforeEach` and override the required mocks to throw the needed exception instead.

jujoramos (dismissed):

filipesoliveira (approved):

anien (approved):
Ok (crypto)

jujoramos (approved):
LGTM

components/ledger/ledger-verification/src/main/kotlin/net/corda/ledger/verification/processor/impl/VerificationRequestProcessor.kt
```diff
@@ -69,16 +80,9 @@ class VerificationRequestProcessor(
...
-            log.error(errorMessage(externalEventContext, ExternalEventResponseErrorType.PLATFORM), exception)
-            responseFactory.platformError(externalEventContext, exception)
-        }
-
-        else -> {
-            log.warn(errorMessage(externalEventContext, ExternalEventResponseErrorType.TRANSIENT), exception)
-            responseFactory.transientError(externalEventContext, exception)
-        }
+    private fun platformErrorResponse(externalEventContext: ExternalEventContext, exception: Exception): Record<String, FlowEvent> {
+        log.error(errorMessage(externalEventContext, ExternalEventResponseErrorType.PLATFORM), exception)
```
LWogan, line 84: I think this should be warn (i think it was wrong in the previous code too)

LWogan (commented):
Looks fine to me from flow perspective

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/processor/LedgerPersistenceRequestProcessor.kt
lankydan: I'm not sure the changes you've made are correct here, unless I'm missing something.

From the Pr it looks like you want a new exception type to be thrown for transient exceptions, but the transient exception handling in response factory handles it slightly differently. So I'm not sure the behaviour matches up exactly the same.

I don't think I explained it very well initially. If I've misunderstood something then let me know.

lankydan (approved):

LWogan (approved):
approved given the promise to fix the log line in a follow up

PR #5401, opened 2024-01-09
CORE-19229 - Remove transactions from StateManager read operations.

driessamyn:


ifrankrui (commented):
First of all, whether a read operation executes within an explicit transaction or not, it still runs within a transaction context. The database implicitly starts a transaction for each statement, even if it's not explicitly wrapped in a `BEGIN` and `COMMIT` block. This is a standard behavior in PostgreSQL ([PostgreSQL Transactions Documentation](https://www.postgresql.org/docs/current/tutorial-transactions.html#TUTORIAL-TRANSACTIONS)).

Secondly, regarding the default isolation level, our specification in Corda is set to `Read Committed` ( [Corda Runtime OS - TransactionIsolationLevel.kt](https://github.com/corda/corda-runtime-os/blob/5d2a0b0d3df790fb30a9f8fde75f4bf04ab70a06/libs/db/db-orm/src/main/kotlin/net/corda/orm/TransactionIsolationLevel.kt#L24)). This matches the default isolation level of PostgreSQL, which is also `Read Committed`. Therefore, from a data integrity perspective, using the default settings does not compromise the data integrity here.

Regarding the intention to remove the transaction for the read operation, using `use { ... }` might indeed be more efficient and simpler, potentially leading to some performance gains. While the impact might be minimal for a few statements, it could be more measurable for a large number of select queries.

Omar-awad (approved):
LGTM

jujoramos (approved):
LGTM

ifrankrui (approved):

PR #5405, opened 2024-01-10
CORE-18537: Add a timeout to event processing

JamesHR3:
### Problem statement

When the event mediator processes events, the mediator will currently wait indefinitely for event processing to finish. If this takes longer than the poll interval then the process will be kicked out the consumer group, leading to instability. This can mostly be avoided by preventing code from blocking indefinitely inside the poll loop, but for flows some executed code is not in the platform's control. In these cases it remains possible for this condition to be hit.

### Solution

Add a timeout to event processing of a group in the event mediator.

This timeout is currently hardcoded. If the timeout is exceeded, processing for that group fails and the corresponding states are marked with a metadata key indicating that processing has failed.

Note that this is only a mitigation and is not foolproof. This is because each event processing thread is blocked on in turn, and so if all fail the timeout the total time waiting will be longer than the individual timeout. A choice of 30s should still keep the total time down below the 5 minutes without a poll required by Kafka provided the number of processing threads is sufficiently small.


jujoramos (approved):
LGTM

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/ConsumerProcessor.kt
```diff
@@ -107,11 +116,25 @@ class ConsumerProcessor<K : Any, S : Any, E : Any>(
...
-                    taskManager.executeShortRunningTask {
+                    val future = taskManager.executeShortRunningTask {
                         eventProcessor.processEvents(group, statesToProcess)
                     }
-                }.map {
-                    it.join()
+                    Pair(future, group.keys)
+                }.map { (future, keys) ->
+                    try {
+                        future.getOrThrow(Duration.ofMillis(EVENT_PROCESSING_TIMEOUT_MILLIS))
```
LWogan, line 125: Should this be configurable via the corda api?

JamesHR3: I'd like to avoid introducing any new config, I already think the configuration in this area is much too complicated to tune effectively. I think we have other options here.

I could use the `processorTimeout` value, which is already there and semantically means the same thing. This is currently set to 1.5s though as a default, which I think is likely too short.

The other approach would be to configure this as some fraction of `max.poll.interval.ms`, which only makes sense if it's Kafka behind the scenes. I think I would prefer doing this and falling back to some default. I think that's something that can be achieved in a follow up if it turns out that this is too short.

LWogan: Hard coded is fine for now. We have a mediator section in the subscription config now, so i think a new value there would make sense in the future.

LWogan (approved):
Looks ok from flow perspective

PR #5406, opened 2024-01-10
CORE-18174: CLI tool change for `initial-rbac`

BenYip123:
Add permission to CLI `initial-rbac` to change user passwords


tools/plugins/initial-rbac/src/main/kotlin/net/corda/cli/plugin/initialRbac/commands/UserAdminSubcommand.kt
```diff
@@ -29,6 +30,8 @@ class UserAdminSubcommand : RestCommand(), Callable<Int> {
         "CreateUsers" to "POST:/api/$VERSION_PATH_REGEX/user",
         "GetUsersV1" to "GET:/api/v1/user\\?loginName=$USER_URL_REGEX",
         "GetUsers" to "GET:/api/$VERSION_PATH_REGEX/user/$USER_URL_REGEX",
+        "ChangeSelfUserPassword" to "POST:/api/$VERSION_PATH_REGEX/user/selfpassword",
```
vkolomeyko: I do not think we need explicit permission for `ChangeSelfUserPassword` as it should be granted to any user.

filipesoliveira (dismissed):

vkolomeyko (approved):
LGTM

PR #5407, opened 2024-01-10
CORE-18691 Implementation of store and fetch merkle proofs

nkovacsx:
### Overview

Add new requests to the persistence worker to persist and fetch merkle proofs.


libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/transaction/MerkleProofDto.kt
```diff
@@ -0,0 +1,23 @@
...
+ * - Size of the original Merkle tree the proof was created from
+ * - List of the revealed hashes
+ */
+@CordaSerializable
+data class MerkleProofDto(
+    val transactionId: String,
+    val groupIndex: Int,
+    val treeSize: Int,
+    val hashes: List<SecureHash>,
+    val leavesWithData: Map<Int, ByteArray>
```
dickon: Do we need the optional nonces  in `leavesWithData`?

Maybe `leavesWithData` should be of type `List<IndexedMerkleLeaf>`.

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/merkle/MerkleProofFactory.kt
```diff
@@ -0,0 +1,26 @@
...
+import net.corda.v5.base.annotations.DoNotImplement
+import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.crypto.SecureHash
+import net.corda.v5.crypto.merkle.MerkleProof
+
+/**
+ * A service used to instantiate [MerkleProof]s.
+ */
+@DoNotImplement
+interface MerkleProofFactory {
```
dickon, line 12: This can, and I think probably should, be a functional SAM interace.

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/merkle/MerkleProofFactory.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * Create a new [MerkleProof] instance using the given data.
+     */
+    @Suspendable
```
dickon, line 25: Needs param definitions in kdoc

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/UtxoRepository.kt
```diff
@@ -161,4 +162,28 @@ interface UtxoRepository {
         signedGroupParameters: SignedGroupParameters,
         timestamp: Instant
     )
+
+    /** Persists a merkle proof and returns its ID */
+    fun persistMerkleProof(
```
dickon, line 168:
```suggestion
    @Suppress("LongParameterList")
    fun persistMerkleProof(
```

components/ledger/ledger-utxo-flow/src/main/kotlin/net/corda/ledger/utxo/flow/impl/persistence/UtxoLedgerPersistenceServiceImpl.kt
```diff
@@ -151,6 +162,31 @@ class UtxoLedgerPersistenceServiceImpl @Activate constructor(
         }
     }
 
+    @Suspendable
+    override fun findMerkleProofs(
```
dickon, line 166: You'll need

@Suppress("LongParameterList", "TooManyFunctions")

on the UtxoLedgerPersistenceServiceImpl class since with this addition we are now over the threshold for detekt warnings.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -119,4 +119,69 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
...
+    /**
+     * This query will join the Merkle proof table and the component group table together to find the leaf data the
+     * Merkle proof has revealed.
+     *
+     * Each Merkle proof has an ID created by concatenating the following properties:
+     * - Transaction ID
+     * - Group Index
+     * - Revealed leaf indices joined to a string
+     *
+     * The indices that the Merkle proof reveals are stored in a join-table. That's the reason we need a sub-query
```
dickon, line 132: I don't see why we can't do a 3-way join and why we are saying that joint able requires a sub-query.

I'd expect (and some v quick googling reinforces this) that subqueries are slowed than multi-way joins.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
```diff
@@ -119,4 +119,69 @@ abstract class AbstractUtxoQueryProvider : UtxoQueryProvider {
...
+                utc.leaf_idx,
+                utc.data
+            FROM {h-schema}utxo_transaction_merkle_proof utmp
+            JOIN {h-schema}utxo_transaction_component utc
+                on utc.transaction_id = utmp.transaction_id
+                and utc.group_idx = utmp.group_idx
+            WHERE utmp.transaction_id = :transactionId
+                and utmp.group_idx = :groupIndex
+                and utc.leaf_idx in (
+                	SELECT utmpl.leaf_index
```
dickon: So I think this would be better as:

```
            SELECT
                utmp.merkle_proof_id,
                utc.transaction_id,
                utc.group_idx,
                utmp.tree_size,
                utmp.hashes,
                utc.leaf_idx,
                utc.data
            FROM {h-schema}utxo_transaction_merkle_proof utmp
            JOIN {h-schema}utxo_transaction_component utc
                on utc.transaction_id = utmp.transaction_id
                and utc.group_idx = utmp.group_idx
            JOIN {h-schema}utxo_transaction_merkle_proof_leaves utmpl
                on utmpl.merkle_proof_id = utmp.merkle_proof_id
            WHERE utmp.transaction_id = :transactionId
                and utmp.group_idx = :groupIndex
```

(which does pass UtxoPersistenceServiceImplTests and is likely to be faster due to avoiding table scans in some places)

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofFactoryImpl.kt
```diff
@@ -0,0 +1,36 @@
...
+import net.corda.v5.base.annotations.Suspendable
+import net.corda.v5.crypto.SecureHash
+import net.corda.v5.crypto.merkle.MerkleProof
+import net.corda.v5.crypto.merkle.MerkleProofType
+import net.corda.v5.serialization.SingletonSerializeAsToken
+import org.osgi.service.component.annotations.Activate
+import org.osgi.service.component.annotations.Component
+import org.osgi.service.component.annotations.ServiceScope
+
+@Component(
```
dickon, line 14: (Aside: I think this could actually not be OSGi and we just construct MerkleProofImpl  directly, with other arrangements (possibly involving other refactoring). If Ledger is using OSGi components for many classes and for the test setup this change doesn't seem the place to depart from that)

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
nkovacsx: yeah that's a good point, thanks

libs/ledger/ledger-utxo-data/src/main/kotlin/net/corda/ledger/utxo/data/transaction/MerkleProofDto.kt
nkovacsx: Good point. But in order to instantiate a `IndexedMerkleLeaf` we'd need to add a `merkle-impl` dependency to the persistence worker which I think is not a good idea so I'll use a different representation here.

libs/crypto/cipher-suite/src/main/kotlin/net/corda/crypto/cipher/suite/merkle/MerkleProofFactory.kt
nkovacsx: The problem is, we don't have access to `merkle-impl` in `UtxoLedgerPersistenceServiceImpl` and we probably shouldn't.

components/ledger/ledger-persistence/src/main/kotlin/net/corda/ledger/persistence/utxo/impl/AbstractUtxoQueryProvider.kt
dickon: Nit: we should now update the comment to say that we use a join rather than a subquery.

dickon (dismissed):
LGTM with detekt changes

dickon (dismissed):
LGTM (recent changes, on top of what I already approved)

blsemo (dismissed):
LGTM

dickon (approved):
LGTM (API bump plus earlier review)

PR #5409, opened 2024-01-10
CORE-19247: Workaround Time Overflow

jujoramos:
Use "Instant.EPOCH" instead of "Instant.MIN" when filtering by
interval of time through State Manager operations.


thiagoviana (approved):
LGTM (as a temporary solution).

PR #5413, opened 2024-01-10
CORE-17032: State Manager Metrics

jujoramos:
Add metrics to record the time taken by each State Manager operation
and verify that they are correctly recorded within integration tests.

### DIFF
--- a/libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
+++ b/libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
@@ -8,6 +8,12 @@ import net.corda.libs.statemanager.api.MetadataFilter
 import net.corda.libs.statemanager.api.State
 import net.corda.libs.statemanager.api.StateManager
 import net.corda.libs.statemanager.impl.lifecycle.CheckConnectionEventHandler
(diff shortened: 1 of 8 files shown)


ronanbrowne (approved):
LGTM (gradle change only)

JamesHR3 (approved):

libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/StateManagerImpl.kt
```diff
@@ -99,44 +81,91 @@ class StateManagerImpl(
...
+        return metricsRecorder.recordProcessingTime(UPDATE) {
+            try {
+                val (_, failedUpdates) = dataSource.connection.transaction { conn ->
+                    stateRepository.update(conn, states.map { it.toPersistentEntity() })
+                }
+
+                if (failedUpdates.isEmpty()) {
+                    emptyMap()
+                } else {
+                    getFailedUpdates(failedUpdates)
```
sprintengine, line 116: Should we add a metric for number of optimistic locking failures?

jujoramos: Probably yes, but in a follow up `PR` as this one has been merged already 😅

PR #5414, opened 2024-01-10
ENT-1764/Log4j json timestamp edited

mag1c-48:
The log4j json format timestamp previously was in seconds since Epoch

Which is not human readable, this has been altered and converted into UTC format.

JSON format timestamp: `"timestamp":"2024-01-11 14:36:11.785"}`

Text format timestamp has been updated to add `yyyy-MM-dd` 
Text format timestamp: `2024-01-11 14:42:30.618`


charts/corda/log4j2.xml
```diff
@@ -7,26 +7,37 @@
...
+                      }
+                    }
+                  }'/>
+                <EventTemplateAdditionalField
+                        key="timestamp"
+                        format="JSON"
+                        value='{
+                    "$resolver": "timestamp",
+                    "pattern": {
+                        "format": "yyyy-MM-dd HH:mm:ss.SSS",
```
davidcurrie: This should be in the ISO 8601 format that you had originally i.e. `yyyy-MM-dd'T'HH:mm:ss.SSS'Z'`

charts/corda/log4j2.xml
```diff
@@ -7,26 +7,37 @@
...
+                        "timeZone": "UTC"
+                        }
+                    }
+                '/>
             </JsonTemplateLayout>
         </Console>
         <Console name="text" target="SYSTEM_OUT">
-            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} %X - %msg%n"/>
+            <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %-5level %logger{36} %X - %msg%n"/>
```
davidcurrie: I believe this should give you the equivalent of what we have in the JSON layout but it needs testing:
```suggestion
            <PatternLayout pattern="%d{ISO8601}{UTC} [%t] %-5level %logger{36} %X - %msg%n"/>
```

davidcurrie (changes requested):

charts/corda/log4j2.xml
```diff
@@ -7,26 +7,35 @@
...
+                        "format": "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'",
+                        "timeZone": "UTC"
+                    }
+                }]]></value>
+                </EventTemplateAdditionalField>
             </JsonTemplateLayout>
         </Console>
         <Console name="text" target="SYSTEM_OUT">
-            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} %X - %msg%n"/>
+            <PatternLayout pattern="%d{ISO8601}{UTC} [%t] %-5level %logger{36} %X - %msg%n"/>
```
davidcurrie: Sorry to mess you around, but looking at the output from the build, I realise that ISO8601 differs from RFC3339 in the preference for a comma rather than a period (although both are supported) and that it also does not require the timezone specifier. Can we use the following to be consistent across the JSON and text formats?
```suggestion
            <PatternLayout pattern="%d{yyyy-MM-dd'T'HH:mm:ss.SSS'Z'}{UTC} [%t] %-5level %logger{36} %X - %msg%n"/>
```

davidcurrie (dismissed):
LGTM

mcgovc (dismissed):
LGTM

ronanbrowne (approved):
LGMT

PR #5415, opened 2024-01-11
CORE-18925 - Apply code formatting on membership modules

nikinagy:
Format membership code owned by the network team. Specifically the following packages:
/components/membership/
/libs/membership/
/libs/layered-property-map/


.git-blame-ignore-revs
```diff
@@ -15,3 +15,5 @@ a05941cd4bc4a92ace31b32d7f1450d8fc6043b2
 477af2cf66a4d1208d3b11664fd3e2c80a6131fa
 # Network processors code
 b983ec7c351ef99e518bf16ef88f9489ed7d818a
+# Network Membership Code formatting
+08373a250cdd70a07b55f05c3e1111346b8b5ac6
```
dimosr, line 19: Are you going to squash-merge this, as usual? If so, should we have a follow-up PR to ignore the single commit that will be the result of it?

dimosr (commented):
Formatting diff looks good - just a comment on the blame ignore.

.git-blame-ignore-revs
nikinagy: According to the page: https://refactored-adventure-z4oe4g4.pages.github.io/ways-we-work/kotlin/kotlin-auto-code-formatting.html
**We should not do a squash merge**. However, I am not too sure what about the second commit as it's not saying anything about that. As far as I've seen everyone added only one commit to the ignore file which was the formatting itself.

dimosr: ah ok, I hadn't noticed that. That makes sense too, I guess we only want to want to ignore the formatting diff, so not ignoring the second commit is not a big deal.

dimosr (approved):

PR #5416, opened 2024-01-11
CORE-13499: clarify state manager update behaviour

dickon:


libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
```diff
@@ -55,7 +55,14 @@ interface StateManager : Lifecycle {
      * is only returned to the caller once all updatable [states] have been updated and replicas of the underlying
      * persistent storage, if any, are synced.
      *
-     * @param states Collection of states to be updated.
+     * Typical usage is to get some states, e.g. using `findByMetadataMatchingAll`, then make changes to the
+     * state content but leave the version alone, then try calling `update`. If the result is non empty,
+     * sleep for a random time, then query and update again.
+     *
```
jujoramos: I would get rid of this part, or rephrase it to remove the `sleep` section. Aside from that, the `State` instances returned by this method are up to date already, caller shouldn't need to retrieve the states again.

jujoramos (commented):
Aside from the minor re-phrasing, can you also update `KDocs` for the `delete` function as well?.

jujoramos (approved):
LGTM

libs/state-manager/state-manager-api/src/main/kotlin/net/corda/libs/statemanager/api/StateManager.kt
dickon: done

PR #5418, opened 2024-01-11
CORE-19282 wrap the mediator subscriptions client state in a MediatorState wrapper object

LWogan:
API: https://github.com/corda/corda-api/pull/1442

New MediatorState wrapper allows us to store a list of bus bound output events per consumer input event.

A hash of ConsumerRecords (e.g Key + FlowEvent) will be used as the key to store bus bound output records of the message processor. e.g SessionEvents/FlowStatus. These will be stored in a follow-up PR. 

This PRs scope is to introduce the wrapper MediaterState


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/StateManagerHelper.kt
```diff
@@ -2,36 +2,44 @@ package net.corda.messaging.mediator
...
      */
     fun createOrUpdateState(
         key: String,
         persistedState: State?,
+        mediatorState: MediatorState,
         newState: StateAndEventProcessor.State<S>?,
     ) = serialize(newState?.value)?.let { serializedValue ->
+        mediatorState.state = ByteBuffer.wrap(serializedValue)
+        val mediatorStateBytes = serializer.serialize(mediatorState)
+            ?: throw IllegalStateException("Serialized mediator state was null. This should not be impossible!")
```
ben-millar: I appreciate the motivational sentiment behind "This should not be impossible!", but I also think it might be a typo. (Can change this whenever you update the API revision)

ben-millar (dismissed):
LGTM

ben-millar (approved):
Looks gooder to me

PR #5422, opened 2024-01-12
CORE-19195: proof merging and test cleanup

dickon:
A naive implementation of Merkle proof mering. We can work on a more efficient version later. 

Add test cases for proofs. This requires random test ordering and care not to use too much memory for the test lists, so we use the Kotlin iterator()/yield() library function. Use this same approach for simple proofs and subsets.

Convert various special cases inside inner function into their won test cases.

Rework test suites to unroll loops into JUnit, and allow larger tree sizes to be tested. Tree size is currently limited by the size of 64 bit unsigned longs.

Fix memory use in tree rendering.

Ensure that `MerkleProofImpl.calculateRootInstrumented` in all cases returns nodes in the expected order. This matters for use when merging.

### DIFF
--- a/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
+++ b/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
@@ -305,6 +305,11 @@ class MerkleProofImpl(
      */
     @Suppress("UnusedParameters")
     fun merge(other: MerkleProofImpl, digest: MerkleTreeHashDigestProvider): MerkleProofImpl {
+        require(this.treeSize == other.treeSize) {
+            "underlying tree sizes must match; left hand side has underlying tree "+
+                "size ${this.treeSize} and right hand side has underlying tree size ${other.treeSize}"
+        }
+
         // First, work out the leaves for the output proof.
         val indexMapThis = leaves.associateBy { it.index }
         val indexMapOther = other.leaves.associateBy { it.index }
--- a/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
+++ b/libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
@@ -105,6 +105,7 @@ class MerkleProofImpl(
                     onNewHash(MerkleNodeInfo(it, treeDepth, null))
                 }
             }
+        val pendingNodes = mutableListOf<MerkleNodeInfo>()
         // need for the number of elements
         var currentSize = treeSize                                 // outer loop variable; the number of
         // leaves left as we roll up the tree
@@ -135,6 +136,9 @@ class MerkleProofImpl(
             // Now walk over the hashes at this tree level, striding over 1 or 2 at a time
             var index = 0
             while (index < nodeHashes.size) {
+                while ((pendingNodes.firstOrNull()?.node?.indexWithinLevel ?: Int.MIN_VALUE) >= index) {
+                    onNewHash(pendingNodes.removeFirst())
+                }
                 val item = nodeHashes[index]
                 // We are at level $treeDepth from the top of the tree (where 1 is the root of the tree),
                 //     and at $index nodes from the left (counting from 0)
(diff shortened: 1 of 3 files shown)


libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -273,6 +294,66 @@ class MerkleProofImpl(
         return MerkleProofImpl(proofType, treeSize, outLeaves, outHashes)
     }
 
+    @Suppress("UnusedParameters")
+    fun merge(other: MerkleProofImpl, digest: MerkleTreeHashDigestProvider): MerkleProofImpl {
+        // First, work out the leaves for the output proof.
+        val indexMapMe = leaves.associateBy { it.index }
+        val indexMapOther = other.leaves.associateBy { it.index }
```
mnesbit, line 315: Needs some more checks that the other proof is compatible. i.e. any overlapped leaf must have exact same IndexedMerkleLeaf, the tree sizes must be equal. Ultimately the root hash needs to be checked for equality, though likely that should be done during processing

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -101,16 +101,30 @@ class MerkleProofImpl(
...
         // work out nodeHashes, which is a list of node information for the current level we operate at
         var nodeHashes: List<MerkleNode> = leaves.sortedBy { it.index }.map {
-                MerkleNode(it.index, digest.leafHash(it.index, it.nonce, it.leafData)).also {
-                    onNewHash(MerkleNodeInfo(it, treeDepth, null))
-                }
+            MerkleNode(it.index, digest.leafHash(it.index, it.nonce, it.leafData))
+        }
+
+        // We will discover some hashes at a given level and some at the level above.
+        // But we want to guarnatee to our uses that the get out hashes on the `onNewHash` callback
```
```suggestion
        // But we want to guarantee to our uses that they get out hashes on the `onNewHash` callback
```
?

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
```diff
@@ -101,16 +101,30 @@ class MerkleProofImpl(
...
+        // in a left to right then bototm to top order, rather than mix up leaves from different levels.
```
```suggestion
        // in a left to right then bottom to top order, rather than mix up leaves from different levels.
```

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
vlajos: Hash digest providers need to be the same as well.

dickon: I don't know how we can implement equality on hash digest providers, though? There may be different instances of the same code, or different implementaitons with the same behaviour, or the same implementation configured differently. Consider a hypothetical digest which hased a configurable number of times, the equality would need to include the configuration of how many repeated hashes to do.

vlajos: I think we can rely on a meaningful `equal()` which gives true for the same classes with the same configuration. 
(NonceHashDigestProvider looks like a good example for me)
This will treat different instances with the same code/config equal, and the same implementations with different configurations different. Which looks correct.
I do not think we can easily detect different implementations with the same behaviour, also, probably, it is the safest to treat them as different.

BTW, the transaction's metadata specifies explicitly the hash digest provider details. So there should not exist filtered transactions with Merkle Proofs with different hash digest providers for the same transaction.

dickon: Even if we did, we only accept one digest provider as an argument, so we trust callers to supply a compatible digest provider. If we had decided on a service reference in each proof instance we could do a check.

nkovacsx (dismissed):
Approving after our pair-programming session with @dickon. It makes sense to me.

libs/crypto/merkle-impl/src/main/kotlin/net/corda/crypto/merkle/impl/MerkleProofImpl.kt
dickon: > Needs some more checks that the other proof is compatible. i.e. any overlapped leaf must have exact same IndexedMerkleLeaf, the tree sizes must be equal. Ultimately the root hash needs to be checked for equality, though likely that should be done during processing

I've implemented all of those.

vlajos (dismissed):
The algorithm looks good to me.
I did not check the tests thoroughly.

anien (approved):
Approving on the terms it's been already approved before.

PR #5424, opened 2024-01-12
CORE-18631, CORE-18630: Add state manager for inbound and outbound sessions

yift-r3:
Add state manager support for the link manager stateful sessions manager for both inbound and outbound sessions.


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -53,5 +102,224 @@ internal class StatefulSessionManagerImpl(coordinatorFactory: LifecycleCoordinat
...
+    }
+
+    private data class TraceableResult<T>(
+        val traceable: T,
+        val message: LinkOutMessage,
+        val stateUpdate: State,
+        val sessionToCache: Session?
+    )
+
+    private val cachedInboundSessions = ConcurrentHashMap<String, SessionManager.SessionDirection>()
```
dimosr: I think we should be using a Guava cache here, so that we can configure max size, eviction policies etc.

That said, I think it could potentially be done as part of the follow-up epic to expedite getting this in.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -30,6 +77,8 @@ internal class StatefulSessionManagerImpl(coordinatorFactory: LifecycleCoordinat
         /*To be implemented CORE-18631 (process InitiatorHelloMessage, InitiatorHandshakeMessage)
          * and CORE-18630 (process ResponderHello + process ResponderHandshake)
          * */
+        val messages = wrappedMessages.map { it to getMessage(it)}
+        processInboundSessionMessages(messages)
         return emptyList()
```
dimosr: needs replacement with actual result?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -53,5 +102,224 @@ internal class StatefulSessionManagerImpl(coordinatorFactory: LifecycleCoordinat
...
+            }
+        }.toMutableList()
+
+        stateManager.update(result.map { it.stateUpdate }).map { (failedUpdateSessionId, _) ->
+            val toRemove = result.find { it.stateUpdate.key == failedUpdateSessionId }
+            result.remove(toRemove)
+        }
+        return result.map{
+            it.sessionToCache?.let { sessionToCache ->
+                cachedInboundSessions.put(
```
dimosr: nit: should we move that outside the `.map` (e.g. in a separate `forEach`) to make it a bit clearer, since that's a separate side-effect?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -53,5 +102,224 @@ internal class StatefulSessionManagerImpl(coordinatorFactory: LifecycleCoordinat
...
+            null -> {
+                sessionManagerImpl.processInitiatorHello(message.initiatorHelloMessage)?.let {
+                    (responseMessage, authenticationProtocol) ->
+                    val timestamp = Instant.now()
+                    val newMetadata = InboundSessionMetadata(
+                        source = responseMessage.header.destinationIdentity.toCorda(),
+                        destination = responseMessage.header.sourceIdentity.toCorda(),
+                        lastSendTimestamp = timestamp,
+                        encryptionKeyId = "",
+                        encryptionKeyTenant = "",
```
dimosr: Can we leave a comment on what needs to be done here (and as part of which ticket it will be done)?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -53,5 +102,224 @@ internal class StatefulSessionManagerImpl(coordinatorFactory: LifecycleCoordinat
...
+                        source = responseMessage.header.sourceIdentity.toCorda(),
+                        destination = responseMessage.header.destinationIdentity.toCorda(),
+                        lastSendTimestamp = timestamp,
+                        encryptionKeyId = "",
+                        encryptionKeyTenant = "",
+                        status = InboundSessionStatus.SentResponderHandshake,
+                        expiry = Instant.now() + Duration.ofDays(7)
+                    )
+                    val session = sessionData.getSession()
+                    val newState = State(
```
dimosr: Do we need to specify `version` here, given there will be an existing state that needs update?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/InboundSessionMetadata.kt
```diff
@@ -0,0 +1,66 @@
...
+}
+
+data class InboundSessionMetadata(
+    val source: HoldingIdentity,
+    val destination: HoldingIdentity,
+    val lastSendTimestamp: Instant,
+    val encryptionKeyId: String,
+    val encryptionKeyTenant: String,
+    val status: InboundSessionStatus,
+    val expiry: Instant
```
dimosr: worth adding some kdocs on the fields here whose semantics is not completely clear. The ones that stand out to me are:
* `status`
* `lastSendTimestamp`
* `expiry`
* `encryptionKeyId` and `encryptionKeyTenant`

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/InboundSessionMetadata.kt
```diff
@@ -0,0 +1,66 @@
...
+        Instant.ofEpochMilli(metadata[LAST_SEND_TIMESTAMP] as Long),
+        metadata[ENCRYPTION_KEY_ID].toString(),
+        metadata[ENCRYPTION_TENANT].toString(),
+        metadata[STATUS].toString().statusFromString(),
+        Instant.ofEpochMilli(metadata[EXPIRY] as Long)
+    )
+
+    fun lastSendExpired(): Boolean {
+        return Instant.now() > lastSendTimestamp + SESSION_EXPIRY_PERIOD
+    }
```
dimosr: What does that represent? 2 things come to mind:
* session that needs rotation / has been rotated. But in this case, we should probably not compare with last send, but the time the session got established.
* whether a session needs refresh (e.g. see CORE-18730). In this case, I think we'd need a much tighter time window (e.g. a few seconds), than 7 days which is the rotation one.

The implementation makes me think it's 1, but looking at how it's used from https://github.com/corda/corda-runtime-os/pull/5438, makes me think that part of the code expects it to be 2.

-----

I think we need 2 functions for each one. The first one for expiry is probably only going to be used when we get to https://r3-cev.atlassian.net/browse/CORE-18632. The second one would be used for replays as part of the changes of the outbound path, not sure we need to use it as part of the inbound path (we can potentially always respond and let the receiving naturally filter out naturally based on the session state if they've already processed one)?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
dimosr: Looking at https://github.com/corda/corda-runtime-os/pull/5438, I see that's being done there.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/config/MultiClusterTestConfigManager.kt
```diff
@@ -10,6 +10,7 @@ class MultiClusterTestConfigManager(
     }
 
     override fun load(section: String, props: Map<String, Any?>): TestConfigManager {
+        configManagers.iterator().apply {  }
```
dimosr: is this needed? Do I understand correctly, it has no effect given block is an empty lambda?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
yift-r3: Done in https://github.com/corda/corda-runtime-os/pull/5441

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
yift-r3: Removed in https://github.com/corda/corda-runtime-os/pull/5441

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/config/MultiClusterTestConfigManager.kt
yift-r3: Removed in https://github.com/corda/corda-runtime-os/pull/5441

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: This has been added in #5441

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/InboundSessionMetadata.kt
williamvigorr3: I've added some Kdoc to everything (except `encryptionKeyId` and `encryptionKeyTenant` which will be deleted when #5441 gets merged).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: Yes. I am now incrementing the version in all places even if we only change the metadata.

seanbrereton (approved):
LGTM (BLT)

williamvigorr3 (approved):
LGTM

YashNabar (approved):

PR #5425, opened 2024-01-12
CORE-18879: Ensure verifying notary implementation

jennyang-r3:
In this PR, I added more unit test cases for verifying notary, and checking any missing checks, and refactored the test.

**added tests:**
- Contract verifying notary should successfully notarise if a signed tx verifies signatory signatures
- Contract verifying notary should respond with error if contract verification failed
- Contract verifying notary should respond with error if a signed transaction failed to signatory signature verification
- Contract verifying notary should respond with error if either current or filtered tx is invalid 
- Contract verifying notary should respond with error if dependency is invalid
- Contract verifying notary should respond with error if dependency reference is missing
- Contract verifying notary should respond with error if dependency input is missing

### DIFF
--- a/notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
+++ b/notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
@@ -194,12 +194,7 @@ class ContractVerifyingNotaryServerFlowImpl() : ResponderFlow {
         filteredTransactionsAndSignatures: List<FilteredTransactionAndSignatures>
     ) {
         try {
-
             val dependentStateAndRefs = filteredTransactionsAndSignatures.flatMap { (filteredTransaction, _) ->
-                require(filteredTransaction.id != initialTransaction.id) {
-                    "Either filtered transaction \"${filteredTransaction.id}\" or " +
-                            "current transaction \"${initialTransaction.id}\" is invalid."
-                }
                 (filteredTransaction.outputStateAndRefs as UtxoFilteredData.Audit<StateAndRef<*>>).values.values
             }.associateBy { it.ref }
 
--- a/notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
+++ b/notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
@@ -196,6 +196,10 @@ class ContractVerifyingNotaryServerFlowImpl() : ResponderFlow {
         try {
 
             val dependentStateAndRefs = filteredTransactionsAndSignatures.flatMap { (filteredTransaction, _) ->
+                require(filteredTransaction.id != initialTransaction.id) {
+                    "Either filtered transaction \"${filteredTransaction.id}\" or " +
+                            "current transaction \"${initialTransaction.id}\" is invalid."
+                }
                 (filteredTransaction.outputStateAndRefs as UtxoFilteredData.Audit<StateAndRef<*>>).values.values
             }.associateBy { it.ref }
 
--- a/notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
+++ b/notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
@@ -16,7 +16,6 @@ import net.corda.uniqueness.datamodel.impl.UniquenessCheckErrorUnhandledExceptio
 import net.corda.uniqueness.datamodel.impl.UniquenessCheckResultFailureImpl
 import net.corda.uniqueness.datamodel.impl.UniquenessCheckResultSuccessImpl
 import net.corda.uniqueness.datamodel.impl.UniquenessCheckStateRefImpl
-import net.corda.v5.application.crypto.DigitalSignatureAndMetadata
 import net.corda.v5.application.membership.MemberLookup
 import net.corda.v5.application.messaging.FlowSession
 import net.corda.v5.application.uniqueness.model.UniquenessCheckError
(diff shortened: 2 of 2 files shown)


notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/main/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImpl.kt
```diff
@@ -196,6 +196,10 @@ class ContractVerifyingNotaryServerFlowImpl() : ResponderFlow {
         try {
 
             val dependentStateAndRefs = filteredTransactionsAndSignatures.flatMap { (filteredTransaction, _) ->
+                require(filteredTransaction.id != initialTransaction.id) {
+                    "Either filtered transaction \"${filteredTransaction.id}\" or " +
+                            "current transaction \"${initialTransaction.id}\" is invalid."
+                }
```
blsemo: Why do we need this check? And if we need this check, shouldn't the error message be more specific to what the problem is, i.e. a transaction depending on itself?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
```diff
@@ -299,8 +401,13 @@ class ContractVerifyingNotaryServerFlowImplTest {
     }
 
     @Test
+    @Order(10)
     fun `Contract verifying notary plugin server should respond with error if transaction verification fails`() {
```
blsemo: That title is wrong - it suggests that the transaction verification (i.e. of the  initital transaction) fails - but it really is the Merkle proof verification of the dependencies that fails here.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
```diff
@@ -351,99 +463,278 @@ class ContractVerifyingNotaryServerFlowImplTest {
...
+        val responseError = responseFromServer.first().error
+        assertThat(responseError).isNotNull
+        assertThat(responseError).isInstanceOf(NotaryExceptionTransactionVerificationFailure::class.java)
+        assertThat((responseError as NotaryExceptionTransactionVerificationFailure).errorText)
+            .contains("Missing reference state and ref from the filtered transaction")
+    }
+
+    @Test
+    @Order(16)
+    fun `Contract verifying notary should respond with error if either current or filtered tx is invalid`() {
```
blsemo: I don't understand the test case - in the real world, the initial transaction cannot be a dependency of itself, as it's not a valid, notarized transaction at this point yet.
If anything, this should complain that there is a circular dependency.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
```diff
@@ -351,99 +463,278 @@ class ContractVerifyingNotaryServerFlowImplTest {
...
+        val responseError = responseFromServer.first().error
+        assertThat(responseError).isNotNull
+        assertThat(responseError).isInstanceOf(NotaryExceptionTransactionVerificationFailure::class.java)
+        assertThat((responseError as NotaryExceptionTransactionVerificationFailure).errorText)
+            .contains("Either filtered transaction \"$SIGNED_TX_ID\" or current transaction \"$SIGNED_TX_ID\" is invalid")
+    }
+
+    @Test
+    @Order(17)
+    fun `Contract verifying notary should respond with error if a signed transaction failed to signatory signature verification`() {
```
blsemo:
```suggestion
    fun `Contract verifying notary should respond with error if a signed transaction failed to verify signatory signature verification`() {
```

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
```diff
@@ -351,99 +463,278 @@ class ContractVerifyingNotaryServerFlowImplTest {
...
+        )
+        val responseError = responseFromServer.first().error
+        assertThat(responseError).isNotNull
+        assertThat(responseError).isInstanceOf(NotaryExceptionGeneral::class.java)
+        assertThat((responseError as NotaryExceptionGeneral).message)
+            .contains("Signature failed error")
+    }
+
+    @Test
+    @Order(Integer.MAX_VALUE)
```
blsemo: Why does this test need to be run last?

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
```diff
@@ -351,99 +463,278 @@ class ContractVerifyingNotaryServerFlowImplTest {
...
+    @Order(19)
+    fun `Contract verifying notary should successfully notarise if a signed tx verifies signatory signatures`() {
+        whenever(mockTransactionSignatureService.signBatch(any(), any())).thenReturn(
+            listOf(listOf(signatorySignatureCharlie))
+        )
+
+        val signedTransaction = mockSignedTransaction()
+        whenever(signedTransaction.verifySignatorySignatures()).thenAnswer {  }
+
+        signedTransaction.verifySignatorySignatures()
```
blsemo: What is the purpose of this call?

blsemo (commented):
I have few questions/remarks about the tests in this PR, see the comments inline, also:
- Why is the ordering of the tests required? Does it make a difference in which order the tests are run? If yes, that is a bad code smell, if no, can we get rid of it?
- A lot of these tests just check that exceptions thrown by other services are converted correctly - they do not check that the real thing would behave as expected on wrong or malicious inputs, so I think we still need to test most of these things as e2e tests. This is especially true for any test where you set a mock API to throw.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
jennyang-r3: I was trying to call the actual function, but it didn't work so can be ignored.

notary-plugins/notary-plugin-contract-verifying/notary-plugin-contract-verifying-server/src/test/kotlin/com/r3/corda/notary/plugin/contractverifying/server/ContractVerifyingNotaryServerFlowImplTest.kt
jennyang-r3: Ordered test sorted out

blsemo (approved):
LGTM

PR #5426, opened 2024-01-12
CORE-19016 store async outputs

LWogan:
Mediator wrapper state pr: https://github.com/corda/corda-runtime-os/pull/5418

The list of asynchronous output records associated with each consumer record polled and processed needs to be saved to the state manager.

The mediator already groups asynchronous outputs together into a collection. This collection is altered to store them keyed by input event.
A new service is provided to handle mediator replay logic

Output events are stored in MediatorState. Events should be keyed by consumer record input hash to allow for replays to be detected.


libs/state-manager/state-manager-db-impl/src/main/kotlin/net/corda/libs/statemanager/impl/repository/StateRepository.kt
```diff
@@ -26,7 +26,7 @@ interface StateRepository {
      * Transaction should be controlled by the caller.
      *
      * @param connection The JDBC connection used to interact with the database.
-     * @param state State entity to persist.
+     * @param states State entity to persist.
```
ben-millar, line 29: Not all heroes wear capes

ben-millar (dismissed):
LGTM

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessor.kt
```diff
@@ -45,20 +46,26 @@ class EventProcessor<K : Any, S : Any, E : Any>(
...
+            val asyncOutputs = mutableMapOf<Record<K, E>, MutableList<MediatorMessage<Any>>>()
+            val allConsumerInputs = groupEntry.value
             val processed = try {
-                while (queue.isNotEmpty()) {
-                    val event = queue.removeFirst()
-                    val response = config.messageProcessor.onNext(processorState, event)
-                    processorState = response.updatedState
-                    val (syncEvents, asyncEvents) = response.responseEvents.map { convertToMessage(it) }.partition {
-                        messageRouter.getDestination(it).type == RoutingDestination.Type.SYNCHRONOUS
+                allConsumerInputs.onEach { consumerInputEvent ->
```
mbrkic-r3:
```suggestion
                allConsumerInputs.forEach { consumerInputEvent ->
```

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessor.kt
```diff
@@ -45,20 +46,26 @@ class EventProcessor<K : Any, S : Any, E : Any>(
...
+                    while (queue.isNotEmpty()) {
+                        val event = queue.removeFirst()
+                        val response = config.messageProcessor.onNext(processorState, event)
+                        processorState = response.updatedState
+                        val (syncEvents, asyncEvents) = response.responseEvents.map { convertToMessage(it) }.partition {
+                            messageRouter.getDestination(it).type == RoutingDestination.Type.SYNCHRONOUS
+                        }
+                        asyncOutputs.compute(consumerInputEvent) { _, oldValue ->
+                            (oldValue?.plus(asyncEvents) ?: asyncEvents).toMutableList()
+                        }
```
mbrkic-r3:
```suggestion
                        asyncOutputs.computeIfAbsent(consumerInputEvent) { mutableListOf() }.addAll(asyncEvents)
```

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessor.kt
mbrkic-r3: Do we need this loop, can we use `event` (below) instead of `consumerInputEvent`?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/MediatorReplayService.kt
```diff
@@ -0,0 +1,71 @@
...
+     * @param newOutputs The new outputs to add to the existingOutputs to be stored in the mediator.
+     * @return Return a new [MediatorReplayOutputEvents] object containing all the existing outputs with the new outputs added to it.
+     */
+    fun <K : Any, E : Any> getOutputEvents(
+        existingOutputs: List<MediatorReplayOutputEvents>,
+        newOutputs: Map<Record<K, E>, MutableList<MediatorMessage<Any>>>
+    ): List<MediatorReplayOutputEvents> {
+        val mediatorOutputs = existingOutputs.toMutableList()
+
+        newOutputs.onEach { entry ->
```
mbrkic-r3:
```suggestion
        newOutputs.forEach { entry ->
```

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/MediatorReplayService.kt
```diff
@@ -0,0 +1,71 @@
...
+     * Generate a unique hash for the consumer input record to be used as a key when storing outputs.
+     * This allows for replay detection of consumer inputs by looking up the hash in the [MediatorReplayOutputEvents]
+     * @param inputEvent The consumer input event polled from the bus
+     * @return A hash of the input event as bytes
+     */
+    private fun <K : Any, E : Any> getInputHash(inputEvent: Record<K, E>): ByteBuffer {
+        val recordKeyBytes = serialize(inputEvent.key)
+        val recordValueBytes = serialize(inputEvent.value)
+        if (recordKeyBytes == null || recordValueBytes == null)
+            throw IllegalStateException("Input record key and value bytes should not be null")
```
mbrkic-r3:
```suggestion
        check (recordKeyBytes != null && recordValueBytes != null) {
            "Input record key and value bytes should not be null"
        }
```

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/MediatorReplayService.kt
```diff
@@ -0,0 +1,71 @@
...
+    ): List<MediatorReplayOutputEvents> {
+        val mediatorOutputs = existingOutputs.toMutableList()
+
+        newOutputs.onEach { entry ->
+            val hash = getInputHash(entry.key)
+            val mediatorOutputList = entry.value.map {
+                val topic = it.properties.getProperty(MSG_PROP_TOPIC)
+                val key = ByteBuffer.wrap(serializer.serialize(it.properties.getProperty(MSG_PROP_KEY)))
+                val payload = ByteBuffer.wrap(serialize(it.payload))
+                MediatorReplayOutputEvent(topic, key, payload)
```
mbrkic-r3, line 44: Should we also preserve all message properties?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/MediatorReplayService.kt
```diff
@@ -0,0 +1,71 @@
...
+            }
+            mediatorOutputs.add(MediatorReplayOutputEvents(hash, mediatorOutputList))
```
mbrkic-r3: Could we have multiple records with the same input (hash) here, should they be merged?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/MediatorReplayService.kt
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Generate a unique hash for the consumer input record to be used as a key when storing outputs.
+     * This allows for replay detection of consumer inputs by looking up the hash in the [MediatorReplayOutputEvents]
+     * @param inputEvent The consumer input event polled from the bus
+     * @return A hash of the input event as bytes
+     */
+    private fun <K : Any, E : Any> getInputHash(inputEvent: Record<K, E>): ByteBuffer {
+        val recordKeyBytes = serialize(inputEvent.key)
```
mbrkic-r3: Do we need to include the key considering that state is always related to a single key?

LWogan: Ah yes this is true. I had originally designed it in my head for storing separately in the state manager. good catch!

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/MediatorReplayService.kt
LWogan: not based on my usage, but your suggestion might be safer longer tern so ill apply it

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/MediatorReplayService.kt
LWogan: At the moment there are no other properties required right now. But a keyvaluepair list might be a better choice for the schema rather than just explicitly choosing the ones i need. Ill raise a followup ticket for this as its not critical. 
https://r3-cev.atlassian.net/browse/CORE-19327

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessor.kt
LWogan: i need to be able to store asyncOutputs per consumerInputEvent, not per event.  event could be a synchronous response. We only store outputs for each polled record. Resulting synchronous responses that have outputs are keyed by the original polled record.

mbrkic-r3: Ah, got it.
With this change sync events will be processed immediately, before other input events (previously all sync events would be processed in the "second round"). Not sure whether this would affect performance (depends on many other things I guess).

mbrkic-r3 (approved):
LGTM

PR #5430, opened 2024-01-12
CORE-18839: Enforce strict constraints for SLF4j

vkolomeyko:


gradle/libs.versions.toml
```diff
@@ -11,9 +11,11 @@ jettyVersion = "9.4.53.v20231009"
 
 log4jVersion = "2.20.0"
 # SLF4J cannot be ugraded to 2.x due to CorDapps requiring the 1.7 <= x < 2.0
-slf4jVersion = "1.7.36"
-# The CLI uses SLF4J version 2
-slf4jV2Version = "2.0.6"
```
simon-johnson-r3, line 16: This stops us getting a later version as a transitive dependency, is that correct? Whilst that seems reasonable, I have a couple of questions.
- When was that happening?
- Is there anything else we need to apply this to?

vkolomeyko: There is an ongoing issue with SLF4j upgrade to v2.0.
This was raised a while back on Principal Engineers' forum: https://r3-cev.atlassian.net/wiki/spaces/CB/pages/4631003156/2023-09-28+-+Principal+Engineers+Weekly+Meeting

Since our sandbox requires SLF4J v1, this will prevent us upgrading to 3rd party libraries which are dependent on v2.x. There is also a Jira to that effect: https://r3-cev.atlassian.net/browse/CORE-15310

Work on this PR is trying to workaround the issue for CLI runtime execution environment.

PR #5434, opened 2024-01-15
CORE-19017 detect when an input record is a replayed event based on data stored in the mediator state

LWogan:
To be used in a followup pr to detect replayed events in the mediator messaging pattern.

Compare a hash of the input record to previously saved hashes of input records whose outputs have been saved to the MediatorState wrapper


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/MediatorReplayService.kt
```diff
@@ -49,6 +51,29 @@ class MediatorReplayService @Activate constructor(
...
+        val savedOutputs = mediatorState.outputEvents
+        val inputHash = getInputHash(inputRecord).array()
+
+        savedOutputs.forEach { mediatorReplayOutputEvents ->
+            if (inputHash.contentEquals(mediatorReplayOutputEvents.inputEventHash.array())) {
+                return mediatorReplayOutputEvents.outputEvents.map { outputEvent ->
+                    outputEvent.toMediatorMessage()
+                }
+            }
+        }
```
ben-millar: We should leverage our knowledge of the underlying data and search from the end of our list, given that our `savedOutputs` are in chronological order and replayed events are more likely to be at the end. This way the lookup time doesn't increase with the size of the list (provided there is a matching value there).
```suggestion
        return mediatorState.outputEvents
            .findLast { inputHash.contentEquals(it.inputEventHash.array()) }
            ?.outputEvents
            ?.map { it.toMediatorMessage() }
```

ben-millar (dismissed):
LGTM, left optional suggestion

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/MediatorReplayService.kt
LWogan: ah yes, i forgot you suggested this, thanks!

ben-millar (approved):

mbrkic-r3 (approved):
LGTM

PR #5436, opened 2024-01-15
CORE-19293: Increase Transient Error Timeout Retry

jujoramos:
Increase the growth factor used by the exponential backoff algorithm
while retrying transient errors at the flow engine. When using the
default configuration settings, this brings the maximum time spent
during retries from ~8 seconds to ~30 seconds total.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
@@ -119,8 +119,8 @@ class FlowEventProcessorImpl(
                 logger = log,
                 maxRetries = flowConfig.getLong(FlowConfig.PROCESSING_MAX_RETRY_ATTEMPTS),
                 maxTimeMillis = flowConfig.getLong(FlowConfig.PROCESSING_MAX_RETRY_WINDOW_DURATION),
-                // Exponential backoff -> 500ms, 1s, 2s, 4s, 8s, etc.
-                backoffStrategy = Exponential(base = 2.0, growthFactor = 250L),
+                // Exponential Backoff + Default maxRetryAttempts -> 1s, 2s, 4s, 8s, 16s = 31s max
+                backoffStrategy = Exponential(base = 2.0, growthFactor = 500L),
                 // Only FlowTransientException will be retried
                 shouldRetry = { _, _, t -> t is FlowTransientException },
                 onRetryAttempt = { n, d, t -> logRetryAndRollbackCheckpoint(pipeline.context.checkpoint, n, d, t) },


components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
```diff
@@ -119,8 +119,8 @@ class FlowEventProcessorImpl(
                 logger = log,
                 maxRetries = flowConfig.getLong(FlowConfig.PROCESSING_MAX_RETRY_ATTEMPTS),
                 maxTimeMillis = flowConfig.getLong(FlowConfig.PROCESSING_MAX_RETRY_WINDOW_DURATION),
-                // Exponential backoff -> 500ms, 1s, 2s, 4s, 8s, etc.
-                backoffStrategy = Exponential(base = 2.0, growthFactor = 250L),
+                // Exponential Backoff + Default maxRetryAttempts -> 1s, 2s, 4s, 8s, 16s = 31s max
+                backoffStrategy = Exponential(base = 2.0, growthFactor = 500L),
```
LWogan, line 123: Should we put this value into the flow corda-api config?

JamesHR3: I personally don't think adding these to the config really achieves very much. We already have fairly granular controls over retry number and window, all we should really do is make sure we get the count in the retry window.

We could try and do some calculation here to make sure we fit the retries in the window, but if that turns out to be complex I'm not sure it's worth it.

JamesHR3 (approved):

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
jujoramos: > Should we put this value into the flow corda-api config? 

Probably not as they are only internal, used by the exponential backoff algorithm only. The user can already configure the maximum amount of retries (`maxRetryAttempts`) and the maximum duration that can be used by retries (`maxRetryAttempts`), it should enough for them.

jujoramos: > I personally don't think adding these to the config really achieves very much. We already have fairly granular controls over retry number and window, all we should really do is make sure we get the count in the retry window.

Agreed.

> We could try and do some calculation here to make sure we fit the retries in the window, but if that turns out to be complex I'm not sure it's worth it.

It's already done within the `tryWithBackoff` implementation, the time spent in retries should never exceed the configured `maxRetryWindowDuration`.

LWogan (approved):

PR #5437, opened 2024-01-15
CORE-19262: Added support for unmananged key rotation status get api

akankshaSharmaR3:


testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
```diff
@@ -226,6 +232,15 @@ fun ClusterInfo.rotateCryptoUnmanagedWrappingKeys(
     }
 }
 
+fun ClusterInfo.cryptoUnmanagedWrappingKeysRotationStatus(
+    requestid: String
+) = cluster {
+    assertWithRetry {
+        command { getCryptoUnmanagedWrappingKeysRotationStatus(requestid) }
+        condition { it.code == ResponseCode.ACCEPTED.statusCode }
```
anien: This is not returning ACCEPTED code, but OK status code. Or some exception.

anien (changes requested):

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
akankshaSharmaR3: Changed it to 'OK'.

anien (dismissed):
LGTM

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
```diff
@@ -226,6 +232,15 @@ fun ClusterInfo.rotateCryptoUnmanagedWrappingKeys(
     }
 }
 
+fun ClusterInfo.cryptoUnmanagedWrappingKeysRotationStatus(
```
simon-johnson-r3: The name of this function is a bit confusing. As far as I can tell it waits for an OK status or fails after timeout?

In that case it would be better named something like `waitForCryptoUnmanagedWrappingKeysRotation`

simon-johnson-r3 (changes requested):

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
akankshaSharmaR3: Changed method name as suggested

anien: I don't think that's completely correct now. We don't wait for key rotation, but for the key rotation status. Other functions use the same `assertWithRetry` but I don't really think the retry will happen much in this case and any other function has the wait in the name. To keep consistency, I would be keen on keeping the original name of the function.

anien: @simon-johnson-r3  WDYT?

akankshaSharmaR3: Changed name to "getStatusForUnmanagedWrappingKeysRotation"

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterBuilder.kt
```diff
@@ -736,13 +740,25 @@ class ClusterBuilder {
...
         )
     }
 
+    fun getCryptoUnmanagedWrappingKeysRotationStatus(requestid: String): SimpleResponse {
+        return get(
+            "/api/$REST_API_VERSION_PATH/wrappingkey/unmanaged/rotation/${requestid}",
+        )
+    }
+
+    fun getWrappingKeysProtocolVersion(): SimpleResponse {
```
anien, line 757: Are we testing this? If yes, I am happy to leave it. Just not sure if we test it in other APIs.

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterBuilder.kt
```diff
@@ -736,13 +740,25 @@ class ClusterBuilder {
...
+        return get(
+            "/api/$REST_API_VERSION_PATH//wrappingkey/getprotocolversion",
```
anien:
```suggestion
            "/api/$REST_API_VERSION_PATH/wrappingkey/getprotocolversion",
```

akankshaSharmaR3: Updated

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
```diff
@@ -226,6 +232,23 @@ fun ClusterInfo.rotateCryptoUnmanagedWrappingKeys(
...
+fun ClusterInfo.getStatusForUnmanagedWrappingKeysRotation(
+    requestid: String
+) = cluster {
+    assertWithRetry {
+        command { getCryptoUnmanagedWrappingKeysRotationStatus(requestid) }
+        condition { it.code == ResponseCode.OK.statusCode }
+    }
+}
+
+fun ClusterInfo.getProtocolVersionForWrappingKeys(
```
anien: I would suggest the name to be something like `getProtocolVersionForUnmanagedKeyRotation`, because the whole section in swagger/API is called Key Rotation (Which might change once we add managed key rotation).

akankshaSharmaR3: Done

testing/e2e-test-utilities/src/main/kotlin/net/corda/e2etest/utilities/ClusterUtils.kt
simon-johnson-r3: I guess the confusing part here is the use of the word "status". We are waiting for the http _status_ to be OK. But then we return the _status_ of key rotation which is something different entirely, contained within the contents of the OK response.

But anyway, I think the name is clearer now.

simon-johnson-r3 (dismissed):

anien (dismissed):

anien (approved):

PR #5438, opened 2024-01-15
CORE-18630 Add state manager for outbound sessions

YashNabar:
Add state manager transitions to the stateful session manager for outbound sessions. Introduce caching for established sessions.
Note: This feature is disabled in this change.


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -38,36 +50,175 @@ internal class StatefulSessionManagerImpl(
...
+            messageAndKeyMap.mapNotNull {
+                OutboundMessageContext(it.value, state[it.key], getMessage(it.value))
+            }
+        }
+        val resultStates = sessionStates.map { traceAndState ->
+            val counterparties = sessionManagerImpl.getSessionCounterpartiesFromMessage(traceAndState.message.message)
+            if (counterparties == null) {
+                traceAndState to SessionManager.SessionState.CannotEstablishSession
+            }
+            val state = traceAndState.state ?: return@map sessionNeeded(
```
yift-r3: Shouldn't that be inside an else?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -38,36 +50,175 @@ internal class StatefulSessionManagerImpl(
...
+                        traceAndState to (state.retrieveEstablishedSession(counterparties!!)
+                            ?: SessionManager.SessionState.CannotEstablishSession)
+                    }
+
+                    else -> traceAndState to SessionManager.SessionState.CannotEstablishSession
+                }
+            }
+        }
+
+        val failures = stateManager.update(updates)
```
yift-r3: We probably need to log the errors

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -38,36 +50,175 @@ internal class StatefulSessionManagerImpl(
...
+    ): SessionManager.SessionState.SessionEstablished? {
+        val sessionData = AvroSessionState.fromByteBuffer(ByteBuffer.wrap(value))
+            .toCorda(
+                schemaRegistry,
+                sessionEncryptionOpsClient,
+                sessionManagerImpl.revocationCheckerClient::checkRevocation
+            ).sessionData
+        return when(sessionManagerImpl.getProtocolMode(counterParties.ourId)) {
+            GroupPolicyConstants.PolicyValues.P2PParameters.ProtocolMode.AUTH -> {
+                SessionManager.SessionState.SessionEstablished(sessionData as AuthenticatedSession, counterParties)
```
yift-r3: Can't we do the when on the sessionData to avoid the casting?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -38,36 +50,175 @@ internal class StatefulSessionManagerImpl(
...
 
     override fun <T> getSessionsById(
         uuids: Collection<T>,
         getSessionId: (T) -> String
     ): Collection<Pair<T, SessionManager.SessionDirection>> {
         val traceable = uuids.associateBy { getSessionId(it) }
-        val sessionFromCache = uuids.map { it to cachedInboundSessions[getSessionId(it)] }
-        val sessionsIdsNotInCache = sessionFromCache.filter { it.second == null }.map { getSessionId(it.first) }
+        val sessionsFromCache = uuids.map {
+            it to (cachedInboundSessions[getSessionId(it)] ?: cachedOutboundSessions[getSessionId(it)])
```
yift-r3: Do we need to keep two caches?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -38,36 +50,175 @@ internal class StatefulSessionManagerImpl(
...
+        }
+        val sessionsNotInCache = sessionsFromCache.filter { it.second == null }.map { it.first to getSessionId(it.first) }
```
yift-r3: instead of map we can do associateWith, so we will have a map (and then we can use it's values and keys instead of maps)

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -38,36 +50,175 @@ internal class StatefulSessionManagerImpl(
...
         }
-        // In CORE-18630 check cache/state manager for outbound sessions if not found for inbound sessions.
-        // We should avoid as unnecessary reads of the state manager, so first check the inbound and outbound session cache,
-        // then the state manager for an inbound session (query with session ID), then the state manager for an outbound session
-        // (query for sessionId in metadata).
+        val sessionsNotInInboundStateManager =
+            (sessionsNotInCache.map { it.first } - inboundSessionsFromStateManager.map { it.first }.toSet()).map {
+                getSessionIdFilter(getSessionId(it))
+            }
+        val outboundSessionsFromStateManager: List<Pair<T, SessionManager.SessionDirection>> =
```
yift-r3: I'm confused; where in the state API we asked for only inbound or outbound?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -78,8 +229,40 @@ internal class StatefulSessionManagerImpl(
          * and CORE-18630 (process ResponderHello + process ResponderHandshake)
          * */
         val messages = wrappedMessages.map { it to getMessage(it)}
-        processInboundSessionMessages(messages)
-        return emptyList()
+        val result = (processInboundSessionMessages(messages) + processOutboundSessionMessages(messages)).toMutableList()
```
yift-r3: We don't need a mutable list; we can keep it as a collection, and filter later one (we do a map anyway in line 238).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
YashNabar: The inbound and outbound sessions are keyed differently, so to retrieve inbound sessions (keyed by session ID) we simply do a `stateManager.get()`, but for outbound sessions we query based on metadata with `stateManager.findByMetadataMatchingAny()`.

That said, if we're happy to have a common metadata type for inbound and outbound (as I have here), we could query for both types of sessions with `stateManager.findByMetadataMatchingAny()`

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -38,36 +50,175 @@ internal class StatefulSessionManagerImpl(
...
+                traceAndState.message.message.header.statusFilter,
+            )?.let {
+                it.second?.let { newState -> updates.add(newState) }
+                traceAndState to it.first
+            } ?: (traceAndState to SessionManager.SessionState.CannotEstablishSession)
+            val metadata = SessionMetadata(state.metadata)
+            if (metadata.lastSendExpired()) {
+                when (metadata.status) {
+                    SessionStatus.SentInitiatorHello, SessionStatus.SentInitiatorHandshake -> {
+                        traceAndState to SessionManager.SessionState.SessionAlreadyPending(counterparties!!)
```
dimosr: Don't we need to force a replay of the session message here?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -38,36 +50,175 @@ internal class StatefulSessionManagerImpl(
...
+                    }
+                    SessionStatus.SessionReady -> {
+                        traceAndState to (state.retrieveEstablishedSession(counterparties!!)
+                            ?: SessionManager.SessionState.CannotEstablishSession)
+                    }
+                    else -> traceAndState to SessionManager.SessionState.CannotEstablishSession
```
dimosr: isn't this an invalid state? Should we log this accordingly?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -38,36 +50,175 @@ internal class StatefulSessionManagerImpl(
...
+                }
+            } else {
```
dimosr: isn't this branch capturing the case where we've recently sent the session message, so we should skip without sending any session message for a while (at which point we'll retry, i.e. the branch above)?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -38,36 +50,175 @@ internal class StatefulSessionManagerImpl(
...
+
+                    else -> traceAndState to SessionManager.SessionState.CannotEstablishSession
+                }
+            }
+        }
+
+        val failures = stateManager.update(updates)
+
+        return resultStates
+            .filterNot { failures.containsKey(it.first.message.calculateHash()) }
```
dimosr: shouldn't we return `CannotEstablishSession` for the ones that failed when trying to update via state manager, instead of filtering them out?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -38,36 +50,175 @@ internal class StatefulSessionManagerImpl(
...
+                SessionManager.SessionState.SessionEstablished(sessionData as AuthenticatedSession, counterParties)
+            }
+            GroupPolicyConstants.PolicyValues.P2PParameters.ProtocolMode.AUTH_ENCRYPT -> {
+                SessionManager.SessionState.SessionEstablished(sessionData as AuthenticatedEncryptionSession, counterParties)
+            }
+            else -> null
+        }
+    }
+
+    private fun sessionNeeded(
```
dimosr: nit: I think`newSessionNeeded` would be a better name for this?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -38,36 +50,175 @@ internal class StatefulSessionManagerImpl(
...
+            filter,
+        )?.firstOrNull() ?: return null
+
+        val newMetadata = SessionMetadata(
+            sessionId = initMessage.first.sessionId,
+            source = counterParties.ourId,
+            destination = counterParties.counterpartyId,
+            lastSendTimestamp = Instant.now(),
+            encryptionKeyId = "",
+            encryptionKeyTenant = "",
```
dimosr: Needs some docs here on what needs to be done and as part of which ticket.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -228,28 +460,83 @@ internal class StatefulSessionManagerImpl(
...
+                        ).message
+                    val newState = State(
+                        key = state.key,
+                        value = state.value,
+                        version = state.version,
+                        metadata = updatedMetadata.toMetadata()
+                    )
+                    initiatorHandshakeToResend to newState
+                } else {
+                    null
```
dimosr, line 852: similar question/comment on this being an invalid state, which requires ERROR/WARN logging.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -293,11 +581,41 @@ internal class StatefulSessionManagerImpl(
...
+                            message.responderHandshakeMessage.header.sessionId,
+                            SessionState(null, session).toAvro(schemaRegistry, sessionEncryptionOpsClient)
+                                .toByteBuffer().array(),
+                            metadata = updatedMetadata.toMetadata()
+                        )
+                        ProcessHandshakeResult(null, newState, session as Session)
+                    }
+                }
+            }
+            else -> null
```
dimosr: we could be a bit more granular here. e.g.:
* `SentInitiatorHello`, `SentInitiatorHandshake`, `SentResponderHello` are invalid states in which case we can log accordingly at WARN/ERROR level.
* `SessionReady` is the case where a responder handshake has already been processed and this is potentially a replay, in which case we could log at INFO since that shouldn't be a frequent scenario and it would be to know it's happening.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
YashNabar: According to this [section in the design](https://github.com/corda/platform-eng-design/blob/2994ee9b90734ba14c908dbfa8d48883aab6cc95/core/corda-5/corda-5.2/network/shared-session-storage/link-session-manager-changes.md#the-outbound-session-state), we want to discard any failed updates I thought?

dimosr: We want to ignore them in the sense that we don't need to retry these persistence requests, since the failure indicates another link manager has progressed this session to the next step already.

But at the same time, I believe the contract of the `processOutboundMessages` is to return an entry in the collection for every message that was passed in.

dimosr: Also, I think there's more to it - not just returning `CannotEstablishSession`, as I said above.

If the update fails, it means the session can be in many potential states which need to be handled separately. For instance, if the session is in the establishment phase (say 2 link managers trying to send initiator hello), then we need to queue the messages for when the session is established. I think filtering them out at this point risks the upstream components not processing the messages at all.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -38,48 +53,208 @@ internal class StatefulSessionManagerImpl(
...
+                    }
+                    SessionStatus.SessionReady -> {
+                        traceAndState to (state.retrieveEstablishedSession(counterparties!!)
+                            ?: SessionManager.SessionState.CannotEstablishSession)
+                    }
+                    else -> traceAndState to SessionManager.SessionState.CannotEstablishSession
+                }
+            } else {
+                when (metadata?.status) {
+                    null, SessionStatus.SentInitiatorHello, SessionStatus.SentInitiatorHandshake -> {
```
dimosr: is this correct?

I would have assumed that `status` would never be null.

Not having a session with a peer would correspond to a missing session in the state manager. Once we initialise a session, the session will always have a `status` value from that point on.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -228,28 +449,98 @@ internal class StatefulSessionManagerImpl(
...
+                        sessionId = message.responderHelloMessage.header.sessionId,
+                        source = metadata.source,
+                        destination = metadata.destination,
+                        lastSendTimestamp = timestamp,
+                        encryptionKeyId = "",
+                        encryptionKeyTenant = "",
+                        status = SessionStatus.SentInitiatorHandshake,
+                        expiry = Instant.now() + Duration.ofDays(7)
+                    )
+                    val newState = State(
```
dimosr: Don't you need to retain the version here, instead of falling back to the default which is suitable when there's no state?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -293,11 +585,56 @@ internal class StatefulSessionManagerImpl(
...
+        }
+    }
+
+    private fun processResponderHandshake(
+        state: State?,
+        message: OutboundSessionMessage.ResponderHandshakeMessage,
+    ): ProcessHandshakeResult?{
+        val metadata = state?.metadata?.let { metadataMap -> SessionMetadata(metadataMap) }
+        return when (metadata?.status) {
+            SessionStatus.SentResponderHandshake -> {
```
dimosr: I think that's the other way around. We should be in `SentInitiatorHandshake` when we receive responder handshake. Being in `SentResponderHandshake` seems an invalid state.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: I didn't think about this in the original design.

So to do what Dimos is saying we can return `SessionAlreadyPending` instead of `CannotEstablishSession` (if the session is pending and the state update fails), which will mean the `OutboundMessageProcessor` will queue the message (from member/flow worker).

This will still work with the new design. The key thing is that we discard the response Session Message if the State Update fails which we are still doing.

YashNabar: Would we need to return `SessionAlreadyPending` considering the updates here will only contain states returned by `newSessionNeeded()`?

dimosr: I think we'd need to check the current status and act accordingly. For example, I can see the following cases:
* current status `SentInitiatorHello` or `SentResponderHello` -> can return `SessionAlreadyPending` so that message is queued.
* current status `SessionReady` -> can return SessionEstablished

wrt to `newSessionNeeded`, I am not sure I am fully following the updated logic. For instance, when there's an existing state at `SentInitiatorHello` or `SentInitiatorHandshake`, why do we need a new session? I would have thought we just check whether there's enough time elapsed since last send, and if so just replay the session message.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -5,51 +5,177 @@ import java.time.Duration
...
+                messagesWithoutKey.add(it)
+                it to null
+            } else {
+                it to calculateOutboundSessionKey(
+                    messageHeader.source.toCorda(), messageHeader.destination.toCorda(), serial
+                )
+            }
+        }
+        val updates = mutableSetOf<State>()
+        val sessionStates = stateManager.get(messageAndKeyMap.values.filterNotNull()).let { state ->
```
dimosr: Reading the state manager docs, it seems it's not returning any entries for keys not found (instead of entries with null values). So, we'll need to make sure to group those together and process them separately, while the current logic below (line 114 at the else conditional) relies on the value being null I think.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -5,51 +5,177 @@ import java.time.Duration
...
+                            counterparties!!,
+                            traceAndState.message.message.header.statusFilter,
+                        )?.let {
+                            updates.add(it.second)
+                            traceAndState to it.first
+                        } ?: (traceAndState to SessionManager.SessionState.CannotEstablishSession)
+                    }
+
+                    SessionStatus.SentInitiatorHello, SessionStatus.SentInitiatorHandshake -> {
+                        state.replaySessionMessage()?.let {
```
dimosr: I think returning `NewSessionNeeded` here can be a bit misleading, since that was supposed to be used when a session doesn't exist at all, while here it exists and we just need to replay. 

I understand why you probably did it, just to comply with the logic in the outbound message processor which otherwise won't send these messages. But, there are a few options that might keep the codebase a bit easier to understand: 
* we could also have a publisher in this component and publish the message here instead of relying on the outbound message processor (we don't require any atomic semantics). That way, we could still return `SessionAlreadyPending` which I think will be slightly cleaner. The downside is if we do that before we try and update the state manager, there's still a chance we'll replay from multiple link managers.
* we could introduce an extra field in `SessionAlreadyPending` for messages that need to be sent by upstream clients. Then, we can include the messages that the outbound message processor needs to send there. That way, we can process the result of the state manager updates and filter out messages in cases where updates failed because another link manager managed to write first (so that we replay only once).

Note: This is more of a cleanliness thing I think, since functionally what you have could work. So, you could do that as a follow-up cleanup if you want.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -342,13 +339,83 @@ internal class StatefulSessionManagerImpl(
...
+
+        val newMetadata = OutboundSessionMetadata(
+            sessionId = initMessage.first.sessionId,
+            source = counterParties.ourId,
+            destination = counterParties.counterpartyId,
+            lastSendTimestamp = Instant.now(),
+            encryptionKeyId = "",
+            encryptionKeyTenant = "",
+            status = SessionStatus.SentInitiatorHello,
+            expiry = Instant.now() + Duration.ofDays(7),
```
dimosr: Judging from the changes here (https://github.com/corda/corda-runtime-os/pull/5441), I think `expiry` is supposed to be a plain timestamp, since the expiration window is added on top of it when calculating when the session has expired.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
dimosr: Resolving, this logic has been swapped around now and is correct.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
dimosr: Resolving, logic has been swapped around so not applicable anymore.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
YashNabar: We map the session states based on the original `messageAndKeyMap`, so `sessionStates` would contain null entries for sessions not found in the state manager - and we process them in the block you pointed out

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
YashNabar: Addressed in [8468379](https://github.com/corda/corda-runtime-os/pull/5438/commits/8468379747f569674657679ca2b0987788d34ab5)

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -5,51 +5,177 @@ import java.time.Duration
...
+        }
+        val updates = mutableSetOf<State>()
+        val sessionStates = stateManager.get(messageAndKeyMap.values.filterNotNull()).let { state ->
+            messageAndKeyMap.mapNotNull {
+                OutboundMessageContext(it.key, state[it.value], getMessage(it.key))
+            }
+        }
+        val resultStates = sessionStates.map { traceAndState ->
+            val counterparties = sessionManagerImpl.getSessionCounterpartiesFromMessage(traceAndState.message.message)
+            if (counterparties == null) {
```
williamvigorr3: This if statement doesn't do anything. I think you need a `return@map` (hence why `counterparties!!`).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: I think we could rename `NewSessionNeeded` to `SendSessionNegotiationMessage` (one button in InteliJ so we can do this quickly now). 

Not sure what the second option would do to the layer above (would the logic become the same for `SessionAlreadyPending` and `NewSessionNeeded `). Not sure I like adding the complexity of adding an extra publisher (first option).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
dimosr: right, I was reading it wrong. I noticed that we filter the null values from `messageAndKeyMap` in the `stateManager.get()`, but not within the `let` block though. Is this intentional? 

Also, does `mapNotNull` have any effect given we always return an `OutboundMessageContext` from the lambda function?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -5,51 +5,177 @@ import java.time.Duration
...
+                    }
+                }
+            }
+        }
+
+        val failures = stateManager.update(updates).also {
+            logger.warn("Failed to update session states for the following keys: ${it.keys}.")
+        }
+
+        // TODO fix
```
williamvigorr3: What do we need to fix?

YashNabar: Need to return the appropriate state here (and also tidy that return a bit)

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -545,10 +548,11 @@ internal class SessionManagerImpl(
         }
     }
 
-    private fun linkOutMessagesFromSessionInitMessages(
+    internal fun linkOutMessagesFromSessionInitMessages(
         sessionCounterparties: SessionCounterparties,
         messages: List<Pair<AuthenticationProtocolInitiator, InitiatorHelloMessage>>,
-        filter: MembershipStatusFilter
+        filter: MembershipStatusFilter,
+        sendHeartbeats: Boolean = true
```
yift-r3: (Nitpick - ) The name confused me a little; I don't think this method sends the heartbeat.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
dimosr: For the second option, it would probably be an extra field in `SessionAlreadyPending` with messages to send optionally & a piece of logic / conditional on the outbound processor that would check in that case whether there are messages to send if a session is already pending.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -621,7 +627,9 @@ internal class SessionManagerImpl(
                 responderMemberInfo,
                 p2pParams.networkType
             )?.let {
-                heartbeatManager.sessionMessageSent(sessionCounterparties, message.first.sessionId)
+                if (sendHeartbeats) {
```
yift-r3: (Nitpick - ) we could do it in a second loop - something like: 
```
        return messages.mapNotNull { message ->
            createLinkOutMessage(
                message.second,
                sessionCounterparties.ourId,
                responderMemberInfo,
                p2pParams.networkType
            )?.let {
                message.first.sessionId to it
            }
        }.also {
            if (sendHeartbeats) {
                it.forEach {(sessionId, _) ->
                    heartbeatManager.sessionMessageSent(
                        sessionCounterparties,
                        sessionId,
                    )
                }
            }
        }
```

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: Note that `stateManager.findByMetadataMatchingAny()` is order O(N) where N is the size of the table as `stateManager.get()` is O(1). So I think we should avoid `stateManager.findByMetadataMatchingAny()` where possible (as we need to lookup outbound sessions by `sessionId` sometimes we can't always avoid this).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
```diff
@@ -649,6 +656,16 @@ internal class SessionManagerImpl(
...
         }
+        return processResponderHello(sessionInfo, session, message, true)?.first
+    }
+
+    @Suppress("ComplexMethod")
+    internal fun processResponderHello(
+        sessionInfo: SessionCounterparties,
+        session: AuthenticationProtocolInitiator,
+        message: ResponderHelloMessage,
+        sendHeartbeatsAndReplaySessionMessages: Boolean = false
```
yift-r3: Maybe the heartbeat flag should be a constructor member of the `SessionManagerImpl`?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerWarnings.kt
```diff
@@ -70,4 +70,11 @@ internal object SessionManagerWarnings {
         this.warn("Could not get session information from message sent from $us" +
                 " to $peer with ID `$messageId`. Peer is not in the members map.")
     }
+
+    internal fun Logger.invalidSessionStatusError(messageName: String, sessionId: String, status: String) {
```
yift-r3: It's easier to find the location of the error in the code when it's in the same class as the logger is defined.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/SessionMetadata.kt
```diff
@@ -0,0 +1,142 @@
...
+import net.corda.v5.base.types.MemberX500Name
+import net.corda.virtualnode.HoldingIdentity
+import java.time.Duration
+import java.time.Instant
+
+enum class SessionStatus{
+    SentInitiatorHello, SentInitiatorHandshake, SentResponderHello, SentResponderHandshake, SessionReady
+}
+
+private fun String.statusFromString(): SessionStatus {
```
yift-r3: Can we put those functions in a companion object?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/SessionMetadata.kt
```diff
@@ -0,0 +1,142 @@
...
+        metadata[ENCRYPTION_KEY_ID].toString(),
+        metadata[ENCRYPTION_TENANT].toString(),
+        metadata[STATUS].toString().statusFromString(),
+        Instant.ofEpochMilli(metadata[EXPIRY] as Long),
+        metadata[SERIAL] as Long,
+        metadata[MEMBERSHIP_STATUS].toString().membershipStatusFromString(),
+        metadata[COMMUNICATION_WITH_MGM].toString().toBoolean(),
+    )
+
+    fun lastSendExpired(): Boolean {
```
yift-r3: Can you call this with a clock?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/SessionMetadata.kt
```diff
@@ -0,0 +1,142 @@
...
+        const val GROUP_ID_KEY = "groupId"
+        const val LAST_SEND_TIMESTAMP = "lastSendTimestamp"
+        const val ENCRYPTION_KEY_ID = "encryptionKeyId"
+        const val ENCRYPTION_TENANT = "encryptionTenant"
+        const val STATUS = "status"
+        const val EXPIRY = "expiry"
+        val SESSION_EXPIRY_PERIOD: Duration = Duration.ofDays(7)
+    }
+
+    constructor(metadata: Metadata): this(
```
yift-r3: Instead of second constructors, can we have a function in the companion objects?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/SessionMetadata.kt
```diff
@@ -0,0 +1,142 @@
...
+        return Metadata(
+            mapOf(
+                SESSION_ID to this.sessionId,
+                SOURCE_VNODE to this.source.x500Name.toString(),
+                DEST_VNODE to this.destination.x500Name.toString(),
+                GROUP_ID_KEY to this.source.groupId,
+                LAST_SEND_TIMESTAMP to this.lastSendTimestamp.toEpochMilli(),
+                ENCRYPTION_KEY_ID to this.encryptionKeyId,
+                ENCRYPTION_TENANT to this.encryptionKeyTenant,
+                STATUS to this.status,
```
yift-r3: toString

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/SessionMetadata.kt
```diff
@@ -0,0 +1,142 @@
...
+                EXPIRY to this.expiry.toEpochMilli(),
+                SERIAL to this.serial,
+                MEMBERSHIP_STATUS to this.membershipStatus,
```
yift-r3: toString

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -5,51 +5,177 @@ import java.time.Duration
...
 import net.corda.schema.registry.AvroSchemaRegistry
+import net.corda.v5.crypto.DigestAlgorithmName
+import net.corda.v5.crypto.SecureHash
 import net.corda.virtualnode.HoldingIdentity
 import net.corda.virtualnode.toCorda
 import org.slf4j.LoggerFactory
+import java.security.MessageDigest
 import net.corda.data.p2p.state.SessionState as AvroSessionState
 
+@Suppress("TooManyFunctions", "LongParameterList", "LargeClass")
```
yift-r3, line 57: (Probably not in this ticket - ) we should break this class up, it's too big and complicated.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -5,51 +5,177 @@ import java.time.Duration
...
     private val sessionEncryptionOpsClient: SessionEncryptionOpsClient,
+    private val membershipGroupReaderProvider: MembershipGroupReaderProvider,
 ): SessionManager {
 
     override fun <T> processOutboundMessages(
         wrappedMessages: Collection<T>,
         getMessage: (T) -> AuthenticatedMessageAndKey
     ): Collection<Pair<T, SessionManager.SessionState>> {
-        return emptyList()
+        val messagesWithoutKey = mutableSetOf<T>()
```
yift-r3: I'm not sure we need this to be mutable; we can just do something like:
```
val messagesWithoutKey = messageAndKeyMap.filter { it.value != null }.keys
```
after the associate

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -5,51 +5,177 @@ import java.time.Duration
...
+                )
+                messagesWithoutKey.add(it)
+                it to null
+            } else {
+                it to calculateOutboundSessionKey(
+                    messageHeader.source.toCorda(), messageHeader.destination.toCorda(), serial
+                )
+            }
+        }
+        val updates = mutableSetOf<State>()
```
williamvigorr3: We should check the cache here (to avoid using the state manager if the Session is already in memory). 

To do this we need to key the cache by session counterparties as well as `sessionId`.  In the old session manager this was done with two concurrent hash maps (in `OutboundSessionPool`).

```
    private val outboundSessions = ConcurrentHashMap<SessionCounterpartiesKey, ConcurrentHashMap<String, SessionType>>()
    private val counterpartiesForSessionId = ConcurrentHashMap<String, SessionCounterpartiesKey>()
```

Then we can do a look up like this `outboundSessions[counterpartiesForSessionId[sessionId]]`. We need to do something similar (which needs to work with the caffeine cache).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -5,51 +5,177 @@ import java.time.Duration
...
+        val sessionStates = stateManager.get(messageAndKeyMap.values.filterNotNull()).let { state ->
+            messageAndKeyMap.mapNotNull {
+                OutboundMessageContext(it.key, state[it.value], getMessage(it.key))
```
yift-r3: Can't we reuse the getMessage from line 63?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -5,51 +5,177 @@ import java.time.Duration
...
+            }
+        }
+        val resultStates = sessionStates.map { traceAndState ->
+            val counterparties = sessionManagerImpl.getSessionCounterpartiesFromMessage(traceAndState.message.message)
+            if (counterparties == null) {
+                traceAndState to SessionManager.SessionState.CannotEstablishSession
```
yift-r3: Don't we want to log different messages for different `CannotEstablishSession` use cases?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -5,51 +5,177 @@ import java.time.Duration
...
+                            metadata.sessionId,
+                            metadata.status.toString()
+                        )
+                        traceAndState to SessionManager.SessionState.CannotEstablishSession
+                    }
+                }
+            }
+        }
+
+        val failures = stateManager.update(updates).also {
```
yift-r3: Can you check if the updates is not empty?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
yift-r3: The `resultStates` could be a collection of triplets, and then we can avoid this mutable set...

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
yift-r3: Can we avoid calling `getSessionId` three times?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -5,51 +5,177 @@ import java.time.Duration
...
         uuids: Collection<T>,
         getSessionId: (T) -> String
     ): Collection<Pair<T, SessionManager.SessionDirection>> {
         val traceable = uuids.associateBy { getSessionId(it) }
-        val sessionFromCache = uuids.map { it to cachedInboundSessions[getSessionId(it)] }
-        val sessionsIdsNotInCache = sessionFromCache.filter { it.second == null }.map { getSessionId(it.first) }
+        val sessionsFromCache = uuids.map {
+            it to (cachedInboundSessions[getSessionId(it)] ?: cachedOutboundSessions[getSessionId(it)])
+        }
+        val sessionsNotInCache = sessionsFromCache.filter { it.second == null }.associate { it.first to getSessionId(it.first) }
```
yift-r3: Four times...

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -5,51 +5,177 @@ import java.time.Duration
...
         val inboundSessionsFromStateManager: List<Pair<T, SessionManager.SessionDirection>> =
-            stateManager.get(sessionsIdsNotInCache).entries.mapNotNull { (sessionId, state) ->
+            stateManager.get(sessionsNotInCache.values).entries.mapNotNull { (sessionId, state) ->
```
yift-r3: I think we can get the inbound and outbound sessions in here. Can't we?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -60,26 +186,69 @@ internal class StatefulSessionManagerImpl(
                 it to SessionManager.SessionDirection.Inbound(InboundSessionMetadata(state.metadata).toCounterparties(), session)
             }
         }
-        // In CORE-18630 check cache/state manager for outbound sessions if not found for inbound sessions.
-        // We should avoid as unnecessary reads of the state manager, so first check the inbound and outbound session cache,
-        // then the state manager for an inbound session (query with session ID), then the state manager for an outbound session
-        // (query for sessionId in metadata).
+        val sessionsNotInInboundStateManager =
+            (sessionsNotInCache.keys - inboundSessionsFromStateManager.map { it.first }.toSet()).map {
+                getSessionIdFilter(getSessionId(it))
```
yift-r3: Is that the fifth time we are calling `getSessionId`?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
yift-r3: Can't we key both by session ID (maybe add input/output before hand?)

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -110,6 +279,37 @@ internal class StatefulSessionManagerImpl(
...
+        val trace: T
+    )
+
+    private data class OutboundMessageContext<T>(
+        val trace: T, val state: State?, val message: AuthenticatedMessageAndKey
+    )
+
+    private fun calculateOutboundSessionKey(
+        source: HoldingIdentity, destination: HoldingIdentity, serial: Long
+    ) = SessionCounterpartiesKey(source, destination, serial).hash.toHexString()
```
yift-r3: base64 will be shorter than hex...

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -110,6 +279,37 @@ internal class StatefulSessionManagerImpl(
...
+    private data class SessionCounterpartiesKey(
+        override val ourId: HoldingIdentity,
+        override val counterpartyId: HoldingIdentity,
+        val serial: Long,
+    ): SessionManager.BaseCounterparties {
+        val hash: SecureHash by lazy(LazyThreadSafetyMode.PUBLICATION) {
+            val s = (ourId.x500Name.toString() + counterpartyId.x500Name.toString() + serial.toString())
+            val digest: MessageDigest = MessageDigest.getInstance(DigestAlgorithmName.SHA2_256.name)
+            val hash: ByteArray = digest.digest(s.toByteArray())
+            SecureHashImpl(DigestAlgorithmName.SHA2_256.name, hash)
```
yift-r3: There's a small risk of a collision here...
Is there a security issue with just using the identities and serial? Is the key raw in the database too short for that?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -119,28 +319,109 @@ internal class StatefulSessionManagerImpl(
         ): InboundSessionMessage()
     }
 
+    private sealed class OutboundSessionMessage {
+        data class ResponderHelloMessage(
+            val responderHelloMessage: net.corda.data.p2p.crypto.ResponderHelloMessage
+        ): OutboundSessionMessage()
+        data class ResponderHandshakeMessage(
+            val responderHandshakeMessage: net.corda.data.p2p.crypto.ResponderHandshakeMessage
```
yift-r3: Can you import those?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -119,28 +319,109 @@ internal class StatefulSessionManagerImpl(
...
         val traceable: T,
-        val message: LinkOutMessage,
+        val message: LinkOutMessage?,
         val stateUpdate: State,
         val sessionToCache: Session?
     )
 
     private val cachedInboundSessions = ConcurrentHashMap<String, SessionManager.SessionDirection>()
 
+    private val cachedOutboundSessions = ConcurrentHashMap<String, SessionManager.SessionDirection>()
```
yift-r3: Change to caffeine...

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -154,26 +435,33 @@ internal class StatefulSessionManagerImpl(
...
-        return result.map{
-            it.sessionToCache?.let { sessionToCache ->
-                cachedInboundSessions.put(
-                    sessionToCache.sessionId,
-                    SessionManager.SessionDirection.Inbound(
-                        InboundSessionMetadata(it.stateUpdate.metadata).toCounterparties(),
-                        sessionToCache
-                    )
-                )
+        val states = stateManager.findByMetadataMatchingAny(messageContexts.map { getSessionIdFilter(it.sessionId) })
```
yift-r3: Can you check that the messageContexts is not empty?

Regardless, I think that this might be a performance issue.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/SessionMetadata.kt
```diff
@@ -0,0 +1,142 @@
...
+
+import net.corda.data.p2p.app.MembershipStatusFilter
+import net.corda.libs.statemanager.api.Metadata
+import net.corda.v5.base.types.MemberX500Name
+import net.corda.virtualnode.HoldingIdentity
+import java.time.Duration
+import java.time.Instant
+
+enum class SessionStatus{
+    SentInitiatorHello, SentInitiatorHandshake, SentResponderHello, SentResponderHandshake, SessionReady
```
yift-r3: Do we want to keep the inbound and outbound statuses in the same enum?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -228,28 +517,116 @@ internal class StatefulSessionManagerImpl(
...
+                    schemaRegistry,
+                    sessionEncryptionOpsClient,
+                    sessionManagerImpl.revocationCheckerClient::checkRevocation
+                ).sessionData as AuthenticationProtocolInitiator
+                val counterparties = state.getSessionCounterparties()
+
+                sessionManagerImpl.processResponderHello(
+                    counterparties, sessionState, message.responderHelloMessage
+                )?.let { (responseMessage, authenticationProtocol) ->
+                    val timestamp = Instant.now()
```
yift-r3: Can you use clock instead of Instant.now?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -228,28 +517,116 @@ internal class StatefulSessionManagerImpl(
...
+                    val updatedMetadata = metadata.copy(
+                        sessionId = message.responderHelloMessage.header.sessionId,
+                        source = metadata.source,
+                        destination = metadata.destination,
+                        lastSendTimestamp = timestamp,
+                        encryptionKeyId = "",
+                        encryptionKeyTenant = "",
+                        status = SessionStatus.SentInitiatorHandshake,
+                        expiry = Instant.now() + Duration.ofDays(7)
```
yift-r3: we have the timestamp (line 544).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -228,28 +517,116 @@ internal class StatefulSessionManagerImpl(
...
+                ).sessionData as AuthenticationProtocolInitiator
+                val counterparties = state.getSessionCounterparties()
+
+                sessionManagerImpl.processResponderHello(
+                    counterparties, sessionState, message.responderHelloMessage
+                )?.let { (responseMessage, authenticationProtocol) ->
+                    val timestamp = Instant.now()
+                    val updatedMetadata = metadata.copy(
+                        sessionId = message.responderHelloMessage.header.sessionId,
+                        source = metadata.source,
```
yift-r3: If we do `metadata.copy` we don't need to pass the source and destination...

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -321,5 +759,30 @@ internal class StatefulSessionManagerImpl(
         }
     }
 
+    private fun <T> Any.getSessionIdIfOutboundSessionMessage(trace: T): OutboundSessionMessageContext<T>? {
+        return when (this) {
+            is ResponderHelloMessage -> OutboundSessionMessageContext(
+                this.header!!.sessionId, OutboundSessionMessage.ResponderHelloMessage(this), trace
```
yift-r3: This function returns nullable result, we can remove the !! and use ?

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: The raw key in the DB is too small (there's no security issue here). Isn't the chance of a collision vanishingly small: (https://en.wikipedia.org/wiki/Birthday_problem#Probability_table) i.e. Looking at the table if we stored 4.8×10^32 sessions in the state manager the probability of a collision would be 10^-12 (this assumes SHA2_256 is random).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: These fields have been deleted now.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/SessionMetadata.kt
williamvigorr3: I've moved them to separate enums. This simplifies the logic in `StatefulSessionManagerImpl`.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: Fixed in 7fdfc27

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
williamvigorr3: I renamed this `trackSessionHealthAndReplaySessionMessages` and moved it to the constructor. In b164401b2c.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
williamvigorr3: I moved it to the constructor in b164401b2c.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerWarnings.kt
williamvigorr3: I moved this into `StatefulSessionManagerImpl` in b164401b2c.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/SessionMetadata.kt
williamvigorr3: Fixed in b164401b2c.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/SessionMetadata.kt
williamvigorr3: Fixed in 7fdfc27a.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/SessionMetadata.kt
williamvigorr3: This is fixed.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/SessionMetadata.kt
williamvigorr3: This is fixed.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/SessionMetadata.kt
williamvigorr3: Fixed in fa43394f3f

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: This has been fixed.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: This has been fixed.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: This has been fixed.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: This has been rewritten (so there is no longer a mutable set).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: This has been fixed.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: > right, I was reading it wrong. I noticed that we filter the null values from `messageAndKeyMap` in the `stateManager.get()`, but not within the `let` block though. Is this intentional?
> 
StateManager.get cannot return a null entry in the Map (not 100% sure what you are asking here).

> Also, does `mapNotNull` have any effect given we always return an `OutboundMessageContext` from the lambda function?

This code has been rewritten (so the mapNotNull is gone). To give some context (to why this has changed) Yiftach worked out we need to grouping all off the messages which can cause a single state update.

dimosr: Don't worry about it given it's been refactored, it looks good now.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: I changed this line: https://github.com/corda/corda-runtime-os/blob/2ac9016e5438a8c88ee9916b8aba120efc3130da/components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/metadata/SessionMetadata.kt#L167 to not add on any expiry time.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/SessionManagerImpl.kt
williamvigorr3: This has been done in 8b27fac17b4cb13c5746fa5f01a2c2e3adf98c2c.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: I've raised https://r3-cev.atlassian.net/browse/CORE-19400 to do this + improve the unit tests.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: We only call this once now.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: We only call this once now.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: Fixed in 8b27fac17b4cb13c5746fa5f01a2c2e3adf98c2c

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: We encode this to base64 now.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: We key them separately. This discussion is a duplicate of https://github.com/corda/corda-runtime-os/pull/5438#discussion_r1452477623 (so I will resolve it and we can discuss on the other comment).

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: Is not empty was added in 7950cbf8.

yift-r3 (approved):

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: This has been fixed now.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: I added this to https://r3-cev.atlassian.net/browse/CORE-19400.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
YashNabar: This has been fixed

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
YashNabar: Addressed in [4e0636c](https://github.com/corda/corda-runtime-os/pull/5438/commits/4e0636c3f206a035876924e692124c349da959d8)

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
williamvigorr3: Should I raise a follow up ticket to do this?

williamvigorr3 (approved):
LGTM

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
dimosr: yeah, if the other people agree, we can raise a follow-up ticket to refactor this a bit.

PR #5439, opened 2024-01-15
CORE-19295: Clean Up Checkpoints Only

jujoramos:
Fix "FlowTimeoutTaskProcessor" implementation so that only states
representing flow checkpoints are marked for deletion. Other types of
states should not be automatically deleted as that might cause issues,
especially when the persistent storage is shared  to store multiple
state types.


anien (approved):
LGTM (only crypto).

filipesoliveira (approved):

JamesHR3 (approved):
This is safer in the existing code, but will prevent cleanup where the processor ends up not getting a say (e.g. timeout of processing an event that has not yet generated a state).

LWogan (approved):
lgtm

PR #5441, opened 2024-01-15
CORE-18631: Start unit tests, some clean up

yift-r3:


components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -153,24 +146,30 @@ internal class StatefulSessionManagerImpl(
...
+        }.onEach { result ->
+            result.sessionToCache?.let { sessionToCache ->
+                cachedInboundSessions[sessionToCache.sessionId] = SessionManager.SessionDirection.Inbound(
+                    result.stateUpdate.metadata.from().toCounterparties(),
+                    sessionToCache,
                 )
             }
-            it.traceable to it.message
+        }.map { result ->
+            result.traceable to result.message
```
williamvigorr3: This problem is already here but you refactored this a lot. (Its a lot clearer now makes sense). This needs to return `result.traceable to null` if `processInitiatorHello/processInitiatorHandshake` returns null.

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -1,94 +1,100 @@
...
-                it to SessionManager.SessionDirection.Inbound(InboundSessionMetadata(state.metadata).toCounterparties(), session)
+            if (sessionsIdsNotInCache.isEmpty()) {
+                emptyList()
+            } else {
+                stateManager.get(sessionsIdsNotInCache.keys).entries.mapNotNull { (sessionId, state) ->
+                    val session = stateConvertor.toCordaSessionState(
+                        state,
+                        sessionManagerImpl.revocationCheckerClient::checkRevocation,
+                    ).sessionData as? Session
+                    session?.let {
```
williamvigorr3, line 66: We should put the found session in the cache here (I missed this in #5424):
```kotlin
session?.let {
    sessionsIdsNotInCache[sessionId]?.let {
        val inboundSession =  SessionManager.SessionDirection.Inbound(state.metadata.from().toCounterparties(), session)
        cachedInboundSessions.put(sessionId, inboundSession)
        it to inboundSession
    }
}
```

components/link-manager/src/main/kotlin/net/corda/p2p/linkmanager/sessions/StatefulSessionManagerImpl.kt
```diff
@@ -1,94 +1,100 @@
...
             }
-        }
         // In CORE-18630 check cache/state manager for outbound sessions if not found for inbound sessions.
         // We should avoid as unnecessary reads of the state manager, so first check the inbound and outbound session cache,
         // then the state manager for an inbound session (query with session ID), then the state manager for an outbound session
         // (query for sessionId in metadata).
 
-        return sessionFromCache.mapNotNull { (traceable, sessionDirection) ->
-            sessionDirection?.let { traceable to it }
+        return sessionFromCache.mapNotNull { (sessionId, sessionDirection) ->
```
williamvigorr3, line 80: We need to return `NoSession` if we can't find the Session in either the state manager or cache (so we will need to refactor this to do that but it probably makes sense to leave this as is and do that in #5438). FYI @YashNabar

williamvigorr3 (commented):
Looks good. I think we should fix one thing now.

williamvigorr3 (approved):
LGTM

PR #5442, opened 2024-01-15
CORE-19020 idempotent flow engine logic

LWogan:
Migrate replay logic to the Flow pipeline. Revert changes to mediator not applicable anymore. Store hash of consumer input record to all records passed to processor.

https://github.com/corda/corda-api/pull/1462


libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessor.kt
```diff
@@ -29,14 +29,20 @@ class EventProcessor<K : Any, S : Any, E : Any>(
 
     /**
      * Process a group of events.
-     * @param group Group of records of various keys
-     * @param retrievedStates states for a group
+     * If an record is one which the mediator has processed before then the asynchronous outputs from the previous invocation
+     * of the processor from the [MediatorState]. The message processor is not executed and the retrieved outputs are returned to be resent.
```
ben-millar:
```suggestion
     * If the mediator has previously processed a record, then the asynchronous outputs from the previous invocation of the processor will be retrieved from the [MediatorState].
     * In this case, the message processor is not executed again, and the previously retrieved outputs are returned for resending.
```

LWogan:
build e2e

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessor.kt
```diff
@@ -46,43 +56,69 @@ class EventProcessor<K : Any, S : Any, E : Any>(
...
+        mediatorState: MediatorState
+    ): Pair<List<Record<K, E>>, List<MediatorMessage<Any>>> {
+        val cachedOutputs = mutableListOf<MediatorMessage<Any>>()
+        val nonReplayOutputs = allConsumerInputs.filter { inputEvent ->
+            mediatorReplayService.getReplayEvents(inputEvent, mediatorState)?.let {
+                cachedOutputs.addAll(it)
+            } == null
         }
+        return Pair(nonReplayOutputs, cachedOutputs)
     }
```
ben-millar: This works, but I have some concerns about the efficiency. In the worst case (no replay events in the state), we compare every single input event we receive to every single input event ever received until that point. Some potential optimisations could be:
1. Retrieve all cached events once from the mediator state, process them into a map, and use this for lookups (requires only a single traversal of past events per batch).
2. Modify `getReplayEvents()` to take a set of `Record<K, E>` and return a map of `Record<K, E> to List<MediatorMessage<Any>>` replay events. (Requires at worst a single traversal; less if replay events are recent)

LWogan:
build e2e

ben-millar (dismissed):
LGTM

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessor.kt
```diff
@@ -46,43 +56,68 @@ class EventProcessor<K : Any, S : Any, E : Any>(
...
-                        processorState = response.updatedState
-                        val (syncEvents, asyncEvents) = response.responseEvents.map { convertToMessage(it) }.partition {
-                            messageRouter.getDestination(it).type == RoutingDestination.Type.SYNCHRONOUS
+            val (nonReplayConsumerInputs, replayOutputs) = getReplayOutputsAndNonReplayInputs(allConsumerInputs, mediatorState)
+            if (nonReplayConsumerInputs.isEmpty()) {
+                EventProcessingOutput(replayOutputs, StateChangeAndOperation.Noop)
+            } else {
+                val asyncOutputs = mutableMapOf<Record<K, E>, MutableList<MediatorMessage<Any>>>()
+                val processed = try {
+                    nonReplayConsumerInputs.onEach { consumerInputEvent ->
```
mbrkic-r3:
```suggestion
                    nonReplayConsumerInputs.forEach { consumerInputEvent ->
```

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/EventProcessor.kt
```diff
@@ -46,43 +56,68 @@ class EventProcessor<K : Any, S : Any, E : Any>(
...
+        state: State?,
+        processed: State?
+    ) = when {
+        state == null && processed != null -> StateChangeAndOperation.Create(processed)
+        state != null && processed != null -> StateChangeAndOperation.Update(processed)
+        state != null && processed == null -> StateChangeAndOperation.Delete(state)
+        else -> StateChangeAndOperation.Noop
+    }
+
+    private fun getReplayOutputsAndNonReplayInputs(
```
mbrkic-r3: With initial version of Event Mediator we had some issues if output events were not sent in order (of related input events). Not sure if that was fixed in the meanwhile so output event could now be sent in any order. If this could still be an issue then we need to keep order.

LWogan: I think this was due to a bug in the flow mapper which was subsequently fixed. Session messages were being blocked when the mapper was in state CLOSING. This meant the order of the last session data messages and close messages mattered.  In general I think we can't rely on the kafka producers to send in order anymore. We currently set some producer client settings in a way that order is still preserved but we should assume it may not be configured that way in the future.

All that aside, within the the context of a single input key, output events order is still the same.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/MediatorReplayService.kt
```diff
@@ -23,26 +25,28 @@ class MediatorReplayService @Activate constructor(
...
+    fun getOutputEvents(
         existingOutputs: List<MediatorReplayOutputEvents>,
-        newOutputs: Map<Record<K, E>, MutableList<MediatorMessage<Any>>>
+        newOutputs: Map<ByteBuffer, MutableList<MediatorMessage<Any>>>
     ): List<MediatorReplayOutputEvents> {
         val mediatorOutputs = existingOutputs.toMutableList()
 
         newOutputs.forEach { entry ->
-            val hash = getInputHash(entry.key)
+            val hash = entry.key
```
mbrkic-r3: Why do we use key instead of hash?

components/flow/flow-mapper-service/src/main/kotlin/net/corda/session/mapper/messaging/mediator/FlowMapperEventMediatorFactoryImpl.kt
```diff
@@ -86,6 +86,7 @@ class FlowMapperEventMediatorFactoryImpl @Activate constructor(
         .threads(messagingConfig.getInt(MEDIATOR_PROCESSING_THREAD_POOL_SIZE))
         .threadName("flow-mapper-event-mediator")
         .stateManager(stateManager)
+        .idempotentProcessor(false)
```
mbrkic-r3: Is this because FlowMapper already handles duplicated messages?

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/MediatorReplayService.kt
```diff
@@ -57,22 +61,51 @@ class MediatorReplayService @Activate constructor(
         return mediatorOutputs
     }
 
+    private fun <T> serializeOutput(any: T?): MediatorReplayOutputEventType? {
```
mbrkic-r3: We might create a separate class that wraps serializers/deserializer and works with `MediatorReplayOutputEventType` (which might be renamed to `SerializedObject`).

LWogan: i couldn't think of a good name. SerializedObject is better. ill update.

components/flow/flow-mapper-service/src/main/kotlin/net/corda/session/mapper/messaging/mediator/FlowMapperEventMediatorFactoryImpl.kt
LWogan: at the moment this feature is only desirable for the flow engine. Flow mapper has logic to block messages which have expired. Duplicates/replays that arrive late, after the TTL expires, we would want to discard.

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/processor/MediatorReplayService.kt
LWogan: I changed the service to take the hash as an input. This allowed me to deduplicate messages in the EventProcessor that had no value in the mediator state but we're duplicates in the same poll batch. I needed to calculate the hash per input to be able to dedupe.

The requirement given recently was that duplicates within a single poll should result in the outputs replayed and message processor executed once,

The code has gone through many iterations. But it works now as follows:
- give mediator replay service all input records and return all replays from mediator state. Traverse once.
- process non-replays(can contain duplicates), store outputs in a map keyed by the input hash in EventProcessor
- detect duplicates via the input hash map in EventProcessor

mbrkic-r3: OK, I missed that hash is sent in.

Regarding processing described in the comment, can it be simplified by having the 2nd step handle deduplication done by the 1st step as well (first step is removed, and we check for duplicates against the stored input hashes that are updated with each processed event)?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/FlowEngineReplayService.kt
```diff
@@ -0,0 +1,33 @@
...
+     * associated with that hash.
+     * @param inputEventHash The hash of the input event
+     * @param checkpoint The checkpoint
+     * @return If the [inputEventHash] is present, return the list of output events associated with this input event.
+     * If it is not present, then it is not replayed event. Null is returned
+     */
+    fun getReplayEvents(
+        inputEventHash: String,
+        checkpoint: Checkpoint?
+    ): List<Record<*, *>>?
```
mbrkic-r3, line 23: We could avoid handling `null` values by returning `emptyList()` instead.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
```diff
@@ -158,11 +165,20 @@ class FlowEventProcessorImpl(
...
+            result.checkpoint.saveOutputs(flowEngineReplayService.generateSavedOutputs(hash, outputs))
+        }
         return flowEventContextConverter.convert(
-            result.copy(outputRecords = result.outputRecords + cleanupEvents)
+            result.copy(outputRecords = outputs)
         )
     }
 
+    private fun getInputEventHash(event: Record<String, FlowEvent>) =
+        event.headers.find { it.first == INPUT_HASH_HEADER }?.second
```
mbrkic-r3: Do we have a use case where input hash header can be missing? If not, we shouldn't return `null` from this function but  throw exception instead.

libs/flows/flow-api/src/main/kotlin/net/corda/flow/pipeline/events/FlowEventContext.kt
```diff
@@ -38,5 +40,6 @@ data class FlowEventContext<T>(
     val mdcProperties: Map<String, String>,
     val flowMetrics: FlowMetrics,
     val flowTraceContext: TraceContext,
-    val metadata: Metadata?
+    val metadata: Metadata?,
+    val inputEventHash: String?,
```
mbrkic-r3, line 44: Why is hash `String` and not a byte array?

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/MediatorInputService.kt
```diff
@@ -0,0 +1,18 @@
...
+    companion object {
+        const val INPUT_HASH_HEADER = "InputHash"
+    }
+
+    /**
+     * Generate a unique SHA256 hash, converted to a UUID, for the consumer input record.
+     * @param inputEvent The consumer input event polled from the bus
+     * @return A hash of the input event as bytes, converted a UUID
+     */
+    fun <K : Any, E : Any> getHash(inputEvent: Record<K, E>): UUID
```
mbrkic-r3: Why do we convert bytes to UUID?

components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
```diff
@@ -146,8 +146,6 @@ class FlowEventMediatorFactoryImpl @Activate constructor(
                     rpcEndpoint(VERIFICATION_WORKER_REST_ENDPOINT, VERIFICATION_PATH), SYNCHRONOUS)
                 is UniquenessCheckRequestAvro -> routeTo(rpcClient,
                     rpcEndpoint(UNIQUENESS_WORKER_REST_ENDPOINT, UNIQUENESS_PATH), SYNCHRONOUS)
-                is FlowEvent -> routeTo(messageBusClient,
```
mbrkic-r3: What is the reason for removing this?

libs/flows/flow-api/src/main/kotlin/net/corda/flow/pipeline/events/FlowEventContext.kt
LWogan: record headers only accept strings.

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/MediatorInputService.kt
LWogan: record headers only allow strings. UUID seemed like the best approach

components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
LWogan: dead code, we dont use this topic. Not sure if perf team have a ticket to clean this up properly and remove all references/api etc ill check with them

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/FlowEngineReplayService.kt
LWogan: i use null to detect when its not a replay.

empty list is a valid replay response. an input in the past might have no outputs

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
LWogan: I was thinking about backwards compatibility. Reading events from a topic with a newer version of Corda. Maybe thats not a valid concern?

components/flow/flow-service/src/main/kotlin/net/corda/flow/messaging/mediator/FlowEventMediatorFactoryImpl.kt
mbrkic-r3: This topic was still used for 5.1, event after new topics were introduced (I think it was some scenario related to retries, maybe external events). If it's really not needed anymore it might be better to remove it with another PR, that would also remove this topic and related permissions from API repo.

libs/flows/flow-api/src/main/kotlin/net/corda/flow/pipeline/events/FlowEventContext.kt
mbrkic-r3: Proper solution would be to use `MediatorMessage` for inputs/consumers as well. This might require considerable refactoring so probably best to have it as a tech debt for now.

libs/messaging/messaging/src/main/kotlin/net/corda/messaging/api/mediator/MediatorInputService.kt
mbrkic-r3: Base64 representation would be more compact than hex, although in this case it might not be a big difference. Since we can store only String at the moment, wouldn't be better to return String (so it doesn't have to be converted every time later)?

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
mbrkic-r3: Don't we set this header after events are polled?

LWogan:
build e2e

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
LWogan: This is true :D I'll update it

libs/messaging/messaging-impl/src/main/kotlin/net/corda/messaging/mediator/MediatorInputServiceImpl.kt
```diff
@@ -0,0 +1,28 @@
...
+    private val cordaAvroSerializationFactory: CordaAvroSerializationFactory,
+): MediatorInputService {
+    private val serializer = cordaAvroSerializationFactory.createAvroSerializer<Any> { }
+
+    override fun <K : Any, E : Any> getHash(inputEvent: Record<K, E>): String {
+        val recordValueBytes = serialize(inputEvent.value)
+        check(recordValueBytes != null) {
+            "Input record key and value bytes should not be null"
+        }
+        return UUID.nameUUIDFromBytes(recordValueBytes.sha256Bytes()).toString()
```
mbrkic-r3: [As commented in other PR](https://github.com/corda/corda-runtime-os/pull/5562#discussion_r1475862568), it's best not to use UUID. Since hash is here used only to detect duplicates (i.e. not for any security purposes), we could use MD5 and store it as bytes in Checkpoint to reduce size per stored input event.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEngineReplayServiceImpl.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    override fun getReplayEvents(
+        inputEventHash: String,
+        checkpoint: Checkpoint?
+    ): List<Record<*, *>>? {
+        if (checkpoint == null) return null
+        val replayEvents = mutableListOf<Record<*, *>>()
+        var isReplay = false
+        checkpoint.savedOutputs.asReversed().forEach { replay ->
```
mbrkic-r3: We might convert to `Map` to check for a match, although with a small number of elements it might not be faster. If it stays like this we could `return` as soon as it's found, no need to continue checking as we should not have duplicate hashes in the `savedOutputs`.

components/flow/flow-service/src/main/kotlin/net/corda/flow/state/impl/FlowCheckpointImpl.kt
```diff
@@ -234,6 +238,11 @@ class FlowCheckpointImpl(
             .build()
     }
 
+    override fun saveOutputs(savedOutputs: SavedOutputs) {
+        val existingOutputs = checkpoint.savedOutputs?.toMutableList() ?: mutableListOf()
+        checkpoint.savedOutputs = existingOutputs.plus(savedOutputs)
```
mbrkic-r3, line 243: Function `plus` will create a new list so in that case `existingOutputs` doesn't need to be mutable, or function `add` could be used instead with mutable list.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEventProcessorImpl.kt
```diff
@@ -87,10 +91,19 @@ class FlowEventProcessorImpl(
...
+        val inputEventHash = getInputEventHash(event)
+        if (!isSyncResponse(event)) {
+            flowEngineReplayService.getReplayEvents(inputEventHash, state?.value)?.let { replays ->
+                log.debug { "Detected input event that has been processed previously for hash :$inputEventHash" }
+                return StateAndEventProcessor.Response(state, replays)
+            }
+        }
+
         val pipeline = try {
             log.trace { "Flow [${event.key}] Received event: ${flowEvent.payload::class.java} / ${flowEvent.payload}" }
```
mbrkic-r3, line 103: We might move this to the top of the function.

LWogan: I think its here on purpose as flow event can be null at the top of this function.

components/flow/flow-service/src/main/kotlin/net/corda/flow/pipeline/impl/FlowEngineReplayServiceImpl.kt
LWogan: refactored to use find and map

LWogan:
build e2e

LWogan:
build e2e

mbrkic-r3 (dismissed):
LGTM

LWogan:
build e2e

LWogan:
build e2e

thiagoviana (approved):

components/flow/flow-service/src/main/kotlin/net/corda/flow/external/events/impl/ExternalEventManagerImpl.kt
```diff
@@ -197,9 +197,7 @@ class ExternalEventManagerImpl(
     private fun generateRecord(externalEventState: ExternalEventState, instant: Instant) : Record<*, *> {
         val eventToSend = externalEventState.eventToSend
         eventToSend.timestamp = instant
-
-        // The below placeholder is needed until the Record.topic field is made optional as part of CORE-17887
-        val topic = eventToSend.topic ?: "EMPTY_TOPIC_PLACEHOLDER"
+        val topic = eventToSend.topic
```
ben-millar, line 200: 👍🏻

ben-millar (approved):

PR #5443, opened 2024-01-15
CORE-18930: make transaction ID deterministic

emilybowe:
The privacy salt currently gets a unique ID for transactions that is random. We need it to be unique but also deterministic. The random aspect will be replaced with a concatenation of FlowID+suspendCount+counter which are obtained from the checkpoint. To get this data, FlowFiberService will be used. A service will be created that uses FlowFiberService to give our privacy salt, so this can be injected in several places. Testing should show that transactions with the same flowID and suspendCount and counter give the same ID (deterministic) even if created multiple times.


components/ledger/ledger-common-flow-api/src/main/kotlin/net/corda/ledger/common/flow/transaction/PrivacySaltProviderService.kt
```diff
@@ -0,0 +1,11 @@
+package net.corda.ledger.common.flow.transaction
+
+import net.corda.ledger.common.data.transaction.PrivacySalt
+
+fun interface PrivacySaltProviderService {
+
+    /**
+     * Returns a privacy salt
```
LWogan: We should expand this kdoc to describe that it creates a deterministic unique salt based on a flows information.

components/ledger/ledger-common-flow/src/main/kotlin/net/corda/ledger/common/flow/impl/transaction/PrivacySaltProviderServiceImpl.kt
```diff
@@ -0,0 +1,23 @@
...
+@Component(service = [PrivacySaltProviderService::class])
+class PrivacySaltProviderServiceImpl @Activate constructor(
+    @Reference(service = FlowFiberService::class)
+    private val flowFiberService: FlowFiberService
+) : PrivacySaltProviderService {
+
+    override fun generatePrivacySalt(): PrivacySalt {
+        val flowID = flowFiberService.getExecutingFiber().getExecutionContext().flowCheckpoint.flowId
+        val suspendCount = flowFiberService.getExecutingFiber().getExecutionContext().flowCheckpoint.suspendCount
+        val input = (flowID + suspendCount).repeat(10)
```
LWogan: Ah i see there's a minimum size requirement for the salt. I think UUIDs length should meet the constraint though

libs/ledger/ledger-common-data/src/test/kotlin/net/corda/ledger/common/data/transaction/factory/WireTransactionFactoryImplTest.kt
```diff
@@ -3,17 +3,51 @@ package net.corda.ledger.common.data.transaction.factory
...
+            ),
+            privacySalt
+        )
+
+        assertThat(transaction1.id).isEqualTo(transaction2.id)
+        assertThat(transaction1.privacySalt).isEqualTo(transaction2.privacySalt)
+    }
+
+    @Test
+    fun `draft`() {
```
LWogan: guessing this is just some code for your local test purposes? leaving a comment to remind you to remove.

libs/ledger/ledger-common-data/src/test/kotlin/net/corda/ledger/common/data/transaction/factory/WireTransactionFactoryImplTest.kt
```diff
@@ -3,17 +3,51 @@ package net.corda.ledger.common.data.transaction.factory
...
 @Suppress("MaxLineLength")
 class WireTransactionFactoryImplTest : CommonLedgerTest() {
     private val metadata = transactionMetadataExample()
     private val metadataJson = jsonMarshallingService.format(metadata)
     private val canonicalJson = jsonValidator.canonicalize(metadataJson)
-    private val privacySalt = getPrivacySalt()
+    private val privacySalt = privacySaltProviderService.generatePrivacySalt()
+
+    @Test
+    fun `transaction ids are deterministic`() {
```
LWogan: lets add a test which generates different salts for different flow ids too

LWogan (commented):
Looks good, few comments

vlajos:
Can two privacy salts be requested while`FlowID+suspendCount` is the same?
Although probably it is not a real problem if that is not possible, since `toSignedTransaction()`s, where the privacy salt gets assigned, suspend anyway. (at least for the signing operation IIRC.)

emilybowe:
> Details

@vlajos I've checked with @LWogan and we would say no, for the reason you mention (because toSignTransaction will always suspend) and as long as we're always signing after creating transactions the suspendCount will increment.

vkolomeyko (dismissed):
LGTM
(on behalf of @corda/rest)

testing/ledger/ledger-common-base-test/src/main/kotlin/net/corda/ledger/common/test/CommonLedgerTest.kt
```diff
@@ -39,10 +40,12 @@ abstract class CommonLedgerTest {
...
 
     val wireTransactionFactory = WireTransactionFactoryImpl(
-        merkleTreeProvider, digestService, jsonMarshallingService, jsonValidator, cipherSchemeMetadata
+        merkleTreeProvider, digestService, jsonMarshallingService, jsonValidator
     )
 
-    private val flowFiberService = TestFlowFiberServiceWithSerialization(currentSandboxGroupContext)
+    val flowFiberService = TestFlowFiberServiceWithSerialization(currentSandboxGroupContext)
+
+    val privacySaltProviderService = PrivacySaltProviderServiceImpl(flowFiberService)
```
LWogan: Its not ideal to have to update a bunch of tests with the same mocks to get flowid and suspend count to work.  

You could make privacySaltProviderService a mock and have it return a salt. Then  you could remove the setups from the additional tests .

e.g
`val privacySaltProviderService = mock<PrivacySaltProviderService>().apply {
        whenever(generatePrivacySalt()).thenReturn(PrivacySaltImpl("someBytes".toByteArray()))
    }`

i didn't test the above, but something like that should work.  I see theres a few other mocks in that class such as mockSigningService() so it should be fine to do the same.

LWogan (commented):
Looks good. One test cleanup comment

libs/ledger/ledger-common-data/src/test/kotlin/net/corda/ledger/common/data/transaction/factory/WireTransactionFactoryImplTest.kt
```diff
@@ -3,25 +3,105 @@ package net.corda.ledger.common.data.transaction.factory
...
 
 @Suppress("MaxLineLength")
 class WireTransactionFactoryImplTest : CommonLedgerTest() {
     private val metadata = transactionMetadataExample()
     private val metadataJson = jsonMarshallingService.format(metadata)
     private val canonicalJson = jsonValidator.canonicalize(metadataJson)
-    private val privacySalt = getPrivacySalt()
+
+    @BeforeEach
+    fun setupPrivacySaltProviderService() {
```
LWogan: Sorry I missed this on my first pass but the tests to check that the PrivacySaltProviderService makes deterministic/unqiue ids based on flowid and suspend count should be in PrivacySaltProviderServiceImpTest.

This WireTransactionFactoryImplTest.kt should just be using a mocked PrivacySaltProviderService.

This WireTransactionFactoryImplTest was testing the PrivacySaltProviderServiceImpl.

privacySaltProviderService should be removed from CommonLedgerTest. We only want to use mockPrivacySaltProviderService there

LWogan (commented):
One more thing! Almost there.

LWogan (dismissed):
LGTM

blsemo (changes requested):
I don't think that approach is good. enough.
We need an iron clad guarantee that two distinct transactions have different ids, even if they're created in the same flow and have the same content.
And even if we always supend in the current implementation, this opens up an alley for surprising bugs in the future.

components/ledger/ledger-consensual-flow/src/main/kotlin/net/corda/ledger/consensual/flow/impl/transaction/factory/ConsensualSignedTransactionFactoryImpl.kt
```diff
@@ -50,20 +51,24 @@ class ConsensualSignedTransactionFactoryImpl @Activate constructor(
...
+    private val privacySaltProviderService: PrivacySaltProviderService
 ) : ConsensualSignedTransactionFactory, UsedByFlow, SingletonSerializeAsToken {
 
     /**
      * Creates the signedTransaction initially
      */
     @Suspendable
     override fun create(
-        consensualTransactionBuilder: ConsensualTransactionBuilder
+        consensualTransactionBuilder: ConsensualTransactionBuilder,
```
blsemo: Why was a trailing comma added here?

libs/ledger/ledger-common-data/src/main/kotlin/net/corda/ledger/common/data/transaction/factory/WireTransactionFactory.kt
```diff
@@ -5,17 +5,6 @@ import net.corda.ledger.common.data.transaction.WireTransaction
...
-     *
-     * @param componentGroupLists The component groups to include.
-     *
-     * @return A [WireTransaction].
-     */
-    fun create(componentGroupLists: List<List<ByteArray>>): WireTransaction
-
     /**
      * Creates a [WireTransaction] from the passed in [componentGroupLists] and [privacySalt].
      *
```
blsemo, line 10: The comment on this method is now wrong!

libs/ledger/ledger-common-data/src/main/kotlin/net/corda/ledger/common/data/transaction/factory/WireTransactionFactoryImpl.kt
```diff
@@ -35,24 +32,13 @@ class WireTransactionFactoryImpl @Activate constructor(
...
 
     private val metadataSchema: WrappedJsonSchema by lazy {
         jsonValidator.parseSchema(getSchema(TransactionMetadataImpl.SCHEMA_PATH))
     }
 
-    override fun create(componentGroupLists: List<List<ByteArray>>): WireTransaction {
-        val metadata =
-            parseMetadata(componentGroupLists[TransactionMetadataImpl.ALL_LEDGER_METADATA_COMPONENT_GROUP_ID].first())
-        check((metadata as TransactionMetadataInternal).getNumberOfComponentGroups() == componentGroupLists.size) {
-            "Number of component groups in metadata structure description does not match with the real number!"
```
blsemo, line 51: You removed this check on creating a new transaction.
Remember that the other function was only present to recreate  wire TX from the database, so has less stringent checks than one  exposed to a user facing API.

testing/ledger/ledger-common-base-test/src/main/kotlin/net/corda/ledger/common/test/CommonLedgerTest.kt
```diff
@@ -39,10 +42,13 @@ abstract class CommonLedgerTest {
...
     val wireTransactionFactory = WireTransactionFactoryImpl(
-        merkleTreeProvider, digestService, jsonMarshallingService, jsonValidator, cipherSchemeMetadata
+        merkleTreeProvider, digestService, jsonMarshallingService, jsonValidator
     )
 
-    private val flowFiberService = TestFlowFiberServiceWithSerialization(currentSandboxGroupContext)
+    val flowFiberService = TestFlowFiberServiceWithSerialization(currentSandboxGroupContext)
+
+    val mockPrivacySaltProviderService = mock<PrivacySaltProviderService>().apply {
+        whenever(generatePrivacySalt()).thenReturn(PrivacySaltImpl("someBytes".repeat(10).toByteArray())) }
```
blsemo: Does this mean that all transactions in all tests using this CommonTests as base will have the same privacy salt?

libs/ledger/ledger-common-data/src/main/kotlin/net/corda/ledger/common/data/transaction/factory/WireTransactionFactory.kt
LWogan: Why is this wrong?

emilybowe (commented):
changes look good @LWogan ! 👍

libs/ledger/ledger-common-data/src/main/kotlin/net/corda/ledger/common/data/transaction/factory/WireTransactionFactoryImpl.kt
LWogan: Whats the best recourse here? Do we need to keep an explicit create NEW transaction function or is it ok to just keep the check in the create function with the signature of :
```
override fun create(
        componentGroupLists: List<List<ByteArray>>,
        privacySalt: PrivacySalt
    )
    `

File: testing/ledger/ledger-common-base-test/src/main/kotlin/net/corda/ledger/common/test/CommonLedgerTest.kt (hunk 10 above)
Comment (LWogan) [reply testing/ledger/ledger-common-base-test/src/main/kotlin/net/corda/ledger/common/test/CommonLedgerTest.kt]: I've updated it to return random bytes now each time its called

File: libs/ledger/ledger-common-data/src/main/kotlin/net/corda/ledger/common/data/transaction/factory/WireTransactionFactoryImpl.kt (hunk 9 above)
Comment (LWogan) [reply libs/ledger/ledger-common-data/src/main/kotlin/net/corda/ledger/common/data/transaction/factory/WireTransactionFactoryImpl.kt:51]: Is there any negatives to having the check present for transactions loaded from the db. If the transaction had to pass the check to be created, does it change after that?
Would it only require updating some ledger tests to provide more realistic data. I put it in and saw a few ledger tests fail as a result.

Comment (LWogan) [reply libs/ledger/ledger-common-data/src/main/kotlin/net/corda/ledger/common/data/transaction/factory/WireTransactionFactoryImpl.kt:51]: I spoke to @lankydan and @vlajos. Neither were aware of any reason why we couldn't have this check in the transaction. I've adjusted  a handful of tests to supply the correct number of groups.

It does leave an open question why this code existed. So far noone knows why. Perhaps legacy code to simplify tests?

```
val completeComponentGroupLists = (0 until metadata.getNumberOfComponentGroups()).map { index ->
            componentGroupLists.getOrElse(index) { arrayListOf() }
        }

blsemo (approved):
I think that should do it.

LWogan (approved):

vkolomeyko (approved):

PR #5445, opened 2024-01-16
CORE-19243 Fix inbound session metric reporting

charlieR3:
Ensure inbound session message metrics are recorded per inbound session message rather than per any inbound message.

**Before:**
Screenshot 2024-01-17 at 09 27 11
Note the high number of session messages in this case. This is from the nightly large network tests using the latest `release/os/5.2` build.

**After:**
Screenshot 2024-01-17 at 09 39 41

This run of the large network tests used a build of this current PR. It was configured to run for a much shorter period of time but it still shows inbound session metrics being reported at a level which is more aligned with expectations.
For the record, these were the parameters used for the large network test run which produced this graph.
* -Dcorda.members.per.cluster.count=5 
* -Dcorda.flow.type=BILATERAL_SETTLEMENT 
* -Dcorda.verify.with.flows.during.network.setup=false 
* -Dcorda.flow.execution.duration=PT15M 
* -Dcorda.start.flow.frequency.in.milliseconds=1000 
* -Dcorda.number.of.flows.to.start.each.time=5 
* -Dcorda.number.of.warmup.flows=100


yift-r3 (approved):

yift-r3 (approved):
LGTM

PR #5446, opened 2024-01-16
CORE-13499: Implement key rotation status

anien:
**Changes**:
* store key rotation status in RekeyBusProcessor when we first get the key rotation request
* update key rotation status after each rewrap, update overall status if we rotated all keys in the current tenantId
* show user the resulting status of the key rotation (In Progress or Done) if the rotation ever started
* add check if the previous key rotation is in progress
* show user the timestamp of the last key rotation status, as the last key rotation might have happened long time ago

### DIFF
--- a/components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
+++ b/components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
@@ -159,19 +159,19 @@ class KeyRotationRestResourceImpl @Activate constructor(
                 .also { it.start() }
     }
 
-    override fun getKeyRotationStatus(rootKeyAlias: String): KeyRotationStatusResponse {
+    override fun getKeyRotationStatus(keyAlias: String): KeyRotationStatusResponse {
         tryWithExceptionHandling(logger, "retrieve key rotation status") {
             checkNotNull(stateManager)
         }
 
         val entries = stateManager!!.findByMetadataMatchingAll(
             listOf(
-                MetadataFilter("rootKeyAlias", Operation.Equals, rootKeyAlias),
+                MetadataFilter("rootKeyAlias", Operation.Equals, keyAlias),
                 MetadataFilter("type", Operation.Equals, "keyRotation")
             )
         )
         // if entries are empty, there is no rootKeyAlias data stored in the state manager, so no key rotation is/was in progress
-        if (entries.isNullOrEmpty()) throw ResourceNotFoundException("No key rotation for $rootKeyAlias is in progress.")
+        if (entries.isNullOrEmpty()) throw ResourceNotFoundException("No key rotation for $keyAlias is in progress.")
 
         var lastUpdatedTimestamp = Instant.MIN
         var rotationStatus = "Done"
@@ -188,7 +188,7 @@ class KeyRotationRestResourceImpl @Activate constructor(
             if (state.modifiedTime.isAfter(lastUpdatedTimestamp)) lastUpdatedTimestamp = state.modifiedTime
             if (state.metadata["status"] != "Done") rotationStatus = "In Progress"
         }
-        return KeyRotationStatusResponse(rootKeyAlias, rotationStatus, lastUpdatedTimestamp, result)
+        return KeyRotationStatusResponse(keyAlias, rotationStatus, lastUpdatedTimestamp, result)
     }
 
     override fun startKeyRotation(oldKeyAlias: String, newKeyAlias: String): ResponseEntity<KeyRotationResponse> {
--- a/components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
+++ b/components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
@@ -8,7 +8,6 @@ import net.corda.crypto.rest.response.KeyRotationResponse
 import net.corda.crypto.rest.response.KeyRotationStatusResponse
 import net.corda.crypto.rest.response.TenantIdWrappingKeysStatus
 import net.corda.data.crypto.wire.ops.key.rotation.KeyRotationRequest
-import net.corda.data.crypto.wire.ops.key.rotation.KeyRotationStatus
 import net.corda.data.crypto.wire.ops.key.rotation.KeyType
 import net.corda.data.crypto.wire.ops.key.status.UnmanagedKeyStatus
 import net.corda.libs.configuration.SmartConfig
@@ -34,6 +33,7 @@ import net.corda.messaging.api.publisher.config.PublisherConfig
 import net.corda.messaging.api.publisher.factory.PublisherFactory
 import net.corda.messaging.api.records.Record
 import net.corda.rest.PluggableRestResource
+import net.corda.rest.exception.ForbiddenException
 import net.corda.rest.exception.ResourceNotFoundException
 import net.corda.rest.messagebus.MessageBusUtils.tryWithExceptionHandling
 import net.corda.rest.response.ResponseEntity
@@ -77,9 +77,7 @@ class KeyRotationRestResourceImpl @Activate constructor(
 
     private var publishToKafka: Publisher? = null
     private var stateManager: StateManager? = null
-    private val serializer = cordaAvroSerializationFactory.createAvroSerializer<KeyRotationStatus>()
-    private val deserializer = cordaAvroSerializationFactory.createAvroDeserializer({}, KeyRotationStatus::class.java)
-    private val deserializer1 = cordaAvroSerializationFactory.createAvroDeserializer({}, UnmanagedKeyStatus::class.java)
+    private val deserializer = cordaAvroSerializationFactory.createAvroDeserializer({}, UnmanagedKeyStatus::class.java)
 
     override val targetInterface: Class<KeyRotationRestResource> = KeyRotationRestResource::class.java
     override val protocolVersion: Int = platformInfoProvider.localWorkerPlatformVersion
@@ -176,10 +174,11 @@ class KeyRotationRestResourceImpl @Activate constructor(
         // if entries are empty, there is no rootKeyAlias data stored in the state manager, so no key rotation is in progress
         if (entries.isNullOrEmpty()) throw ResourceNotFoundException("No key rotation for $rootKeyAlias is in progress.")
 
+        var lastUpdatedTimestamp = Instant.MIN
         var rotationStatus = "Done"
         val result = mutableListOf<Pair<String, TenantIdWrappingKeysStatus>>()
         entries.forEach { (key, state) ->
-            val keyRotationStatus = deserializer1.deserialize(state.value)
+            val keyRotationStatus = deserializer.deserialize(state.value)
             println("XXX: key: $key, state.key: ${state.key}, state.value: $keyRotationStatus")
             result.add(
                 state.metadata["rootKeyAlias"].toString() to TenantIdWrappingKeysStatus(
@@ -187,13 +186,15 @@ class KeyRotationRestResourceImpl @Activate constructor(
                     keyRotationStatus.rotatedKeys
                 )
             )
+            if (state.modifiedTime > lastUpdatedTimestamp) lastUpdatedTimestamp = state.modifiedTime
             if (state.metadata["status"] != "Done") rotationStatus = "In Progress"
         }
-
-        return KeyRotationStatusResponse(rootKeyAlias, result, rotationStatus)
+        return KeyRotationStatusResponse(rootKeyAlias, result, rotationStatus, lastUpdatedTimestamp)
     }
 
     override fun startKeyRotation(oldKeyAlias: String, newKeyAlias: String): ResponseEntity<KeyRotationResponse> {
+        if (!hasPreviousRotationFinished(oldKeyAlias)) throw ForbiddenException("Previous key rotation for $oldKeyAlias is in progress.")
+
         tryWithExceptionHandling(logger, "start key rotation") {
             checkNotNull(publishToKafka)
         }
@@ -203,6 +204,19 @@ class KeyRotationRestResourceImpl @Activate constructor(
             newKeyAlias,
             publishRequests = { publishToKafka!!.publish(it) }
         )
+    }
+
+    private fun hasPreviousRotationFinished(oldKeyAlias: String): Boolean {
+        stateManager!!.findByMetadataMatchingAll(
+            listOf(
+                MetadataFilter("rootKeyAlias", Operation.Equals, oldKeyAlias),
+                MetadataFilter("type", Operation.Equals, "keyRotation")
+            )
+        ).forEach {
+            // if we find one In Progress status, we know we are not done
+            if (it.value.metadata["status"] != "Done") return false
+        }
+        return true
     }
 }
 
@@ -222,11 +236,6 @@ fun doKeyRotation(
 ): ResponseEntity<KeyRotationResponse> {
     // We cannot validate oldKeyAlias or newKeyAlias early here on the client side of the RPC since
     // those values are considered sensitive.
-
-
-    // TODO: we should now be able to check if another key rotation is in progress by checking the state manager db
-    // TODO: we can possibly add status in the key rotation metadata and if any rotation for tenantId hasn't yet finished,
-    // it can say so.
 
     val requestId = UUID.randomUUID().toString()
     val keyRotationRequest = KeyRotationRequest(
--- a/components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
+++ b/components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
@@ -189,7 +189,7 @@ class KeyRotationRestResourceImpl @Activate constructor(
             if (state.modifiedTime > lastUpdatedTimestamp) lastUpdatedTimestamp = state.modifiedTime
             if (state.metadata["status"] != "Done") rotationStatus = "In Progress"
         }
-        return KeyRotationStatusResponse(rootKeyAlias, result, rotationStatus, lastUpdatedTimestamp)
+        return KeyRotationStatusResponse(rootKeyAlias, rotationStatus, lastUpdatedTimestamp, result)
     }
 
     override fun startKeyRotation(oldKeyAlias: String, newKeyAlias: String): ResponseEntity<KeyRotationResponse> {
--- a/components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
+++ b/components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
@@ -5,12 +5,17 @@ import net.corda.configuration.read.ConfigChangedEvent
 import net.corda.configuration.read.ConfigurationReadService
 import net.corda.crypto.rest.KeyRotationRestResource
 import net.corda.crypto.rest.response.KeyRotationResponse
+import net.corda.crypto.rest.response.KeyRotationStatusResponse
+import net.corda.crypto.rest.response.TenantIdWrappingKeysStatus
 import net.corda.data.crypto.wire.ops.key.rotation.KeyRotationRequest
 import net.corda.data.crypto.wire.ops.key.rotation.KeyRotationStatus
 import net.corda.data.crypto.wire.ops.key.rotation.KeyType
+import net.corda.data.crypto.wire.ops.key.status.UnmanagedKeyStatus
 import net.corda.libs.configuration.SmartConfig
 import net.corda.libs.configuration.helper.getConfig
 import net.corda.libs.platform.PlatformInfoProvider
+import net.corda.libs.statemanager.api.MetadataFilter
+import net.corda.libs.statemanager.api.Operation
 import net.corda.libs.statemanager.api.StateManager
 import net.corda.libs.statemanager.api.StateManagerFactory
 import net.corda.lifecycle.DependentComponents
@@ -29,6 +34,7 @@ import net.corda.messaging.api.publisher.config.PublisherConfig
 import net.corda.messaging.api.publisher.factory.PublisherFactory
 import net.corda.messaging.api.records.Record
 import net.corda.rest.PluggableRestResource
+import net.corda.rest.exception.ResourceNotFoundException
 import net.corda.rest.messagebus.MessageBusUtils.tryWithExceptionHandling
 import net.corda.rest.response.ResponseEntity
 import net.corda.schema.Schemas.Crypto.REKEY_MESSAGE_TOPIC
@@ -73,6 +79,7 @@ class KeyRotationRestResourceImpl @Activate constructor(
     private var stateManager: StateManager? = null
     private val serializer = cordaAvroSerializationFactory.createAvroSerializer<KeyRotationStatus>()
     private val deserializer = cordaAvroSerializationFactory.createAvroDeserializer({}, KeyRotationStatus::class.java)
+    private val deserializer1 = cordaAvroSerializationFactory.createAvroDeserializer({}, UnmanagedKeyStatus::class.java)
 
     override val targetInterface: Class<KeyRotationRestResource> = KeyRotationRestResource::class.java
     override val protocolVersion: Int = platformInfoProvider.localWorkerPlatformVersion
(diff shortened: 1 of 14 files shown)


components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -154,35 +154,78 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
     }
 
-    override fun getKeyRotationStatus(requestId: String): List<Pair<String, String>> {
+    @VisibleForTesting
+    fun initialiseStateManager(config: Map<String, SmartConfig>) {
+        val stateManagerConfig = config.getConfig(ConfigKeys.STATE_MANAGER_CONFIG)
+
+        stateManager?.stop()
+        stateManager = stateManagerFactory.create(stateManagerConfig).also { it.start() }
+        logger.debug("State manager created and started ${stateManager!!.name}")
```
dickon: nit: `stateManager?.name` is better here.

simon-johnson-r3: surprised it's even needed, stateManager couldn't be null here, usually Kotlin understands this, although maybe it requires an explicit blocking check to realise...

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -154,35 +154,78 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
         tryWithExceptionHandling(logger, "retrieve key rotation status") {
             checkNotNull(stateManager)
         }
 
-        val entries = stateManager!!.get(listOf(requestId))
-        val result = mutableListOf<Pair<String, String>>()
-        entries.forEach { entry ->
-            val keyRotationStatus = deserializer.deserialize(entry.value.value)!!
-            result.add(keyRotationStatus.oldParentKeyAlias to keyRotationStatus.requestId.toString())
+        val entries = stateManager!!.findByMetadataMatchingAll(
```
simon-johnson-r3: the convention is `stateManager!!` -> `checkNotNull(stateManager)` because it gives a better error message

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -154,35 +154,78 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
+            listOf(
+                MetadataFilter("rootKeyAlias", Operation.Equals, keyAlias),
+                MetadataFilter("type", Operation.Equals, "keyRotation")
+            )
+        )
+        // if entries are empty, there is no rootKeyAlias data stored in the state manager, so no key rotation is/was in progress
+        if (entries.isNullOrEmpty()) throw ResourceNotFoundException("No key rotation for $keyAlias is in progress.")
+
```
simon-johnson-r3, line 184: there are a lot of strings used in here, should they be defined somewhere?

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -154,35 +154,78 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
+        var lastUpdatedTimestamp = Instant.MIN
+        var rotationStatus = "Done"
+        val result = mutableListOf<Pair<String, TenantIdWrappingKeysStatus>>()
+        entries.forEach {
+            val state = it.value
+            val keyRotationStatus = deserializer.deserialize(state.value)!!
```
simon-johnson-r3: I think we need better handling of the null here, that's a platform error as I understand it and we should log this and make the end user aware something very bad has gone wrong

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -154,35 +154,78 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
+
         return doKeyRotation(
             oldKeyAlias,
             newKeyAlias,
             publishRequests = { publishToKafka!!.publish(it) }
         )
     }
+
+    private fun hasPreviousRotationFinished(oldKeyAlias: String): Boolean {
+        stateManager!!.findByMetadataMatchingAll(
```
simon-johnson-r3: checkNotNull

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -154,35 +154,78 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
+            listOf(
+                MetadataFilter("rootKeyAlias", Operation.Equals, oldKeyAlias),
+                MetadataFilter("type", Operation.Equals, "keyRotation")
+            )
+        ).forEach {
+            // if we find one In Progress status, we know we are not done
```
simon-johnson-r3: minor: the comment doesn't match the logic, which is the other way round (i.e. it doesn't check for an `In Progress` state). I think you can actually just remove the comment as the code is quite clear.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -154,35 +154,78 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
+        entries.forEach {
+            val state = it.value
+            val keyRotationStatus = deserializer.deserialize(state.value)!!
+            result.add(
+                state.metadata["tenantId"].toString() to TenantIdWrappingKeysStatus(
+                    keyRotationStatus.total,
+                    keyRotationStatus.rotatedKeys
+                )
+            )
+            if (state.modifiedTime.isAfter(lastUpdatedTimestamp)) lastUpdatedTimestamp = state.modifiedTime
```
simon-johnson-r3, line 198: I'm not sure about this line. Firstly I doubt `state.modifiedTime` would ever pre-date the beginning of 1970, and secondly no matter what `state.modifiedTime` was, I don't think here we should potentially try and change it.

I think passing `state.modifiedTime` directly into `KeyRotationStatusResponse` is fine?

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
```diff
@@ -154,35 +154,78 @@ class KeyRotationRestResourceImpl @Activate constructor(
...
+            if (state.metadata["status"] != "Done") rotationStatus = "In Progress"
```
simon-johnson-r3: the mapping between internal states and external states isn't really clear to me, do we need to do any translation? what internal state are we worried about propagating to the end user?

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/response/KeyRotationStatusResponse.kt
```diff
@@ -0,0 +1,31 @@
...
+    val rootKeyAlias: String,
+    val status: String,
+    val lastUpdatedTimestamp: Instant,
+    val wrappingKeys: List<Pair<String, TenantIdWrappingKeysStatus>>,
+)
+
+/**
+ * The key rotation status for wrapping keys per particular tenantId.
+ *
+ * @param total Total number of wrapping keys to be rotated for particular tenantId.
```
simon-johnson-r3, line 24: I don't think you meant `to be rotated` here. It's just the total number of wrapping keys for a particular `tenantId` I believe (i.e. in effect this number is unrelated to key rotation)?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -76,6 +83,37 @@ class CryptoRekeyBusProcessor(
...
+                                STATE_TYPE to status::class.java.name
+                            )
+                        )
+                    )
+                )
+            }
+
+            // Only delete previous key rotation status if we are actually going to rotate something
+            if (records.isNotEmpty()) deleteStateManagerRecords(request.oldParentKeyAlias)
+            stateManager!!.create(records)
```
simon-johnson-r3: checkNotNull

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -76,6 +83,37 @@ class CryptoRekeyBusProcessor(
...
+
+            // Group by tenantId/vNode
+            targetWrappingKeys.groupBy { it.first }.forEach {
+                logger.debug("Grouping wrapping keys by vNode/tenantId ${it.key}")
+                val status = UnmanagedKeyStatus(request.oldParentKeyAlias, it.value.size, 0)
+                records.add(
+                    State(
+                        // key is set as a unique string to prevent table search in re-wrap bus processor
+                        request.oldParentKeyAlias + it.key + "keyRotation",  // rootKeyAlias + tenantId + keyRotation
+                        serializer.serialize(status)!!,
```
simon-johnson-r3: checkNotNull

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -76,6 +83,37 @@ class CryptoRekeyBusProcessor(
                     wrappingRepo.findKeysWrappedByAlias(request.oldParentKeyAlias).map { wki -> tenantId to wki }
                 }
             }.flatten()
+
+            // First update state manager, then publish re-wrap messages, so the state manager db is already populated
+            val records = mutableListOf<State>()
+
+            // Group by tenantId/vNode
+            targetWrappingKeys.groupBy { it.first }.forEach {
```
simon-johnson-r3: the use of `it` on line 97 isn't that obvious, I assume it's tenantId, would be possible to add `{ tenantId ->` etc. here?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -88,38 +126,44 @@ class CryptoRekeyBusProcessor(
...
-            stateManager?.create(
-                listOf(
-                    State(
-                        request.requestId,
-                        flattened, 1,
-                        Metadata(mapOf(STATE_TYPE to status::class.java.name))
-                    )
-                )
+    private fun hasPreviousRotationFinished(oldKeyAlias: String): Boolean {
+        stateManager!!.findByMetadataMatchingAll(
```
simon-johnson-r3: checkNotNull

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -88,38 +126,44 @@ class CryptoRekeyBusProcessor(
...
+        ).forEach {
+            // if we find one In Progress status, we know we are not done
+            if (it.value.metadata["status"] != "Done") return false
         }
+        return true
+    }
 
-        return emptyList()
+    private fun deleteStateManagerRecords(oldParentKeyAlias: String) {
+        val toDelete = stateManager!!.findByMetadataMatchingAll(
```
simon-johnson-r3: checkNotNull

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -88,38 +126,44 @@ class CryptoRekeyBusProcessor(
...
+                MetadataFilter("rootKeyAlias", Operation.Equals, oldParentKeyAlias),
+                MetadataFilter("type", Operation.Equals, "keyRotation")
+            )
+        )
+        logger.info("Deleting following records ${toDelete.keys} for previous key rotation for rootKeyAlias $oldParentKeyAlias.")
+        val failedToDelete = stateManager.delete(
+            toDelete.values
+        )
+
+        // TODO should we throw the exception?
```
simon-johnson-r3: They have failed delete because they have been updated since you fetched them above. The pattern here should be to repeat this process until it stops failing. I would suggest having some reasonable retry count, and only after then should you throw. If we cannot delete these states I'd suggest we shouldn't be trying to rotate, not least because we rely on the states to block duplicate attempts.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
```diff
@@ -76,6 +83,37 @@ class CryptoRekeyBusProcessor(
...
+            val records = mutableListOf<State>()
+
+            // Group by tenantId/vNode
+            targetWrappingKeys.groupBy { it.first }.forEach {
+                logger.debug("Grouping wrapping keys by vNode/tenantId ${it.key}")
+                val status = UnmanagedKeyStatus(request.oldParentKeyAlias, it.value.size, 0)
+                records.add(
+                    State(
+                        // key is set as a unique string to prevent table search in re-wrap bus processor
+                        request.oldParentKeyAlias + it.key + "keyRotation",  // rootKeyAlias + tenantId + keyRotation
```
simon-johnson-r3: Instead of the comment, perhaps we should have a helper function to generate the key called from both places?

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRewrapBusProcessor.kt
```diff
@@ -14,20 +21,79 @@ private const val REWRAP_KEYS_OPERATION_NAME = "rewrapKeys"
...
                 cryptoService.rewrapWrappingKey(request.tenantId, request.targetKeyAlias, request.newParentKeyAlias)
             }
+
+            // Once re-wrap is done, we can update the state manager
+            var statusUpdated = false
+            while (!statusUpdated) {
+                // rootKeyAlias + tenantId + keyRotation is the unique key, therefore we don't need to do the table
+                // search through state manager
+                val tenantIdWrappingKeysRecords =
+                    stateManager!!.get(listOf(request.oldParentKeyAlias + request.tenantId + "keyRotation"))
```
simon-johnson-r3: checkNotNull

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRewrapBusProcessor.kt
```diff
@@ -14,20 +21,79 @@ private const val REWRAP_KEYS_OPERATION_NAME = "rewrapKeys"
...
+                    val newMetadata = if (deserializedStatus.total == deserializedStatus.rotatedKeys + 1) {
+                        updateMetadata(state.metadata, "status", "Done")
+                    } else {
+                        state.metadata
+                    }
+                    val failedToUpdate =
+                        stateManager.update(listOf(State(state.key, newValue!!, state.version, newMetadata)))
+                    if (failedToUpdate.isNotEmpty()) {
+                        logger.debug("Failed to update following states ${failedToUpdate.keys}, retrying.")
+                        Thread.sleep(Random.nextLong(0, 1000))
```
simon-johnson-r3: Hmm, not sure about this...

gradle.properties
```diff
@@ -82,6 +82,10 @@ picocliVersion = 4.7.3
 protonjVersion=0.34.1
 quasarVersion = 0.9.1_r3-SNAPSHOT
 reflectAsmVersion = 1.11.9
```
simon-johnson-r3: is this some sort of merge issue?

simon-johnson-r3 (changes requested):

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: 2 lines above this I have `checkNotNull(stateManager)`, it's part of  the `tryWithExceptionHandling`, but it still forces me to check on stateManager being not null.

gradle.properties
anien: It is.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: This was meant to get the latest modified time between all records in state manager as they are grouped by tenantId. So if one tenant Id was updated later than the other, with this, we would get that latest modified time.
Also, users might then compare when the rotation started (if they made the note of the start time) with when the last rotation has finished.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: I just thought that this gives us flexibility on how we are showing data to the user. And we can change it easily anytime.
If we use avro schema for returning it to the user, we would be tied to it and it would be harder to change. Also, current avro schema doesn't contain all the information we store in the metadata that we show to the user.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/response/KeyRotationStatusResponse.kt
anien: No, it does return the total number of keys that will be rotated for the tenant. Because tenant can have multiple wrapping keys, but not all of them must be wrapped by the current root wrapping key.

gradle.properties
anien: Hm, there doesn't seem to be a change, so not sure why github picked it up.

anien:
@simon-johnson-r3 I have addressed most of your comments, few more to go. PTAL.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: I have added private val for state manager checking if state manager is null. `!!` is gone.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: Sorted.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: I have created an objects with string values (I can't use `enum` in some places because `State's` `Metadata` doesn't support `enum`.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: I did `checkNotNull` here.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: Sorted.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: Comment removed.

components/crypto/crypto-rest/src/main/kotlin/net/corda/crypto/rest/impl/KeyRotationRestResourceImpl.kt
anien: I have added comment too.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
anien: Done.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
anien: I named lambda parameters.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
anien: Done.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
anien: Done.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
anien: I have created a helper function.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRewrapBusProcessor.kt
anien: Done.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
anien: Done.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRewrapBusProcessor.kt
anien: I have removed sleep. Elsewhere in the code when retries for state manager are happening, we don't have sleeps either.

components/crypto/crypto-service-impl/src/main/kotlin/net/corda/crypto/service/impl/bus/CryptoRekeyBusProcessor.kt
anien: I added retries. And if we cannot delete records, we won't even start new key rotation.

gradle.properties
anien: Reverted.

simon-johnson-r3 (approved):

vkolomeyko (approved):
LGTM from @corda/rest point of view

PR #5450, opened 2024-01-17
CORE-19009 Highest wrapping key generation

simon-johnson-r3:
Ensure that when signing keys are created the wrapping key with the highest generation count for the specified alias is always picked.

### DIFF
--- a/components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
+++ b/components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
@@ -15,7 +15,8 @@ import net.corda.crypto.core.parseSecureHash
 import net.corda.crypto.core.publicKeyIdFromBytes
 import net.corda.crypto.persistence.SigningKeyOrderBy
 import net.corda.crypto.persistence.SigningWrappedKeySaveContext
-import net.corda.crypto.persistence.WrappingKeyInfo
+import net.corda.crypto.persistence.db.model.SigningKeyEntity
+import net.corda.crypto.persistence.db.model.SigningKeyMaterialEntity
 import net.corda.crypto.persistence.db.model.WrappingKeyEntity
 import net.corda.crypto.softhsm.impl.SigningRepositoryImpl
 import net.corda.crypto.softhsm.impl.toDto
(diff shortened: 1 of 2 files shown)


Sakpal (commented):
LGTM

components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
```diff
@@ -103,15 +104,31 @@ class SigningRepositoryTest : CryptoRepositoryTest() {
...
-    private val wrappingKeys = mutableMapOf<EntityManagerFactory, WrappingKeyInfo>()
-    private fun saveWrappingKey(emf: EntityManagerFactory, alias: String): WrappingKeyInfo {
-        return wrappingKeys.computeIfAbsent(emf) {
+    data class UniqueAliasAndGeneration(
+        val emf: EntityManagerFactory,
+        val alias: String,
+        val generation: Int
+    )
+
+    private val uniqueWrappingKeys = mutableSetOf<UniqueAliasAndGeneration>()
```
anien: Is this used anywhere? We do add new alias+generation to it, but I can't see if we do anything with it further.

simon-johnson-r3: It's to make sure we never try and write the same wrapping key with the same alias and generation to the in memory db more than once. The framework does not clear the db between each test, but many tests call the same function with the same parameters to set up the wrapping key.

anien: Got it now. I can also see how it's used below. Saw the braces wrong!

components/crypto/crypto-softhsm-impl/src/integrationTest/kotlin/repository/SigningRepositoryTest.kt
```diff
@@ -103,15 +104,31 @@ class SigningRepositoryTest : CryptoRepositoryTest() {
...
+
+    private fun saveWrappingKey(
+        emf: EntityManagerFactory,
+        alias: String,
+        generation: Int = 1
+    ): UUID {
+        val uuid = UUID.randomUUID()
```
anien: We don't seem to be returning the existing uuid if the wrapping key already exists in the `uniqueWrappingKeys`

anien (approved):
LGTM.

PR #5452, opened 2024-01-17
CORE-19133: Change AssertWithRetry to store list of retries in failure exception

josephzunigadaly:
Because the list of retries is getting lost with concurrent test execution as there is only one stdout being shared among tests.

Also remove todo comment because the current state of the code is good.


vkolomeyko (approved):
LGTM

PR #5455, opened 2024-01-17
CORE-19017 migrate flow TimeoutEventCleanupProcessor to expect the MediatorState wrapper class

LWogan:
States saved by the mediator message pattern use a MediatorState wrapper object. The TimeoutEventCleanupProcessor does not currently expect this.

### DIFF
--- a/components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/TimeoutEventCleanupProcessor.kt
+++ b/components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/TimeoutEventCleanupProcessor.kt
@@ -13,6 +13,7 @@ import net.corda.messaging.api.records.Record
 import net.corda.utilities.debug
 import org.slf4j.LoggerFactory
 
+@Suppress("LongParameterList")
 class TimeoutEventCleanupProcessor(
     private val checkpointCleanupHandler: CheckpointCleanupHandler,
     private val stateManager: StateManager,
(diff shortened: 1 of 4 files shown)


ben-millar (dismissed):
LGTM

components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/TimeoutEventCleanupProcessor.kt
```diff
@@ -29,8 +32,10 @@ class TimeoutEventCleanupProcessor(
         val statesToRecords = stateManager.get(events.mapNotNull {
             it.value?.checkpointStateKey
         }).mapNotNull { (_, state) ->
-            avroDeserializer.deserialize(state.value)?.let {
-                state to generateCleanupRecords(it)
+            mediatorDeserializer.deserialize(state.value)?.let { mediatorState ->
+                checkpointDeserializer.deserialize(mediatorState.state.array())?.let {
+                    state to generateCleanupRecords(it)
+                }
```
jujoramos, line 38: This will likely flood the logs with lots of `WARN` messages registered by the `CordaAvroDeserializer` (see https://r3-cev.atlassian.net/browse/CORE-19295), and we probably don't want that. Aside from the above, we're now filtering the states that this processor can receive [here](https://github.com/corda/corda-runtime-os/blob/release/os/5.2/components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/FlowTimeoutTaskProcessor.kt#L74), can we use that to prevent receiving instances of `MediatorState`?.

LWogan: I did a manual test and there was no warns. The stateType metadata will remain the client supplied states objects class. Lookups based on this metadata will work. The actual real type of the serialized object will be `MediatorState ` which has a field `state`. This contains the client supplied bytes avro serialized. The MediatorState acts as a wrapper.

emilybowe (approved):
LGTM

LWogan:
This is breaking a lot of builds so im going to merge. Any further comments on the design can be addressed in followups

components/flow/flow-service/src/main/kotlin/net/corda/flow/maintenance/TimeoutEventCleanupProcessor.kt
jujoramos: 👍

